PR #766, opened 2015-01-19
HSEARCH-1767 Optimize delete by term

emmanuelbernard:
@Sanne , this only encompasses the optimization from the Hibernate Search metadata. But I need your advice, in particular on 3c273f5 as I had to create the visitor lazily. Let me know what you think.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/backend/impl/lucene/AbstractWorkspaceImpl.java
+++ b/engine/src/main/java/org/hibernate/search/backend/impl/lucene/AbstractWorkspaceImpl.java
@@ -6,12 +6,16 @@
  */
 package org.hibernate.search.backend.impl.lucene;
 
+import java.util.Collections;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicLong;
 
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.index.IndexWriter;
+
+import org.hibernate.search.cfg.spi.IdUniquenessResolver;
+import org.hibernate.search.engine.service.spi.ServiceManager;
 import org.hibernate.search.engine.spi.DocumentBuilderIndexedEntity;
 import org.hibernate.search.exception.impl.ErrorContextBuilder;
 import org.hibernate.search.indexes.impl.PropertiesParseHelper;
@@ -34,8 +38,8 @@ public abstract class AbstractWorkspaceImpl implements Workspace {
 	private static final Log log = LoggerFactory.make();
 
 	private final OptimizerStrategy optimizerStrategy;
-	private final Set<Class<?>> entitiesInIndexManager;
 	private final DirectoryBasedIndexManager indexManager;
+	private final ServiceManager serviceManager;
 
 	protected final IndexWriterHolder writerHolder;
 	private final boolean deleteByTermEnforced;
@@ -49,10 +53,10 @@ public abstract class AbstractWorkspaceImpl implements Workspace {
 	public AbstractWorkspaceImpl(DirectoryBasedIndexManager indexManager, WorkerBuildContext context, Properties cfg) {
 		this.indexManager = indexManager;
 		this.optimizerStrategy = indexManager.getOptimizerStrategy();
-		this.entitiesInIndexManager = indexManager.getContainedTypes();
 		this.writerHolder = new IndexWriterHolder( context.getErrorHandler(), indexManager );
 		this.indexMetadataIsComplete = PropertiesParseHelper.isIndexMetadataComplete( cfg, context );
 		this.deleteByTermEnforced = context.isDeleteByTermEnforced();
+		this.serviceManager = context.getServiceManager();
 	}
 
 	@Override
@@ -82,7 +86,10 @@ public abstract class AbstractWorkspaceImpl implements Workspace {
 
 	@Override
 	public Set<Class<?>> getEntitiesInIndexManager() {
-		return entitiesInIndexManager;
+		// Do not cache it as an IndexManager receiving a new type should return an updated list
+		// and will trigger a LuceneBackendResources rebuild and by side effect
+		// a new LuceneWorkVisitor which will need the new list
+		return Collections.unmodifiableSet( indexManager.getContainedTypes() );
 	}
 
 	@Override
@@ -107,12 +114,70 @@ public abstract class AbstractWorkspaceImpl implements Workspace {
 
 	@Override
 	public boolean areSingleTermDeletesSafe() {
-		return indexMetadataIsComplete && entitiesInIndexManager.size() == 1;
+		return indexMetadataIsComplete && getEntitiesInIndexManager().size() == 1;
 	}
 
 	@Override
 	public boolean isDeleteByTermEnforced() {
-		return deleteByTermEnforced;
+		// if artificially forced: go ahead
+		if ( deleteByTermEnforced ) {
+			return true;
+		}
+		// Optimize only if we have all the metadata
+		if ( indexMetadataIsComplete ) {
+			Set<Class<?>> entitiesInvolved = getEntitiesInIndexManager();
+			// a single entity is always safe
+			if ( entitiesInvolved.size() == 1 ) {
+				return true;
+			}
+			// ask the source of data for some extra info to be sure it's safe
+			IdUniquenessResolver idUniquenessResolver = this.serviceManager.requestService( IdUniquenessResolver.class );
+			try {
+				// Check all tuple of entities with one another with the following rules:
+				//
+				// either:
+				// 1. the id field name of one is not used as field name of the second
+				// 2. if they use JPA ids, the range of ids is shared for both entity types
+				//    i.e. two instances of either types are unique if and only if they have the same id value
+				for ( Class<?> firstEntity : entitiesInvolved ) {
+					DocumentBuilderIndexedEntity firstDocumentBuilder =
+							indexManager.getIndexBinding( firstEntity ).getDocumentBuilder();
+					String firstIdField = firstDocumentBuilder.getIdKeywordName();
+					boolean followingEntities = false;
+					for ( Class<?> secondEntity : entitiesInvolved ) {
+						// Skip all entities already processed and the same entity
+						if ( firstEntity == secondEntity ) {
+							followingEntities = true;
+						}
+						else if ( followingEntities ) {
+							//core of the validation rules
+							DocumentBuilderIndexedEntity secondDocumentBuilder =
+									indexManager.getIndexBinding( secondEntity ).getDocumentBuilder();
+							// both use JPA id and they share the same id uniqueness set
+							boolean uniqueIdEqualityMeansEntityEquality =
+									firstDocumentBuilder.getTypeMetadata().isJpaIdUsedAsDocumentId() &&
+									secondDocumentBuilder.getTypeMetadata().isJpaIdUsedAsDocumentId() &&
+									idUniquenessResolver.areIdsUniqueForClasses( firstEntity, secondEntity );
+							String secondIdField = secondDocumentBuilder.getIdKeywordName();
+							// id field collision?
+							boolean idFieldCollision =
+									firstDocumentBuilder.getTypeMetadata().getDocumentFieldMetadataFor( secondIdField ) != null ||
+									secondDocumentBuilder.getTypeMetadata().getDocumentFieldMetadataFor( firstIdField ) != null;
+							if ( !uniqueIdEqualityMeansEntityEquality && idFieldCollision ) {
+								return false;
+							}
+						}
+					}
+				}
+			}
+			finally {
+				this.serviceManager.releaseService( IdUniquenessResolver.class );
+			}
+			return true;
+		}
+		else {
+			return false;
+		}
 	}
 
 	@Override
--- a/engine/src/main/java/org/hibernate/search/backend/impl/lucene/AbstractWorkspaceImpl.java
+++ b/engine/src/main/java/org/hibernate/search/backend/impl/lucene/AbstractWorkspaceImpl.java
@@ -187,4 +187,9 @@ public abstract class AbstractWorkspaceImpl implements Workspace {
 	public void flush() {
 		getCommitPolicy().onFlush();
 	}
+
+	@Override
+	public String getIndexName() {
+		return this.indexManager.getIndexName();
+	}
 }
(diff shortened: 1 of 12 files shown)


Sanne:
Specifically about 3c273f5 : do we really need the visitor to be lazily initialized?
This mutations stuff is getting complex, the _LuceneBackendResources_ was meant to be "simple" and immutable.
I see I get an NPE in _isDeleteByTermEnforced()_ if I try reverting that commit as you can't access the list of known indexed entities yet, but I'm wondering if the problem would be solved when you address my comment at https://github.com/emmanuelbernard/hibernate-search/commit/3c273f53e04c47e67703766625d948fb6cd481c9#commitcomment-9346792 : you'd have the full list of types already provided, so you'd avoid the chicken-egg problem with the SearchFactory not being fully available yet.

emmanuelbernard:
@Sanne I am not sure I can avoid the laziness problem. Can I 100% ignore the list of @Indexed and only focus on the list of ORM entities? I guess that should be safe.
Also the current optimization works for all source events, not only ORM. But that's not an immediate concern today.

emmanuelbernard:
@Sanne another concern I have is that I am far from `SearchConfiguration` and:

1. either I would need to provide a static list of entities + subclasses that HSearch reasons about.
2. or I would receive an implementation of "E is not related to M or viceversa" from `SearchConfiguration`

I prefer 2 but it also means that I am dragging the ORM metadata for an unknown time. When / how to release the reference to be GCed?

engine/src/main/java/org/hibernate/search/backend/impl/lucene/AbstractWorkspaceImpl.java
```diff
@@ -107,12 +115,41 @@ public IndexWriter getIndexWriter(ErrorContextBuilder errorContextBuilder) {
 
 	@Override
 	public boolean areSingleTermDeletesSafe() {
-		return indexMetadataIsComplete && entitiesInIndexManager.size() == 1;
+		return indexMetadataIsComplete && getEntitiesInIndexManager().size() == 1;
 	}
 
 	@Override
 	public boolean isDeleteByTermEnforced() {
```
gunnarmorling: Could you add a description to `Workspace#isDeleteByTermEnforced()`? Specifically, I'm curious about the interaction of `areSingleTermDeletesSafe` and `isDeleteByTermEnforced`.

Sanne:
> Can I 100% ignore the list of @Indexed and only focus on the list of ORM entities? I guess that should be safe.

I think it would be the only safe option, as compared to ORM we only have a subset.

> Also the current optimization works for all source events, not only ORM. But that's not an immediate concern today.

I don't think we should generalize this to the non-ORM space - unless you implement the next suggestion 2 - as we're relying here on the specific understanding that in the ORM case the inheritance implies there is no ambiguity on the keys. A different engine might allow that, by using an alternative strategy. Infinispan for example is in a different camp: the optimisation is always valid and does not depend on the mutual inheritance relation among indexed types.

> 1. either I would need to provide a static list of entities + subclasses that HSearch reasons about.

It doesn't seem like a problem to me. Remember we'd be totally fine in limiting this optimisation to the case in which we don't allow for dynamic entity discovery.

> 2. or I would receive an implementation of "E is not related to M or viceversa" from SearchConfiguration

Nice one. What about you make a Service which has a function like 

    areIdValuesPotentiallyAmbiguous(Type A, Type B);

Looking at the requirements in 3c273f53e04c47e67703766625d948fb6cd481c9, such a function is all you really need from the types.

You'd not explicitly point to "inheritance rules" as that's ORM specific, so you could allow other engines to plug a different conceptual rule in theory. Also we might be able to cleanup the two booleans being exposed today, as Infinispan Query could simply implement this by returning _true_ for any request.

Actually such an _areIdValuesPotentiallyAmbiguous_ function would be awesome as it would allow us to easily implement the alternative optimisations I've mentioned on JIRA (those which I said we shouldn't focus on as they seem tricky) such as knowing that a Date Id type will never conflict with a Short Id type. Not something I'd want to address now, but it seems the abstraction of the function would go in the right direction.

>I prefer 2 but it also means that I am dragging the ORM metadata for an unknown time. When / how to release the reference to be GCed?

I don't think ORM actually GCs its metadata? Does it? I'm expecting it to need quite some metadata runtime, but I don't know. We're not making a copy, just referring to it so not consuming more memory.

engine/src/main/java/org/hibernate/search/backend/impl/lucene/AbstractWorkspaceImpl.java
emmanuelbernard: OK

emmanuelbernard: `areSingleTermDeletesSafe` should be named `isSingleClassInIndex`
`isDeleteByTermEnforced` should be named `isDeleteByTermEnforcedOrSafe`

gunnarmorling: > areSingleTermDeletesSafe should be named isSingleClassInIndex
> isDeleteByTermEnforced should be named isDeleteByTermEnforcedOrSafe

Yes, these names are much easier to grasp. I had figured the the semantics
after studying the code for a while, but those names would help.

emmanuelbernard:
@Sanne 

I've been thinking about it some more.
I can make it work with the existing `SearchConfiguration` contract and implementation of `SearchConfigurationFromHibernateCore` I can collect all the classes coming from `getClassMappings()`
and check that at least a superclass is present in that `Iterator` for each pair of classes in the index. So I don't really need a new contract areIdValuesPotentiallyAmbiguous.

Still that's based on assumption and only works for ORM as you point out. So adding `areIdValuesPotentiallyAmbiguous` or something approaching is a better idea. thoughts?

This method relies on keeping the ORM `Configuration` object around. That is not expected in the contract of ORM, that's the metadata waste I am mentioning. I don't even know if we keep around the `SearchConfiguration` implementation when we mutate the `SearchFactory`

Second point: The current algo only checking `@Indexed` classes is actually generic and goes beyond just ORM because I think we agreed that a hierarchy cannot have different `@DocumentId`s in a given class hierarchy. So the same ORM logic applies. Still want to give away this optimisation for the sake of non lazy visitors?

Sanne:
> Still that's based on assumption and only works for ORM as you point out. So adding areIdValuesPotentiallyAmbiguous or something approaching is a better idea. thoughts?

Yes I'd like that better too. Especially as it replaces the other two booleans in `SearchConfiguration`. I'm fine with an iterative approach but the risk is we never get to clean it up when it's not fresh in your mind.

> This method relies on keeping the ORM Configuration object around. That is not expected in the contract of ORM, that's the metadata waste I am mentioning. I don't even know if we keep around the SearchConfiguration implementation when we mutate the SearchFactory

I remember you took great care for that in old times, but I'm afraid that when we implemented the "dynamic reconfiguration" we dropped all such efforts to not keep references to ORM's configuration around as we really need it to "reboot".
It's tricky, I'm wondering how we could add a test to make sure we clean that object up, as keeping it always in mind isn't going to work.
Assuming that the damages wasn't done already - or that we might want to clean it up and with the current patch you don't want to make it harder - could you just make a copy into a hashset of all mapped entities? Such a thing could then implement the nice contract of _areIdValuesPotentiallyAmbiguous_ .

> Second point: The current algo only checking @Indexed classes is actually generic and goes beyond just ORM because I think we agreed that a hierarchy cannot have different `@DocumentId`s in a given class hierarchy.

I don't remember about that restriction. You have a reference to refresh my mind?
But even if they all use the same property being marked as `@DocumentId`, you could have entities {A, B, C}, where A is the parent of {B, C}, and only {B, C} are marked with `@Indexed`. If the engine component doesn't know about the existance of entity A, he can't figure out that {B, C} are not conflicting.

Sanne:
> Still want to give away this optimisation for the sake of non lazy visitors?

Of course not :) I'm not that strict on not wanting to change the lazy visitor. Just pushing to see if there is a better solution, as we'd be missing out on optimisations && incidentally keep the visitor simpler.

emmanuelbernard:
@Sanne and all. The PR is in good shape to be candidate for master. Review for inclusion.

Sanne:
Thanks! reviewing ..

orm/src/main/java/org/hibernate/search/cfg/impl/HibernateCoreIdUniquenessResolver.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Implementation extracting entity classes from the configuration
+ * and release the configuration reference for future Garbage Collection.
+ *
+ * @author Emmanuel Bernard &lt;emmanuel@hibernate.org&gt;
+ */
+public class HibernateCoreIdUniquenessResolver implements IdUniquenessResolver {
+	private final Set<Class<?>> entities;
+
+	public HibernateCoreIdUniquenessResolver(Configuration cfg) {
```
gunnarmorling: What happens if a user works with [shared indexes](https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#section-sharing-indexes)? Can we account for that here?

gunnarmorling: Ah, I guess that's covered by the call to `areSingleTermDeletesSafe()` before.

orm/src/main/java/org/hibernate/search/cfg/impl/HibernateCoreIdUniquenessResolver.java
```diff
@@ -0,0 +1,64 @@
...
+		 * That should be the root entity for that given class.
+		 */
+		Class<?> rootOfEntityInIndex = getRootEntity( entityInIndex );
+		Class<?> rootOfOtherEntityInIndex = getRootEntity( otherEntityInIndex );
+		return rootOfEntityInIndex == rootOfOtherEntityInIndex;
+	}
+
+	private Class<?> getRootEntity(Class<?> entityInIndex) {
+		if ( ! entities.contains( entityInIndex ) ) {
+			// the entity is not a mapped entity ?
```
gunnarmorling: Or even raise an `IllegalArgumentException`? Seems like a programming error if the class is not known?

Sanne: maybe, but even returning self is going to be safe enough.
Remember it's an optimisation, fall back to "slow mode" is always ok; I can't think of when this assumption could possibly fail, but let's assume there is an exotic environment in which it would fail, I'd rather have it go slow than have to debug that :)

gunnarmorling: Mh, dunno. I'd prefer "fail fast" so to analyze the reason rather than
silently live with this slowness.

Sanne: Good point.. I'll leave that to @emmanuelbernard if he wishes to polish on this, otherwise I'll pull and you can send a follow-up ;)

Sanne:
I'm done reviewing, it looks great and I'm tempted to merge right away.
Two micro suggestions are open though:
 - https://github.com/hibernate/hibernate-search/pull/766#commitcomment-9439634
 - https://github.com/hibernate/hibernate-search/pull/766#discussion_r23526939

Let me know if I should merge, or you want to address those.

Sanne:
mm, linking to the comments doesn't seem very helpful as there is no highlight of which one I mean.
I mean these points:
 - should we rather fail fast and throw `IllegalArgumentException` as in Gunnar's suggestion
 - the classname for `SubANonIndexed` (and `SubBNonIndexed`) seems odd.

orm/src/main/java/org/hibernate/search/cfg/impl/HibernateCoreIdUniquenessResolver.java
emmanuelbernard: This entire patch is about handling shared indexes :)
For a given physical index "index1", we look for all entity types contained and we try and see if they can safely be deleted via a single term query.

orm/src/main/java/org/hibernate/search/cfg/impl/HibernateCoreIdUniquenessResolver.java
emmanuelbernard: I think I prefer to return the entity itself rather than this exception. It is not inherently invalid for Hibernate Search to index non JPA entities.

We could add a log in the system to say whether or not we had to disable that optimization though when we build the visitor. That's probably a good idea. A user will spot the safe but slow mode and will be able to investigate.

emmanuelbernard:
> @gustavonalle you'll want to have a look at this, it's a better replacement for the booleans we added two weeks ago.

That is not correct @Sanne. We will be able to get rid of the `isDeleteByTermEnforced()` in a SPI breaking version. But I don't think it is safe to get rid of `isIndexMetadataComplete()`. If you look at `AbstractWorkspaceImpl.areSingleTermDeletesSafe()` you will see that it has to know whether the metadata is complete to know that a single entity in an index is safe. We could twist the `IdUniquenessResolver.areIdsUniqueForClasses` to encompass the metadata compete trick but I think that would be a bad idea.

emmanuelbernard:
> Why the name SubANonIndexed when it's actually @Indexed ?

I'll work on a better name.

emmanuelbernard:
Rebased
Change test entity names for @Sanne 
Added logs when the index cannot use the optimization for @gunnarmorling

Sanne:
I'm aware on the changes not being enough to replace the use case from Gustavo yet, not least because of our SPI contract. Just thinking ahead on what it could become.

orm/src/main/java/org/hibernate/search/cfg/impl/HibernateCoreIdUniquenessResolver.java
gunnarmorling: Yes, sorry, this was a brain fart ;) Considered to delete it but then I
thought I'll take your bashing :) The two-step algorithm and its weird
method names make it a bit tough to digest admittedly. Thanks for the
commenting btw. much appreciated.

emmanuelbernard: Yes it's quite complex and that concerns me a bit. That's why I went as far as commenting my code for once ;)

PR #767, opened 2015-01-20
HSEARCH-1324 Test combination Infinispan, JMS and JBoss Modules

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-1324

Is this good enough or do  you have some other use case in mind?


Sanne:
merged!

PR #768, opened 2015-01-21
HSEARCH-997: Adding numeric bridges for short/byte

gunnarmorling:
I've added a bridge for byte, too, for the sake of completeness.

### DIFF
--- a/documentation/src/main/asciidoc/ch04.asciidoc
+++ b/documentation/src/main/asciidoc/ch04.asciidoc
@@ -57,7 +57,7 @@ configure various aspects of the indexing process. Without `@Field` the property
 the indexing process.
 
 Hibernate Search tries to determine the best way to index your property. In most
-cases this will be as string, but for the types int, long, double and float (and their respective
+cases this will be as string, but for the types short, int, long, double and float (and their respective
 Java wrapper types) Lucene's numeric field encoding (see <<numeric-field-annotation>>) is used.
 This numeric encoding uses a so called link:http://en.wikipedia.org/wiki/Trie[Trie structure] which
 allows for efficient range queries and sorting, resulting in query response times
@@ -168,7 +168,7 @@ indexing of null values (see JavaDocs of `LuceneOptions.indexNullAs()`).
 ===== @NumericField
 
 `@NumericField` is a companion annotation to `@Field`. It can be specified in the
-same scope as `@Field`, but only on properties of numeric type like int, long, double and float
+same scope as `@Field`, but only on properties of numeric type like short, int, long, double and float
 (and their respective Java wrapper types). It allows to define a custom `precisionStep` for the
 numeric encoding of the property value.
 
@@ -188,7 +188,8 @@ numeric encoding of the property value.
                   value is 4.
 |===============
 
-Lucene supports the numeric types: `Double`, `Long`, `Integer` and `Float`. Other numeric types
+Lucene supports the numeric types: `Double`, `Long`, `Integer` and `Float`. For properties of type
+`Short`, an `Integer` field will be used in the index. Other numeric types
 should use the default string encoding (via `@Field`), unless the application can deal with a
 potential loss in precision, in which case a custom `NumericFieldBridge` can be used. See
 <<example-custom-numericfieldbridge>>.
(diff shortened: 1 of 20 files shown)


hferentschik:
Is this really a good idea? If I understand this change correctly, then byte and short will automatically be now indexed numerically. As @Sanne was saying, this involved quite some overhead and indexing as string might be better.

As of Search 5, `@NumericValue` is optional. The engine decides based on type, which type to index numerically. The default of using string encoding for short and byte seems correct.

I guess the question is, what to do, if a byte/short is explicitly annotated with `@NumericValue`?
* Ignore the annotation
* Throw a configuration error
* Only in this case index numerically

gunnarmorling:
>  If I understand this change correctly, then byte and short will automatically be now indexed numerically.

Yes, that's true. Indeed we may consider to use String-based indexing by default for byte and short if you guys think that makes more sense. OTOH it'd be kind of inconsistent to the other numeric types and thus may come at user's suprise.

Btw. how does one disable numeric indexing for int if it enabled by default now?

> I guess the question is, what to do, if a byte/short is explicitly annotated with @NumericValue?

IMO it should be indexed numerically. Why would one mark it as numeric otherwise?

hferentschik:
> Yes, that's true. Indeed we may consider to use String-based indexing by default for byte and short if you guys think that makes more sense.

I think so

> OTOH it'd be kind of inconsistent to the other numeric types and thus may come at user's suprise.

Well, the docs in this case describes the types which are numerically indexed. There is a reason why Lucene does not do it. In this context it would be interesting to have some performance measurements. If string indexing indeed performs better than numerical indexing, then it seems wrong to index numerically just for consistency. 

Btw, with the automatic encoding discovery of Search 5, the surprise comes really at query time where you have to choose the right query type for an indexed field. So for byte/short you need to use query types which target string encoded fields. There might indeed be some 'surprise'. However we have some error reporting around this (eg targeting a string field with a numeric query). And then there is the docs of course.

> Btw. how does one disable numeric indexing for int if it enabled by default now?

By explicitly setting a non numeric field bridge using `@FieldBridge`. It's in the docs. Maybe read it and let us know it it makes sense. This is quite new.

> IMO it should be indexed numerically. Why would one mark it as numeric otherwise?

true, true - the user add `@NumericField`, because he wants it to be indexed numerically. But why does he do it? Because he things if something is a number, it is "better" to index them as such. However, in this case this might just be wrong. But I guess, if you want to do things like sorting etc, it might make sense. AFAIR, an explicit `@NumericValue` is honored anyways at the moment. So indexing byte/short numerically, if explicitly requested is probably ok. I still think, however, we should stick to string encoding for the default behaviour.

gunnarmorling:
> If string indexing indeed performs better than numerical indexing, then it seems wrong to index numerically just for consistency. 

I'm not sure. Just when looking at a class which has several numeric properties, isn't reasonable to expect that all these numeric props are indexed as numbers (if it is doable of course), allowing number-based queries to run against them?

So in that light, if string-based indexing is faster, it seems like an optimization which one could opt-in if really needed.

> And then there is the docs of course.

Sure, but let's try to make the API work as one would reasonably expect it, so to avoid the need for checking with the docs in the first place.

Anyways, that's just my view as a newcomer to the field. If you folks say that using string-based fields for byte/short by default is more sensible, I'm all for it. Seems we agree on numeric indexing if `@NumericField` is given explicitly.

hferentschik:
> Sure, but let's try to make the API work as one would reasonably expect it, so to avoid the need for checking with the docs in the first place.

 I see your point. As said, to make a sounds decision, we would need to know/measure the actual overhead

> Seems we agree on numeric indexing if @NumericField is given explicitly.

+1

gunnarmorling:
As we agreed on IRC, using numeric fields for byte/short is generally a good idea for the sake of consistency and to avoid surprises to the user. Only we shouldn't do that change by default in a micro release in order to ensure index compatibility.

Added therefore some more commits which make byte/short mapped to String fields by default; Only if `@NumericField` is given, the new bridges will be used. There is [HSEARCH-1779](https://hibernate.atlassian.net/browse/HSEARCH-1779) to change the default behavior at the next suitable time.

@hferentschik, @Sanne: Can you give it another look? Thx!

gunnarmorling:
Btw. also added a bridge for encoding bytes in Strings, as it apparently was missing.

gunnarmorling:
@hferentschik Is there anything else you think should be done? Or is it good to go? Thx!

hferentschik:
> Is there anything else you think should be done? Or is it good to go? Thx!

Looking (did not see your addititional commit)

hferentschik:
rebased and merged

gunnarmorling:
Sweet, thx!

PR #770, opened 2015-01-21
Hsearch 1763 full text filter on mapped superclass

DavideD:
First draft for [HSEARCH-1763](https://hibernate.atlassian.net/browse/HSEARCH-1763)
It will look for @FullTextFilterDef(s) in the super classes of the indexed classes.

Not sure if it makes sense, let me know.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/spi/SearchIntegratorBuilder.java
+++ b/engine/src/main/java/org/hibernate/search/spi/SearchIntegratorBuilder.java
@@ -7,8 +7,6 @@
 
 package org.hibernate.search.spi;
 
-import java.beans.Introspector;
-import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -27,11 +25,9 @@ import org.hibernate.annotations.common.reflection.XClass;
 import org.hibernate.annotations.common.reflection.java.JavaReflectionManager;
 import org.hibernate.search.annotations.AnalyzerDef;
 import org.hibernate.search.annotations.AnalyzerDefs;
-import org.hibernate.search.annotations.Factory;
 import org.hibernate.search.annotations.FullTextFilterDef;
 import org.hibernate.search.annotations.FullTextFilterDefs;
 import org.hibernate.search.annotations.Indexed;
-import org.hibernate.search.annotations.Key;
 import org.hibernate.search.backend.impl.BatchedQueueingProcessor;
 import org.hibernate.search.backend.impl.QueueingProcessor;
 import org.hibernate.search.backend.impl.WorkerFactory;
@@ -64,7 +60,6 @@ import org.hibernate.search.exception.ErrorHandler;
 import org.hibernate.search.exception.SearchException;
 import org.hibernate.search.exception.impl.LogErrorHandler;
 import org.hibernate.search.filter.FilterCachingStrategy;
-import org.hibernate.search.filter.ShardSensitiveOnlyFilter;
 import org.hibernate.search.filter.impl.CachingWrapperFilter;
 import org.hibernate.search.filter.impl.MRUFilterCachingStrategy;
 import org.hibernate.search.indexes.impl.IndexManagerHolder;
@@ -74,7 +69,6 @@ import org.hibernate.search.spi.impl.SearchFactoryState;
 import org.hibernate.search.util.StringHelper;
 import org.hibernate.search.util.configuration.impl.ConfigurationParseHelper;
 import org.hibernate.search.util.impl.ClassLoaderHelper;
-import org.hibernate.search.util.impl.ReflectionHelper;
 import org.hibernate.search.util.logging.impl.Log;
 import org.hibernate.search.util.logging.impl.LoggerFactory;
 
@@ -310,7 +304,7 @@ public class SearchIntegratorBuilder {
 		ConfigContext configContext = new ConfigContext( searchConfiguration, buildContext, searchMapping );
 
 		initProgrammaticAnalyzers( configContext, searchConfiguration.getReflectionManager() );
-		initProgrammaticallyDefinedFilterDef( searchConfiguration.getReflectionManager() );
+		initProgrammaticallyDefinedFilterDef( configContext, searchConfiguration.getReflectionManager() );
 		final PolymorphicIndexHierarchy indexingHierarchy = factoryState.getIndexHierarchy();
 		final Map<Class<?>, EntityIndexBinding> documentBuildersIndexedEntities = factoryState.getIndexBindings();
 		final Map<Class<?>, DocumentBuilderContainedEntity> documentBuildersContainedEntities = factoryState.getDocumentBuildersContainedEntities();
@@ -355,8 +349,6 @@ public class SearchIntegratorBuilder {
 					documentBuildersContainedEntities.put( mappedClass, documentBuilder );
 				}
 			}
-			bindFilterDefs( mappedXClass );
-			//TODO should analyzer def for classes at their same level???
 		}
 
 		IndexManagerHolder indexesFactory = factoryState.getAllIndexesManager();
@@ -400,6 +392,8 @@ public class SearchIntegratorBuilder {
 				documentBuildersIndexedEntities,
 				documentBuildersContainedEntities
 		);
+
+		factoryState.setFilterDefinitions( configContext.initFilters() );
 		factoryState.setAnalyzers( configContext.initLazyAnalyzers() );
 	}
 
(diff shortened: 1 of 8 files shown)


engine/src/main/java/org/hibernate/search/spi/SearchIntegratorBuilder.java
```diff
@@ -403,6 +407,35 @@ else if ( metadataProvider.containsSearchMetadata( mappedClass ) ) {
...
+	 * if it's not annotated with @Indexed and it's not already in the collection.
+	 *
+	 * @param superClasses the list containing the super classes
+	 * @param mappedXClass the current class
+	 */
+	private void saveSuperClasses(Set<XClass> superClasses, XClass mappedXClass) {
+		if ( mappedXClass.getSuperclass() != null ) {
+			saveSuperClasses( superClasses, mappedXClass.getSuperclass() );
+		}
+		if ( !mappedXClass.isAnnotationPresent( Indexed.class ) && !OBJECT_CLASS_NAME.equals( mappedXClass.getName() ) ) {
```
hferentschik: why the condition of `!mappedXClass.isAnnotationPresent( Indexed.class )`? Why can it not be annotated with `@Indexed`. Filter definitions are global and the the only way to identify them is by their unique name. It does not matter where the definition is placed. For simplicity we always assumed that we just check the configured class itself, but now we want to consider the whole hierarchy. 

That said, there is a `ReflectionHelper.createXClassHierarchy( parseContext.getCurrentClass() );` which gives you the hierarchy as a list (bottom up). I think you could use that.

Last but not least, we are already processing the full hierarchy for indexed classes in `AnnotationMetadataProvider`. Now we do it twice as bootstrap. Once for the filter definitions and once when we process the rest of the configuration. Seems to be a waste. Technically there would be no problem to find the filter definitions in `AnnotationMetadataProvider` instead and only process the class hierarchies once.

engine/src/main/java/org/hibernate/search/spi/SearchIntegratorBuilder.java
```diff
@@ -324,6 +326,8 @@ private void initDocumentBuilders(SearchConfiguration searchConfiguration, Build
 		final org.hibernate.search.engine.metadata.impl.MetadataProvider metadataProvider =
 				new AnnotationMetadataProvider( searchConfiguration.getReflectionManager(), configContext );
 
+		bindFilterDefsInSuperClasses( classMappings );
```
hferentschik: See here you are processing now the filter definitions, after creating creating the `AnnotationMetadataProvider`. It could do the job of discovering the filter definitions. However, you would need to defer the retrieval of the filter definitions until `metadataProvider.getTypeMetadataFor( mappedClass )`, since the parsing of annotations is on request. See line 348 and 385. One would need to add the filter defs to `TypeMetadata`, which is maybe a bit fishy. `TypeMetadata` is supposed to hold all information related to a given type and it hierarchy, but the filer definitions are as mentioned type independent. That said, they are defined within this type hierarchy, so I guess we could add it. They need to be aggregated within `SearchIntegratorBuilder` then anyways.

engine/src/main/java/org/hibernate/search/spi/SearchIntegratorBuilder.java
DavideD: > why the condition of !mappedXClass.isAnnotationPresent( Indexed.class )? 

Initially, I didn't want to change the existing code and I kept the patch separated. The classes annotated with `@Indexed` are processed later so I decided to ignore them here. I will change this so that filters are processed in one part of the code only.

> That said, there is a ReflectionHelper.createXClassHierarchy( parseContext.getCurrentClass() ); which gives you the hierarchy as a list (bottom up). I think you could use that.

ok

> Last but not least, we are already processing the full hierarchy for indexed classes in AnnotationMetadataProvider. Now we do it twice as bootstrap. ... Seems to be a waste. Technically there would be no problem to find the filter definitions in AnnotationMetadataProvider instead and only process the class hierarchies once. 

I think I prefer the way it's done now, because the logic is going to be in one place. If I move it in the AnnotationMetadataProvider, I would be able to find the filters defined on one hierarchy but then I would have to check duplicates in `SearchIntegratorBuilder` anyway.
This way instead global elements are parsed in one place while type elements in another. I don't see it as waste unless there is reason to believe that this is going to become a performance bottleneck.

hferentschik: > Initially, I didn't want to change the existing code and I kept the patch separated. The classes annotated with @Indexed are processed later so I decided to ignore them here. I will change this so that filters are processed in one part of the code only.

Sounds good. One wonders otherwise what's going on. At least it should be commented.



>> That said, there is a ReflectionHelper.createXClassHierarchy( parseContext.getCurrentClass() ); which gives you the hierarchy as a list (bottom up). I think you could use that.

> ok

Cool

> I think I prefer the way it's done now, because the logic is going to be in one place. 

Really the metadata provider is where it should all happen and it contains the bulk of annotation processing

> but then I would have to check duplicates in SearchIntegratorBuilder anyway.

true, because the `SearchIntegrator` has the global view of all filters and analyzers. I see no contradiction here

> This way instead global elements are parsed in one place while type elements in another. I don't see it as waste unless there is reason to believe that this is going to become a performance bottleneck.

I agree, it is not a confirmed bottleneck, however, we used to be concerned about minimizing the amount of reflection calls. At least back in the days (right @emmanuelbernard !?). This will all change a bit with the introduction of Jandex, but that is still somewhere down the road.

Looking further into the code, I actually noticed that `@AnalyzerDef(s)` is also already handled in `AnnotationMetaDataProvider`. It uses `org.hibernate.search.engine.impl.ConfigContext` to keep track of the discovered analyzers definitions. The context also takes care of the de-duplication. From a configuration point of view analyzer and filter definitions are the same thing. IMO the approach should be aligned. If `ConfigContext` is used, there is even no need to do any de-duplication in `SearchIntegratorBuilder`.

DavideD: Sounds good,
I'll have a look at the ConfigContext and see what happens.

hferentschik: > I'll have a look at the ConfigContext and see what happens.

Thanks. I think now we are nailing it ;-)

Sanne:
How is this doing?

hferentschik:
> How is this doing?

Waiting for an update ...

DavideD:
@Sanne @hferentschik 
I've changed the code so that the logic for finding FullTextFilterDef is now in one place. I left everything in `SearchIntegratorBuilder` though. I've tried to move the logic in AnnotationMetadataProvider but it seems wrong since it only provides data relative to one class making it harder to detect duplicated filter names.

I think we should apply the same logic for `@AnalyzerDef`, I've created [HSEARCH-1781](https://hibernate.atlassian.net/browse/HSEARCH-1781) for it. I'll start to work on it when/if this pull is integrated.

hferentschik:
> I've changed the code so that the logic for finding FullTextFilterDef is now in one place. 

Ok

> I left everything in SearchIntegratorBuilder though. I've tried to move the logic in AnnotationMetadataProvider but it seems wrong since it only provides data relative to one class making it harder to detect duplicated filter names.

Interesting. I think it goes fine with `AnnotationMetadataProvider`, but I am fine if you think it should go into `SearchIntegratorBuilder`. However, in this case `AnalyzerDef` should be aligned. Conceptually, both definition types should behave the same way. The same "issues" exist around their discovery. Having to different ways of doing it seems wrong. So HSEARCH-1781 is really a must in this case imo.

orm/src/test/java/org/hibernate/search/test/filter/FilterTest.java
```diff
@@ -160,6 +161,16 @@ public void testMultipleFiltersOfSameTypeWithDifferentParameters() {
 	}
 
 	@Test
+	public void testFilterDefinedOnSuperClass() {
+		TermQuery query = new TermQuery( new Term( "employer", "Red Hat" ) );
+		FullTextQuery ftQuery = fullTextSession.createFullTextQuery( query, Employee.class );
+		ftQuery.enableFullTextFilter( "roleFilter" )
+				.setParameter( "role", Role.ADMINISTRATOR );
+
+		assertEquals( "Should find the filter defined in the super class", 1, ftQuery.getResultSize() );
```
hferentschik: I am not so font of this indirect testing. Really it should be enough to configure the filter and bootstrap Search. Afterwards one can just use `ExtendedSearchIntegrator#getFilterDefinition` to check whether the filter is there. Really that is what we want to confirm.

DavideD:
> However, in this case AnalyzerDef should be aligned. Conceptually, both definition types should behave the same way. The same "issues" exist around their discovery. 

I find weird that they use a different target time for the annotations, should not be the same?

* `@AnalyzerDef` has `@Target({ ElementType.TYPE, ElementType.FIELD, ElementType.METHOD })`

* `@FullTextFilterDef` has `@Target( { ElementType.TYPE } )`

hferentschik:
> I find it weird that they have different targets, should it not be the same?

Good catch. Odd indeed. It might be a cut&paste error from when creating `@AnalyzerDef`. I think `ElementType.TYPE` is really what it should be in both cases. In practice it does not really matter where you place the annotation, but given the 'global' aspect of these annotations, the type level just makes sense. I think this is what we mainly advocated.

One potential reason to have it this way, is if you define a analyzer def just for a single field and use it as this field as well. In this case, it might be nice to write:

    @Field
    @AnalyzerDef(name = "en",
		tokenizer = @TokenizerDef(factory = StandardTokenizerFactory.class),
		filters = {
			@TokenFilterDef(factory = LowerCaseFilterFactory.class),
			@TokenFilterDef(factory = SnowballPorterFilterFactory.class, params = {
			@Parameter(name = "language", value = "English")
    })
    @Analyzer(definition="en")
    public String myText

For a filter there is no such grouping possible, since filters apply at query time. 

Either way, we cannot reduce the target for `AnalyzerDef` and I guess for filters it just does not make sense to add new targets.

DavideD:
Ok, by the way, I think I was wrong about keep the code in SearchIntegratorBuilder, I just didn't understand  the analyzerDefinitionPoints role in ConfigContext. I'll change the code so that everything in the end is going to be consistent

hferentschik:
Ok

Sanne:
> Good catch. Odd indeed. It might be a cut&paste error from when creating @AnalyzerDef. I think ElementType.TYPE is really what it should be in both cases.

Odd indeed. I just checked, it seems it even works :)

> Either way, we cannot reduce the target for `AnalyzerDef`

I assume you mean we can't, because it would break API, right? I wonder though how much this is really used. When fixing a bug, in some cases it's always breaking APIs..
I wouldn't be too strict on this one, it's a subtle line. But yea let's keep it as-is since it doesn't do any harm to keep.

Sanne:
https://hibernate.atlassian.net/browse/HSEARCH-1782

DavideD:
In SearchIntegratorBuilder there is also a `initProgrammaticAnalyzers`, should it also exist a `initProgrammaticFilters`?

hferentschik:
> In SearchIntegratorBuilder there is also a initProgrammaticAnalyzers, should it also exist a initProgrammaticFilters?

There is right? At least in the original code there was `initProgrammaticAnalyzers` and `initProgrammaticallyDefinedFilterDef`

DavideD:
Yes, sorry

DavideD:
This should be fixed now, and the logic is consistent among the two annotations.
I'm not sure how to test the filters using `ExtendedSearchIntegrator`, if you guys can point me to an example I will add the test if it's needed

hferentschik:
> This should be fixed now, and the logic is consistent among the two annotations.

sweet

> I'm not sure how to test the filters using ExtendedSearchIntegrator, if you guys can point me to an example I will add the test if it's needed

An example would be `ErrorHandlerGivenAsInstanceTest`. Basically you would just configure the classes, bootstrap `SearchIntegrator` and then just verify that the filter god picked up and initialized via `ExtendedSearchIntegrstor#getFilterDefinition`. No need for indexing, searching and making assumptions based on the number of returned results. Apart from the fact that the result size could also depend on a million other things, the direct test is much easier to grasp. IMO we need to strife to get away from these end to end tests and write more focused unit tests. However, I think we can just move on for now and merge this. I have a final look...

hferentschik:
merged, thanks for your patience

PR #765, opened 2015-01-19
HSEARCH-1765 Add support for deleting by Query

s4ke:
this is not the final version, but a work in progress

PR #772, opened 2015-01-23
HSEARCH-1765 Add support for deleting by Query

s4ke:
a rebased version of the original Pull Request.


s4ke:
original can be found here: https://github.com/hibernate/hibernate-search/pull/765

s4ke:
@hferentschik @Sanne This is the new version of the pull request. I deleted the old branch while rebasing. Hope this is okay.

s4ke:
retest this please

engine/src/main/java/org/hibernate/search/backend/DeleteByQueryLuceneWork.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Representation of deleteByQuery(...) from Lucene. Currently not all functionality of Lucene is supported but can be
+ * added
+ *
+ * @author Martin Braun
+ */
+public class DeleteByQueryLuceneWork extends LuceneWork implements Serializable {
+
+	private static final long serialVersionUID = -5384359572337884952L;
```
Sanne: I know the IDE recommends having these.. but there are actually good reasons to not specify a serialVersionUID.
We decided to never decplare them.

Sanne: http://lists.jboss.org/pipermail/hibernate-dev/2008-May/003097.html

engine/src/main/java/org/hibernate/search/backend/DeleteByQuerySupport.java
```diff
@@ -0,0 +1,232 @@
...
+/**
+ * This class has means to convert all (by default) supported DeletionQueries back to Lucene Queries and to their
+ * String[] representation and back.
+ *
+ * @author Martin Braun
+ */
+public final class DeleteByQuerySupport {
+
+	private DeleteByQuerySupport() {
+		throw new AssertionError( "can't touch this!" );
```
Sanne: no need, just add a comment to clarify it's intentionally private.

engine/src/main/java/org/hibernate/search/backend/DeleteByQuerySupport.java
```diff
@@ -0,0 +1,232 @@
...
+									booleanQuery.add( new TermQuery( new Term( query.getFieldName(), value ) ), Occur.MUST );
+								}
+								return booleanQuery;
+							}
+							finally {
+								tokenStream.close();
+							}
+						}
+						catch (IOException e) {
+							throw new AssertionError( "no IOException can occur while using a TokenStream " + "that is generated via String" );
```
Sanne: If you really need such an exception, we'd better use `org.hibernate.search.exception.AssertionFailure`

engine/src/main/java/org/hibernate/search/backend/DeleteByQuerySupport.java
```diff
@@ -0,0 +1,232 @@
...
+	public static final Map<Integer, ToLuceneQuery> TO_LUCENE_QUERY_CONVERTER;
+	static {
+		{
+			Map<Integer, ToLuceneQuery> map = new HashMap<>();
+
+			map.put( SingularTermQuery.QUERY_KEY, new ToLuceneQuery() {
+
+				@Override
+				public Query build(DeletionQuery deletionQuery, ScopedAnalyzer analyzerForEntity) {
+					SingularTermQuery query = (SingularTermQuery) deletionQuery;
```
Sanne: I'm confused here. Why do you think it needs to be analyzed?

engine/src/main/java/org/hibernate/search/backend/DeleteByQuerySupport.java
```diff
@@ -0,0 +1,232 @@
...
+import org.hibernate.search.backend.SingularTermQuery.Type;
+import org.hibernate.search.util.impl.ScopedAnalyzer;
+
+/**
+ * This class has means to convert all (by default) supported DeletionQueries back to Lucene Queries and to their
+ * String[] representation and back.
+ *
+ * @author Martin Braun
+ */
+public final class DeleteByQuerySupport {
```
Sanne: this one doesn't belong in public API right? So it shouldn't be in this package.

engine/src/main/java/org/hibernate/search/backend/DeleteByQuerySupport.java
s4ke: oh. this was needed because of the CustomBehaviour Thingy. Will remove this.

engine/src/main/java/org/hibernate/search/backend/impl/lucene/works/DeleteByQueryWorkDelegate.java
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	@Override
+	public void performWork(LuceneWork work, IndexWriter writer, IndexingMonitor monitor) {
+		DeleteByQueryLuceneWork deleteWork = (DeleteByQueryLuceneWork) work;
+
+		final Class<?> entityType = work.getEntityClass();
+		final DeletionQuery query = deleteWork.getDeletionQuery();
+
+		log.tracef( "Removing all %s matching Query: %s", entityType.toString(), query.toString() );
```
Sanne: you should guard this tracef method with an "if (log.isTraceEnabled()) " to make sure those `toString()` methods don't get invoked in vain.

engine/src/main/java/org/hibernate/search/backend/impl/lucene/works/DeleteByQueryWorkDelegate.java
```diff
@@ -0,0 +1,95 @@
...
+
+		BooleanQuery entityDeletionQuery = new BooleanQuery();
+
+		// TODO: we could let experts supply their own means to convert back to
```
Sanne: I'd drop this all :)
If there's need to make things pluggable, it should not be too hard to figure out.

engine/src/main/java/org/hibernate/search/backend/impl/lucene/works/DeleteByQueryWorkDelegate.java
```diff
@@ -0,0 +1,95 @@
...
+		}
+
+		Term classNameQueryTerm = new Term( ProjectionConstants.OBJECT_CLASS, entityType.getName() );
+		TermQuery classNameQuery = new TermQuery( classNameQueryTerm );
+		entityDeletionQuery.add( classNameQuery, BooleanClause.Occur.MUST );
+
+		try {
+			writer.deleteDocuments( entityDeletionQuery );
+		}
+		catch (Exception e) {
```
Sanne: `deleteDocuments` only possibly throws an `IOException`. Don't catch more than that!

engine/src/main/java/org/hibernate/search/backend/impl/lucene/works/DeleteByQueryWorkDelegate.java
```diff
@@ -0,0 +1,95 @@
...
+			String message = String.format( "unable to remove all %s matching Query: %s", entityType.toString(), query.toString() );
+			throw new SearchException( message, e );
```
Sanne: our logger is able to throw nicely formatted errors too. Which is nice, as people might eventually internationalize them all..

For an example, see an entry in `Log` which returns a `SearchException`.
You have to throw the returned value from the logger.

engine/src/main/java/org/hibernate/search/backend/impl/DeleteByQuerySupport.java
```diff
@@ -0,0 +1,235 @@
...
+ *
+ * @author Martin Braun
+ */
+public final class DeleteByQuerySupport {
+
+	private DeleteByQuerySupport() {
+		throw new AssertionError( "can't touch this!" );
+	}
+
+	public static final Map<Integer, ToLuceneQuery> TO_LUCENE_QUERY_CONVERTER;
```
Sanne: I'm wondering here if we really want to keep the Map.
I understand it's an implementation level class, and that sometimes it's convenient to expose public fields. But being a public Map with an `Integer` key, that's not a nice contract for clients (even if it's impl).
I'd rather expose a better API, and internally use a more efficient array. That's if we really need to use the integer based key.

Looking at the different parse functions, it looks like you have to validate for a specific lenght of the array. Maybe a fully typesafe API is better?
You'd be able to stricly mandate the expected parameters at compile time.

engine/src/main/java/org/hibernate/search/engine/impl/WorkPlan.java
```diff
@@ -54,8 +57,8 @@
 	private final InstanceInitializer instanceInitializer;
 
 	/**
-	 * most work is split in two, some other might cancel one or more existing works,
-	 * we don't track the number accurately as that's not needed.
+	 * most work is split in two, some other might cancel one or more existing works, we don't track the number
```
Sanne: please revert these!

engine/src/main/java/org/hibernate/search/backend/impl/lucene/works/DeleteByQueryWorkDelegate.java
gunnarmorling: Or just pass the objects as is, i.e. without the `toString()`.

engine/src/main/java/org/hibernate/search/backend/impl/DeleteByQuerySupport.java
s4ke: About the parse functions:

These are only used for serialization right now. I kinda don't want these in the serialization modules because they are needed in both and I don't want to spread the DeleteByQuery logic in too many modules.

engine/src/main/java/org/hibernate/search/backend/DeleteByQueryLuceneWork.java
hferentschik: > http://lists.jboss.org/pipermail/hibernate-dev/2008-May/003097.html

true true

engine/src/main/java/org/hibernate/search/backend/impl/DeleteByQuerySupport.java
hferentschik: > I'd rather expose a better API

+1

engine/src/main/java/org/hibernate/search/engine/impl/WorkPlan.java
s4ke: done :).

hferentschik:
There seem to be some still some `AssertionErrors` thrown - http://ci.hibernate.org/job/hibernate-search-PR/667/console

Maybe it is better to close this pull request and create a new one from the latest working branch? WDYD? Sort of a clean slate.

PR #773, opened 2015-01-26
HSEARCH-1624 Allowing error handlers to be given as instances

gunnarmorling:


engine/src/main/java/org/hibernate/search/cfg/Environment.java
```diff
@@ -134,8 +134,11 @@
 	public static final String MODEL_MAPPING = "hibernate.search.model_mapping";
 
 	/**
-	 * Set to a fully qualified classname of a type implementing org.hibernate.search.exception.ErrorHandler
-	 * to override the error strategy used during processing of the Lucene updates.
+	 * Option for specifying an error handler used during processing of the Lucene updates. Supported value types are:
+	 * <ul>
+	 * <li>{@code String}: the fully qualified name of an {@link org.hibernate.search.exception.ErrorHandler} implementation</li>
```
hferentschik: nice, Proper `ul` and `li` :-)

engine/src/main/java/org/hibernate/search/spi/SearchIntegratorBuilder.java
```diff
@@ -571,16 +572,28 @@ private static String defineIndexingStrategy(SearchConfiguration cfg) {
...
+		else if ( configuredErrorHandler instanceof ErrorHandler ) {
+			return (ErrorHandler) configuredErrorHandler;
+		}
+		else {
+			throw log.unsupportedErrorHandlerConfigurationValueType( configuredErrorHandler.getClass() );
+		}
+	}
+
+	private ErrorHandler createErrorHandlerFromString(String errorHandlerClassName, ClassLoaderService classLoaderService) {
+		if ( StringHelper.isEmpty( errorHandlerClassName ) || "log".equals( errorHandlerClassName.trim() ) ) {
```
hferentschik: Should 'log' maybe be a constant, making it clear that it is the shortcut name of the logging error handler?

engine/src/main/java/org/hibernate/search/util/logging/impl/Log.java
```diff
@@ -736,4 +737,6 @@
 	@Message(id = 248, value = "WorkList should never be empty. Stacktrace below \n %s" )
 	void workListShouldNeverBeEmpty(String stackTrace);
 
+	@Message(id = 249, value = "Unsupported value type for configuration property " + Environment.ERROR_HANDLER + ": %1$s")
+	SearchException unsupportedErrorHandlerConfigurationValueType(@FormatWith(ClassObjectFormatter.class) Class<?> errorHandlerValueType);
```
hferentschik: Interesting, I have not seen `@FormatWith` yet. Not sure if you gain much here though!? I am bit confused about the 'Object' in `ClassObjectFormatter`? Would `ClassFormatter`, `ClassInstanceFormatter` or `LogClassFormatter` be better?

engine/src/main/java/org/hibernate/search/util/logging/impl/ClassObjectFormatter.java
```diff
@@ -0,0 +1,26 @@
...
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.util.logging.impl;
+
+/**
+ * Used with JBoss Logging to display class names in log messages.
+ *
+ * @author Gunnar Morling
+ */
+public class ClassObjectFormatter {
```
Sanne: Why do you need this?

orm/src/test/java/org/hibernate/search/test/errorhandling/ErrorHandlerGivenAsInstanceTest.java
```diff
@@ -0,0 +1,26 @@
...
+import org.hibernate.search.cfg.Environment;
+import org.hibernate.search.testsupport.TestForIssue;
+
+/**
+ * Test for making sure error handlers can also be passed as instances rather than via their FQN.
+ *
+ * @author Gunnar Morling
+ */
+@TestForIssue(jiraKey = "HSEARCH-1624")
+public class ErrorHandlerGivenAsInstanceTest extends ErrorHandlingDuringDocumentCreationTest {
```
hferentschik: I am not a big fan of these inheritance based tests. Hard to see what and how you tests things in this class and changes to `ErrorHandlingDuringDocumentCreationTest` might have side effects here. 

Also, from a testing POV, would it not be enough to just configure the error handler as you do here and then just boostrap and check that `SearchIntegrator#getErrorHandler` returns the right instance? A much more direct test imo.

hferentschik:
Looking good apart of the few comments I made. Up to you which ones you think are worth integrating.

engine/src/main/java/org/hibernate/search/util/logging/impl/Log.java
gunnarmorling: I like it because it gives a consistent "rendering" of types used in logs,
without the need for the calling code to do so.

For Class<?> objects the proposed impl returns the FQN (i.e. not "class
FQN" as Class#toString() does). Of course you only "gain" something when
using it across the board which is what we do in OGM. So we don't have to
remember at all call sites to consistently pass Class#getName() to a String
param but just use the class object and be done with it.

engine/src/main/java/org/hibernate/search/spi/SearchIntegratorBuilder.java
gunnarmorling: I guess so :)

orm/src/test/java/org/hibernate/search/test/errorhandling/ErrorHandlerGivenAsInstanceTest.java
gunnarmorling: Ok, that was the quickest to do but I can change it the way you describe.
Probably better :)

hferentschik: > Ok, that was the quickest to do but I can change it the way you describe. Probably better :)

thanks

engine/src/main/java/org/hibernate/search/util/logging/impl/Log.java
Sanne: I don't understand this. Class#toString() delegates to `getName()`, exactly the same. The only difference is it prefixes it with "interface " or "class" , which shouldn't really be a problem.

gunnarmorling: Hum, didn't expect this to cause a big discussion tbh.

Consider these other random examples from the same logger class:

* @Message(id = 46, value = "FieldCache was enabled on class %1$s but for
this type of identifier we can't extract values from the FieldCache: cache
disabled"))
* @Message(id = 125, value = "Interceptor enforces skip index operation
%2$s on instance of class %1$s")

Here "class" will be printed out twice in the rendered message.

* @Message(id = 156, value = "Cannot read %1$s field from a %2$s object :
does getter exists and is it public ?")

Here the class prefix reads odd.

* @Message(id = 134, value = "Unable to instantiate ClassBridge of type
%1$s defined on %2$s")

This one requires the calling code to call getName()

Etc. etc.

The formatter helps to avoid this sort of inconsistencies.

hferentschik: Since you are pointing these cases out so nicely. What's about fixing them as well using `@FormatWith`? I really dislike the printing of `class`.

Personally, I prefer to only pass strings to the log interface. `Class` might be a special case, but I want to also make sure that no custom types make it into the Log interface.

gunnarmorling: > Since you are pointing these cases out so nicely. What's about fixing
them as well using @FormatWith? I really dislike the printing of class.

I knew that would be coming over me ;)

Yes, I can do that.

Sanne: :+1: 

I didn't mean to make it a big discussion.. just curious.
I would object though about introducing pointless code like a `toString()` service :)
But it doesn't seem the case, I like your idea now that I'm not startled by the surprise of not having a clue about that annotation.

gunnarmorling:
Pushed some more commits to address your remarks.

hferentschik:
looking ...

engine/src/main/java/org/hibernate/search/util/logging/impl/Log.java
```diff
@@ -91,11 +92,11 @@
 
 	@LogMessage(level = WARN)
 	@Message(id = 25, value = "Service provider has been used but not released: %1$s")
-	void serviceProviderNotReleased(Class<?> class1);
+	void serviceProviderNotReleased(@FormatWith(ClassFormatter.class) Class<?> class1);
```
hferentschik: Thanks for fixing these as well

hferentschik: Now we have already one more than a single use of the formatter ;-)

engine/src/test/java/org/hibernate/search/test/errorhandling/ErrorHandlerGivenAsInstanceTest.java
```diff
@@ -0,0 +1,59 @@
...
+	public void canPassErrorHandlerInstanceToConfiguration() {
+		// given
+		ErrorHandler errorHandler = new MyErrorHandler();
+
+		// when
+		SearchConfiguration cfg = getSearchConfiguration( errorHandler );
+		SearchIntegrator searchIntegrator = new SearchIntegratorBuilder().configuration( cfg ).buildSearchIntegrator();
+
+		// then
+		assertSame( errorHandler, searchIntegrator.getErrorHandler() );
```
hferentschik: +1 I think this is much more to the point

PR #774, opened 2015-01-27
HSEARCH-1324 Integration test fixes

DavideD:
This pull request add some fixes to the inlcuded tests and add a new test case that uses Infinispan as second level cache with JMS


Sanne:
merged!
 - needed to expand PermGen on Java7
 - fixed a javadoc type

PR #779, opened 2015-01-30
HSEARCH-1791

gunnarmorling:
Hey @hferentschik, any chance you could check whether you can reproduce that one? As said on IRC, it creates a term range query for me, as expected.

What I've done here is the auto-analyzer disablement for ids.


hferentschik:
Looking ...

hferentschik:
Actually I am going to close this one for now. I started some work (so far just test changes) and somehow I think there might be something fishy going on after all. I'll re-open a new pull request later.

gunnarmorling:
Mh, I'd prefer we close a PR in such case only once there actually is another one superseding it. Afterall there is a value in it (also if possibly something else must be added) which I do not want to put at risk of getting lost.

hferentschik:
> which I do not want to put at risk of getting lost.

no worries, it will not be lost

PR #784, opened 2015-02-03
HSEARCH-1775: Avoiding references to engine "impl" classes from hibernate-search-infinispan

gunnarmorling:
Hi guys, could you check out this one? That's what I have done to get rid of the "impl" dependencies from hs-infinispan to hs-engine:

* Created an SPI service `LockFactoryCreator` for obtaining lock factories, which is consumable from within hs-infinispan
* Created an SPI helper class `DirectoryHelper` providing the required functionality for dealing with directories
* Duplicated some methods from `ConfigurationParseHelper` and `JNDIHelper` so hs-infinispan has these locally

My rationale was this:

* `LockFactoryCreator` and `DirectoryHelper` seem generally useful and stable, so it seems reasonable to commit to an SPI for them; The former is a service as it consumes another service itself (I disliked passing the service manager through a method parameter), the latter is a simple helper as it's sufficient in this case
* `ConfigurationParseHelper` and `JNDIHelper` seem awkward as general SPIs and we might not want to commit to them; As hs-infinispan is subject to relocation to ISPN anyways, I felt the duplication of these parts is the better choice (maybe there is even equivalent tuff in ISPN which may be used down the road)

I'm not 100% familiar with HSEARCH's packaging structure and philosophy, so if you have any better ideas for certain parts, please let me know and we can discuss. Thx!


engine/src/main/java/org/hibernate/search/cfg/Environment.java
```diff
@@ -218,6 +218,37 @@
...
+	 * <li><code>none</code></li>
+	 * <li>The fully-qualified name of a custom {@link org.hibernate.search.store.LockFactoryProvider} implementation.</li>
+	 * </ul>
+	 * Can be given globally or for specific indexes:
+	 * <ul>
+	 * <li><code>hibernate.search.default.locking_strategy=simple</code></li>
+	 * <li><code>hibernate.search.Books.locking_strategy=org.custom.components.MyLockingFactory</code></li>
+	 * </ul>
+	 */
+	public static final String LOCKING_STRATEGY = "locking_strategy";
```
hferentschik: I like that you add these to `Environment`. Not all properties are added there :-(

engine/src/main/java/org/hibernate/search/store/impl/DefaultLockFactoryCreator.java
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	@Override
+	public LockFactory createLockFactory(File indexDir, Properties dirConfiguration) {
+		//For FS-based indexes default to "native", default to "single" otherwise.
+		String defaultStrategy = indexDir == null ? "single" : "native";
+		String lockFactoryName = dirConfiguration.getProperty( Environment.LOCKING_STRATEGY, defaultStrategy );
+		if ( "simple".equals( lockFactoryName ) ) {
+			if ( indexDir == null ) {
+				throw new SearchException( "To use \"simple\" as a LockFactory strategy an indexBase path must be set" );
```
hferentschik: not via the JBoss logger?

hferentschik:
I like :-)

> Created an SPI service LockFactoryCreator for obtaining lock factories, which is consumable from within hs-infinispan

+1

> Created an SPI helper class DirectoryHelper providing the required functionality for dealing with directories

Why not a service?

> Duplicated some methods from ConfigurationParseHelper and JNDIHelper so hs-infinispan has these locally

+1 I think the simple solution of duplication would be applicable at other places as well (for example dependencies to our internal classes for the sake of avoiding some duplication). I think in some cases duplication is acceptable to keep other design goals (separation and isolation of modules). 

> I'm not 100% familiar with HSEARCH's packaging structure and philosophy, so if you have any better ideas for certain parts, please let me know and we can discuss

Looking good. Check out some of my comments.

engine/src/main/java/org/hibernate/search/cfg/Environment.java
gunnarmorling: Yes, I'm doing it as I come across them.

engine/src/main/java/org/hibernate/search/store/impl/DefaultLockFactoryCreator.java
gunnarmorling: I just took this code as is from the original class. I'll add a commit for using JBL.

engine/src/main/java/org/hibernate/search/cfg/Environment.java
hferentschik: +1

engine/src/main/java/org/hibernate/search/store/impl/DefaultLockFactoryCreator.java
hferentschik: Sure. What did you say yesterday? "Making the world a better place" ;-)

gunnarmorling: Done.

gunnarmorling:
Pushed an update based on your comments, @hferentschik. Only remaining this is the service discussion. I don't think we should do it, see my comment here: https://github.com/gunnarmorling/hibernate-search/commit/84439dcfda1fdd2f6f441da0fb8e1348c79d5b2a#commitcomment-9599030.

hferentschik:
I am happy to merge. @Sanne, do you have more/any objections?

Sanne:
So Lucene uses "locks" today. OK.. so we have a LockFactory. We make that pluggable as a LockFactoryFactory ... :-/ starts to sound like a joke already.
We then evolve even further into a "LockFactoryFactoryService" abstraction to override these? What's the point?

And this seems like very inconvenient:

        try {
			return serviceManager.requestService( LockFactoryCreator.class ).createLockFactory( indexDir, properties );
		}
		finally {
			serviceManager.releaseService( LockFactoryCreator.class );
 		}


guys I'm not happy about how everything is turning into a `Service`.
I tried to refactor the `ClassLoaderService` back in November and my PRs were refused as you all wanted it to be a `Service` but reality is that for an "aspect" like that one, the `Service` model isn't a good fit at all.

The `Service` contract was originally designed to solve some practical problems, let's please stick to that and not use it as a model to which all code needs to conform to.

Sanne:
> I am happy to merge.

Not happy with those Service mantra spreading all over without reason

hferentschik:
> And this seems like very inconvenient ...

So what's so terribly inconvenient here?

> guys I'm not happy about how everything is turning into a Service.

Are you not exaggerating a bit now? 

> The Service contract was originally designed to solve some practical problems, 

It is not just there, because of a "practical problem", it is a pattern which helps us decoupling things. A very useful thing to have.

> let's please stick to that and not use it as a model to which all code needs to conform to.

I don't understand what you mean here, but ok - what is your suggestion then?

gunnarmorling:
I also find the service stuff cumbersome, that's exactly why I did *not* use it everywhere.

I did it for `LockFactoryCreator` because it depends on another service (for class-loading). Injecting dependencies seems reasonable to me, better than passing them via method parameters which would make APIs just ugly to use. You also wouldn't do that in a "normal" Spring or CDI application. I did not create a service for the directory helper as it doesn't have any dependencies.

Also see my original comment:

> LockFactoryCreator and DirectoryHelper seem generally useful and stable [...] The former is a service as it consumes another service itself (I disliked passing the service manager through a method parameter), the latter is a simple helper as it's sufficient in this case

If the class loader stuff wasn't a service I'd be all for not having this service either.

> I tried to refactor the ClassLoaderService back in November and my PRs were refused as you all wanted it to be a Service

I for my part didn't refuse anything. I'd be all for a simpler model for injecting dependencies.

> The Service contract was originally designed to solve some practical problems

It would help if that purpose was stated somewhere (so something more than "Marker interface for services") giving advice when something should be a service and when not.

> let's please stick to that and not use it as a model to which all code needs to conform to.

Again, that's exactly not what has been done. Not sure what makes you say that.

Sanne:
merged, thanks all!

gunnarmorling:
Thanks!

PR #785, opened 2015-02-04
HSEARCH-1795

hferentschik:
So, here is comes the first step towards faceting - the ability to serialize `DocValues`. The change set seems large, but remember, I removed a whole module and needed to copy a lot of avro resources to create the 1.2 version of the protocol. Some more comments:

* I start of with a fix for the Java serialization module and then in the next commit remove the whole module. Obviously this could be squashed. However, this would also get rid of the fixes I already made to the protocol. I was just thinking that if we ever have a need to go back to this code, it would be nice to get back to a working version (at the time it was removed)
* Only the last commit is really making functional changes to the protocol. The others are java serialization related or refactoring of exiting code, especially the tests
* Last but not least, I am skeptical around this whole SPI. I just made the changes inline with the current approach. However, I am skeptical in particular around things like `LuceneWorkHydrator` and `LuceneWorksBuilder`. The whole protocol seems very fragile to change. On changes to the content of `LuceneWork` instances, the SPI classes need to change as well. IMO, the actual contract  is in `LuceneWorkSerializer`. That's all, that should be exposed. `LuceneWorkHydrator` and `LuceneWorksBuilder` put unnecessary constraints on 'how' you do the serialization (anyways, that was just a sidebar)


integrationtest/wildfly/src/test/java/org/hibernate/search/test/integration/VersionTestHelper.java
```diff
@@ -62,10 +65,20 @@ public static void addDependencyToSearchModule(Archive<?> archive, String... opt
...
-		return System.getProperty( "dependency.version.Lucene" );
+		return getRequiredSystemProperty( "dependency.version.Lucene" );
 	}
 
 	public static String getDependencyVersionHibernateCommonsAnnotations() {
-		return System.getProperty( "dependency.version.HibernateCommonsAnnotations" );
+		return getRequiredSystemProperty( "dependency.version.HibernateCommonsAnnotations" );
+	}
+
+	private static String getRequiredSystemProperty(String property) {
```
Sanne: Nice!

integrationtest/wildfly/src/test/java/org/hibernate/search/test/integration/jms/SearchNewEntityJmsMasterSlaveIT.java
```diff
@@ -41,21 +45,44 @@
...
-			libraryFiles.add( dependency( "org.hibernate:hibernate-search-serialization-java:" + currentVersion ) );
-			libraryFiles.add( dependency( "org.apache.lucene:lucene-core:" + VersionTestHelper.getDependencyVersionLucene() ) );
-			libraryFiles.add( dependency( "org.apache.lucene:lucene-analyzers-common:" + VersionTestHelper.getDependencyVersionLucene()) );
-			libraryFiles.add( dependency( "org.hibernate.common:hibernate-commons-annotations:" + VersionTestHelper.getDependencyVersionHibernateCommonsAnnotations() ) );
-			return libraryFiles.toArray( new File[0] );
+			libraryFiles.addAll( dependency( "org.hibernate:hibernate-search-orm:" + currentVersion, false ) );
+			libraryFiles.addAll( dependency( "org.hibernate:hibernate-search-engine:" + currentVersion, false ) );
+			libraryFiles.addAll( dependency( "org.hibernate:hibernate-search-backend-jms:" + currentVersion, false ) );
+			libraryFiles.addAll(
+					dependency( "org.hibernate:hibernate-search-serialization-avro:" + currentVersion, true )
```
Sanne: could you add a comment on why you need this to use Maven's "transitivity" ?

Sanne:
I only had a quick glance as it's very late, but have a question already:
Why deleting the `ProtocolVersion10Test` ?
By getting rid of the `persistent-avro-1.0` stored resource, we have no way of verifying that this version of Search is able to de-serialize what's written by an older version of Search. And we went to a great lenght of research and testing effort to make sure we could do that.
When bumping the protocol to 1.2, I would rather expect to see a new `persistent-avro-1.1` resource to be created *before* any changes to the protocol are made.

gunnarmorling:
The docs of `LuceneWork` still describe requirements for its serializability (and it implements `Serializable`). I understand that's not needed anymore with the removal of the Java serialization impl?

integrationtest/wildfly/src/test/java/org/hibernate/search/test/integration/jms/SearchNewEntityJmsMasterSlaveIT.java
hferentschik: i find that quite obvious, but I will add a comment

hferentschik:
> I only had a quick glance as it's very late, but have a question already: Why deleting the ProtocolVersion10Test ?

It is renamed to `ProtocolBackwardCompatibilityTest`. Also adding another test for 1.2 backward compatibility testing. I am now generating the "old" model on the fly (I did some refactoring in the serialization provider to make this possible). I has issues with the serialized file which triggered this refactoring. When I renamed one of the test helpers `ProtocolVersion10Test` suddenly broke. For some reason de-serializing this file required the test helper to be available. Something was fishy there.

> By getting rid of the persistent-avro-1.0 stored resource, we have no way of verifying that this version of Search is able to de-serialize what's written by an older version of Search. 

We do. At least we have a way to verify what is written out by earlier versions of the Avro protocol.

> When bumping the protocol to 1.2, I would rather expect to see a new persistent-avro-1.1 resource to be created before any changes to the protocol are made.

See `ProtocolBackwardCompatibilityTest`

This might create problems, once classes like the LuceneWorkHydrator changes in an incompatible way, but in this case you are out of luck with storing the serialized version as well. The whole framework is really fragile. So far we are "lucky" that we just need to make simple additions which are easy to handle from a backwards compatibility point of view.

gunnarmorling:
+1 for having tests to make sure we can read back messages in the old format versions.

Regarding the approach for dealing with evolution in general:

Do we actually need to keep around the old schema versions for that? E.g. here in 1.2 there is only a new possible element type for the `fieldables` array. I don't see a way how this change may cause incompatabilities when reading back 1.1 or 1.0 messages (they will never present this new type and it's also not that a type they may present has been removed in 1.2). For other kind of changes Avro allows for default values (in case new fields are added) or aliasing (in case field are renamed).

So it might suffice to only have the latest version of the schema present as long as it can be used to read back previous versions. That would avoid the need for building and keeping the different protocol builders. Also changes to the protocol will be more apparent in the history. Only drawback I see is that we couldn't create 1.0/1.1 test data on the fly. But one could use a binary stream created by an older HS version.

hferentschik:
> The docs of LuceneWork still describe requirements for its serializability (and it implements Serializable). I understand that's not needed anymore with the removal of the Java serialization impl?

It was needed before either. In the end even the native Java serialization needed to take the instances apart and create serializable parts. I will remove the comment and actually I think we can remove the implement clause as well. Good catch with the docs.

Sanne:
> I am now generating the "old" model on the fly 

If you create it "on the fly", it's hard to tell if the code which is generating it is going to respect the exact coding of what was produced in the *previous version*.
Especially in a large changeset like this one, I'm having a hard time to verify that there's no way possible for your changes to affect the format of the older encoding scheme?
Also, it might really not depend on our changes. Say we upgrade some dependency (like Avro), which changes the encoding ordering of some bytes.. an "on the fly" encoding wouldn't be able to tell us.

That's why we decided to actually commit in the repository a small - but hopefully representative - message. By simply verifying that
 - you don't delete that test ;-)
 - we never change a committed payload after a `.Final` release
we can have full peace of mind.

Let's remember this is way more tricky than keeping attention for API backwards compatibility; unfortunately it means maintenance for this area is quite complex, which is why we're helping ourselves with Avro rather than simply serializing some custom objects.

Sanne:
> So it might suffice to only have the latest version of the schema present as long as it can be used to read back previous versions. That would avoid the need for building and keeping the different protocol builders. Also changes to the protocol will be more apparent in the history. Only drawback I see is that we couldn't create 1.0/1.1 test data on the fly. But one could use a binary stream created by an older HS version.

Interesting idea. Indeed as I said above, it gives us far better reassurances to commit the binaries as generated in each protocol version, so the drawback of your proposal doesn't really apply.
It's damn hard though to proof that your new version of the Avro schema is able to read the old schema.. ProtoBuf is better in this field, as it provides tooling to verify this.
You're right though that - provided our "committed payload" is covering it all - we might not need the old schemas around.

Sanne:
> The docs of LuceneWork still describe requirements for its serializability (and it implements Serializable).

This was actually used in the past, before we introduced the serialization services, we'd use straight serialization of the `LuceneWork` types. Which is why the constructors are bit funky: they needed to support transfers of various kinds of payloads, but we had a very strict rule about not making changes to the class at all, so to keep the backwards serialization protocol standing ... poor man's Avro :)

hferentschik:
> Do we actually need to keep around the old schema versions for that? ...

Technically probably not. As you say, we could just get away with a single version.

> That would avoid the need for building and keeping the different protocol builders. 

The overhead seems minimal to me. Sure, it is not DRY, but still straight forward

>  Only drawback I see is that we couldn't create 1.0/1.1 test data on the fly. 

Right, that is indeed a drawback imo. This way it is very easy to test backward compatibility. 

> But one could use a binary stream created by an older HS version.

I got burned there with the changes I made now. Also it seems cumbersome to get this binary files.

hferentschik:
>  it gives us far better reassurances to commit the binaries as generated in each protocol version

provided the correctness of the binaries. As said, I suspect there is something wrong with the version we had. 

Anyways, what are you guys saying here? You want a single protocol version and binary test data? I started added some more javadocs and changed the one from LuceneWork (also removing Serializable). Depending on what you guys thing, I will then also refactor the test.

Sanne:
> provided the correctness of the binaries. As said, I suspect there is something wrong with the version we had.

I see.. but how do you know? It's getting tricky with this large changeset to figure it out. On the other hand, we know that that binary blob was working fine so far, so it's also possible that the "storage approach" is proofing itself.

And alternative is you download an Hibernate Search 4.4 distribution, have it send some messages over JMS, and receive them on a master node running your PR.. isn't that super inconvenient?
Of course I'm not asking you to do that, but that should give the picture of what that test was supposed to do.

If we had strict tests in place, I'd not need to investigate the split-hair of your code trying to figure if you thought about every detail, and it should be easier to move on.

hferentschik:
>  but how do you know? It

I know, because when I tried to deserialize the old binary file after just renaming `AvroTestHelper` I got an exception, saying that `AvroTestHelper` could not be loaded. That made me think, that there was more in this file as we thought.

> On the other hand, we know that that binary blob was working fine so far, so it's also possible that the "storage approach" is proofing itself.

What?

> If we had strict tests in place, I'd not need to investigate the split-hair of your code trying to figure if you thought about every detail, and it should be easier to move on.

You realize the changes to the actual serialization code are minimal. I added a couple of methods to the SPI and added a few new types to the Avro protocol. I agree the test classes and names have changes, but that is hardly rocker science. 

So bottom line, what do you think we should do? Three options:

* Keep as is (keep the full copies of the protocols and do testing of backward compatibility on the fly
* Keep the full copies of the protocol, but use binary files checked into version control to make the compatibility test
* Keep only the latest Avro protocol which will force us to use the binary testing approach as a consequence

hferentschik:
Pushed a new version with some more docs and changes to LuceneWork.

hferentschik:
> All subclasses of LuceneWork also implement Serializable btw.

should probably be removed as well

s4ke:
> should probably be removed as well

see https://hibernate.atlassian.net/browse/HSEARCH-1787?filter=-2

hferentschik:
Ok, yet one more version. I went back to the storing of a serialized version on disk for the compatibility test. 

There are still all three schema versions around. I actually tried to remove all except the latest, but things are not so simple. Schema evolution with Avro is a prickly beast and  I am not sure if you in the end not still need the original writer schemas (that's what I somehow read out of some relevant SO posts). ATM, I don't think this is worth perusing.

Sanne:
> Ok, yet one more version. I went back to the storing of a serialized version on disk for the compatibility test.

Huge thanks. I realise it's big work, but seems to be the safest approach.

> There are still all three schema versions around. I actually tried to remove all except the latest, but things are not so simple. Schema evolution with Avro is a prickly beast and I am not sure if you in the end not still need the original writer schemas (that's what I somehow read out of some relevant SO posts). ATM, I don't think this is worth perusing.

+1

hferentschik:
@Sanne merge!?

Sanne:
I definitely want this in. But it's huge, and I'm struggling a bit.
I'm going to split it up, and apply the straight-forward part already, for example the last commit is a no-brainer and is nicely separated from the other commits :-)

https://hibernate.atlassian.net/browse/HSEARCH-1805

hferentschik:
> I definitely want this in. But it's huge, and I'm struggling a bit.

Honestly, it not so huge at all. We discussed most of it already. @gunnarmorling, had a pretty good look. First two commits just ignore. Java serialization is gone anyways. The actual DocValue changes just look at the actual production code changes. The rest is test refactoring. Not sure where you are struggling. I think you make it harder than it has to be.

gunnarmorling:
Yes, please let's move forward with this and not leave it lying around until the next HS iteration. It will only be harder to remember about all the details. Let us know if there is anything to help you to get a better feeling. I also can take another closer look at the change.

Sanne:
@gunnarmorling no worries, the week isn't over yet :waning_gibbous_moon:
Hardy just gave me the last details I needed.

Sanne:
merged!

PR #786, opened 2015-02-04
HSEARCH-1774 Removing lucene-sandbox from features.xml

gunnarmorling:


integrationtest/osgi/karaf-it/src/test/java/org/hibernate/search/test/integration/osgi/HibernateSearchWithKarafIT.java
```diff
@@ -86,6 +92,13 @@
 				.type( "xml" )
 				.versionAsInProject();
 
+		MavenUrlReference hibernateSearchFeature = maven()
```
hferentschik: nice

integrationtest/osgi/karaf-it/src/test/java/org/hibernate/search/test/integration/osgi/HibernateSearchWithKarafIT.java
```diff
@@ -138,6 +148,24 @@ public void tearDown() {
 		bundleContext.ungetService( serviceReference );
 	}
 
+	/**
+	 * Starts all bundles to make sure there are no "uses constraint violations" caused by packages exported from the
+	 * bundles contributed by the HSEARCH Karaf feature.
+	 */
+	@Test
+	@TestForIssue(jiraKey="HSEARCH-1774")
+	public void testStartingAllBundles() throws Exception {
```
hferentschik: Interesting. So the problem was that not all bundles are started by default. Nice test.

integrationtest/osgi/karaf-it/src/test/java/org/hibernate/search/test/integration/osgi/HibernateSearchWithKarafIT.java
```diff
@@ -138,6 +148,24 @@ public void tearDown() {
...
+		for ( Bundle bundle : bundleContext.getBundles() ) {
+			if ( !isFragmentBundle( bundle ) ) {
+				bundle.start();
+			}
+		}
+	}
+
+	private boolean isFragmentBundle(Bundle bundle) {
```
hferentschik: what's a fragment bundle?

pom.xml
```diff
@@ -37,6 +37,7 @@
         <module>integrationtest/wildfly</module>
         <module>integrationtest/performance</module>
         <module>integrationtest/osgi/karaf-features</module>
+        <module>integrationtest/osgi/karaf-it</module>
```
hferentschik: why did you do this? There is a profile 'integration-karaf' which gets activates for JDK 7 and 8. The test does not work on JDK 9, but Sanne still wants to build against it, so he added this additional profile. Admittedly one can easily get lost with all the profiles defined in this pom. I would not mind a clean-up (in particular removing the db profiles, but that's a different story).

If you are ok with it,  I remove this line and merge. Other than that all looks good.

integrationtest/osgi/karaf-it/pom.xml
```diff
@@ -50,6 +50,13 @@
 
         <dependency>
             <groupId>${project.groupId}</groupId>
+            <artifactId>hibernate-search-engine</artifactId>
```
hferentschik: That's for getting `@TestForIssue` right?

integrationtest/osgi/karaf-features/src/main/features/features.xml
```diff
@@ -21,7 +21,7 @@
         <bundle start-level="30">mvn:org.hibernate.javax.persistence/hibernate-jpa-2.1-api/1.0.0.Final</bundle>
 
         <!-- JBoss Logging and Hibernate Commons Annotations -->
-        <bundle>mvn:org.jboss.logging/jboss-logging/3.1.0.GA</bundle>
+        <bundle>mvn:org.jboss.logging/jboss-logging/${jbossLoggingVersion}</bundle>
```
hferentschik: nice

integrationtest/osgi/karaf-it/src/test/java/org/hibernate/search/test/integration/osgi/HibernateSearchWithKarafIT.java
gunnarmorling: > what's a fragment bundle?

It's a bundle which does not live on its own but attaches to another bundle
(the "fragment host") and makes its contents available to that host (see
http://wiki.osgi.org/wiki/Fragment). So it's a way to add contents to
another bundle. I'm excluding them here as they cannot be started on their
own.

Fragments may actually be useful for us if e.g. engine is a fragment host
and other HS modules are fragments of that. That way they can use e.g.
internal stuff from engine. I didn't go for this with the ISPN module
though as it's subject to extraction from the code base anyways.

pom.xml
gunnarmorling: Ah, I see. I thought it'd be missing, indeed didn't see the profile
(wouldn't it make sense to add the actual feature definition to that
profile then as well; seems less confusing). So feel free to remove that
one.

integrationtest/osgi/karaf-it/pom.xml
gunnarmorling: Yes, right.

integrationtest/osgi/karaf-it/src/test/java/org/hibernate/search/test/integration/osgi/HibernateSearchWithKarafIT.java
hferentschik: > It's a bundle which does not live on its own but attaches to another bundle (the "fragment host")

Ahh, something learned here. I did not know about this. Thanks.

> Fragments may actually be useful for us if e.g. engine is a fragment host and other HS modules are fragments of that. 

Interesting indeed. This could have been useful for engine + orm. Well, on the other hand it is nice to have a "clean" package separation and no impl dependencies.

hferentschik:
rebased and merged

hferentschik:
@gunnarmorling The CI server does not seem to be happy about the change. Any idea why? I wipe the workspace and trigger one more build. On my machine the build works fine.

gunnarmorling:
Yes, saw it already. Looking into it. I've got a vague idea but need
further investigation.

PR #788, opened 2015-02-04
HSEARCH-1798 EAR

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-1798

EAR deployment integration tests

### DIFF
--- a/integrationtest/wildfly/src/test/java/org/hibernate/search/test/integration/wildfly/ModuleMemberRegistrationEarArchiveWithJbossDeploymentIT.java
+++ b/integrationtest/wildfly/src/test/java/org/hibernate/search/test/integration/wildfly/ModuleMemberRegistrationEarArchiveWithJbossDeploymentIT.java
@@ -1,0 +1,113 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.test.integration.wildfly;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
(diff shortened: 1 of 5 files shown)


integrationtest/wildfly/src/test/java/org/hibernate/search/test/integration/wildfly/ModuleMemberRegistrationEarArchiveWithJbossDeploymentIT.java
```diff
@@ -0,0 +1,113 @@
...
+import org.jboss.shrinkwrap.api.spec.EnterpriseArchive;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.jboss.shrinkwrap.descriptor.api.Descriptors;
+import org.jboss.shrinkwrap.descriptor.api.application6.ApplicationDescriptor;
+import org.jboss.shrinkwrap.descriptor.api.persistence20.PersistenceDescriptor;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+/**
+ * Example of an integration test using JBoss AS 7 and Arquillian.
```
gunnarmorling: Why is it an example? Also it's WF 8, right?

gunnarmorling:
Looks good to me apart from the class-level comments on the test classes.

integrationtest/wildfly/src/test/java/org/hibernate/search/test/integration/wildfly/ModuleMemberRegistrationEarArchiveWithJbossDeploymentIT.java
Sanne: It's an example as we're using some of these tests to give advice to users on how to run such integration tests..

Sanne:
I've merged it, as the issue pointed out by Gunnar is a more widespread problem.
Could you send a follow-up PR changing the various "JBoss 7" descriptions to WildFly, and also add a one line description on each of the various integration tests?
It's starting to get a bit hard to figure what the purpose of each IT is - the class names are a good hint but a small description would be more useful to remind what's the "special purpose" of each.

PR #789, opened 2015-02-04
HSEARCH-633 Package annotations

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-633

With this pull request it's possible to add filter and analyzers on packages containing indexed classes.


Sanne:
Always nice to see we can add some value at minimal complexity :+1: 

The comments in the tests are very minor things, up to you what you want to polish.
Only request I have is to add a (very minimal) mention in the documentation of this being possible at all?

Sanne:
> Should we add a note that all these definitions generally not only apply to the element at which they are given but globally? So e.g. giving an analyzer def on a package-info.java does not limit its usage to that package but makes it available in general.

+1 to add a note but that's not limited to the ones applied on packages. They are always "global"

Sanne:
great, merged

PR #791, opened 2015-02-04
HSEARCH-1802: add the lucene grouping feature to hibernate search

saschagrebe:
My basic implementation of the lucene grouping is inspired by the facetted search. It does only extract the entity infos from the index. The use of document extractor should be checked by someone with a deeper understanding of the hibernate search internals.

I think the API needs to be discused. Maybe the list() method in the full text query would be better to retrieve the grouping result. I could imagine it would make sense to resolve the entities directly.

I provided some basic tests (inspired by the tests for facetting) for the feature.

If you have any question don't hesitate to contact me.

Greetings,
Sascha Grebe


Sanne:
Hi Sascha, thanks!
I'll have an actual look tomorrow, as it's very late for me now; most of our team are in Europe.

Could you provide some high level overview of what your use case is?
We generally describe new ideas on the developer's mailing list, or on IRC.
(All contact details are here: http://hibernate.org/community/) and then create a new JIRA issue for HSEARCH here: https://hibernate.atlassian.net/browse/HSEARCH/

Also I see you're probably new to Git, feel free to send us a traditional patch file if you prefer, you could attach one to JIRA once we created an issue for this.

Sanne:
@saschagrebe this is looking very promising!

Could you start  by signing the contributor's agreement at https://cla.jboss.org ? It's a very quick process, just a checkbox on a web page to make sure you understand you're donating the code to the project.
It will ask you for which project, please pick "Hibernate" from the list. Also, then please let me know when you've signed it as it usually takes weeks before someone notifies me about that being done.

saschagrebe:
@Sanne Thanks for the information. This is my first contribution to an open source project.

I just filed an issue in JIRA (https://hibernate.atlassian.net/browse/HSEARCH-1802) and signed the contributor's agreement.

engine/src/main/java/org/hibernate/search/query/grouping/SimpleGroup.java
```diff
@@ -0,0 +1,80 @@
...
+
+import org.apache.lucene.search.ScoreDoc;
+import org.apache.lucene.search.TopDocs;
+import org.hibernate.search.query.engine.spi.DocumentExtractor;
+import org.hibernate.search.query.engine.spi.EntityInfo;
+
+/**
+ * @author Sascha Grebe
+ */
+public class SimpleGroup implements Group {
```
DavideD: I think we usually call this kind of implementations with the suffix impl, like 'GroupImpl'.
Shouldn't this also be in the impl package?

Some javadoc would be nice

engine/src/main/java/org/hibernate/search/query/grouping/SimpleGroupingResult.java
```diff
@@ -0,0 +1,71 @@
...
+import java.io.IOException;
+import java.util.LinkedList;
+import java.util.List;
+
+import org.hibernate.search.query.engine.spi.DocumentExtractor;
+
+/**
+ * @author Sascha Grebe
+ */
+public class SimpleGroupingResult implements GroupingResult {
```
DavideD: Similar to the comment to SimpleGroup, I think this should be in an impl package

DavideD:
@saschagrebe Thanks for the pull request, it looks really good.

It just need some minor adjustments to be more consistent with our code style.
But I'm going to do that so for now ignore my previous comments.

DavideD:
I'm going to close this pull request and send a new one with some additional commits

saschagrebe:
@DavideD 

Great to hear that you continue my work. Please have a closer look to the usage of document extractor. Maybe there will be a bottleneck when many documents will be returned.

If you have any questions don't hesitate to contact me.

PR #792, opened 2015-02-05
HSEARCH-1800 St thomas

emmanuelbernard:


emmanuelbernard:
https://hibernate.atlassian.net/browse/HSEARCH-1800

s4ke:
Since this API is aimed at Integrators I'd like to have a way to have access to the Low-Level methods of the ExtendedSearchIntegrator/SearchIntegrator used.

Also, I don't think WorkLoad is a good name for a class that as far as I can tell is a replacement of FullTextSession.

emmanuelbernard:
Hi @s4ke 

The parallel is more like this:

* `WorkLoadManager` == `EntityManagerFactory`
* `WorkLoad` == `EntityManager`

I remained quite conservative on what is exposed because this API would be a user API, not only for integrators. It can be useful for "lazy" integrators that don't want the full work of integrating with their own native API (like say EntityManager) but that's not the only target.

emmanuelbernard:
BTW I'm working on a first documentation draft.

s4ke:
@emmanuelbernard: about the conservatism: that's fine for me as long as I can somehow get hold of the SearchIntegrator. Are you against that? If yes, maybe we can make a "contract" that integrators that want to have acces to it have to cast the WorkLoadManager to its implementation?

s4ke:
After running mvn install I am left with a changed git repository:

manualsource/org.hibernate.search.manualsource.test.Star/ is added.

Should this be fixed?

Sanne:
Good catch @s4ke : yes the index should be either stored in memory or cleaned up.

s4ke:
@emmanuelbernard :

I thought about the naming a bit and I really don't like it if this is supposed to be user API. WorkLoadManager doesn't really suggest Search capabilities. Why not call it SearchManager and SearchManagerFactory?

gunnarmorling:
When would a user benefit from the proposed functionality? Specifically, if they don't work with JPA entities to begin with, what do they gain from this API over working with Lucene (or e.g. ES) directly?

s4ke:
This would work perfectly together with this: https://github.com/Hotware/Hibernate-Search-JPA . This is an important thing for people that want to use Hibernate-Search with JPA providers that happen to not be Hibernate-ORM.

hferentschik:
For my part, I thought we get a lot of these things as part of the free form entity stuff @Sanne is working on. In my world we start with @Sanne's refactoring and gradually extend it. I don't quite understand the need for another module here.

s4ke:
Just a follow up on my comments that I wanted this:

As of now I have no need for this anymore. Please disregard my comments saying otherwise.

Sanne:
@emmanuelbernard do you still think this is going to be useful?

If so I'm ok to merge it as an experimental module, provided you are able rebase it

gunnarmorling:
Ok, so I had a closer look this morning. Some general thoughts:

* Overall, I like the idea of having a simple API for indexing objects and getting them back from queries. That said, the story around this needs to be sharpened IMHO. If people have ES, getting their objects in there through JSON mapping a la Jackson is very simple. So I think it's not a super-attractive toy for this croud per se. We still give them association denormalization, switching between backends, and different work modes (JMS etc.), so these are things that provide a benefit.
* Embedded Lucene usage is a bit different, there the aspect of "API simplification" takes more affect.
* The limitation to actual Java classes is a severe one. St. Thomas would benefit from free-form very much, so this should be a close follow-up. I know, resources etc. ...

So yes, let's add it as a module and see whether it gains traction. As @emmanuelbernard is saying, there is not to loose much, the code is largely in place.

I'll add some more comments inline on specifics.

manualsource/src/main/java/org/hibernate/search/manualsource/FullTextQuery.java
```diff
@@ -0,0 +1,211 @@
...
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+
+package org.hibernate.search.manualsource;
+
+import java.util.Iterator;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.lucene.search.Explanation;
```
gunnarmorling: Can we avoid Lucene exposure in the API? It'll be a general theme in HS 6, but it'd be good to avoid it in newly written code to begin with.

manualsource/src/main/java/org/hibernate/search/manualsource/SearchFactory.java
```diff
@@ -0,0 +1,107 @@
...
+	/**
+	 * Retrieve an analyzer instance by its definition name
+	 *
+	 * @param name the name of the analyzer
+	 *
+	 * @return analyzer with the specified name
+	 *
+	 * @throws org.hibernate.search.exception.SearchException if the definition name is unknown
+	 */
+	Analyzer getAnalyzer(String name);
```
gunnarmorling: Could we skip that? I understand it's copied essentially, but it'd be great to avoid the leakage of Lucene.

manualsource/src/main/java/org/hibernate/search/manualsource/WorkLoad.java
```diff
@@ -0,0 +1,115 @@
...
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+
+package org.hibernate.search.manualsource;
+
+import java.io.Serializable;
+
+/**
+ * Represents an indexing and querying work load.
```
gunnarmorling: I'm not sure I like the mix of indexing and querying in a single contract. Specifically, what are semantics when running queries during an un-finished batch? On first thought, I'd split the two up into two things, e.g. `IndexingTask` and `Query` (yes, I know that name is taken).

manualsource/src/main/java/org/hibernate/search/manualsource/WorkLoad.java
```diff
@@ -0,0 +1,115 @@
...
+	 * @param entity The entity to index - must not be <code>null</code>.
+	 *
+	 * @throws IllegalArgumentException if entity is null or not an @Indexed entity
+	 */
+	<T> void index(T entity);
+
+	/**
+	 * @return the <code>SearchFactory</code> instance.
+	 */
+	SearchFactory getSearchFactory();
```
gunnarmorling: Should one rather get this from the `WorkloadManager` (or some other, more "service-like" contract)? It seems odd to expose it on the level of individual works.

manualsource/src/main/java/org/hibernate/search/manualsource/WorkLoad.java
```diff
@@ -0,0 +1,115 @@
...
+	 *
+	 * Any {@link org.hibernate.search.indexes.interceptor.EntityIndexingInterceptor} registered on the entity type will be ignored.
+	 *
+	 * @param entityType The class of the entities to remove.
+	 *
+	 * @throws IllegalArgumentException if entityType is <code>null</code> or not a class or superclass annotated with <code>@Indexed</code>.
+	 */
+	<T> void purgeAll(Class<T> entityType);
+
+	// TODO should we add the mass indexer?
```
gunnarmorling: I'd like that in a follow-up, as it provides good benefit over simple manual integration.

manualsource/src/main/java/org/hibernate/search/manualsource/impl/FullTextQueryImpl.java
```diff
@@ -0,0 +1,268 @@
...
+
+	private static final Log log = LoggerFactory.make();
+	private final WorkLoadImpl workLoadImpl;
+
+	private final HSQuery hSearchQuery;
+
+	/**
+	 * Constructs a  <code>FullTextQueryImpl</code> instance.
+	 */
+	public FullTextQueryImpl(org.apache.lucene.search.Query query,
```
gunnarmorling: Ok, so this would not work with ES. I suppose we are fine on the indexing side, but querying would need some enabling work.

manualsource/src/test/java/org/hibernate/search/manualsource/test/ManualSourceIndexingTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+		// add the data to the database
+		database.put( sun.getId(), sun );
+		database.put( alphaCentauri.getId(), alphaCentauri );
+
+		WorkLoad workLoad = workLoadManager.createWorkLoad();
+		workLoad.startBatch();
+		workLoad.index( sun );
+		workLoad.index( alphaCentauri );
+		workLoad.commitBatch();
```
gunnarmorling: How about making it more Lambda-ish:

    workLoadManager.executeBulkOfWork( workLoad -> {
        workLoad.index( sun );
        workLoad.index( alpha );
    });

Removes two methods from `WorkLoad`.

Actually, could it always be that style, also if there only is a single index task? I.e. why distinguish between single and batch?

manualsource/src/main/java/org/hibernate/search/manualsource/source/ObjectInitializer.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Interface implemented by the source of data to load specific entities.
+ *
+ * @author Emmanuel Bernard &lt;emmanuel@hibernate.org&gt;
+ */
+public interface ObjectInitializer {
+	/**
+	 * Initializes the set of requested objects and put them in the map.
+	 */
+	void initializeObjects(List<EntityKeyForLoad> keys, Map<EntityKeyForLoad, Object> idsToObjects, Context context);
```
gunnarmorling: Out of interest, what's the reason for passing in the result map as a parameter instead of returning it?

A pattern growing on me recently is passing in a "collector" accepting result(s):

    void initializeObjects(List<EntityKeyForLoad> keys, EntityCollector collector, Context context) {
        for(EntityKeyForLoad key : keys) {
            Object entity = ...;
            collector.addEntity( key, entity );
        }
    }

Essentially, it's a more abstract flavor of yours, exposing only those methods actually needed. Plus, it gives our caller the freedom of how to implement it and we can add more methods to that down the road if needed:

gsmet:
@yrodiere should we close this one? I think we have all the infrastructure needed in the Search 6 POC and we won't make any progress on this on the Search 5 codebase.

yrodiere:
@gsmet I think we can close it, yes: most of what is achieved here could be achieved with a custom mapper derived from the POJO mapper in Search 6.
I will move the ticket to 6 and won't close it yet, though, because it seems the ticket is supposed to benefit Hibernate Search users, not integrators, and if so, we need to make things simpler, or at least add documentation about how to write a (simple) mapper.

PR #793, opened 2015-02-05
HSEARCH-1663

gunnarmorling:
This picks some low-hanging fruits to get rid of several references to HS "impl" classes from ISPN.

The remaining references in "main" code are (going to ignore dependencies in test code for now):

infinispan-query
* SearchableCacheConfiguration
  * org.hibernate.search.engine.service.classloading.impl.DefaultClassLoaderService (there is a TODO already to replace it with something from ISPN)
* LuceneWorkTransformationVisitor
  * org.hibernate.search.backend.impl.WorkVisitor

infinispan-remote-query-server
* IndexingTagHandler
  * org.hibernate.search.engine.impl.LuceneOptionsImpl
  * org.hibernate.search.engine.metadata.impl.DocumentFieldMetadata
* QueryFacadeImpl
  * org.hibernate.search.bridge.builtin.impl.NullEncodingTwoWayFieldBridge
  * org.hibernate.search.bridge.builtin.impl.TwoWayString2FieldBridgeAdaptor

Any ideas for resolving some or all of these are welcome. I guess though we'll need to tackle them at another time, as the HS iteration is nearly over.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/backend/impl/batch/DefaultBatchBackend.java
+++ b/engine/src/main/java/org/hibernate/search/backend/impl/batch/DefaultBatchBackend.java
@@ -16,6 +16,7 @@ import org.hibernate.search.backend.OptimizeLuceneWork;
 import org.hibernate.search.backend.impl.StreamingSelectionVisitor;
 import org.hibernate.search.backend.impl.TransactionalSelectionVisitor;
 import org.hibernate.search.backend.impl.WorkQueuePerIndexSplitter;
+import org.hibernate.search.backend.spi.BatchBackend;
 import org.hibernate.search.batchindexing.MassIndexerProgressMonitor;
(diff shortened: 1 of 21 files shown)


hferentschik:
> I guess though we'll need to tackle them at another time, as the HS iteration is nearly over.

+1 I think we should wrap things up and merge the things we have to start with

gunnarmorling:
@Sanne Pushed an update which introduces `IndexingMode getIndexingMode()`. I removed `getIndexingStrategy()` from impl types where possible but I left it in place (and deprecated) where it's exposed through the SPI (`BuildContext`) as we keep higher promises there.

Sanne:
I've merged most of it: skipped the documentbuilder buildcontext changes as that's a bit inconvenient to polish now.
Thanks! that string/enum thing had been bothering me since a while..

PR #794, opened 2015-02-05
HSEARCH-1791

hferentschik:


hferentschik:
In the end I basically kept all your changes. Just made the tests a bit more concise just focusing on the issue. 

That said, we have some underlying issues and I think we just have not nailed yet what it is.

Sanne:
merged, thanks @hferentschik @gunnarmorling  !

PR #799, opened 2015-03-02
Fix Projection handling code to handle recursion in metadata properly

Sanne:
This should address the issues raised by various users
- https://hibernate.atlassian.net/browse/HSEARCH-1814
- https://hibernate.atlassian.net/browse/HSEARCH-1786

### DIFF
--- a/engine/src/main/java/org/hibernate/search/query/engine/spi/EntityInfo.java
+++ b/engine/src/main/java/org/hibernate/search/query/engine/spi/EntityInfo.java
@@ -7,7 +7,6 @@
 package org.hibernate.search.query.engine.spi;
 
 import java.io.Serializable;
-import java.util.List;
 
 /**
  * Wrapper class describing the loading of an element.
@@ -23,9 +22,11 @@ public interface EntityInfo {
 
 	Object[] getProjection();
 
-	List<Integer> getIndexesOfThis();
+	Iterable<Integer> getIndexesOfThis();
 
 	boolean isProjectThis();
 
 	void populateWithEntityInstance(Object entity);
+
+	void setIndexesOfThis(int x);
 }
--- a/engine/src/test/java/org/hibernate/search/test/projection/ProjectionConversionTest.java
+++ b/engine/src/test/java/org/hibernate/search/test/projection/ProjectionConversionTest.java
@@ -1,0 +1,241 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.test.projection;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Locale;
+
+import org.apache.lucene.document.Document;
+import org.apache.lucene.index.IndexableField;
+import org.apache.lucene.search.Query;
+import org.hibernate.search.annotations.DocumentId;
+import org.hibernate.search.annotations.Field;
+import org.hibernate.search.annotations.FieldBridge;
+import org.hibernate.search.annotations.Indexed;
+import org.hibernate.search.annotations.IndexedEmbedded;
+import org.hibernate.search.annotations.Store;
+import org.hibernate.search.backend.spi.Work;
+import org.hibernate.search.backend.spi.WorkType;
+import org.hibernate.search.bridge.LuceneOptions;
+import org.hibernate.search.bridge.TwoWayFieldBridge;
+import org.hibernate.search.bridge.builtin.LongBridge;
+import org.hibernate.search.engine.ProjectionConstants;
+import org.hibernate.search.engine.integration.impl.ExtendedSearchIntegrator;
+import org.hibernate.search.query.dsl.QueryBuilder;
+import org.hibernate.search.query.engine.spi.EntityInfo;
+import org.hibernate.search.testsupport.TestForIssue;
+import org.hibernate.search.testsupport.concurrency.ConcurrentRunner;
+import org.hibernate.search.testsupport.concurrency.ConcurrentRunner.TaskFactory;
+import org.hibernate.search.testsupport.junit.SearchFactoryHolder;
+import org.hibernate.search.testsupport.setup.TransactionContextForTest;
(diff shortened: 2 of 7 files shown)


engine/src/test/java/org/hibernate/search/test/projection/ProjectionConversionTest.java
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	@Test
+	public void projectionWithCustomBridge() {
+		projectionTestHelper( "customBridgedKeyword", "lowercase-keyword" );
+	}
+
+	@Test
+	public void projectingEmbeddedIdByPropertyName() {
+		projectionTestHelper( "embedded.id", Long.valueOf( 2l ) ); //FAILS OFTEN
```
hferentschik: FAILS OFTEN!? What is that comment supposed to mean?

engine/src/test/java/org/hibernate/search/test/projection/ProjectionConversionTest.java
```diff
@@ -0,0 +1,241 @@
...
+						}
+					};
+				}
+			}
+		).execute();
+	}
+
+	void projectionTestHelper(String projectionField, Object expectedValue) {
+		ExtendedSearchIntegrator searchFactory = sfHolder.getSearchFactory();
+		QueryBuilder guestQueryBuilder = searchFactory.buildQueryBuilder().forEntity( ExampleEntity.class ).get();
```
hferentschik: 'guestQueryBuilder'!? where is 'guest' coming from?

engine/src/test/java/org/hibernate/search/test/projection/ProjectionConversionTest.java
```diff
@@ -0,0 +1,241 @@
...
+
+	@Test
+	public void projectingOnConflictingMappedIdField() {
+		projectionTestHelper( "second.id", "a string" );
+	}
+
+	@Test
+	public void concurrentMixedProjections() throws Exception {
+		//The point of this test is to "simultaneously" project multiple different types
+		new ConcurrentRunner( 1000, 20,
```
hferentschik: Didn't you say that the metadata is 'stable' per search factory? There is no bootstrapping in the different threads, so what are we actually testing here?

engine/src/test/resources/log4j.properties
```diff
@@ -23,7 +23,7 @@ log4j.logger.org.jboss=info
 #log4j.logger.com.jboss=debug
 
 log4j.logger.org.hibernate=info
-log4j.logger.org.hibernate.search=debug
+log4j.logger.org.hibernate.search=info
```
hferentschik: +1

engine/src/main/java/org/hibernate/search/query/engine/spi/EntityInfo.java
```diff
@@ -23,9 +22,11 @@
 
 	Object[] getProjection();
 
-	List<Integer> getIndexesOfThis();
+	Iterable<Integer> getIndexesOfThis();
```
hferentschik: Why the switch to `Iterable`? After all you are changing an spi and imo an `Iterable` is a bad choice of return type.

emmanuelbernard: Ah yes I forgot it's a SPI. Are you breaking Infinispan?

engine/src/test/java/org/hibernate/search/test/projection/ProjectionConversionTest.java
Sanne: I'll remove it :)
It's a left over from notes on debugging. It was very puzzling to figure out, so I had some notes and some ugly system-outs in various places

thanks, good catch!

engine/src/test/java/org/hibernate/search/test/projection/ProjectionConversionTest.java
Sanne: +1 another left over from various other tests I have been playing with

engine/src/test/java/org/hibernate/search/test/projection/ProjectionConversionTest.java
Sanne: It turns out that concurrency wasn't the issue, but it's good to test for this down the road.
Since it would only fail "occasionally", I suspected we could have an issue with mixing up parallel projection requests

It only takes 0.2 seconds to run (especially after fixing the logger threshold) so I'd rather keep it.

engine/src/main/java/org/hibernate/search/query/engine/spi/EntityInfo.java
Sanne: I'll remove this commit as it's not really part of the solution.

Sanne: But `Iterable` is a great return type if you don't want to expose neither what kind of collection it is, nor mutators ;-)

engine/src/test/java/org/hibernate/search/test/projection/ProjectionConversionTest.java
hferentschik: Ok

engine/src/main/java/org/hibernate/search/query/engine/spi/EntityInfo.java
hferentschik: The only thing you can do is to iterate over it. Not even a simple empty/size check is possible. If you are afraid about mutation, wrapping into a unmodifiable collection is the better option imo.

Sanne: I've already removed the commit, so we're talking about theory here.

In this specific case, we don't need empty/size checks. All we need is the client to iterate over it, so that's good. I understand it depends on the specific situation, I just think that in this specific case we don't need any of that, and even the unmodifiable collection is an unneeded complexity.

Sanne: Also, exposing any form of List wouldn't allow us to have second thoughs on what kind of collection we'd need. In this specific case, I wanted to replace the List with a Set.

Sanne:
merged after applying suggestions.
Removed the proposed changes to EntityInfo

PR #804, opened 2015-03-12
HSEARCH-1824 Have the new state of a MutableSearchFactoryState not nuke ...

Sanne:
...the previously defined Filter(s) and Analyzer(s)

 - https://hibernate.atlassian.net/browse/HSEARCH-1824

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/impl/MutableSearchFactoryState.java
+++ b/engine/src/main/java/org/hibernate/search/engine/impl/MutableSearchFactoryState.java
@@ -9,6 +9,7 @@ package org.hibernate.search.engine.impl;
 
 import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.lucene.analysis.Analyzer;
 import org.hibernate.search.backend.spi.BackendQueueProcessor;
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/hibernate/search/engine/impl/MutableSearchFactoryState.java
```diff
@@ -41,9 +42,9 @@
 	private IndexingMode indexingMode;
 	private Worker worker;
 	private BackendQueueProcessor backendQueueProcessor;
-	private Map<String, FilterDef> filterDefinitions;
+	private Map<String, FilterDef> filterDefinitions = new ConcurrentHashMap<>();
```
gunnarmorling: I would declare all the map members of type `ConcurrentMap` to make their nature apparent. Callers I would just let pass plain maps whose contents I'd put to the internally used concurrent maps. I.e. would not rely on the fact that a concurrent map is passed in (e.g. see factoryState.setDocumentBuildersIndexedEntities(...) further below in this PR).

engine/src/test/java/org/hibernate/search/test/configuration/mutablefactory/DefinitionsOnHotRebootTest.java
```diff
@@ -0,0 +1,96 @@
...
+import org.hibernate.search.testsupport.setup.SearchConfigurationForTest;
+import org.junit.Test;
+
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Makes sure that Filter and Analyzer definitions survive a hot reboot
+ */
+@TestForIssue(jiraKey = "HSEARCH-1230")
```
gunnarmorling: s/1230/1824/?

gunnarmorling:
@Sanne Looks good to me apart from thos two remarks inline.

engine/src/test/java/org/hibernate/search/test/configuration/mutablefactory/DefinitionsOnHotRebootTest.java
Sanne: doh :) thanks!

engine/src/main/java/org/hibernate/search/engine/impl/MutableSearchFactoryState.java
Sanne: I agree but I'd rather keep that out of this specific change. It has the potential to go viral and get me to refactor all of it, which I should definitely do but not in the scope of this micro fix which we need ASAP.

Sanne:
merged

PR #805, opened 2015-03-17
Fix HSEARCH-655

yge45:
An attempt to fix https://hibernate.atlassian.net/browse/HSEARCH-655. 

Unit test is OK for me : index is not corrupted (can be open with luke). My use case describe in https://forum.hibernate.org/viewtopic.php?f=9&t=1038268 seems to be also OK.

### DIFF
--- a/orm/src/main/java/org/hibernate/search/batchindexing/impl/BatchCoordinator.java
+++ b/orm/src/main/java/org/hibernate/search/batchindexing/impl/BatchCoordinator.java
@@ -6,12 +6,17 @@
  */
 package org.hibernate.search.batchindexing.impl;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


Sanne:
ok to test

Sanne:
Thanks a lot @yge45 , the patch look excellent!

I'm merging it, I've only made two small changes in the unit test:
 - removed the '@RunWith(BMUnitRunner.class)' annotation and the 'hibernate.search.default.indexwriter.infostream' properties as they don't seem to be used
 - I've removed the 'Thread.sleep' methods by checking instead on the thread progress.. just to keep our testsuite execution as short as we can.

Did you already sign the contributor's agreement for Hibernate at cla.jboss.org ?

yge45:
Contributor's agreement for Hibernate signed

Sanne:
thanks!
Now you're free to send as many pull requests as you want ;)

PR #807, opened 2015-03-19
HSEARCH-1837 Allow Integrators to customize default directory providers ...

gustavonalle:
...for shortcuts

https://hibernate.atlassian.net/browse/HSEARCH-1837


Sanne:
Hi @gustavonalle , I've merged it, thanks! Looks like you know exactly how we structure things.

I've made one single change: please see my second commit, just moved one class to a different package.

Sanne:
BTW, didn't want to delay the fix for a minor question, but still have one doubt: why didn't you actually implement all of the default methods in the base class?

gustavonalle:
> why didn't you actually implement all of the default methods in the base class

Mainly to make extensions easier, like in infinispan I will be doing this:
https://github.com/gustavonalle/infinispan/blob/ISPN-5311/query/src/main/java/org/infinispan/query/backend/DirectoryProviderService.java

Sanne:
I guessed you were going to extend the base class, but you could still override any method even if it was implemented there right?

Do you mean you prefer them to be abstract? Ok, I was just curious.

PR #811, opened 2015-04-08
HSEARCH-1793 Making CriteriaObjectInitializer work correctly for cross-h...

gunnarmorling:
...ierarchy query results by creating one Criteria for each id space

### DIFF
--- a/orm/src/main/java/org/hibernate/search/query/hibernate/impl/CriteriaObjectInitializer.java
+++ b/orm/src/main/java/org/hibernate/search/query/hibernate/impl/CriteriaObjectInitializer.java
@@ -8,15 +8,23 @@ package org.hibernate.search.query.hibernate.impl;
 
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 
 import org.hibernate.Criteria;
 import org.hibernate.annotations.common.reflection.XMember;
+import org.hibernate.criterion.Criterion;
 import org.hibernate.criterion.Disjunction;
 import org.hibernate.criterion.Restrictions;
+import org.hibernate.engine.spi.SessionFactoryImplementor;
+import org.hibernate.search.cfg.spi.IdUniquenessResolver;
 import org.hibernate.search.engine.integration.impl.ExtendedSearchIntegrator;
+import org.hibernate.search.engine.service.spi.ServiceManager;
 import org.hibernate.search.engine.spi.DocumentBuilderIndexedEntity;
 import org.hibernate.search.exception.AssertionFailure;
 import org.hibernate.search.query.engine.spi.EntityInfo;
@@ -30,6 +38,7 @@ import org.hibernate.search.util.logging.impl.LoggerFactory;
  * Initialize object using one or several criteria queries.
  *
  * @author Emmanuel Bernard <emmanuel@hibernate.org>
+ * @author Gunnar Morling
  */
 public class CriteriaObjectInitializer implements ObjectInitializer {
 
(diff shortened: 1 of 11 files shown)


orm/src/main/java/org/hibernate/search/query/hibernate/impl/CriteriaObjectInitializer.java
```diff
@@ -105,33 +111,124 @@ private void setCriteriaTimeout(Criteria criteria, TimeoutManager timeoutManager
...
+
+			criteria.add( getIdListCriterion( infosByIdSpace.values().iterator().next(), objectInitializationContext ) );
+
+			return Collections.singletonList( criteria );
+		}
+		// entities originate from different id spaces -> criteria per space
+		else {
+			// Cannot use external criteria for fetching entities from different spaces
+			if ( objectInitializationContext.getCriteria() != null ) {
+				log.givenCriteriaObjectCannotBeApplied();
```
gunnarmorling: AFAICS we cannot make use of an externally provided `Criteria` in case we need to fetch entities from several id spaces (i.e. I don't think we can fetch all entities using a single criteria object in this case).

Unless there actually is a way to do this, I could imagine to provide a way for registering a call back which returns user-provided criteria for given entity types:

    public interface CriteriaProvider {
        Criteria getCriteria(EntityType entityType);
    }

and then:

    final FullTextSession fullTextSession = ...;
    List results = fullTextSession.createFullTextQuery( query )
        .setCriteriaProvider(
            new CriteriaProvider {
                public Criteria getCriteria(EntityType entityType) {
                    // for Book add a fetch join, everything else is defaulted
                    if ( entityType == Book.class ) {
                        return fullTextSession.createCriteria( Book.class ).setFetchMode( "authors", FetchMode.JOIN );
                    }
                    else {
                        return fullTextSession.createCriteria( entityType );
                    }
                }
            }
        )
        .list();

This callback would be invoked by the initializer if requiring a `Criteria` for a given type.

Any thoughts? Anything much simpler I didn't see?

orm/src/main/java/org/hibernate/search/query/hibernate/impl/CriteriaObjectInitializer.java
```diff
@@ -105,33 +111,124 @@ private void setCriteriaTimeout(Criteria criteria, TimeoutManager timeoutManager
...
+		IdUniquenessResolver resolver = serviceManager.requestService( IdUniquenessResolver.class );
+		SessionFactoryImplementor sessionFactory = (SessionFactoryImplementor) objectInitializationContext.getSession().getSessionFactory();
+
+		try {
+			Map<Class<?>, List<EntityInfo>> infosByIdSpace = new HashMap<>();
+
+			for ( EntityInfo entityInfo : entityInfos ) {
+				boolean addedToExistingGroup = processEntityInfo( infosByIdSpace, resolver, entityInfo );
+
+				if ( !addedToExistingGroup ) {
```
hferentschik: why do you need to return a boolean from `processEntityInfo`? Could you not handle the adding of the new `EntityInfo` list in `processEntityInfo` directly?

orm/src/main/java/org/hibernate/search/query/hibernate/impl/CriteriaObjectInitializer.java
hferentschik: Sounds reasonable what you are saying. Not sure whether it is worth the effort. Maybe worth creating a new issue!?

orm/src/test/java/org/hibernate/search/test/query/dsl/DSLTest.java
```diff
@@ -890,6 +890,57 @@ public void testUseMatchQueryOnEmbeddedNumericIdCreatesTermQuery() throws Except
 		);
 	}
 
+	@Test
```
hferentschik: For my taste `DSLTest` gets a big too big to be easily understandable. It gets quite hard to figure out which test classes are involved where. Why not create a single test class for these criteria issue, maybe even in a dedicated directory.

orm/src/test/java/org/hibernate/search/test/query/dsl/DSLTest.java
```diff
@@ -890,6 +890,57 @@ public void testUseMatchQueryOnEmbeddedNumericIdCreatesTermQuery() throws Except
...
+	@TestForIssue(jiraKey = "HSEARCH-1793")
+	public void testQueryingForEntitiesFromDifferentIdSpaces() throws Exception {
+		QueryBuilder queryBuilder = fullTextSession.getSearchFactory()
+				.buildQueryBuilder().forEntity( EducationalInstitution.class ).get();
+
+		Query query = queryBuilder.bool()
```
hferentschik: I think some comments either here at the query construction or at the assert step would help. Why does this setup and the result you are asserting verify that HSEARCH-1793?

hferentschik: Personally I would love to have a more direct test, which for example just tests `CriteriaObjectInitializer` and that for example x amount of `Criteria` queries get created/executed. The contract of  `CriteriaObjectInitializer` is of course hard to test stand-alone. Maybe one could do something with a Byteman test!?

hferentschik: Maybe add a test for the exceptional case as well?

orm/src/main/java/org/hibernate/search/query/hibernate/impl/CriteriaObjectInitializer.java
gunnarmorling: Ok.

orm/src/test/java/org/hibernate/search/test/query/dsl/DSLTest.java
gunnarmorling: Good idea, I'll move it to a separate class/directory.

orm/src/test/java/org/hibernate/search/test/query/dsl/DSLTest.java
gunnarmorling: I prefer to have a test which makes sure the "whole thing" actually works. To assert on the created `Criteria` I'd have to use Byteman as you say or similar which I personally find quite cumbersome. And I still wouldn't know whether these criteria work as intended in the end, so I'd still need a test in the current form.

I'll add some more comments to the test itself.

> Maybe add a test for the exceptional case as well?

What exactly DYM by "exceptional case"?

orm/src/main/java/org/hibernate/search/query/hibernate/impl/CriteriaObjectInitializer.java
gunnarmorling: Ok, I'll create a separate issue for that one.

gunnarmorling: Created [HSEARCH-1852](https://hibernate.atlassian.net/browse/HSEARCH-1852).

gunnarmorling:
@hferentschik, pushed an update.

hferentschik:
Merged, thanks.

PR #812, opened 2015-04-08
HSEARCH-1792 Multi-tenancy support

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-1792

This patch will add support for multi-tenancy to hibernate search by keeping track of the tenant id in the `Work` class and storing it in the index,

Documentaitno is currently missing but I will add it if the solution make sense.

@Sanne what do you think?


Sanne:
merged after some pair programming

PR #813, opened 2015-04-10
HSEARCH-1804 Propagating boost value given for embeddables to contained fields and sub-embeddables

gunnarmorling:


hferentschik:
merged, thanks

PR #814, opened 2015-04-13
HSEARCH-1773, HSEARCH-1356: Establishing public IndexWorkVisitor contract

gunnarmorling:
This PR removes the `WorkVisitor` interface from "impl" in favor of a new public `IndexWorkVisitor` contract. The change is mainly in method and type names, making the nature of the visitor pattern more easily to grasp.

There is no change in numbers of allocated objects. Esp. I didn't change the fact that most visitors return a delegate which does the actual work. While this easily could be done in the visitors themselves, it would require allocation of parameter context objects which I wanted to avoid. Visitors expecting only a single object as input can do so at no additional cost, so that's what I did for `FlushStrategySelector` (same can be done in the ISPN visitor implementation).


engine/src/main/java/org/hibernate/search/backend/LuceneWork.java
```diff
@@ -61,7 +59,14 @@ public String getIdInString() {
 		return idInString;
 	}
 
-	public abstract <T> T getWorkDelegate(WorkVisitor<T> visitor);
```
gunnarmorling: Note that this removes a previously publicly exposed method. IMO that's acceptable though, given that the parameter type resided in "impl", making that previous signature kind of "faulty" anyways.

hferentschik:
> There is no change in numbers of allocated objects. Esp. I didn't change the fact that most visitors return a delegate which does the actual work. While this easily could be done in the visitors themselves,

Right, this is what I would expect (the work being done by the visitors), which imo would make so much more sense. Take `StreamingOperationExecutorSelector` for example, you are passing all these different work instance and make no use of them. it is just a fancy switch statement to select the "right" executor. To me this still does not make sense and is not in the spirit of a `Visitor`.

gunnarmorling:
> Right, this is what I would expect (the work being done by the visitors), which imo would make so much more sense. Take StreamingOperationExecutorSelector for example, you are passing all these different work instance and make no use of them. it is just a fancy switch statement to select the "right" executor. To me this still does not make sense and is not in the spirit of a Visitor.

I will leave this battle to you and @Sanne :) I agree it's a bit "nicer" to do the actual work in the visitor itself, but I also can see that returning an executor with the right method signature helps in avoiding object allocations. So if that's beneficial on the hot path, why not.

But this is now a decision for specific visitor implemenations, the proposed design allows for what you describe (and it's actually done for flush strategies).

gunnarmorling:
Pushed an update.

engine/src/main/java/org/hibernate/search/backend/IndexWorkVisitor.java
```diff
@@ -0,0 +1,35 @@
...
+ * <p>
+ * Implementations can receive a specific input parameter and return a specific output type. Use {@link Void} as
+ * parameterization for each of the two not needed.
+ *
+ * @author Sanne Grinovero
+ * @author Gunnar Morling
+ * @param <P> Context parameter type expected by a specific visitor
+ * @param <R> Return type provided by a specific visitor
+ */
+public interface IndexWorkVisitor<P, R> {
```
hferentschik: What about my comment that this interface cab be simplified if one just slightly changes the code in `NRTWorkspaceImpl`?

gunnarmorling: I replied to it in https://github.com/gunnarmorling/hibernate-search/commit/4f64f6943cbe06629a736c1647386b456f71323c#commitcomment-10700416:

> [The parameter] is needed to pass in visitor-specific context where required. The FlushStrategySelector uses it, but also one implementation in ISPN can take advantage of it. This allows to have the actual work done in the visitor itself rather than returning a delegate.
>
> I only did not do that for the other existing implementations in order to avoid allocation of a context parameter object (which would be needed since more than one contextual object need to be passed in). But the design should allow for this in cases where a) there only is a single context object needed anyways (so no further allocation required) or b) we don't mind the allocation of such wrapper for passing in several context parameters.

It enables exactly what you seem to be after: have visitors which do the actual work. Whether that's not done in specific cases due to performance considerations should not pose a limitation to the design in general IMO.

gunnarmorling:
Can this one be merged?

Sanne:
hi @gunnarmorling , could you rebase it please?

gunnarmorling:
> could you rebase it please?

Rebased and force-pushed.

hferentschik:
> Can this one be merged?

for my part it still falls short of what I had envisioned. Not sure what @Sanne thinks

Sanne:
@hferentschik I don't know what you had envisioned, but this seems to fulfil the requirements of not having Infinispan rely on the '.impl' package and seems rather safe (bar all the renames) so I'll take it.

Of course this doesn't have to be the definitive shape if you feel like making proposals, but I'll always have some resistance to change in this area as long as we don't have automated performance regressions tests :)

Sanne:
rebased and merged!

hferentschik:
> I don't know what you had envisioned, but this seems to fulfil the requirements of not having Infinispan rely on the '.impl' package and seems rather safe (bar all the renames) so I'll take it.

Your call. It is a step forward, but as said, it does not address all of my concerns.

PR #816, opened 2015-04-13
HSEARCH-1841 Refactor DeleteByQuerySupport

s4ke:
DeleteByQuerySupport has slimmed down.

Instead of having 3 different interfaces, I have merged the toString(), fromString() and toLuceneQuery() logic into "DeletionQueryMapper".

Furthermore, DeleteByQuerySupport now doesn't host the converting logic anymore. It's just used to redirect to the correct DeletionQueryMapper.

All DeletionQueries are now required to have a corresponding DeletionQueryMapper.


engine/src/main/java/org/hibernate/search/backend/impl/DeletionQueryMapper.java
```diff
@@ -0,0 +1,38 @@
...
+package org.hibernate.search.backend.impl;
+
+import org.apache.lucene.search.Query;
+import org.hibernate.search.backend.DeletionQuery;
+import org.hibernate.search.util.impl.ScopedAnalyzer;
+
+/**
+ * @author Martin Braun
+ */
+public interface DeletionQueryMapper {
```
gunnarmorling: I'm wondering why this is an interface, there seems to be only one implementation? Is this meant to be extended in some way?

engine/src/main/java/org/hibernate/search/backend/impl/DeletionQueryMapper.java
```diff
@@ -0,0 +1,38 @@
...
+
+	Query toLuceneQuery(DeletionQuery deletionQuery, ScopedAnalyzer analyzerForEntity);
+
+	String[] toString(DeletionQuery deletionQuery);
+
+	DeletionQuery fromString(String[] string);
+
+
+	/*
```
gunnarmorling: I would move this to the class-level docs of this interface (next to a general description of its purpose), e.g. by adding a paragraph "design notes: ...".

engine/src/main/java/org/hibernate/search/backend/impl/DeletionQueryMapper.java
s4ke: Yes, this is intended for further extension. Why should it be a problem if there is only one implementation at the moment?

engine/src/main/java/org/hibernate/search/backend/impl/DeletionQueryMapper.java
s4ke: Seems like a good idea.

engine/src/main/java/org/hibernate/search/backend/impl/DeleteByQuerySupport.java
```diff
@@ -41,221 +29,55 @@ private DeleteByQuerySupport() {
...
-		{
-			Map<Integer, StringToQueryMapper> map = new HashMap<>();
-
-			map.put( SingularTermQuery.QUERY_KEY, new StringToQueryMapper() {
+		if ( SUPPORTED_TYPES == null ) {
+			//LAZY init, but don't to it more than once
+			SUPPORTED_TYPES_LOCK.lock();
+			try {
+				if ( SUPPORTED_TYPES == null ) {
+					Map<Integer, Class<? extends DeletionQuery>> map = new HashMap<>();
```
gunnarmorling: I'm a bit lost here; Why is this map created? In the end only the value set seems to be relevant. Also is the lazy creation worth all the effort (locking etc.)? On first thought I'd say that a simple map should suffice for both the methods of this class:

    private static Map<Class<? extends DeletionQuery>, DeletionQueryMapper> MAPPERS_BY_QUERY=
        Collections.unmodifiableMap(
            Collections.singletonMap(SingularTermDeletionQuery.class, new SingularTermDeletionQueryMapper() )
        );

    public static boolean isSupported(Class<? extends DeletionQuery> type) {
        return MAPPERS_BY_QUERY.containsKey( type );
    }

    public static DeletionQueryMapper getMapper(Class<? extends DeletionQuery> type) {
        return MAPPERS_BY_QUERY.get( type );
    }

That'd be much simpler, but maybe I'm just missing some piece? One thing I'm unsure is how would alternative `DeletionQuery` implementations be registered? Is that supported at all?

engine/src/main/java/org/hibernate/search/backend/SingularTermDeletionQuery.java
```diff
@@ -11,7 +11,7 @@
  *
  * @author Martin Braun
  */
-public final class SingularTermQuery implements DeletionQuery {
+public final class SingularTermDeletionQuery implements DeletionQuery {
```
gunnarmorling: Is this meant to be a public type (i.e. to be referenced by users)? If so, that renaming would break existing usages. But if it is actually not meant to be a user-accessible class, it should be moved to some "impl" package.

engine/src/main/java/org/hibernate/search/backend/impl/DeleteByQuerySupport.java
s4ke: Why it's lazy:

@Sanne doesn't want this stuff initialized when the class is loaded (see: https://hibernate.atlassian.net/browse/HSEARCH-1841)

Why it's no direct Map access:

@Sanne wanted me to change this to an array because of performance reasons. We could drop the Map, that's right.

engine/src/main/java/org/hibernate/search/backend/impl/DeletionQueryMapper.java
gunnarmorling: It's not a problem per-se, I'm more trying to figure the rationale behind the design. What other implementations do you envision? A comment would be good, describing possible means of extension. Otherwise, an interface with a single implemenation will aways make us (at least me;)) wonder what it is about and whether the interface/implementation split is actually needed.

engine/src/main/java/org/hibernate/search/backend/SingularTermDeletionQuery.java
s4ke: It's not public. Will move this to an impl package.

engine/src/main/java/org/hibernate/search/backend/impl/DeletionQueryMapper.java
s4ke: Okay. I am a big fan of putting everything behind an interface. This enforces a coding style that can easily be unit tested.

engine/src/main/java/org/hibernate/search/backend/impl/DeletionQueryMapper.java
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.lucene.search.Query;
+import org.hibernate.search.backend.DeletionQuery;
+import org.hibernate.search.util.impl.ScopedAnalyzer;
+
+/**
+ * @author Martin Braun
+ */
+public interface DeletionQueryMapper {
+
+	Query toLuceneQuery(DeletionQuery deletionQuery, ScopedAnalyzer analyzerForEntity);
```
gunnarmorling: Have you considered to move `toLuceneQuery()` and `toString()` to `DeletionQuery` itself? That way you'd save one interaction (map look-up or similar) when you have the `DeletionQuery` instance in your hands already. The support-class (or "factory") would only be needed during de-serialization then.

engine/src/main/java/org/hibernate/search/backend/impl/DeleteByQuerySupport.java
gunnarmorling: Ok, an array is a bit nicer as you do the look-up by the numeric key during de-serialization. But I don't see any problem at all in doing this at class initialization time. Maybe Sanne referred to several initialization blocks being used instead of a single one.

engine/src/main/java/org/hibernate/search/backend/impl/DeletionQueryMapper.java
s4ke: yes. I think we discussed this in depth in the JIRA issue already :). I don't want this stuff in separate places to be honest. fromString(...) and toString(...) are complete opposites and "undo" each other, so it's a good idea to keep them in the same place imo.

https://hibernate.atlassian.net/browse/HSEARCH-1841

engine/src/main/java/org/hibernate/search/backend/impl/DeleteByQuerySupport.java
s4ke: He was talking about wrong versions of Lucene being loaded:

"If there ever is an exception in this code, that's going to result in hard to diagnose error messages.. for example that might happen if the user is attempting to use a wrongly matching version of Lucene jars." (from the JIRA issue)

engine/src/main/java/org/hibernate/search/backend/impl/DeletionQueryMapper.java
gunnarmorling: > This enforces a coding style that can easily be unit tested.

In which way does the interface help with that? I'm not against the
interface per-se, but I only would bother adding it if a) several
implementations are possible or b) the implementation logic should be
"hidden" in some private package. Either way it should be documented
because the reasoning is not apparent atm.

engine/src/main/java/org/hibernate/search/backend/impl/DeletionQueryMapper.java
gunnarmorling: > I don't want this stuff in separate places to be honest. fromString(...)
and toString(...) are complete opposites and "undo" each other, so it's a
good idea to keep them in the same place imo.

Hum, ok. It's not uncommon though to have serialization code on the
type-to-be-serialized itself and de-serialization code on some accompanying
factory class (or a static method in said class). I don't really mind
either way, but keep in mind that this approach requires the look-up of the
mapper type which would not be needed otherwise. Would be interesting to
learn what the others think?

engine/src/main/java/org/hibernate/search/backend/impl/DeleteByQuerySupport.java
gunnarmorling: > He was talking about wrong versions of Lucene being loaded

I don't know how this would be easier or harder to detect depending on
whether this eagerly or lazily initialized? This class does not even import
any Lucene types at all?

engine/src/main/java/org/hibernate/search/backend/impl/DeletionQueryMapper.java
s4ke: Okay. In this case several implementations are possible. I will add a comment.

In my experience, by declaring interfaces first and hiding the logic completely, you are thinking more modular, even if you are not using mock-ups.

engine/src/main/java/org/hibernate/search/backend/impl/DeleteByQuerySupport.java
s4ke: But it instantiates at least one mapper which imports Lucene code which in turn might fail (that's the reasoning I understood).

engine/src/main/java/org/hibernate/search/backend/impl/DeletionQueryMapper.java
s4ke: I know it's not uncommon and I don't really mind it, but in this case I like the "everything together" approach more.

Also, you also have to look-up the mapper type if you do it with the split approach, but you don't have to lookup the serialization/deserialization code in two different classes.

To be honest, I don't think where this logic is placed shouldn't be that big of a problem for this use case :).

gunnarmorling: Look, my general feeling with this is that it is way more complex than it actually has to be. There is a single piece of logic (delete by query), but this is split into an two interfaces, two implementations, a "support class", uses lazy initialiation with locking, needs map/array look-ups etc. etc.

I would go for a much simpler design:

* Have `getLuceneQuery()` and `toString()` as instance methods on the query class (I don't mind splitting it up into interface/impl, so to allow for alternative impls down the road; you'd still need to figure a way though to configure which implementation to use, or is something for that already in place?)
* Have `fromString()` as a static method on the implementation class(es) (so you have your "everything-together" pattern)
* Invoke that static method during de-serialization if you see the right query key, otherwise raise an exception

That way you get rid of quite some code, making things much easier to understand and you still can evolve from there if actually needed.

s4ke: Hmm. I kinda understand what you mean now.

At the moment serialization is done by storing the key of a query that can 100% be matched back to a class. But this class doesn't have to stay the same between Hibernate-Search versions, enabling us to change the implementations between versions however we'd like.

To get this straight, you want a (atm straight forward) switch-case block (to remove the need for a array lookup) combined with manual calls to the right fromString() method. I guess this logic can then stay in DeleteByQuerySupport until Hibernate-Search uses Java8 (then we can move it to the DeletionQuery interface).

s4ke:
@gunnarmorling: I think this pull request has gotten too many comments as of now. This should be closed and I will open another PR after I have fixed the things we agreed on.

engine/src/main/java/org/hibernate/search/backend/impl/DeleteByQuerySupport.java
```diff
@@ -41,221 +29,36 @@ private DeleteByQuerySupport() {
...
-							return new SingularTermQuery( string[1], Long.parseLong( string[2] ) );
-						case DOUBLE:
-							return new SingularTermQuery( string[1], Double.parseDouble( string[2] ) );
-						default:
-							throw new AssertionFailure( "wrong Type!" );
-					}
-
+		if ( SUPPORTED_TYPES == null ) {
+			// LAZY init, but don't to it more than once
+			SUPPORTED_TYPES_LOCK.lock();
```
gunnarmorling: You still do the locking thingy here. I'd if/else on the given class and you are done. Esp. since there is no way to plug in another type anyways. Or is there? Also the single user of this method raises an exception in case the type would be "not supported". This all seems still a bit odd to me, at least I don't see how it would work.

engine/src/main/java/org/hibernate/search/backend/DeletionQuery.java
```diff
@@ -15,10 +19,23 @@
...
 	 * @return the unique query key
 	 */
 	int getQueryKey();
 
+	/**
+	 * converts this DeletionQuery to a Lucene Query
+	 *
+	 * @param analyzerForEntity the analyzer to be used for this query
+	 */
+	Query toLuceneQuery(ScopedAnalyzer analyzerForEntity);
```
gunnarmorling: @Sanne: I'm wondering whether this interface (and its impl) are rightly located in an API package. Isn't it more SPI? Also how strict are we about breaking changes here. I don't think adding these new methods is a big deal, as it all seems quite experimental to me anyways (should we tag it as such btw.), but WDYT?

engine/src/main/java/org/hibernate/search/backend/impl/DeleteByQuerySupport.java
s4ke: Oh. forgot that one. Thanks. :)

engine/src/main/java/org/hibernate/search/backend/DeletionQuery.java
s4ke: I am okay with moving this to spi as long it's not impl.

engine/src/main/java/org/hibernate/search/backend/impl/DeleteByQuerySupport.java
gunnarmorling: @s4ke You still plan to send an update for this? The next release is approaching quickly :)

Sanne:
thanks!

PR #817, opened 2015-04-13
HSEARCH-1762 Improve built-in bridges javadoc

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-1762

@hferentschik Do you think this is enough?


hferentschik:
Rebased and merged. Thanks.

PR #819, opened 2015-04-15
HSEARCH-809

hferentschik:
NOT TO BE MERGED YET!

Here is the first cut of the new faceting. All the old tests pass again and I just want to look at some more minor details. I thought I create a pull request to get some feedback which I can then work into my current work. 

Right now, the documentation needs still be updating and surprise surprise the OSGi tests are failing again (my guess is that I use some faceting classes which are not properly exported, not sure).

So what has changed:
* Most notably we have now two new annotations `@Facet` and `@Facets`. Contrary to before, one needs to configure now which fields will be used for faceting ( we need to know to index the right doc values)
* Faceting is not using a field cache anymore, but Lucene's dynamic faceting (no taxonomy index) using `DocValues`
* The API exposed via our faceting DSL is still intact apart from two changes
 * You cannot do a range facet request on strings anymore (controversial anyways)
 * the default for `includeZeroCounts` is now `false`.

Thoughts?


gunnarmorling:
> Contrary to before, one needs to configure now which fields will be used for faceting

Would it make sense to have a class-level annotation or similar setting to enable facetting for all properties of a given entity? This may help people migrating from previous versions.

engine/src/main/java/org/hibernate/search/annotations/Facet.java
```diff
@@ -0,0 +1,42 @@
...
+	 * @return the field name this annotation refers to. It can be omitted in case there is only a single {@code @Field}
+	 * annotation
+	 */
+	String forField() default "";
+
+	/**
+	 * @return the encoding type to use for this facet. Per default the encoding type is chosen based on the type of the
+	 * entity property.
+	 */
+	FacetEncodingType encoding() default FacetEncodingType.AUTO;
```
gunnarmorling: What is the use case for choosing a different encoding than for the property?

engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
```diff
@@ -946,6 +991,22 @@ private void bindFieldAnnotation(
 					.indexNullAs( nullToken );
 		}
 
+		if ( !facetAnnotations.isEmpty() ) {
```
gunnarmorling: The `isEmpty()` check should not be needed.

engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
```diff
@@ -577,6 +588,158 @@ else if ( currentFieldValue instanceof Map ) {
...
+							facetMetadata.getFacetName(),
+							calendar.getTime().getTime()
+					);
+				}
+				else {
+					throw new AssertionFailure( "Unexpected value type for faceting: " + value.getClass().getName() );
+				}
+				break;
+			}
+			case DOUBLE: {
```
gunnarmorling: Better check the type prior to casting?

engine/src/main/java/org/hibernate/search/query/engine/impl/QueryHits.java
```diff
@@ -252,19 +297,192 @@ private void updateTopDocs(int n) throws IOException {
...
+			}
+		}
+		else {
+			throw log.unsupportedFacetRangeParameter( facetRequest.getFacetValueType().getName() );
+		}
+
+		return facets;
+	}
+
+	private FacetResult getFacetResult(RangeFacetRequest facetRequest) throws IOException {
```
gunnarmorling: `getFloatingPointFacetResult()`? The name seems more generic than what it does.

gunnarmorling:
@hferentschik Looks good to me overall. My only question really is when one would use a different encoding for the facet field.

Apart from that migration is an interesting topic. E.g. I assume people need to re-build their index after upgrading if they use facet queries? Also a "globally enable faceting" mode may be interesting for some, at least for a migration period.

engine/src/main/java/org/hibernate/search/query/engine/impl/QueryHits.java
```diff
@@ -252,19 +297,192 @@ private void updateTopDocs(int n) throws IOException {
...
+			final TermsEnum iterator = terms.iterator( null ); //we have no TermsEnum to reuse
+			BytesRef byteRef;
+			while ( ( byteRef = iterator.next() ) != null ) {
+				termValues.add( byteRef.utf8ToString() );
+			}
+		}
+		return termValues;
+	}
+
 	private Collector optionallyEnableFacetingCollectors(Collector collector) {
```
gunnarmorling: The "s" can go away now: optionallyEnableFacetingCollector *s*

Sanne:
> Apart from that migration is an interesting topic. E.g. I assume people need to re-build their index after upgrading if they use facet queries?

Probably, but that's ok. It's very often a requirement.

engine/src/main/java/org/hibernate/search/util/impl/ReflectionHelper.java
```diff
@@ -200,6 +200,36 @@ public static Object createInstance(Class<?> clazz, boolean checkForFactoryAnnot
...
 	}
 
+	/**
+	 * Checks whether the specified type is a floating point type.
+	 *
+	 * @param type the type to check
+	 * @return {@code true} if the specified type is an floating point type or a wrapper thereof. {@code false} otherwise.
+	 */
+	public static boolean isFloatingPointType(Class<?> type) {
+		return float.class.equals( type )
```
gunnarmorling: So btw. one cannot do facet queries e.g. on `BigDecimal`? Is this a limitation in Lucene or on our side? It seems interesting when it comes to prices for instance.

Sanne:
@hferentschik I haven't looked at the details yet, but from high level and your description it looks awesome. It's exciting that you've verified (as we hoped) that we don't need to change the API and we get to drop field caches :+1: 

Could you summarize why this should not be merged? (assuming we don't spot any need during a more thourough review)

hferentschik:
> Would it make sense to have a class-level annotation or similar setting to enable facetting for all properties of a given entity? This may help people migrating from previous versions.

No, I don't think so. This would just unnecessarily bloat the index. I think in this case it is really better to force a explicit migration. This way one is also forced to think about the encoding.

engine/src/main/java/org/hibernate/search/annotations/Facet.java
hferentschik: Discrete vs Range facets. Per default strings are discrete (distinct values) whereas numbers are numeric doc values which can be used for ranges. Now think about a case where your domain contains a int value, but the values are really just a set of distinct values (or at least you want to treat them as such).  For example, a rating 1-10. You can also work with ranges here of course, but if you just want the count for each of the different rating a discrete facet (using string encoding) is much simpler). Does this make sense? The documentation will contain some details about this of course.

engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
hferentschik: true, true - that was something from before I did some refactoring. Removing it.

engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
hferentschik: ok

engine/src/main/java/org/hibernate/search/query/engine/impl/QueryHits.java
hferentschik: meant to be 'getFacetResultForFloatingPointRange'

hferentschik:
> Looks good to me overall.

Cool. I'll push an update addressing your comments.

> My only question really is when one would use a different encoding for the facet field.

I hope the answer I gave on your other comment answers your question

> Apart from that migration is an interesting topic.

Sure. This is not a effortless migration. People will have to make actual changes. In many cases it will be done with adding a `@Facet` annotation, but there are other border cases. Obviously the documentation needs to be updated and we need a decent section in the migration guide.

An interesting question here is about versioning. What version would one expect for the type changes this pull request introduces. For example, going Final after merging this is a no no imo. 

> E.g. I assume people need to re-build their index after upgrading if they use facet queries? 

Sure. A full rebuild it required. They won't have `DocValues` in their index.

> Also a "globally enable faceting" mode may be interesting for some, at least for a migration period.

What would that do?

engine/src/main/java/org/hibernate/search/query/engine/impl/QueryHits.java
hferentschik: ok

engine/src/main/java/org/hibernate/search/util/impl/ReflectionHelper.java
hferentschik: No, not at the moment. I was thinking a bit about it, but in order to use range faceting one would need to convert the `BigDecimal` to a double. This is potentially a conversion where precision is a problem. One would need a way to handle this. See also `BigDecimalNumericFieldBridge` in our docs which suggests a lossy numeric field bridge for `BigDecimal`. We can for sure think a bit more about it.

hferentschik:
Could you summarize why this should not be merged? (assuming we don't spot any need during a more thourough review)

The short answer - it is not ready yet :-)

The long answer:
* I need to add some more tests to verify some border cases, including forcing a string encoding for a number
* Add some more consistency checks in the metadata provider to avoid problems during runtime faceting (eg the targeted field cannot be analyzed, etc)
* I am also in the process to verify that the problems with *ToMany associations are fixed as expected (eg HSEARCH-900)
* Above all, the documentation for all these changes is absent so far. We can hardly merge this and then even release w/o any information on how this is working now.

engine/src/main/java/org/hibernate/search/annotations/Facet.java
gunnarmorling: > if you just want the count for each of the different rating a discrete facet (using string encoding) is much simpler

Not sure how easy it is, but couldn't our implementation handle discrete facet queries in case the facet field is numeric (via a collector)? In the end, discrete values can be considered as a list of ranges of size 1. If that's possible, then we'd had one less knob exposed (which can be configured inconsistently etc.).

gunnarmorling:
>> Would it make sense to have a class-level annotation or similar setting to enable facetting for all properties of a given entity?
>
> No, I don't think so. This would just unnecessarily bloat the index.

Not really; It'd be a setting people would consciously use if that's what they need. Only that they would not be forced to sprinkle `@Facet` annotations all over their entity.

> What would that do? 

Basically what I described above: allow for facet queries on all the properties of an entity in a less verbose way than annotating each property.

engine/src/main/java/org/hibernate/search/annotations/Facet.java
hferentschik: > Not sure how easy it is, but couldn't our implementation handle discrete facet queries in case the facet field is numeric (via a collector)

Sure we can, but this is what we want to get away from, right? Our facet implementation so far was based on a custom collector. It basically did what you suggest and to a certain degree it was much more flexible. It could basically "on the fly" handle discrete vs range facets. After all it was just loading (collecting) all possible values. The question and rub in our case is how to do this efficiently. The answer so far (and pre Lucene 5) was `FieldCache`s , but again this is something which we want and have to get away from. 

IMO there is no problem with exposing this encoding type. In fact I find it consistent with the latest changes we did for general indexing of numbers. Similarly numbers are now numerically encodied by default and one can force a string encoding by choosing an appropriate bridge. This is really the same thing. 

We have so far not been able to abstract the choice  of string vs numeric encoding away from the user and I am not sure we even can w/o restricting some use cases. 

> In the end, discrete values can be considered as a list of ranges of size 1.

It could be a set of discrete floating point values as well.

hferentschik:
> Basically what I described above: allow for facet queries on all the properties of an entity in a less verbose way than annotating each property.

Apart from the already mentioned index bloat (and probably runtime effects), there is also a question of consistency. There is also no class level annotation to mark all properties of a class for indexing. One needs to add explicit `@Field` annotations. I think `@Facet` belongs to the property level together with the `@Field` annotation and field bridges, etc. Introducing a class level facet annotation seems wrong.

Sanne:
I agree with the consistency argument, having an "everything is indexed" would be quite a change of direction.

Still, it's an interesting new perspective. It did strike me in other occasions that with Hibernate Search you need to craft the indexing mapping in advance, and only then the query can work, and you have to learn how to match them correctly.
It's definitely material for a different JIRA but it would be nice to keep in mind a "developer mode" in which we index everything (up to some reasonable level of nesting), so people can focus on developing the queries and playing with them, before having to face the index mapping problem.

In short, taking a good thing from SQL.

hferentschik:
Superseded  by #821

PR #821, opened 2015-04-17
HSEARCH-809 Updating faceting API

hferentschik:
Includes fixes for:

* HSEARCH-1600
* HSEARCH-809
* HSEARCH-900
* HSEARCH-812


Sanne:
Awesome :)

Could I extract and apply things like HSEARCH-1600 separately?

hferentschik:
> Could I extract and apply things like HSEARCH-1600 separately?

I am confused, HSEARCH-1600 is already separate. The faceting code has also already been reviewed in most parts. Can we not discuss this pull request as is.

Sanne:
I'm not understanding the Jenkins error. Documentation builds fine here..
I'll apply the commit for HSEARCH-1600 already just to get rid of the distraction, then I'll test the build of the manual again.

hferentschik:
You realize I pushed an update already? There was a problem with the doc ok build which I resolved. Did not show via asciidoctor. Just make sure to get the latest branch and take it from there.

Sanne:
right, thanks. I got confused as I had checked out the latest - which worked fine - just after you pushed apparently, as Jenkins finished the build marking it "green" again ten minutes later :)

Sanne:
I'd be very eager to merge this, not only because users are asking for it but also as I'd hope to use it as a trampoline to get to Lucene 5.

The big question is IMO about the backwards compatibility change in a minor release - by our own rules we can't do that, although I guess it's debatable if requiring the new annotations is indeed an API change.

I'd really like to say it's not and apply it, but I kinda have to admit it's because I want to, and not necessarily a fair judgement. Let's talk about it on the mailing list?

A safer approach could be to start a branch 6 already.. we could make good use of that to include Lucene 5 patches and ORM 5 too WIP too. But it will be a long wait to actually release a 6 since we want many more breaking changes in there. Let's chat/email about it with everyone next week?

hferentschik:
IMO we can merge and make this a feature of 5.2. As you say, a lot of people are waiting for this, in particular since it also solved the *ToMany faceting issue we always had with the collector approach. And even though one needs to add new annotations, the API itself is pretty much unchanged. Waiting for Search 6 seems to long of a wait for this feature. 

+1 for merging and releasing as part of 5.2

hferentschik:
What the plan on this now. IMO we should merge asap

Sanne:
I'll merge this later today, need to apply some changes before this.

hferentschik:
> I'll merge this later today, need to apply some changes before this.

Ok

Sanne:
merged!

PR #824, opened 2015-04-28
HSEARCH-1861 Enable IndexedEmbedded#includeEmbeddedObjectId to be enforced via SearchConfiguration Properties

s4ke:
https://hibernate.atlassian.net/browse/HSEARCH-1861

### DIFF
--- a/engine/src/main/java/org/hibernate/search/cfg/spi/impl/ExtendedSearchConfiguration.java
+++ b/engine/src/main/java/org/hibernate/search/cfg/spi/impl/ExtendedSearchConfiguration.java
@@ -1,0 +1,25 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.cfg.spi.impl;
+
+import org.hibernate.search.cfg.spi.SearchConfiguration;
+
+/**
+ * Additional SearchConfiguration-Features that are not that commonly used are in here. These methods are in this
+ * separate interface to not break exisiting integrator code
+ *
+ * @author Martin Braun
+ */
+public interface ExtendedSearchConfiguration extends SearchConfiguration {
+
+	/**
+	 * @return {@code true} if the value of all
+	 * {@link org.hibernate.search.annotations.IndexedEmbedded#includeEmbeddedObjectId} should be overwritten to true
+	 */
+	boolean isEnforceIncludeEmbeddedObjectId();
+
+}
--- a/engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
+++ b/engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
@@ -89,11 +89,18 @@ public class AnnotationMetadataProvider implements MetadataProvider {
 	private final ReflectionManager reflectionManager;
 	private final ConfigContext configContext;
 	private final BridgeFactory bridgeFactory;
+	private final boolean enforceIncludeEmbeddedObjectIds;
 
 	public AnnotationMetadataProvider(ReflectionManager reflectionManager, ConfigContext configContext) {
+		this( reflectionManager, configContext, false );
+	}
+
+	public AnnotationMetadataProvider(ReflectionManager reflectionManager, ConfigContext configContext,
+			boolean enforceIncludeEmbeddedObjectIds) {
 		this.reflectionManager = reflectionManager;
 		this.configContext = configContext;
 		this.bridgeFactory = new BridgeFactory( configContext.getServiceManager() );
+		this.enforceIncludeEmbeddedObjectIds = enforceIncludeEmbeddedObjectIds;
 	}
 
 	@Override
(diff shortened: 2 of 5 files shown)


hferentschik:
Not sure about this, should it not be a configuration error (speak exception) if you set the property and encounter `includeEmbeddedObjectId = false`? This can get very counter intuitive if the configuration via annotations tells you one thing, but something else is happening entirely.

s4ke:
Some backstory: In my Hibernate Search Adaption for any JPA provider I do asynchronous Index updating. Now, if a entity is deleted a DELETE event is logged. At the time this event is handled, there is no way of obtaining an object corresponding to that entity anymore so I need to use a DeleteByQuery (seems familiar :)?) to delete every occurence of the id in the index (every possible fieldName will be queried and the corresponding Document will be deleted).

Now, when annotating new Entities to work with this integration a user would need to set `includeEmbeddedObjectId = false` on EVERY SINGLE `@IndexedEmbedded` which in my book is annoying.

engine/src/main/java/org/hibernate/search/cfg/spi/impl/ExtendedSearchConfiguration.java
```diff
@@ -0,0 +1,25 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.cfg.spi.impl;
```
s4ke: @Sanne @hferentschik : should this stay in this package or maybe be moved to org.hibernate.search.cfg.spi?

hferentschik: -1 for moving. It is intentionally there

s4ke: Okay, I was just wondering whether I placed it correctly.

Sanne:
@hferentschik right and therefore I wouldn't add it to the public configuration properties, but only allow it to be overridden by custom integrators like @s4ke 's research project. That should be ok right?

hferentschik:
> right and therefore I wouldn't add it to the public configuration properties, but only allow it to be overridden by custom integrators like @s4ke 's research project.

not sure. I still introduces inconsistencies. I find it more transparent, if the docs would just say that if you want to use the Search with non Hibernate JPA providers, you needs to do this and that. At least it is transparent. And how many entities and relations does an average application really have? 
Transparently changing the default value is imo a potential source for trouble.

At the very least, it of course not be a public option and something should be logged. 

Personally, however, I would wait with this until users start complaining why they have to add too many `includeEmbeddedObjectId = false`. 

Let's keep changes to the core behavior (settings, defaults, etc) to a minimum to get this to work and then make incremental improvements, preferably based on user feedback.

s4ke:
@hferentschik: If someone is going to use Hibernate Search without ORM it is _mandatory_ for them to store the ids in the index. Therefore I as a user would find it quite unnecessary to add `includeEmbeddedObjectId = true`.

hferentschik:
> If someone is going to use Hibernate Search without ORM it is mandatory for them to store the ids in the index. Therefore I as a user would find it quite unnecessary to add includeEmbeddedObjectId = true.

It is necessary, because the default value of the annotation is `true`. Yes we can ignore that and enforce the behavior you want, but that will never be transparent and is ambiguous. We also, for example say that  a field you want to facet on cannot be tokenized. We could if we see a `@Facet` just ignore any indexing options which won't work for faceting, but that would again be ambiguous. I think one has to look at the configuration and behavior as a whole and not introduce some options for the sake of saving a bit of configuration. 

> Something will always have to be communicated to the user, be it that the includeEbmeddedObjectI is always true or that a user has to set it manually. But I think the latter one would be a bigger nuissance.

I disagree.

Also, not having the option for the first cut is the safest option. If you get feedback around how cumbersome the configuration is, you can add this improvement later. Adding something and then later taking it away is much harder.

s4ke:
> It is necessary, because the default value of the annotation is true.

The default value is false. That's the whole reason I need it.

> Yes we can ignore that and enforce the behavior you want, but that will never be transparent and is 
> ambiguous. We also, for example say that a field you want to facet on cannot be tokenized. 
> We could if we see a @Facet just ignore any indexing options which won't work for faceting, 
> but that would again be ambiguous.

Why shouldn't that be transparent? I mean when people want to use my project they will have to read up on the requirements anyway. And it doesn't do something for a specific case and something completely different for another, because when the option is set to true, every `includeEmbeddedObjectId` is set to true. That's where your `@Facet` example differs to this one.

> Also, not having the option for the first cut is the safest option. If you get feedback around how 
> cumbersome the configuration is, you can add this improvement later. Adding something and then 
> later taking it away is much harder.

I get your point, but for this one I don't think that the feature will ever be removed since it's a hard requirement that the embedded Object's id is stored in the index. I think I can live with this not being enabled for now, but I am almost 100% sure that we will get complaints :D.

hferentschik:
> The default value is false. That's the whole reason I need it.

Whatever. One way or the other you are changing the default. 

> I get your point, but for this one I don't think that the feature will ever be removed since it's a hard requirement that the embedded Object's id is stored in the index.

Not the fact that you need the embedded object is is what might change, but the fact whether or not the integrator should change the defaults. 

> I think I could live with this not being enabled for now 

Good.

s4ke:
> Whatever. One way or the other you are changing the default.

Just wanted to make that clear :)

> Not the fact that you need the embedded object is is what might change, but the fact whether or 
> not the integrator should change the defaults.

I get it. This _could_ clutter the integrator API unnecessarily.

I think I will keep this branch for now, and we'll see what happens in the future.

engine/src/main/java/org/hibernate/search/cfg/spi/impl/ExtendedSearchConfiguration.java
gunnarmorling: I don't think it's right to place a class in `impl` if it's intended by external components/integrators. Couldn't you just control this via a property? That wouldn't require any public API/SPI at all. If it's for experimenting for now, I wouldn't mind if there wasn't a constant for it.

PR #826, opened 2015-05-05
Hsearch 1848

Sanne:
# DO NOT MERGE

Removing most of the Infinispan related code; still keeping for now:
 - documentation
 - unit tests
 - integration tests
 - a deprecated version of the SPIs to make the constants still available

Because of those elements we've still not killed the circular dependency, but at least we start for a smooth transition.

While I don't think we should merge it now, I'd love to see if we're all on the same page on were this is going.


Sanne:
@emmanuelbernard do you agree to release w/o this PR or would you prefer us to consider including it?

It's tempting to include since it's well covered by integration tests, but the new module structure would need more than a trivial version bump to patch WildFly.

hferentschik:
> While I don't think we should merge it now

Why not?

> Because of those elements we've still not killed the circular dependency, but at least we start for a smooth transition.

So why not go the full monty and do a complete removal. Is this not what we have discussed all along?

emmanuelbernard:
Looks like a bit too much for a final release indeed,
Reading this PR, it also means that we will "break" applications by moving the Maven coordinates etc. Maybe better do it all in a 5.3.

> but the new module structure would need more than a trivial version bump to patch WildFly

I am not sure I follow entirely this, try and ping me. But if I understand you, we would be forced to keep the circularly dependent approach for WF 9 and won't likely fix this until WF 10? Or am I over interpreting?

Sanne:
> So why not go the full monty and do a complete removal. Is this not what we have discussed all along?

If we make any non-trivial changes compared to 'Beta1', I'll need another round of testing from all parties; that might get ugly. We already have the green light regarding 5.2.0.Beta1 + Infinispan 7.2 + WildFly 9 combination.
I'd rather release a 5.3.0.Beta1 *as well* and let's see if that works..

Sanne:
>  it also means that we will "break" applications by moving the Maven coordinates etc. Maybe better do it all in a 5.3.

No, our old Maven module still exists, it merely has a dependency to the new module. It's empty other than having the deprecated classes which define constants.

But indeed I plan to do better when the time comes to remove it and learn how to publish one of these:
http://maven.apache.org/guides/mini/guide-relocation.html

> But if I understand you, we would be forced to keep the circularly dependent approach for WF 9 and won't likely fix this until WF 10? Or am I over interpreting?

The affected modules are flagged "private" even in the application server and we can change them anytime. Actually the user won't even notice they are changed, but I would need to change the build scripts within WF9, which is a bit more complex than merely bump the version number of a structure which has been verified by other parties.

Sanne:
This is now merged

PR #831, opened 2015-06-01
Fixing Infinispan module, optionally depend on Tika

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-1880
 - https://hibernate.atlassian.net/browse/HSEARCH-1883

### DIFF
--- a/modules/pom.xml
+++ b/modules/pom.xml
@@ -23,6 +23,7 @@
         <aliases.xml.basedir>src/main/aliases</aliases.xml.basedir>
         <hibernate.search.version>${project.version}</hibernate.search.version>
         <hibernate.search.module.slot>${parsedVersion.majorVersion}.${parsedVersion.minorVersion}</hibernate.search.module.slot>
+        <infinispan.integration.slot>for-hibernatesearch-${hibernate.search.module.slot}</infinispan.integration.slot>
     </properties>
 
     <dependencies>
@@ -79,17 +80,6 @@
                         <goals>
                             <goal>single</goal>
                         </goals>
-                    </execution>
-                </executions>
-            </plugin>
-            <plugin>
-               <!-- Disable Enforcer as we're overriding the version to an external version of
-                org.hibernate:hibernate-search-infinispan -->
-                <artifactId>maven-enforcer-plugin</artifactId>
-                <executions>
-                    <execution>
-                        <id>enforce-rules</id>
-                        <phase>none</phase>
                     </execution>
                 </executions>
             </plugin>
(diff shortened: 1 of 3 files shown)


modules/pom.xml
```diff
@@ -23,6 +23,8 @@
         <aliases.xml.basedir>src/main/aliases</aliases.xml.basedir>
         <hibernate.search.version>${project.version}</hibernate.search.version>
         <hibernate.search.module.slot>${parsedVersion.majorVersion}.${parsedVersion.minorVersion}</hibernate.search.module.slot>
+        <infinispan.integration.slot>for-hibernatesearch-${hibernate.search.module.slot}</infinispan.integration.slot>
```
Sanne: @gustavonalle WDYT of this naming pattern to express the "provided for" idea?

The module below renders as

    <module name="org.infinispan.hibernate-search.directory-provider" slot="for-hibernatesearch-5.3" services="import" optional="true" />

gunnarmorling: What's the rationale for encoding the expected consumer in the name?

Sanne: https://hibernate.atlassian.net/browse/HSEARCH-1880?focusedCommentId=68345&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-68345

Sanne: In short, to break out of circular dependency, we should be releasing w/o assumptions on which Infinispan version is going to provide that module.
Encoding our version allows other projects to "provide something compatible to X", not least allows other projects to provide such modules for multiple versions (as needed and as tested).

gunnarmorling: Hum, ok. I am wondering why there is a dependency to a specific directory provider in the first place? I would rather expect some automatic discovery of providers present at runtime using some SPI discovery scheme?

Sanne: We do use SPI, but if you don't have it on your classpath you won't find it..

DavideD:
Merged https://github.com/hibernate/hibernate-search/pull/831

modules/pom.xml
gunnarmorling: Sure, my question more is why HS engine itself declares that dependency instead of an app using that provider. It's not that you can "know" each possible provider. Is it just as a convincience for this specific one which you expect to be used frequently?

Sanne: Right it's a convenience, but especially as if your app were to depend on that module you'd be forcing it to depend on a specific version of Infinispan.
By letting the module depend on it, it becomes an "implementation detail", allowing the user to use a different Infinispan version.

In this case it's a critical convenience, as Hibernate ORM will often use a different version of Infinispan for its caching purposes.. Finally, this way we can somewhat restrain the feature usage to specific versions we've tested as "fit for this purpose".

PR #833, opened 2015-06-02
HSEARCH-1882: Logging executed Lucene queries

gunnarmorling:


gunnarmorling:
@Sanne, is this one good to go?

Sanne:
yes it looks good to me, but @emmanuelbernard asked for it so I'll have him check it too.

gunnarmorling:
Hey @emmanuelbernard: Can we merge it before it's becoming one month of age ;) ?

emmanuelbernard:
I have the category anme comment but applied it nevertheless. We can discuss the name and decide to change it if needed.

Sanne:
Which alternative query category do you have in mind?

emmanuelbernard:
@Sanne something like that I think.

- `hibernate.search.database_query` (internal ones)
- `hibernate.search.faceting_query` (does that make sense?)
- `hibernate.search.autocomplete_query` (does that make sense?)

Sanne:
The Faceting in particular.. it needs to be run "within" a full-text query so not sure if we can separate it.
But the other are very concrete possibilities, so +1 to rename now and keep it stable down the road.

https://hibernate.atlassian.net/browse/HSEARCH-1909

PR #834, opened 2015-06-03
HSEARCH-1891 Spanning transaction per id list processed in batch indexing pipeline

gunnarmorling:


Sanne:
@gunnarmorling the commits should refer to HSEARCH-1891 right? (not HSEARCH-1791)

Sanne:
There's something wrong with exception handling, especially InterruptedException:

Method 'loadAllFromQueue' is allowed to catch the exception and reset the interrupted flag of the thread, as that would cause an early return from the loop and that honours the shutdown request for the job.

But method 'indexAllQueue' shouldn't be using the same pattern but should rather let the exception be unhandled, to make sure to jump out of the outher loops as well.

Ultimately this highlights that the transaction handling isn't cleaning up the transaction in case of exceptions: the commit/rollback would need to happen in a finally block to avoid leaks.

'rollbackTransaction()' : Why do you think rolling back is better than a commit in this case?
I understand it's just a read-only operation but why would that be better than a straight-forward commit?

gunnarmorling:
> the commits should refer to HSEARCH-1891 right

Ah, yes, that's true.

gunnarmorling:
> Ultimately this highlights that the transaction handling isn't cleaning up the transaction in case of exceptions

Pushed an update to address that.

> I understand it's just a read-only operation but why would that be better than a straight-forward commit?

It makes it very obvious to the reader that this TX is not meant to change anything. "Committing" a read-only TX seems odd to me, but really it's just a documentary preference, as you say there is no functional difference.

gunnarmorling:
Btw. this change renders `SessionAwareRunnable` un-used. Not sure whether to remove it or leave it for possible other use cases? Removing it simplifies the `OptionallyWrapInJTATransaction` a bit.

Sanne:
that's nice, I'd be happy to remove 'SessionAwareRunnable' and simplify 'OptionallyWrapInJTATransaction' as much as possible.

Sanne:
> "Committing" a read-only TX seems odd to me, but really it's just a documentary preference, as you say there is no functional difference.

I'm not the expert on those details, but it doesn't feel right to me. I understand that the end state of the database and the locks is the same, but the path to get there is different. At the very least, this will cause statistics to highlight a large number of rollbacks.
Is this common practice? The way I worked, people would strictly commit in such a case.

@emmanuelbernard ?

gunnarmorling:
> that's nice, I'd be happy to remove 'SessionAwareRunnable' and simplify 'OptionallyWrapInJTATransaction' as much as possible.

Cool; Pushed one more commit for that.

Sanne:
thanks @gunnarmorling !
I've added an additional commit - trivial enough to not need a review - to remove some dead code.

gunnarmorling:
Nice, thanks!

PR #835, opened 2015-06-03
Hibernate ORM 5

Sanne:
All main tests have been migrated, including the WildFly modules tests are running fine now but while that gives us good feedback, their setup complexity is high so I've disabled the release of the modules and added a highlight about this in the documentation.

Similarly the performance tests also require WildFly and have been disabled, in this case altogether, as it's easier to revert later and having them running now doesn't give me any value, at least the functional ones allow us to monitor integration changes in the next few weeks while ORM matures.

The tasks to re-enable bits and pieces are tracked at HSEARCH-1893: blocker before we do a Beta.

@gunnarmorling you might recognize some code :) Hopefully it's all in much better shape now, I reverted all cases of tests which had been disabled to be fixed later and actually fixed all similar TODOs we had in previous drafts.

# Some review guidelines

Many tests needed refactoring; unless they have a dedicated commit that has generally been a long but  braindead conversion so probably not too interesting to loose your sight on the details.
The gist is :
 - ORM does no longer take untyped Properties for configurations, so that affects all test configurations
 - it's strictly disallowed to reuse some Services
 - defined configurations are strictly immutable

I did consider a Properties conversion layer, but it was very tricky with our tests approach to guarantee the second and third points, so bit the bullet and made sure that each test has a single opportunity to define all configuration settings (per SessionFactory), can't change those settings anymore (the properties are no longer exposed) and finally that the configuration-defining function will be invoked again for each SessionFactory creation rather than try and cache or reuse things.
By doing so I found several issues :) especially as the testing framework is now much stricter on, for example but not least, unclosed Session instances.

Since I had to do a better job in guaranteeing resource closing, you might notice some methods now use the AutoClose syntax, finally available in ORM 5 ;-)

That's it for the "large volume" changes. The other commits are much more to the point and should be self-explanatory.


DavideD:
Merged https://github.com/hibernate/hibernate-search/pull/835

DavideD:
@Sanne Have you created all the related issues about re-enabling the modules and  removing the hard coded slot version in the test?

DavideD:
Could you link them to this issue as follow up?

Sanne:
> Could you link them to this issue as follow up?

I mentioned HSEARCH-1893 in the above description ;)
It's also mentioned as a code comment in several places.

PR #837, opened 2015-06-05
HSEARCH-1812 Document how to get the SearchIntegrator

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-1812

### DIFF
--- a/documentation/src/main/asciidoc/advanced-features.asciidoc
+++ b/documentation/src/main/asciidoc/advanced-features.asciidoc
@@ -5,13 +5,13 @@ In this final chapter we are offering a smörgåsbord of tips and tricks which m
 you dive deeper and deeper into Hibernate Search.
 
 [[AccessingSearchFactory]]
-=== Accessing the SearchFactory
+=== Accessing the `SearchFactory`
 
-The SearchFactory object keeps track of the underlying Lucene resources for Hibernate Search. It is
+The `SearchFactory` object keeps track of the underlying Lucene resources for Hibernate Search. It is
 a convenient way to access Lucene natively. The `SearchFactory` can be accessed from a
 FullTextSession:
 
-.Accessing the SearchFactory
+.Accessing the `SearchFactory`
 ====
 [source, JAVA]
 ----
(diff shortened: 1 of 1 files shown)


documentation/src/main/asciidoc/advanced-features.asciidoc
```diff
@@ -19,6 +20,22 @@ SearchFactory searchFactory = fullTextSession.getSearchFactory();
 ----
 ====
 
+[[AccessingSearchIntegrator]]
+=== Accessing the SearchIntegrator
```
hferentschik: `SearchIntegrator` - back quotes for formatting missing

documentation/src/main/asciidoc/advanced-features.asciidoc
```diff
@@ -19,6 +20,22 @@ SearchFactory searchFactory = fullTextSession.getSearchFactory();
...
+The SearchIntegrator is an object that gives access to lower level API of Hibernate Search.
```
hferentschik: The interface `SearchIntegrator` gives access to lower level APIs of Hibernate Search.

documentation/src/main/asciidoc/advanced-features.asciidoc
```diff
@@ -19,6 +20,22 @@ SearchFactory searchFactory = fullTextSession.getSearchFactory();
...
+Frameworks can use it to integrate with Hibernate Search. It should not be exposed as a
+public API.
```
hferentschik: "It should not be exposed as a public API"  - not sure what this is supposed to mean. It sounds like a recommendation to the user to now expose it as public API, but that is really not in the hands of the users. Given its package name (`org.hibernate.search.spi`), I would just say it is a SPI.

documentation/src/main/asciidoc/advanced-features.asciidoc
DavideD: `IndexReader` and `SearchFactory` appear in the same page and none of them use back quotes for formatting. Should it be changed everywhere?

documentation/src/main/asciidoc/advanced-features.asciidoc
DavideD: I don't understand how

> It should **not** be exposed as a public API

can be confused as a recommendation  to the user to now expose it as public API.

I will just remove this sentence anyway, it's related to frameworks and I don't think it's needed here.

documentation/src/main/asciidoc/advanced-features.asciidoc
Sanne: Yes ideally it should. But let's focus on your fix now.. if you want to change the others too, that's great but keep it as as a separate commit.

DavideD:
I've updated the documentation and added the backquotes where missing in a separate commit

DavideD:
Reopend...I've pressed the wrong button

documentation/src/main/asciidoc/advanced-features.asciidoc
hferentschik: > I will just remove this sentence anyway, it's related to frameworks and I don't think it's needed here.

+1

hferentschik:
Re-based and merged

PR #843, opened 2015-06-29
HSEARCH-1903

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-1903

PR #845, opened 2015-06-29
Speedup integration tests

Sanne:
The only purpose for this change is to reduce the execution time of wildfly tests.

revised/simplified version of PR #843


gunnarmorling:
Rebased and applied. Added one more commit to move one comment back to right place.

PR #846, opened 2015-06-29
The missing JGroups module and tests

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-1907

This might be troublesome on OSX.. I hope not?

### DIFF
--- a/integrationtest/wildfly/src/test/java/org/hibernate/search/test/integration/jgroups/JGroupsDeploymentHelper.java
+++ b/integrationtest/wildfly/src/test/java/org/hibernate/search/test/integration/jgroups/JGroupsDeploymentHelper.java
@@ -11,7 +11,7 @@ import static org.hibernate.search.test.integration.VersionTestHelper.addDepende
 import java.io.File;
 
 import org.hibernate.search.test.integration.VersionTestHelper;
-import org.hibernate.search.test.integration.jms.SearchNewEntityJmsMasterSlave;
+import org.hibernate.search.test.integration.jms.MasterSlaveTestTemplate;
 import org.hibernate.search.test.integration.jms.controller.RegistrationController;
 import org.hibernate.search.test.integration.jms.model.RegisteredMember;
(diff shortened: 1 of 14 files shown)


integrationtest/wildfly/src/test/java/org/hibernate/search/test/integration/jgroups/JGroupsDeploymentHelper.java
```diff
@@ -0,0 +1,129 @@
...
+		return commonUnitDef( name, "filesystem-slave", refreshPeriod, tmpDir )
+					.createProperty()
+						.name( "hibernate.search.default.worker.backend" )
+						.value( "jgroupsSlave" )
+						.up()
+					.up()
+				.up();
+	}
+
+	private static Properties<PersistenceUnit<PersistenceDescriptor>> commonUnitDef(
```
gunnarmorling: Out of curiosity, do you actually feel that this style of descriptor declaration is advantageous? Granted, it may help preventing some duplications, but personally I would be willing to accept those for the sake of a better readable actual `persistence.xml` file (prepped with variables via resource filtering as needed). Which also can be easily used as template for other examples or hints to users etc.

No need to change it, just interested in your opinion, maybe there are advantages I miss.

Sanne: I'm not against using the files, just that we would have three of them for
each master / slave test? But right they would probably be more readable
:-)

gunnarmorling:
> This might be troublesome on OSX.. I hope not?

It fails indeed for me on OS X. IIRC, some route config foo is needed? Could you share the link again? I can go and configure my machine, but obviously it's going to be an obstacle for contributors who may run into the same issue

Sanne:
Thanks for testing that. I don't remember the link, it's in the JGroups JIRA.. but don't bother, I agree with the obstacle point and will try to figure out an alternative.
So.. don't fix your machine yet or I run out of guinea pigs ;)

PR #853, opened 2015-07-02
HSERCH-1910 USe docVAlues instead of FieldCache for Spatial

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-1910


Sanne:
merged

PR #854, opened 2015-07-02
HSEARCH-1708 Throw an exception when sorting on a field not indexed

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-1708

### DIFF
--- a/engine/src/main/java/org/hibernate/search/query/engine/impl/HSQueryImpl.java
+++ b/engine/src/main/java/org/hibernate/search/query/engine/impl/HSQueryImpl.java
@@ -24,6 +24,7 @@ import org.apache.lucene.search.Explanation;
 import org.apache.lucene.search.Filter;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.Sort;
+import org.apache.lucene.search.SortField;
 import org.apache.lucene.search.TermQuery;
 import org.apache.lucene.search.similarities.Similarity;
 import org.hibernate.search.engine.ProjectionConstants;
@@ -43,6 +44,9 @@ import org.hibernate.search.filter.impl.ChainedFilter;
 import org.hibernate.search.filter.impl.DefaultFilterKey;
 import org.hibernate.search.filter.impl.FullTextFilterImpl;
 import org.hibernate.search.indexes.spi.IndexManager;
+import org.hibernate.search.metadata.FieldDescriptor;
+import org.hibernate.search.metadata.FieldSettingsDescriptor.Type;
+import org.hibernate.search.metadata.IndexedTypeDescriptor;
 import org.hibernate.search.query.engine.spi.DocumentExtractor;
 import org.hibernate.search.query.engine.spi.EntityInfo;
 import org.hibernate.search.query.engine.spi.HSQuery;
@@ -50,6 +54,7 @@ import org.hibernate.search.query.engine.spi.TimeoutExceptionFactory;
 import org.hibernate.search.query.engine.spi.TimeoutManager;
 import org.hibernate.search.reader.impl.MultiReaderFactory;
 import org.hibernate.search.spatial.Coordinates;
+import org.hibernate.search.spatial.DistanceSortField;
 import org.hibernate.search.spi.SearchIntegrator;
 import org.hibernate.search.store.IndexShardingStrategy;
 import org.hibernate.search.util.StringHelper;
@@ -638,24 +643,27 @@ public class HSQueryImpl implements HSQuery, Serializable {
 					false, false
 			);
 		}
-		else if ( this.sort != null && projection != null ) {
-			boolean activate = false;
-			for ( String field : projection ) {
-				if ( SCORE.equals( field ) ) {
-					activate = true;
-					break;
+		else if ( this.sort != null ) {
+			validateSortFields( extendedIntegrator );
+			if ( projection != null ) {
+				boolean activate = false;
+				for ( String field : projection ) {
+					if ( SCORE.equals( field ) ) {
+						activate = true;
+						break;
+					}
 				}
-			}
-			if ( activate ) {
-				return new LazyQueryState(
-						filteredQuery,
-						compoundReader,
-						searcherSimilarity,
-						extendedIntegrator,
-						classesAndSubclasses,
-						true,
-						false
-				);
+				if ( activate ) {
+					return new LazyQueryState(
+							filteredQuery,
+							compoundReader,
+							searcherSimilarity,
+							extendedIntegrator,
+							classesAndSubclasses,
+							true,
+							false
+					);
+				}
 			}
 		}
 		//default
@@ -670,7 +678,35 @@ public class HSQueryImpl implements HSQuery, Serializable {
 		);
 	}
 
+	private void validateSortFields(ExtendedSearchIntegrator extendedIntegrator) {
+		SortField[] sortFields = sort.getSort();
+		for ( SortField sortField : sortFields ) {
+			if ( sortField instanceof DistanceSortField ) {
+				validateDistanceSortField( extendedIntegrator, sortField );
+			}
+		}
+	}
 
+	private void validateDistanceSortField(ExtendedSearchIntegrator extendedIntegrator, SortField sortField) {
+		boolean indexedField = false;
+		String field = sortField.getField();
+		if ( field != null ) {
+			for ( Class<?> clazz : classesAndSubclasses ) {
+				IndexedTypeDescriptor indexedTypeDescriptor = extendedIntegrator.getIndexedTypeDescriptor( clazz );
+				FieldDescriptor fieldDescriptor = indexedTypeDescriptor.getIndexedField( field );
+				if ( fieldDescriptor != null ) {
+					indexedField = true;
+					if ( fieldDescriptor.getType() != Type.SPATIAL ) {
+						throw log.distanceSortRequiresSpatialField( field );
+					}
+					break;
+				}
+			}
+		}
+		if ( !indexedField ) {
+			throw log.sortRequiresIndexedField( sortField.getClass(), field );
+		}
+	}
 
 	private Similarity checkSimilarity(Similarity similarity, Similarity entitySimilarity) {
 		if ( similarity == null ) {
--- a/engine/src/main/java/org/hibernate/search/spatial/impl/CoordinateHelper.java
+++ b/engine/src/main/java/org/hibernate/search/spatial/impl/CoordinateHelper.java
@@ -1,0 +1,25 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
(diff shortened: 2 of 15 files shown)


engine/src/main/java/org/hibernate/search/query/engine/impl/HSQueryImpl.java
```diff
@@ -670,7 +680,51 @@ else if ( this.sort != null && projection != null ) {
...
+			if ( !indexedField ) {
+				throw log.cannotSortBecauseFieldIsNotIndexed( field );
+			}
+		}
+	}
 
+	private void validateDistanceSortField(String field, Class<?> clazz) {
+		EntityIndexBinding indexBinding = extendedIntegrator.getIndexBinding( clazz );
+		PropertyMetadata propertyMetadata = indexBinding.getDocumentBuilder().getMetadata().getPropertyMetadataForProperty( field );
+		if ( propertyMetadata != null ) {
```
hferentschik: Would it not make more sense to enhance the metadata itself so that the `PropertyMetadata` can already tell you whether you are dealing with coordinates? It seems wrong to work with reflection this late in the game. Also it is quite expensive.

hferentschik: If you check `DocumentFieldMetadata` it contains `isNumeric`. I would somehow expect that there is something like `isSpatial` as well.

engine/src/main/java/org/hibernate/search/query/engine/impl/HSQueryImpl.java
```diff
@@ -670,7 +680,51 @@ else if ( this.sort != null && projection != null ) {
...
+			}
+			if ( !indexedField ) {
+				throw log.cannotSortBecauseFieldIsNotIndexed( field );
+			}
+		}
+	}
 
+	private void validateDistanceSortField(String field, Class<?> clazz) {
+		EntityIndexBinding indexBinding = extendedIntegrator.getIndexBinding( clazz );
+		PropertyMetadata propertyMetadata = indexBinding.getDocumentBuilder().getMetadata().getPropertyMetadataForProperty( field );
```
Sanne: this line here is problematic, the field name is not the same as the property name.

engine/src/main/java/org/hibernate/search/query/engine/impl/HSQueryImpl.java
```diff
@@ -639,6 +648,7 @@ else if ( Boolean.FALSE.equals( forceScoring ) ) {
 			);
 		}
 		else if ( this.sort != null && projection != null ) {
+			validate( sort, classesAndSubclasses );
```
Sanne: Shouldn't we need to validate even if _sort_ is set but _projection_ is not?

engine/src/main/java/org/hibernate/search/query/engine/impl/HSQueryImpl.java
```diff
@@ -670,7 +680,51 @@ else if ( this.sort != null && projection != null ) {
...
 	}
 
+	private void validate(Sort sort, Collection<Class<?>> classes) {
+		SortField[] sortFields = sort.getSort();
+		for ( SortField sortField : sortFields ) {
+			String field = sortField.getField();
+			boolean indexedField = false;
+			for ( Class<?> clazz : classes ) {
+				IndexedTypeDescriptor indexedTypeDescriptor = extendedIntegrator.getIndexedTypeDescriptor( clazz );
+				FieldDescriptor fieldDescriptor = indexedTypeDescriptor.getIndexedField( field );
```
hferentschik: Here you are using the public metadata API opposed to the internal one (as you do later). `FieldDescriptor` seems to be the right spot to add additional metadata. There is for example already a `NumericFieldSettingsDescriptor`, maybe a `SpatialFieldSettingsDescriptor` would make sense. This way you might be able to skip accessing the internal metadata alltogether (obviously the internal metadata would need to change as well, since the public one is built from it, just saying you might be able to just use one type of API in this particular usecase)

engine/src/main/java/org/hibernate/search/query/engine/impl/HSQueryImpl.java
```diff
@@ -670,7 +680,51 @@ else if ( this.sort != null && projection != null ) {
...
+			}
+		}
+	}
 
+	private void validateDistanceSortField(String field, Class<?> clazz) {
+		EntityIndexBinding indexBinding = extendedIntegrator.getIndexBinding( clazz );
+		PropertyMetadata propertyMetadata = indexBinding.getDocumentBuilder().getMetadata().getPropertyMetadataForProperty( field );
+		if ( propertyMetadata != null ) {
+			// The field is a property of the entity
+			XClass elementClass = propertyMetadata.getPropertyAccessor().getElementClass();
```
Sanne: You shouldn't use the _XClass_ helpers here, as they will ignore other sources of metadata - for example the programmatic configuration.
Also, we shouldn't do reflection at runtime but somehow have information to validate this using our model.

DavideD:
Rebased and fixed.

I might have found a nicer solution: I validate the value when I read it; If it does not have a NumDocValues, it means that the field is invalid.

engine/src/main/java/org/hibernate/search/spatial/impl/SpatialIndexReader.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.lucene.index.NumericDocValues;
+import org.hibernate.search.util.logging.impl.Log;
+import org.hibernate.search.util.logging.impl.LoggerFactory;
+
+/**
+ * Utility class containing functions to read the value of a spatial field from an index.
+ *
+ * @author Davide D'Alto
+ */
+public final class SpatialIndexReader {
```
hferentschik: We usually use `XYZHelper` to indicate helper classes with static methods. So I'd prefer `SpatialIndexReaderHelper` or just `SpatialHelper`. This avoids also the initial thought that we are dealing with a subclass of Lucene's `IndexReader` which was my first thought.

hferentschik:
> I might have found a nicer solution: I validate the value when I read it; If it does not have a NumDocValues, it means that the field is invalid. 

But can the `NumDocValue` not also just indicate another (non spatial) related field? A spatial fields is for sure a numeric field, but not every numeric field is use as spatial coordinate. I still think the metadata approach is more promising, especially since it adds more value to the (metadadata) API as well.

Sanne:
+1 to use the metadata.. you might not be able to read it if the index is empty anyway ;)

DavideD:
Remarks applied and rebased

engine/src/main/java/org/hibernate/search/spatial/impl/CoordinateHelper.java
```diff
@@ -0,0 +1,33 @@
...
+ * double.
+ *
+ * @author Davide D'Alto
+ */
+public final class CoordinateHelper {
+
+	private CoordinateHelper() {
+	}
+
+	public static NumericDocValues loadCoordinateDocValues(AtomicReader atomicReader, String field) throws IOException {
```
gunnarmorling: What's the benefit of this over the direct method invocation?

engine/src/main/java/org/hibernate/search/query/engine/impl/HSQueryImpl.java
```diff
@@ -670,7 +678,32 @@ else if ( this.sort != null && projection != null ) {
...
+		SortField[] sortFields = sort.getSort();
+		for ( SortField sortField : sortFields ) {
+			if ( sortField instanceof DistanceSortField ) {
+				validateDistanceSortField( extendedIntegrator, sortField );
+			}
+		}
+	}
 
+	private void validateDistanceSortField(ExtendedSearchIntegrator extendedIntegrator, SortField sortField) {
+		String field = sortField.getField();
```
gunnarmorling: This can be null for some sorts (e.g. by doc id).

engine/src/main/java/org/hibernate/search/query/engine/impl/HSQueryImpl.java
```diff
@@ -670,7 +678,32 @@ else if ( this.sort != null && projection != null ) {
...
+			IndexedTypeDescriptor indexedTypeDescriptor = extendedIntegrator.getIndexedTypeDescriptor( clazz );
+			FieldDescriptor fieldDescriptor = indexedTypeDescriptor.getIndexedField( field );
+			if ( fieldDescriptor != null ) {
+				indexedField = true;
+				if ( fieldDescriptor.getType() != Type.SPATIAL ) {
+					throw log.distanceSortRequiresSpatialField( field );
+				}
+			}
+		}
+		if ( !indexedField ) {
```
gunnarmorling: The method name `validateDistanceSortField()` is a bit too specific with this check here.

gunnarmorling:
@DavideD, added some more minor comments. Looks alright apart from those.

engine/src/main/java/org/hibernate/search/spatial/impl/CoordinateHelper.java
DavideD: right,  initially it was longer but after some clean up it became smaller.
I'll remove it

engine/src/main/java/org/hibernate/search/query/engine/impl/HSQueryImpl.java
DavideD: Not necessarily.
Take for example `SortTest#testCustomFieldComparatorAscendingSort`, this part in particular:
```Sort sort = new Sort( new SortField( "sum", new SumFieldComparatorSource() ) );```

"sum" is not an indexed field, but everything still works because `SumFieldComparatorSource` does not use the passed field name.

So, at the moment, I don't feel like imposing this validation for all sorts.
Probably, this is something for a separate issue.

engine/src/main/java/org/hibernate/search/query/engine/impl/HSQueryImpl.java
DavideD: I would like to add a test for this scenario. Do I just need to specify the document id as sort field?

hferentschik:
> Looks alright apart from those.

+1 That's what I had in mind solution wise. I also think it is a nice improvement for the metadata API. I guess you should still follow up on @gunnarmorling comment regarding the sort field being potentially `null`.

DavideD:
Remarks applied + few other changes:

* Null check for sortField
* Remove method from Coordinateshelper
* Squashed commits
* Log message print the SortField class

engine/src/main/java/org/hibernate/search/query/engine/impl/HSQueryImpl.java
gunnarmorling: Yes, I think so.

gunnarmorling:
Rebased and applied. Nice work, @DavideD!

PR #857, opened 2015-07-03
HSEARCH-1908 Avoid the need to dynamically override the Analyzer to fieldname mapping

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-1908

This is all to avoid invoking
 IndexWriter.addDocument( document, analyzer )

as the only option in Lucene 5 is

 IndexWriter.addDocument( document )

### DIFF
--- a/engine/src/main/java/org/hibernate/search/backend/impl/lucene/analysis/ConcurrentlyMutableAnalyzer.java
+++ b/engine/src/main/java/org/hibernate/search/backend/impl/lucene/analysis/ConcurrentlyMutableAnalyzer.java
@@ -1,0 +1,89 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.impl.lucene.analysis;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.apache.lucene.analysis.Analyzer;
+import org.apache.lucene.analysis.DelegatingAnalyzerWrapper;
+import org.hibernate.search.exception.AssertionFailure;
+import org.hibernate.search.util.impl.ScopedAnalyzer;
+
+/**
+ * An Analyzer implementation which can dynamically be reconfigured. We use this as temporary solution to workaround
+ * LUCENE-6212, until a better solution is found.
+ *
+ * @author Sanne Grinovero <sanne@hibernate.org> (C) 2015 Red Hat Inc.
+ */
+public final class ConcurrentlyMutableAnalyzer extends DelegatingAnalyzerWrapper {
+
+	private final AtomicReference<ScopedAnalyzer> current = new AtomicReference<>();
+
+	public ConcurrentlyMutableAnalyzer(Analyzer initialAnalyzer) {
+		super( new ResettableReuseStrategy() );
+		if ( initialAnalyzer instanceof ScopedAnalyzer ) {
+			current.set( (ScopedAnalyzer) initialAnalyzer );
+		}
+		else {
+			current.set( new ScopedAnalyzer( initialAnalyzer ) );
+		}
+	}
+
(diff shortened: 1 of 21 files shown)


engine/src/main/java/org/hibernate/search/backend/impl/lucene/IndexWriterDriver.java
```diff
@@ -0,0 +1,91 @@
...
+import org.apache.lucene.document.Document;
+import org.apache.lucene.index.IndexWriter;
+import org.apache.lucene.index.Term;
+import org.apache.lucene.search.Query;
+import org.hibernate.search.backend.impl.lucene.analysis.ConcurrentlyMutableAnalyzer;
+import org.hibernate.search.util.impl.ScopedAnalyzer;
+
+/**
+ * Encapsulates various operations to be performed on a single IndexWriter.
+ * Avoid using {@link Workspace#getIndexWriter()} directly as it bypasses lifecycle
```
hferentschik: `Workspace` needs to be fully qualified to make this link work

engine/src/main/java/org/hibernate/search/backend/impl/lucene/IndexWriterDriver.java
```diff
@@ -0,0 +1,91 @@
...
+ * management of the IndexWriter such as reference counting, potentially leading to leaks.
+ *
+ * @author Sanne Grinovero <sanne@hibernate.org> (C) 2015 Red Hat Inc.
+ */
+public final class IndexWriterDriver {
```
hferentschik: Not sure I like the `Driver` part so much. Why not `IndexWriterDelegate`?

engine/src/main/java/org/hibernate/search/backend/impl/lucene/analysis/ConcurrentlyMutableAnalyzer.java
```diff
@@ -0,0 +1,89 @@
...
+	public boolean isCompatibleWith(ScopedAnalyzer analyzer) {
+		ScopedAnalyzer currentAnalyzer = current.get();
+		return currentAnalyzer.isCompositeOfSameInstances( analyzer );
+	}
+
+	/**
+	 * Correct concurrency control requires external locking!
+	 * @param analyzer
+	 */
+	public void deployNewAnalyzer(ScopedAnalyzer analyzer) {
```
hferentschik: `deploy`? Is it not just a simple `set(Scoped)Analyzer`?

hferentschik: or `updateAnalyzer`

hferentschik:
Looking good to me apart from the minor naming comments I made

engine/src/main/java/org/hibernate/search/backend/impl/lucene/IndexWriterDriver.java
Sanne: Ok, that's the same name I thought about too but I then changed it to `xDriver` as it does a bit more complex _driving_ of the IW. I'll change it to your suggestion, the Driver aspect is probably too unusual / flashy.

engine/src/main/java/org/hibernate/search/backend/impl/lucene/analysis/ConcurrentlyMutableAnalyzer.java
Sanne: I'll use `updateAnalyzer` as in various experiments this was doing much more complex things than just the _set_ it turned into.. that might still need to change, I wouldn't want to make a promise about it being just a set

engine/src/main/java/org/hibernate/search/backend/impl/lucene/IndexWriterDriver.java
Sanne: +1

Sanne:
Remarks applied; it took a while to rename the core class in the second commit, still have to figure out a clever way to do that. Maybe I should sed the patch files ;)

hferentschik:
Merged.

PR #858, opened 2015-07-03
HSEARCH-668 Transactional JMS

emmanuelbernard:
The tests don't work. It seems the JMS messages are not sent which leads to the test failures.
It could be for many reasons:

* test is not doing what I think it does
* configuration of the test is incorrect (anything to do to make a transactional queue on the client side?, ...)
* the actual core code does not do what I think it does
    * is the Hibernate flush actually called before the beforeTransaction of Hibernate Search
    * is the JMS calls to a transactional queue correct?

I need a second pair of eyes, I'm out of ideas on how and what to look.

### DIFF
--- a/backends/jms/src/main/java/org/hibernate/search/backend/jms/impl/JmsBackendQueueTask.java
+++ b/backends/jms/src/main/java/org/hibernate/search/backend/jms/impl/JmsBackendQueueTask.java
@@ -27,6 +27,7 @@ import org.hibernate.search.util.logging.impl.LoggerFactory;
 /**
  * @author Emmanuel Bernard
  * @author Sanne Grinovero <sanne@hibernate.org> (C) 2011 Red Hat Inc.
+ * @author Yoann Gendre
  */
 public class JmsBackendQueueTask implements Runnable {
 
@@ -60,23 +61,31 @@ public class JmsBackendQueueTask implements Runnable {
 		LuceneWorkSerializer serializer = indexManager.getSerializer();
 		byte[] data = serializer.toSerializedModel( filteredQueue );
 		QueueSender sender;
-		QueueSession session;
-		QueueConnection connection;
+		QueueSession session = null;
+		QueueConnection connection = null;
 		try {
 			connection = processor.getJMSConnection();
-			//TODO make transacted parameterized
-			session = connection.createQueueSession( false, QueueSession.AUTO_ACKNOWLEDGE );
+			// TODO make transacted parameterized
+			if ( processor.isTransactional() ) {
+				session = connection.createQueueSession( true, QueueSession.SESSION_TRANSACTED );
+			}
+			else {
+				session = connection.createQueueSession( false, QueueSession.AUTO_ACKNOWLEDGE );
+			}
 			ObjectMessage message = session.createObjectMessage();
 			message.setObject( data );
 			message.setStringProperty( Environment.INDEX_NAME_JMS_PROPERTY, indexName );
 
 			sender = session.createSender( processor.getJmsQueue() );
 			sender.send( message );
-
+			sender.close();
 			session.close();
 		}
 		catch (JMSException e) {
 			throw log.unableToSendJMSWork( indexName, processor.getJmsQueueName(), e );
+		}
+		finally {
+			processor.releaseJMSConnection( connection );
 		}
 	}
 }
--- a/engine/src/main/java/org/hibernate/search/backend/impl/InTransactionWorkQueueSynchronization.java
+++ b/engine/src/main/java/org/hibernate/search/backend/impl/InTransactionWorkQueueSynchronization.java
@@ -4,11 +4,10 @@
  * License: GNU Lesser General Public License (LGPL), version 2.1 or later
  * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
  */
+
 package org.hibernate.search.backend.impl;
 
 import java.util.concurrent.ConcurrentMap;
-
-import javax.transaction.Status;
 
 import org.hibernate.search.backend.spi.Work;
 import org.hibernate.search.engine.integration.impl.ExtendedSearchIntegrator;
@@ -16,11 +15,11 @@ import org.hibernate.search.util.logging.impl.Log;
 import org.hibernate.search.util.logging.impl.LoggerFactory;
 
 /**
- * Execute final work in the after transaction synchronization.
+ * Execute final work inside a transaction.
  *
  * @author Emmanuel Bernard
  */
-public class PostTransactionWorkQueueSynchronization implements WorkQueueSynchronization {
+public class InTransactionWorkQueueSynchronization implements WorkQueueSynchronization {
 
 	private static final Log log = LoggerFactory.make();
 
@@ -40,9 +39,9 @@ public class PostTransactionWorkQueueSynchronization implements WorkQueueSynchro
 	/**
 	 * in transaction work
 	 */
-	public PostTransactionWorkQueueSynchronization(Object transactionIdentifier, QueueingProcessor queueingProcessor,
-			ConcurrentMap<Object, WorkQueueSynchronization> queuePerTransaction,
-			ExtendedSearchIntegrator extendedIntegrator) {
+	public InTransactionWorkQueueSynchronization(Object transactionIdentifier, QueueingProcessor queueingProcessor,
+												ConcurrentMap<Object, WorkQueueSynchronization> queuePerTransaction,
+												ExtendedSearchIntegrator extendedIntegrator) {
 		this.transactionIdentifier = transactionIdentifier;
 		this.queueingProcessor = queueingProcessor;
 		this.queuePerTransaction = queuePerTransaction;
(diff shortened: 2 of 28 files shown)


integrationtest/wildfly/src/test/java/org/hibernate/search/test/integration/jms/transaction/TransactionalJmsMasterSlaveIT.java
```diff
@@ -0,0 +1,121 @@
...
+	private static final File tmpDir = RegistrationConfiguration.createTempDir();
+
+	/**
+	 * Lazy initialization of the libraries since Maven is painfully slow.
+	 */
+	private static class LibrariesLoader {
+		public static final File[] LIBRARIES = init();
+
+		private static File[] init() {
+			final String currentVersion = Version.getVersionString();
```
DavideD: I think you should use `VersionTestHelper` instead of `Version`

backends/jms/src/main/java/org/hibernate/search/backend/jms/impl/JmsBackendQueueTask.java
```diff
@@ -60,23 +61,30 @@ public void run() {
...
-		QueueConnection connection;
+		QueueSession session = null;
+		QueueConnection connection = null;
 		try {
 			connection = processor.getJMSConnection();
-			//TODO make transacted parameterized
-			session = connection.createQueueSession( false, QueueSession.AUTO_ACKNOWLEDGE );
+			// TODO make transacted parameterized
+			if ( processor.isTransactional() ) {
+				session = connection.createQueueSession( true, QueueSession.SESSION_TRANSACTED );
```
gunnarmorling: What be the motivation actually to use it without TX?

engine/src/main/java/org/hibernate/search/backend/impl/InTransactionWorkQueueSynchronization.java
```diff
@@ -0,0 +1,133 @@
...
+	public boolean isConsumed() {
+		return consumed;
+	}
+
+	@Override
+	public void beforeCompletion() {
+		// we are doing all the work in the before completion phase so that it is part of the transaction
+		try {
+			if ( prepared ) {
+				if ( log.isTraceEnabled() ) {
```
gunnarmorling: All the `isTraceEnabled()` business should not be needed if you just pas `is` as parameter value.

engine/src/main/java/org/hibernate/search/backend/impl/InTransactionWorkQueueSynchronization.java
```diff
@@ -0,0 +1,133 @@
...
+	@Override
+	public void afterCompletion(int i) {
+		try {
+			if ( Status.STATUS_COMMITTED == i ) {
+				if ( log.isTraceEnabled() ) {
+					log.tracef(
+							"Processing Transaction's afterCompletion() phase for %s. Performing work.", this.toString()
+					);
+				}
+				queueingProcessor.performWorks( queue );
```
gunnarmorling: Is it right you do `performWorks()` here again?

engine/src/main/java/org/hibernate/search/backend/impl/InTransactionWorkQueueSynchronization.java
```diff
@@ -0,0 +1,133 @@
...
+			//clean the Synchronization per Transaction
+			//not needed stricto sensus but a cleaner approach and faster than the GC
+			if ( queuePerTransaction != null ) {
+				queuePerTransaction.remove( transactionIdentifier );
+			}
+		}
+	}
+
+	public void flushWorks() {
+		WorkQueue subQueue = queue.splitQueue();
```
gunnarmorling: Is gained anything by doing thus flush? IIUC, if JMS is used transactionally, messages will be "sent" to clients only upon commit anyways?

Sanne: If you don't flush first, nothing will be in the JMS queue to send..

gunnarmorling: Why that? He is doing `prepareWorks()` + `performWorks()` in `beforeCompletion()`.

On a more general note, I am unsure why this internal queuing is done at all in the transactional case. Couldn't messages be sent straight to the JMS queue? If the TX is rolled back, the messages should be dropped as well. Maybe I am still missing a piece though. Either way it probably warrants some comments explaining the overall interaction of pieces.

Sanne: > Why that? He is doing prepareWorks() + performWorks() in beforeCompletion().

Ok I didn't read it all yet. But for sure we need to flush at least once "into" the queue.

> Couldn't messages be sent straight to the JMS queue?

That might be an option for some small and trivial session usage cases, but in the way we currently collect just references to the dirty indexed entities we do avoid re-processing a complex graph multiple times.. we only do it once, on the latest version. Not doing this two-phase was once a big problem as in some cases you'd have an exponential complexity and also potentially re-process expensive blobs to be indexed. The current design avoids that.. I've just spent 10 minutes on JIRA to find a reference and I'm sad I couldn't find it - maybe it was on the forums but we had a test contributed which would execute a commit in 35 minutes, which went back to 1 second after changing to the two-phase model :)
I'm sure I attached jprofiler graphs somewhere, can't find it now.. it's a quite old story.

backends/jms/src/main/java/org/hibernate/search/backend/jms/impl/JmsBackendQueueTask.java
emmanuelbernard: Not having to use 2PC and a few subtle things that might go wrong depending on how the JMS queue implementation and the JTA implementation works.

engine/src/main/java/org/hibernate/search/backend/impl/InTransactionWorkQueueSynchronization.java
emmanuelbernard: oops, copy paste.

engine/src/main/java/org/hibernate/search/backend/impl/InTransactionWorkQueueSynchronization.java
emmanuelbernard: this {{flushWork}} method is used by the explicit {{ftsession.flushToIndexes()}}, does your questions still stand?

emmanuelbernard:
I've rebased to fix the massive copy paste error. Now the {{afterCompletion}} code is empty

integrationtest/wildfly/src/test/java/org/hibernate/search/test/integration/jms/transaction/TransactionalJmsMasterSlaveIT.java
emmanuelbernard: `VersionTestHelper.injectVariables( "${dependency.version.HibernateSearch}" );`?
I don't see a non public method to access the Hibernate Search version

emmanuelbernard: looks like it did the trick and the failure in the IDE is now consistent with the CLI. I've made one of the private method public

engine/src/main/java/org/hibernate/search/backend/impl/InTransactionWorkQueueSynchronization.java
emmanuelbernard: I've added a commit to do this @gunnarmorling. Let me know if that's not what you had in mind.

emmanuelbernard:
Pushed by rebasing to latest master, fixing IDE tests and removing isTraceEnabled.
Still the tests fail and I don't know why. Help welcome.

engine/src/main/java/org/hibernate/search/backend/impl/InTransactionWorkQueueSynchronization.java
gunnarmorling: Looks good.

engine/src/main/java/org/hibernate/search/backend/impl/InTransactionWorkQueueSynchronization.java
gunnarmorling: > this flushWork method is used by the explicit ftsession.flushToIndexes(), does your questions still stand?

But what difference does that logic called upon expicit flushes make here? The messages will be submitted to the JMS queue but AFAICU not passed on to the consumer yet (which happens only upon TX commit). So is there any functional difference if `flushToIndexes()` was a no-op in this case?

gunnarmorling: > Not doing this two-phase was once a big problem as in some cases

Interesting. I agree each entity should only get one index update per TX, but I assumed this entire code would only be invoked once per changed entity anyways. I haven't fully understood why the manual queuing is needed for de-duplication, but I believe you it being the right thing ;)

emmanuelbernard: > > this flushWork method is used by the explicit ftsession.flushToIndexes(), does your questions still stand?

> But what difference does that logic called upon expicit flushes make here? The messages will be submitted to the JMS queue but AFAICU not passed on to the consumer yet (which happens only upon TX commit). So is there any functional difference if flushToIndexes() was a no-op in this case?

One use case is the use of session clear not to undo a work but to clear the memory usage (typical flush() then clear()). If you don't flushToIndexes() before the clear, you will not be able to build the index structure. There is a separate issue to consider merging the flush and flushToIndexes operations if the backend is transactional but we need to think about all the consequences.

Also, if the JMS backend or the JTA engine are doing passivation of onging tx work, calling fllushToIndexes frees memory usage.

emmanuelbernard:
I finally found the issue. It was a configuration problem as expected. One need to use the right JNDI bound XA JMS connection factory or else it does not do what you want silently. Of course, documentation on the internet is very sparse.

Unfortunately my test (test the non addition of an entry) takes ~35s extra on the test suite. If someone can find a better approach, that would be better.

The documentation is missing but I'd rather get a go (and even a push) before looking into this.
Documentation wise I think we need to say that:
* it

Sanne:
> One need to use the right JNDI bound XA JMS connection factory or else it does not do what you want silently.

In this case it's just our tests, and we could debug it. End users will have to match this up correctly using their own queue names.. is there really no way to WARN if we're taking the wrong kind from JNDI?

> only needed on JMS client side, the master does not need / should not have a tx worker

Shouldn't we make an option to make consumptions transactional as well? i.e. if the master takes a batch of Works `A`, then can not manage to flush that successfully to the IndexWriter, and fails, the next master node should still get `A` in his TODO list?
Certainly a different issue.

> The documentation is missing but I'd rather get a go (and even a push) before looking into this.

+1 to break up in other tasks

> If you look at the code, you'll see that a single property is necessary. Kind of nice.

Awesome!

integrationtest/wildfly/src/test/java/org/hibernate/search/test/integration/jms/transaction/TransactionalJmsMasterSlave.java
```diff
@@ -0,0 +1,179 @@
...
+		memberRegistration.rollbackedRegister();
+
+		assertNotNull( "A registered member should have an ID", newMember.getId() );
+	}
+
+	@Test
+	@InSequence(8)
+	@OperateOnDeployment("slave-2")
+	public void searchRollbackedMemberAfterSynchronizationOnSlave2() throws Exception {
+		// we need to explicitly wait because we need to detect the *effective* non operation execution (rollback)
```
DavideD: I see that you are only testing this on "slave-2", would it make sense to check from "slave-1" and  "master" as well?

DavideD:
superseeded by https://github.com/hibernate/hibernate-search/pull/862

integrationtest/wildfly/src/test/java/org/hibernate/search/test/integration/jms/transaction/TransactionalJmsMasterSlave.java
emmanuelbernard: No I don't see any reason. It would only slow down the test.

PR #859, opened 2015-07-06
HSEARCH-1681 Index optimisation should commit to publish the performed optimisation

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-1681


Sanne:
merged

PR #860, opened 2015-07-07
HSEARCH-1919 Avoid using FieldCache in SortTest

gunnarmorling:
@Sanne, another fallout of the L5 migration which can go right to master already.


Sanne:
merged, thanks

PR #861, opened 2015-07-08
HSEARCH-1922 Using standard JMS header JMSXGroupID for Environment.IN…

hferentschik:
…DEX_NAME_JMS_PROPERTY

Small change which will allow for a true HA JMS master/slave setup.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/cfg/Environment.java
+++ b/engine/src/main/java/org/hibernate/search/cfg/Environment.java
@@ -214,9 +214,10 @@ public final class Environment {
 	public static final String INDEX_MANAGER_IMPL_NAME = "indexmanager";
 
 	/**
-	 * Name of the JMS message property containing the index name to which to apply remote work.
+	 * Name of the JMS message property containing the index name to which to apply remote work. <i>JMSXGroupID</i> is
+	 * actually a standard JMS header field which is used for message grouping. See HSEARCH-1922.
 	 */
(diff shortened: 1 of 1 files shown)


Sanne:
Do you think it has any backwards compatibility impact?

I'm thinking about the case in which one upgrades the master, but still has pending messages in the queue (previously persisted, or still being produced).
These messages will have the index name encoded in a different property, and fail.

We could simply encode both for a transitionally time, and have the master code try read the new one as first option but fallback to the deprecated name?

hferentschik:
> I'm thinking about the case in which one upgrades the master, but still has pending messages in the queue (previously persisted, or still being produced).
These messages will have the index name encoded in a different property, and fail.

True, that would be a problem. But do we even support such a incremental upgrade. Also I expect that for our scenario the queue is not storing messages for a long time. The whole idea is to process them directly. Also, we are talking about a Lucene upgrade for the next version. If you ask me, I would stay away from any such incremental upgrade attempts. 

> We could simply encode both for a transitionally time, and have the master code try read the new one as first option but fallback to the deprecated name?

Sure, but that's again more code for imo minimal benefit.

Sanne:
I added two lines to fallback to the old property name, but I agree with you on not spending much more on it so I'll not add explicit upgrading tests nor bother you with an additional PR.. integrating!

hferentschik:
Fair enough.

PR #866, opened 2015-07-09
HSEARCH-1849

gunnarmorling:
@Sanne, everything compiles and tests pass as of this PR. Any feedback welcome.

I managed to remove the OpenBitSet stuff, DocIdBitSet is still there. I may find the time to look into it tomorrow, not sure yet though. So in case you'd like to take over, just let me know. You also may want to rework your earlier temp commit and merge it with the last one from me.

I think that's the last remaining issue, apart from that I think we are largely good to go, Refinements may be done in follow-up PRs IMO, unless you discover some major glitches.


Sanne:
Thanks @gunnarmorling ! No worries, I'll take over from here: you're off. Also, to be clear: current master is still 5.4, we should branch before merging this (however we can branch retroactively).
From my side, I had trouble with HSEARCH-668, not finished with that yet.

Sanne:
I've rebased this and will resume working on this

Sanne:
replaced by #880

PR #869, opened 2015-07-13
HSEARCH-1893 Fix and re-enable the performante integration tests

Sanne:
The missing task for HSEARCH-1893


Sanne:
I had to add the hibernate-search-testing dependencies to the Arquillian deployed performante test. It's using it, so I'm not sure why it would work before.
We had other cases in which WildFly10 shows to be much stricter than 8 and 9 with classloading, I guess it's one of those cases.

Sanne:
Seems it's not working correctly on Gunnar's machine:

    org.hibernate.search.exception.SearchException: Unable to create index directory: /path/to/Development/hibernate-search/integrationtest/performance/./org.hibernate.search.test.jpa.Bretzel for index org.hibernate.search.test.jpa.Bretzel"

Sanne:
@gunnarmorling turns out this seems to be a rather nasty bug we had and nobody noticed!

Sanne:
This is now https://hibernate.atlassian.net/browse/HSEARCH-1945

PR #874, opened 2015-07-22
HSEARCH-1936 do not force storeTermVectorPayloads flag when termVector is YES

bguillon-td:
@Sanne here is a pull request to fix HSEARCH-1936 I created recently. Not sure this fix won't create regressions but all unit tests still pass after the change.


Sanne:
Hi @bguillon-td , it looks correct, I'm merging it. Thanks a lot, good catch!

PR #876, opened 2015-07-29
HSEARCH-1940 Use Maven Relocation to deprecate the old hibernate-sear…

Sanne:
…ch artifactId

 - https://hibernate.atlassian.net/browse/HSEARCH-1940

PR #877, opened 2015-07-31
HSEARCH-1941 Passing a null to Search.getFullTextSession causes a con…

Sanne:
…fusing error message

 - https://hibernate.atlassian.net/browse/HSEARCH-1941


DavideD:
Merged https://github.com/hibernate/hibernate-search/pull/876

PR #879, opened 2015-08-03
Improve directory creation and validation, enable performance tests again

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-1932
 - https://hibernate.atlassian.net/browse/HSEARCH-1945

I didn't figure out a way to unit test the "." path resolution which wouldn't simply test that we use the API which we use.. a more reliable test is the fact that we re-enable the performance tests, which would previously fail because of the same "." resolution issue.

See https://github.com/hibernate/hibernate-search/pull/869 for the failure.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/store/impl/DirectoryProviderHelper.java
+++ b/engine/src/main/java/org/hibernate/search/store/impl/DirectoryProviderHelper.java
@@ -86,13 +86,13 @@ public final class DirectoryProviderHelper {
 			log.debug( "No root directory, go with relative " + relative );
 			sourceDirectory = FileSystems.getDefault().getPath( relative );
 			if ( Files.notExists( sourceDirectory ) ) {
-				throw new SearchException( "Source directory does not exist: " + relative );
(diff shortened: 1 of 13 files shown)


Sanne:
Marked as urgent as we'd like to merge this before #880

gunnarmorling:
@Sanne I see the following index directories created outside of `target`. I don't think that's intended:

    integrationtest/performance/ProductArticle/
    integrationtest/performance/org.hibernate.search.test.event.autoindexembeddable.Book/
    integrationtest/performance/org.hibernate.search.test.jpa.Bretzel/
    integrationtest/performance/org.hibernate.search.test.jpa.ToStringTest$Foo/
    integrationtest/performance/org.hibernate.search.test.query.timeout.Clock/
    integrationtest/performance/org.hibernate.search.test.spatial.POI/

gunnarmorling:
Merged. Looks good apart from those directories being created. Can you open a follow-up JIRA for fixing this, please?

Sanne:
Thanks!
As reference, it's https://hibernate.atlassian.net/browse/HSEARCH-1946

PR #880, opened 2015-08-03
Lucene5 reworked 3 august

Sanne:
I'm satisfied with this version. Please have a last review, but don't merge it yet!
We want to merge the 5.4 related patches first.


Sanne:
@gustavonalle this is the latest proposal, which we plan to release on wednesday if there are no major concerns. (Minor concerns will be resolved as follow-up issues)

gunnarmorling:
The latest commits (after mine) look good to me as well. It needs another rebase after the latest changes in 5.4. Can we remove all the deprecated methods using `File` now?

Sanne:
> The latest commits (after mine) look good to me as well.

right that's where to focus. Some early commits also changed, but mostly:
 - remove commits which are no longer required
 - fixed up some stuff to polish the patches (like remove TODO markers which have already been solved) - but no code changes which you didn't already see 

> Can we remove all the deprecated methods using File now?

+1 but as a next step? I'd be careful as some helper methods are often used by client code, were it's no major pain for us I'd rather keep those around for a while.

gustavonalle:
@Sanne I will test it today together with my Lucene 5 changes

Sanne:
!!  I'm rebasing this again and needs quite some conflicts fixed

engine/src/main/java/org/hibernate/search/reader/impl/ManagedMultiReader.java
```diff
@@ -25,15 +32,94 @@
...
+					break;
+					case STRING:
+					case STRING_VAL: mappings.put( sortField.getField(), Type.SORTED );
+					break;
+					case BYTES: mappings.put( sortField.getField(), Type.BINARY );
+					break;
+					case CUSTOM: // Nothing to do; expecting doc value fields created by the user
+					break;
+					default: log.sortFieldTypeUnsupported( sortField.getField(), sortField.getType() );
+				}
```
gustavonalle: NumericFields require Type.SORTED_SET type apparently

gustavonalle: Also, bridges are being ignored. My entity can have an ```int``` field and also use ```org.hibernate.search.bridge.builtin.IntegerBridge``` and it will end up being not stored as a ```NumericField```: in this case the Uninverter fails

gunnarmorling: To be sure, can you show an entity exposing this issue?

gustavonalle: https://github.com/infinispan/infinispan/blob/master/query/src/test/java/org/infinispan/query/test/Person.java

Sanne:
Closing, replaced by PR #885

PR #887, opened 2015-08-11
HSEARCH-1947 Java 8 Date and Time

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-1947

This patch will enforce the use of JDK 8 for the build and add the new bridges under the package:
`org.hibernate.search.bridge.builtin.time.impl`

After the discussion on the mailing list ,and off-line with @Sanne I've created bridges for:

* `LocalDate`, `LocalTime`, `Year`, `YearMonth`, `MonthDay`:
  they get stored as numeric fields where the digits represents the year, month or day of the date.
  For example, 2012-12-1 becomes the long 20121201
* `Instant`, `LocalDateTime`: they are stored as number of milliseconds from epoch.
* `Duration`: duration in milliseconds
* `ZoneId` and `ZoneOffset`: they get store as string 

Note that the precision for `Duration`, `LocalDateTime` and `Instant` is the millisecond.
This should cover most cases.

None of this bridges support `@DateBridge`. It seems there  is no real need for the user to select the encoding, and the new java api provides methods and classes for date and time truncation. Maybe it could be useful for the time to have a @TimeBridge annotation to set the resolution using `java.time.temporal.ChronoUnit`. A user can always create a custom bridge or we can add feature if they are requested.

We decided to not create a bridge for OffsetDateTime, OffsetTime, ZonedDateTIme and so on.
Probably the most useful way to store the offset and the zone is in separte fields, but with the current bridges the user can store these information as separate properties of the entity.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/DurationNumericBridge.java
+++ b/engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/DurationNumericBridge.java
@@ -1,0 +1,50 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.bridge.builtin.time.impl;
+
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+
+/**
+ * Converts an {@link Duration} to the number of milliseconds from the epoch of 1970-01-01T00:00:00Z.
+ * <p>
+ * If the duration is too large to fit in a {@code long} milliseconds, then an
+ * exception is thrown.
+ * <p>
+ * If the duration has greater than millisecond precision, then the conversion will drop any excess precision information
+ * as though the amount in nanoseconds was subject to integer division by one million.
+ * <p>
+ * The biggest instant that can be converted using {@link Long#MAX_VALUE} is w * The biggest instant that can be converted using {@link Long#MAX_VALUE} is 292278994-08-17T07:12:55.807Z.
+ * <p>
+ * The biggest duration that can be converted using {@link Long#MAX_VALUE} is PT2562047788015H12M55.807S.
+ * The smallest duration that can be converted using {@link Long#MIN_VALUE} is PT-2562047788015H-12M-55.808S.
+ *
+ * @author Davide D'Alto
+ */
+public class DurationNumericBridge extends JavaTimeNumericBridge<Duration, Long> {
+
+	public static final DurationNumericBridge INSTANCE = new DurationNumericBridge();
+
+	private DurationNumericBridge() {
+	}
+
+	/**
+	 * @see Duration#toMillis()
+	 */
+	@Override
+	public Long encode(Duration accessor) {
+		return accessor.toMillis();
+	}
+
+	/**
+	 * @see Duration#of(long, java.time.temporal.TemporalUnit)
+	 */
+	@Override
+	public Duration decode(Long value) {
+		return Duration.of( value, ChronoUnit.MILLIS );
+	}
+}
--- a/engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/LocalDateNumericBridge.java
+++ b/engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/LocalDateNumericBridge.java
@@ -1,0 +1,52 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.bridge.builtin.time.impl;
+
+import java.time.LocalDate;
+
+/**
+ * Converts a {@link LocalDate} to a {@link Long}.
+ * <p>
+ * The digits of the resulting number will have the following pattern: yyyyMMdd
+ * <p>
+ * For example, the date 2001-3-25 will become the number 20010325.
+ * <p>
+ * The values during encoding and decoding must be between the boundaries of a {@link LocalDate};
+ * if some of the values are out of range, a {@link java.time.DateTimeException} is thrown.
+ *
+ * @author Davide D'Alto
+ */
+public class LocalDateNumericBridge extends JavaTimeNumericBridge<LocalDate, Long> {
+
+	public static final LocalDateNumericBridge INSTANCE = new LocalDateNumericBridge();
+
+	private LocalDateNumericBridge() {
+	}
+
+	@Override
+	public Long encode(LocalDate date) {
+		long year = (long) date.getYear() * 10_000L;
+		long month = (long) date.getMonthValue() * 100L;
+		long day = (long) date.getDayOfMonth();
+
+		return year < 0
+				? year - month - day
+				: year + month + day;
+	}
+
+	@Override
+	public LocalDate decode(Long number) {
+		long absValue = Math.abs( number );
+		int year = (int) ( absValue / 10_000L );
+		int month = (int) ( ( absValue / 100L ) % 100L );
+		int day = (int) ( absValue % 100 );
+
+		return number.longValue() < 0
+				? LocalDate.of( -year, month, day )
+				: LocalDate.of( year, month, day );
+	}
+}
--- a/engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/LocalDateTimeNumericBridge.java
+++ b/engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/LocalDateTimeNumericBridge.java
@@ -1,0 +1,45 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.bridge.builtin.time.impl;
+
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+
+/**
+ * Converts a {@link LocalDateTime} to number of millisecond from Epoch using a UTC offset.
+ * <p>
+ * The LocalDateTime is converted to an {@link Instant} using {@link ZoneOffset#UTC}:
+ * <p>
+ * If the instant represents a point on the time-line too far in the future or past to fit in a {@code long}
+ * milliseconds, then an exception is thrown.
+ * <p>
+ * If the instant has greater than millisecond precision, then the conversion will drop any excess precision information
+ * as though the amount in nanoseconds was subject to integer division by one million.
+ * <p>
+ * The biggest date that can be converted using {@link Long#MAX_VALUE} is 292278994-08-17T07:12:55.807.
+ * The smallest date that can be converted using {@link Long#MIN_VALUE} is -292275055-05-16T16:47:04.192.
+ *
+ * @author Davide D'Alto
+ */
+public class LocalDateTimeNumericBridge extends JavaTimeNumericBridge<LocalDateTime, Long> {
+
+	public static final LocalDateTimeNumericBridge INSTANCE = new LocalDateTimeNumericBridge();
+
+	private LocalDateTimeNumericBridge() {
+	}
+
+	@Override
+	public Long encode(LocalDateTime dateTime) {
+		return dateTime.atOffset( ZoneOffset.UTC ).toInstant().toEpochMilli();
+	}
+
+	@Override
+	public LocalDateTime decode(Long value) {
+		return Instant.ofEpochMilli( value ).atOffset( ZoneOffset.UTC ).toLocalDateTime();
+	}
+}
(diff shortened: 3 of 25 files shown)


DavideD:
For the other type that we cannot/won't convert to numeric field, should we provide a StringBridge? I'm talking about Period, Offset*, Zoned*?

DavideD:
For some of this types, it is also prossible to index null as a value out of the allowed type (for example Long.MAX), should we do  that?

Sanne:
+1 to review this now, but let's fix https://hibernate.atlassian.net/browse/HSEARCH-1954 before merging this

Sanne:
> For some of this types, it is also prossible to index null as a value out of the allowed type (for example Long.MAX), should we do that?

It's an interesting but I'd like to finish my experimentation with null tokens - going on in parallel before answering that. Could you track that as as follow up JIRA for this and mark it as a blocker for 5.5.0.Beta1 ?

engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/DurationNumericBridge.java
```diff
@@ -0,0 +1,50 @@
...
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.bridge.builtin.time.impl;
+
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+
+/**
+ * Converts an {@link Duration} to the number of milliseconds from the epoch of 1970-01-01T00:00:00Z.
```
gunnarmorling: A duration is not related to the epoch, right? Rather it's stored as the duration's length in ms.

engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/DurationNumericBridge.java
```diff
@@ -0,0 +1,50 @@
...
+ * <p>
+ * If the duration is too large to fit in a {@code long} milliseconds, then an
+ * exception is thrown.
+ * <p>
+ * If the duration has greater than millisecond precision, then the conversion will drop any excess precision information
+ * as though the amount in nanoseconds was subject to integer division by one million.
+ * <p>
+ * The biggest instant that can be converted using {@link Long#MAX_VALUE} is w * The biggest instant that can be converted using {@link Long#MAX_VALUE} is 292278994-08-17T07:12:55.807Z.
```
gunnarmorling: Instant?

engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/LocalDateNumericBridge.java
```diff
@@ -0,0 +1,52 @@
...
+package org.hibernate.search.bridge.builtin.time.impl;
+
+import java.time.LocalDate;
+
+/**
+ * Converts a {@link LocalDate} to a {@link Long}.
+ * <p>
+ * The digits of the resulting number will have the following pattern: yyyyMMdd
+ * <p>
+ * For example, the date 2001-3-25 will become the number 20010325.
```
gunnarmorling: I feel it's very confusing to store this as a numeric value that way, as it mixes different numbering systems. Saving it as a String seems preferably to me.

engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/LocalDateTimeNumericBridge.java
```diff
@@ -0,0 +1,45 @@
...
+ * <p>
+ * If the instant has greater than millisecond precision, then the conversion will drop any excess precision information
+ * as though the amount in nanoseconds was subject to integer division by one million.
+ * <p>
+ * The biggest date that can be converted using {@link Long#MAX_VALUE} is 292278994-08-17T07:12:55.807.
+ * The smallest date that can be converted using {@link Long#MIN_VALUE} is -292275055-05-16T16:47:04.192.
+ *
+ * @author Davide D'Alto
+ */
+public class LocalDateTimeNumericBridge extends JavaTimeNumericBridge<LocalDateTime, Long> {
```
gunnarmorling: Shouldn't the strategies for `LocalDate` and `LocalDateTime` be the same more ore less? What's the reasoning for storing them that differently?

engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/LocalTimeNumericBridge.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.bridge.builtin.time.impl;
+
+import java.time.LocalTime;
+
+/**
+ * Converts a {@link LocalTime} to a {@link Long}.
+ * <p>
+ * The digits of the resulting number will have the following pattern: hhMMssnnnnnnnnn
```
gunnarmorling: I would store that rather as a String.

engine/src/main/java/org/hibernate/search/bridge/util/impl/NumericFieldUtils.java
```diff
@@ -24,6 +41,9 @@
 
 	private static final Log log = LoggerFactory.make();
 
+	private static final Class<?>[] NUMERIC_JAVA_TIME_TYPES = { Year.class, Instant.class, LocalDate.class, LocalDateTime.class, YearMonth.class,
```
gunnarmorling: Does that work on JDKs < 8? I'd assume the class load + static initialisation will stumble upon these types here on older JDKs?

engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
```diff
@@ -1058,7 +1058,15 @@ private boolean isNumericField(NumericField numericFieldAnnotation, FieldBridge
 
 		// either @NumericField is specified explicitly or we are dealing with a implicit numeric value encoded via a numeric
 		// field bridge
-		return numericFieldAnnotation != null || fieldBridge instanceof NumericFieldBridge || fieldBridge instanceof NumericEncodingDateBridge;
+		return numericFieldAnnotation != null
+				|| fieldBridge instanceof NumericFieldBridge
+				|| fieldBridge instanceof NumericEncodingDateBridge
```
gunnarmorling: Could `JavaTimeNumericBridge` be a `NumericFieldBridge`?

engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/DurationNumericBridge.java
DavideD: I thinnk this is a copy and paste error, I will fix it

engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/DurationNumericBridge.java
DavideD: I thinnk this is a copy and paste error, I will fix it

gunnarmorling:
@DavideD Added some comments inline.

> LocalDate, LocalTime, Year, YearMonth, MonthDay: they get stored as numeric fields where the digits represents the year, month or day of the date. For example, 2012-12-1 becomes the long 20121201

Storing these as numbers that way seems not right to me, it's source of potential usage mistakes (e.g. consider date arithmetics). I think they should either be stored as millis since epoch in UTC or as String.

> Instant, LocalDateTime: they are stored as number of milliseconds from epoch.

How comes `LocaleDateTime` is handled differently than the ones above? They seem to fall into the same category.

> Note that the precision for Duration, LocalDateTime and Instant is the millisecond.
> This should cover most cases.

I guess so. Or we could store the nano-seconds part optionally in a second field?

> Maybe it could be useful for the time to have a @TimeBridge annotation to set the resolution using java.time.temporal.ChronoUnit.

That's a nice idea.

We decided to not create a bridge for OffsetDateTime, OffsetTime, ZonedDateTIme and so on.

Hum, but isn't `ZonedDateTime` a very central type in the time API? Not supporting it seems like a shortcoming and usage issue to me. Similar to `Calendar` so far it could be stored in millis since epoch (projecting it back would be done in UTC) or as an ISO8601-formatted String, including the TZ information.

engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/LocalDateNumericBridge.java
DavideD: > I feel it's very confusing to store this as a numeric value that way,

Why do you think it is confusing?
This solution has the readability of the string approach plus the performance of the numeric approach. Seems the best of the two worlds to me.

engine/src/main/java/org/hibernate/search/bridge/util/impl/NumericFieldUtils.java
DavideD: I didn't have error when building with JDK7.
Isn't static only initialized when you call it? This shouldn't get called if you are using JDK7

DavideD: I will double check though

engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/LocalDateNumericBridge.java
DavideD: Also by using this approach you are not affected by problems that you would have when adding and substracting dates from different time zone

engine/src/main/java/org/hibernate/search/bridge/util/impl/NumericFieldUtils.java
Sanne: it's not initialized when you call it but when the class is initialized: i.e. whenever any code calls any method of _NumericFieldUtils_ .

> I didn't have error when building with JDK7.

How do you build with JDK7?

engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/LocalDateNumericBridge.java
gunnarmorling: My concern is that this pattern may trick people into doing wrong
assumptions and date calculations based on the indexed values, e.g.
20120901-20120830 to determine the number of days between two dates.
Admittedly it's unlikely that someone accesses the raw indexed value, but
still it seems problematic to me. It's a bit like storing numbers to the
power of 8 in a numeric field to the power of 10.

engine/src/main/java/org/hibernate/search/bridge/util/impl/NumericFieldUtils.java
DavideD: Right, I was making confusion.

I didn't build it with JDK 7

engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/LocalDateNumericBridge.java
Sanne: interesting POV. I had suggested the numeric encoding to Davide as it's more efficient from both space and query perspective.
When we do projections - or anything - we return the original type to the user, so this would only be problematic with low-level raw access (like you say): I'm not sure we should encourage that? 

It boils down to a need of project mission statement I think; we should decide if we're *using* the index to provide additional powerful query capabilities, or *map* to the index and allow the user to choose their own mapping.

I'd be inclined to say that we don't do mapping and that "raw index access" is at your own risk...

engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
DavideD: probably, I'll check

engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/LocalDateNumericBridge.java
gunnarmorling: I'd agree as far as the current Lucene backends are concerned, but even then I find the proposed mapping a bit weird. It's just not the "universal" way for persisting date information. If you are concerned about performance, then we should go for the millis in epoch approach. Personally I'd prefer strings, though.

But it's a different story when taking ES into the picture. There are much higher odds people will access the index directly through the REST API. There we also have to consider inter-op with other apps accessing the index and really should go for a more standard mapping (Dates are ISO-String-based in ES).

engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/LocalDateTimeNumericBridge.java
DavideD: I thought, LocalDateTime was too big to be represented as number.
I was wrong as it can be expressed as double or float.

engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/LocalDateNumericBridge.java
Sanne: Ok let's go with a String then. Can we make sure the padding is going to be "universally right" in terms of sorting?

Sanne: > But it's a different story when taking ES into the picture. 

right in the ES / Solr case we can no longer assume to fully own the schema.. that's a big change in mindset.

DavideD: The biggest year and nano are [-]999_999_999, for LocalDate* it should not be a problem.

DavideD:
There are many things to change in this pull request.
I'll send a new one

PR #889, opened 2015-08-13
HSEARCH-1954 Solve issue with JDK 8

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-1954

This pull request changes all the javadocs that are causing exceptions during the javadoc generation.
It also enforce the use of JDK 8,

I'm still not sure how we plan to check if everything works with JDK7 without the creation of a separate project.

I will update the wiki and the website if the pull is merged.

### DIFF
--- a/backends/jgroups/src/main/java/org/hibernate/search/backend/jgroups/impl/AutoNodeSelector.java
+++ b/backends/jgroups/src/main/java/org/hibernate/search/backend/jgroups/impl/AutoNodeSelector.java
@@ -22,7 +22,7 @@ import org.jgroups.View;
  * <p>Limitation: make sure all nodes in the group are having the same application running
  * and use the same configuration, or the master might ignore incoming messages.</p>
  *
- * @author Sanne Grinovero <sanne@hibernate.org> (C) 2012 Red Hat Inc.
+ * @author Sanne Grinovero (C) 2012 Red Hat Inc.
  */
 public class AutoNodeSelector implements NodeSelectorStrategy {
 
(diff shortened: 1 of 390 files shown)


Sanne:
The first replacement went wrong: surname is not "Bernard>"

Sanne:
on top of the very minor comments I made above, there are a couple more tasks to be done:

### profiles cleanup
some profiles get only activate on minimum JDK8, these could be simplified (not longer need them as a profile, default would be good enough).

### Wiki updates
The how-to-contribute wikis and the "how to release" wikis need to be updated.

### Jenkins jobs
We'll have to disable the JDK7 specific jobs, and reconfigure the "main" jobs to use JDK8. Better to do that before we merge it to prevent failure notifications.

Sanne:
@DavideD I'll do the website updates and the Jenkins jobs.. can you update this PR related to the other things?

Optional: do you think there's some good part in these (thinking javadoc fixes) which we could apply to 5.4 w/o too much trouble? worth trying?

DavideD:
> @DavideD I'll do the website updates and the Jenkins jobs.. can you update this PR related to the other things?

Yep, thanks

> Optional: do you think there's some good part in these (thinking javadoc fixes) which we could apply to 5.4 w/o too much trouble? worth trying?

Except for the last 2 commits, all the changes are related to the javadocs, so I'm not expecting to many issues.  I'll try it

DavideD:
I've applied the changes and rebased:
* Fix author for Emmanuel
* Fix changes to the pom alignement
* Integration tests with WildFly are now part of the default build

Sanne:
Thanks, CI jobs should be updated.. merged it, let's see the build outcomes.

PR #890, opened 2015-08-14
30% Review - Enabling to use ElasticSearch as indexing/query backend

gunnarmorling:
@Sanne, @hferentschik, @DavideD et al. this is the request for early feedback on the ES backend. At this point I am interested in feedback on the overall approach and strategy, so please no nit-picking about formatting, JavaDoc and the like. And yes, there are CS violations ;)

Some thoughts on the approach:

* This implements an `IndexManager`+ `LuceneWorkVisitor` on the indexing side, and `HSQuery` on the query side
* Queries are supported in form of the ES query DSL (JSON notation and the more concise, yet less powerful "query string syntax")
* It uses JEST as a client, so to avoid coupling with specific ES/Lucene versions; JEST is supposedly compatible with several versions of the ES REST API
* At creation time of the index manager the ES index + type mapping is created, later at run time data to be indexed is posted, adhering to the schemas created upfront
* The entity id is used as ES document id (not to be mixed up with the internal Lucene document id!)
* Indexing is quite featureful: proper type mapping, field boosting, embeddables etc. Querying also with paging, sorting, scrolling, projection
* I aimed at altering no SPI contracts incompatbly in HSEARCH; this works so far, but it requires some nasty hacks to make things fit

Open issues and next steps:
* Refine currently implemented things (e.g. "index null as", make timeouts configurable etc.)
* Cover more features as far as possible: faceting, filtering, multi-tenancy etc. Some are "just" work to be done, not sure about conceptual mismatches in other cases, that needs more investigation
* Re-work current IM to use a BackendQueryProcessor so one e.g. can use the combo of JMS (e.g. via a persistent queue) + ES; Currently the IM sends requests straight to ES
* Investigate more performant (bulk) indexing/scrolling via suitable APIs in ES
* Index sharing
* Allow to pass in queries via ES query builder (making it an *optional* dependency, so users can add them in the right version if the want to)

Longer term tasks are:
* Rework HS SPIs to make things work out better for non-Lucene backends; E.g. I create temporary `Document` instances atm. as that's what field bridges operate on; A more abstract representation would be useful here to avoid this kind of hacks
* Generalize HS query DSL so it's not bound to Lucene queries but can emit other (ES) query representations as well
* ... ?

Ok, let the battle begin, or "Hashime" as they say, but be nice everyone :)


Sanne:
> The entity id is used as ES document id (not to be mixed up with the internal Lucene document id!)

You mean our "normal Id logic applies" right? I.e. I might need to override them with @DocumentId, and we don't want to depend on JPA.

Sanne:
An important one: the ElasticSearch "backend" and the helpers to run queries on it should not depend on packages such as

 - org.hibernate:hibernate-search-orm
 - org.hibernate:hibernate-search-orm / testing classifier
 - org.hibernate:hibernate-testing
 - JPA interfaces
 - etc...

We want to be able to use that w/o requiring an Hibernate ORM/OGM or JPA application.

gunnarmorling:
> You mean our "normal Id logic applies" right? I.e. I might need to
override them with @DocumentId, and we don't want to depend on JPA.

Yes, exactly.

gunnarmorling:
> An important one: the ElasticSearch "backend" and the helpers to run
queries on it should not depend on packages such as

Good point. It should be doable by shuffling around some parts.

Sanne:
I wanted to play more with it but don't want to install ES at your same IP address so had some fun myself too ;-)
https://github.com/gunnarmorling/hibernate-search/pull/1

gunnarmorling:
> An important one: the ElasticSearch "backend" and the helpers to run queries on it should not depend on packages such as

Thinking about it, having these as _test_ dependencies should be no problem, right? Otherwise, testing will become a bit tedious.

hferentschik:
> Thinking about it, having these as test dependencies should be no problem, right? Otherwise, testing will become a bit tedious.

+1

Sanne:
> Thinking about it, having these as test dependencies should be no problem, right? Otherwise, testing will become a bit tedious.

Sure not a problem to have it as a test dependency if you prefer that.

gunnarmorling:
I've pushed some more commits to make null tokens, class-level bridges and some things more work.

Any thoughts?

gunnarmorling:
@Sanne, @hferentschik, @DavideD, ANY feedback at all? I know we all are busy, but this PR is sitting here now for almost a week and I really would like to make sure we agree on the large picture of this. Thanks!

Sanne:
I'm sorry, you're right. Back on it again right now..

Sanne:
Hi @gunnarmorling , looks like ES2 was released now: https://www.elastic.co/blog/elasticsearch-2-0-0-released

You think you could update this and get the integration tests to download/start ES automatically?
I remember sending you an experimental PR to get that done but I think you couldn't use that approach because ES was still in flux.

gunnarmorling:
Hey @Sanne, ys, saw it and am about to do the update. Also will update those integration tests.

But please let's move forward with this then, this PR is sitting here since August :/ I know I didn't push for it myself either, but it's depressing how long it takes us sometimes to integrate such useful additions.

gunnarmorling:
@Sanne Rebased and updated to ES 2.0. The Maven build also starts/stops an ES node before/after test execution now. If you want to use a different ES host, configure it in hibernate.properties.

Sanne:
I started by running the tests, and it's failing `JGroupsElasticSearchTest`; I assume because it has a hard-coded IP address.

But looking at it, I'm confused on the purpose of the tests as it's mixing both JGroups and ES. What's the architecture you have in mind?

gunnarmorling:
@Sanne, it relates to the split of means of transport (local/JGroups/JMS) and means of indexing (Lucene/ES) I also would like to address with #896 (we should discuss that at the Paris meeting btw.) 

Mixing JGroups and ES indeed is not going to be super-helpful, it's more JMS + ES which I consider as a very useful variant to address situations where a remote ES is temporarily unavailable and the indexing tasks are automatically re-sent later on by means of a persistent JMS queue. I chose the JGroups + ES mix only for testing purposes as JGroups seemed simpler to set up than JMS.

So this particular test generally shows that the ES index manager can be used with other means of transport than local. #896 seeks to further simplify that, e.g. the ES index manager class would look simpler then, but as said we could discuss that during the meeting.

I hope this clarifies things a bit?

Sanne:
Thanks yes that helps. And sorry, I realize you had already explained that now but I didn't remember the idea when looking at the test.

I do think your JMS+ES idea is interesting, but we need to start breaking this all down in smaller pieces.
Could we rework the separation of transport vs IndexManager at a second time?

I'm asking you as I didn't figure out yet if this a necessary step for the core of the issue "get stuff indexed in ES". If you think it's necessary then I guess we need to discuss the architecture in detail now. I'd hope this could be treated as a separate thing.

I'd hope to get a bit more practical than discuss architecture: what about we start merging some of this?
For example the first commit: I see no reason to not open a JIRA right now "Avoiding dependency to Lucene's Query type  in TimeoutManagerImpl" and send a PR for that.
Incidentally, that's a small one but I'd ask you to make the toString evaluation of the Query a lazy evaluation. Now I do suspect that you're not looking forward for that kind of detailed feedback within this large PR :) So that's why I'd break down those little blockers and try to get rid of them in break-out issues: that's the goal of 5.6, to collect all changes already which are not API breaking..

If you could extract several of those and send them as PR, then rebase this, rinse and repeat I think we should be able to integrate most of it and so have a "natural selection" of what needs to be postponed for 6?

In the meantime, I think the next steps on the "mainline work" should be to focus on the testing utilities and integration test fixtures. For example that hardcoded IP address in the JGroups test?

gunnarmorling:
> but we need to start breaking this all down in smaller pieces.
> Could we rework the separation of transport vs IndexManager at a second time?

Really that's the case already :) There is my other PR #896 for doing this separation.

In the PR at bay nothing really has changed in this regard (which is why I had to create an ES-specific backend queue processor, which would not have been necessary with that split already in place). I've only replaced some references to `DirectoryBasedIndexManager` with `IndexManager` at places which should be agnostic of the specific implementation (`BackendQueueProcessor` contract).

gunnarmorling:
> If you could extract several of those and send them as PR, then rebase this, rinse and repeat I think we should be able to integrate most of it and so have a "natural selection" of what needs to be postponed for 6?

Ok, I can try and extract some PRs for the preparatory commits at the beginning. But also note that the entire change is meant to be acceptable for 5.6, there should be no breakages really (with the only exception being - but I will double check - the change to `BackendQueueProcessor` which I'd deem acceptable given it should rarely, if at all, be implemented by clients). I did not change `FieldBridge` etc. for exactly that reason, that's the sort of changes which should go into 6.

gunnarmorling:
> For example that hardcoded IP address in the JGroups test?

Ah yes, that's an oversight. There is a property for this now used in the other tests, I just forgot to remove that fixed setting from the JGroups test.

gunnarmorling:
@Sanne Force-pushed an update. ES is now started/stopped for the test in the JGroups module, too. The host is configured through hibernate.properties, no need to change it normally.

gunnarmorling:
Rebased to latest master and force-pushed. I am continuing now with executing the "orm" module test suite and addressing its test failures (as far as they apply to ES being used as a backend) one by one.

I keep that separate from that PR though in order to not let it grow indefinite. Let me know in case you are ready for more :)

Sanne:
It now passes tests on my machine! Nice progress

But it still fails on jenkins.. could you have a look at these?
http://ci.hibernate.org/view/Pull%20Requests/job/hibernate-search-PR/1008/

gunnarmorling:
Jenkins, retest this please.

gunnarmorling:
> It now passes tests on my machine!

Cool, glad to hear!

> But it still fails on jenkins.. could you have a look at these?

Weird. Looking into it...

gunnarmorling:
@Sanne: Force-pushed an update, the module's build passes on CI now, too. There is just this intermittent test failure `ScheduledCommitPolicyTest` which we really should fix or disable.

Sanne:
now checkstyle is complaining.
Yea, about that failure there's a JIRA open for 5.6.. I haven't looked at it yet but I feel it might be a race in the test.

Sanne:
BTW ScheduledCommitPolicyTest never fails me locally

gunnarmorling:
Ok, now it's all green on CI.

Sanne:
> this is the request for early feedback on the ES backend. At this point I am interested in feedback on the overall approach and strategy, so please no nit-picking about formatting, JavaDoc and the like. And yes, there are CS violations ;)

Do you think I could consider it for merging at this stage?

gunnarmorling:
Pushed another commit for adding some to-do markers. IMO it's good to push with that; Not yet for releasing, I'll send another PR with many improvements based on the current one.

Sanne:
it's merged!

PR #894, opened 2015-08-18
HSERCH-1946 left directories after test

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-1946

The problem is caused by some persistence unit that are included in the testing jar.

### DIFF
--- a/integrationtest/performance/src/test/java/org/hibernate/search/test/performance/TestRunnerStandalone.java
+++ b/integrationtest/performance/src/test/java/org/hibernate/search/test/performance/TestRunnerStandalone.java
@@ -22,7 +22,6 @@ import org.junit.Test;
 /**
  * @author Tomas Hradec
  */
-@SuppressWarnings("deprecation")
 public class TestRunnerStandalone {
 
 	private final TestScenario scenario = TestScenarioFactory.create();
--- a/pom.xml
+++ b/pom.xml
@@ -715,6 +715,7 @@
                             <excludes>
                                 <exclude>**/hibernate.properties</exclude>
                                 <exclude>**/log4j.properties</exclude>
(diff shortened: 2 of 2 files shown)


DavideD:
@Sanne do you also want the change for the selection of the target dir in this pull request?

The problem didn't seem related to that so I haven't committed it

DavideD:
Actually I jsut reilized that this pull request is not complited.
It works for the default case though

DavideD:
Ok it should work for the other scenarios as well

Sanne:
Strange, I totally forgot that there is a persistence.xml in the testing jar. Shouldn't we remove that instead? It might throw off people tryingto use our testing utilities too.

How does it cause this problem?

Sanne:
What would you think of this alternative fix: 
 - https://github.com/hibernate/hibernate-search/commit/afdd14f059867c36631eb414cd195c010154bfef

I think I like it better as others won't risk to fall in the same trap ("others" could be other people using our testing utilities, or other modules we work on)

DavideD:
I've rebase the pull request to use your commit.
I just changed the commit message.

Is that ok, to merge?

Sanne:
+1 yes please fix my commit message
You'll remove your second commit right?

DavideD:
Merged https://github.com/hibernate/hibernate-search/pull/894

@Sanne The issue on the tracker seems bigger that the problem solved by this pull request, should I leave it close it and create a more specific pull request (read the target dir from the property)?

Sanne:
no, looks good enough, I've resolved it. Thanks!

Sanne:
(In case I ever have the need to run it on another drive, I'll open a new RFE)

PR #895, opened 2015-08-19
HSERACH-1947 Bridge for classes in java.time

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-1947

The bridges are now going to convert the objects to string padded with zero to allow the sorting.
The provider will check for the types on the classpath instead of checking for the jdk version

By the way, do I need to use the ClassLoader  service to check for the presence of a class or can I use the Class.forName(,..)? I'm asking because in NumericFieldUtils I don't have access to the ClassLoaderService at the moment.

Previous pull request: https://github.com/hibernate/hibernate-search/pull/887

### DIFF
--- a/engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/DurationBridge.java
+++ b/engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/DurationBridge.java
@@ -1,0 +1,63 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.bridge.builtin.time.impl;
+
+import java.time.Duration;
+import java.util.Locale;
+
+import org.hibernate.search.bridge.TwoWayStringBridge;
+import org.hibernate.search.util.logging.impl.Log;
+import org.hibernate.search.util.logging.impl.LoggerFactory;
+
+/**
+ * Converts a {@link Duration} to a {@link String}.
+ * <p>
+ * The string is obtained concatenating the number of seconds with the nano of seconds.
+ * The values are padded with 0 to allow field sorting.
+ *
+ * @author Davide D'Alto
+ */
+public class DurationBridge implements TwoWayStringBridge {
+
+	private static final Log log = LoggerFactory.make();
+	private static final int SECONDS_PADDING = 20;
+	private static final int NANOS_PADDING = 9;
+	private static final String FORMAT = "%+0" + SECONDS_PADDING + "d%0" + NANOS_PADDING + "d";
+
+	public static final DurationBridge INSTANCE = new DurationBridge();
+
+	private DurationBridge() {
+	}
+
+	@Override
+	public String objectToString(Object object) {
+		if ( object == null ) {
+			return null;
+		}
+
+		Duration duration = (Duration) object;
+		long seconds = duration.getSeconds();
+		int nano = duration.getNano();
+		return String.format( Locale.ROOT, FORMAT, seconds, nano );
+	}
+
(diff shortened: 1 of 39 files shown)


hferentschik:
> By the way, do I need to use the ClassLoader service to check for the presence of a class or can I use the Class.forName(,..)

I guess the intention is to always use the service. In this case you are looking for Java classes so Class.forName should be fine as well, I guess. Preferably `ClassLoaderService` though imo.

DavideD:
The idea behind using the Class.forName was to be able to add the classes for other version of JDK.
While I was testing that everything is working fine with a separate project, I discovered that the backports have a different package name. @Sanne: should I create a BridgeProvider (and relative bridges) for the backport as well? I can probably resuse most of the code, so tit should not take too much time

This is the link to the backport page: http://www.threeten.org/
The backport sources: https://github.com/ThreeTen/threetenbp

Sanne:
I just checked this build using the demo as suggested at https://hibernate.atlassian.net/browse/HSEARCH-1969?focusedCommentId=69872&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-69872

and found that this is introducing a hard dependency on the Time types, we should fix that first.

DavideD:
I made a change yesterday evening, could you make sure that you have the latest branch and using the latest building. The problem was that I was catching the wrong exception.

I'm asking because I've tried the demo and it seems to work.

Sanne:
> I made a change yesterday evening, 

Yeah sorry my snapshot build was from early yesterday. I'll try again now.

engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/DurationBridge.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Converts a {@link Duration} to a {@link String}.
+ * <p>
+ * The string is obtained concatenating the number of seconds with the nano of seconds.
+ * The values are padded with 0 to allow field sorting.
+ *
+ * @author Davide D'Alto
+ */
+public class DurationBridge implements TwoWayStringBridge {
```
gunnarmorling: As durations are naturally a numeric value, I think it makes more sense to map it to a numeric field really.

engine/src/main/java/org/hibernate/search/bridge/builtin/time/impl/InstantBridge.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Converts a {@link Instant} to a {@link String}.
+ * <p>
+ * The string is obtained concatenating the number of seconds from Epoch with the nano of seconds.
+ * The values are padded with 0 to allow field sorting.
+ *
+ * @author Davide D'Alto
+ */
+public class InstantBridge extends TemporalAccessorStringBridge<Instant> {
```
gunnarmorling: The same for instants; Storing it as a numeric field seems more appropriate to me.

engine/src/main/java/org/hibernate/search/bridge/impl/JavaTimeBridgeProvider.java
```diff
@@ -0,0 +1,140 @@
...
+
+	/**
+	 * @return {@code true} if at least one of the supported classes in the package java.time exists on the classpath,
+	 * {@code false} otherwise.
+	 */
+	public boolean hasFoundSomeJavaTimeTypes() {
+		return !builtInBridges.isEmpty();
+	}
+
+	private static boolean classExists(ClassLoaderService classLoaderService, String string) {
```
gunnarmorling: Better name it "className" than "string".

DavideD:
I've pushed an update, now the following type are converted to numeric fields:
- Instant: truncated to the milliseconds from Epoch
- Duration: in nanoseconds, throw exception if it cannot be represented as long
- Year as integer

I'm updating the documentation ...

Sanne:
rebasing this

PR #896, opened 2015-08-20
HSEARCH-1970 Untangling IndexManager and BackendQueueProcessor contracts

gunnarmorling:
@Sanne, @hferentschik This PR does the untangling of `(DirectoryBased)IndexManager` and `BackendQueueProcessor` we discussed. Responsibilities are much nicer defined that way and it will e.g. allow to use the ES index manager via JMS without further changes in the new ES index manager being required.

On a tangent this change also allowed to remove some methods from `BQP` which did not really make sense there (lock handling).

Of course it's a braking change to existing `IndexManager` and `BackendQueueProcessor` implementations. Not sure how much that is in reality as I'd expect those contracts only to be implemented by rather advanced integrators?

Anyways, let me know what you think. Maybe it's not as overwhelming as the ES PR :)

### DIFF
--- a/engine/src/main/java/org/hibernate/search/backend/impl/lucene/WorkspaceHolder.java
+++ b/engine/src/main/java/org/hibernate/search/backend/impl/lucene/WorkspaceHolder.java
@@ -13,7 +13,6 @@ import java.util.concurrent.locks.Lock;
 import org.hibernate.search.backend.BackendFactory;
 import org.hibernate.search.backend.IndexingMonitor;
 import org.hibernate.search.backend.LuceneWork;
-import org.hibernate.search.backend.spi.BackendQueueProcessor;
 import org.hibernate.search.indexes.spi.DirectoryBasedIndexManager;
 import org.hibernate.search.indexes.spi.IndexManager;
 import org.hibernate.search.spi.WorkerBuildContext;
@@ -30,7 +29,7 @@ import org.hibernate.search.util.logging.impl.LoggerFactory;
  * @author Emmanuel Bernard
  * @author Sanne Grinovero
  */
-public class LuceneBackendQueueProcessor implements BackendQueueProcessor {
+public class WorkspaceHolder {
 
 	private static final Log log = LoggerFactory.make();
 
@@ -40,7 +39,6 @@ public class LuceneBackendQueueProcessor implements BackendQueueProcessor {
 	private LuceneBackendTaskStreamer streamWorker;
 	private WorkProcessor workProcessor;
 
-	@Override
 	public void initialize(Properties props, WorkerBuildContext context, IndexManager indexManager) {
 		sync = BackendFactory.isConfiguredAsSync( props );
 		if ( workspaceOverride == null ) {
@@ -63,13 +61,11 @@ public class LuceneBackendQueueProcessor implements BackendQueueProcessor {
 		}
 	}
 
-	@Override
 	public void close() {
 		resources.shutdown();
 		workProcessor.shutdown();
 	}
 
-	@Override
 	public void applyStreamWork(LuceneWork singleOperation, IndexingMonitor monitor) {
 		if ( singleOperation == null ) {
 			throw new IllegalArgumentException( "singleOperation should not be null" );
@@ -77,7 +73,6 @@ public class LuceneBackendQueueProcessor implements BackendQueueProcessor {
 		streamWorker.doWork( singleOperation, monitor );
 	}
 
-	@Override
 	public void applyWork(List<LuceneWork> workList, IndexingMonitor monitor) {
 		if ( workList == null ) {
 			throw new IllegalArgumentException( "workList should not be null" );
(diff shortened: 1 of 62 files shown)


engine/src/main/java/org/hibernate/search/backend/impl/lucene/WorkspaceHolder.java
```diff
@@ -29,7 +29,7 @@
  * @author Emmanuel Bernard
  * @author Sanne Grinovero
  */
-public class LuceneBackendQueueProcessor implements BackendQueueProcessor {
+public class WorkspaceHolder {
```
gunnarmorling: This might even be pulled into `DirectoryBasedIndexManager`. I left it separately for now to keep the diff a bit nicer.

engine/src/main/java/org/hibernate/search/backend/spi/BackendQueueProcessor.java
```diff
@@ -60,17 +59,6 @@
 	void applyStreamWork(LuceneWork singleOperation, IndexingMonitor monitor);
 
 	/**
-	 * @return a Lock instance which will block index modifications when acquired
-	 */
-	Lock getExclusiveWriteLock();
```
gunnarmorling: Not needed anymore, as `DirectoryBasedIndexManager` only needs this internally and invokes that functionality on the `WorkspaceHolder`.

engine/src/main/java/org/hibernate/search/indexes/impl/IndexManagerHolder.java
```diff
@@ -211,6 +213,14 @@ public IndexManager getIndexManager(String targetIndexName) {
 		return indexManagersRegistry.get( targetIndexName );
 	}
 
+	public BackendQueueProcessor getBackendQueueProcessor(String indexName) {
```
gunnarmorling: I also considered a class `Backend` comprising the index manager and the queue processor of a given index but I decided against it as that would have been much more invasive.

engine/src/main/java/org/hibernate/search/indexes/spi/IndexManager.java
```diff
@@ -119,10 +117,4 @@
 	 * To optimize the underlying index. Some implementations might ignore the request, if it doesn't apply to them.
 	 */
 	void optimize();
-
-	/**
-	 * @return the Serializer implementation used for this IndexManager
-	 */
-	LuceneWorkSerializer getSerializer();
```
gunnarmorling: Made a service, as it's not needed for users of the `IndexManager` contract. It's only needed internally within remoting  `BackendQueueProcessor` implementations. They also could manage the serializer themselves, but then I thought that's what services are for in the end.

engine/src/main/java/org/hibernate/search/indexes/impl/IndexManagerHolder.java
hferentschik: But potentially a new idea

engine/src/main/java/org/hibernate/search/indexes/spi/IndexManager.java
hferentschik: +1

hferentschik:
I like. I did not look at all the details, but looking at the various commits + rough browsing of the code I give this a +1

hferentschik:
I like these fallouts from your ES prototype :-)

gunnarmorling:
> where exactly is the code which makes this decision?

It's specific to BQP implementations. E.g. for JMS the BQP always will send a JMS message, whereas for JGroups there is a if/else in its BQP which either serializes + sends or invokes locally depending on whether it's the master node or not.

In engine it's `WorkQueuePerIndexSplitter` which triggers the BQP as per the index config which then does the required things from there.

gunnarmorling:
> I like these fallouts from your ES prototype :-)

Glad to hear that. Indeed I am quite happy with that change, it straightens out some stumbling blocks.

gunnarmorling:
@Sanne, any thoughts? Would be a good step into the right direction, also taking the new DB backend into account, which should be usable with ES.

Sanne:
Let's not merge such changes between a CR and the Final, we'll resume this in the 5.6 timeframe.

gunnarmorling:
@Sanne, what is the rationale for the "Need wait for another PR" label?

Sanne:
@gunnarmorling 

> what is the rationale for the "Need wait for another PR" label?

I just meant to postpone this for after the 5.5.0.Final release, reverted the label now.

gunnarmorling:
Hey @Sanne, did you have a chance to test this with ISPN query? Just again stumbled upon the unhappy correlation of backend queue processor and index manager during the ES follow-up work: The "leaking queue processor" was tied to Lucene - which it doesn't really have to because for op counting the actual backend doesn't matter.

Essentially, there is a circular dependency between the two things: BQP expects IM in `initialize()`, and IMs are supposed to interact with BQP. This circle causes several issues, amongst them the exposure of methods on the respective contracts which were not needed otherwise. This PR gets rid of this circle, by making it a one-way relationship: BQP knows the concept of IM but not the other way around.

I still think this change makes things easier to grasp and combine and thus would love to see it merged. Let's come to a final verdict please :)

gunnarmorling:
@Sanne Any progress testing this? Please decide one way or the other, these long radio silences are killing me ;)

gunnarmorling:
Closing this in favor of the rebased #1042.

PR #898, opened 2015-08-21
Fixjavadocs

DavideD:
@Sanne This should fix the javadocs errors.

Could you have  a look at the link I changed? I'm not sure I've added the right class


Sanne:
Thanks! yes that's the right types.. merged.

PR #899, opened 2015-08-26
HSEARCH-1973 ConnectedMultiFieldsTermQueryBuilder must unwrap a NullE…

anistor:
…ncodingTwoWayFieldBridge before checking for NumericFieldBridge

Jira: https://hibernate.atlassian.net/browse/HSEARCH-1973


Sanne:
Hi @anistor you're right the fix seems trivial and this class already seems to imply supporting a Numeric field, but I'm having an hard time to figure when in practice this combination would occur.

Am I correct to guess that the Parser generates the equivalent of something like

    queryBuilder.keyword().onField( "age" ).matching( [int] ).createQuery() 

?

My doubt really is on the *keyword()* clause, which strongly suggests the user is forcing a *TermQuery*.

hferentschik:
Is it not really so that numeric fields can only be targeted by a range 
query?
Using a term query on a numerical encoded field will cause problems further 
down
the line.

gunnarmorling:
A test would be nice. 

Specifically, does it work as expected for a query such as

    queryBuilder.keyword().onField( "age" ).matching( null ).createQuery() 

? Or is one supposed to pass the null token as String in that case?

Sanne:
I've incorporated this fix into pr #902 : slightly different approach as we need the marker also for HSEARCH-1956

PR #901, opened 2015-09-01
Simple doc changes to reflect the new WildFly deployer

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-1963
 - https://hibernate.atlassian.net/browse/HSEARCH-1979
 - https://hibernate.atlassian.net/browse/HSEARCH-1980

See also https://docs.jboss.org/author/display/WFLY10/JPA+Reference+Guide#JPAReferenceGuide-UsingHibernateSearch


README.md
```diff
@@ -79,13 +79,9 @@ The project is split in several Maven modules:
 
 * _engine_: The engine of the project. Most of the beef is here.
 
-* _infinispan_: Backend storing indexes in [Infinispan](http://infinispan.org/).
-
 * _integrationtest_: Integration tests with various technologies like WildFly, Spring or Karaf.
 Also includes performance tests.
 
-* _legacy_: Old Maven GAV kept for backward compatibility.
```
gunnarmorling: To be sure: You removed it from the docs but still keep it around as an undocumented artefact until the next major?

gunnarmorling:
Merged. Some integration tests were failing for me, caused by HS being added as module *and* as lib contained in the deployment unit. I've adapted the affected tests to not have the HS module being added automatically by the deployer (so we have tests of both flavours).

Sanne:
your commit is clearly correct but I double checked my version and it doesn't fail here.. puzzling.
Maybe we should add explicit testing of the version being run? I guess it's possible for some tests to succeed even though it's using the version included within Wildfly.
thanks!

README.md
Sanne: I'd prefer to keep it around for longer, especially as still recently some documentation was advising those GAVs.. but since it's now using an explicit "relocation" tag I didn't feel like needing to scare first time contributors who are reading the README with a long list of modules, when these can be ignored.

gunnarmorling:
It failed for me with some class-loading issue, I think it basically somehow caught classes from both modules which then were not compatible with each other. So I fixed this to have only one version on the class path.

PR #903, opened 2015-09-02
HSEARCH-1973 Avoid double wrapping bridges for Null marker support wh…

Sanne:
…ich already support the functionality

### DIFF
--- a/engine/src/main/java/org/hibernate/search/bridge/builtin/impl/NullEncodingTwoWayFieldBridge.java
+++ b/engine/src/main/java/org/hibernate/search/bridge/builtin/impl/NullEncodingTwoWayFieldBridge.java
@@ -10,22 +10,33 @@ package org.hibernate.search.bridge.builtin.impl;
 import org.apache.lucene.document.Document;
 
 import org.apache.lucene.index.IndexableField;
+import org.hibernate.search.bridge.FieldBridge;
 import org.hibernate.search.bridge.LuceneOptions;
 import org.hibernate.search.bridge.TwoWayFieldBridge;
+import org.hibernate.search.bridge.impl.NullEncodingCapable;
 import org.hibernate.search.util.logging.impl.Log;
 import org.hibernate.search.util.logging.impl.LoggerFactory;
 
 /**
  * @author Hardy Ferentschik
  */
-public class NullEncodingTwoWayFieldBridge implements TwoWayFieldBridge {
+public class NullEncodingTwoWayFieldBridge implements TwoWayFieldBridge, NullEncodingCapable {
 
 	private static final Log LOG = LoggerFactory.make();
 
 	private final TwoWayFieldBridge fieldBridge;
 	private final String nullMarker;
 
-	public NullEncodingTwoWayFieldBridge(TwoWayFieldBridge fieldBridge, String nullMarker) {
+	public static FieldBridge wrapForNullEncodingIfNeeded(FieldBridge fieldBridge, String nullMarker) {
+		if ( nullMarker != null && fieldBridge instanceof TwoWayFieldBridge && ! ( fieldBridge instanceof NullEncodingCapable ) ) {
+			return new NullEncodingTwoWayFieldBridge( (TwoWayFieldBridge) fieldBridge, nullMarker );
+		}
+		else {
+			return fieldBridge;
+		}
+	}
+
+	private NullEncodingTwoWayFieldBridge(TwoWayFieldBridge fieldBridge, String nullMarker) {
 		this.fieldBridge = fieldBridge;
 		this.nullMarker = nullMarker;
 	}
--- a/engine/src/main/java/org/hibernate/search/bridge/impl/NullEncodingCapable.java
+++ b/engine/src/main/java/org/hibernate/search/bridge/impl/NullEncodingCapable.java
@@ -1,0 +1,15 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.bridge.impl;
+
+/**
+ * Marker interface to indicate that a StringBridge or FieldBridge implementation
+ * does not need wrapping to handle encoding of null values.
+ */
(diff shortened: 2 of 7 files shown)


engine/src/test/java/org/hibernate/search/test/dsl/NumericTypeWithNullEncodingTest.java
```diff
@@ -0,0 +1,121 @@
...
+		Assert.assertEquals( "title-two", entityInfo.getProjection()[0] );
+	}
+
+
+	@Test
+	public void verifyExplicitKeywordQuery() {
+		Query query = getQueryBuilder()
+					.keyword()
+					.onField( "age" )
+					.matching( 2 )
```
gunnarmorling: Would `matching( null )` be a supported invocation as well?

engine/src/test/java/org/hibernate/search/test/sorting/SortingTest.java
```diff
@@ -0,0 +1,121 @@
...
+import org.hibernate.search.testsupport.setup.TransactionContextForTest;
+import org.junit.Rule;
+import org.junit.Test;
+
+/**
+ * Test to verify we apply the right sorting strategy for non-trivial mapped entities
+ *
+ * @author Sanne Grinovero
+ */
+public class SortingTest {
```
gunnarmorling: Looks nice, but is it strictly related to the issue at hand?

engine/src/main/java/org/hibernate/search/bridge/impl/NullEncodingCapable.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.bridge.impl;
+
+/**
+ * Marker interface to indicate that a StringBridge or FieldBridge implementation
```
gunnarmorling: StringBridge? It seems only used for field bridges? You envision more general usage down the road?

engine/src/main/java/org/hibernate/search/bridge/impl/NullEncodingCapable.java
```diff
@@ -0,0 +1,15 @@
...
+ * does not need wrapping to handle encoding of null values.
+ */
+public interface NullEncodingCapable {
```
gunnarmorling: I'd name it `NullEncodingCapable(Field)Bridge` to make apparent it's a contract for (field) bridges.

gunnarmorling:
@Sanne Added some comments/questions. Please merge yourself once you think it's good.

engine/src/test/java/org/hibernate/search/test/dsl/NumericTypeWithNullEncodingTest.java
anistor: Nope. I'm just testing this in infinispan and get 
>org.hibernate.search.exception.SearchException: HSEARCH000237: Cannot create numeric range query for field 'age', since from and to values are null
	at org.hibernate.search.bridge.util.impl.NumericFieldUtils.createNumericRangeQuery(NumericFieldUtils.java:43)

anistor: This is why I handled numeric nulls as if they were strings here (https://github.com/hibernate/hibernate-search/pull/899/files#diff-6c790eb30d42e35f5a375e7c759dfff6R81). Lucene did not complain.

engine/src/main/java/org/hibernate/search/bridge/builtin/impl/NullEncodingTwoWayFieldBridge.java
```diff
@@ -10,22 +10,33 @@
...
 	private final TwoWayFieldBridge fieldBridge;
 	private final String nullMarker;
 
-	public NullEncodingTwoWayFieldBridge(TwoWayFieldBridge fieldBridge, String nullMarker) {
+	public static FieldBridge wrapForNullEncodingIfNeeded(FieldBridge fieldBridge, String nullMarker) {
+		if ( nullMarker != null && fieldBridge instanceof TwoWayFieldBridge && ! ( fieldBridge instanceof NullEncodingCapable ) ) {
+			return new NullEncodingTwoWayFieldBridge( (TwoWayFieldBridge) fieldBridge, nullMarker );
+		}
+		else {
+			return fieldBridge;
```
anistor: This null-capable bridge could use a different null marker.

Sanne: great point, my solution is wrong

engine/src/test/java/org/hibernate/search/test/dsl/NumericTypeWithNullEncodingTest.java
Sanne: Looks like I made a mess by trying to extract "the good parts" of a larger work. The null-query works in my full-picture branch but requires many more changes, but looks like this part alone is troublesome already.

engine/src/test/java/org/hibernate/search/test/sorting/SortingTest.java
Sanne: It's a backport from master to keep the test coverage regarding sorting aligned (I wanted to make sure I wouldn't break this)

Sanne:
Closing this to just got with @anistor's original fix and my tests. Attempting to already go in the direction of supporting the numeric tokens is just making it all complicated, I'll do that a a second step even if it means reverting some code here: after all the value here is in the added tests, and needing a fix for 5.4 which isn't affected by the tokens encoding issue.

engine/src/test/java/org/hibernate/search/test/dsl/NumericTypeWithNullEncodingTest.java
Sanne: > Lucene did not complain.

FWIW that's only an acceptable encoding if one doesn't sort on that field.. so that patch is good to apply on 5.4 but will need further fixing for (Lucene 5 && sorting).

Sanne:
Replaced with #905

PR #911, opened 2015-09-08
HSEARCH-1989 Let maven build support maven-color

emmanuelbernard:


emmanuelbernard:
Forced push to avoud the move to Checkstyle 2.14 as it breaks our build

Sanne:
ah, the priorities :)

PR #912, opened 2015-09-08
null tokens on Numeric fields shall have a numeric encoding

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-1956

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/impl/nullencoding/NullMarkerCodec.java
+++ b/engine/src/main/java/org/hibernate/search/engine/impl/nullencoding/NullMarkerCodec.java
@@ -1,0 +1,55 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.engine.impl.nullencoding;
+
+import org.apache.lucene.document.Document;
+import org.apache.lucene.index.IndexableField;
+import org.apache.lucene.search.Query;
+import org.hibernate.search.bridge.LuceneOptions;
+
+/**
+ * Contract to manage coding and decoding (querying) of null values.
+ * Null values can be stored as special markers or skipping the field
+ * altogether.
+ * When encoding a marker the index field type shall match the other fields,
+ * as index types often need to be consistent across the whole index.
+ * Also the strategy to find a match for such a null-encoded value might vary.
+ *
+ * @author Sanne Grinovero
+ */
+public interface NullMarkerCodec {
+
+	/**
+	 * This is mostly a requirement for integration with other old-style
+	 * contracts which expect a strongly String based strategy.
+	 * @return a string representation of the null-marker, or null if no marker is used.
+	 */
+	String nullRepresentedAsString();
+
+	/**
+	 * Store the null marker in the Document.
+	 * Some implementations might not do anything
+	 * @param fieldName the
+	 * @param document
+	 * @param luceneOptions
+	 */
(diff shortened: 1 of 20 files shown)


engine/src/main/java/org/hibernate/search/engine/impl/nullencoding/NullMarkerCodec.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.lucene.index.IndexableField;
+import org.apache.lucene.search.Query;
+import org.hibernate.search.bridge.LuceneOptions;
+
+/**
+ * Contract to manage coding and decoding (querying) of null values.
+ * Null values can be stored as special markers or skipping the field
+ * altogether.
+ * When encoding a marker the index field type shall match the other fields,
+ * as index types often need to be consistent across the whole index.
```
gunnarmorling: "index types"="field types"

engine/src/main/java/org/hibernate/search/engine/impl/nullencoding/NullMarkerCodec.java
```diff
@@ -0,0 +1,55 @@
...
+	String nullRepresentedAsString();
+
+	/**
+	 * Store the null marker in the Document.
+	 * Some implementations might not do anything
+	 * @param fieldName the
+	 * @param document
+	 * @param luceneOptions
+	 */
+	void encodeNullValue(String fieldName, Document document, LuceneOptions luceneOptions);
```
gunnarmorling: I wish that contract was a bit less dependent on Lucene. Couldn't there just be methods for converting the given marker value into the required format and vice versa. Otherwise we make our lives harder for alternative backends. I'd like to get away from having to create fake `Document`s for ES eventually :)

engine/src/test/java/org/hibernate/search/test/dsl/NumericTypeWithNullEncodingTest.java
```diff
@@ -36,7 +35,8 @@
 public class NumericTypeWithNullEncodingTest {
 
 	@Rule
-	public SearchFactoryHolder sfHolder = new SearchFactoryHolder( SomeEntity.class );
+	public SearchFactoryHolder sfHolder = new SearchFactoryHolder( SomeEntity.class )
+			.withProperty( org.hibernate.search.cfg.Environment.DEFAULT_NULL_TOKEN, "-7" );
```
gunnarmorling: Should there be a property to give a default per field type? Or e.g. `Environment.DEFAULT_NUMERIC_NULL_TOKEN`. Applying a shared default for string and numeric fields appears problematic, as it likely forces people into overriding it in many cases.

gunnarmorling:
@Sanne added some remarks. Looks functionally alright to me, but I don't like the fact that it's bound to Lucene APIs that heavily.

engine/src/main/java/org/hibernate/search/engine/impl/nullencoding/NullMarkerCodec.java
Sanne: good point, but it's an internal contract and we can changes this whenever we want. I did try other paths but the FieldBridge contract of today didn't leave me much freedom .. I'd need more invasive changes.

engine/src/test/java/org/hibernate/search/test/dsl/NumericTypeWithNullEncodingTest.java
Sanne: +1 opened https://hibernate.atlassian.net/browse/HSEARCH-1995

engine/src/main/java/org/hibernate/search/engine/impl/nullencoding/NullMarkerCodec.java
Sanne: fixed

Sanne:
applied remarks and merged

PR #914, opened 2015-09-09
HSEARCH-1992 Validating consistency of sort fields

gunnarmorling:


gunnarmorling:
This does the validation of requested sorts per concerned entity now. For queries targeting multiple indexes, only those indexes will be uninverted that actually have to, i.e. an index will be queried as is if it has the doc value fields for all affected the entity types it hosts while another index targeted in the same query may be uninverted.

For a shared index the sort field configuration must be consistent, i.e. either all entities do declare the required sort fields or none of them does (in which case the index will be uninverted), but there must not be a mixture of both. Sorting by doc value fields from an uninverted index gave me incorrect results, so I've added a check to prevent that.

Sanne:
awesome!

engine/src/main/java/org/hibernate/search/reader/impl/ManagedMultiReader.java
```diff
@@ -67,47 +70,42 @@ static ManagedMultiReader createInstance(IndexManager[] indexManagers, Iterable<
...
 		}
 
-		IndexReader[] effectiveReaders = getEffectiveReaders( indexManagers, subReaders, configuredSortableFields, sort );
+		IndexReader[] effectiveReaders = getEffectiveReaders( indexManagers, subReaders, configuredSorts, sort );
 		return new ManagedMultiReader( effectiveReaders, subReaders, readerProviders );
 	}
 
 	/**
-	 * Gets the readers to be effectively used. Will be the given readers, if:
+	 * Gets the readers to be effectively used. A given reader will be returned itself if:
```
DavideD: Can this be expressed with: "A reader will return itself..."?

orm/src/test/java/org/hibernate/search/test/query/SortTest.java
```diff
@@ -272,6 +278,67 @@ public void testResultOrderedByEmbeddedAuthorNameAscending() throws Exception {
...
+
+		tx.commit();
+	}
+
+	@Test
+	public void testQueryOnIndexSharedByEntityWithRequiredSortFieldAndEntityWithoutRaisesException() throws Exception {
+		Transaction tx = fullTextSession.beginTransaction();
+
+		Query query = queryParser.parse( "name:Bill" );
+		FullTextQuery hibQuery = fullTextSession.createFullTextQuery( query, Thatcher.class, BrickLayer.class );
```
DavideD: What's the difference between `Thatcher` and `Plumber`?
As far as I understand they both have a sortName field?

I don't understand why the previous test passes while this one fails.

DavideD:
I couldn't figure out why you are expecting one of the tests to fail, some additional comments might help.

@Sanne made me noticed that there should be a test checking that we are actually skipping the uninverting when it's not needed.

engine/src/main/java/org/hibernate/search/reader/impl/ManagedMultiReader.java
gunnarmorling: No, not really. Because it's not the reader which returns something but this method.

orm/src/test/java/org/hibernate/search/test/query/SortTest.java
gunnarmorling: I hoped the method name would say it all :) The difference is that `BrickLayer` (not `Plumber`!) defines the sort field whereas `Thatcher` does not. Both entities are mapped to the same index ("index sharing") and this test aims at discovering this inconsistent configuration. I first thought that uninverting the index in that case still would work, but apparently it doesn't: sorting across the uninverted and the doc value fields doesn't give the right order.

I'll add a more descriptive comment.

gunnarmorling:
>  there should be a test checking that we are actually skipping the uninverting

Any idea on how to do this (apart from fiddling with Byteman which I strongly dislike)?

Sanne:
> Any idea on how to do this (apart from fiddling with Byteman which I strongly dislike)?

You'll probably need to unit test (proper unit test, not end-to-end) the core function which makes the decision about inverting or not.

orm/src/test/java/org/hibernate/search/test/query/SortTest.java
DavideD: > I first thought that uninverting the index in that case still would work

That's what confused me as well. It seems counter intuitive as it works for different indexes.

gunnarmorling:
Pushed one more commits with more tests and docs.

gunnarmorling:
Added one more commit for [HSEARCH-2000](https://hibernate.atlassian.net/browse/HSEARCH-2000).

Sanne:
rebased and merged, I added one micro commit just to close an IndexReader in a test.

PR #916, opened 2015-09-10
HSEARCH-1966 Fix javadocs generation

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-1966

This pull request apply the following changes:

  * Move the plugin execution in the aggregator project: the build will fails on errors
  * Execute the plugin for each module: no need for sourcepath and it will run for every new module
  * Generate the javadoc when the profile for distribution is active (like before):
     ```mvn -Pdist```
  * Make it possible to disable the failure on error:
     ```mvn -Pdist -DfailOnJavaDocError=false ```


engine/src/main/java/org/hibernate/search/engine/metadata/impl/SortableFieldMetadata.java
```diff
@@ -21,6 +21,7 @@ private SortableFieldMetadata(String fieldName) {
 
 	/**
 	 * The name of the field whose field bridge will be used to determine the value for the doc value field.
+	 * @return the name of the field
```
gunnarmorling: So the build fails if that's not there? I left it out on purpose, because it's not useful and just adds clutter IMHO. If possible, I'd prefer to avoid this sort of comments.

pom.xml
```diff
@@ -1165,9 +1161,73 @@
     <profiles>
         <profile>
             <id>dist</id>
+
+            <properties>
+                <failOnError>true</failOnError>
```
gunnarmorling: I'd name it "failOnJavaDocError" to give some more context.

engine/src/main/java/org/hibernate/search/engine/metadata/impl/SortableFieldMetadata.java
gunnarmorling: Thinking about it, I believe that's only reported as WARN by JavaDoc. It's some other actual ERROR reports (such as unpaired tags, stale links) which make it fail. I don't think we need (or even should) fix everything reported at WARN level. That's not to suggest to undo it, just to describe the strategy I'd follow in general.

DavideD: I'm pretty sure it's an error in this case.
As far as I understand, one cannot add a description withouth also describing the  parameters in the signature.

I will check again, just in case.

pom.xml
DavideD: ok

engine/src/main/java/org/hibernate/search/engine/metadata/impl/SortableFieldMetadata.java
DavideD: Right, I made confusion again.
The problem is that, if there is a different error in the build, this appear in the log in the following way:
```
[ERROR] Exit code: 1 - /home/ddalto/workspace/projects/hibernate-search/engine/src/main/java/org/hibernate/search/engine/metadata/impl/SortableFieldMetadata.java:25: warning: no @return
```

If other people reading thi s don't spot it, the warning symbol is at the end of the sentence.

I dislike these kind of comments as well but I don't want to be in the situation where I have a huge list of warning and then I need to look for the real error.
I also tend to forget to pay close attention to the bug 
I would prefer if we can agree on a way to remove the warning as well for these cases.

Maybe the following is better?
```
	/**
	 * @return the name of the field whose field bridge will be used to determine the value for the doc value field.
	 */
	public String getFieldName() {
```

DavideD:
Pushed forced
* Changed the name of the variable for failure on javadoc generation
* Changed javadoc comments for `SortableField`, `SortableFields` and `SortableFieldMetadata`

engine/src/main/java/org/hibernate/search/engine/metadata/impl/SortableFieldMetadata.java
gunnarmorling: Agreed, the tool output is horrible, although it seems to list the actual errors at the end in recent JDK versions (at least in some cases, sometimes it's missing :/). Still I think we should not sacrifice readability of the generated docs for the sake of making the tool happy.

> Maybe the following is better?

No, IMO not. The method description (or its first sentence) is shown in the "method summary" block, whereas the return value description is only visible in the details block.

gunnarmorling:
When running `mvn clean install -DskipTests=true -Pdist`I am getting very many messages in the form of "Forking Hibernate Search ORM" (i.e. several times for one module). Is this expected?

DavideD:
> When running mvn clean install -DskipTests=true -PdistI am getting very many messages in the form of "Forking Hibernate Search ORM" (i.e. several times for one module). Is this expected?

I guess it's due to the fact that it is using the aggregate goal instead of the javadoc one.

Sanne:
I could not figure out why, but it seems that if I'm building it with
 > mvn clean install -Pdist,docs

the distribution tarballs contain the javadoc, but not the reference documentation. Am I using the wrong command?

DavideD:
s/docs/docbook/

DavideD:
mvn clean install -Pdist,**docbook**

Sanne:
right :)

Sanne:
merged!

PR #917, opened 2015-09-10
HSEARCH-1933 Warnings during build

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-1933

### DIFF
--- a/backends/jms/pom.xml
+++ b/backends/jms/pom.xml
@@ -41,6 +41,12 @@
             <type>test-jar</type>
             <scope>test</scope>
         </dependency>
+
+        <!-- Avoid warnings during build -->
+        <dependency>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>jboss-logging-annotations</artifactId>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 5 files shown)


backends/jms/pom.xml
```diff
@@ -41,6 +41,12 @@
             <type>test-jar</type>
             <scope>test</scope>
         </dependency>
+
+        <!-- Avoid warnings during build -->
+        <dependency>
```
Sanne: but we don't want to have this as a transitive dependency?

DavideD: It's provided, as it is defined as provided in the dependency management of the aggregator

DavideD: I can add the scope to make it explicit

Sanne: ok but is "provided" correct? That maven definition always confuses me, but we don't really expect it to be provided at runtime.

DavideD: To be precise, it's provided and optional: https://github.com/hibernate/hibernate-search/blob/master/pom.xml#L549
I don't think there is a better scope in maven that match what we need for our use case.

Sanne: ok

Sanne:
rebased and merged

PR #918, opened 2015-09-10
HSEARCH-1997 Avoid checking for Java8 time classes multiple times

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-1997


DavideD:
Would this work?

Sanne:
yes, nice. Rebased and merged

PR #919, opened 2015-09-10
HSEARCH-1955 ScheduledCommitPolicy helper threads not handling except…

Sanne:
…ions

 - https://hibernate.atlassian.net/browse/HSEARCH-1955

### DIFF
--- a/engine/src/main/java/org/hibernate/search/backend/impl/lucene/ExclusiveIndexWorkspaceImpl.java
+++ b/engine/src/main/java/org/hibernate/search/backend/impl/lucene/ExclusiveIndexWorkspaceImpl.java
@@ -9,6 +9,7 @@ package org.hibernate.search.backend.impl.lucene;
 import org.hibernate.search.backend.BackendFactory;
 import org.hibernate.search.backend.LuceneWork;
 import org.hibernate.search.backend.impl.CommitPolicy;
+import org.hibernate.search.exception.ErrorHandler;
 import org.hibernate.search.indexes.impl.PropertiesParseHelper;
 import org.hibernate.search.indexes.spi.DirectoryBasedIndexManager;
 import org.hibernate.search.spi.WorkerBuildContext;
@@ -27,7 +28,8 @@ public class ExclusiveIndexWorkspaceImpl extends AbstractWorkspaceImpl {
 		boolean async = ! BackendFactory.isConfiguredAsSync( cfg );
 		if ( async ) {
 			int commitInterval = PropertiesParseHelper.extractFlushInterval( indexManager.getIndexName(), cfg );
-			commitPolicy = new ScheduledCommitPolicy( writerHolder, indexManager.getIndexName(), commitInterval );
+			ErrorHandler errorHandler = context.getErrorHandler();
+			commitPolicy = new ScheduledCommitPolicy( writerHolder, indexManager.getIndexName(), commitInterval, errorHandler );
 		}
 		else {
 			commitPolicy = new PerChangeSetCommitPolicy( writerHolder );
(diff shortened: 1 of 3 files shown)


engine/src/test/java/org/hibernate/search/test/backend/lucene/ScheduledCommitPolicyTest.java
```diff
@@ -80,6 +80,26 @@ public boolean evaluate() {
...
+			action = "throw new NullPointerException(\"Fake internal error\")",
+			name = "timerDisruptingError")
+	public void testErrorHandlingOnBackgroundThread() throws Exception {
+		writeData( sfAsyncExclusiveIndex, 2 );
+		final CountingErrorHandler errorHandler = (CountingErrorHandler) sfAsyncExclusiveIndex.getSearchFactory().getErrorHandler();
+		assertConditionMet( new Condition() {
+			@Override
+			public boolean evaluate() {
+				// It's going to commit once each millisecond, and produce a failure each time.
+				// So "4" is just a random number higher than 0, but high enough to
```
DavideD: Do you mean "2"?

Not critical, but maybe you can use a method or a constant do describe this better

DavideD: No need to change it.
I'll replace it with 2 before merging it

engine/src/main/java/org/hibernate/search/backend/impl/lucene/ExclusiveIndexWorkspaceImpl.java
```diff
@@ -27,7 +28,8 @@ public ExclusiveIndexWorkspaceImpl(DirectoryBasedIndexManager indexManager, Work
 		boolean async = ! BackendFactory.isConfiguredAsSync( cfg );
 		if ( async ) {
 			int commitInterval = PropertiesParseHelper.extractFlushInterval( indexManager.getIndexName(), cfg );
-			commitPolicy = new ScheduledCommitPolicy( writerHolder, indexManager.getIndexName(), commitInterval );
+			ErrorHandler errorHandler = context.getErrorHandler();
+			commitPolicy = new ScheduledCommitPolicy( writerHolder, indexManager.getIndexName(), commitInterval, errorHandler );
 		}
 		else {
 			commitPolicy = new PerChangeSetCommitPolicy( writerHolder );
```
DavideD: I suppose it's a different issue but shouldn't we handle errors for this case as well?

engine/src/test/java/org/hibernate/search/test/backend/lucene/ScheduledCommitPolicyTest.java
Sanne: that's a rebase/conflict mistake. I actually meant to use "4" in the code, which is safer than 2 as t here are 2 updates being done, I want to verify the scheduler does more than that.

engine/src/main/java/org/hibernate/search/backend/impl/lucene/ExclusiveIndexWorkspaceImpl.java
Sanne: It's not needed for that one as it's not asynchronous. The problem we're fixing is that an exception would kill the timer thread, and never commit again.

engine/src/test/java/org/hibernate/search/test/backend/lucene/ScheduledCommitPolicyTest.java
DavideD: I'll replace the threshold to 4 then

Sanne:
pushed again to fix the '4' :)

DavideD:
Merged https://github.com/hibernate/hibernate-search/pull/919

Sanne:
I'll resolve the JIRA then

PR #920, opened 2015-09-11
HSEARCH-1987 Making sure @NumericField works correctly if a property …

gunnarmorling:
…has several fields

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/metadata/impl/NumericFieldsConfiguration.java
+++ b/engine/src/main/java/org/hibernate/search/engine/metadata/impl/NumericFieldsConfiguration.java
@@ -1,0 +1,100 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.engine.metadata.impl;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import org.hibernate.annotations.common.reflection.XProperty;
+import org.hibernate.search.annotations.NumericField;
+import org.hibernate.search.annotations.NumericFields;
+import org.hibernate.search.util.logging.impl.Log;
+import org.hibernate.search.util.logging.impl.LoggerFactory;
+
+/**
+ * Represents the configuration of numeric fields of a specific entity property.
+ *
+ * @author Gunnar Morling
+ */
+class NumericFieldsConfiguration {
+
+	private static final Log LOG = LoggerFactory.make();
+
+	private final Class<?> indexedType;
+	private final XProperty member;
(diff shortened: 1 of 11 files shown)


engine/src/main/java/org/hibernate/search/engine/metadata/impl/NumericFieldsConfiguration.java
```diff
@@ -0,0 +1,100 @@
...
+			if ( fieldsOfProperty.contains( fieldMarkedAsNumeric ) ||
+					( fieldMarkedAsNumeric.isEmpty() && !fieldsOfProperty.isEmpty() ) ) {
+				continue;
+			}
+
+			throw LOG.numericFieldAnnotationWithoutMatchingField( indexedType, member.getName() );
+		}
+	}
+
+	private String unqualify(String fieldName) {
```
Sanne: Is this meant to handle embeddable fields? I'm not sure if you can rely on a strong guarantee of them being separated via a '.'

engine/src/main/java/org/hibernate/search/engine/metadata/impl/NumericFieldsConfiguration.java
```diff
@@ -0,0 +1,100 @@
...
+		int separatorIdx = fieldName.lastIndexOf( '.' );
+		if ( separatorIdx != -1 ) {
+			fieldName = fieldName.substring( separatorIdx + 1 );
+		}
+
+		return fieldName;
+	}
+}
```
Sanne: no need to fix this, but I noticed it other times in your PRs. Something wrong in your IDE settings?

engine/src/main/java/org/hibernate/search/engine/metadata/impl/NumericFieldsConfiguration.java
```diff
@@ -0,0 +1,100 @@
...
+	 */
+	NumericField getNumericFieldAnnotation(String fieldName) {
+		fieldName = unqualify( fieldName );
+
+		fieldsOfProperty.add( fieldName );
+
+		NumericField numericFieldAnnotation = fieldsMarkedAsNumeric.get( fieldName );
+
+		if ( numericFieldAnnotation == null ) {
+			numericFieldAnnotation = fieldsMarkedAsNumeric.get( "" );
```
Sanne: This one was tricky :) I've been checking some corner cases but indeed I couldn't fool it.
:+1:

engine/src/main/java/org/hibernate/search/engine/metadata/impl/NumericFieldsConfiguration.java
Sanne: The problem is that org.hibernate.search.annotations.IndexedEmbedded.prefix() is customizable :-(

Sanne: lol, I tried to look back in history to find out why that was done, and the answer was interesting: it's there since the first ever commit!

Sanne:
closing as replaced by #921

PR #921, opened 2015-09-11
Hsearch 1987

Sanne:
building on #920 : added one more fix and saved a bit of memory


Sanne:
Removed the last commit. Still need to process the other remark..

gunnarmorling:
@Sanne You still plan to follow-up on this? Or should I take it over again? Is there anything else missing for the release?

Sanne:
Hi Gunnar,
I could not follow up on that last issue with ids, I think it's ok to open
that as a separate issue to be looked at later on. Could you open the jira
and describe it please (I don't have my IDE now and don't remember the
details).

Shall we merge this then? Your fix looks good for the issue at hand.

gunnarmorling:
Rebased and applied. I've added one more commit to fix the remaining issue about numeric embedded id fields. All should be good now. Cheers :)

PR #926, opened 2015-09-14
HSEARCH-2004 Queries executed outside a transaction in tests

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-2004

CI cannot complete the build when running the mariadb profile because of some tests not executing queries inside a transaction.


DavideD:
Jenkins, retest this please

PR #927, opened 2015-09-15
HSEARCH-2003, HSEARCH-2005

gunnarmorling:
Consider the field name when configuring numeric fields through the DSL. Also deprecates the method `ContainedInMapping#numericField()` as discussed with @emmanuelbernard. Plus one addition to the sortable field configuration through the DSL: It's now possible for the implicit id field, too.


orm/src/test/java/org/hibernate/search/test/configuration/Item.java
```diff
@@ -8,17 +8,16 @@
...
 
 @Entity
 public class Item {
 
-	@Id @GeneratedValue
+	@Id
 	private Integer id;
 	private String description;
-	private Double price;
+	private Short price;
```
DavideD: How come the change from double to short?

gunnarmorling: It's to facilitate testing: for short (and byte) I can control String vs. numeric mapping via `@NumericField` (Double will be a numeric field by default, no way to disable it through `@NumericField`). Then I can assert on the right field type in the Lucene document.

engine/src/main/java/org/hibernate/search/cfg/NumericFieldMapping.java
```diff
@@ -14,10 +14,11 @@
  */
 public class NumericFieldMapping extends FieldMapping {
 
-	private final Map<String, Object> numericField = new HashMap<String, Object>();
+	private final Map<String, Object> numericField = new HashMap<String, Object>( 2 );
 
-	public NumericFieldMapping(PropertyDescriptor property, EntityDescriptor entity, SearchMapping mapping) {
+	public NumericFieldMapping(String fieldName, PropertyDescriptor property, EntityDescriptor entity, SearchMapping mapping) {
```
gunnarmorling: Technically it's an API change, but I'd say it's ok as it's not supposed to be called by clients really?

It'd be much better if there was a proper public API / impl split for the DSL as it's the case in the HV DSL.

DavideD:
Merged https://github.com/hibernate/hibernate-search/pull/927

DavideD:
Rebased to the latest changes

PR #935, opened 2015-11-03
HSEARCH-2029 Hibernate Search does not reuse JavaReflectionManager from Hibernate.

golovnin:
Changes the implementation of SearchConfigurationFromHibernateCore#getReflectionManager() to reuse in Hibernate Search an instance of JavaReflectionManager from Hibernate.

### DIFF
--- a/orm/src/main/java/org/hibernate/search/cfg/impl/SearchConfigurationFromHibernateCore.java
+++ b/orm/src/main/java/org/hibernate/search/cfg/impl/SearchConfigurationFromHibernateCore.java
@@ -17,6 +17,7 @@ import java.util.Set;
 import org.hibernate.annotations.common.reflection.ReflectionManager;
 import org.hibernate.annotations.common.reflection.java.JavaReflectionManager;
 import org.hibernate.boot.Metadata;
+import org.hibernate.boot.spi.MetadataImplementor;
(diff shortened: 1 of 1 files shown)


gunnarmorling:
Hey @golovnin, thanks for this PR.

I stumbled upon your comment on the issue regarding the module export. I don't think it's the correct way of doing this.

The failing tests adds the HSEARCH libs to the deployment unit (i.e. they are not obtained as modules from the container). In `PackagerHelper` some transitive dependencies are excluded from the deployment unit (e.g. ORM), but accidentally the HCANN JAR gets added to the WAR. That way two versions of HCANN are present: the one from the WAR and the container-provided one, as ORM depends on it. This gives the linkage error.

To me the correct way appears to be to exclude HCANN from the deployment unit and instead have the test depend on the container's HCANN module. I got this sorted out locally and will amend your PR accordingly. Good thing is that no change on the WF side is needed.

golovnin:
Hi Gunnar,
thank you for looking at it and finding the right solution for the LinkageError in the failing tests. I must admit I have not looked into the tests. I assumed they are correct per se and thought it must be an error in WildFly. :-)

gunnarmorling:
No problem, the export just made me curious :) I'm going to close this one, as it's superseded #938. Thanks!

PR #936, opened 2015-11-03
HSEARCH-2030 DSL range/below on a string field creates a TermRangeQue…

anistor:
…ry with lower bound equal to the null token

jira: https://hibernate.atlassian.net/browse/HSEARCH-2030

### DIFF
--- a/engine/src/main/java/org/hibernate/search/query/dsl/impl/ConnectedMultiFieldsRangeQueryBuilder.java
+++ b/engine/src/main/java/org/hibernate/search/query/dsl/impl/ConnectedMultiFieldsRangeQueryBuilder.java
@@ -113,15 +113,17 @@ public class ConnectedMultiFieldsRangeQueryBuilder implements RangeTerminationEx
 
 	private static Query createKeywordRangeQuery(String fieldName, RangeQueryContext rangeContext, QueryBuildingContext queryContext, ConversionContext conversionContext, FieldContext fieldContext) {
 		final Analyzer queryAnalyzer = queryContext.getQueryAnalyzer();
-		final Object fromObject = rangeContext.getFrom();
-		final Object toObject = rangeContext.getTo();
 		final DocumentBuilderIndexedEntity documentBuilder = Helper.getDocumentBuilder( queryContext );
-		final String fromString = fieldContext.objectToString( documentBuilder, fromObject, conversionContext );
+		final String fromString = rangeContext.hasFrom() ?
+				fieldContext.objectToString( documentBuilder, rangeContext.getFrom(), conversionContext ) :
+				null;
 		final String lowerTerm = fromString == null ?
 				null :
 				Helper.getAnalyzedTerm( fieldName, fromString, "from", queryAnalyzer, fieldContext );
 
-		final String toString = fieldContext.objectToString( documentBuilder, toObject, conversionContext );
+		final String toString = rangeContext.hasTo() ?
+				fieldContext.objectToString( documentBuilder, rangeContext.getTo(), conversionContext ) :
+				null;
 		final String upperTerm = toString == null ?
 				null :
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/hibernate/search/query/dsl/impl/ConnectedMultiFieldsRangeQueryBuilder.java
```diff
@@ -116,12 +116,16 @@ private static Query createKeywordRangeQuery(String fieldName, RangeQueryContext
...
-		final String fromString = fieldContext.objectToString( documentBuilder, fromObject, conversionContext );
+		final String fromString = fromObject == null ?
+				null :
+				fieldContext.objectToString( documentBuilder, fromObject, conversionContext );
 		final String lowerTerm = fromString == null ?
 				null :
 				Helper.getAnalyzedTerm( fieldName, fromString, "from", queryAnalyzer, fieldContext );
 
-		final String toString = fieldContext.objectToString( documentBuilder, toObject, conversionContext );
+		final String toString = toObject == null ?
```
gunnarmorling: I am wondering whether we should consider the case where explicitly `above( null )` has been invoked. E.g. if the null token is "-1", that may make sense to exclude some negative values existing in the database for some reason. Maybe it's contrived, but I'd like to hear what others think.

anistor: My feeling is ```below(null)``` and ```above(null)``` should be forbidden. Because technically only one of the bounds of ```TermRangeQuery``` can be null, not both. We are making an error by using the null token to fill in. It should not play any role in range queries because it is only a marker so it only makes sense for == or "is null' type of comparisons. 

There is also a deeper reason to disallow it. Because such a query would either match all or nothing, so it would be pointless. I support my theory by looking at how others handle this, specifically JPA and SQL. Here is a piece of jpa spec regarding the results of 'order by':

>SQL rules for the ordering of null values apply: that is, all null values must appear before all non-null values in the ordering or all null values must appear after all non-null values in the ordering, but it is not specified which.

anistor: So I would add another jira for banning below(null) and above(null) if you agree.

anistor:
I've softened the approach a bit. Please take a look at second commit. I'll squash them afterwards.

anistor:
Updated and squashed.

anistor:
Updated, added ```@TestForIssue``` annotation.

anistor:
I renamed the test to be more descriptive and added an assert for the below(null) case also.

gunnarmorling:
Merged; Thanks, @anistor!

anistor:
Thanks @gunnarmorling! I would backport this to Hibernate Search 5.5.x also. WDYT?

Sanne:
@anistor I backported too. Thanks!

gunnarmorling:
Sorry, that really makes no sense to me. Feel free to do it of course, I
for myself will only apply changes to master until there is an actual
deviation between the two branches.

Sanne:
@gunnarmorling it makes perfect sense: @anistor needs this fix soon, so unless you want to force him to use a 5.6.0-SNAPSHOT or Alpha1 - as we definitely have no reason to publish a 5.6.0.Final this implies we'll soon be releasing a 5.5.1.Final.
And since he's doing maintenance bugfixing on downstream projects, it's not desirable to force them to upgrade minors or use an unstable version.

gunnarmorling:
Sorry, but you seem to misunderstand on purpose; The obvious solution of course would be to release (or branch and diverge) a 5.5.1 off master *once* there is the actual need for it.

Anyways, I have had my say; I don't care about any unneeded merges you may want to do, I just won't take part in it :)

PR #937, opened 2015-11-05
HSEARCH-2027 Making sure dependency versions used in Karaf features f…

gunnarmorling:
…ile match the ones from ORM

Follow-up to the recent ORM version update which did not update the Jandex version, too.


Sanne:
awesome, thanks!
I've been rebuilding it many times now and was getting confused as I was sure it passed before I merged the upgrade.. something wrong with my build process.

PR #941, opened 2015-11-12
HSEARCH-1964 Clarify documentation around should, must and must not

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-1964


gunnarmorling:
Merged to master and 5.5.

PR #942, opened 2015-11-12
HSEARCH-2034

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2034

Allow users of the Query DSL to disable scoring, to improve performance when the intent is just to filter by some criteria.

### DIFF
--- a/orm/src/test/java/org/hibernate/search/test/query/dsl/DSLTest.java
+++ b/orm/src/test/java/org/hibernate/search/test/query/dsl/DSLTest.java
@@ -417,6 +417,33 @@ public class DSLTest extends SearchTestBase {
 		transaction.commit();
 	}
 
+	@Test
+	@SuppressWarnings("unchecked")
+	public void testBooleanWithoutNegationQueries() throws Exception {
+		Transaction transaction = fullTextSession.beginTransaction();
+		final QueryBuilder monthQb = fullTextSession.getSearchFactory()
+				.buildQueryBuilder().forEntity( Month.class ).get();
+
+		//must + disable scoring
+		Query query = monthQb
+				.bool()
+				.must( monthQb.keyword().onField( "mythology" ).matching( "colder" ).createQuery() )
+					.not() //expectation: exclude January
+				.must( monthQb.keyword().onField( "mythology" ).matching( "snowboarding" ).createQuery() )
+					.not() //expectation: exclude February
+				.createQuery();
+
+		List<Month> results = fullTextSession.createFullTextQuery( query, Month.class ).list();
+		assertEquals( 1, results.size() );
+		assertEquals( "March", results.get( 0 ).getName() );
+		assertTrue( query instanceof BooleanQuery );
+		BooleanQuery bq = (BooleanQuery) query;
+		BooleanClause firstBooleanClause = bq.clauses().get( 0 );
+		assertFalse( firstBooleanClause.isScoring() );
+
+		transaction.commit();
+	}
+
 	@Test(expected = SearchException.class)
 	public void testIllegalBooleanJunction() {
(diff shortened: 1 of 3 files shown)


orm/src/test/java/org/hibernate/search/test/query/dsl/DSLTest.java
```diff
@@ -351,6 +354,22 @@ public void testBoolean() throws Exception {
 		assertEquals( 1, results.size() );
 		assertEquals( "January", results.get( 0 ).getName() );
 
+		//must + disable scoring
+		query = monthQb
+				.bool()
+				.must( monthQb.keyword().onField( "mythology" ).matching( "colder" ).createQuery() )
+				.disableScoring()
+				.should( monthQb.keyword().onField( "mythology" ).matching( "colder" ).createQuery() )
```
gunnarmorling: Why is it you have this `should()` clause here? Is it related to that optimization for single-clause boolean queries? If so, we either should document the requirement for such pseudo-phrase to enable the FILTER mode or (my preference) disable the optimization for single FILTER clauses.

orm/src/test/java/org/hibernate/search/test/query/dsl/DSLTest.java
```diff
@@ -351,6 +354,22 @@ public void testBoolean() throws Exception {
 		assertEquals( 1, results.size() );
 		assertEquals( "January", results.get( 0 ).getName() );
 
+		//must + disable scoring
```
gunnarmorling: Just a style-thing: I'd rather create a dedicated test method for this (also for the other logical groups of assertions within that method btw.). That way e.g. a test failure on CI makes more apparent which usage is causing the error. The inline comments give a good indication what should be separate methods.

Sanne: +1

orm/src/test/java/org/hibernate/search/test/query/dsl/DSLTest.java
Sanne: iGood point, that's a bug.
The 'should()' clause is indeed needed to avoid it being optimised as a non-boolean query.

But I didn't think that it would prevent this to be run as a non-scoring FILTER; further the optimisation logic only checks for "MUST" and not for FILTER clauses, so I need to fix that too.

Sanne:
Re-pushed, completely different implementation which fixes a bug too.

 - https://hibernate.atlassian.net/browse/HSEARCH-2034
 - https://hibernate.atlassian.net/browse/HSEARCH-2037

orm/src/test/java/org/hibernate/search/test/query/dsl/DSLTest.java
```diff
@@ -386,6 +389,57 @@ public void testBoolean() throws Exception {
...
+		BooleanQuery bq = (BooleanQuery) query;
+		BooleanClause firstBooleanClause = bq.clauses().get( 0 );
+		assertFalse( firstBooleanClause.isScoring() );
+
+		transaction.commit();
+	}
+
+	@Test
+	@SuppressWarnings("unchecked")
+	public void testBooleanWithoutNegationQueries() throws Exception {
```
gunnarmorling: Should this rather be `testBooleanWithOnlyNegationQueries()`? I will change it.

gunnarmorling:
Merged to master and 5.5. Added one more commit for fixing the test method name and adding issue references.

orm/src/test/java/org/hibernate/search/test/query/dsl/DSLTest.java
Sanne: +1 thanks for the fix

PR #943, opened 2015-11-12
HSEARCH-2035 Avoiding dependency to Lucene's Query type in TimeoutMan…

gunnarmorling:
…agerImpl

### DIFF
--- a/engine/src/main/java/org/hibernate/search/query/engine/impl/TimeoutManagerImpl.java
+++ b/engine/src/main/java/org/hibernate/search/query/engine/impl/TimeoutManagerImpl.java
@@ -8,11 +8,9 @@ package org.hibernate.search.query.engine.impl;
 
 import java.util.concurrent.TimeUnit;
 
-import org.apache.lucene.search.Query;
 import org.apache.lucene.util.Counter;
-
-import org.hibernate.search.exception.SearchException;
 import org.hibernate.search.engine.spi.TimingSource;
+import org.hibernate.search.exception.SearchException;
 import org.hibernate.search.query.engine.spi.TimeoutExceptionFactory;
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/hibernate/search/query/engine/impl/HSQueryImpl.java
```diff
@@ -238,7 +238,7 @@ private TimeoutManagerImpl getTimeoutManagerImpl() {
 			if ( luceneQuery == null ) {
 				throw new AssertionFailure( "Requesting TimeoutManager before setting luceneQuery()" );
 			}
-			timeoutManager = new TimeoutManagerImpl( luceneQuery, timeoutExceptionFactory, this.extendedIntegrator.getTimingSource() );
+			timeoutManager = new TimeoutManagerImpl( String.valueOf( luceneQuery ), timeoutExceptionFactory, this.extendedIntegrator.getTimingSource() );
```
Sanne: I think we should avoid invoking the toString() this early, as the code in many Query implementations is a bit stressing.

Could we make it lazy? Worst case just accepting the 'Object' type would do, and we invoke toString() on that as necessary.. but maybe you have a more elegant idea which doesn't take you more than 5 minutes?

Sanne: BTW the reasoning is that the string is only needed when and ever if the timeout exception is thrown.

gunnarmorling: `Object` it is then. Rebased and force-pushed an update.

Sanne:
merged

PR #944, opened 2015-11-12
HSEARCH-2036 Extracting base class for HSQuery implementations

gunnarmorling:
(so ES can re-use some bits of it)

I expect to be able to move some more logic there (the determination of affected classes / index managers), but I'll come to that when honing the ES-specific implementation.


Sanne:
great! merging..

PR #946, opened 2015-11-13
HSEARCH-2039

gunnarmorling:
Hey @Sanne, this uses temporary directories as index dirs for the tests in "orm".

Other modules I still let use their "target" dir as that's generally safer to clean-up (the temp dirs used in "orm" are deleted upon normal JVM shut-down, but e.g. not when killing a test run). I've pushed the determination of target dirs to the individual modules, as a common re-used helper just seemed to easy to be abused in cases where it shouldn't (tests in "orm").

I'm also adding the caller name to index dirs now so to make it easier to identify tests potentially causing file leaks.

### DIFF
--- a/engine/src/test/java/org/hibernate/search/test/configuration/sharding/ShardingConfigurationTest.java
+++ b/engine/src/test/java/org/hibernate/search/test/configuration/sharding/ShardingConfigurationTest.java
@@ -8,6 +8,10 @@ package org.hibernate.search.test.configuration.sharding;
 
 import java.io.File;
 import java.io.Serializable;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -15,13 +19,12 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 
-import org.junit.Assert;
 import org.apache.lucene.document.Document;
-import org.hibernate.search.exception.SearchException;
 import org.hibernate.search.annotations.DocumentId;
 import org.hibernate.search.annotations.Indexed;
 import org.hibernate.search.engine.impl.MutableSearchFactory;
 import org.hibernate.search.engine.spi.EntityIndexBinding;
+import org.hibernate.search.exception.SearchException;
 import org.hibernate.search.filter.FullTextFilterImplementor;
 import org.hibernate.search.indexes.spi.DirectoryBasedIndexManager;
 import org.hibernate.search.indexes.spi.IndexManager;
@@ -32,12 +35,13 @@ import org.hibernate.search.store.ShardIdentifierProvider;
 import org.hibernate.search.store.impl.FSDirectoryProvider;
 import org.hibernate.search.store.impl.IdHashShardingStrategy;
 import org.hibernate.search.store.impl.NotShardedStrategy;
-import org.hibernate.search.testsupport.TestConstants;
 import org.hibernate.search.testsupport.BytemanHelper;
-import org.hibernate.search.testsupport.setup.SearchConfigurationForTest;
+import org.hibernate.search.testsupport.TestConstants;
 import org.hibernate.search.testsupport.TestForIssue;
+import org.hibernate.search.testsupport.setup.SearchConfigurationForTest;
 import org.jboss.byteman.contrib.bmunit.BMRule;
 import org.jboss.byteman.contrib.bmunit.BMUnitRunner;
+import org.junit.Assert;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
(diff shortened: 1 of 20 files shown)


engine/src/test/java/org/hibernate/search/test/configuration/sharding/ShardingConfigurationTest.java
```diff
@@ -262,6 +266,21 @@ private MutableSearchFactory getSearchFactory(Map<String, String> shardingProper
 		).buildSearchIntegrator();
 	}
 
+	private static Path getTargetDir() {
+		URI classesDirUri;
+		try {
+			classesDirUri = ShardingConfigurationTest.class.getProtectionDomain()
+					.getCodeSource()
+					.getLocation()
```
Sanne: I'm not familiar with this API. This would - like the existing test helper - attempt to write into the JAR's URL when referring to the class of a module which isn't the one currently being compiled by Maven right?

If so, what's the benefit in changing this from using the current helper?

engine/src/test/java/org/hibernate/search/testsupport/TestConstants.java
```diff
@@ -44,43 +43,40 @@ public static Version getTargetLuceneVersion() {
...
-	 * @param testClass the test class for which the index directory is requested.
+	 * @param requester The test class requesting the index directory, its name will be part of the returned directory
 	 * @return Return the root directory to store test indexes
 	 */
-	public static String getIndexDirectory(Class<?> testClass) {
-		Path targetDir = getTargetDir( testClass );
-		Path indexDirPath = targetDir.resolve( "indextemp" );
+	public static String getIndexDirectory(Path parent, Class<?> requester) {
+		Path indexDirPath = parent.resolve( "indextemp-" + requester.getSimpleName() );
+		indexDirPath.toFile().deleteOnExit();
```
Sanne: 'deleteOnExit' is just an additional safety right?

I don't think we could rely on it, as I presume it would fail if it's not empty?

engine/src/test/java/org/hibernate/search/testsupport/TestConstants.java
```diff
@@ -89,4 +85,11 @@ public static String getIndexDirectory(Class<?> testClass) {
 	public static boolean arePerformanceTestsEnabled() {
 		return Boolean.getBoolean( "org.hibernate.search.enable_performance_tests" );
 	}
+
+	private static class CallerProvider extends SecurityManager {
+
+		public Class<?> getCallerClass() {
+			return getClassContext()[2];
```
Sanne: The static initialization will happen only once per JVM.
Does this imply that our testsuite now requires surefire to run with "fork" ?

engine/src/test/java/org/hibernate/search/test/configuration/sharding/ShardingConfigurationTest.java
gunnarmorling: Yes, it must not be invoked for classes from the classpath. Hence I moved that into modules (see my introductory comment at this PR) and changed it to not accept a class object which suggests more flexibility than actually present.

Regarding the implementation, I'd say it's much simpler. No functional benefit apart from that, just my "canonical way" of doing this :)

Sanne:
I identified some minor concerns, but since it's just test helpers I'm ok to merge this as-is if you prefer.. I'm sure it's not the definitive framework anyway ;)

Why still rely on classloader detection? Seems such a fragile approach.

Sanne:
Marked as "just push it": I could push now myself but giving you the chance to reply on my comments if you think any question is actually warranting a change.

Tests pass for me.

engine/src/test/java/org/hibernate/search/test/configuration/sharding/ShardingConfigurationTest.java
Sanne: Ok, yes no doubts it looks better.

engine/src/test/java/org/hibernate/search/testsupport/TestConstants.java
gunnarmorling: If everything goes well, the directory is empty. I ran the full suite and all temp dirs were gone afterwards.

engine/src/test/java/org/hibernate/search/testsupport/TestConstants.java
gunnarmorling: How does the fork mode relate to this? This is just to obtain the calling class of the `getIndexDirectory()` method. How the class is loaded should not make a difference.

gunnarmorling:
> Why still rely on classloader detection? Seems such a fragile approach.

Not sure I am following? To which part specifically do you refer?

engine/src/test/java/org/hibernate/search/testsupport/TestConstants.java
Sanne: Let's assume Surefire's Fork is disabled: the same JVM would run all modules and initialize the static fields of this class only once.
The value of TestConstants.callerProvider would be initialized once, at first use, and that might be pointing to the target directory of a different module than the one being tested.

gunnarmorling: But `CallerProvider` is stateless, it does not point to any directory. It is just used to obtain the method of the calling test which wants to create an index directory (so its name can be added to the dir name). It's not related to the target dir.

Sanne: oops, sorry got confused by the diff. All clear now :)

gunnarmorling:
Rebased and applied.

PR #947, opened 2015-11-13
HSEARCH-2028

gunnarmorling:
Some doc fixes

### DIFF
--- a/documentation/src/main/asciidoc/mapping.asciidoc
+++ b/documentation/src/main/asciidoc/mapping.asciidoc
@@ -309,6 +309,29 @@ Should you want to make a property sortable but not searchable, still an `@Field
 bridge configuration can be inherited). It can be marked with `store = Store.NO` and `index = Index.NO`, causing
 only the doc value field required for sorting to be added, but not a regular index field.
 
+Fields added through class-level bridges or custom field bridges cannot be marked as sortable by means of the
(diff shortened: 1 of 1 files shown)


Sanne:
regarding the sorting issue: did you see the user's comment?
https://hibernate.atlassian.net/browse/HSEARCH-2021?focusedCommentId=73504&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-73504

I'm afraid he's right. He can add the fields, but we don't allow the sorting to be defined as it fails validation vs the metadata which we know of.

gunnarmorling:
Ah, didn't see his comment, but I suppose he has a valid point there :/ I think we would allow sorting but would transparently bring in the uninverting reader. Seems there is no easy work-around then (besides of relying on the uninverting reader, but then no field for that reader must be marked as sortable).

Sanne:
I'll push the first commit only, which resolves HSEARCH-2028 nicely. With the second commit you're sneaking in an unrelated task ;-)
Don't delete the commit yet though, it's a nice one for HSEARCH-2021, we just need some more step to figure how it should be done properly.

Sanne:
HSEARCH-2028 is merged; let's repurpose the remaining part?

gunnarmorling:
Going to close this one; I will send the doc update for HSEARCH-2021 as part of #949.

PR #949, opened 2015-11-16
HSEARCH-2021 et al.: Allowing to sort on fields contributed by custom bridges

gunnarmorling:
HSEARCH-2021: Adds a new optional `FieldBridge` sub-contract which allows implementations to expose the names of the sortable fields they created; Experimental, to be evolved into more complete means of exposing related meta-data down the road

HSEARCH-2043: Adding option to fail upon uncovered sorts which is useful for testing (by users and us); Defaults to not fail atm. which may be altered later on if we want to encourage users more actively to configure all sorts

HSEARCH-2042: Minor en-passant fix which avoids index uninverting in some cases

### DIFF
--- a/documentation/src/main/asciidoc/mapping.asciidoc
+++ b/documentation/src/main/asciidoc/mapping.asciidoc
@@ -316,25 +316,40 @@ required doc value fields, in addition to the document fields it adds. Furthermo
 the sortable fields created by this bridge:
 
 [[example-adding-docvaluefield]]
-.Adding a sortable field via a custom field bridge
+.Adding sortable fields via a custom field bridge
 ====
 [source, JAVA]
 ----
+/***
+  * Custom field bridge for a Map property which creates sortable fields with the values
+  * of two keys from the map.
+  */
 public class MyClassBridge implements MetadataProvidingFieldBridge {
 
     @Override
     public void set(
             String name, Object value, Document document, LuceneOptions luceneOptions) {
 
-        // add other document field(s)
-        // ...
+        Map<String, String> map = (Map<String, String>) value;
 
-        // add doc value field
-        document.add( new NumericDocValuesField( "myField", 311 );
+        String firstName = map.get( "firstName" );
+        String lastName = map.get( "lastName" );
+
+        // add regular document fields
+        luceneOptions.addFieldToDocument( name + "_firstName", lastName, document );
+        luceneOptions.addFieldToDocument( name + "_lastName", lastName, document );
+
+        // add doc value fields to allow for sorting
+        document.add( new SortedDocValuesField( name + "_firstName", new BytesRef( firstName ) ) );
+        document.add( new SortedDocValuesField( name + "_lastName", new BytesRef( lastName ) ) );
     }
 
-    Set<String> getSortableFieldNames(String name) {
-        return Collections.singleton( "myField" );
+    public Set<String> getSortableFieldNames(String name) {
+        Set<String> sortableFields = new HashSet<>();
+        sortableFields.add( name + "_firstName" );
+        sortableFields.add( name + "_lastName" );
+
+        return sortableFields;
     }
 }
 ----
--- a/documentation/src/main/asciidoc/mapping.asciidoc
+++ b/documentation/src/main/asciidoc/mapping.asciidoc
@@ -350,6 +350,26 @@ The `MetadataProvidingFieldBridge` contract is under active development and cons
 may be altered in future revisions, e.g. by adding further methods, thus breaking existing implementations.
 ====
 
+[[flagging-uncovered-sorts]]
+====== Flagging uncovered sorts
+By default Hibernate Search will transparently create an uninverting index reader when running a query with sorts not
+covered by the sortable fields configured as described above. While this allows to execute the query, relying on index
+uninverting negatively impacts performance.
+
+You thus can optionally advice Hibernate Search to raise an exception when detecting uncovered sorts. To do so, specify
+the following option:
+
+.Disabling automatic index uninverting for uncovered sorts
+====
+[source]
+----
+hibernate.search.index_uninverting_allowed = false
+----
+====
+
+You e.g. may set this to "false" during testing to identify the sortable fields required for your queries and set it to
+"true" in production environments to fall back to index uninverting for uncovered sorts accidentally left over.
+
 [[id-annotation]]
 ===== @Id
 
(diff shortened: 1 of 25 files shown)


gunnarmorling:
@DavideD, @Sanne, force-pushed an update.

The default field name is passed now to the bridges and the config property value is only retrieved once.  Also index inverting is disabled by default now. Added some sortable fields in order to do so, only for some geo-spatial search tests it's enabled (see [HSEARCH-2044](https://hibernate.atlassian.net/browse/HSEARCH-2044)). Also updated the ref guide and ironed out some other glitches.

gunnarmorling:
Pushed one more commit to address HSEARCH-2044. The spatial field bridges expose the meta-data required for sort field validation now; As sorting by distance happens through a custom comparator, no physical doc value field is required, we just needed the right meta-data to let queries use the pseudo-field.

documentation/src/main/asciidoc/mapping.asciidoc
```diff
@@ -309,6 +309,67 @@ Should you want to make a property sortable but not searchable, still an `@Field
...
+            String name, Object value, Document document, LuceneOptions luceneOptions) {
+
+        // add other document field(s)
+        // ...
+
+        // add doc value field
+        document.add( new NumericDocValuesField( "myField", 311 );
+    }
+
+    Set<String> getSortableFieldNames(String name) {
```
DavideD: What's the purpose of the `name`  parameter?

DavideD: Ah, I think I got it. The fields returned are the one created for a particular field. isn't it?

gunnarmorling: Check out the comment at https://github.com/gunnarmorling/hibernate-search/commit/e75d12a7f6f0fb46d6e021426db986454033850d#commitcomment-14427397 for the rationale behind this parameter.

documentation/src/main/asciidoc/mapping.asciidoc
```diff
@@ -309,6 +309,67 @@ Should you want to make a property sortable but not searchable, still an `@Field
...
+.Disabling automatic index uninverting for uncovered sorts
+====
+[source]
+----
+hibernate.search.index_uninverting_allowed = false
+----
+====
+
+You e.g. may set this to "false" during testing to identify the sortable fields required for your queries and set it to
+"true" in production environments to fall back to index uninverting for uncovered sorts accidentally left over.
```
DavideD: I don't know if I would do this, If one has test why should it disable the feature in production.

I think legacy code that you cannot change or without test might be a better explanation

By the way, is a warning logged somewhere instead of an exception in case the property is not set?

documentation/src/main/asciidoc/mapping.asciidoc
DavideD: It think having an example that shows how to use it would make it simpler for a reader to understand what's going on

engine/src/main/java/org/hibernate/search/bridge/MetadataProvidingFieldBridge.java
```diff
@@ -0,0 +1,29 @@
...
+ * @author Gunnar Morling
+ * @hsearch.experimental This contract is currently under active development and may be altered in future releases,
+ * breaking existing implementations.
+ */
+public interface MetadataProvidingFieldBridge extends FieldBridge {
+
+	/**
+	 * Returns the names of the sortable fields created by this field bridge.
+	 *
+	 * @param The default field name; Should be used consistently with
```
DavideD: Is it the default fieldName? Shouldn't be something like `the name of the field, for which the sortable fields are created`?

documentation/src/main/asciidoc/mapping.asciidoc
DavideD: > By the way, is a warning logged somewhere instead of an exception in case the property is not set?

RIght, the javadoc tells that there is warning. Maybe we write it in the documentation as well

engine/src/main/java/org/hibernate/search/query/engine/impl/SortConfigurations.java
```diff
@@ -67,7 +67,7 @@ public SortConfiguration(String indexName, Map<Class<?>, List<SortableFieldMetad
 
 			for ( SortField sortField : sort.getSort() ) {
 				// no doc value field needed for these
-				if ( sortField.getType() == SortField.Type.DOC && sortField.getType() == SortField.Type.SCORE ) {
+				if ( sortField.getType() == SortField.Type.DOC || sortField.getType() == SortField.Type.SCORE ) {
```
DavideD: Is there a test affected by this?

DavideD:
Nice feature :)

Looks good to me,
I think there is just a minor error in a javadoc.

As for the documentation, I think the example could be clearer if it would use the name parameter of the `getSortablefieldNames(String name)`.

documentation/src/main/asciidoc/mapping.asciidoc
gunnarmorling: Ok, I'll expand the example a bit.

documentation/src/main/asciidoc/mapping.asciidoc
gunnarmorling: Dunno, don't think it's worth mentioning every warning in the ref guide.

engine/src/main/java/org/hibernate/search/bridge/MetadataProvidingFieldBridge.java
gunnarmorling: A bridge can create any number of fields in `set()`. It can choose to ignore the name passed to `set()` (which is the name as configured through `@Field`, using the property name as default), or it could e.g. use the passed name as a prefix for all the fields it creates. It's just important that `getSortableFieldNames()` reports the same names as `set()` will create them.

engine/src/main/java/org/hibernate/search/query/engine/impl/SortConfigurations.java
gunnarmorling: Yes, there is `org.hibernate.search.test.query.sorting.SortTest#testResultOrderedByDocId()`.

gunnarmorling:
I've pushed one more commit for expanding the example as you have suggested, @DavideD. Thanks for reviewing this!

engine/src/main/java/org/hibernate/search/bridge/MetadataProvidingFieldBridge.java
DavideD: Ok, anyway the build will fail during the creation of the javadoc if `@param name` is not there.
(now is  `@param The default field name`)

gunnarmorling: Thx, force-pushed an update to fix the comment.

DavideD:
Merged https://github.com/hibernate/hibernate-search/pull/949

Thanks

DavideD:
Do I need to merge this on 5.5?

gunnarmorling:
Yes, please. The sort fix is intended for 5.5. Thx!

DavideD:
> Yes, please. The sort fix is intended for 5.5. Thx!

Done

PR #950, opened 2015-11-23
Performance

Sanne:
New performance tests, and fixing some low hanging fruits already:

 - https://hibernate.atlassian.net/browse/HSEARCH-1867
 - https://hibernate.atlassian.net/browse/HSEARCH-2050
 - https://hibernate.atlassian.net/browse/HSEARCH-2051
 - https://hibernate.atlassian.net/browse/HSEARCH-2052
 - https://hibernate.atlassian.net/browse/HSEARCH-2054

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/impl/FacetHandling.java
+++ b/engine/src/main/java/org/hibernate/search/engine/impl/FacetHandling.java
@@ -1,0 +1,52 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.engine.impl;
+
+import java.io.IOException;
+
+import org.apache.lucene.document.Document;
+import org.apache.lucene.facet.FacetsConfig;
+import org.hibernate.search.util.logging.impl.Log;
+import org.hibernate.search.util.logging.impl.LoggerFactory;
+
+/**
+ * Helper class to manage a {@literal FacetsConfig} instance:
+ * this needs to be lazily allocated as it's a quite heavy object.
+ * Ideally it should be reused across threads, but since field options
+ * are index specific that doesn't suit our current model.
+ */
+public final class FacetHandling {
+
+	private static final Log log = LoggerFactory.make();
+
+	private FacetsConfig facetConfig = null;
+
+	public Document build(Document doc) {
+		if ( facetConfig == null ) {
+			return doc;
+		}
+		else {
+			try {
+				return facetConfig.build( doc );
+			}
+			catch (IOException e) {
+				throw log.errorDuringFacetingIndexing( e );
+			}
+		}
+	}
+
+	public void setMultiValued(String facetName) {
+		facetConfig.setMultiValued( facetName, true );
+	}
+
+	public void enableFacetProcessing() {
+		if ( facetConfig == null ) {
+			facetConfig = new FacetsConfig();
+		}
+	}
+
+}
--- a/engine/src/main/java/org/hibernate/search/engine/metadata/impl/PropertyMetadata.java
+++ b/engine/src/main/java/org/hibernate/search/engine/metadata/impl/PropertyMetadata.java
@@ -64,9 +64,16 @@ public class PropertyMetadata {
 		return propertyAccessorName;
 	}
 
-	// TODO HSEARCH-1867 change return type to set
+	/**
+	 * @deprecated use getFieldMetadataSet() instead
+	 */
+	@Deprecated
 	public List<DocumentFieldMetadata> getFieldMetadata() {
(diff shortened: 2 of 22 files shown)


Sanne:
Added HSEARCH-2050 as well; although there's a better solution for that one which will need a Lucene upgrade if I get my fix in it.

Sanne:
Added the ORM upgrade as well, which aligns the performance tests with Wildfly's ORM.

engine/src/main/java/org/hibernate/search/engine/metadata/impl/PropertyMetadata.java
```diff
@@ -64,11 +64,18 @@ public String getPropertyAccessorName() {
...
-	// TODO HSEARCH-1867 change return type to set
+	/**
+	 * @deprecated use getFieldMetadataSet() instead
+	 */
+	@Deprecated
 	public List<DocumentFieldMetadata> getFieldMetadata() {
 		return new ArrayList<DocumentFieldMetadata>( documentFieldMetadataList );
 	}
 
+	public Set<DocumentFieldMetadata> getFieldMetadataSet() {
```
gunnarmorling: Why did you not just change the return type of the existing method? Please don't tell me you are trying to keep backwards compat for internal types again ;)

engine/src/main/java/org/hibernate/search/engine/impl/FacetHandling.java
```diff
@@ -0,0 +1,49 @@
...
+package org.hibernate.search.engine.impl;
+
+import java.io.IOException;
+
+import org.apache.lucene.document.Document;
+import org.apache.lucene.facet.FacetsConfig;
+import org.hibernate.search.util.logging.impl.Log;
+import org.hibernate.search.util.logging.impl.LoggerFactory;
+
+public final class FacetHandling {
```
gunnarmorling: A short comment describing its purpose would be nice. Probably just move that comment below up here so it's displayed at usage sites of this type.

integrationtest/engine-performance/src/main/java/org/hibernate/search/engineperformance/Launcher.java
```diff
@@ -0,0 +1,68 @@
...
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+/**
+ * Launches all the JMH benchmarks within this project.
+ * <p>
+ * In order to run the benchmarks, do the following:
+ * <ul>
+ * <li>Generate the JMH benchmark classes by running {@code mvn compile -pl performance} from the root dir</li>
```
gunnarmorling: `mvn compile -pl :hibernate-search-engine-performance`; Also the next bullet point is obsolete for HSEARCH,

pom.xml
```diff
@@ -1093,6 +1099,7 @@
                                         <versionRange>[2.1,)</versionRange>
                                         <goals>
                                             <goal>unpack</goal>
+                                            <goal>build-classpath</goal>
```
gunnarmorling: Is that actually needed? Or is it some left-over?

gunnarmorling:
@Sanne Nice, one, looks good overall. See inline comments for some minor remarks.

integrationtest/engine-performance/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+
+# Notes
+
+For best results disable features such as power management, dynamic CPU scaling,
+and run it on a dedicated box which has no other significant services running.
+So the "run it from your IDE" approach is just meant for development of new tests.
+
+
+## TODO
```
gunnarmorling: Let's capture that in a JIRA issue.

pom.xml
Sanne: yes that's needed for a clean Eclipse import

integrationtest/engine-performance/src/main/java/org/hibernate/search/engineperformance/Launcher.java
Sanne: thanks

engine/src/main/java/org/hibernate/search/engine/metadata/impl/PropertyMetadata.java
Sanne: I don't remember the details, but the fact that we had a full JIRA for it rather than having it fixed in the past makes me unformfortable in fixing it in a Final micro release. This needs to be backported as it's a performance issue.

engine/src/main/java/org/hibernate/search/engine/impl/FacetHandling.java
Sanne: +1 done

integrationtest/engine-performance/README.md
Sanne: I'll follow up on these soon.. I have a whole other batch of benchmarks and fixes to cleanup and send PRs for.

Sanne:
Pushed a new version

gunnarmorling:
Merged to master and 5.5.

PR #951, opened 2015-11-24
HSEARCH-2048 Generalizing MetadataProvidingFieldBridge in order to al…

gunnarmorling:
…so expose the type of created fields

### DIFF
--- a/engine/src/main/java/org/hibernate/search/bridge/MetadataProvidingFieldBridge.java
+++ b/engine/src/main/java/org/hibernate/search/bridge/MetadataProvidingFieldBridge.java
@@ -6,7 +6,7 @@
  */
 package org.hibernate.search.bridge;
 
-import java.util.Set;
+import org.hibernate.search.bridge.spi.FieldMetadataBuilder;
 
 /**
  * Optional contract to be implemented by {@link FieldBridge} implementations wishing to expose meta-data related to the
@@ -19,11 +19,11 @@ import java.util.Set;
 public interface MetadataProvidingFieldBridge extends FieldBridge {
 
 	/**
-	 * Returns the names of the sortable fields created by this field bridge.
+	 * Allows this bridge to expose meta-data about the fields it creates.
 	 *
 	 * @param name The default field name; Should be used consistently with
 	 * {@link FieldBridge#set(String, Object, org.apache.lucene.document.Document, LuceneOptions)}.
-	 * @return The names of the sortable fields created by this field bridge. Never {@code null}.
+	 * @param builder Builder for exposing field-related meta-data
 	 */
-	Set<String> getSortableFieldNames(String name);
+	void configureFieldMetadata(String name, FieldMetadataBuilder builder);
 }
--- a/engine/src/main/java/org/hibernate/search/bridge/MetadataProvidingFieldBridge.java
+++ b/engine/src/main/java/org/hibernate/search/bridge/MetadataProvidingFieldBridge.java
@@ -11,6 +11,11 @@ import org.hibernate.search.bridge.spi.FieldMetadataBuilder;
 /**
  * Optional contract to be implemented by {@link FieldBridge} implementations wishing to expose meta-data related to the
  * fields they create.
+ * <p>
+ * Field bridges should implement this contract if they create field(s) with custom names and wish them to mark as
+ * sortable. The required doc value fields will be added by the Hibernate Search engine in that case. Otherwise users
+ * may not (efficiently) sort on such custom fields.
+ * <p>
  *
  * @author Gunnar Morling
  * @hsearch.experimental This contract is currently under active development and may be altered in future releases,
--- a/engine/src/main/java/org/hibernate/search/bridge/spi/FieldMetadataCreationContext.java
+++ b/engine/src/main/java/org/hibernate/search/bridge/spi/FieldMetadataCreationContext.java
@@ -19,7 +19,9 @@ public interface FieldMetadataCreationContext {
 	 * Adds a field to the list of all meta-data.
 	 *
 	 * @param name The name of the field
-	 * @param type The type of the field
+	 * @param type The type of the field; The type of the field created in
+	 * {@link org.hibernate.search.bridge.FieldBridge#set(String, Object, org.apache.lucene.document.Document, org.hibernate.search.bridge.LuceneOptions)}
+	 * must match the type declared for the field here.
 	 * @return a context object for fluent API invocations
 	 */
 	FieldMetadataCreationContext field(String name, FieldType type);
--- a/engine/src/main/java/org/hibernate/search/bridge/spi/FieldMetadataCreationContext.java
+++ b/engine/src/main/java/org/hibernate/search/bridge/spi/FieldMetadataCreationContext.java
@@ -1,0 +1,34 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.bridge.spi;
+
+/**
+ * Allows to configure field-related meta-data.
+ *
+ * @author Gunnar Morling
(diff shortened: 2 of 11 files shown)


gunnarmorling:
Btw. the consumption of the added meta-data is limited to the sorting aspect for now. Obviously, the type info (and any other meta-data we may add down the road) needs to be consumed as well later on.

engine/src/main/java/org/hibernate/search/bridge/MetadataProvidingFieldBridge.java
```diff
@@ -19,11 +19,11 @@
...
-	 * Returns the names of the sortable fields created by this field bridge.
+	 * Allows this bridge to expose meta-data about the fields it creates.
 	 *
 	 * @param name The default field name; Should be used consistently with
 	 * {@link FieldBridge#set(String, Object, org.apache.lucene.document.Document, LuceneOptions)}.
-	 * @return The names of the sortable fields created by this field bridge. Never {@code null}.
+	 * @param builder Builder for exposing field-related meta-data
 	 */
-	Set<String> getSortableFieldNames(String name);
+	void configureFieldMetadata(String name, FieldMetadataBuilder builder);
```
Sanne: great idea to expose the builder!

Could you add a note in the javadoc about  the purpose of this contract?
As a know-nothing user, I don't think I'd understand what benefit I get from implementing this contract vs its parent.
Whilte at it, might be a good idea to warn about possible consequences of a mis-defined metadata?

engine/src/main/java/org/hibernate/search/bridge/spi/FieldType.java
```diff
@@ -0,0 +1,52 @@
...
+
+	/**
+	 * A date, mapped to a String in Lucene.
+	 */
+	STRING_MAPPED_DATE,
+
+	/**
+	 * An integer field.
+	 */
+	INTEGER,
```
Sanne: you're implying these are NumericField(s) right?

engine/src/main/java/org/hibernate/search/bridge/spi/FieldType.java
```diff
@@ -0,0 +1,52 @@
...
+package org.hibernate.search.bridge.spi;
+
+/**
+ * The type of an indexed field.
+ *
+ * @author Gunnar Morling
+ * @hsearch.experimental This contract is currently under active development and may be altered in future releases,
+ * breaking existing users.
+ */
+public enum FieldType {
```
Sanne: This class name is quite overloaded. Do you think we could find a name which implies the (narrow) usage scope?

engine/src/main/java/org/hibernate/search/bridge/spi/FieldType.java
gunnarmorling: Yes, exactly, in Lucene these would end up as `IntField` and the like.

engine/src/main/java/org/hibernate/search/bridge/MetadataProvidingFieldBridge.java
gunnarmorling: +1, I'll expand the comment a bit.

engine/src/main/java/org/hibernate/search/bridge/spi/FieldType.java
gunnarmorling: No good idea really. Open for any suggestions :)

Sanne: I guess it depends if we envision this to be more widely used, or just limited to the FieldBridge metadata?
I'm confused by the existance of details such as STRING_MAPPED_BOOLEAN and _DATE ..

engine/src/main/java/org/hibernate/search/bridge/spi/FieldType.java
Sanne: We should make that clear, as the user needs to make sure the choice is consistent with the writing.
Or we evolve this further and take the different writing methods away from him, so to only have the choice from the metadata available ;) Not sure if we can though, we'd need to accept an Object type for the value.

Sanne: Hummm but we can certainly validate the runtime operations of the FieldBridge to only invoke the methods he promised he would :)

gunnarmorling: Yes, I suppose we could flag a mismatch between anounced meta-data and actual runtime data. I will add a comment on required consistency to the "type" parameter of `field()`.

engine/src/main/java/org/hibernate/search/bridge/spi/FieldType.java
gunnarmorling: How widely we'll use it, not sure yet.

`STRING_MAPPED_BOOLEAN` and `...DATE` are needed for ES as boolean and date are supported types there (i.e. not only String and Number as in plain Lucene; At leas on the ES-API level, I suppose the internal document fields may be the old known ones).

Thinking about it, I probably should rename it to BOOLEAN and DATE.

It's all a bit guess-work at this point really. Yet I think that change will go more into the right direction than the original `getSortableFieldName()` one. We e.g. may expose more options by adding methods to the builder without breaking existing impls. But it's still marked as experimental, so IMO it's no problem if we refine (aka change) stuff afterwards.

Sanne: Ha, thank you now I understand. Makes sense, and +1 to shorten their names. 

Wondering if we could (not now but in a future iteration) essentially have the builder provide the "adder" hook to the user, and replace the LuceneOptions as only way to add things to the "document".

gunnarmorling:
@Sanne Pushed an update with some doc clarifications/fixes.

engine/src/main/java/org/hibernate/search/bridge/spi/FieldType.java
gunnarmorling: Interesting, could you describe a bit more how you would imagine this? Probably a good idea to do it at the meeting :)

DavideD: How does  this enum  relate to `FieldSettingsDescripto$Type`?

Sanne:
merged in both master and 5.5.
I've added one more commit to make the example implementation in the docs fit in the rendered area w/o needing horizontal scrollbars.

engine/src/main/java/org/hibernate/search/bridge/spi/FieldType.java
gunnarmorling: Good question, they are related concepts. `FieldSettingsDescriptor` is used in the public metadata API. We may be possible to unify the two in HS 6.

engine/src/main/java/org/hibernate/search/bridge/spi/FieldMetadataCreationContext.java
```diff
@@ -0,0 +1,36 @@
...
+	 */
+	FieldMetadataCreationContext field(String name, FieldType type);
+
+	/**
+	 * Marks the last added field as sortable.
+	 *
+	 * @param sortable whether the last added field is sortable or not.
+	 * @return this context for fluent API invocations
+	 */
+	FieldMetadataCreationContext sortable(boolean sortable);
```
DavideD: Can we remove the sortable parameter?

I don't think we need it, we could just says that by default is not sortable or add a method `notSortable()`?

gunnarmorling: I thought about it, but we had cases in the past where we also set out without the parameter and ended up needing it e.g. for overriding a config inherited from somewhere else. Not sure whether that may happen here, too, but it seemed a bit more future-proof that way.

DavideD: > for overriding a config inherited from somewhere else

You could do it calling `notSortable()` (maybe there is a better name). At least, we should add a default sortable() without parameter that set it to true. I suppose that is going to be the most common case.

PR #954, opened 2015-11-26
HSEARCH-2056 Some fixes around id handling

gunnarmorling:
https://hibernate.atlassian.net/browse/HSEARCH-2056
https://hibernate.atlassian.net/browse/HSEARCH-1282

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
+++ b/engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
@@ -107,10 +107,33 @@ public class AnnotationMetadataProvider implements MetadataProvider {
 	private final ConfigContext configContext;
 	private final BridgeFactory bridgeFactory;
 
+	private final Class<? extends Annotation> jpaIdClass;
+	private final Class<? extends Annotation> jpaEmbeddedIdClass;
+
 	public AnnotationMetadataProvider(ReflectionManager reflectionManager, ConfigContext configContext) {
 		this.reflectionManager = reflectionManager;
 		this.configContext = configContext;
 		this.bridgeFactory = new BridgeFactory( configContext.getServiceManager() );
+
+		if ( configContext.isJpaPresent() ) {
+			this.jpaIdClass = loadAnnotationClass( "javax.persistence.Id", configContext );
+			this.jpaEmbeddedIdClass = loadAnnotationClass( "javax.persistence.EmbeddedId", configContext );
+		}
+		else {
+			this.jpaIdClass = null;
+			this.jpaEmbeddedIdClass = null;
+		}
+	}
+
+	private Class<? extends Annotation> loadAnnotationClass(String className, ConfigContext configContext) {
+		try {
+			@SuppressWarnings("unchecked")
+			Class<? extends Annotation> idClass = ClassLoaderHelper.classForName( className, configContext.getServiceManager() );
+			return idClass;
+		}
+		catch (ClassLoadingException e) {
+			throw new SearchException( "Unable to load class " + className + " even though it should be present?!" );
+		}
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
```diff
@@ -1751,4 +1779,22 @@ boolean stateInspectionOptimizationsEnabled(TypeMetadata.Builder typeMetadataBui
 		metadataProvidingFieldBridge.configureFieldMetadata( fieldName, builder );
 		return builder.getSortableFields();
 	}
+
+	private boolean hasExplicitDocumentId(XClass type) {
```
gunnarmorling: I am not particularly happy with that additional pass, but I cannot think of a way for avoiding it without re-writing large parts here. We'd have to delay the validation against id default field redefinition from retrieval of the annotation to a later point where we already have visited all properties. But even then we'd have to do another pass of the previously built model, so likely not much would be gained.

Also I am avoiding the repeated class loading of the id types by pulling that out, so that should compensate a bit ;)

engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
```diff
@@ -1751,4 +1779,22 @@ boolean stateInspectionOptimizationsEnabled(TypeMetadata.Builder typeMetadataBui
...
+		List<XProperty> methods = type.getDeclaredProperties( XClass.ACCESS_PROPERTY );
+		for ( XProperty method : methods ) {
+			if ( method.getAnnotation( DocumentId.class ) != null ) {
+				return true;
+			}
+		}
+
+		List<XProperty> fields = type.getDeclaredProperties( XClass.ACCESS_FIELD );
+		for ( XProperty field : fields ) {
+			if ( field.getAnnotation( DocumentId.class ) != null ) {
```
golovnin: Hi Gunnar,
cosmetic improvement (same applies to the line 1786):

isAnnotationPresent( DocumentId.class ) is easier to read than getAnnotation( DocumentId.class ) != null

Best regards,
Andrej Golovnin

orm/src/test/java/org/hibernate/search/test/id/Article.java
```diff
@@ -22,8 +22,10 @@
 @Entity
 @Indexed
 public class Article {
+
 	@Id
 	@GeneratedValue
+	@Field
```
DavideD: How come do you need to add @Field on this test now?

DavideD: Ok, got it

DavideD:
Merged on master https://github.com/hibernate/hibernate-search/pull/954

engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
gunnarmorling: Hey @golovnin, just seeing this comment now. I'd agree if this was using the reflection API, but `XProperty` is part of Hibernate Commons Annotations and there we only have `getAnnotation()`.

Thanks for checking on this PR!

golovnin: XProperty  extends XAnnotatedElement and XAnnotatedElement has a method isAnnotationPresent(Class<T>).

gunnarmorling: Ah, you are right. Was looking for `hasAnnotation()` for some reason...
I'll change it. Thanks!

PR #956, opened 2015-12-09
Test case for HSEARCH-2069

fenrhil:
Here it is.
I put this test case in the ORM package, because the engine does not depend on JPA, which prevents me from adding the @Id annotation that triggers implicit creation of an "id" field and then the whole issue.


DavideD:
Thanks for the test!
I'll close this pull request and send a new one that includes a fix.

PR #957, opened 2015-12-09
Resolving race conditions during flush vs shutdown

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2023


engine/src/test/java/org/hibernate/search/test/backend/lucene/ResourcesClosedInOrderTest.java
```diff
@@ -0,0 +1,129 @@
...
+import org.hibernate.search.testsupport.setup.SearchConfigurationForTest;
+import org.hibernate.search.testsupport.setup.TransactionContextForTest;
+import org.jboss.byteman.contrib.bmunit.BMRule;
+import org.jboss.byteman.contrib.bmunit.BMRules;
+import org.jboss.byteman.contrib.bmunit.BMUnitRunner;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import static org.hibernate.search.testsupport.BytemanHelper.isEventStackEmpty;
```
DavideD: A curiosity: is it your IDE default configuration that sets static import after all the others?

DavideD:
Looks good.
It's great that you have solved this issue. I will merge it.

engine/src/test/java/org/hibernate/search/test/backend/lucene/ResourcesClosedInOrderTest.java
Sanne: No, I actually type that "manually" and like them to be grouped last.

gunnarmorling: Didn't want to jump on that, but FTR it's causing me headaches, too.

By default Eclipse is sorting them first, so this requires a specific set-up for the import organizer. Unfortunately, at this point both styles are used throughout the code base, causing frequent re-orgs when editing files.

I'd be perfectly fine with the default setting (statics first), but if that's your preference for HS, let's do it consistently.

DavideD: > I'd be perfectly fine with the default setting (statics first), but if that's your preference for HS, let's do it consistently. Just realize people are very likely to do it different by default, causing addition of files with the "wrong style" accidentally (as in my case).

Yes, that is what happens to me very often, expecially when I remove unused import with Ctrl-O in eclipse.

I would prefer the static first as well, but at least now I know why I had this problem before.

Sanne: I'm fine either way, I think I always had them last (long before Eclipse did mess with them) so the habit (and somewhat my expectation) has them on the bottom but we can change that if you prefer.

DavideD:
Merge https://github.com/hibernate/hibernate-search/pull/957

engine/src/test/java/org/hibernate/search/test/backend/lucene/ResourcesClosedInOrderTest.java
Sanne: @gunnarmorling if this has to be come a rule, then it's your turn to make it a checkstyle rule: I don't want all the hate directed to me only :-)

DavideD: THis might work: http://checkstyle.sourceforge.net/apidocs/com/puppycrawl/tools/checkstyle/checks/imports/ImportOrderCheck.html

DavideD:
I've merged it on master, I have is a conflict on 5.5. Could you have a look at it?

Sanne:
> I've merged it on master, I have is a conflict on 5.5. Could you have a look at it?

Ok, it was just some whitespace in the Logger. Pushed now.

DavideD:
Thanks

PR #959, opened 2015-12-10
HSEARCH-2069 SortableField on id field will make initialization crash if the entity is @IndexedEmbedded in another

DavideD:
Follow up for https://github.com/hibernate/hibernate-search/pull/956

https://hibernate.atlassian.net/browse/HSEARCH-2069

### DIFF
--- a/orm/src/test/java/org/hibernate/search/test/embedded/sorting/EmbeddedSortableIdFieldTest.java
+++ b/orm/src/test/java/org/hibernate/search/test/embedded/sorting/EmbeddedSortableIdFieldTest.java
@@ -1,0 +1,26 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.test.embedded.sorting;
+
+import org.hibernate.search.test.SearchTestBase;
+import org.hibernate.search.testsupport.TestForIssue;
+import org.junit.Test;
+
+@TestForIssue(jiraKey = "HSEARCH-2069")
+public class EmbeddedSortableIdFieldTest extends SearchTestBase {
+
+	@Test
+	public void test() {
+		// Nothing to do here: if the test fails, it will fail during initialization.
+	}
+
+	@Override
+	public Class<?>[] getAnnotatedClasses() {
+		return new Class[] { Level1.class, Level2SortableId.class };
+	}
+
+}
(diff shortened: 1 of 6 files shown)


gunnarmorling:
Hey @fenrhil, could you sign the JBOSS contributor license agreement if you haven't done so? It's a simple web app living at http://cla.jboss.org/. Thanks!

fenrhil:
Hi @gunnarmorling , I already did quite a while ago, in the context of this pull request: https://github.com/hibernate/hibernate-search/pull/523
So I thinks everything's OK. Let me know if it isn't!

gunnarmorling:
Ok, cool, thanks!

orm/src/test/java/org/hibernate/search/test/embedded/sorting/EmbeddedSortableIdFieldTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+			session.save( hero );
+			session.save( lex );
+
+			transaction.commit();
+		}
+	}
+
+	@Test
+	public void testExceptionForNotEmbeddedField() {
```
gunnarmorling: I am confused a bit by this one :) What's the rationale for expecting this exception here?

For the issue at hand, to me it'd make more sense to test the following:

* one can sort on an embedded id if `@Sortable` is present (and the embedded id is configured to be indexed)
* a `@Sortable` id does not cause harm if its entity is embedded into another one but the embedded id is not to be indexed

orm/src/test/java/org/hibernate/search/test/embedded/sorting/EmbeddedSortableIdFieldTest.java
```diff
@@ -0,0 +1,100 @@
...
+			FullTextQuery fullTextQuery = fullTextSession.createFullTextQuery( q, Hero.class );
+			fullTextQuery.setSort( sort );
+			fullTextQuery.list();
+
+			transaction.commit();
+		}
+	}
+
+	@Test
+	public void testSortableOnIndexedFieldId() {
```
gunnarmorling: @DavideD: How weird: this test fails if I disable the other test method?!

The failure seems correct though (so no idea why the test passes if the other method runs, too): 

    org.hibernate.search.exception.SearchException: HSEARCH000301: Requested sort field(s) id are not configured for entity type org.hibernate.search.test.embedded.sorting.Hero

Indeed the `Hero` id is not marked as sortable. If I do that, the test still fails because the sort field is of the wrong type (ids are indexed as String by default). If I use the correct sort type, the test still fails because `Hero` has no field `name`,  only `villain.name`.

I believe some re-work is due ;)

DavideD:
Thank @gunnarmorling I'll fix the test and re-open this pull request

PR #960, opened 2015-12-10
Elasticsearch backend - Follow-Up

gunnarmorling:
@Sanne, @DavideD: Next round for the ES backend :) What it does:

* Lets tests from "orm" module run for ES as well (failing ones are excluded in the POM)
* Translates (simple) Lucene queries into ES ones
* Take bridge-provided metadata into account for creating the ES schema (via the recently created contract)
* Multi-tenancy, purge-all
* Many more improvements across the board, I enabled previously failing tests as I moved along

I think with this one in, we may work with dedicated JIRA issues for further tasks, they should be focused enough now to warrant that and we should be able to work on them in parallel. It should be driven by looking into remaining failing tests from "orm", I expect many more to pass with the right adjustments to the ES backend. Only a hand-ful are really specific to Lucene/directory stuff etc. and would be excluded from the ES run on a permanent basis.

Let me know what you think :)


orm/src/test/java/org/hibernate/search/test/batchindexing/DatabaseMultitenancyTest.java
```diff
@@ -203,20 +200,20 @@ public void shouldSearchOtherTenantsDocuments() throws Exception {
 	private void rebuildIndexWithMassIndexer(Class<?> entityType, String tenantId) throws Exception {
 		FullTextSession session = Search.getFullTextSession( openSessionWithTenantId( tenantId ) );
 		session.createIndexer( entityType ).purgeAllOnStart( true ).startAndWait();
-		IndexReader indexReader = session.getSearchFactory().getIndexReaderAccessor().open( entityType );
-		long tenantIdTermFreq = indexReader.docFreq( new Term( DocumentBuilderIndexedEntity.TENANT_ID_FIELDNAME, tenantId ) );
+//		IndexReader indexReader = session.getSearchFactory().getIndexReaderAccessor().open( entityType );
```
gunnarmorling: Just seeing these, I'll re-work that one. Should be easily doable with the more abstract test helper I've created.

gunnarmorling: Reworked that commit now.

gunnarmorling:
@Sanne, any thoughts?

Sanne:
glad to see this still rebases cleanly.. that was bothering me ;-)

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticSearchHSQueryImpl.java
```diff
@@ -100,11 +115,8 @@ public DocumentExtractor queryDocumentExtractor() {
 
 	@Override
 	public int queryResultSize() {
-		if ( resultSize == null ) {
```
gsmet: This change introduces a bug as searchResult is not reset in clearCachedResults(). I fixed it in my branch by also resetting it but if you plan to merge this sooner rather than later, I thought it was worth mentioning.

Sanne:
@gunnarmorling could you rebase it please? there's a compilation error, should be trivial to fix:

    [ERROR] COMPILATION ERROR : 
    [ERROR] /home/sanne/workspaces/hibernate/hibernate-search-root/engine/src/test/java/org/hibernate/search/testsupport/backend/LeakingBackendQueueProcessor.java:[32,8] org.hibernate.search.testsupport.backend.LeakingBackendQueueProcessor is not abstract and does not override abstract method closeIndexWriter() in org.hibernate.search.backend.spi.BackendQueueProcessor
    [INFO] 1 error

Sanne:
nevermind, since it looks like you're not around I'll do it.

Sanne:
@Hibernate-CI retest this please

Sanne:
Rebasing introduced some problems; some were existing problems which weren't noticeable, some issues were related to new tests needing to be applied some of the new helpers.

But also several tests were failing already before the rebase - which I hadn't noticed initially.
Among these:
 - the existing JGroups / ES tests didn't set 'index_management_strategy' so the index isn't being created.
 - the OSGi / Karaf integration tests were broken by new imports having the -orm module depend on an -impl package within -engine.
 - CollectionUpdateEventTest fails for me, even on your version of the branch. I disabled it so we can follow up later.

Sanne:
Replaced by PR #980

gsmet:
I confirm CollectionUpdateEventTest also fails for me as I reported it in a previous email.

PR #963, opened 2015-12-16
HSEARCH-2069 Follow-up regarding validation of the @NumericFields

gsmet:


Sanne:
Thanks! I've merge it on both master and 5.5, and added one more commi to apply my own remark from the code review (as Davide is on holidays already)

PR #964, opened 2015-12-18
HSEARCH-2073 LuceneHSQuery cannot be deserialized

gustavonalle:
https://hibernate.atlassian.net/browse/HSEARCH-2073

### DIFF
--- a/engine/src/main/java/org/hibernate/search/query/engine/impl/LuceneHSQuery.java
+++ b/engine/src/main/java/org/hibernate/search/query/engine/impl/LuceneHSQuery.java
@@ -7,7 +7,6 @@
 package org.hibernate.search.query.engine.impl;
 
 import java.io.IOException;
-import java.io.Serializable;
 import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/hibernate/search/query/engine/impl/LuceneHSQuery.java
```diff
@@ -66,7 +65,7 @@
  * @author Emmanuel Bernard
  * @author Hardy Ferentschik
  */
-public class LuceneHSQuery extends AbstractHSQuery implements HSQuery, Serializable {
```
Sanne: why removing the Serializable in this case?

gustavonalle: It was redundant since I made the parent Serializable

Sanne: yes I realize it's redundant, but wondering what is the contract you're needing.
In practice, it's `LuceneHSQuery` which you need to be serializable, right?

We could make it serializable without having to mandate the contract on its parent, or we might want to mark both as Serializable as an explicit marker (reminder).

gustavonalle: > In practice, it's LuceneHSQuery which you need to be serializable, right?
 
Yep, that's the one causing trouble, although I haven't tested yet with ```ElasticSearchHSQueryImpl``` so I though it'd make sense for both to be Serializable

Sanne:
Replaced by #968

PR #965, opened 2015-12-18
HSEARCH-2074 Test case

fenrhil:
... for "ContainedIn triggers too much work when there is not "depth" on the corresponding @IndexedEmbedded"

https://hibernate.atlassian.net/browse/HSEARCH-2074


Sanne:
I'm closing the PR for now as we need a fix still. Let's not delete this though, as people working on HSEARCH-2074 will need this. Thanks!

PR #967, opened 2015-12-22
HSEARCH-2074 Bugfix

fenrhil:
Follow-up for #965
See https://hibernate.atlassian.net/browse/HSEARCH-2074 for details.

I ran all the tests I could locally, and it seems OK, but a full CI build is required to test everything.

**About the fix**
When dealing with includePaths, working with `depth` only was not enough. I had to keep in memory the fact that only some paths were included in the last `@IndexedEmbedded`, which means that if the next `@IndexedEmbedded` does not include any of these, we can safely stop recursing.

**Also**
Please note the following modifications:
 * I had trouble understanding what AbstractDocumentBuilder.updateDepth (now AbstractDocumentBuilder.updateContainedInRecursionContext) did with depths, so I took the  liberty of refactoring the depth mechanism a little bit. I think it's a bit clearer, but I can try to reverse that if you disagree.
 * I think there was a bug in AnnotationMetadataProvider.isCorrespondingIndexedEmbedded: when no `mappedBy` was set on the property holding the `@ContainedIn` annotation, then every property of the contained type on the container entity was deemed "corresponding". I added a check to only deem them "corresponding" if there is a matching `mappedBy` on the reverse property.


Sanne:
Hi @fenrhil , is it possible something is missing?
If I run the tests, it fails:

    Failed tests: 
          WorkDoneOnEntitiesTest.testWorkShouldBeExecutedOnPerson:108 expected:<1> but was:<0>
    
     Tests run: 656, Failures: 1, Errors: 0, Skipped: 2

The approach looks good at an high level, I need to look into details yet. Thanks for taking this complex change!

fenrhil:
Strange. I ran the test again, and it failed. I ran it again for debug, and it succeeded. Ran it again, fail. And so on, without a clear pattern.

I'll see what I can do. Something is really wrong here.

fenrhil:
@Sanne, here you are:  	35e6c54. Very simple, yet a bit hard to spot :)
Turns out it was a pre-existing bug, though; my changes may have made it more frequent, but it was already there. Maybe it was already mentioned in a JIRA ticket?

Sanne:
Interesting! no there are no issues explicitly pointing to this.

Thanks a lot, I'll try to understand this better tomorrow, and hopefully add yet another test as clearly something isn't being covered.

Sanne:
Hi @fenrhil it took me a bit to understand it all as well, it's been a while that I had not patched this logic :)

I've merged your changes but squashing your commits in one, no need to keep the formatting changes separate.

Merged in `master` and `5.5`

PR #968, opened 2015-12-22
Query serialization

Sanne:
https://hibernate.atlassian.net/browse/HSEARCH-2073

@gustavonalle your same patch, but adding a unit test to catch Serialization issues on this class in the future.

The Lucene queries are not serializable, so I'm surprised you need this. Does my test model the Infinispan use case closely, i.e. by re-setting the Lucene Query instance?

### DIFF
--- a/engine/src/test/java/org/hibernate/search/test/query/serialization/QuerySerializationTest.java
+++ b/engine/src/test/java/org/hibernate/search/test/query/serialization/QuerySerializationTest.java
@@ -1,0 +1,81 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.test.query.serialization;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+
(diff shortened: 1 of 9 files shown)


engine/src/test/java/org/hibernate/search/test/query/serialization/QuerySerializationTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+
+/**
+ * Verifies Serialization of the LuceneHSQuery object.
+ * Infinispan needs to serialize these.
+ *
+ * @author Sanne Grinovero
+ */
+public class QuerySerializationTest {
+
```
gustavonalle: FYI, there's a disabled test in  ```org.hibernate.search.test.engine.QuerySerializationTest``` that looks very similar to this one

Sanne: Ah, thanks! I should clean that up. Let me see if I delete one ore merged them.

gustavonalle:
> Does my test model the Infinispan use case closely, i.e. by re-setting the Lucene Query instance?

It's close enough. To match it exactly, you'd need to use JBoss Marshaller and write a custom externalizer for the Lucene Query  :smile:

Sanne:
ok, fair enough. I've merged it, and deleted the other test: I preferred my new test as it was living in the right module (/engine).
The other JIRA is closed as outdated.

PR #970, opened 2016-01-08
HSEARCH-2078 Upgrade to Apache Lucene 5.4.0

Sanne:
https://hibernate.atlassian.net/browse/HSEARCH-2078


gunnarmorling:
Looks good, merge yourself? Only a property for the gazillion repetitions of the version in the OSGi header maybe?

Sanne:
> Only a property for the gazillion repetitions of the version in the OSGi header maybe?

Yes, I was tempted to do that too, but couldn't finish it. I can give it a shot now..

Sanne:
I've merged this one, will prepare a separate issue and PR for the OSGi properties as it became non-trivial..

PR #973, opened 2016-01-18
HSEARCH-1461 Statistics copied over during addClasses(..) invocation

mrizzi:
Added `MutableFactoryTest.testAddingClassSimpleAPIwithJMX()` to test that the statistics are still collected after the method `addClasses(Class<?>... classes)` invocation, verifying that the total number of queries (`searchQueryCount`) and the size of the set of the indexed classes (`indexedClasses`) are both updated in the right way.
The JMX console and the statistics generation are both activated.

`SearchFactoryTest` added the method `Statistics getStatistics()` to this interface
`ImmutableSearchFactory` added the code to handle the copy of the previous statistics. The field `isStatisticsEnabled` is setted only in case of new creation.
`MutableSearchFactoryState` added the copy of the statistics and the getter and setter for the Statistics object itself

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/impl/ImmutableSearchFactory.java
+++ b/engine/src/main/java/org/hibernate/search/engine/impl/ImmutableSearchFactory.java
@@ -131,15 +131,20 @@ public class ImmutableSearchFactory implements ExtendedSearchIntegratorWithShare
 		this.timeoutExceptionFactory = state.getDefaultTimeoutExceptionFactory();
 		this.timingSource = state.getTimingSource();
 		this.mapping = state.getProgrammaticMapping();
-		this.statistics = new StatisticsImpl( this );
+		if ( state.getStatistics() == null ) {
+			this.statistics = new StatisticsImpl( this );
+			boolean statsEnabled = ConfigurationParseHelper.getBooleanValue(
+					configurationProperties, Environment.GENERATE_STATS, false
+			);
+			this.statistics.setStatisticsEnabled( statsEnabled );
+		}
+		else {
+			this.statistics = (StatisticsImpl) state.getStatistics();
+		}
 		this.indexMetadataIsComplete = state.isIndexMetadataComplete();
 		this.isDeleteByTermEnforced = state.isDeleteByTermEnforced();
 		this.isIdProvidedImplicit = state.isIdProvidedImplicit();
 		this.indexManagerFactory = state.getIndexManagerFactory();
-		boolean statsEnabled = ConfigurationParseHelper.getBooleanValue(
-				configurationProperties, Environment.GENERATE_STATS, false
-		);
-		this.statistics.setStatisticsEnabled( statsEnabled );
 
 		this.enableDirtyChecks = ConfigurationParseHelper.getBooleanValue(
 				configurationProperties, Environment.ENABLE_DIRTY_CHECK, true
--- a/engine/src/main/java/org/hibernate/search/spi/impl/SearchFactoryState.java
+++ b/engine/src/main/java/org/hibernate/search/spi/impl/SearchFactoryState.java
@@ -25,6 +25,7 @@ import org.hibernate.search.indexes.impl.IndexManagerHolder;
 import org.hibernate.search.query.engine.spi.TimeoutExceptionFactory;
 import org.hibernate.search.spi.IndexingMode;
 import org.hibernate.search.spi.InstanceInitializer;
+import org.hibernate.search.stat.Statistics;
 
 /**
  * Represents the sharable state of a search factory
@@ -77,4 +78,6 @@ public interface SearchFactoryState {
 	IndexManagerFactory getIndexManagerFactory();
 
 	boolean enlistWorkerInTransaction();
+
+	Statistics getStatistics();
 }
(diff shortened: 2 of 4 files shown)


engine/src/main/java/org/hibernate/search/spi/impl/SearchFactoryState.java
```diff
@@ -77,4 +78,6 @@
 	IndexManagerFactory getIndexManagerFactory();
 
 	boolean enlistWorkerInTransaction();
-}
+
+	Statistics getStatistics();
+}
```
DavideD: Could you add a new line here?

Sanne:
add to whitelist

engine/src/main/java/org/hibernate/search/engine/impl/ImmutableSearchFactory.java
```diff
@@ -131,15 +131,20 @@ public ImmutableSearchFactory(SearchFactoryState state) {
...
-		this.statistics = new StatisticsImpl( this );
+		if ( state.getStatistics() == null ) {
+			this.statistics = new StatisticsImpl( this );
+			boolean statsEnabled = ConfigurationParseHelper.getBooleanValue(
+					configurationProperties, Environment.GENERATE_STATS, false
+			);
+			this.statistics.setStatisticsEnabled( statsEnabled );
+		}
+		else {
+			this.statistics = (StatisticsImpl) state.getStatistics();
```
gunnarmorling: Is that cast safe?

Sanne:
hi @mrizzi Thanks a lot for fixing this!

engine/src/test/java/org/hibernate/search/test/configuration/mutablefactory/MutableFactoryTest.java
```diff
@@ -152,6 +155,77 @@ public void testAddingClassSimpleAPI() throws Exception {
 		sf.close();
 	}
 
+	@Test
+	public void testAddingClassSimpleAPIwithJMX() throws Exception {
+		SearchIntegrator sf = new SearchIntegratorBuilder().configuration( new SearchConfigurationForTest().addClass( A.class ).addProperty( "hibernate.search.jmx_enabled" , Boolean.TRUE.toString() ).addProperty( "hibernate.search.generate_statistics", Boolean.TRUE.toString() ).addProperty( "com.sun.management.jmxremote", Boolean.TRUE.toString() ) ).buildSearchIntegrator();
+
```
Sanne: right.. I realize our style formatter is a bit pedantic, but it will allow you to break down lines in a more understandable shape.

could you reformat it for better readability?

    SearchIntegrator sf = new SearchIntegratorBuilder()
				.configuration(
						new SearchConfigurationForTest()
							.addClass( A.class )
							.addProperty( "hibernate.search.jmx_enabled" , Boolean.TRUE.toString() )
							.addProperty( "hibernate.search.generate_statistics", Boolean.TRUE.toString() )
							.addProperty( "com.sun.management.jmxremote", Boolean.TRUE.toString() ) )
				.buildSearchIntegrator();

engine/src/test/java/org/hibernate/search/test/configuration/mutablefactory/MutableFactoryTest.java
```diff
@@ -152,6 +155,77 @@ public void testAddingClassSimpleAPI() throws Exception {
...
+		entityInfos = hsQuery.queryEntityInfos();
+		assertEquals( 1 , entityInfos.size() );
+		assertEquals( 3 , sf.getStatistics().getSearchQueryExecutionCount() );
+		assertEquals( 3 , sf.getStatistics().getIndexedClassNames().size() );
+
+		hsQuery.getTimeoutManager().stop();
+
+		sf.close();
+	}
```
Sanne: the test looks great!

engine/src/main/java/org/hibernate/search/engine/impl/ImmutableSearchFactory.java
Sanne: right, I have the same question. Would be nice if you could improve the patch to not need the cast, or if that's not possible to add a comment to explain why it is safe.

mrizzi: The cast is safe because `StatisticsImpl` is created only in the `ImmutableSearchFactory`'s constructor itself, 4 rows before the cast.

Otherwise, setting up `StatisticsImpl` as the return type of the the `SearchFactoryState.getStatistics()` method avoids the use of the cast but, honestly, i prefer to have the `Statistics` interface as return type rather than a class implementing that interface (i.e. `StatisticsImpl`).
Anyway, if we change `SearchFactoryState.getStatistics()`'s return type to `StatisticsImpl`, we have to change the `MutableSearchFactoryState` class that implements `SearchFactoryState` and the two factories `ImmutableSearchFactory` and `MutableSearchFactory`.
In this way in the `ImmutableSearchFactory` class constructor there's no need to cast anymore.

engine/src/test/java/org/hibernate/search/test/configuration/mutablefactory/MutableFactoryTest.java
mrizzi: Yes Sanne, you're right.
Just wrote it down at the beginning of the work and forgot to format in the right style.

Just pushed  the class well formatted.

Sanne:
I didn't see the new commit: remember that github doesn't send notifications if you don't add a comment.

Merged! Thanks a lot.

PR #953, opened 2015-11-25
Preview of the async backend

s4ke:
I am currently in the progress of merging my GenericJPA project into a submodule of hibernate-search.  As of now I am finished with integrating the async trigger based backend into Hibernate Search ORM.

This pull request is a preview of my work on this part. (The merge is by no means finished, so please don't merge this PR)

PR #979, opened 2016-02-15
HSEARCH-2104 preview of the async backend (only the async parts, no generification)

s4ke:
This is the follow up PR for this:

https://github.com/hibernate/hibernate-search/pull/953

It's still quite large, but only contains the async backend.

Also keep in mind that this PR contains some intermediary work needed for the generic version of Hibernate Search which cannot be removed. All the abstraction layers needed to add the generic version are contained in this.


s4ke:
@gunnarmorling sorry, i can't make this much smaller.

s4ke:
Also, this does not contain the integration tests for the triggers as these require the generified version to run.

genericjpa/LICENSE.txt
```diff
@@ -0,0 +1,502 @@
+ GNU LESSER GENERAL PUBLIC LICENSE
```
Sanne: We don't need this, the license is in the root already.

engine/src/test/java/org/hibernate/search/testsupport/TestConstants.java
```diff
@@ -92,4 +92,4 @@ public static boolean arePerformanceTestsEnabled() {
 			return getClassContext()[2];
 		}
 	}
-}
+}
```
Sanne: here is proof that the PR could be smaller :)

genericjpa/async-updates/src/main/java/org/hibernate/search/genericjpa/jpa/util/impl/JPAEntityManagerFactoryWrapper.java
```diff
@@ -0,0 +1,29 @@
+package org.hibernate.search.genericjpa.jpa.util.impl;
```
Sanne: It looks like these classes are missing the license header.. which makes me think that somehow you got checkstyle disabled here as it would have failed the build and pointed it out.
Could you please make sure that checkstyle is run on your modules too? It might find some formatting issues too.. I hope not too many to be a PITA but it's certainly nicer to send the initial contribution with the right formatting than having to fix it later.

genericjpa/async-updates/src/main/java/org/hibernate/search/genericjpa/jpa/util/impl/JPAQueryWrapper.java
```diff
@@ -0,0 +1,43 @@
+package org.hibernate.search.genericjpa.jpa.util.impl;
+
+import javax.persistence.Query;
+import java.util.List;
+
+import org.hibernate.search.genericjpa.jpa.util.impl.QueryWrapper;
+
+/**
+ * Created by Martin on 11.11.2015.
```
Sanne: These comments are not very useful :) Feel free to use the "@author" tag, that's the standard we use.

genericjpa/copyright.txt
```diff
@@ -0,0 +1,4 @@
+Martin Braun
```
Sanne: this is also redundant, the copyright is applied from the root of the repository. Your name is there already right?

genericjpa/db/src/main/java/org/hibernate/search/genericjpa/annotations/IdColumn.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.lang.annotation.Retention;
+
+import org.hibernate.search.genericjpa.db.ColumnType;
+
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * @author Martin Braun
+ * @hsearch.experimental
```
Sanne: :+1:

genericjpa/async-updates/src/main/java/org/hibernate/search/genericjpa/jpa/util/impl/JPAQueryWrapper.java
s4ke: yeah, seems like you're right, but these were automatically generated (and i haven't changed the default).

genericjpa/db/src/main/java/org/hibernate/search/genericjpa/db/events/index/impl/IndexUpdater.java
```diff
@@ -0,0 +1,320 @@
...
+					}
+					finally {
+						latch.countDown();
+					}
+
+				}
+		);
+		try {
+			latch.await();
+			//while we did things on a different thread we still
```
Sanne: So.. it's not really async is it?
If you just want to make sure underlying errors get reported correctly you could use the org.hibernate.search.exception.ErrorHandler

genericjpa/async-updates/src/main/java/org/hibernate/search/genericjpa/jpa/util/impl/JPAEntityManagerFactoryWrapper.java
s4ke: yeah. If i remember correctly, Hibernate-Search core requires Java 7 and checkstyle therefore doesn't work on my code which currently requires Java 8.

genericjpa/db/src/main/java/org/hibernate/search/genericjpa/db/events/index/impl/IndexUpdater.java
Sanne: Also notice that with this pattern you're essentially introducing a bottleneck: all threads coming here which have some update to write will have to delegate to the single thread of your executor, and then wait for it to complete the task. That's a very significant throughput issue.

genericjpa/db/src/main/java/org/hibernate/search/genericjpa/db/events/index/impl/IndexUpdater.java
```diff
@@ -0,0 +1,320 @@
...
+			//want to throw the exceptions from there
+			//so the AsyncUpdateSource stumbles on this Exception
+			if ( exception[0] != null ) {
+				throw exception[0];
+			}
+		}
+		catch (InterruptedException e) {
+			throw new SearchException( e );
```
Sanne: Careful: you can't just catch an InterruptedException and ignore the interruption state of the thread. You probably want to reset that.

genericjpa/db/src/test/java/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,3 @@
+Manifest-Version: 1.0
+Class-Path: 
+
```
Sanne: ?

genericjpa/settings-example.xml
```diff
@@ -0,0 +1,294 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Sanne: you should delete this one.

genericjpa/db/src/main/java/org/hibernate/search/genericjpa/db/events/index/impl/IndexUpdater.java
s4ke: Hmm. sounds about right, but if we scroll updates from the database, we should only be doing things from one thread only.

genericjpa/standalone/src/main/java/org/hibernate/search/SearchFactory.java
```diff
@@ -0,0 +1,107 @@
...
+import org.hibernate.search.stat.Statistics;
+
+/**
+ * Provide application wide operations as well as access to the underlying Lucene resources.
+ *
+ * @author Emmanuel Bernard
+ * @author Hardy Ferentschik
+ * @author Sanne Grinovero
+ */
+public interface SearchFactory {
```
Sanne: We need to find an alternative name for this one. It's a very bad idea to have two classes with the same name - even if they are in the same project and the two modules are not meant to be used together!

s4ke: ah good catch. In my code i used this in the same way as the one from orm. This has to be put into some common module.

s4ke: some "API" module maybe?

genericjpa/db/src/main/java/org/hibernate/search/genericjpa/db/events/index/impl/IndexUpdater.java
```diff
@@ -0,0 +1,320 @@
...
+ * @author Martin Braun
+ */
+public final class IndexUpdater {
+
+	// TODO: think of a clever way of doing batching here
+	// or maybe leave it as it is
+
+	// TODO: unit test this with several batches
+
+	private static final Logger LOGGER = Logger.getLogger( IndexUpdater.class.getName() );
```
s4ke: Also, I should probably replace all of these with the default way of logging in Hibernate Search.

genericjpa/README.md
```diff
@@ -0,0 +1,84 @@
+Hibernate-Search-JPA
+====================
```
gunnarmorling: I am a bit confused, I wouldn't expect this entire module ("genericjpa") as part of this PR, given you say it's the async backend only. Maybe the split between the two things is not entirely clear to me. What is it you consider as "async backend" and how's that different from "genericjpa"?

s4ke: oh, this is a relic from the old version that had its own repository.

As of now the async backend is a side product of my genericjpa efforts and is therefore in the genericjpa folder.

s4ke: The async backend was originally created while building genericjpa.

The "async-updates" (asynchronous updates for Hibernate Search ORM) module is intended to be used in a context where at least the modules "util" (general utility classes nothing fancy), "db" (everything trigger related that is not orm specific) and "standalone" (standalone abstraction of the engine, you can already see that this is designed to be used by the actual genericjpa module, which will come later) are available.

s4ke: Maybe the first few pages of the chapters 5,6 and 7 from here can help to understand why I did things like this and where this code comes from:

https://github.com/s4ke/Bachelor-Thesis/raw/Update-System-rework/tex/src/bachelor_thesis.pdf

gunnarmorling:
@s4ke I still see the "genericjpa" module as part of this PR. I assumed this would not be the case anymore as we discussed earlier today only the async backend would be contained?

Also is there a high-level description of the overall architecture? Like one page or so. I'm looking for an angle to dive into this.

genericjpa/common-apis/src/main/java/org/hibernate/search/genericjpa/metadata/impl/MetadataRehasher.java
```diff
@@ -0,0 +1,124 @@
...
+import org.hibernate.search.exception.AssertionFailure;
+import org.hibernate.search.metadata.NumericFieldSettingsDescriptor.NumericEncodingType;
+
+/**
+ * @author Martin Braun
+ */
+public final class MetadataRehasher {
+
+	public List<RehashedTypeMetadata> rehash(List<TypeMetadata> originals) {
+		return originals.stream().map( this::rehash ).collect( Collectors.toList() );
```
gunnarmorling: This is using Java 8. Is that really what we want? As said before I'd vote against it if there is no real advantage from the user's PoV. So I'd like to make sure we all agree on this move.

/CC @Sanne.

genericjpa/common-apis/src/main/java/org/hibernate/search/annotations/InIndex.java
```diff
@@ -0,0 +1,27 @@
...
+ * This annotation is needed to specify that a class is located in an index. <br>
+ * <br>
+ * This is only needed (but always allowed), if objects of subclasses are supplied to indexing related operations. Then
+ * the "least top level" class that is annotated with @InIndex is used. operations
+ *
+ * @author Martin Braun
+ */
+@Target({TYPE})
+@Retention(RUNTIME)
+public @interface InIndex {
```
gunnarmorling: Why is that needed? Making a user annotate an entity with `@Indexed` and `@InIndex` seems rather odd.

genericjpa/common-apis/src/main/java/org/hibernate/search/genericjpa/metadata/impl/RehashedTypeMetadata.java
```diff
@@ -0,0 +1,128 @@
...
+
+import org.hibernate.annotations.common.reflection.XProperty;
+import org.hibernate.search.backend.spi.SingularTermDeletionQuery;
+import org.hibernate.search.engine.metadata.impl.DocumentFieldMetadata;
+import org.hibernate.search.engine.metadata.impl.TypeMetadata;
+
+/**
+ * @author Martin Braun
+ */
+public final class RehashedTypeMetadata {
```
gunnarmorling: What's the purpose of this class? There should always be a meaningful class-level comment. Also some more high-level documentation is needed (e.g. package-level JavaDocs) to describe the overall mechanism. That's the first thing I'd really need to grasp what's going on here :) I.e. no need for polished JavaDocs everywhere, but something providing an overview would help.

genericjpa/async-updates/src/main/java/org/hibernate/search/db/events/jpa/impl/SQLJPAAsyncUpdateSourceProvider.java
```diff
@@ -0,0 +1,204 @@
...
+			String query,
+			boolean canFail) {
+		try {
+			//we use a new EntityManager here everytime, because
+			//if we get an error during trigger creation
+			//(which is allowed, since we don't have logic
+			//to check with IF EXISTS on every database)
+			//the EntityManager can be in a RollbackOnly state
+			//which we dont want
+			EntityManagerWrapper em = emf.createEntityManager();
```
gunnarmorling: Why use an EM at all? Seems plain JDBC suffices if all are after is executing raw statements. No entities seem involved here?

genericjpa/db/src/main/java/org/hibernate/search/db/EventType.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.hibernate.search.exception.AssertionFailure;
+
+/**
+ * Contains constants that describe the different Database-Events that are relevant to the index
+ *
+ * @author Martin Braun
+ * @hsearch.experimental
+ */
+public final class EventType {
```
gunnarmorling: How comes that this isn't an enum?

genericjpa/db/src/main/java/org/hibernate/search/db/util/impl/EntityManagerWrapper.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.db.util.impl;
+
+/**
+ * Created by Martin on 11.11.2015.
+ */
+public interface EntityManagerWrapper {
```
gunnarmorling: What's the purpose of all these wrappers?

genericjpa/db/src/main/java/org/hibernate/search/annotations/UpdateInfo.java
```diff
@@ -0,0 +1,50 @@
...
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * @author Martin Braun
+ * @hsearch.experimental
+ */
+@Target({FIELD, METHOD, TYPE})
+@Retention(RUNTIME)
+@Repeatable(UpdateInfos.class)
+public @interface UpdateInfo {
```
gunnarmorling: What's the meaning of this? Btw. I'd prefer another package name than "annotations". I wouldn't package by technical aspects (you don't want a package for all your enums either for instance). Rather something more meaningful like "async" or whatever this _semantically_ represents. Also the split package gives bad vibes for OSGi.

gunnarmorling: It seems all this information could be obtained from the entity itself (id type etc.). It appears to introduce quite some redundancy for the user.

genericjpa/common-apis/src/main/java/org/hibernate/search/genericjpa/transaction/TransactionContext.java
```diff
@@ -0,0 +1,11 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.genericjpa.transaction;
+
+public interface TransactionContext extends org.hibernate.search.backend.TransactionContext {
```
gunnarmorling: What's that needed for?

genericjpa/common-apis/src/main/java/org/hibernate/search/genericjpa/factory/Transaction.java
```diff
@@ -0,0 +1,66 @@
...
+package org.hibernate.search.genericjpa.factory;
+
+import javax.transaction.Status;
+import javax.transaction.Synchronization;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.hibernate.search.genericjpa.transaction.TransactionContext;
+
+public class Transaction implements TransactionContext {
```
gunnarmorling: What is it doing?

s4ke:
@gunnarmorling 

> @s4ke I still see the "genericjpa" module as part of this PR. I assumed this would not be the case anymore as we discussed earlier today only the async backend would be contained?

The classes that are in the *genericjpa* subpackages you mean? I kept those so when we merge the genericjpa version we don't have to move code around again. These are mostly needed in only 1 or 2 places in the async backend, but the genericjpa version needs them all over the place.

s4ke:
@gunnarmorling please take a look at pages 68-75 in here:

https://github.com/s4ke/Bachelor-Thesis/raw/Update-System-rework/tex/src/bachelor_thesis.pdf

This illustrates the general idea and should give an introduction into the most important classes. If there are still questions I can write something up.

genericjpa/common-apis/src/main/java/org/hibernate/search/genericjpa/factory/Transaction.java
s4ke: I just removed it.

genericjpa/common-apis/src/main/java/org/hibernate/search/genericjpa/transaction/TransactionContext.java
s4ke: I just removed it.

genericjpa/db/src/main/java/org/hibernate/search/annotations/UpdateInfo.java
s4ke: I think we had this discussion already and we came to the conclusion that for now the extra annotations can be kept with a @hsearch.experimental and that we will later add a more "intelligent" mechanism that tries to guess more.

genericjpa/db/src/main/java/org/hibernate/search/db/util/impl/EntityManagerWrapper.java
s4ke: These are needed so that I can plug in the generic version seamlessly. Also I didn't want to implement the full JPA contract when using a Hibernate Session (Hibernate Search only uses these internal afaik) when there is no JPA available.

genericjpa/db/src/main/java/org/hibernate/search/db/EventType.java
s4ke: Probably because I can use these easier in SQL generation.

genericjpa/async-updates/src/main/java/org/hibernate/search/db/events/jpa/impl/SQLJPAAsyncUpdateSourceProvider.java
s4ke: I cannot unwrap to a JDBC connection from every JPA provider. (I believe Hibernate is the one that restricts this)

genericjpa/common-apis/src/main/java/org/hibernate/search/genericjpa/metadata/impl/RehashedTypeMetadata.java
s4ke: added one.

genericjpa/db/src/main/java/org/hibernate/search/annotations/UpdateInfo.java
gunnarmorling: Yes, we had it before, and as I said back then, I just think this approach is not really what a user would expect or be willing to do. It's just too much work on their side.

Also it's not clear to me what benefit these annotations provide at all: Their info can be retrieved from the model itself: table name, id type etc. Why asking the user to express this info a second time?

s4ke: > Yes, we had it before, and as I said back then, I just think this approach is not really what a  user would expect or be willing to do. It's just too much work on their side.

I don't see why having a mechanism that covers all use cases is bad as a starting point. You have to remember that this mechanism is intended to be used in the generic version as well. After the general mechanism is merged, I am planning on doing a less "hands-on" approach for the users.

> Also it's not clear to me what benefit these annotations provide at all: Their info can be retrieved from the model itself: table name, id type etc. Why asking the user to express this info a second time?

Again, this is intended to be used by both the normal Hibernate Search and the generic version (which I plan on merging after this backend). The normal JPA annotations don't provide enough information so I decided on having everything that is Update mechanism related in the same spot to not confuse people.

s4ke: @gunnarmorling 

I think I have an alternative that can make us both happy:

For the sake of _this_ pullrequest we keep the extra annotations for Hibernate Search when used with Hibernate ORM. I found out that it should be possible to build a "EventModelParser" (that's what I named the interface that gathers the information about the tables) that uses information from Hibernate ORM's persisting system to construct the needed meta-objects (and falls back to the annotations if needed?).

For me this is actually quite hard as I don't know Hibernate ORM's code-base well at all and I will need some help with that.

/CC @Sanne

s4ke: To make it clear:

The plan is to keep the @UpdateInfo annotations and build a separate system that is used in conjunction with Hibernate ORM.

gunnarmorling: Seems like we are going in circles.

As I have said several times now, I don't believe this approach is in the best interest of users. It's @Sanne's call to decide eventually, but personally I am not convinced that this is the right path forwarded.

Users will find it terrible to be forced to specify all that meta-data again, although you can determine it from the JPA provider's meta-model. Don't think so much about shortcuts to make implementation simpler but more what actually is helpful to a user.

You can obtain some information through the JPA meta-model (`EntityManager#getMetamodel()`), e.g. on id types. For other things (table name etc.) you need to go provider-specific APIs. I don't think it'll be super-complicated; For Hibernate ORM we could give you the right starting point, for others you'd have to dig a bit. But essentially that's what I think should be done here.

Personally, I'd push back the thought of generification for now and rather focus on delivering a great async backend for Hibernate ORM. Then the solution can be grown from there, but I wouldn't start with a less than optimal API as that's what drives users away and you'll have a hard time to make them come back.

s4ke: @gunnarmorling:

> You can obtain some information through the JPA meta-model (EntityManager#getMetamodel()), e.g. on id types. For other things (table name etc.) you  need to go provider-specific APIs. I don't think it'll be super-complicated; For Hibernate ORM we could give you the right starting point, for others you'd have to dig a bit. But essentially that's what I think should be done here. 

I already looked into that and I think you underestimate how much work this is (apart from Hibernate ORM).

> Personally, I'd push back the thought of generification for now and rather focus on delivering > a great async backend for Hibernate ORM. Then the solution can be grown from there, but I > wouldn't start with a less than optimal API as that's what drives users away and you'll have a > hard time to make them come back.

That's more or less what I intended to say. Hibernate Search (with ORM) gets its own approach, while the generic version stays back. I more or less intended this PR to be a starter to get things going.

s4ke: Additionally the JPA metamodel only gives information about the Java side of things. No notion of where the information is stored on the database level.

With using the annotations as a fallback (NOT with Hibernate ORM, but as I said for the sake of this PR I want to leave it as is and enlist the work of someone that knows the Hibernate Persister better than me for a follow up issue) we make sure every JPA provider is supported. Even small ones like BatooJPA (which I do not intend writing a custom integration for). For the big JPA implementations I intend to have a custom solution that works with the least amount of extra work possible.

s4ke: Also just as a side-note. These annotations can be of use if the user wants to override in which table the contents get stored. But I repeat, I want to have a mechanism that is as automatic as possible for Hibernate Search with Hibernate ORM (but for that I need help).

s4ke: My plan is to have the annotations only for certain edge cases in the Hibernate Search + Hibernate ORM case.

s4ke:
@gunnarmorling I think this is ready to be reviewed again. Want me to open a fresh pull request?

gunnarmorling:
As you prefer; For me, sticking to the current one is fine.

s4ke:
@gunnarmorling Okay, then I guess we can keep the discussion here

gunnarmorling:
There are several binary files (an index dir). Can you remove them from this PR?

genericjpa/common-apis/src/main/java/org/hibernate/search/genericjpa/annotations/InIndex.java
```diff
@@ -0,0 +1,27 @@
...
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * This annotation is needed to specify that a class is located in an index. <br>
+ * <br>
+ * This is only needed (but always allowed), if objects of subclasses are supplied to indexing related operations. Then
+ * the "least top level" class that is annotated with @InIndex is used. operations
```
gunnarmorling: "operations"?

It's still not clear to me why this is needed in addition to `@Indexed`?

genericjpa/db/src/main/java/org/hibernate/search/db/EventType.java
gunnarmorling: Ok, what I really meant to say is: Please make it an enum ;) You can have static and instance methods for these, too.

genericjpa/common-apis/src/main/java/org/hibernate/search/genericjpa/metadata/impl/RehashedTypeMetadata.java
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * This class represents a "rehashed" version of the
+ * TypeMetadata the engine gives us. These are mostly needed
+ * for the async backend to know where to find the id fields
+ * (on the class and in the index so the index can be held up-to-date
+ * properly) and the types of the fields
+ *
+ * @author Martin Braun
+ */
+public final class RehashedTypeMetadata {
```
gunnarmorling: Is this shared between threads? If so, it should be immutable preferably. Btw. I found the "rehashed" name confusing, maybe "Enriched..." or "Extended..."? It's not just representing `TypeMetadata` in a different form but adds more metadata to it.

genericjpa/common-apis/src/main/java/org/hibernate/search/genericjpa/metadata/impl/MetadataUtil.java
```diff
@@ -0,0 +1,138 @@
...
+import org.hibernate.search.exception.ErrorHandler;
+import org.hibernate.search.exception.impl.LogErrorHandler;
+import org.hibernate.search.indexes.impl.IndexManagerHolder;
+import org.hibernate.search.spi.BuildContext;
+import org.hibernate.search.spi.IndexingMode;
+
+/**
+ * @author Martin Braun
+ */
+public class MetadataUtil {
```
gunnarmorling: What's it doing? IIRC, we discussed before that there should be a type-level comment for each class.

genericjpa/db/src/main/java/org/hibernate/search/db/util/impl/EntityManagerFactoryWrapper.java
```diff
@@ -0,0 +1,25 @@
...
+ * This interface is used so we can have an abstraction from the actual ORM used for the async backend.
+ * This is needed because the async backend is intended to be usable with other JPA providers than
+ * Hibernate ORM as well. We also cannot be sure whether Hibernate ORM is used in conjunction with JPA
+ * so we cannot rely on these interfaces, either.
+ *
+ * Only methods that are required for the async backend can be found in this class
+ *
+ * @author Martin Braun
+ */
+public interface EntityManagerFactoryWrapper {
```
gunnarmorling: Hum, so this still contains code for abstracting from JPA providers; Didn't you want to get rid of this in this first PR, just focusing on the async backend for Hibernate ORM?

pom.xml
```diff
@@ -40,6 +40,14 @@
         <module>integrationtest/osgi/karaf-features</module>
         <module>integrationtest/sandbox</module>
         <module>integrationtest/engine-performance</module>
+
+        <module>genericjpa/util</module>
+        <module>genericjpa/common-apis</module>
+        <module>genericjpa/db</module>
+		<module>genericjpa/async-updates</module>
```
gunnarmorling: The indentation looks odd (also below).

genericjpa/common-apis/src/main/java/org/hibernate/search/genericjpa/metadata/impl/RehashedTypeMetadata.java
s4ke: I can make it immutable. Good idea on the name.

genericjpa/common-apis/src/main/java/org/hibernate/search/genericjpa/metadata/impl/MetadataUtil.java
s4ke: Seems like I missed this one (and maybe some others), will add more comments this weekend.

orm/src/main/java/org/hibernate/search/hcore/impl/HibernateSearchSessionFactoryObserver.java
```diff
@@ -71,7 +76,14 @@ public void sessionFactoryCreated(SessionFactory factory) {
 			HibernateSessionFactoryService sessionService = new DefaultHibernateSessionFactoryService( factory );
 			if ( extendedIntegrator == null ) {
 				SearchIntegrator searchIntegrator = new SearchIntegratorBuilder()
-						.configuration( new SearchConfigurationFromHibernateCore( metadata, configurationService, classLoaderService, sessionService ) )
+						.configuration(
```
gunnarmorling: Hum, seems there are many formatting changes in this file that can be reverted?

orm/src/main/java/org/hibernate/search/hcore/impl/HibernateSearchSessionFactoryObserver.java
```diff
@@ -93,8 +110,40 @@ public void sessionFactoryCreated(SessionFactory factory) {
 		}
 	}
 
+	private void startTriggerAsyncBackendService(SessionFactory factory) {
```
gunnarmorling: Is there a strong need for starting the async backend from within the ORM module? Could there not be a specific observer in the async backend module?

genericjpa/common-apis/src/main/java/org/hibernate/search/genericjpa/annotations/InIndex.java
s4ke: This is just in here because it is needed in the SubclassSupportInitializer which is in turn used in StandaloneSearchConfiguration. This is a "relic" from genericjpa which I don't really want to remove now because I would have change code now and revert these changes for my genericjpa code. I hope this is okay. (btw all classes which are needed exclusively for genericjpa should be in genericjpa packages).

orm/src/main/java/org/hibernate/search/backend/triggers/impl/TriggerAsyncBackendService.java
```diff
@@ -0,0 +1,34 @@
...
+import org.hibernate.search.engine.integration.impl.ExtendedSearchIntegrator;
+import org.hibernate.search.engine.service.classloading.spi.ClassLoaderService;
+import org.hibernate.search.engine.service.spi.Service;
+
+/**
+ * @author Martin Braun
+ * <p>
+ * Service that controls the trigger based asynchronous index updating mechanism
+ * <p>
+ * Does not implement Startable and Stoppable as its lifecycle needs to be controlled
```
gunnarmorling: Can you expand the reasoning on this a bit?

s4ke:
> There are several binary files (an index dir). Can you remove them from this PR?

Oh, I added those by mistake. Good catch.

genericjpa/db/src/main/java/org/hibernate/search/db/util/impl/EntityManagerFactoryWrapper.java
s4ke: No, I don't want to break too much existing code. I want to keep the abstractions as I don't think they are such a big deal to be honest and don't break my genericjpa code (which I keep rebasing on this code constantly).

pom.xml
s4ke: it looks odd indeed. Will check on my PC tomorrow.

orm/src/main/java/org/hibernate/search/hcore/impl/HibernateSearchSessionFactoryObserver.java
s4ke: I will have to take a look into why I did this the way I did it. Maybe it was related to me not getting a proper reference to the SessionFactory, but I don't know right now.

genericjpa/db/src/main/java/org/hibernate/search/db/util/impl/EntityManagerFactoryWrapper.java
gunnarmorling: > No, I don't want to break too much existing code.

Not sure what you mean?

> I want to keep the abstractions 

As said, this PR should not contain anything related to that generification. Let's focus on getting one thing done (async backend), this will increase chances to have something merge-able at some point. The other stuff just adds complexity not needed for achieving that first goal.

s4ke: I know you would have started this the other way around: First the async backend and then the generification (which would be the right way if I started from scratch and this abstraction wouldn't have made it into this PR).

But the matter of the fact is that the generification code is quite mature now and mostly in place (and I keep that code up-to-date with this PR so its easier to keep track of changes). The average Hibernate ORM + Hibernate Search user doesn't see any of this code and it just helps in the longterm.

s4ke: If I remove the abstraction now, I am 100% guaranteed to undo these changes into the state they are now. This plus the fact that I cannot keep rebasing my current genericjpa code on this PR makes this a no-go for me.

orm/src/main/java/org/hibernate/search/hcore/impl/HibernateSearchSessionFactoryObserver.java
s4ke: Good catch.

s4ke:
All major concerns should have been addressed by now.

s4ke:
Things that still have to be done:

- create-drop for triggers (needs a PR in orm to be merged) 
- with the new persister in the next hibernate orm release, the EventModelParser for ORM needs to be adapted.

s4ke:
Also: the orm EventModelParser should be able to return similar column Definitions for the Update Tables to the ones hibernate orm uses for the original Tables

s4ke:
I think this is currently in a state that can be reviewed quite well. The things that still need some work will have to wait for the next Hibernate ORM release (and the adaption of that in Search).

yrodiere:
Unfortunately, we've taken so long to merge this PR that merging it now, after the extensive changes from Search 6, would require a lot of work. I'm going to close this for now, but I will add a link to this PR from https://hibernate.atlassian.net/browse/HSEARCH-2364, because this work will be very interesting when we try to introduce fully asynchronous indexing in Search 6, even if the technology might be different (Kafka, so that we can eventually use Debezium?).

s4ke:
I am happy to hear that this is being tackled even though the code does not make it to master. At least it will help when exploring what has to be done. Thanks for the udpate :).

PR #981, opened 2016-02-16
HSEARCH-2065 Progress on faceting with Elasticsearch

gsmet:
The main limit atm is that the faceting is done on the indexed fields (see Car.java used by the WebShopTest for an example of what is the current limitation). IMHO, we should build fields specific to the faceting.

We cannot do so for now as the LuceneDocument does not contain enough information to reconstruct the desired values of the facets - especially for numerics as the values are transformed as binary and strings by the Lucene faceting code.

It is linked to a discussion we had with Gunnar about a major overhaul of how the documents are built.

Another point I'd like to mention is that currently HS considers that the document type of a given Java type is fixed. This is not really the case as Dates are stored as Longs with Lucene but as Strings with Elasticsearch.

### DIFF
--- a/elasticsearch/pom.xml
+++ b/elasticsearch/pom.xml
@@ -249,24 +249,17 @@
                                 <exclude>**/ToStringTest.java</exclude>
                                 <exclude>**/DynamicBoostingTest.java</exclude>
                                 <exclude>**/DSLTest.java</exclude>
+                                <exclude>**/DSLEmbeddedSearchTest.java</exclude>
                                 <exclude>**/MoreLikeThisTest.java</exclude>
                                 <exclude>**/NumericEncodingQueriesTest.java</exclude>
                                 <exclude>**/ExplanationTest.java</exclude>
+                                <exclude>**/EmbeddedSortableIdFieldTest.java</exclude>
 
                                 <!-- Needs range queries on nested fields -->
                                 <exclude>**/DslEmbeddedSearchTest.java</exclude>
 
                                 <!-- Faceting -->
-                                <exclude>**/CollectionFacetingTest.java</exclude>
-                                <exclude>**/EdgeCaseFacetTest.java</exclude>
-                                <exclude>**/FacetFilteringTest.java</exclude>
-                                <exclude>**/ManyToOneFacetingTest.java</exclude>
-                                <exclude>**/NoQueryResultsFacetingTest.java</exclude>
-                                <exclude>**/NumberFacetingTest.java</exclude>
-                                <exclude>**/RangeFacetingTest.java</exclude>
-                                <exclude>**/SimpleFacetingTest.java</exclude>
-                                <exclude>**/StringFacetingTest.java</exclude>
-                                <exclude>**/WebShopTest.java</exclude>
+                                <exclude>**/FacetIndexingFailureTest.java</exclude>
 
                                 <exclude>**/IndexAndQueryNullTest.java</exclude>
                                 <exclude>**/ProgrammaticIndexAndQueryNullTest.java</exclude>
--- a/elasticsearch/pom.xml
+++ b/elasticsearch/pom.xml
@@ -259,7 +259,6 @@
                                 <!-- Faceting -->
                                 <exclude>**/FacetIndexingFailureTest.java</exclude>
                                 <exclude>**/FacetUnknownFieldFailureTest.java</exclude>
-                                <exclude>**/WebShopTest.java</exclude>
 
                                 <exclude>**/IndexAndQueryNullTest.java</exclude>
                                 <exclude>**/ProgrammaticIndexAndQueryNullTest.java</exclude>
--- a/elasticsearch/pom.xml
+++ b/elasticsearch/pom.xml
@@ -260,6 +260,8 @@
 
                                 <!-- Faceting -->
                                 <exclude>**/FacetIndexingFailureTest.java</exclude>
+                                <exclude>**/FacetUnknownFieldFailureTest.java</exclude>
+                                <exclude>**/WebShopTest.java</exclude>
 
                                 <exclude>**/IndexAndQueryNullTest.java</exclude>
                                 <exclude>**/ProgrammaticIndexAndQueryNullTest.java</exclude>
--- a/elasticsearch/pom.xml
+++ b/elasticsearch/pom.xml
@@ -249,11 +249,9 @@
                                 <exclude>**/ToStringTest.java</exclude>
                                 <exclude>**/DynamicBoostingTest.java</exclude>
                                 <exclude>**/DSLTest.java</exclude>
-                                <exclude>**/DSLEmbeddedSearchTest.java</exclude>
                                 <exclude>**/MoreLikeThisTest.java</exclude>
                                 <exclude>**/NumericEncodingQueriesTest.java</exclude>
                                 <exclude>**/ExplanationTest.java</exclude>
-                                <exclude>**/EmbeddedSortableIdFieldTest.java</exclude>
 
                                 <!-- Needs range queries on nested fields -->
                                 <exclude>**/DslEmbeddedSearchTest.java</exclude>
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/ElasticsearchQueries.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/ElasticsearchQueries.java
@@ -35,7 +35,7 @@ public class ElasticsearchQueries {
 	}
 
 	/**
-	 * Creates an ElasticSearch query from the given JSON query representation. See the <a
+	 * Creates an Elasticsearch query from the given JSON query representation. See the <a
 	 * href="https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html">official
 	 * documentation</a> for the complete query syntax.
 	 */
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/ElasticsearchQueries.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/ElasticsearchQueries.java
@@ -7,6 +7,7 @@
 package org.hibernate.search.backend.elasticsearch;
 
 import org.hibernate.search.backend.elasticsearch.impl.ElasticsearchHSQueryImpl;
+import org.hibernate.search.backend.elasticsearch.json.JsonBuilder;
 import org.hibernate.search.engine.integration.impl.ExtendedSearchIntegrator;
 import org.hibernate.search.query.engine.spi.HSQuery;
 import org.hibernate.search.query.engine.spi.QueryDescriptor;
@@ -30,9 +31,15 @@ public class ElasticsearchQueries {
 	 * documentation</a> for the complete query syntax.
 	 */
 	public static QueryDescriptor fromJson(String jsonQuery) {
-		// TODO Parse + Re-render using Gson for now to leverage single quote support
-		jsonQuery = new JsonParser().parse( jsonQuery ).toString();
+		return new ElasticsearchJsonQuery( new JsonParser().parse( jsonQuery ).getAsJsonObject() );
+	}
 
+	/**
+	 * Creates an ElasticSearch query from the given JSON query representation. See the <a
+	 * href="https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html">official
+	 * documentation</a> for the complete query syntax.
+	 */
+	public static QueryDescriptor fromJson(JsonObject jsonQuery) {
 		return new ElasticsearchJsonQuery( jsonQuery );
 	}
 
@@ -46,23 +53,18 @@ public class ElasticsearchQueries {
 		// Payload looks like so:
 		// { "query" : { "query_string" : { "query" : "abstract:Hibernate" } } }
 
-		JsonObject query = new JsonObject();
-		query.addProperty( "query", queryStringQuery );
+		JsonBuilder.Object query = JsonBuilder.object().add( "query",
+				JsonBuilder.object().add( "queryString",
+						JsonBuilder.object().addProperty( "query", queryStringQuery ) ) );
 
-		JsonObject queryString = new JsonObject();
-		queryString.add( "query_string", query );
-
-		JsonObject queryObject = new JsonObject();
-		queryObject.add( "query", queryString );
-
-		return new ElasticsearchJsonQuery( queryObject.toString() );
+		return new ElasticsearchJsonQuery( query.build() );
 	}
 
 	private static class ElasticsearchJsonQuery implements QueryDescriptor {
 
-		private final String jsonQuery;
+		private final JsonObject jsonQuery;
 
-		public ElasticsearchJsonQuery(String jsonQuery) {
+		public ElasticsearchJsonQuery(JsonObject jsonQuery) {
 			this.jsonQuery = jsonQuery;
 		}
 
@@ -73,7 +75,7 @@ public class ElasticsearchQueries {
 
 		@Override
 		public String toString() {
-			return jsonQuery;
+			return jsonQuery.toString();
 		}
 	}
 }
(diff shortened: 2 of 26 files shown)


elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/ElasticsearchQueries.java
```diff
@@ -46,23 +53,18 @@ public static QueryDescriptor fromQueryString(String queryStringQuery) {
...
+		return new ElasticsearchJsonQuery( query.build() );
 	}
 
 	private static class ElasticsearchJsonQuery implements QueryDescriptor {
 
-		private final String jsonQuery;
+		private final JsonObject jsonQuery;
 
-		public ElasticsearchJsonQuery(String jsonQuery) {
+		public ElasticsearchJsonQuery(JsonObject jsonQuery) {
```
gunnarmorling: Hi, as we discussed I'd prefer to not expose the usage of Gson on the public API at this point. This should be an implementation detail only.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchHSQueryImpl.java
```diff
@@ -553,6 +586,78 @@ private JsonElement getFieldValue(JsonObject parent, String projectedField) {
...
+		return facets;
+	}
+
+	private List<Facet> updateStringFacets(JsonObject aggregations, DiscreteFacetRequest facetRequest) {
+		JsonElement aggregation = aggregations.get( facetRequest.getFacetingName() );
+		if ( aggregation == null ) {
+			return Collections.emptyList();
+		}
+		// deal with nested aggregation for nested documents
+		if ( facetRequest.getFieldName().contains( "." ) ) {
```
gunnarmorling: The "." could be configured to something else by the user as I learned. This should be resolved from the meta-data.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/json/JsonBuilder.java
```diff
@@ -0,0 +1,162 @@
...
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+
+/**
+ * Builder used to make the Gson API chainable.
+ *
+ * @author Guillaume Smet
+ */
+public class JsonBuilder {
```
gunnarmorling: This should not be public.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/json/JsonBuilder.java
```diff
@@ -0,0 +1,162 @@
...
+			}
+			else if ( value instanceof Boolean ) {
+				jsonObject.addProperty( property, (Boolean) value );
+			}
+			else if ( value instanceof Character ) {
+				jsonObject.addProperty( property, (Character) value );
+			}
+			else if (value instanceof Date) {
+				// we use this Gson formatter to be sure the date format is in sync with the JestClient configuration
+				jsonObject.addProperty( property, JestClient.GSON.toJson( value ).replace( "\"", "" ) );
```
gunnarmorling: The reference to `JestClient` here seems odd. Should the GSON object be extracted to a separate holder class or service?

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/json/ToElasticsearch.java
```diff
@@ -0,0 +1,287 @@
...
+import com.google.gson.JsonArray;
+import com.google.gson.JsonObject;
+
+/**
+ * Various utilities to transform Hibernate Search API into Elasticsearch JSON.
+ *
+ * @author Guillaume Smet
+ * @author Gunnar Morling
+ */
+public class ToElasticsearch {
```
gunnarmorling: This should not be public API AFAICS:

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/json/ToElasticsearch.java
```diff
@@ -0,0 +1,287 @@
...
+				sort.addProperty( "_term", "asc" );
+				break;
+			case RANGE_DEFINITION_ORDER:
+				throw LOG.cannotSendRangeDefinitionOrderToElasticsearchBackend();
+		}
+		return sort;
+	}
+
+	public static JsonObject condition(String operator, JsonArray conditions) {
+		JsonObject jsonCondition = new JsonObject();
```
gunnarmorling: That instantiation seems unneeded given the subsequent assignments.

gunnarmorling:
Nice stuff, @gsmet! My understanding of the entire faceting code is a bit limited, so maybe you could provide some more details on the following:

> The main limit atm is that the faceting is done on the indexed fields (see Car.java used by the WebShopTest for an example of what is the current limitation). IMHO, we should build fields specific to the faceting.
>
> We cannot do so for now as the LuceneDocument does not contain enough information to reconstruct the desired values of the facets - especially for numerics as the values are transformed as binary and strings by the Lucene faceting code.

What exactly are the limitations in the car example? What fields do we build and which ones would we need instead?

> It is linked to a discussion we had with Gunnar about a major overhaul of how the documents are built.

Indeed, the current retrieval from Lucene `Document` structures is just a quick hack for getting started. I'd only like to better understand the limitations in the faceting case.

> Another point I'd like to mention is that currently HS considers that the document type of a given Java type is fixed. This is not really the case as Dates are stored as Longs with Lucene but as Strings with Elasticsearch.

Where/when is that exactly a problem? It's still "fixed" for a given backend/configuration. Are there assumptions on dates always being stored as date in core?

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/json/ToElasticsearch.java
```diff
@@ -0,0 +1,287 @@
...
+		String path = getFieldNestedPath( field );
+
+		return JsonBuilder.object().add( "nested",
+				JsonBuilder.object()
+						.addProperty( "path", path )
+						.add( "query", query ) )
+				.build();
+	}
+
+	public static JsonObject fromLuceneFilter(Filter luceneFilter) {
```
gunnarmorling: Ah, nice. So with that, `Query#filter()` should be implementable?

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/ElasticsearchQueries.java
gsmet: This particular class is private (but I should have made its constructor private too) but I suppose what bugs you is the fromJson(JsonObject) method just above.

It's only used in ElasticsearchLuceneQueryTranslator atm.

Maybe I should move JsonBuilder and ToElasticsearch to the impl package? This way, I could make the fromJson(JsonObject) method package protected.

WDYT?

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchHSQueryImpl.java
gsmet: Atm there are a lot of other assumptions about the fact that a "." in the field name means an embedded field. See FieldHelper for instance or ToElasticsearch.

I'm not sure we will be able to use nested documents if we don't have an easy way to know if the field is nested or not.

gunnarmorling: > Atm there are a lot of other assumptions about the fact that a "." in the
field name means an embedded field

Yes, I know. I had started with that pattern to keep it simple, but also
not really considering that users could choose something different. Sanne
reminded me of it currently, which is why I am bringing it up. I don't mind
going with that for now, but let's not forget about it.

gsmet:
> What exactly are the limitations in the car example? What fields do we build and which ones would we need instead?

See this facet definition:
```java
	@Field(name = "cubicCapacity", analyze = Analyze.NO, bridge = @FieldBridge(impl = IntegerBridge.class))
	@Facets({
			@Facet(name = "cubicCapacity", forField = "cubicCapacity", encoding = FacetEncodingType.STRING),
			@Facet(name = "cubicCapacity_Numeric", forField = "cubicCapacity", encoding = FacetEncodingType.LONG)
	})
	private int cubicCapacity;
```
With the current code, the faceting is based on the field and only the field. So you can facet on cubicCapacity, except it's a numeric instead of a string.

The test currently failing is the one using cubicCapacity_Numeric as we don't have a field with this name in the Elasticsearch index. I thought about trying to rebuild it from the Lucene Document but it's not that easy as the Lucene facet code transforms the facet fields to something barely usable and you can't use the values indexed in the field itself as you might have a fieldBridge transforming the values.

> Where/when is that exactly a problem? It's still "fixed" for a given backend/configuration. Are there assumptions on dates always being stored as longs in core?

See AnnotationMetadataProvider#determineFacetEncodingType :
```java
		else if ( Date.class.isAssignableFrom( indexedType ) || Calendar.class.isAssignableFrom( indexedType ) ) {
			facetEncodingType = FacetEncodingType.LONG;
		}
```

I think the only problem with it is in DocumentBuilderIndexedEntity#addFacetDocValues where I added the workaround for now.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/json/ToElasticsearch.java
gsmet: Haven't tested it yet but I think so. At least for the filters built from a query.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchHSQueryImpl.java
gsmet: Yes, my point was just that I don't think it's something we should fix in this PR.

I can take a shot at it if I'm not stepping on your toes?

Created https://hibernate.atlassian.net/browse/HSEARCH-2097 so that we don't forget about it.

gunnarmorling: Thanks for opening this issue! I am working on stuff related to embedded
mapping, so I suppose it'd be better to wait with addressing HSEARCH-2097
for a bit.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/json/JsonBuilder.java
gsmet: Maybe it would make more sense to move the GSON object to JsonBuilder and use it in JestClient?

gunnarmorling: Does JestClient otherwise need to deal with JsonBuilder? If not I'd push it
to a separate holder class.

gunnarmorling:
> The test currently failing is the one using cubicCapacity_Numeric as we don't have a field with this name in the Elasticsearch index.

Ok, I see. Thanks for the thorough explanation, that's much appreciated! I need to think about this a bit.

gunnarmorling:
> The test currently failing is the one using cubicCapacity_Numeric as we don't have a field with this name in the Elasticsearch index.

Couldn't we just add this field? It can be retrieved in `ElasticsearchIndexWorkVisitor.convertToJson(Document, Class<?>)`, it's just that we exclude all doc value fields (it escapes me right now why). Maybe you can try and change that?

gsmet:
Yeah, well, that's what I thought and then I opened `FacetHandling`, I saw return `facetConfig.build( doc )` and I discovered that Lucene is rewriting all the faceted fields in a different way (see `FacetsConfig.build(TaxonomyWriter taxoWriter, Document doc))` :).

I don't think we can really exploit the result of the Lucene mangling. You end up with binary and strings.

gunnarmorling:
Hum, ok. I can get the numeric value for cubic capacity from the facet doc
value field, but I suppose it may not be working for other types.

gsmet:
From the field returned by Lucene after the FacetsConfig.build() call? Where did you find a numeric value?

gsmet:
We also need a way to make the difference between a numeric (LONG or DOUBLE) facet and a string facet.

gunnarmorling:
In ElasticsearchIndexWorkVisitor#convertToJson() I am seeing a
NumericDocValuesField whose value I can get via numericValue().

gsmet:
OK, I'll take another look next week!

gsmet:
@gunnarmorling done. The difficulty was for the String facets but I think I got it. All the faceting tests now runs, except the ones using a non existing field.

elasticsearch/pom.xml
```diff
@@ -257,16 +257,8 @@
...
-                                <exclude>**/CollectionFacetingTest.java</exclude>
-                                <exclude>**/EdgeCaseFacetTest.java</exclude>
-                                <exclude>**/FacetFilteringTest.java</exclude>
-                                <exclude>**/ManyToOneFacetingTest.java</exclude>
-                                <exclude>**/NoQueryResultsFacetingTest.java</exclude>
-                                <exclude>**/NumberFacetingTest.java</exclude>
-                                <exclude>**/RangeFacetingTest.java</exclude>
-                                <exclude>**/SimpleFacetingTest.java</exclude>
-                                <exclude>**/StringFacetingTest.java</exclude>
-                                <exclude>**/WebShopTest.java</exclude>
```
gunnarmorling: Nice progress :)

engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
```diff
@@ -636,20 +636,26 @@ private void addFacetDocValues(Document document,
...
+						facetField = new SortedSetDocValuesFacetField( facetMetadata.getFacetName(),
+								( (TwoWayFieldBridge) fieldBridge ).objectToString( value ) );
+					}
+					else {
 						log.numericDateFacetForNonNumericField(
 								facetMetadata.getFacetName(),
-								fieldMetadata.getFieldName()
-						);
+								fieldMetadata.getFieldName() );
+						throw new AssertionFailure( "Invalid FieldBridge for date faceting: " + fieldBridge );
```
gunnarmorling: I believe it was just forgotten to actually raise the exception returned by the previously invoked log method. I'll change that.

engine/src/main/java/org/hibernate/search/query/dsl/impl/FacetRange.java
```diff
@@ -87,6 +87,20 @@ private String buildRangeString() {
 	}
 
 	@Override
+	public int hashCode() {
```
gunnarmorling: Overriding hashCode() but not equals()? That seems not right. Is there a need for this (i.e. is it used in sets or maps as key or so)?

gunnarmorling:
Hey @gsmet, Looks good to me apart from that hashCode()/equals() anomaly. Once that's fixed, I'll merge it. Thanks a lot, great to see this one tackled!

engine/src/main/java/org/hibernate/search/query/dsl/impl/FacetRange.java
gsmet: No, it's not really. I only use it to have a string representation of the FacetRange. I'll go change it to something more understandable.

I also will fix the facetComparators duplicated code.

gunnarmorling: Ok, cool. Just let me know then when you're done. Any other loose ends from your PoV related to faceting (apart from the overall re-design we discussed ;)?

gsmet:
Hi @gunnarmorling . We should be good to go now. I fixed the last infelicities.

gunnarmorling:
Rebased and applied; Many thanks, @gsmet!

Sanne:
Great, I'll merge the other one as soon as you rebase it.

gunnarmorling:
Just rebased and force-pushed the feature branch.

PR #982, opened 2016-02-16
Add Travis support

gsmet:
Hi!

As mentioned on the list, I think we could give Travis a try for building the PRs.

Note that what is interesting with Travis is that it can also run for our individual forks so we can get our branches tested by the CI before creating the PR.

The .travis.yml build the tests for 3 different RDBMS, maybe we could just build it for PostgreSQL if we only use Travis for the PR tests?

Thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,37 @@
+language: java
+jdk:
+  - oraclejdk8
+addons:
+  postgresql: '9.4'
+  mariadb: '10.1'
+  apt:
+    packages:
+      - oracle-java8-installer
+# might be useful to push reports to an S3 bucket
+#  artifacts:
+#    paths:
+#      - $(find $HOME -name surefire-reports | tr "\n" ":")
+#    debug: true
+#  s3_region: 'us-west-2'
+cache:
+  directories:
+    - $HOME/.m2
+env:
+  - DB=h2
+# for now, let's test only with H2
+#  - DB=pgsql
+#  - DB=mariadb
+install:
+  - mvn -v
+  # first run to download all the Maven dependencies without logging
+  - mvn -s settings-example.xml -B -q clean compile -DskipTests=true
+before_script: 
+  - if [[ "$DB" == "pgsql" ]]; then psql -U postgres -f 'travis/pgsql.sql';fi 
+  - if [[ "$DB" == "mariadb" ]]; then mysql -u root < 'travis/mariadb.sql'; fi
+script:
+  # we run checkstyle first to fail fast if there is a styling error
+  - mvn -s settings-example.xml checkstyle:check
+  - if [[ "$DB" == "h2" ]]; then mvn -s settings-example.xml -Ph2 verify;fi 
+  - if [[ "$DB" == "pgsql" ]]; then mvn -s settings-example.xml -Pci-postgresql verify;fi 
+  - if [[ "$DB" == "mariadb" ]]; then mvn -s settings-example.xml -Pci-mariadb verify; fi
+  
--- a/backends/jgroups/src/test/java/org/hibernate/search/backend/jgroups/impl/SyncJGroupsBackendTest.java
+++ b/backends/jgroups/src/test/java/org/hibernate/search/backend/jgroups/impl/SyncJGroupsBackendTest.java
@@ -43,12 +43,13 @@ public class SyncJGroupsBackendTest {
 
 	private static final Log log = LoggerFactory.make();
 	private static final String JGROUPS_CONFIGURATION = "testing-flush-loopback.xml";
+	private static final int JGROUPS_MESSAGES_TIMEOUT = 1000;
 
 	@Rule
 	public SearchFactoryHolder slaveNode = new SearchFactoryHolder( Dvd.class, Book.class, Drink.class, Star.class )
 		.withProperty( "hibernate.search.default.worker.backend", "jgroupsSlave" )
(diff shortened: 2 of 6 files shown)


gsmet:
Note that we can commit the .travis.yml file without enabling Travis on the main project. But contributors could easily enable CI for their forks!

gsmet:
@Sanne I thought about https://github.com/hibernate/hibernate-search/pull/1008 and I think it would be cool to add a checkstyle:check at the beginning of the Travis build to fail fast if something is wrong with checkstyle.

I was thinking about adding that, only test for h2 for now and merge this PR. Anything against this plan?

Sanne:
I think we could merge this; I'm not sure yet if I'll like Travis but these changes look harmless so why not..

Could you create a JIRA for this and refer to it in the commit log?

Sanne:
> I thought about #1008 and I think it would be cool to add a checkstyle:check at the beginning of the Travis build to fail fast if something is wrong with checkstyle.

I think after our checkstyle fixes from today this should not be necessary anymore, right?
AFAIK we run the check as part of each build now.

gsmet:
What I like with doing a checkstyle:check before is that we fail fast in case of style errors as we check all the code source before running the tests.

Sanne:
ok.

.travis.yml
```diff
@@ -0,0 +1,32 @@
...
+  - DB=pgsql
+  - DB=mariadb
+install:
+  - mvn -v
+  - mvn -s settings-example.xml -B -q clean compile -DskipTests=true
+before_script: 
+  - if [[ "$DB" == "pgsql" ]]; then psql -U postgres -f 'travis/pgsql.sql';fi 
+  - if [[ "$DB" == "mariadb" ]]; then mysql -u root < 'travis/mariadb.sql'; fi
+script:
+  - if [[ "$DB" == "h2" ]]; then mvn -s settings-example.xml -Ph2 test;fi
```
gunnarmorling: How comes you do the tests after compile? Seems btw. it'd miss the integration tests that way?

backends/jgroups/src/test/java/org/hibernate/search/backend/jgroups/impl/SyncJGroupsBackendTest.java
```diff
@@ -114,10 +114,15 @@ public void testSynchAsConfigured() {
 		}
 		catch (SearchException se) {
 			//Expected: we're inducing the RPC into NPE
-			Throwable cause = se.getCause().getCause().getCause();
-			Assert.assertTrue( "Cause was not a NullPointerException but a " + cause, cause instanceof NullPointerException );
-			Assert.assertEquals( "Simulated Failure", cause.getMessage() );
-			npeTriggered = true;
+			try {
```
gunnarmorling: How is this related to the Travis set-up?

gunnarmorling:
+1 for merging and getting it off our plates :)

backends/jgroups/src/test/java/org/hibernate/search/backend/jgroups/impl/SyncJGroupsBackendTest.java
gsmet: Well, it's related because I needed the exception to understand what was going on and fix it :)

Sanne: Yea, I know were you're coming from with this fix.. but I think that @gunnarmorling was suggesting that such improvements should live in a separate commit at least, sometimes in their own JIRA.

For example in this case I don't think this is the best improvement for that NPE but asking to improve on that would delay integration of the other improvements in this PR too.

gsmet:
OK, so I fixed everything you reported and added comments to the .travis.yml file.

The bad news is that the elasticsearch integration tests don't seem to be stable enough. I had SocketTimeoutExceptions.

I committed higher timeouts to see how it goes.

Sanne:
> I committed higher timeouts to see how it goes.

What do you prefer,shall we merge this or are you going to try the new timeouts out?

gsmet:
It's better with the higher timeout but I still have 2 tests failing:
- one because the test run before the cluster was green: so we really need what you talked about in your last email;
- as for the other, it looks like a real failure but it might be a timing problem too:
```
testScenario(org.hibernate.search.test.engine.optimizations.CollectionUpdateEventsSecondTest)  Time elapsed: 0.241 sec  <<< FAILURE!
java.lang.AssertionError: expected:<0> but was:<1>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:743)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:555)
	at org.junit.Assert.assertEquals(Assert.java:542)
	at org.hibernate.search.test.engine.optimizations.CollectionUpdateEventsSecondTest.assertOperationsPerformed(CollectionUpdateEventsSecondTest.java:102)
```

Sanne:
I rebased and merged the first 3 commits as they were nice and independent;

The 4th commit "Try to use a higher timeout for Elasticsearch" sounds like an experiment and doesn't have a JIRA yet.
Let me know what you think we should do with that commit? Or shall we close this PR?

gsmet:
Gunnar sets the timeout higher in https://github.com/hibernate/hibernate-search/pull/1023 so I'm closing this one for now.

PR #997, opened 2016-02-17
Fix FieldType javadoc: dates are now stored as long in Lucene

gsmet:
We might have to change the comments with the Elasticsearch integration
but we might as well have them accurate for the time being.


Sanne:
+1 merged
Thanks!

PR #998, opened 2016-02-21
HSEARCH-2103 Fix Checkstyle for Java 8 syntax	

s4ke:
This pr bumps the maven-checkstyle-plugin to 2.17 and puppycrawl checkstyle to 6.15.

### DIFF
--- a/build-config/pom.xml
+++ b/build-config/pom.xml
@@ -37,7 +37,7 @@
         <dependency>
             <groupId>com.puppycrawl.tools</groupId>
             <artifactId>checkstyle</artifactId>
-            <version>6.11.2</version>
+            <version>${puppycrawl.checkstyle.version}</version>
             <exclusions>
                 <!-- Excluding com.sun:tools since it causes errors when importing this module into an IDE on Windows (HSEARCH-1374) -->
                 <exclusion>
--- a/build-config/pom.xml
+++ b/build-config/pom.xml
@@ -5,7 +5,8 @@
  ~ License: GNU Lesser General Public License (LGPL), version 2.1 or later
  ~ See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
  -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.hibernate</groupId>
@@ -36,7 +37,7 @@
         <dependency>
             <groupId>com.puppycrawl.tools</groupId>
             <artifactId>checkstyle</artifactId>
-            <version>5.7</version>
+            <version>6.11.2</version>
             <exclusions>
                 <!-- Excluding com.sun:tools since it causes errors when importing this module into an IDE on Windows (HSEARCH-1374) -->
                 <exclusion>
--- a/build-config/src/main/resources/checkstyle.xml
+++ b/build-config/src/main/resources/checkstyle.xml
@@ -100,8 +100,11 @@
         <module name="EqualsAvoidNull" />
 
         <!-- <module name="DeclarationOrder" /> -->
-        <module name="RedundantModifier"/>
-        <module name="MissingOverride"/>
+        <module name="RedundantModifier">
+            <property name="tokens"
+                      value="METHOD_DEF, VARIABLE_DEF, ANNOTATION_FIELD_DEF, INTERFACE_DEF, CLASS_DEF, ENUM_DEF"/>
+        </module>
+        <module name="MissingOverride" />
 
         <module name="MethodParamPad">
             <property name="option" value="nospace" />
@@ -116,9 +119,12 @@
             <property name="option" value="alone" />
         </module>
 
+        <!-- METHOD_CALL breaks inline casts
+            (i.e. Iterable<?> collection = ((Map<?,?>) value).values();)
+            in 6.15-->
         <module name="ParenPad">
-            <property name="tokens" value="CTOR_CALL, METHOD_CALL, SUPER_CTOR_CALL" />
-            <property name="option" value="space" />
+            <property name="tokens" value="CTOR_CALL, SUPER_CTOR_CALL"/>
+            <property name="option" value="space"/>
         </module>
 
         <module name="org.hibernate.checkstyle.checks.regexp.DoubleSpaces">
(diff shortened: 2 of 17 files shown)


s4ke:
Seems like jenkins is broken for this PR

gunnarmorling:
Looks good to me so far, but what you said in the isssue itself should be discussed:

> I am using Java 8 for some classes in my async backend pull request 

Did we already agree on requiring Java 8?

It'd be fine for me, provided we actually have a good benefit from it, making potential exclusion of users stick to older versions worthwile. If it's just to benefit from some syntactical sugar, I'd rather stick to 7.

s4ke:
@gunnarmorling :

Hmm, I think I use some Java 8 classes, we can discuss whether that's a good idea on that PR I guess. Java 8 is also only required in my new modules.

Anyhow, the Checkstyle upgrade brings some improvements in code-style checking as well, which I think is great.

Sanne:
Let's keep in mind that many getters were written in the "one line" style intentionally,
I'm not a big fan of enforcing otherwise.

I generally like Checkstyle as having some basic ground rules allows us to
 - quickly understand larger blocks of code
 - avoid people introducing many style related changes in PR

But we've also seen that making rules too strict backfires: a pedantic build tool is quite annoying. We need to find a good balance, and I'm actually having the impression that we should be relaxing some rules rather than introducing more checks?

s4ke:
I did not include any new style checks, the default behaviour just changed and I figured since the 1 line getters were only present in tests that this would not be a big deal to keep.

s4ke:
@Sanne: i reverted the 1 line property access checks.

Sanne:
>  i reverted the 1 line property access checks.

Thanks!

>Also, I think that the checkstyle.xml needs some more intensive care in general and needs to be updated to our needs. Why not keep a list of suggestions open?

I'm not sure I understand. What do you have in mind?

BTW the long term plan was to see how this checkstyle.xml and the one living in Hibernate OGM would "stabilize" and mature, and hopefully also converge to extract them in their own repository to be used across Hibernate projects but maintained "out of the way".

i.e. There's no benefit for Search to rebuild that module every time it's being built.

s4ke:
It seems like the checkstyle documentation lacks at some points and that the actual code behaves quite differently. Therefore the settings used in checkstyle.xml might not be optimal 100% at all times and I would propose that we keep enhancing it more in the future.

For this some kind of TODO list would be a good idea where people can list their complaints with the current setting.

Sanne:
> For this some kind of TODO list would be a good idea where people can list their complaints with the current setting.

Sure. I'm not aware of any though; maybe you have some as Windows user?

s4ke:
> Sure. I'm not aware of any though; maybe you have some as Windows user?

Not related to Windows, but as I already said I don't think that checking whether an empty line contains whitespace is beneficial to the build process. As I cannot just auto-format all the code this is quite the issue for me sometimes. I remember some spending my fair share deleting whitespace where it didn't help the readability.

Sanne:
consistency with whitespaces are actually important to avoid conflicts with git, and warnings with git when it's at EOF. But it's the conflicts part which is more important.

Any sensible IDE and editor has options around how to deal with whitespaces, so being strict on that isn't a significant coding burden but rather to check that you've configured your editors correctly.

s4ke:
Hmm, sounds reasonable?

I am using the hibernate-style.xml for IntelliJ. Isn't that enough configuration to prevent that?

Sanne:
> I am using the hibernate-style.xml for IntelliJ. Isn't that enough configuration to prevent that?

I'm not sure. I would assume the IntelliJ styles are not aligned with our checkstyle rules, but my guess is that the IntelliJ styles need more work rather than the checkstyle files.
N.B. those IntelliJ styles are mostly used by Steve Ebersole on Hibernate ORM, while most people use Eclipse when working on Hibernate Search.

s4ke:
If I use autoformat with the IntelliJ xml file everything works great. But sometimes during normal editing it seems to slip up. I will keep watching out for these issues, maybe they have gone away in the meantime (or at least the big ones).

Sanne:
> But sometimes during normal editing it seems to slip up. I will keep watching out for these issues, maybe they have gone away in the meantime (or at least the big ones).

I am not sure what that is, since I mostly use Eclipse. My guess is that beyond "formatting rules" it might have separate options about adding default whitepace for indentation.

Sanne:
Hi @s4ke 
thanks for this, there are some nice improvements in here but I'd prefer you could do one more polishing step: I know git doesn't scare you (as it does to some occasional contributors) so I'd love it if you could do one more step when preparing pull request: to actually squash "reverting changes" in their original commit so that it looks like they never happened.

It's quite easier to review if you can keep the changes to the minimum;
to that end, commits like "1453adf" could also use some love as you state "updated version for checkstyle" but for that I expect a single line to change.. reformatting the whole file is a no-go ;)

In this case I see that indeed there was a need to fix some indentation, but that really should be in a separate commit.
Also you changed many XML elements which end with " />" to remove the separating space.. the space is intentional in our style.

For sake of HSEARCH-2103 could you revert most of the non-related cosmetic changes?

For example in the checkstyle.xml file, I'd really like to be able to open the diff and have it point out what you did change.. it's hard to point fingers since it's significantly reformatted.

In such a case I'd avoid reformatting altogether (generally), but if formatting really needs some love, make sure you format it once w/o any other change, commit with a clear message and then apply your changes so that one can "trust" that the first commit isn't changing anything and read the second commit with more care.

N.B. this is more important in the long term than just for the one -off review. Sometimes one needs to dig in history and see when something was changed.. it's very time consuming when there are changes which aren't relevant so we need to keep those out of the main history.

Sanne:
all nice now :+1:

gsmet:
@s4ke when I run mvn -DskipTests checkstyle:check, I have the following errors:
```
[INFO] Building Hibernate Search ORM 5.6.0-SNAPSHOT
[INFO] 
[INFO] --- maven-checkstyle-plugin:2.17:check (default-cli) @ hibernate-search-orm ---
[INFO] Starting audit...
[ERROR] /data/home/gsmet/git/gsmet/hibernate-search/orm/src/test/resources/META-INF/persistence.xml:2: Line does not match expected header line of '<!--'. [Header]
[ERROR] /data/home/gsmet/git/gsmet/hibernate-search/orm/src/test/resources/META-INF/persistence.xml:22: White spaces at the end of line [RegexpSingleline]
[ERROR] /data/home/gsmet/git/gsmet/hibernate-search/orm/src/test/resources/META-INF/persistence.xml:27: White spaces at the end of line [RegexpSingleline]
[ERROR] /data/home/gsmet/git/gsmet/hibernate-search/orm/src/test/resources/org/hibernate/search/test/spatial/hibernate.cfg.xml:1: Line does not match expected header line of '<?xml version="1.0" encoding="UTF-8"?>'. [Header]
[ERROR] /data/home/gsmet/git/gsmet/hibernate-search/orm/src/test/resources/org/hibernate/search/test/spatial/hibernate.cfg.xml:35: White spaces at the end of line [RegexpSingleline]
```

Could you take a look at it?

Thanks!

s4ke:
@gsmet:

Huh that's weird. It seems like the error only occurs when used with the direct checkstyle:check goal. I will fix this and provide a PR.

gsmet:
Yup. That  surprised me but it looks like it comes up with things we should really fix.

s4ke:
see https://github.com/hibernate/hibernate-search/pull/1008

PR #999, opened 2016-02-22
HSEARCH-2105 Several improvements around ES embeddable mapping

gunnarmorling:
Hey @Sanne can you quickly check out this PR, specifically the last three commits of it (I based this on @gsmet's not yet merged faceting work).

It does two things:

* Don't use nested mappings by default with ES for now (as discussed, we can make nested the default for index-embedded collections with more than one field and/or make it user-configurable eventually)
* Ensures that for index-embedded collections those fields that belong to one collection element are mapped to the same array element in the resulting ES JSON array (again, that's not a real nested mapping with sub-docs, i.e. the index looks essentially the same as with plain HS+Lucene, but the field should be correctly correlated in the JSON array, otherwise users will hate it).

For the second issue I came up with a hack which I think is acceptable but I'd like your feedback before driving it further. The idea is to insert "marker fields" which denote the array index of the subsequent document fields. This allows to restore the properly structured JSON by examining the "flattened" document. Note that this only happens in the ES case (I service-yified this part, only ES plugs in a proper impl, whereas in plain Lucene it's no-ops).

To give an example, say I have an association from `GolfPlayer` to a collection of `GolfCourse` which is index-embedded. That's the relevant parts of the document as built for ES:

    $nesting=playedCourses[0] // marker field
    playedCourses.name=Purbeck
    playedCourses.rating=127.3
    $nesting=playedCourses[1]
    playedCourses.name=Mount Maja
    playedCourses.rating=111.9

The marker fields themselves won't be added to the JSON document of course, they are only info used to asemble the required structure.

It's hacky but it does the trick and IMHO is good enough for the time being. Pushing potentially wrongly structured data to ES is very bad, hence I'd go with that solution for now. In the next step we should look into abandoning using plain Lucene `Document`s as the actual datastructure which will make all that unneeded. ~~Note that there are two test failures I still need to sort out.~~

Please let me know what you think.

### DIFF
--- a/elasticsearch/pom.xml
+++ b/elasticsearch/pom.xml
@@ -299,6 +299,7 @@
                                 <exclude>**/SortWithIndexUninvertingTest.java</exclude>
                                 <exclude>**/SortOnFieldsFromCustomBridgeTest.java</exclude>
                                 <exclude>**/RollbackTransactionTest.java</exclude>
+                                <exclude>**/PrefixedEmbeddedCaseInPathTest.java</exclude>
 
                                 <!-- Specific to other backends -->
                                 <exclude>**/CustomBackendTest.java</exclude>
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchHSQueryImpl.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchHSQueryImpl.java
@@ -629,8 +629,9 @@ public class ElasticsearchHSQueryImpl extends AbstractHSQuery {
 		if ( aggregation == null ) {
 			return Collections.emptyList();
 		}
+
 		// deal with nested aggregation for nested documents
-		if ( FieldHelper.isEmbeddedField( facetRequest.getFieldName() ) ) {
+		if ( isNested( facetRequest ) ) {
 			aggregation = aggregation.getAsJsonObject().get( facetRequest.getFacetingName() );
 		}
 		if ( aggregation == null ) {
@@ -644,6 +645,12 @@ public class ElasticsearchHSQueryImpl extends AbstractHSQuery {
 					bucket.getAsJsonObject().get( "doc_count" ).getAsInt() ) );
 		}
 		return facets;
+	}
+
+	private boolean isNested(DiscreteFacetRequest facetRequest) {
+		//TODO Drive through meta-data
+//		return FieldHelper.isEmbeddedField( facetRequest.getFieldName() );
+		return false;
 	}
 
 	// TODO: Investigate scrolling API:
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchHSQueryImpl.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchHSQueryImpl.java
@@ -533,6 +533,11 @@ public class ElasticsearchHSQueryImpl extends AbstractHSQuery {
 			}
 			// Should only be the case for custom bridges
 			else {
+				// TODO: should we do it?
+				if ( !value.isJsonPrimitive() ) {
+					throw new SearchException( "Projection of non-JSON-primitive field values is not supported: " + value );
+				}
+
 				JsonPrimitive primitive = value.getAsJsonPrimitive();
 
 				if ( primitive.isBoolean() ) {
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
@@ -398,7 +398,13 @@ public class ElasticsearchIndexManager implements IndexManager {
 			JsonObject property = parentProperties.getAsJsonObject( part );
 			if ( property == null ) {
 				property = new JsonObject();
-				property.addProperty( "type", "nested" );
+
+				// TODO enable nested mapping as needed:
+				// * only needed for embedded *-to-many with more than one field
+				// * for these, the user should be able to opt out (nested would be the safe default mapping in this
+				// case, but they could want to opt out when only ever querying on single fields of the embeddable)
+
+//				property.addProperty( "type", "nested" );
 
 				JsonObject properties = new JsonObject();
 				property.add( "properties", properties );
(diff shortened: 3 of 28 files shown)


Sanne:
> as discussed, we can make nested the default for index-embedded collections with more than one field and/or make it user-configurable eventually

This reminds me of yet another problem we have with the current FieldBridge API: we don't really know how many fields are being added.

gunnarmorling:
> This reminds me of yet another problem we have with the current FieldBridge API: we don't really know how many fields are being added.

`MetadataProvidingFieldBridge` goes into this direction, i.e. telling *what* fields to expect? Note that for the ES schema it makes no difference whether a field is there once or n times. So I am not seeing a strong problem here right now, but I may be missing something.

gunnarmorling:
Pushed an update; @Sanne, any thoughts?

Sanne:
something wrong here:

Tests in error: 
  ElasticsearchNullValueIT>SearchTestBase.setUp:46 » Annotation Use of @OneToMan...
  ElasticsearchNullValueIT.deleteTestData:63->SearchTestBase.openSession:81 » IllegalState
  ElasticsearchNullValueIT>SearchTestBase.setUp:46 » Annotation Use of @OneToMan...
  ElasticsearchNullValueIT.deleteTestData:63->SearchTestBase.openSession:81 » IllegalState
  ElasticsearchIT.testEmbeddedIndexing:207 » AssertionFailure Exception releasin...
  ElasticsearchIT.deleteTestData:140->SearchTestBase.openSession:81 » IllegalState
  NestedEmbeddedTest.testNestedEmbeddedIndexingWithContainedInOnCollection:105 » AssertionFailure
  PrefixedEmbeddedCaseInPathTest.setUp:48->persistEntity:124 » AssertionFailure ...
  PrefixedEmbeddedCaseInPathTest.setUp:48->persistEntity:124 » AssertionFailure ...
  PrefixedEmbeddedCaseInPathTest.setUp:48->persistEntity:124 » AssertionFailure ...
  PrefixedEmbeddedCaseInPathTest.setUp:48->persistEntity:124 » AssertionFailure ...

Tests run: 328, Failures: 0, Errors: 11, Skipped: 0


This was after a rebase. Trying again on your exact same commit...

Sanne:
Re-run the testsuite on your exact same commit and it passed all tests.
Checking again after rebase...

Sanne:
It worked fine twice after the rebase. I probably messed up the previous update, sorry..

Sanne:
> MetadataProvidingFieldBridge goes into this direction, i.e. telling what fields to expect? Note that for the ES schema it makes no difference whether a field is there once or n times. So I am not seeing a strong problem here right now, but I may be missing something.

I didn't mean that the same field gets added twice. I mean that there are cases in which a single property contributes multiple different index fields:
 - https://github.com/hibernate/hibernate-search/blob/e2f3e1e98eb903e08f34339ee4408bef31d18e3a/orm/src/test/java/org/hibernate/search/test/bridge/DateSplitBridge.java#L42

So I don't see how you can plan to have ES use one kind of mapping vs the other one based on the fact if we're adding more than one field for a _ToMany IndexedEmbedded relation.

Sanne:
I'm trying to understand why you modelled `ElasticsearchNestingContextFactoryProvider` as a Service, and it using a CHM to "cache" the `ContextCreationStratey`. You also decided to make sure these are lazily initialized, surely as the mapping isn't complete on service start.

But rather than having a cache, shouldn't this strategy be (conceptually) a field of the `DocumentBuilderIndexedEntity` ?
(if that was not an -engine specific building block).

Wondering if we shouldn't have a different implementation of `DocumentBuilderIndexedEntity` for the Elasticsearch case.

Let's assume you made this as a temporary work around; then why bother with the pitfalls of caching? Is it a performance related thought?

gunnarmorling:
> I'm trying to understand why you modelled ElasticsearchNestingContextFactoryProvider as a Service, and it using a CHM to "cache" the ContextCreationStratey. You also decided to make sure these are lazily initialized, surely as the mapping isn't complete on service start.

Many good questions :)

It's a service so I can plug in the implementation which resides in the _elasticsearch_ module. If no service implementation is present, the no-op default implementation will be used.

The implementation is lazy because I don't know before whether a given type is mapped to ES (thus requiring the specific nesting context) or not (thus allowling the no-op impl to be used). It uses the CHM to avoid doing this decision again and again.

> But rather than having a cache, shouldn't this strategy be (conceptually) a field of the DocumentBuilderIndexedEntity ?
> (if that was not an -engine specific building block).

Yes, having specific `DocumentBuilderIndexedEntity` implementations would be a better approach and the one we should pursue eventually. Then the ES backend could apply its specific DBIE implementation and we can get rid of that nesting context scheme. I only wen that route to keep changes in a digestable size and not do all at once.

> Wondering if we shouldn't have a different implementation of DocumentBuilderIndexedEntity for the Elasticsearch case.

Yes, absolutely. It'd do that probably when getting rid of `Document` as the central "format" used internally as discussed.

> Let's assume you made this as a temporary work around; then why bother with the pitfalls of caching? Is it a performance related thought?

Right, it's temporary, and yes, the caching is meant to avoid determining the type of backend repeatedly at runtime.

gunnarmorling:
@Sanne Rebased to latest master. Now it's just the four actual commits :)

gunnarmorling:
> I mean that there are cases in which a single property contributes multiple different index fields:

Assuming `DateSplitBridge` would implement `MetadataProvidingFieldBridge` and properly announce the three fields it creates, it should be doable.

Sanne:
fixing a typo in a follow-up commit! tsk tsk .. :)

Thanks for the explanation, it all makes sense now.

I'm wondering if we should add a bit more context or "what the intent is" in PRs, especially when it's random progress not clearly defined in a JIRA.

gunnarmorling:
> I'm wondering if we should add a bit more context or "what the intent is" in PRs, especially when it's random progress not clearly defined in a JIRA.

I had hoped my explanation at the top of this PR would make the intention clear. I'll try harder next time :)

Sanne:
> Assuming DateSplitBridge would implement MetadataProvidingFieldBridge and properly announce the three fields it creates, it should be doable.

The problem is we don't mandate to implement `MetadataProvidingFieldBridge`.. so what do you do when there are custom fieldbridges around which don't? Default to nesting as only safe option?
Either way, it would be "surprising behaviour" for some people.

gunnarmorling:
Without this, we cannot create the proper ES index structure in such cases
anyways. So using this bridge as-is would fail at indexing time as thing
stand (I opted for "strict" index structure checking for the time being).

I think as of HSEARCH 6 we should make proper announcement of created
fields mandatory. I.e. merge MetadataProvidingFieldBridge into FB 2.0
essentially.

Sanne:
merged!

gunnarmorling:
Jippie!

PR #1002, opened 2016-02-24
HSEARCH-2091 Reuse the ElasticsearchIndexWorkVisitor

gsmet:


Sanne:
this is failing several tests! puzzling, I was also assuming it would be as simple as you made it.

Sanne:
The `ExtendedSearchIntegrator` which you are passing to the constructor of the Visitor isn't initialized yet.
This is only materialized after `org.hibernate.search.backend.elasticsearch.impl.ElasticsearchIndexManager.setSearchFactory(ExtendedSearchIntegrator)`

Sanne:
I guess you didn't test it?

gsmet:
Might have tested the wrong tree. Didn't surprise me it worked as I didn't think it might not.

Sanne:
np, it surprised me too :)

gsmet:
Time to commit https://github.com/hibernate/hibernate-search/pull/982 so that we can have CI in our forks :)!

Sanne:
Ok this is passing the tests, but now we're getting in trouble with concurrency.

Technically there are no issues: even if the _indexWorkVisitor_ field is not marked 'volatile' that means that in worst case if there are multiple parallel operations entering method we'll be creating a couple additional `ElasticsearchIndexWorkVisitor` instances. With the current implementation this is not a problem as we know there aren't resources in there which wouldn't be cleaned up, but can we rely on that in the long term?

For example I expect that we'll be reusing the JestClient instance rather than keep requesting a new one from the ServiceManager, when we'll be more interested in performance. That day we might start leaking resources because the `ElasticsearchBackendQueueProcessor` will be at risk of not closing all the `ElasticsearchIndexWorkVisitor` it initialized.

The correct solution would be a volatile field and a double checked lock... hugh :-( I start thinking I'd rather keep @gunnarmorling 's original approach unless we can significantly change this area.

I'd highly prefer we could make sure that we have all what we need to initialize what we need at constructor time than using lazy initialization... or simply postpone this issue for "sometime in the future".

gsmet:
OK, closing for now.

PR #1004, opened 2016-02-24
HSEARCH-2114 Enabling usage of "elasticsearch" short name for index m…

gunnarmorling:
…anager configuration

### DIFF
--- a/elasticsearch/src/test/resources/hibernate.properties
+++ b/elasticsearch/src/test/resources/hibernate.properties
@@ -22,5 +22,5 @@ hibernate.cache.region_prefix hibernate.test
 hibernate.cache.provider_class org.hibernate.cache.HashtableCacheProvider
 
 hibernate.search.elasticsearch.host http://127.0.0.1:9200
-hibernate.search.default.indexmanager org.hibernate.search.backend.elasticsearch.impl.ElasticsearchIndexManager
+hibernate.search.default.indexmanager elasticsearch
 hibernate.search.elasticsearch.index_management_strategy CREATE_DELETE
--- a/engine/src/main/java/org/hibernate/search/engine/impl/DefaultIndexManagerFactory.java
+++ b/engine/src/main/java/org/hibernate/search/engine/impl/DefaultIndexManagerFactory.java
@@ -9,8 +9,11 @@ package org.hibernate.search.engine.impl;
 import java.util.Properties;
 
 import org.hibernate.search.cfg.spi.IndexManagerFactory;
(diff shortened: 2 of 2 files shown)


elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManagerFactory.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Index manager capable of resolving the "elasticsearch" alias. All other names/aliases will be handled by the default
+ * implementation.
+ * <p>
+ * Registered through the Java service loader mechanism.
+ *
+ * @author Gunnar Morling
+ */
+public class ElasticsearchIndexManagerFactory extends DefaultIndexManagerFactory {
```
Sanne: Man it's tempting as it's a nice short and clean PR, but this would make all the Elasticsearch work completely unusable for Infinispan Query!

The services system isn't designed for this as it would blow up if it were to find multiple alternative implementations; Infinispan Query also overrides the DefaultIndexManagerFactory to add its own short names.

We need to create a "short name contributors" concepts and be able to apply all of them, not picking only one.

gunnarmorling: Oh, I see. I think for now I'll just add the "elasticsearch" name to the actual default factory then (yes, I know, knowledge of ES in core and all that... ;-)). But yes, let's generalize that scheme quickly.

Sanne: +1

Sanne: @gustavonalle you might want to keep this problem in mind too ^

https://hibernate.atlassian.net/browse/HSEARCH-2115

Sanne: @gunnarmorling as explained in the above JIRA, writing it up reminded me of another thing: the Infinispan case goes beyond contributing it actually overrides some names from the default factory. It extends the default to "pass on" all names it doesn't need to override though, so introducing "elasticsearch" in the default would work.

gunnarmorling:
Ok, I've pushed an update to encode the alias in the default factory for now.

Sanne:
thanks, merged!

PR #1008, opened 2016-02-26
HSEARCH-2126 Checkstyle fails when explicitly called via goal checkstyle:check

s4ke:
@gsmet: can you run this on your machine as well?

### DIFF
--- a/build-config/src/main/resources/suppressions.xml
+++ b/build-config/src/main/resources/suppressions.xml
@@ -10,5 +10,7 @@
     "http://www.puppycrawl.com/dtds/suppressions_1_1.dtd">
 <suppressions>
     <suppress checks="." files="ConcurrentReferenceHashMap\.java$" />
+    <!-- ignore everything but the header checks for xml files-->
+    <suppress checks="^((?!Header).)*$" files=".*\.xml" />
     <suppress files="[\\/]generated-sources[\\/]" checks="[a-zA-Z0-9]*"/>
 </suppressions>
--- a/elasticsearch/src/test/resources/META-INF/persistence.xml
+++ b/elasticsearch/src/test/resources/META-INF/persistence.xml
@@ -1,4 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!--
+ ~ Hibernate Search, full-text search for your domain model
+ ~
+ ~ License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ ~ See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+  -->
 <persistence xmlns="http://java.sun.com/xml/ns/persistence" version="1.0">
     <persistence-unit name="EntityManagerSerializationTestPU">
         <class>org.hibernate.search.test.jpa.Bretzel</class>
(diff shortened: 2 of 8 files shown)


gsmet:
I still have an error
```
[ERROR] /data/home/gsmet/git/gsmet/hibernate-search/orm/src/test/resources/org/hibernate/search/test/spatial/hibernate.cfg.xml:1: Line does not match expected header line of '<?xml version="1.0" encoding="UTF-8"?>'. [Header]
```

utf-8 should be UTF-8 in the header. Once this one is fixed, it works OK.

s4ke:
I added some additional fixes that showed up (for some reason).

Sanne:
We never meant to run checkstyle on XML files. The formatting improvements in these changed XML look fine, but do we want to enforce it on XML?

gsmet:
I don't have a strong opinion about it but it doesn't look like too much of a hassle.

Just wanted to have checkstyle:check working.

s4ke:
@Sanne:

Well, I haven't added any rules regarding xml. Also, this

`<module name="Header">
        <property name="header" value="&lt;?xml version=&quot;1.0&quot; encoding=&quot;UTF-8&quot;?&gt;\n&lt;!--\n ~ Hibernate Search, full-text search for your domain model\n ~\n ~ License: GNU Lesser General Public License (LGPL), version 2.1 or later\n ~ See the lgpl.txt file in the root directory or &lt;http://www.gnu.org/licenses/lgpl-2.1.html&gt;.\n  --&gt;"/>
        <property name="fileExtensions" value="xml"/>
    </module>`

suggests that at some point this was intended.

s4ke:
@Sanne: bump

Sanne:
yes we can merge this, but beware I think people will eventually make XML formatting mistakes and the regular build wouldn't catch it (the same way it didn't catch these before).

Could we adjust this to make sure the `checkstyle:check` coverage is the same as when doing a normal (full) build?

s4ke:
Errors in .xml files are also reported during a normal build, but as it seems the default behaviour has changed. We could just disable all checkstyle checks on .xml files apart from the header check, this would do it as well, right?

s4ke:
@Sanne : i changed it so only the Header check is run on xml files.

Sanne:
@s4ke try this: remove the header from one of the files you fixed, then run the build again. It won't complain...

That's what I meant by saying that it's not regression-proof.

Sure, it's an improvement in style but I'm pretty sure it won't last if it's not automatically prevented ;)

If you want I can merge this.. not a problem. Just saying that if we could enable the automatic check as well, then you'll have solved the problem for longer term.

Sanne:
This is what I'm testing:
 - I fetch your branch
 - I revert your commit 3541607
 - I run `mvn clean install`

I would expect it to fail, but it's not. Am I running the wrong Maven goal?

BTW please squash 2232f66c4f74c7c58f6e547564139274f1975086 into its parent ;)

s4ke:
@Sanne:

During the build checkstyle:checkstyle is used which generates reports. For some reason this apparently behaves differently to checkstyle:check when it comes to error checking.

Sanne:
Awesome, seems you untangled the mystery :+1: 

Merged

PR #1009, opened 2016-02-29
HSEARCH-2131 Include basic Elasticsearch getting started elements in …

Sanne:
…the documentation

 - https://hibernate.atlassian.net/browse/HSEARCH-2131


Sanne:
self merged to get the release out: the wording might not be polished but it's better than nothing.

PR #1010, opened 2016-02-29
Elasticsearch - Spatial support

gsmet:


Sanne:
Seems you catched an existing bug. Could you split this commit up in such a way that the missing Spatial metadata is in its own PR, and we can then backport that to 5.5 ?

Also opened https://hibernate.atlassian.net/browse/HSEARCH-2139 as an observation on the code (an existing problem too)... it would be trivial to fix but I'll wait for you to rework this PR first so we avoid conflicts.

pom.xml
```diff
@@ -181,6 +181,11 @@
         <activeMQVersion>5.8.0</activeMQVersion>
         <commonsLangVersion>2.6</commonsLangVersion>
 
+        <!-- Elasticsearch -->
+        <elasticsearchMavenPluginVersion>2.0</elasticsearchMavenPluginVersion>
+        <!-- Keep in sync with the dependency required by Elasticsearch -->
+        <elasticsearchGroovyVersion>2.4.4</elasticsearchGroovyVersion>
```
gunnarmorling: What exactly is Groovy used for here?

gunnarmorling:
Hey @gsmet, I'm reall excited to see this one!

Could you describe the overall approach in a few words? E.g. I see you try to use the ES "geo_point" type (which is great!), but when looking at the `POI` mapping, I see "longitude" and "latitude" as fields, too:

"mappings" : {
      "org.hibernate.search.test.spatial.POI" : {
        "dynamic" : "strict",
        "properties" : {
          "__HSearch_TenantId" : {
            "type" : "string"
          },
          "latitude" : {
            "type" : "double",
            "store" : true
          },
          "location" : {
            "type" : "geo_point"
          },
          "longitude" : {
            "type" : "double",
            "store" : true
          },
          "name" : {
            "type" : "string",
            "store" : true
          },
          "type" : {
            "type" : "string",
            "store" : true
          }
        }
      }
    }
  }

Is this exected? I'd only have expected "location" here.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/DistanceSort.java
```diff
@@ -0,0 +1,51 @@
...
+		this.center = center;
+	}
+
+	@SuppressWarnings("unchecked")
+	@Override
+	public Map<String, Object> toMap() {
+		Map<String, Object> rootMap = super.toMap();
+		Map<String, Object> innerMap = (Map<String, Object>) rootMap.get(GEO_DISTANCE_FIELD);
+
+		Map<String, Double> location = new HashMap<String, Double>(2);
```
gunnarmorling: This should better be initialized with a size of 3. The reason being the load factor defaulting to .75. So unlike intended, the second `put()` will trigger a resize of the internal table.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchHSQueryImpl.java
```diff
@@ -400,28 +399,55 @@ private JsonObject getTenantIdFilter() {
...
+			if ( isPartOfProjectedFields( ProjectionConstants.SPATIAL_DISTANCE ) ) {
+				// we add a script_field to return the distance
+				query.add( "script_fields",
+						JsonBuilder.object().add( SPATIAL_DISTANCE_FIELD, JsonBuilder.object()
+								.add( "params",
+										JsonBuilder.object()
+												.addProperty( "lat", spatialSearchCenter.getLatitude() )
+												.addProperty( "lon", spatialSearchCenter.getLongitude() )
+								)
+								.addProperty( "script", "doc[\u0027" + spatialFieldName + "\u0027].arcDistanceInKm(lat,lon)" )
```
gunnarmorling: Ok, I suppose that's where Groovy comes into play.

Did you look into [sorting by distance](https://www.elastic.co/guide/en/elasticsearch/guide/current/sorting-by-distance.html) on the ES side? This looks exactly like what we need, so we'd avoid that custom script.

orm/src/test/java/org/hibernate/search/test/jpa/SpatialQueryingJPATest.java
```diff
@@ -74,12 +74,12 @@ public void testDistanceProjection() throws Exception {
 		Object[] fourthResult = (Object[]) results.get( 3 );
 		Object[] fifthResult = (Object[]) results.get( 4 );
 		Object[] sixthResult = (Object[]) results.get( 5 );
-		Assert.assertEquals( 0.0, (Double) firstResult[1], 0.0001 );
-		Assert.assertEquals( 10.1582, (Double) secondResult[1], 0.0001 );
-		Assert.assertEquals( 11.1195, (Double) thirdResult[1], 0.0001 );
-		Assert.assertEquals( 15.0636, (Double) fourthResult[1], 0.0001 );
-		Assert.assertEquals( 22.239, (Double) fifthResult[1], 0.001 );
-		Assert.assertEquals( 24.446, (Double) sixthResult[1], 0.001 );
+		Assert.assertEquals( 0.0, (Double) firstResult[1], 0.01 );
```
gunnarmorling: Do you have any idea why we loose one digit of precision here?

orm/src/test/java/org/hibernate/search/test/spatial/SpatialIndexingTest.java
```diff
@@ -180,6 +182,7 @@ public void testDistanceSort() throws Exception {
 
 	@Test
 	@TestForIssue(jiraKey = "HSEARCH-1708")
+	@Category(ElasticsearchSupportInProgress.class)
```
gunnarmorling: Probably a short comment why it's part of this category will help to remember the context when getting back to it.

gunnarmorling:
@gsmet, done with my review. Overall it looks great. Some remarks/questions I have:

* Dedicated latitude/longitude fields should not be added to the index as there is the location field
* Can the built-in sorting used instead of the custom script
* The `HASH` strategy seems to be mapped via `geo_point`. There is support [for geo hashes](https://www.elastic.co/guide/en/elasticsearch/guide/current/geohash-cell-filter.html), which seems more appropriate on first sight? This could be done as a follow-up PR, if you think it's a good idea.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchHSQueryImpl.java
gsmet: I use the sorting by distance to sort.

The script_field is used to get the distance from the index when you asked for the SPATIAL_DISTANCE projection. You can get it in the response only *if* you sort by distance which might not be the case. That's why I chose this option.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/DistanceSort.java
gsmet: Fixed.

orm/src/test/java/org/hibernate/search/test/jpa/SpatialQueryingJPATest.java
gsmet: Nope. They probably use different algorithms to calculate the distance. Information about it are sparse. It was far worse when I used the distanceInKm function and became acceptable with the arcDistanceInKm function.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchHSQueryImpl.java
gunnarmorling: Ah, I see. I think a short comment of sorts of this routine would be helpful.

Also, as an optimization, the script field could be omitted when sorting by distance.

We'd also need a note in the docs on the Groovy requirement on the ES side when projecting distance without sorting.

orm/src/test/java/org/hibernate/search/test/jpa/SpatialQueryingJPATest.java
gunnarmorling: Ok, I suppose it's fine for now.

orm/src/test/java/org/hibernate/search/test/spatial/SpatialIndexingTest.java
gsmet: I spent some time in LuceneHSQuery yesterday and I think I might as well fix it. Commit on the way.

gsmet:
Thanks for the review.

1/ They should be added as they are indexed explicitely with `@Field`. It's not the longitude/latitude fields derived from the location field which are effectively removed in favor of the geo_point.
2/ Answered inline.
3/ Yes, I need to take a look at that. I wanted to keep the approach simple for now. Will create a JIRA so that we track this.
4/ By native, you mean a native Elasticsearch query?

gunnarmorling:
> 1/ They should be added as they are indexed explicitely with @Field

Ah, I was fooled by this test then. Are these fields even needed? Seems confusing to have them both as "normal" fields and as geo. Seems we could remove them?

> 3/ Yes, I need to take a look at that. I wanted to keep the approach simple for now. Will create a JIRA so that we track this.

+1

> 4/ By native, you mean a native Elasticsearch query?

Yepp. "geo_shape", "geo_distance", something like that.

pom.xml
gsmet: @Sanne I added the versions and the dependencies in the root pom.xml but you did it in the elasticsearch/pom.xml file for the Elasticsearch test dependency. We should be consistent: what do we prefer? Personnally I like to have all the versions in the root pom.xml so that I don't forget to update a dependency and it looks consistent with the fact that arquillian versions are in the root pom.xml.

Thanks for the feedback.

Sanne: yes we can move them to the root pom, we have no strict rules about that, more like a pragmatic approach:

My general habit is to have them in the root pom when the property (the version) needs to be shared across multiple modules.
When it comes to integration testing details, sometimes we intentionally needed (using the past as I think all such cases are gone nowadays) different versions in different modules.

In this case since I tend to believe that the version doesn't "belong" in the parent pom as it's just a testing detail (we're attempting to not couple support to a specific ES version).. although we probably want to test with the same version when it will come to integration tests within Wildfly so then it would be more practical to have these defined in the parent pom.

gunnarmorling: +1 for moving this to the root POM: There is an ES integration test in the JGroups module, too, which should be using the same version.

gunnarmorling:
Hey @gsmet, where are we with this one? Are you still planning to add some more stuff, e.g. the native tests? Or is it good go from your PoV? Best is to remove the "Changes Required" label once done, so not everyone is waiting for everyone :) Thanks!

gsmet:
Hi @gunnarmorling I plan to add the native tests and clean up the POI test class. That's why I haven't changed the label. I'm not working full time yet ;)

gunnarmorling:
Sure, all cool, didn't mean to rush you. Just wanted to make sure we are
not waiting on each other. Take the time you need :) Looking forward to
April btw.!

gsmet:
@gunnarmorling should be OK now.

gsmet:
The upgrade to Elasticsearch 2.2.0 broke the Groovy support.

See http://stackoverflow.com/questions/35582520/elasticsearch-script-lang-not-supported-groovy .

I think we'll need to push something to the elasticsearch-maven-plugin. Will look at it tomorrow.

elasticsearch/src/test/java/org/hibernate/search/backend/elasticsearch/test/ElasticsearchSpatialIT.java
```diff
@@ -0,0 +1,150 @@
...
+		s.close();
+	}
+
+	@Test
+	public void testGeoDistanceQuery() {
+		Session s = openSession();
+		FullTextSession session = Search.getFullTextSession( s );
+		Transaction tx = s.beginTransaction();
+
+		String geoDistanceQuery = "{\n" +
```
gunnarmorling: No need really for the line breaks. But they don't do harm either.

elasticsearch/src/test/resources/elasticsearch/config/elasticsearch.yml
```diff
@@ -0,0 +1 @@
+script.engine.groovy.inline.search: on
```
gunnarmorling: This seems redundant; I'll remove it.

gunnarmorling:
Rebased and applied. Many thanks, @gsmet! I've added two more commits for doing some clean-up in the POMs, removing the redundant config file and adding the dependency to JNA, getting rid of that nasty exception during start-up :)

Sanne:
:+1:

PR #1012, opened 2016-03-02
Reliability of JGroups integration tests

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2141
 - https://hibernate.atlassian.net/browse/HSEARCH-2142


gunnarmorling:
@Sanne there are two test failures. I cannot judge on these: Is this PR still an improvement or does it actually make things worse?

Sanne:
those failures are the recurring failures I was hoping to fix with this :)

So it's existing failures, and while I had hoped this would have been enough it seems it's not but it still is a good step in  the right direction: our previous testing configuration files had mistakes.

gunnarmorling:
Rebased and merged. I'll leave HSEARCH-2141 open then for further investigation/fixes to these failures.

Sanne:
+1 thanks!

PR #1014, opened 2016-03-02
Spatial FieldBridge fixes extracted from #1010

gsmet:
As requested by @Sanne , I extracted the fixes to the FieldBridge contribution to the metadata.


Sanne:
looks great,  thanks!
running tests...

gsmet:
It's probably worth it to wait if we can do something for HSEARCH-2139 at the same time. Commenting in the issue right now.

Sanne:
wait to merge? too late I just pushed it :)

Closing, but not backported until there's real need for it.

PR #1016, opened 2016-03-02
HSEARCH-2145 Fix NPE when running checkstyle:check

gsmet:


s4ke:
Can you provide me with a snippet so I can confirm the NPE behaviour?

gsmet:
I just run mvn checkstyle:check on the tree. 

That being said I'm wondering if it might be due to the checkstyle upgrade and if a mvn install before running the check would have fixed it.

gsmet:
Confirmed.

@Sanne should we still commit this to make our check consistent with the official checkstyle ones?

If not, I'll close it.

Sanne:
looks good, merged!

PR #1017, opened 2016-03-03
HSEARCH-2108 Support Elasticsearch custom and Built-in analyzers

DavideD:

With this patch if Elasticsearch is used as IndexManager the check on the analyzer defintions is cancelled. 
If there is a wrong definition name in the code an error is thrown during the craetion of the index. Sadly I couldn't find any reliable way to check if an analyzer definition exists in on elasticsearch.

The catch is that it is not possible to use different indexmanagers with the elasticsearch one

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/analyzer/impl/ElasticsearchAnalyzer.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/analyzer/impl/ElasticsearchAnalyzer.java
@@ -1,0 +1,32 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.elasticsearch.analyzer.impl;
+
+import org.apache.lucene.analysis.Analyzer;
+
+/**
+ * Wraps analyzer implementations defined on Elasticsearch.
+ *
+ * @author Davide D'Alto
+ */
+public class ElasticsearchAnalyzer extends Analyzer {
+
+	private final String name;
+
+	public ElasticsearchAnalyzer(String name) {
+		this.name = name;
+	}
+
+	public String getName() {
+		return name;
+	}
+
+	@Override
+	protected TokenStreamComponents createComponents(String fieldName) {
+		return null;
+	}
+}
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
@@ -13,10 +13,12 @@ import java.util.Set;
 
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.search.similarities.Similarity;
+import org.hibernate.search.analyzer.spi.AnalyzerDefiner;
 import org.hibernate.search.annotations.Store;
 import org.hibernate.search.backend.BackendFactory;
 import org.hibernate.search.backend.IndexingMonitor;
 import org.hibernate.search.backend.LuceneWork;
+import org.hibernate.search.backend.elasticsearch.analyzer.impl.ElasticsearchAnalyzer;
 import org.hibernate.search.backend.elasticsearch.cfg.ElasticsearchEnvironment;
 import org.hibernate.search.backend.elasticsearch.cfg.IndexManagementStrategy;
 import org.hibernate.search.backend.elasticsearch.client.impl.JestClient;
@@ -41,6 +43,7 @@ import org.hibernate.search.indexes.spi.IndexManager;
 import org.hibernate.search.indexes.spi.ReaderProvider;
 import org.hibernate.search.metadata.NumericFieldSettingsDescriptor.NumericEncodingType;
 import org.hibernate.search.spi.WorkerBuildContext;
+import org.hibernate.search.util.impl.DelegateNamedAnalyzer;
 import org.hibernate.search.util.logging.impl.Log;
 import org.hibernate.search.util.logging.impl.LoggerFactory;
 
@@ -56,7 +59,7 @@ import io.searchbox.indices.mapping.PutMapping;
  *
  * @author Gunnar Morling
  */
-public class ElasticsearchIndexManager implements IndexManager {
+public class ElasticsearchIndexManager implements IndexManager, AnalyzerDefiner {
 
 	private static final Log LOG = LoggerFactory.make();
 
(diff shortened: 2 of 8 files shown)


gunnarmorling:
> Sadly I couldn't find any reliable way to check if an analyzer definition exists in on elasticsearch.

Hum,  that's sad. Did you have a look into the "_analyze" API which Sanne had pointed out? I found it currently fails when testing an unknown analyzer:

    curl -XGET 'localhost:9200/_analyze' -d '{ "analyzer" : "foo", "text" : "" }'

This will never return due to an exception on the ES side. The reason seems to be https://github.com/elastic/elasticsearch/issues/15148 or one of the bugs linked to it. I tested with ES 2.3.0-SNAPSHOT and there it works: I get back a meaningful exception message which we can interpret.

So I'd say we should use that once 2.3.0 is finalized. Could you open a follow-up JIRA?

engine/src/main/java/org/hibernate/search/analyzer/spi/AnalyzerDefiner.java
```diff
@@ -0,0 +1,35 @@
...
+ * @author Davide D'Alto
+ */
+public interface AnalyzerDefiner {
+
+	/**
+	 * Checks if the index manager can managed additional analyzer
+	 *
+	 * @return true if it supports additional analyzers without the need of an analyzer defintion
+	 */
+	boolean supportsAdditionalAnalyzer();
```
gunnarmorling: In which case would one implement this to return `false`? Isn't the fact that an IM implements this facet enough?

DavideD: Right, make sense.
it's a left over from one of my experiments, when I initially wanted to pass the name of the analyzer to check the existence,  I will remove it since it seems it cannot be done at the moment

DavideD:
> Hum, that's sad. Did you have a look into the "_analyze" A ...

Yes, I've tried the _analyze API and the result is the exception you pointed out on the server (and a TimeOut exception on the client). I left a message on their forum but I didn't find the issue ticket.

> I tested with ES 2.3.0-SNAPSHOT and there it works: I get back a meaningful exception message which we can interpret.

Cool

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -499,6 +523,16 @@ public String getActualIndexName() {
 		return actualIndexName;
 	}
 
+	@Override
+	public Analyzer createAnalyzer(String name) {
+		return new ElasticsearchAnalyzer( name );
```
gunnarmorling: I wished we'd avoid creating such "fake" implementations. The stuff gets tougher and tougher to grasp.

Did you consider to establish an abstraction, say `AnalyzerReference`. For ES, the implementation would just refer to the name, whereas for embedded Lucene there'd be another implementation referencing the actual Lucene `Analyzer` instance.

I'm not sure whether it's doable without breaking SPIs, but if it is, I'd prefer that. Otherwise it should go to the 6.0 to-do list.

engine/src/main/java/org/hibernate/search/analyzer/spi/AnalyzerDefiner.java
```diff
@@ -0,0 +1,35 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.analyzer.spi;
```
gunnarmorling: I would not make that an public SPI for now, esp. as the solution might change again if we can do the proper validation of the analyzer. Or at least marking it as experimental.

DavideD:
By the way, I'm not sure the _analyze API will work with custom analyzer, as it seems that you need to useit with an index before it is recognized (I cannot find the article on stackoverflow now, but I'll mention this in the JIRA)

engine/src/main/java/org/hibernate/search/analyzer/spi/AnalyzerDefiner.java
DavideD: I will mark it as experimental, as I think we need some sort of SPI anyway

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
gsmet: Just wanted to point out that we need a way to tell if an Analyzer is remote or if it should be locally used so an AnalyzerReference with this information would be nice.

For instance, see `ConnectedMultiFieldsPhraseQueryBuilder#createQuery`: we need a way to tell that we don't want to analyze the phrase here but pass it directly to Elasticsearch.

The need will be similar for all external search servers, Solr included.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/analyzer/impl/ElasticsearchAnalyzer.java
```diff
@@ -0,0 +1,32 @@
...
+		this.name = name;
+	}
+
+	public String getName() {
+		return name;
+	}
+
+	@Override
+	protected TokenStreamComponents createComponents(String fieldName) {
+		return null;
```
gsmet: I don't think it's safe to return null here. See `Analyzer#tokenStream`. Or we need to be sure that we won't call `tokenStream` with this `Analyzer`.

gunnarmorling: Yes, that's the issue I see with this. This "analyzer" will not be used
locally. So we should abstract and each backend could work with it own
implementation.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
DavideD: I'll dig into this a bit more.
I'm wondering though what would you expect when somebody define a field with
```
@AnalyzerDef( definition = "ec-analyzer )
private String field;
```
Where ec-analyzer is a custom analyzer defined on Elasticsearch.
Would you expect this to work only on elastic search and be ignored on the other backends?

Sanne: N.B. AnalyzerDef is not an annotation which makes much sense to have on a field.

Remember it purely *defines* the analyzer, it doesn't apply it.

I see the source code lists

    @Target({ ElementType.PACKAGE, ElementType.TYPE, ElementType.FIELD, ElementType.METHOD })

but I think that's a mistake.
Think of it as something similar to `@NamedQuery`.

DavideD: My bad, obviously what I meant is:
```
		// Defined in the elasticsearch.yml configuration file
		@Analyzer(definition = "custom-analyzer")
		private String customTweet;
```

DavideD:
Superseeded by https://github.com/hibernate/hibernate-search/pull/1034

PR #1018, opened 2016-03-03
HSEARCH-2152 Bulk requests per list of Lucene works

gunnarmorling:
Cuts the build time for the _elasticsearch_ module by one third on my machine :)

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchBackendQueueProcessor.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchBackendQueueProcessor.java
@@ -6,15 +6,26 @@
  */
 package org.hibernate.search.backend.elasticsearch.impl;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Properties;
 import java.util.concurrent.locks.Lock;
 
 import org.hibernate.search.backend.IndexingMonitor;
 import org.hibernate.search.backend.LuceneWork;
+import org.hibernate.search.backend.elasticsearch.client.impl.JestClient;
 import org.hibernate.search.backend.spi.BackendQueueProcessor;
+import org.hibernate.search.engine.integration.impl.ExtendedSearchIntegrator;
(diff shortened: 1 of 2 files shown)


Sanne:
very nice! merged

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchBackendQueueProcessor.java
```diff
@@ -24,10 +35,12 @@
 public class ElasticsearchBackendQueueProcessor implements BackendQueueProcessor {
 
 	private ElasticsearchIndexManager indexManager;
+	private ExtendedSearchIntegrator searchIntegrator;
 
 	@Override
 	public void initialize(Properties props, WorkerBuildContext context, IndexManager indexManager) {
 		this.indexManager = (ElasticsearchIndexManager) indexManager;
+		this.searchIntegrator = context.getUninitializedSearchIntegrator();
```
gsmet: Sorry to be late to the party! I have a question about this one: is there a reason why you don't use the searchIntergrator from the indexManager as it's done below?

If so would it be an acceptable way to initialize the ElasticsearchIndexWorkVisitor in the constructor once and for all?

gunnarmorling: The SI in the index manager is still null at this point, so this doesn't work.

But I can take the SI from the context and initialize the visitor here with it once. At the time it will acutally be used, the SI should be fully set up. I've filed PR #1022 for this.

gsmet: > The SI in the index manager is still null at this point, so this doesn't work.

I sure know it now that I made the mistake the other day :). But when you use it in the apply* methods, it is, so you could have used the indexManager.searchIntegrator without initializing a specific one in the constructor.

Either this or your subsequent PR is fine IMHO.

PR #1020, opened 2016-03-04
HSEARCH-2155 Make sure Elasticsearch exceptions are logged in the bui…

Sanne:
…ld output

 - https://hibernate.atlassian.net/browse/HSEARCH-2155

### DIFF
--- a/elasticsearch/pom.xml
+++ b/elasticsearch/pom.xml
@@ -362,7 +362,7 @@
                     <httpPort>9200</httpPort>
                     <pluginsPath>${project.build.directory}/_ES_PLUGINS_</pluginsPath>
                     <outputDirectory>${project.build.directory}/elastisearchnode1</outputDirectory>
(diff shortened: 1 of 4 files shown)


Sanne:
This trick doesn't work if the following is defined:

    -Dlog4j.configuration=

Should we explicitly check for this to not be set? (is that doable with Maven?)

gunnarmorling:
>  Should we explicitly check for this to not be set? (is that doable with Maven?)

You could add a Maven profile enabled by the presence of this poperty and then e.g. add the Groovy Maven plug-in in that profile to fail the build. Or you create a custom enforcer rull doing the opposite of this one: http://maven.apache.org/enforcer/enforcer-rules/requireProperty.html

Whether it's worth it? Don't know :)

gunnarmorling:
Rebased and applied. If you think the property checking is worth it, just send a follow-up PR. Thanks!

PR #1021, opened 2016-03-07
HSEARCH-2139 Reuse target field names in SpatialFieldBridges

gsmet:


engine/src/main/java/org/hibernate/search/spatial/SpatialFieldBridgeByHash.java
```diff
@@ -54,7 +56,7 @@ public void configureFieldMetadata(String name, FieldMetadataBuilder builder) {
 		super.configureFieldMetadata( name, builder );
 		if ( spatialHashIndex ) {
 			for ( int i = topSpatialHashLevel; i <= bottomSpatialHashLevel; i++ ) {
-				builder.field( SpatialHelper.formatFieldName( i, name ), FieldType.STRING );
+				builder.field( hashIndexedFieldNames[i], FieldType.STRING );
 			}
 		}
 		if ( numericFieldsIndex ) {
```
gunnarmorling: `latitudeIndexedFieldName` and `longitudeIndexedFieldName` could be re-used here.

engine/src/main/java/org/hibernate/search/spatial/SpatialFieldBridgeByHash.java
```diff
@@ -63,6 +65,16 @@ public void configureFieldMetadata(String name, FieldMetadataBuilder builder) {
 		}
 	}
 
+	@Override
+	protected void initializeIndexedFieldNames(String fieldName) {
+		super.initializeIndexedFieldNames( fieldName );
+
+		hashIndexedFieldNames = new String[bottomSpatialHashLevel + 1];
```
gunnarmorling: This seems to create the hash field names also if not needed (in case of numericFieldsIndex).

gunnarmorling:
I've created #1028 with some clean-ups, so I'm going to close this one. Thanks!

PR #1022, opened 2016-03-07
HSEARCH-2091 Caching and re-using ES index work visitor

gunnarmorling:
I don't think it makes a huge difference, but it should be a tad better, at least until the visitor contains some actual state :)


gsmet:
This one has its own JIRA: https://hibernate.atlassian.net/browse/HSEARCH-2091 .

Looks good to me.

gunnarmorling:
Ah, thanks. Changed commit message and force-pushed.

Sanne:
merged!

PR #1023, opened 2016-03-07
HSEARCH-2146 Waiting until ES indexes have become green

gunnarmorling:
The second commit aims at parallelizing index creation. While it works it feels a bit ad-hoc-ish. Let me know what you think. It cuts down test execution time a bit more, but of obviously the effect will be larger if there are more indexes involved.

Note that I'm handling (index creation + mapping creation) as one execution unit, one could think about splitting these two up, too. This wouldn't improve the end-to-end time, but it'd free threads when waiting for index creation to finish. I don't think it's needed right now.

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
@@ -103,7 +103,7 @@ public class ElasticsearchIndexManager implements IndexManager {
 
 	private IndexManagementStrategy getIndexManagementStrategy(Properties properties) {
 		String strategy = properties.getProperty( ElasticsearchEnvironment.INDEX_MANAGEMENT_STRATEGY );
-		return strategy != null ? IndexManagementStrategy.valueOf( strategy ) : IndexManagementStrategy.NONE;
+		return strategy != null ? IndexManagementStrategy.valueOf( strategy ) : ElasticsearchEnvironment.Defaults.INDEX_MANAGEMENT_STRATEGY;
 	}
 
 	private String getIndexManagementWaitTimeout(Properties properties) {
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
@@ -42,11 +42,15 @@ import org.hibernate.search.indexes.spi.ReaderProvider;
 import org.hibernate.search.metadata.NumericFieldSettingsDescriptor.NumericEncodingType;
 import org.hibernate.search.spatial.impl.SpatialHelper;
 import org.hibernate.search.spi.WorkerBuildContext;
+import org.hibernate.search.util.configuration.impl.ConfigurationParseHelper;
 import org.hibernate.search.util.logging.impl.Log;
 import org.hibernate.search.util.logging.impl.LoggerFactory;
 
 import com.google.gson.JsonObject;
 
+import io.searchbox.client.JestResult;
+import io.searchbox.cluster.Health;
+import io.searchbox.cluster.Health.Builder;
 import io.searchbox.indices.CreateIndex;
 import io.searchbox.indices.DeleteIndex;
 import io.searchbox.indices.IndicesExists;
@@ -64,6 +68,7 @@ public class ElasticsearchIndexManager implements IndexManager {
 	private String indexName;
 	private String actualIndexName;
 	private IndexManagementStrategy indexManagementStrategy;
+	private String indexManagementWaitTimeout;
 	private Similarity similarity;
 
 	ExtendedSearchIntegrator searchIntegrator;
@@ -84,6 +89,7 @@ public class ElasticsearchIndexManager implements IndexManager {
 		this.indexName = getIndexName( indexName, properties );
 		try ( ServiceReference<ConfigurationPropertiesProvider> propertiesProvider = serviceManager.requestReference( ConfigurationPropertiesProvider.class ) ) {
 			this.indexManagementStrategy = getIndexManagementStrategy( propertiesProvider.get().getProperties() );
+			this.indexManagementWaitTimeout = getIndexManagementWaitTimeout( propertiesProvider.get().getProperties() );
 		}
 		this.actualIndexName = IndexNameNormalizer.getElasticsearchIndexName( this.indexName );
 		this.similarity = similarity;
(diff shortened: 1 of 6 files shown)


elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -139,7 +179,31 @@ public void addContainedEntity(Class<?> entity) {
...
+
+		waitForIndexCreation();
+	}
+
+	private void waitForIndexCreation() {
+		Builder healthBuilder = new Health.Builder()
+				.setParameter( "wait_for_status", "green" )
+				.setParameter( "timeout", indexManagementWaitTimeout );
+
+		Health health = new Health(healthBuilder ) {
```
gsmet: Minor nitpicking: looks like it doesn't respect the coding styles.

engine/src/main/java/org/hibernate/search/spi/SearchIntegratorBuilder.java
```diff
@@ -225,10 +234,41 @@ private ExtendedSearchIntegrator buildNewSearchFactory() {
...
 		factoryState.setActiveSearchIntegrator( factory );
 		rootFactory.setDelegate( factory );
+
+		postInitializeIndexManagers( documentBuildersIndexedEntities.values() );
+
 		return rootFactory;
 	}
 
 	/**
+	 * Does post-initialization of entity managers. Retrieved callables are executed in parallel, then we wait until all
```
gsmet: Shouldn't it be "index managers"?

gunnarmorling: True, I'll change it :)

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
gunnarmorling: Indeed; Wondering why CS didn't complain. I'll change it.

gsmet:
Once the checkstyle issue is fixed, I think we should merge the first commit.

As the second commit goes,

gunnarmorling:
Addressed @gsmet's comments. Thanks for checking!

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -100,6 +127,11 @@ private IndexManagementStrategy getIndexManagementStrategy(Properties properties
 		return strategy != null ? IndexManagementStrategy.valueOf( strategy ) : IndexManagementStrategy.NONE;
 	}
 
+	private String getIndexManagementWaitTimeout(Properties properties) {
+		String timeout = properties.getProperty( ElasticsearchEnvironment.INDEX_MANAGEMENT_WAIT_TIMEOUT );
+		return timeout != null ? Integer.valueOf( timeout ) + "ms" : DEFAULT_INDEX_MANAGEMENT_WAIT_TIMEOUT;
```
Sanne: Do we need better validation, i.e. being a positive number and a reasonable range?

Sanne: And testing for `NumberFormatException`

Sanne: org.hibernate.search.util.configuration.impl.ConfigurationParseHelper.parseInt(String, int, String)

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -115,7 +147,15 @@ public void setSearchFactory(ExtendedSearchIntegrator boundSearchIntegrator) {
...
 		this.clientReference = searchIntegrator.getServiceManager().requestReference( JestClient.class );
 
-		initializeIndex();
+		// for dynamic sharding postInitialize() will not be invoked by the builder, so doing it here ourselves
+		if ( postInitialize != null ) {
+			try {
+				postInitialize.call();
+			}
+			catch (Exception e) {
+				throw new SearchException( "Index manager initialization failed", e );
```
Sanne: Use the logger to throw exceptions?

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -141,7 +181,31 @@ public void addContainedEntity(Class<?> entity) {
...
+			@Override
+			protected String buildURI() {
+				return super.buildURI() + actualIndexName;
+			}
+		};
+
+		JestResult result = clientReference.get().executeRequest( health, false );
+
+		if ( !result.isSucceeded() ) {
+			throw new SearchException( "Index " + actualIndexName + " wasn't created in time; Reason: " + result.getErrorMessage() );
```
Sanne: Use the logger to throw exceptions?

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
gunnarmorling: Ok, I'll use that helper.

Let's not go overboard with this stuff, though. The prop is documented, and people will get a meaningful exception when giving something else than a number. If someone configures a negative timeout, I think I don't care ;)

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
gunnarmorling: What I head in mind was to `Logger`-ify everything once we finalize all this work and the dust has settled.

As long as things are in flow, I noticed one moves forth and back with creating these logger methods which seems like a waste of time.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
Sanne: I might be too strict with user input validation, but in my experience people always have huge amounts of issues with our properties.
I've considered moving to a "typesafe" configuration format like XML but having seen how much pain it introduces when the schemas have to be updated for new features or better formats (i.e. Infinispan) I'm glad we're using properties.. we just need to make sure we're paranoid with validation.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
Sanne: Ok that's fine by me.

gunnarmorling: I've opened [HSEARCH-2168](https://hibernate.atlassian.net/browse/HSEARCH-2168) for log interface usage so we don't forget about it.

gunnarmorling:
@emmanuelbernard Test

gunnarmorling:
Further comment, not mentioning the one whose name must not be mentioned.

gunnarmorling:
@Sanne Reworked some of the things you mentioned and force-pushed.

Note that I've removed the commit which parallelized index creation as I noticed it was causing data races when it comes to indexes shared between different entities. This will need some more consideration.

Sanne:
thanks ! And +1 for the safe approach.

gunnarmorling:
It's not that it wasn't "safe", it just would drop/create the index multiple times during initialization. Tests were all passing.

Sanne:
merged, except the test as discussed on hipchat.

PR #1025, opened 2016-03-07
HSEARCH-2162 Use the Hibernate JDBC properties consistently in TestRu…

gsmet:
…nnerStandalone

Followup of HSEARCH-2149: we can't use a mix of Hibernate and HikariCP properties
as they are merged by HikariCP and two of them conflict.

In Eclipse, HSEARCH-2149 made the test work but probably because the classpath is not clean and it loaded an hibernate.properties. HSEARCH-2149 also made the test work with Maven when a database profile was set.

But with Maven and without a database profile, the test didn't run anymore. This commit fixes it in all cases by using consistently the Hibernate JDBC properties instead of mixing Hibernate properties and HikariCP specific properties: you cannot set a driverClassName and a datasourceClassName and the driverClassName is automatically copied from the Hibernate JDBC properties if defined.


gunnarmorling:

Sanne:
Using the properties in this way didn't work with Hikari. Does this mean this was fixed in ORM?

PR #1030, opened 2016-03-08
HSEARCH-2147 Implement hash strategy for spatial queries with Elasticsearch

gsmet:


elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -276,8 +276,12 @@ private void addFieldMapping(JsonObject payload, BridgeDefinedField bridgeDefine
...
-				// we only add the geo_point once
+			if ( SpatialHelper.isSpatialFieldLongitude( simpleFieldName ) ) {
+				// we ignore the longitude field, we will create the geo_point mapping only once with the latitude field
+				return;
+			}
+			else if ( SpatialHelper.isSpatialFieldLatitude( simpleFieldName ) ) {
+				// we only add the geo_point for the latitude field
 				JsonObject field = new JsonObject();
 
 				field.addProperty( "type", "geo_point" );
```
gunnarmorling: Do we need the "geo_point" field *and* the hash fields?

gsmet: Yup. The range and hash approaches have 2 phases:
- in the case of range, the first phase uses a bounding box filter then an exact filter with the distance so you only need longitude and latitude
- in the case of hash, the first phase uses hashes to filter most of the results then an exact filter with the distance so you need the hashes and latitude/longitude.

In both cases, the results are exact and the difference is just the way the first phase is optimized.

gunnarmorling: Ok, I see. Should we then (probably in a separate issue, though) offer an option to make that first phase work with actual ES geohashes?

gunnarmorling:

PR #1032, opened 2016-03-08
HSEARCH-2143 Showing request/response in case of failures upon calling ES

gunnarmorling:


elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/GsonHolder.java
```diff
@@ -12,18 +12,17 @@
...
  */
-public class GsonBuilderHolder {
+public class GsonHolder {
 
-	private GsonBuilderHolder() {
+	private GsonHolder() {
 	}
 
-	public static final Gson BUILDER = new GsonBuilder()
+	public static final Gson INSTANCE = new GsonBuilder()
```
gsmet: Do you mind calling it GSON instead? I put an instance of the Gson JsonParser here in a following patch.

gunnarmorling: Done.

gunnarmorling:
Added some more commits to address HSEARCH-2151. I had to re-organize some things around the bulk handling, but everything should be correctly passed to the error handler now.

I hope I am using the error handler as intended, took me a bit to arrive at the current state :)

Let me know what you think and if you have any questions.

gunnarmorling:
Rebased to latest master and force-pushed.

Sanne:
merged! thanks

PR #1034, opened 2016-03-09
HSEARCH-2108 Elasticsearch analyzers v2

DavideD:
https://github.com/hibernate/hibernate-search/pull/1017 Follows up

I've tried to follow @gunnarmorling suggestion, this patch will add a new `AnalyzerReference` class to keep track of analyzers, locals or remotes. A remote analyzer is represented using the class `RemoteAnalyzer`.

It seems to work with custom-analyzers defined on Elasticsearch and the built-in one.


engine/src/main/java/org/hibernate/search/analyzer/impl/RemoteAnalyzerProvider.java
```diff
@@ -0,0 +1,18 @@
...
+package org.hibernate.search.analyzer.impl;
+
+/**
+ * Applies to index managers that are bound to backends supporting built-in analyzer definitions.
+ * <p>
+ * For example, Elasticsearch comes with whitespace and languages analyzers without a definition in the classpath.
+ *
+ * @author Davide D'Alto
+ */
+public interface RemoteAnalyzerProvider {
```
gunnarmorling: How about adding a method

    AnalyzerReference getRemoteAnalyzer(String name);

For now this would just return a new instance without any validation, but as of ES 2.3 (or whenever that bug is fixed), it can be enriched to actually check the given name is valid and raise an exception otherwise.

engine/src/main/java/org/hibernate/search/util/impl/ScopedAnalyzer.java
```diff
@@ -24,26 +27,28 @@
  */
 public final class ScopedAnalyzer extends AnalyzerWrapper {
```
gunnarmorling: As this class actually *is* a Lucene `Analyzer`, I think the reference should be resolved at this boundary. I.e. this class should ideally remain unchanged, and callers make sure they only invoke it when dealing with Lucene-backed `Analyzer` references.

engine/src/main/java/org/hibernate/search/analyzer/impl/AnalyzerReference.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.lucene.analysis.Analyzer;
+
+/**
+ * Reference an analayzer implementation.
+ * <p>
+ * It can be a reference to an {@link Analyzer} or a {@link RemoteAnalyzer}.
+ *
+ * @author Davide D'Alto
+ */
+public class AnalyzerReference {
```
gunnarmorling: I am wondering why you chose this delegation approach instead of an interface and two implementations. That's more what I head in mind:

    interface AnalyzerReference {
        String getName(); // if that's needed here at all?
        void close();
    }

    class LuceneAnalyzerReference implements AnalyzerReference {
        private Analyzer luceneAnalyzer;

        Analyzer getAnalyzer() { ... }

        ...
    }

    // Could live in the ES module, returned by the provider impl there
    class RemoteAnalyzerReference implements AnalyzerReference {
        ...
    }

The advantage being, that a given reference either can be one *or* the other, this becomes more apparent by having a dedicated type for each. Also you can work with the specific implementation type if you are sure of the type in a given context.

DavideD:
> interface AnalyzerReference {
>     String getName(); // if that's needed here at all?
>     void close();
> }
> 

Analyzers can be lazy initialized, at the moment this is done using the `DelegateNamedAnalyzer`.
My plan was to assume that a reference to an analyzer can be lazy and therefore remove the need for `DelegateNamedAnalyzer` and therefore an additional wrapper.

As for the other remarks, I'll try to apply them.

elasticsearch/src/test/java/org/hibernate/search/backend/elasticsearch/test/ElasticsearchAnalyzerIT.java
```diff
@@ -0,0 +1,184 @@
...
+import org.hibernate.search.FullTextSession;
+import org.hibernate.search.Search;
+import org.hibernate.search.annotations.Analyzer;
+import org.hibernate.search.annotations.Field;
+import org.hibernate.search.annotations.Indexed;
+import org.hibernate.search.test.SearchTestBase;
+import org.junit.Test;
+
+/**
+ * Test the the use of ELasticsearch built-in and custom analyzers.
```
gsmet: s/ELasticsearch/Elasticsearch/

gsmet: s/the the/the/

engine/src/main/java/org/hibernate/search/analyzer/impl/AnalyzerReference.java
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.analyzer.impl;
+
+import org.apache.lucene.analysis.Analyzer;
+
+/**
+ * Reference an analayzer implementation.
```
gsmet: s/analayzer/analyzer/

engine/src/main/java/org/hibernate/search/analyzer/impl/AnalyzerReference.java
gsmet: I think we might as well keep the RemoteAnalyzerReference in the core module. We're going to need it for Solr (or any other search engine) implementation if we do it one day.

gunnarmorling: Ok, what I'd find more important is pushing instantiation of specific
reference types to the chosen backend instead of just using the provider
interface as some marker and acting based on that in core.

gsmet: I agree with Gunnar. I think the pattern you use is more error prone as even if you have to use a RemoteAnalyzer, you can also have an analyzer set up. See my comments on ConnectedMultiFieldsPhraseQueryBuilder below.

engine/src/main/java/org/hibernate/search/query/dsl/impl/ConnectedMultiFieldsPhraseQueryBuilder.java
```diff
@@ -74,7 +72,7 @@ public Query createQuery(FieldContext fieldContext) {
 		final String sentence = phraseContext.getSentence();
```
gsmet: Mmmh, the logic below looks rather convoluted.

Why don't you simply do something like that:
```java
if ( queryContext.getQueryAnalyzer().isRemote() ) {
     return new RemotePhraseQuery(sentence);
}
```
?
with RemotePhraseQuery simply storing the sentence to be reused in ToElasticSearch.

This would simplify the logic in ToElasticSearch too.

engine/src/main/java/org/hibernate/search/query/dsl/impl/ConnectedMultiFieldsTermQueryBuilder.java
```diff
@@ -190,7 +192,15 @@ private Query createTermQuery(FieldContext fieldContext, String term) {
...
 	}
 
-	private List<String> getAllTermsFromText(String fieldName, String localText, Analyzer analyzer) {
+	private List<String> getAllTermsFromText(String fieldName, String localText, AnalyzerReference analyzerReference) {
+		Analyzer analyzer = null;
+		if ( analyzerReference.getAnalyzer() != null ) {
+			analyzer = analyzerReference.getAnalyzer();
+		}
+		else if ( analyzerReference.getRemote() != null ) {
+			return Arrays.asList( localText );
```
gsmet: This won't work. If someone is looking for "several words", it's going to create a term query with term "several words" which won't match anything as "several words" has probably been tokenized and separated in several terms by the analyzer configured on Elasticsearch. Unless you count on having defined an analyzer locally to tokenize the text.

I think for the EXACT case, we can probably use a multi_match query which is the best fit IMHO.

As for the FUZZY case, a multi_match with fuzziness should do the trick.

For wildcard queries, it's a bit more complicated as it's not that easy to find something equivalent in Elasticsearch. And, moreover both the following cases make perfect sense: the one where you want the words tokenized (typically for a search ) or where you don't want them tokenized (especially useful for autocompletion). So automatically tokenizing the search terms is not a solution either. The query string query allows too much features and you can't disable any of them. The simple query string query only allows for prefix queries, not wildcards. I think our only way to make this feature work in the same way with Elasticsearch is to call the _analyze API in this case and exploit the tokens returned.

Thoughts?

(I don't think this should delay this patch though)

DavideD: > The simple query string query only allows for prefix queries, not wildcards. I think our only way to make this feature work in the same way with Elasticsearch is to call the _analyze API in this case and exploit the tokens returned.
> 
> Thoughts?

 It make sense, I'll create a seprate issue for this

PR #1036, opened 2016-03-10
HSEARCH-2175 Enabling more tests which pass now

gunnarmorling:


Sanne:
Running some more tests, but it seems reliable now. My guess is I was still affected by the other stability issues which are fixed now.

Sanne:
merged!

PR #1039, opened 2016-03-11
HSEARCH-2108 Define analyzer in Elasticsearch for indexing 

DavideD:

I've applied the remarks on the previous pull request: https://github.com/hibernate/hibernate-search/pull/1034

I've added an `unwrap` method to check if an `AnalyzerReference` is of the type the backend is expecting.

I've noticed that when using the DSL the analyzer is applied anyway for match phrase queries, is that a bug?


gsmet:
Hi @DavideD 

> I've noticed that when using the DSL the analyzer is applied anyway for match phrase queries, is that a bug?

Not sure I understand you on this?

DavideD:
@gsmet  What I meant is that the analyzer is applied even if one specifies `ignoreAnalyzer()`:
```
Query query = queryBuilder
					.phrase()
						.onField( "message" )
							.ignoreAnalyzer()
...
					.createQuery();
```

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -263,6 +266,14 @@ private void createIndexMappings() {
 		}
 	}
 
+	private String analyzerName(DocumentFieldMetadata fieldMetadata) {
+		AnalyzerReference analyzer = fieldMetadata.getAnalyzer();
```
gunnarmorling: Should we raise an exception if it's an actual Lucene analyzer? I'd say let's at least raise a warning that's going to take no effect.

Sanne:
I've asked @DavideD to add a comment in _initRemoteAnalyzer_ as the various IndexManagers are not all known in advance, so the code is assuming that we don't have a "mixed" use case of some entities using ES and some other entities using Lucene.

This limitation also applies to validation, it's tricky at  this point to figure that an analyzer definition which is valid in the context of ES is not being (wrongly) used in the Lucene embedded case.

elasticsearch/src/test/java/org/hibernate/search/backend/elasticsearch/test/ElasticsearchDSLIT.java
```diff
@@ -0,0 +1,144 @@
...
+
+			Query query = queryBuilder
+					.phrase()
+						.onField( "message" )
+						.sentence( "A very important matter" )
+					.createQuery();
+
+			FullTextQuery fullTextQuery = fullTextSession.createFullTextQuery( query, Letter.class );
+			String queryString = fullTextQuery.getQueryString();
+			assertThat( queryString ).isEqualTo( "{\"query\":{\"match_phrase\":{\"message\":{\"query\":\"very important matter\",\"slop\":0,\"boost\":1.0}}}}" );
```
gunnarmorling: No need to change right now, but it's still good to know: We have `JsonHelper#assertJsonEquals()`  which allows you to use _'_ instead of the escaped _"_, making it a tad nicer to read.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
DavideD: Ok

engine/src/main/java/org/hibernate/search/engine/metadata/impl/TypeMetadata.java
```diff
@@ -427,13 +428,13 @@ public boolean isJpaIdUsedAsDocumentId() {
...
-			this( indexedType, new ScopedAnalyzer( configContext.getDefaultAnalyzer() ) );
+			this( indexedType, new ScopedAnalyzerReference( configContext.getDefaultAnalyzer() ) );
 		}
 
-		public Builder(Class<?> indexedType, ScopedAnalyzer scopedAnalyzer) {
+		public Builder(Class<?> indexedType, ScopedAnalyzerReference scopedAnalyzer) {
 			this.indexedType = indexedType;
 			this.scopedAnalyzer = scopedAnalyzer;
-			this.passThroughAnalyzer = PassThroughAnalyzer.INSTANCE;
+			this.passThroughAnalyzer = new LuceneAnalyzerReference( PassThroughAnalyzer.INSTANCE );
```
gunnarmorling: I suppose we could even have a pass-through singleton instance of `LuceneAnalyzerReference`?

elasticsearch/src/test/java/org/hibernate/search/backend/elasticsearch/test/ElasticsearchDSLIT.java
DavideD: I will update it, it is simple enough

engine/src/main/java/org/hibernate/search/util/impl/ScopedAnalyzerReference.java
```diff
@@ -0,0 +1,121 @@
...
+import org.hibernate.search.analyzer.impl.LuceneAnalyzerReference;
+
+/**
+ * A wrapper containing all the analyzers for a given class.
+ *
+ * @author Davide D'Alto
+ */
+public final class ScopedAnalyzerReference implements AnalyzerReference {
+
+	private final Map<String, AnalyzerReference> analyzers = new HashMap<>();
```
gunnarmorling: That's never shared between threads, right? So we don't need to ensure safe publication and the like?

gsmet:
@DavideD it could be considered a bug but I don't think a phrase query makes sense if you're not going to tokenize it. And for the time being, the tokenizer is tied to the analyzer. That's something I would like to consider changing in the future.

engine/src/main/java/org/hibernate/search/util/impl/ScopedAnalyzerReference.java
Sanne: That's exactly the same that I was referring to when mentioning thread dodginess on HipChat ;-)
It happens to be used by different threads but it seems fine for the specific use case. Definitely something to improve on, but also something that Davide just inherited from the original code here.

DavideD:
> @DavideD it could be considered a bug but I don't think a phrase query makes sense if you're not going to tokenize it. And for the time being, the tokenizer is tied to the analyzer. That's something I would like to consider changing in the future.

Yes, no need to change it now. I've open an issue for it: https://hibernate.atlassian.net/browse/HSEARCH-2185

DavideD:
Applied the remarks and rebased

Sanne:
it looks good. Are you going to squash the correction commits?

DavideD:
OK, I'll squashed them

DavideD:
Squashed

gunnarmorling:
Going to close this one, I've sent #1043 as a replacement which adds one more commit. Thanks, @DavideD!

PR #1040, opened 2016-03-12
HSEARCH-2134 Elasticsearch distribution packages

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-2134

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -59,6 +59,10 @@
         </dependency>
         <dependency>
             <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-search-backend-elasticsearch</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.hibernate</groupId>
             <artifactId>hibernate-search-backend-jms</artifactId>
         </dependency>
         <dependency>
--- a/distribution/src/main/assembly/dist.xml
+++ b/distribution/src/main/assembly/dist.xml
@@ -109,7 +109,7 @@
         </dependencySet>
 
         <dependencySet>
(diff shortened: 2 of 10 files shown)


DavideD:
It seems that this test does not pass anymore now: MassIndexingTimeoutIT

Any idea what could cause the problem?

DavideD:
I'm closing this for now, I want to investigate it a bit more.
It seems that adding module in wildfly and making it an optional dependency to search engine is enough to make this test fails, even if it does not use elasticsearch

DavideD:
Added a commit to change the distribution folder

DavideD:
By the way, does the integration test fail for you as well?

Sanne:
> By the way, does the integration test fail for you as well?

Yes, but couldn't debug it yet.

Sanne:
The test is failing because `org.hibernate.search.backend.elasticsearch.impl.ElasticsearchBridgeProvider` is overriding the Bridge, even though we're not using Elasticsearch in this case.

I'll disable this test as this PR is otherwise good, and will raise the priority for
 - https://hibernate.atlassian.net/browse/HSEARCH-2067 cc/ @gsmet @gunnarmorling

Sanne:
Well HSEARCH-2067 was "Blocker" already so that's not changing ;-)

Sanne:
> Sounds good, shouldn't we also change the name of the artifact as well, then?

I think so, but ask the mailing list so we can get a quorum. We can do it as a different issue.

Sanne:
merged with an additional commit to disable the failing IT tests

PR #1042, opened 2016-03-14
HSEARCH-1970 rebased

gunnarmorling:
Supersedes #896.

I rebased that work to ease reviewing. @Sanne, anything else I can do to have is finalize this (in one way or the other)? I'm feeling frustrated that this change which I deem very useful sits around here for more than six months without much progress. Were I the regular contributor, I'd be gone long time ago.

Please see the original PR for the motivation and high-level description of this patch. In a nutshell, it allows to remove several methods from the `BackendQueueProcessor` and `IndexManager` contracts which essentially only were there because of the circular dependency between the two concepts. Breaking up this circle also also allows to avoid some hacks in `BackendFactory`.


backends/jms/src/main/java/org/hibernate/search/backend/impl/jms/AbstractJMSHibernateSearchController.java
```diff
@@ -69,8 +70,16 @@ public void onMessage(Message message) {
 				log.messageReceivedForUndefinedIndex( indexName );
 				return;
 			}
-			queue = indexManager.getSerializer().toLuceneWorks( (byte[]) objectMessage.getObject() );
-			indexManager.performOperations( queue, null );
+
+			LuceneWorkSerializer luceneWorkSerializer = integrator.getServiceManager().requestService( LuceneWorkSerializer.class );
```
Sanne: Do you think you can avoid this change?

Sanne: One of my concerns is that this is a change that inpacts several integration points. I'm aware we're ok to change SPIs, but not all SPI are equal.. 

But I see your point in wanting to decouple it. Let's say we agree on moving the reference to `LuceneWorkSerializer`, I think what we need is:

    integrator.getLuceneWorkSerializer();

I would still keep the IndexManager#getSerializer() method around to keep such users happy, and deprecate it. (It would be implemented by simply delegating to  `integrator.getLuceneWorkSerializer()`

Sanne: P.S. one reason for the current `DirectoryBasedIndexManager` to cache the Serializer in a field is that we really need to for performance reasons. Serialization / deserialization of streams of operations is very hot for Infinispan, and I guess can be problematic for JMS too.

Sanne:
So, my high level comments:
 - I like the spirit of 7cb47ad : great cleanup. Let's apply it, but there are consequences on how we hope in future to provide "configuration sets" which make sense. Definitely need a plan for docs and architecture rewrite. I still think I wanted this the other way around when proposing `IndexManager`, but stuff evolved differently and your change fixes it.. too bad the names no longer match what I wanted but you're right it's a clear improvement.
 - 38fbf1a +1
 - 33672d3 Moving the lookup to LuceneWorkSerializer: I'd rather not do this now. Good ideas, but I'd take it even a step further, and yet not break backwards compatibility on this one yet.
 - 5815c27 +1

Teasing: at an high level, can you explain now why `ElasticsearchIndexManager` should be an `IndexManager` ? 

We'll need to rebase again, at least the good parts. Sorry for that, let me know if I can help with that.

Also: please please always keep a separate commit for renaming or moving classes. Let's also see if we can avoid renaming some.

backends/jms/src/main/java/org/hibernate/search/backend/impl/jms/AbstractJMSHibernateSearchController.java
gunnarmorling: The thing is, with the way I envision pieces working together, there is just no need for an index manager to deal with the serializer at all. What I had in mind is this:

* Let index managers deal with actual indexing work: talk to Lucene, talk to ES etc.
* Let backend queue processors deal with propagating work: apply it locally (by going to the IM straight away), or distribute in in the cluster (by sending over JMS/JGroups). It's that latter part where the serializer only is needed. Of course a specific BQP could cache the serializer if it benefits from it.

So bottom line is: One could keep that method, but it's the very nature of the proposed change to remove it as it doesn't really make sense in that proposed design. That said, one could just keep the method on the contract temporarily for the sake of not breaking existing impls.

Sanne: > That said, one could just keep the method on the contract temporarily for the sake of not breaking existing impls.

Yes that's what I'm suggesting, as I mentioned to deprecate it.

I understand and like the long term vision.

gunnarmorling: Ok, that sounds great. I'll rework it then, adding back that method to the {{IndexManager}} contract, it would just not be used really.

gunnarmorling:
> Teasing: at an high level, can you explain now why ElasticsearchIndexManager should be an IndexManager? 

`IndexManager` would still be the unified contract for dealing with an actual indexing backend. E.g. for the JMS-based BQP implementation, on the side that processes messages from the JMS queue, it'd talk in a unified way to the indexing backend, be it embedded Lucene or ES. We would get rid of the `ElasticsearchBackendQueueProcessor` though, as it would become obsolete.

backends/jms/src/main/java/org/hibernate/search/backend/impl/jms/AbstractJMSHibernateSearchController.java
Sanne: You could use it to keep e.g. the JMS implementations untouched.

Your call, I'm just preferring smaller patches and have a bit of concern about the performance of those serializer lookups.

gunnarmorling:
Superseded by #1099.

PR #1045, opened 2016-03-15
HSEARCH-2108 Analyzer should be ignore if the field is not analyzed

DavideD:
This should fix the problem about the warning log Chris pointed out before.

The problem is caused by checking the analyzer even if the field is not analyzed.
It also add a couple of additional tests.


Sanne:
merged!

PR #1046, opened 2016-03-16
HSEARCH-2188 Fixed type checking in NumericFieldUtilsTest

gfouquet:
`isAssignableFrom` in `NumericFieldUtils` was used the wrong way. I turned it around and added a unit test, which did not pass with the previous implementation.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/bridge/util/impl/NumericFieldUtils.java
+++ b/engine/src/main/java/org/hibernate/search/bridge/util/impl/NumericFieldUtils.java
@@ -9,10 +9,11 @@ package org.hibernate.search.bridge.util.impl;
 import java.util.Calendar;
 import java.util.Date;
 
-import org.apache.lucene.search.NumericRangeQuery;
-import org.apache.lucene.search.Query;
 import org.hibernate.search.util.logging.impl.Log;
 import org.hibernate.search.util.logging.impl.LoggerFactory;
+
+import org.apache.lucene.search.NumericRangeQuery;
+import org.apache.lucene.search.Query;
 
 /**
  * Utility class to handle numeric fields.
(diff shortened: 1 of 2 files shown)


Sanne:
ok to test

Sanne:
thanks @gfouquet ! I'll gladly apply your patches, but could you please sign the Hibernate contribution agreement?
It's as simple as creating an account, reading the terms and clicking on a checkbox: https://cla.jboss.org

gfouquet:
Hi, 
I just signed the CLA

engine/src/main/java/org/hibernate/search/bridge/util/impl/NumericFieldUtils.java
```diff
@@ -99,11 +100,10 @@ public static boolean requiresNumericRangeQuery(Object value) {
...
-		return numericClass.isAssignableFrom( Double.class ) ||
-				numericClass.isAssignableFrom( Long.class ) ||
-				numericClass.isAssignableFrom( Integer.class ) ||
-				numericClass.isAssignableFrom( Float.class ) ||
-				numericClass.isAssignableFrom( Calendar.class );
+		return value instanceof Double ||
+				value instanceof Long ||
+				value instanceof Integer ||
+				value instanceof Float ||
+				value instanceof Calendar;
```
gsmet: Hmmm, I'm surprised this is not consistent with the code above. Shouldn't we also have Byte, Date and Short here?

s4ke: Sounds reasonable to me.

gunnarmorling:
Where are we with this one?

gfouquet:
I'm willing to make any change the Hibernate team see fit. On the other hand I'm mostly clueless about the side effects of any change.

gsmet:
@gunnarmorling it's already fixed in master for a while and I extracted the test case and the additional cleanup from Gregory's PR and pushed it to master: https://github.com/hibernate/hibernate-search/commit/a5ad4f586c713df198beb797a9f4f5f9237cd3b7

I asked @Sanne if we still maintain 5.3 and 5.4 but I haven't had a reply so far.

If we do, I'll prefer to only cherrypick Davide's commit fixing this: https://github.com/hibernate/hibernate-search/commit/4c0415c1f11b7e26fdc4dc95f4b822bf9e983139 . No need for the further cleanup in the old branches.

gsmet:
@gfouquet I will backport Davide's fix in 5.3 and 5.4. That being said, we haven't planned to release new versions of these 2 branches.

gsmet:
OK, so I pushed the test and the cleanup to master a few weeks ago and backported the original Davide's fix to 5.3 and 5.4 just now.

Closing. Thanks!

PR #1049, opened 2016-03-30
HSEARCH-2197 Loading of search results produces useless garbage.

golovnin:
see https://hibernate.atlassian.net/browse/HSEARCH-2197


Sanne:
great catch! merged

PR #1050, opened 2016-04-01
HSEARCH-2198 Avoid using maven-injection-plugin to retrieve project V…

Sanne:
…ersion

 - https://hibernate.atlassian.net/browse/HSEARCH-2198


s4ke:
Looks great. I like using the standard way of doing things instead of the maven plugin really much. merged it.

Sanne:
thanks!

PR #1051, opened 2016-04-04
HSEARCH-2202 Fix SearchIntegrator camel casing throughout the codebase

gsmet:


Sanne:
Looks good, feel free to integrate if Jenkins is happy.

N.B. don't do such things if it affects an API or an SPI as several other projects rely on our SPI to be stable, and our tests won't highlight what kind of pain we cause; so generally be careful with such changes: they are not as cheap as they look.

gsmet:
Yes, I was careful about it not changing API and SPI.

(I noticed this while working on another patch)

gsmet:
Merged!

PR #1053, opened 2016-04-04
HSEARCH-2203 Reduce memory consumption of EntityInfoImpl objects.

golovnin:
see https://hibernate.atlassian.net/browse/HSEARCH-2203

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchHSQueryImpl.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchHSQueryImpl.java
@@ -541,9 +541,10 @@ public class ElasticsearchHSQueryImpl extends AbstractHSQuery {
 							}
 							break;
 						case ProjectionConstants.THIS:
-							// Use THIS as placeholder. It will be replaced
-							// when we populate the EntityInfo with the real entity.
-							projections[i] = ProjectionConstants.THIS;
+							// Use EntityInfo.ENTITY_PLACEHOLDER as placeholder.
+							// It will be replaced when we populate
+							// the EntityInfo with the real entity.
+							projections[i] = EntityInfo.ENTITY_PLACEHOLDER;
 							break;
 						default:
 							projections[i] = getFieldValue( binding, hit, field );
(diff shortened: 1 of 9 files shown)


elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchHSQueryImpl.java
```diff
@@ -540,23 +541,17 @@ EntityInfo convertQueryHit(JsonObject hit) {
 							}
 							break;
 						case ProjectionConstants.THIS:
-							indexesOfThis.add( i );
+							// Use THIS as placeholder. It will be replaced
+							// when we populate the EntityInfo with the real entity.
+							projections[i] = ProjectionConstants.THIS;
```
Sanne: Rather than reusing the String from `ProjectionConstants.THIS`, I was considering using a marker object.

Would you prefer to create a (static final) singleton having just a nice toString() implementation so it doesn't get too confusing during debug?
You would still use reference equality to check for the marker, but it makes me more comfortable than doing it on a string.

engine/src/main/java/org/hibernate/search/query/engine/impl/EntityInfoImpl.java
```diff
@@ -62,31 +66,21 @@ public String getIdName() {
...
-		return indexesOfThis;
-	}
-
-	@Override
-	public boolean isProjectThis() {
-		return indexesOfThis.size() != 0;
+	public Object getEntityInstance() {
+		for ( Object obj : projection ) {
+			if ( clazz.isInstance( obj ) ) {
+				return obj;
```
Sanne: I think I'd rather store the instance as a field than to have to loop through the projections with instanceof attempts?

You can easily store the EntityInstance during `populateWithEntityInstance`.

Sanne:
Thanks @golovnin , looks great! I made some minor comments, not sure if we should change them. Up to you, so I'll keep this open for some more days.

Seems like you're pushing performance requirements hard? Would be nice to hear more about what you're doing and aiming at.

gunnarmorling:
So does this loose the ability "this" more than once? I doubt anyone would be doing it, but it was allowed before, so we'd at least need an entry in the migration notes.

Sanne:
> So does this loose the ability "this" more than once? 

Why do you say that? I didn't notice.

gunnarmorling:
Yeah, seems alright. The removal of `getIndexesOfThis()` made me wonder, but it's handled as part of that array now.

golovnin:
@Sanne I changed the code as you suggested. I hope that the toString()-method of EntityInfo.ENTITY_PLACEHOLDER is nice enough. :-)

Sanne:
Thanks! I integrated it.

PR #1057, opened 2016-04-11
HSEARCH-2210 Add missing types to NumericFieldUtils#requiresNumericRangeQuery

gsmet:


engine/src/main/java/org/hibernate/search/bridge/util/impl/NumericFieldUtils.java
```diff
@@ -118,9 +118,12 @@ public static boolean requiresNumericRangeQuery(Object value) {
 			return false;
 		}
 		return value instanceof Double ||
+				value instanceof Byte ||
+				value instanceof Short ||
 				value instanceof Long ||
 				value instanceof Integer ||
 				value instanceof Float ||
+				value instanceof Date ||
```
gunnarmorling: Regarding `Date` and `Calendar`, we'd have to take the backend into account as well, in order to a account for a different mapping in ES. Should we track this as a separate issue?

engine/src/test/java/org/hibernate/search/test/bridge/util/NumericFieldUtilsTest.java
```diff
@@ -4,7 +4,7 @@
  * License: GNU Lesser General Public License (LGPL), version 2.1 or later
  * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
  */
-package org.hibernate.search.test.bridge.util.impl;
```
gunnarmorling: How comes this changed?

gsmet: It wasn't consistent with the other tests, that's why I changed it.

engine/src/main/java/org/hibernate/search/bridge/util/impl/NumericFieldUtils.java
gsmet: Not really. This method is only used with Infinispan (when I did it, I wondered how it could have already been working with Es). See `ConnectedMultiFieldsRangeQueryBuilder` and the comment "//we need to guess the proper type from the parameter types (Fallback logic required by the protobuf integration in Infinispan)".

That being said, it might be worth it to add a comment here?

gunnarmorling: Ok, that sounds a bit scary :) Yes, adding some comment would be good. The method appears to be very general-purpose, so it was unexpected it's only used with ISPN.

gsmet: Done!

gunnarmorling:

PR #1060, opened 2016-04-12
HSEARCH-2217 Move Travis build to Trusty infrastructure and use Maven wrapper

gsmet:
With this commit, we should be able to enable Travis on the main repo with all the integration tests running!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,12 @@
+sudo: required
+dist: trusty
 language: java
 jdk:
   - oraclejdk8
 addons:
-  postgresql: '9.4'
-  mariadb: '10.1'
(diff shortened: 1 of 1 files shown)


gunnarmorling:
Looks good as far as I can judge; Out of interest, why is it that the Maven wrapper is needed?

gsmet:
The Maven version provided by Travis is outdated and it doesn't respect our enforcer rules. The wrapper allows to use the latest version of Maven without the hassle to install it. I didn't know about it but found that in related Travis issues and found it convenient.

gunnarmorling:
Cool, makes sense. Can you add a quick comment on that and push yourself? Thx!

PR #1062, opened 2016-04-13
HSEARCH-2067 Introduce the notion of backend specific FieldBridges

gsmet:
So here is the work I did for HSEARCH-2067.

It introduces the notion of backend specific FieldBridges which are only enabled if a type is managed by a given backend.

There are a couple of shortcomings we should probably discuss:

## Building TypeMetadata

As the initialization of FieldBridges now depends on the backend which manages a type, we need to have this information to build them. We can have it when we build the TypeMetadata for an indexed type as it is attached to a backend but we don't have it when we build the TypeMetadata for ContainedIn types as they are not attached to a specific backend (and might even be used by indexed entities managed by 2 different backends).

At first, I tried to implement a ContainedInTypeMetadata with only the relevant information (and I hoped there were only a few of them) but we really need most of them for the dirty checking and optimizations so I ended up implementing a skipFieldBridges method in ParseContext (see AnnotationMetadataProvider).

## Change in SPI

I had to change the IndexManagerFactory SPI to make it a 2 steps process. I'm pretty sure it's going to be subject to discussions (Sanne?).

Maybe it's the good time to also add a method to be able to register new aliases?

### DIFF
--- a/engine/src/main/java/org/hibernate/search/cfg/spi/IndexManagerFactory.java
+++ b/engine/src/main/java/org/hibernate/search/cfg/spi/IndexManagerFactory.java
@@ -28,14 +28,14 @@ public interface IndexManagerFactory extends Service {
 	 *
 	 * @return the type chosen
 	 */
-	Class<? extends IndexManager> determineIndexManagerImpl(Class<?> mappedClass, String indexManagerImplementationName);
+	Class<? extends IndexManager> determineIndexManagerType(Class<?> mappedClass, String indexManagerImplementationName);
 
 	/**
 	 * @param mappedClass the mapped entity class
-	 * @param indexManagerImpl a given {@code IndexManager} type
+	 * @param indexManagerType a given {@code IndexManager} type
 	 *
 	 * @return a new {@code IndexManager} instance of the given type
 	 */
-	IndexManager createIndexManager(Class<?> mappedClass, Class<? extends IndexManager> indexManagerImpl);
+	IndexManager createIndexManager(Class<?> mappedClass, Class<? extends IndexManager> indexManagerType);
 
 }
--- a/engine/src/main/java/org/hibernate/search/cfg/spi/IndexManagerFactory.java
+++ b/engine/src/main/java/org/hibernate/search/cfg/spi/IndexManagerFactory.java
@@ -21,21 +21,17 @@ public interface IndexManagerFactory extends Service {
 
 	/**
 	 * Determine the {@code IndexManager} implementation which will be used for this entity type.
-	 *
-	 * @param mappedClass the mapped entity class
 	 * @param indexManagerImplementationName how this is resolved to an {@code IndexManager} type
 	 * is left to the implementor.
 	 *
 	 * @return the type chosen
 	 */
-	Class<? extends IndexManager> determineIndexManagerType(Class<?> mappedClass, String indexManagerImplementationName);
+	Class<? extends IndexManager> determineIndexManagerType(String indexManagerImplementationName);
 
 	/**
-	 * @param mappedClass the mapped entity class
 	 * @param indexManagerType a given {@code IndexManager} type
-	 *
 	 * @return a new {@code IndexManager} instance of the given type
 	 */
-	IndexManager createIndexManager(Class<?> mappedClass, Class<? extends IndexManager> indexManagerType);
+	IndexManager createIndexManager(Class<? extends IndexManager> indexManagerType);
 
 }
--- a/engine/src/main/java/org/hibernate/search/cfg/spi/IndexManagerFactory.java
+++ b/engine/src/main/java/org/hibernate/search/cfg/spi/IndexManagerFactory.java
@@ -15,20 +15,27 @@ import org.hibernate.search.indexes.spi.IndexManager;
  * define new short-hand aliases, change the default implementation.
  *
  * @author Sanne Grinovero (C) 2012 Red Hat Inc.
+ * @author Guillaume Smet
  */
 public interface IndexManagerFactory extends Service {
 
 	/**
-	 * @return a new instance of the default IndexManager
-	 */
-	IndexManager createDefaultIndexManager();
-
-	/**
-	 * @param indexManagerImplementationName how this is resolved to an IndexManager type
+	 * Determine the {@code IndexManager} implementation which will be used for this entity type.
+	 *
+	 * @param mappedClass the mapped entity class
+	 * @param indexManagerImplementationName how this is resolved to an {@code IndexManager} type
 	 * is left to the implementor.
 	 *
-	 * @return a new IndexManager instance of the chosen type
+	 * @return the type chosen
 	 */
-	IndexManager createIndexManagerByName(String indexManagerImplementationName);
+	Class<? extends IndexManager> determineIndexManagerImpl(Class<?> mappedClass, String indexManagerImplementationName);
+
+	/**
+	 * @param mappedClass the mapped entity class
+	 * @param indexManagerImpl a given {@code IndexManager} type
+	 *
+	 * @return a new {@code IndexManager} instance of the given type
+	 */
+	IndexManager createIndexManager(Class<?> mappedClass, Class<? extends IndexManager> indexManagerImpl);
 
 }
--- a/engine/src/main/java/org/hibernate/search/engine/impl/DefaultIndexManagerFactory.java
+++ b/engine/src/main/java/org/hibernate/search/engine/impl/DefaultIndexManagerFactory.java
@@ -6,9 +6,7 @@
  */
 package org.hibernate.search.engine.impl;
 
-import java.util.Map;
 import java.util.Properties;
-import java.util.concurrent.ConcurrentHashMap;
 
 import org.hibernate.search.cfg.spi.IndexManagerFactory;
 import org.hibernate.search.engine.service.classloading.spi.ClassLoaderService;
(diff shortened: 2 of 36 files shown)


engine/src/main/java/org/hibernate/search/engine/spi/EntityIndexBinding.java
```diff
@@ -50,6 +50,16 @@
 	void postInitialize(Set<Class<?>> indexedClasses);
 
 	/**
+	 * @return the {@code IndexManager} impl used by this entity type
+	 */
+	Class<? extends IndexManager> getIndexManagerImpl();
```
DavideD: What's the difference between this method and `IndexManager[] getIndexManagers();`?
Shouldn't this be called `getIndexManager()`?

orm/src/test/java/org/hibernate/search/test/embedded/EmbeddedEntityNotIndexedTest.java
```diff
@@ -6,46 +6,29 @@
...
-	@Before
-	public void setUp() {
-		SearchConfiguration searchConfiguration = new SearchConfigurationForTest();
-		ConfigContext configContext = new ConfigContext(
-				searchConfiguration,
-				new BuildContextForTest( searchConfiguration )
-		);
-		metadataProvider = new AnnotationMetadataProvider( new JavaReflectionManager(), configContext );
-	}
+public class EmbeddedEntityNotIndexedTest extends AbstractAnnotationMetadataTest {
```
DavideD: Are all these refactoring releated to this issue?

engine/src/main/java/org/hibernate/search/engine/spi/EntityIndexBinding.java
gsmet: Yes, that's what I used first.

But the fact is that with dynamic sharding strategy, it does not work as `getIndexManagers()` starts the index managers and it is too soon to create the index managers in this case. See for instance `IndexManagerHolder#buildEntityIndexBinding` where there is a special case for dynamic sharding to avoid starting the index managers.

That's why I use the class and that's why the name of the method is `getIndexManagerImpl`.

orm/src/test/java/org/hibernate/search/test/embedded/EmbeddedEntityNotIndexedTest.java
gsmet: Yes, the initialization of the AnnotationMetadataProvider is more complicated now as we need a MutableSearchFactoryState so I made sure it was done only once.

engine/src/main/java/org/hibernate/search/engine/spi/EntityIndexBinding.java
DavideD: What about `getIndexManagerType()` instead?

engine/src/main/java/org/hibernate/search/cfg/spi/IndexManagerFactory.java
```diff
@@ -15,20 +15,27 @@
...
+	 * Determine the {@code IndexManager} implementation which will be used for this entity type.
+	 *
+	 * @param mappedClass the mapped entity class
+	 * @param indexManagerImplementationName how this is resolved to an {@code IndexManager} type
+	 * is left to the implementor.
+	 *
+	 * @return the type chosen
 	 */
-	IndexManager createDefaultIndexManager();
+	Class<? extends IndexManager> determineIndexManagerImpl(Class<?> mappedClass, String indexManagerImplementationName);
```
gunnarmorling: How comes the mapped class is needed here, too? It seems to be used for caching, but couldn't this happen using the implementation name as a key? Same for the method below, the implementation I could find does not use the `mappedClass` at all.

gunnarmorling: Naming-wise, I'd rather use the "Type" fix over "Impl" in cases where a class object is returned: `determineIndexManagerType()` (applies here and in some other places).

engine/src/main/java/org/hibernate/search/engine/impl/DefaultIndexManagerFactory.java
```diff
@@ -28,47 +29,79 @@
...
+		return DirectoryBasedIndexManager.class;
+	}
+
+	/**
+	 * Add a mapping between an entity type and the {@code IndexManager} managing this entity type.
+	 *
+	 * @param mappedClass the type of the entity
+	 * @param indexManager the {@code IndexManager} managing this entity
+	 */
+	protected void addMapping(Class<?> mappedClass, Class<? extends IndexManager> indexManager) {
```
gunnarmorling: indexManagerType

engine/src/main/java/org/hibernate/search/engine/spi/EntityIndexBinding.java
```diff
@@ -50,6 +50,18 @@
...
+
+	/**
+	 * @param indexManagerCandidate the tested {@code IndexManager} implementation
+	 *
+	 * @return whether this entity is managed by the given {@code IndexManager} impl
+	 */
+	boolean isManagedBy(Class<? extends IndexManager> indexManagerCandidate);
```
gunnarmorling: Seems this one can be avoided by just locally implementing it in the factories call this method. Tiny duplication, but prevents adding another SPI-level method.

engine/src/main/java/org/hibernate/search/indexes/impl/IndexManagerHolder.java
```diff
@@ -237,15 +253,9 @@ private IndexManager createIndexManager(String indexName,
 		IndexManagerFactory indexManagerFactory = serviceManager.requestService( IndexManagerFactory.class );
 
 		// create IndexManager instance via the index manager factory
-		String indexManagerImplementationName = properties.getProperty( Environment.INDEX_MANAGER_IMPL_NAME );
 		final IndexManager manager;
 		try {
-			if ( StringHelper.isEmpty( indexManagerImplementationName ) ) {
```
gunnarmorling: How comes these bits have changed here? It's not fully clear to me yet whether that's strictly needed for the actual issue.

gunnarmorling:
Hey Guillaume, on first sight I am surprised by the number of changes.

Taking a step back, what we need is the index manager type of a given entity, so we can filter on the right backend-specific provider in `BridgeFactory`. Is that all that's to it?

If so, when looking at `SearchIntegratorBuilder.initDocumentBuilders`, we first create the `EntityIndexBinding` before going to the `MetadataProvider` which in turn invokes the bridge factory. So couldn't we take the type of the first IM returned by `EntityIndexBinding.getIndexManagers()` (assuming there is no mixture of IM types for one entity) and pass it to the metadata provider and from there into the bridge?

That might avoid all the changes to existing SPIs. But then it seems that simple, so I may be missing something :)

engine/src/main/java/org/hibernate/search/cfg/spi/IndexManagerFactory.java
gsmet: Yes, you're right. It's a leftover of another approach. Will fix.

OK for the name change.

engine/src/main/java/org/hibernate/search/engine/impl/DefaultIndexManagerFactory.java
gsmet: Fixed.

engine/src/main/java/org/hibernate/search/engine/spi/EntityIndexBinding.java
gsmet: Changed!

engine/src/main/java/org/hibernate/search/engine/spi/EntityIndexBinding.java
gsmet: I agree it can be avoided but I think it's something that should be available in EntityIndexBinding. That's why I added it there.

Probably better to wait for @Sanne for the SPI changes as I'm pretty sure he will have a strong opinion about them.

gsmet:
> Taking a step back, what we need is the index manager type of a given entity, so we can filter on the right backend-specific provider in BridgeFactory. Is that all that's to it?

Yes! So simple, isn't it? :)

> So couldn't we take the type of the first IM returned by EntityIndexBinding.getIndexManagers()

That's what I did in my first approach even if I remembered @Sanne had warned me about it.

This doesn't work when you use dynamic sharding. In the dynamic sharding case, it's too early to start the index managers (see the special case here https://github.com/hibernate/hibernate-search/blob/master/engine/src/main/java/org/hibernate/search/indexes/impl/IndexManagerHolder.java#L90). You can't use `EntityIndexBinding.getIndexManagers()` at this early stage and having the `IndexManager` type is the way to go.

It took me quite a lot of time to come up with this approach so I understand it might feel a bit overdesigned at first sight.

The SPI changes are discussible of course but I think the overall approach is good.

gunnarmorling:
Hey @gsmet can you check out this commit: https://github.com/hibernate/hibernate-search/commit/928d17887c85c7a1253c960f8e8a554ddd5df138

It's a quick sketch for exposing the IM type to the bridge factory which does avoid SPI changes. It doesn't compile tests and could need some caching etc. when doing it properly, it's just to show it to you quickly. Together with your new `BackendSpecificBridgeProvider` it might do the trick (btw. `ExtendedBridgeProvider` lives in "impl", so it should be possible to add the new method there?).

WDYT?

gsmet:
Closing as there is a new PR pending.

PR #1063, opened 2016-04-13
HSEARCH-2100 Cache the DocumentBuilder in the QueryBuildingContext

gsmet:


gunnarmorling:
Merged, thanks @gsmet! A world with less map look ups is a better one :)

PR #1064, opened 2016-04-15
Fix various typos noted while browsing the code

gsmet:


gunnarmorling:
Looks good, merge yourself if you like. I only don't remember our policy for micro-fixes without JIRA, @Sanne? From my side that's fine.

Sanne:
>  I only don't remember our policy for micro-fixes without JIRA

Not very well defined, and it's bothering me. So let's give it a shot:

We historically followed Steve Ebersole's example "any change needs a JIRA", which I think has the benefit of affecting changelogs and removes any form of argument among the team of the form "I think this needs a JIRA".

On the other side, I struggle with this myself. I just did some cleanup of local branches and identified some dozen of improving commits which I've never shared because they were not "significant enough" to deserve a JIRA.
To be fair, that's a lot of waste.. I should have merged these patches long ago.

So I'm inclined to say "go ahead" but please let's not take this as a precedent to not create JIRAs for things which could use one... as a rule I guess we could say that a JIRA is necessary if anyone thinks so.

My "rule of thumb" is: does recording the change in JIRA add value for those reading the JIRA generated changelogs vs does it add noise?

"Various typos" is noise of course so I guess not having a JIRA is the better way.

Examples of things which will always need a JIRA, even though they are micro:
 - deprecated a method
 - any incompatibility potential (even .impl API changes)
 - dependency version changes
 - anything which might change semantics, affect default encoding / index state (even if it's a fix)

gunnarmorling:
Pretty much how I'd handle this, too. +1.

gsmet:
Merged!

PR #1065, opened 2016-04-18
HSEARCH-2207 Treat BooleanQuery as an immutable Query

mincong-h:
`BooleanQuery` has been used traditionally like this:

> ```java
BooleanQuery q = new BooleanQuery();
q.add( otherQuery, Occur.SOMETHING );
```

But now the no-arguments constructor is deprecated, and adding clauses to the instance is deprecated too.
All our code needs to be updated to use `org.apache.lucene.search.BooleanQuery.Builder` instead, as in Lucene 6 all queries become immutable. `BooleanQuery` is used in many areas, including tests.

This merge contains most of the classes in which `BooleanQuery` is used, except the following : 
* [engine/src/main/java/org/hibernate/search/query/dsl/impl/ConnectedMultiFieldsRangeQueryBuilder.java](https://github.com/hibernate/hibernate-search/blob/master/engine/src/main/java/org/hibernate/search/query/dsl/impl/ConnectedMultiFieldsRangeQueryBuilder.java)
* [engine/src/main/java/org/hibernate/search/query/dsl/impl/MoreLikeThisBuilder.java](https://github.com/hibernate/hibernate-search/blob/master/engine/src/main/java/org/hibernate/search/query/dsl/impl/MoreLikeThisBuilder.java)
* [engine/src/main/java/org/hibernate/search/query/engine/impl/FieldNameCollector.java](https://github.com/hibernate/hibernate-search/blob/master/engine/src/main/java/org/hibernate/search/query/engine/impl/FieldNameCollector.java)

Because I don't know how to modify them correctly.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/query/dsl/impl/ConnectedMultiFieldsRangeQueryBuilder.java
+++ b/engine/src/main/java/org/hibernate/search/query/dsl/impl/ConnectedMultiFieldsRangeQueryBuilder.java
@@ -67,11 +67,11 @@ public class ConnectedMultiFieldsRangeQueryBuilder implements RangeTerminationEx
 			return queryCustomizer.setWrappedQuery( createQuery( fieldContexts.get( 0 ), conversionContext ) ).createQuery();
 		}
 		else {
-			BooleanQuery.Builder builder = new BooleanQuery.Builder();
+			BooleanQuery.Builder aggregatedFieldsQueryBuilder = new BooleanQuery.Builder();
 			for ( FieldContext fieldContext : fieldContexts ) {
-				builder.add( createQuery( fieldContext, conversionContext ), BooleanClause.Occur.SHOULD );
+				aggregatedFieldsQueryBuilder.add( createQuery( fieldContext, conversionContext ), BooleanClause.Occur.SHOULD );
 			}
-			return queryCustomizer.setWrappedQuery( builder.build() ).createQuery();
+			return queryCustomizer.setWrappedQuery( aggregatedFieldsQueryBuilder.build() ).createQuery();
 		}
 	}
 
--- a/engine/src/main/java/org/hibernate/search/query/dsl/impl/ConnectedMultiFieldsRangeQueryBuilder.java
+++ b/engine/src/main/java/org/hibernate/search/query/dsl/impl/ConnectedMultiFieldsRangeQueryBuilder.java
@@ -67,11 +67,11 @@ public class ConnectedMultiFieldsRangeQueryBuilder implements RangeTerminationEx
 			return queryCustomizer.setWrappedQuery( createQuery( fieldContexts.get( 0 ), conversionContext ) ).createQuery();
 		}
 		else {
-			BooleanQuery aggregatedFieldsQuery = new BooleanQuery();
+			BooleanQuery.Builder builder = new BooleanQuery.Builder();
 			for ( FieldContext fieldContext : fieldContexts ) {
-				aggregatedFieldsQuery.add( createQuery( fieldContext, conversionContext ), BooleanClause.Occur.SHOULD );
+				builder.add( createQuery( fieldContext, conversionContext ), BooleanClause.Occur.SHOULD );
 			}
-			return queryCustomizer.setWrappedQuery( aggregatedFieldsQuery ).createQuery();
+			return queryCustomizer.setWrappedQuery( builder.build() ).createQuery();
 		}
 	}
 
--- a/engine/src/main/java/org/hibernate/search/query/dsl/impl/ConnectedMultiFieldsRangeQueryBuilder.java
+++ b/engine/src/main/java/org/hibernate/search/query/dsl/impl/ConnectedMultiFieldsRangeQueryBuilder.java
@@ -67,7 +67,7 @@ public class ConnectedMultiFieldsRangeQueryBuilder implements RangeTerminationEx
 			return queryCustomizer.setWrappedQuery( createQuery( fieldContexts.get( 0 ), conversionContext ) ).createQuery();
 		}
 		else {
-			BooleanQuery aggregatedFieldsQuery = new BooleanQuery( );
+			BooleanQuery aggregatedFieldsQuery = new BooleanQuery();
 			for ( FieldContext fieldContext : fieldContexts ) {
 				aggregatedFieldsQuery.add( createQuery( fieldContext, conversionContext ), BooleanClause.Occur.SHOULD );
 			}
--- a/engine/src/main/java/org/hibernate/search/query/dsl/impl/MoreLikeThisBuilder.java
+++ b/engine/src/main/java/org/hibernate/search/query/dsl/impl/MoreLikeThisBuilder.java
@@ -213,18 +213,18 @@ public class MoreLikeThisBuilder<T> {
 			return createQuery( q.get( 0 ), fieldsContext.getFirst() );
 		}
 		else {
-			BooleanQuery.Builder builder = new BooleanQuery.Builder();
+			BooleanQuery.Builder queryBuilder = new BooleanQuery.Builder();
 			//the fieldsContext indexes are aligned with the priority queue's
 			Iterator<FieldContext> fieldsContextIterator = fieldsContext.iterator();
 			for ( PriorityQueue<Object[]> queue : q ) {
 				try {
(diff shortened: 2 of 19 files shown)

### CODE REVIEW

**Comment by gunnarmorling** [comment]
ok to test

**Comment by Sanne** [comment]
> This merge contains most of the classes in which BooleanQuery is used, except the following :

Well done, I like that you opted for a safe patch first!

We will still need a fix for the remaining cases though. We can break it up into a follow-up task, to keep the focus now on the process of merging your first patch.

**Comment by gunnarmorling** [comment]
@mincong-h, if you haven't done so, can you "sign" the JBoss contributor license agreement, please. It's a small web app living at https://cla.jboss.org/. We need that in order to merge your changes. Thanks!

**Comment by mincong-h** [comment]
@gunnarmorling thanks for your notice, I've signed it. Please check.

By the way, I only signed for issue HSEARCH-2207 in _Step 3 What do you want to contribute ?_. I don't know if I should sign again for a higher scope.

@Sanne, thanks. I'll keep looking into the remaining.

**Comment by Sanne** [comment]
> By the way, I only signed for issue HSEARCH-2207

Thanks, that's enough for all contributions. You're signing for the whole project, the question in step3 is just to give the approvers some context.

**Comment by gsmet** [comment]
@mincong-h what's your issue with ConnectedMultiFieldsRangeQueryBuilder exactly?

As for MoreLikeThisBuilder, I suppose your problem is with https://github.com/hibernate/hibernate-search/blob/master/engine/src/main/java/org/hibernate/search/query/dsl/impl/MoreLikeThisBuilder.java#L186 ? If so, I think you should simply remove the optimization and always create a new BooleanQuery with the builder.

**Comment by mincong-h** [comment]
@gsmet, I fixed the problem of `MoreLikeThisBuilder` by following your advice. The problem was maybe due to `BooleanQuery`'s mutability. Queries built by default constructor are mutable, but those built by builder aren't. I can't remember much detail about it, but essential is the error got fixed. By the way, should `BooleanQuery.TooManyClauses` be used during query building, e.g. [line 194](https://github.com/hibernate/hibernate-search/blob/master/engine/src/main/java/org/hibernate/search/query/dsl/impl/MoreLikeThisBuilder.java#L194)?

File: engine/src/main/java/org/hibernate/search/query/dsl/impl/ConnectedMultiFieldsRangeQueryBuilder.java [hunk 1]
```diff
@@ -67,11 +67,11 @@ public Query createQuery() {
 			return queryCustomizer.setWrappedQuery( createQuery( fieldContexts.get( 0 ), conversionContext ) ).createQuery();
 		}
 		else {
-			BooleanQuery aggregatedFieldsQuery = new BooleanQuery();
+			BooleanQuery.Builder builder = new BooleanQuery.Builder();
```
gsmet: Would you mind naming the variable aggregatedFieldsQueryBuilder instead of builder?

engine/src/main/java/org/hibernate/search/query/dsl/impl/MoreLikeThisBuilder.java
```diff
@@ -220,18 +213,18 @@ else if ( length == 1 ) {
 			return createQuery( q.get( 0 ), fieldsContext.getFirst() );
 		}
 		else {
-			BooleanQuery query = new BooleanQuery();
+			BooleanQuery.Builder builder = new BooleanQuery.Builder();
```
gsmet: Same here, queryBuilder would be better.

engine/src/main/java/org/hibernate/search/query/dsl/impl/ConnectedMultiFieldsRangeQueryBuilder.java
mincong-h: Actually this is my doubt. On one side, I admit that `builder` is confusing, especially in complex code, because we lose visibility about its utility (e.g. aggregation) or type (e.g. boolean). On the other side, the naming with 4 words seems to be difficult to read. @gsmet, could you share your vision about it ?

gsmet: I think aggregatedFieldsQueryBuilder is fine. Better long than ambiguous.

mincong-h: Got it. In previous commit c760655, there're same kind of naming problems. I'll apply the same role.

engine/src/main/java/org/hibernate/search/query/dsl/impl/MoreLikeThisBuilder.java
```diff
@@ -188,18 +188,11 @@ private Query maybeExcludeComparedEntity(Query query) {
...
-				booleanQuery.add( query, BooleanClause.Occur.MUST );
-			}
-			else {
-				booleanQuery = (BooleanQuery) query;
-			}
-			booleanQuery.add(
-					new ConstantScoreQuery( findById ),
-					BooleanClause.Occur.MUST_NOT );
-			return booleanQuery;
+			return new BooleanQuery.Builder()
```
gunnarmorling: @mincong-h, previously this would have re-used an incoming `BooleanQuery`, whereas now it always creates a new builder. Was there any specific reason for getting rid of this optimization?

gunnarmorling: Ah, this relates to the discussion above with @gsmet. Makes sense.

gunnarmorling:
@mincong-h, looks very good, great start!

Are any usages of the deprecated constructor fixed with that? Or is there anything remaining?

mincong-h:
Everything should be fixed and nothing remaining. `BooleanQuery` is used in [26 classes][26], in which 
- 20 classes have been changed
- 6 classes hava not been changed, because they're not concerned.

Here're the entire list of those 6 classes : 

- [engine/src/main/java/org/hibernate/search/query/dsl/impl/BooleanQueryBuilder.java][1] 
- [elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ToElasticsearch.java ][2]
- [engine/src/main/java/org/hibernate/search/query/engine/impl/LuceneHSQuery.java][3]
- [orm/src/test/java/org/hibernate/search/test/configuration/ProgrammaticMappingTest.java][4]
- [orm/src/test/java/org/hibernate/search/test/query/dsl/MoreLikeThisTest.java ][5]
- [engine/src/main/java/org/hibernate/search/util/logging/impl/Log.java][6]

[26]: https://github.com/hibernate/hibernate-search/search?utf8=%E2%9C%93&q=BooleanQuery
[1]: https://github.com/hibernate/hibernate-search/blob/master/engine/src/main/java/org/hibernate/search/query/dsl/impl/BooleanQueryBuilder.java
[2]: https://github.com/hibernate/hibernate-search/blob/master/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ToElasticsearch.java/impl/ToElasticsearch.java 
[3]: https://github.com/hibernate/hibernate-search/blob/master/engine/src/main/java/org/hibernate/search/query/engine/impl/LuceneHSQuery.java
[4]: https://github.com/hibernate/hibernate-search/blob/master/orm/src/test/java/org/hibernate/search/test/configuration/ProgrammaticMappingTest.java
[5]: https://github.com/hibernate/hibernate-search/blob/master/orm/src/test/java/org/hibernate/search/test/query/dsl/MoreLikeThisTest.java
[6]: https://github.com/hibernate/hibernate-search/blob/master/engine/src/main/java/org/hibernate/search/util/logging/impl/Log.java

@gunnarmorling As this is my first commit, it would be better to check again theses 6 classes by you.

gunnarmorling:
Will do. Thanks!

gsmet:
@mincong-h just a quick tip: you'd better keep your master in sync with upstream master and create branches to do your work.

Typically with a clean master:
```
git checkout master
git pull --rebase upstream master
git checkout -b HSEARCH-XXX
```
then do your work in this branch and commit, then:
```
git push origin HSEARCH-XXX
```
and build your PR from this branch.

This way, it will be easier for you to keep in sync with upstream.

gsmet:
@mincong-h I made a few additional fixes and squashed the commits.

Merged! Thanks for your first contribution.

PR #1066, opened 2016-04-18
HSEARCH-2226 FullTextQuery does not override some returned Query type…

Sanne:
…s to allow method chaining


gsmet:
Merged! Thanks!

PR #1067, opened 2016-04-18
HSEARCH-2229 Improve error message when we get a Lucene analyzer instead of a remote one

gsmet:


elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -269,12 +269,12 @@ private void createIndexMappings() {
...
 	}
 
-	private String analyzerName(DocumentFieldMetadata fieldMetadata) {
+	private String analyzerName(Class<?> entityType, DocumentFieldMetadata fieldMetadata) {
 		AnalyzerReference analyzer = fieldMetadata.getAnalyzer();
 		if ( analyzer.is( RemoteAnalyzerReference.class ) ) {
 			return analyzer.unwrap( RemoteAnalyzerReference.class ).getName();
 		}
-		LOG.analyzerIsNotRemote( String.valueOf( analyzer ) );
+		LOG.analyzerIsNotRemote( entityType, fieldMetadata.getFieldName(), analyzer.toString() );
```
Sanne: if you simply delete the "toString", you can have the formatter invoke that if it's needed.

That will avoid creating the String when the log level is disabled.. although it's WARN in this case so it shouldn't matter much, just pointing out that a `toString()` looks suspicious

Sanne:
Will you add a test?

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
gsmet: OK, will do.

gsmet:
I didn't plan to add one. Do you think it might be useful?

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
gsmet: Done.

Sanne:
merged, thanks!

About the unit test: not required, just thinking that we might eventually have a regression. We could add one later, to keep the focus now on the more important tasks to deliver the 5.6 release. I won't even create a JIRA for this as it's very minor matter.

PR #1068, opened 2016-04-18
HSEARCH-2067 Introduce the notion of backend specific FieldBridges (v2)

gsmet:
Following Gunnar's feedback, here is a v2 of HSEARCH-2067.

I still think we should add the indexManagerType to EntityIndexBinding and a isManagedBy method in ExtendedSearchIntegrator but I suppose it can be done later (in 6?).


gunnarmorling:
> I still think we should add the indexManagerType to EntityIndexBinding...

+1. When we allow ourselves breaking/amending some public contracts, we should be able to find a better way.

On first sight it seems odd that the index manager type is needed for meta-data retrieval, but it is due to the link to field bridges. Maybe the `TypeMetadata` should be agnostic of that and only the `EntityIndexBinding` has references to a specific IM, the corresponding field bridges etc.? Anyways, something to consider for 6.

engine/src/main/java/org/hibernate/search/bridge/impl/BridgeFactory.java
```diff
@@ -177,19 +174,28 @@ public FieldBridge buildSpatialBridge(Spatial spatial, XClass clazz, String lati
...
 
 	public FieldBridge buildFieldBridge(Field field,
 			XMember member,
 			boolean isId,
 			boolean isExplicitlyMarkedAsNumeric,
+			Class<? extends IndexManager> indexManagerType,
 			ReflectionManager reflectionManager,
 			ServiceManager serviceManager
 	) {
+		// if we don't know to which IndexManager the entity is rattached, we cannot build the right
```
gunnarmorling: rattached=attached

engine/src/main/java/org/hibernate/search/bridge/spi/BackendSpecificBridgeProvider.java
```diff
@@ -0,0 +1,25 @@
...
+ * @author Guillaume Smet
+ */
+public interface BackendSpecificBridgeProvider extends BridgeProvider {
+
+	/**
+	 * Returns the backend for which we register the bridge
+	 *
+	 * @return the backend for which we register the bridge
+	 */
+	Class<? extends IndexManager> getBackend();
```
gunnarmorling: `getIndexManagerType()`? The term "backend" currently is used for something like "JMS" or "JGroups".

engine/src/main/java/org/hibernate/search/bridge/spi/BackendSpecificBridgeProvider.java
```diff
@@ -0,0 +1,25 @@
...
+package org.hibernate.search.bridge.spi;
+
+import org.hibernate.search.indexes.spi.IndexManager;
+
+/**
+ * Bridge provider specific to a given backend.
+ *
+ * @author Guillaume Smet
+ */
+public interface BackendSpecificBridgeProvider extends BridgeProvider {
```
gunnarmorling: Should we mark it as experimental as we expect it's merged into `BridgeProvider` as of 6?

engine/src/main/java/org/hibernate/search/indexes/impl/IndexManagerHolder.java
```diff
@@ -65,6 +65,10 @@
 
 	private static final Similarity DEFAULT_SIMILARITY = new ClassicSimilarity();
 
+	private static final String DEFAULT_INDEX_MANAGER_KEY = "__DEFAULT__";
+
+	private final Map<String, Class<? extends IndexManager>> indexManagerImplementationsRegistry
```
gunnarmorling: Can be declared as `ConcurrentMap`, making it easier for the reader to reason about the thread-safety (same below).

gunnarmorling: Can you clear it in `stop()`?

engine/src/main/java/org/hibernate/search/indexes/impl/IndexManagerHolder.java
```diff
@@ -547,4 +551,34 @@ private boolean isShardingDynamic(Properties indexProperty, WorkerBuildContext b
...
+
+	public Class<? extends IndexManager> getIndexManagerType(XClass entity, SearchConfiguration cfg, WorkerBuildContext buildContext) {
+		ServiceManager serviceManager = buildContext.getServiceManager();
+		IndexManagerFactory indexManagerFactory = serviceManager.requestService( IndexManagerFactory.class );
+
+		String indexName = getIndexName( entity, cfg );
+		Properties[] indexProperties = getIndexProperties( cfg, indexName );
+		String indexManagerImplementationName = indexProperties[0].getProperty( Environment.INDEX_MANAGER_IMPL_NAME );
+
+		String indexManagerImplementationKey = indexManagerImplementationName != null ?
```
gunnarmorling: Better use `StringHelper#isEmpty()´ to be consistent with the usage below.

engine/src/main/java/org/hibernate/search/util/logging/impl/Log.java
```diff
@@ -934,4 +934,8 @@
 
 	@Message(id = 305, value = "Analyzer reference with name '%2$s' not initialized for field '%1$s' ")
 	SearchException analyzerReferenceNotInitialized(String fieldName, String name);
+
+	@Message(id = 306, value = "IndexManager type is required to build a FieldBridge for %2$s in %1$s")
```
gunnarmorling: Given it's meant to be read by users, I'd avoid the camel-casing: "Index manager", "field bridge".

gunnarmorling:
@gsmet looks good to me!

I'll leave the final judgement to @Sanne, specifically whether he sees any issue with instantiating (but not starting/initializing) the IM types. The alternative would be a new method on `IndexManagerFactory` which again would lead us to an SPI break.

gsmet:
Done. Thanks for the feedback.

@Sanne waiting for your blessing!

gsmet:
Jenkins, retest this please

gsmet:
@Sanne I was wondering if we could add a method in EntityIndexBinding to get the IndexManager type? That's what I did in my first patch but we removed this part as it did change the SPI.

The fact is that I end up with this type of code in 3 places in the DSL (upcoming patch):
```
queryContext.getFactory().getIndexBinding( queryContext.getEntityType() ).getIndexManagers()[0] instanceof RemoteAnalyzerProvider
```
which is kinda ugly.

Should we keep the code like that for now and add this to the 6 todo list?

Sanne:
Yes, `EntityIndexBinding` seems better than implying that IndexManager(s) are of the same type.

The email I just sent to the mailing list goes a step further, something along the lines of:

     getIndexBinding( clazz ).getIndexFamily().getStorageType();

(exact names need improving: just to give you a brief idea of what I had in mind)

gunnarmorling:
@Sanne, what is `getStorageType()`? What does it return?

gustavonalle:
I tested this patch and it solves the issue we are facing at https://issues.jboss.org/browse/ISPN-6522
Is it possible to release another Alpha with this fix soon?

gunnarmorling:
@gustavonalle We've planned another release at the end of next week (likely Beta1). Would that be alright for you?

Btw. ISPN-6522 reveals another issue on our side, specifically an entry is missing from module.xml for the ES backend. I've filed [HSEARCH-2234](https://hibernate.atlassian.net/browse/HSEARCH-2234).

gustavonalle:
@gunnarmorling a Beta1 next week sounds good, thanks

Sanne:
@gustavonalle just wondering.. would everything work fine in Infinispan if you avoided to deploy the new ES module? (i.e. filter it out when uncompressing the modules)

gustavonalle:
@Sanne That was my first thought to workaround the issue (plan B actually), but if someone adds the backend jar to the classpath in library mode, the issue would still happen.

Sanne:
@gustavonalle sure, I didn't mean to suggest that as a solution, but as a quick way to confirm that this was the problem (and that there are no other problems as well)

Sanne:
> what is getStorageType()? What does it return?

I didn't think it through. A simple `enum` could work, it's not extensible but then again I doubt people will be able to plug in a whole alternative storage type w/o making changes to source code anyway.

Better ideas? Or we can go with the enum as a starting point.

gunnarmorling:
What would the enum represent? Something like `Lucene`, `Elasticsearch`, `Solr` etc.?

I'm a bit reluctant on doing conditional logic in core based on that (if that's the intended purpose). Rather I'd suggest to have methods on the specific `IndexFamily` implementation (living in the module of the respective "backend") which do their thing.

Anyways, I think we can discuss that separately, not to block this PR.

Sanne:
> I'm a bit reluctant on doing conditional logic in core based on that (if that's the intended purpose). Rather I'd suggest to have methods on the specific IndexFamily implementation (living in the module of the respective "backend") which do their thing.

Right. That's why I brought the family concept up. Not sure how to model that yet, but I'd like to avoid "instanceof" of aspecific IndexManager class type; sketching some ideas for later.

> Anyways, I think we can discuss that separately, not to block this PR.

Yes, I wasn't waiting on this point. Just delayed as I had some failures to inspect, seem unrelated though.

Sanne:
merged!

PR #1069, opened 2016-04-19
HSEARCH-2214 Only wait for status "yellow" after index creation

gunnarmorling:
[HSEARCH-2214](https://hibernate.atlassian.net/browse/HSEARCH-2214).

### DIFF
--- a/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
+++ b/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
@@ -97,6 +97,7 @@ Hostname and port for Elasticsearch:: `hibernate.search.elasticsearch.host \http
 Select Elasticsearch as the backend:: `hibernate.search.default.indexmanager elasticsearch`
 Selects the index creation strategy:: `hibernate.search.elasticsearch.index_management_strategy CREATE_DELETE`
 Maximum time to wait for the indexes to become available before failing (in ms):: `hibernate.search.elasticsearch.index_management_wait_timeout 10000`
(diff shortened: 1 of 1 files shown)


Sanne:
No I don't think we can accept to write on "yellow" by default.
HSEARCH-2214 reads: Provide **an option** to wait for Elasticsearch status yellow rather than green

If people want to set some "unsafe" option and disable seatbelts, that's their problem.. but safe defaults should be our responsibility.

gunnarmorling:
Did you see my last comment on the Jira issue? How would you deal with
status degradation after start-up? I had hoped we'd quickly agree, but if
not, let's have a discussion over there.

Sanne:
> Did you see my last comment on the Jira issue?

Saw it now, sorry for the delay :-/ replied..

happy to pick up the phone it it helps? I feel like my very question about the property is important but still ignored by all replies..

gunnarmorling:
@Sanne pushed an update, making it an option as discussed. The question is how to test it in an automated fashion (I've done some manual testing which looked good). We'd likely need a separate _elasticsearch.yml_ and thus a separate execution of the ES Maven plug-in. Doable, but it'll take time. Should we go for it?

DavideD:
I'm OK to post-pone the testing using a separate configuration for now and
create a separate JIRA so that we don't forget to add it later.

gunnarmorling:
Rebased and force-pushed an update to cover the new property in docs, too.

Sanne:
too late! I had just merged it, and was about to open a jira to track the need for docs..

Sanne:
cool, all merged. Seemed you had rebased on your own changes so it was linear still :)

PR #1070, opened 2016-04-19
HSEARCH-2071 Support EmbeddedId

DavideD:
https://hibernate.atlassian.net/projects/HSEARCH/issues/HSEARCH-2071

I checked the test `EmbeddableIdTest` and `ProgrammaticEmbeddedItTest` and they won't work with elasticsearch for the following reasons:
* the field bridge used for the id is not a `MetadataProvidingFieldBridge`; because of this, the new fields are not defined during the schema creation and this causes some errors when we need to create them;
* the fields name contain a `.` ;

I've created new tests without this limitations and checked that the Lucene and Elasticsearch backend are consistent. This resulted in another test failure because on Lucene a field was store not analyzed while in Elasticsearch was stored analyzed.

I tried to fix the issue by adding some information in the `BridgeDefinedField`. Let me know if it makes sense.

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
@@ -74,6 +74,7 @@ public class ElasticsearchIndexManager implements IndexManager, RemoteAnalyzerPr
 	private static final Log LOG = LoggerFactory.make( Log.class );
 
 	private static final String ANALYZED = "analyzed";
+	private static final String NOT_ANALYZED = "not_analyzed";
 
 	private String indexName;
 	private String actualIndexName;
@@ -284,7 +285,7 @@ public class ElasticsearchIndexManager implements IndexManager, RemoteAnalyzerPr
 			// TODO HSEARCH-2256 Should we make this configurable?
 			JsonObject field = new JsonObject();
 			field.addProperty( "type", "string" );
-			field.addProperty( "index", "not_analyzed" );
+			field.addProperty( "index", NOT_ANALYZED );
 			properties.add( DocumentBuilderIndexedEntity.TENANT_ID_FIELDNAME, field );
 
 			// normal document fields
@@ -421,7 +422,7 @@ public class ElasticsearchIndexManager implements IndexManager, RemoteAnalyzerPr
 				// the fields potentially created for the spatial hash queries
 				JsonObject field = new JsonObject();
 				field.addProperty( "type", "string" );
-				field.addProperty( "index", "not_analyzed" );
+				field.addProperty( "index", NOT_ANALYZED );
 
 				getOrCreateProperties( payload, fieldName ).add( fieldName, field );
 			}
@@ -435,7 +436,7 @@ public class ElasticsearchIndexManager implements IndexManager, RemoteAnalyzerPr
 		JsonObject field = new JsonObject();
 		field.addProperty( "type", getFieldType( facetMetadata ) );
 		field.addProperty( "store", false );
-		field.addProperty( "index", "not_analyzed" );
+		field.addProperty( "index", NOT_ANALYZED );
 
 		getOrCreateProperties( payload, fullFieldName).add( simpleFieldName, field );
 		return field;
@@ -447,7 +448,7 @@ public class ElasticsearchIndexManager implements IndexManager, RemoteAnalyzerPr
 		if ( FieldHelper.isBoolean( binding, fieldMetadata.getName() ) ||
 				FieldHelper.isDate( binding, fieldMetadata.getName() ) ||
 				FieldHelper.isNumeric(fieldMetadata) ) {
-			return "not_analyzed";
+			return NOT_ANALYZED;
 		}
 
 		return getIndexValue( fieldMetadata.getIndex() );
@@ -459,7 +460,7 @@ public class ElasticsearchIndexManager implements IndexManager, RemoteAnalyzerPr
 		if ( FieldHelper.isBoolean( bridgeDefinedField )
 				|| FieldHelper.isNumeric( bridgeDefinedField )
 				|| FieldHelper.isDate( bridgeDefinedField ) ) {
-			return "not_analyzed";
+			return NOT_ANALYZED;
 		}
 
 		Field.Index index = bridgeDefinedField.getIndex();
@@ -473,7 +474,7 @@ public class ElasticsearchIndexManager implements IndexManager, RemoteAnalyzerPr
 				return ANALYZED;
 			case NOT_ANALYZED:
 			case NOT_ANALYZED_NO_NORMS:
-				return "not_analyzed";
+				return NOT_ANALYZED;
 			case NO:
 				return "no";
 			default:
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
@@ -442,7 +442,6 @@ public class ElasticsearchIndexManager implements IndexManager, RemoteAnalyzerPr
 		return field;
 	}
 
-	@SuppressWarnings("deprecation")
 	private String getIndex(EntityIndexBinding binding, DocumentFieldMetadata fieldMetadata) {
 		// Never analyze boolean, date or numeric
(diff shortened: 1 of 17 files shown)


Sanne:
@DavideD why is this assigned to you? Please give yourself advice in how to proceed :)

DavideD:
Problem solved :)

Sanne:
ouch :) :+1:

gsmet:
Jenkins, please retest this

gsmet:
Jenkins, retest this please

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -383,7 +390,11 @@ private void addFieldMapping(JsonObject payload, BridgeDefinedField bridgeDefine
 			JsonObject field = new JsonObject();
 
 			field.addProperty( "type", getFieldType( bridgeDefinedField ) );
-			field.addProperty( "index", ANALYZED );
+
+			String index = getIndex( bridgeDefinedField );
+			field.addProperty( "index", index );
+
+			// TODO: We need to add the analyzer name
```
gsmet: Looks like this TODO should be fixed.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -439,7 +450,22 @@ private String getIndex(EntityIndexBinding binding, DocumentFieldMetadata fieldM
...
 		}
 
-		switch ( fieldMetadata.getIndex() ) {
+		return getIndexValue( fieldMetadata.getIndex() );
+	}
+
+	@SuppressWarnings("deprecation")
+	private String getIndex(BridgeDefinedField bridgeDefinedField) {
+		// Never analyze boolean
+		if ( bridgeDefinedField.getType().equals( FieldType.BOOLEAN ) ) {
```
gsmet: Same for dates and numerics, see a few lines above.

engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
```diff
@@ -1271,6 +1255,22 @@ private void bindFieldAnnotation(
...
+					.fieldName( sortableField )
+					.build();
+				propertyMetadataBuilder.addSortableField( sortableFieldMetadata );
+			}
+
+			for ( BridgeDefinedField field : bridgeContributedMetadata.getFields() ) {
+				propertyMetadataBuilder.addBridgeDefinedField( field );
+			}
+		}
+
```
gsmet: Any particular reason why you moved this code around?

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -77,4 +79,7 @@
 			value = "Elasticsearch connection time-out; check the cluster status, it should be 'green';\n Request:\n========\n%1$sResponse:\n=========\n%2$s" )
 	SearchException elasticsearchRequestTimeout(String request, String response);
 
+	@Message(id = ES_BACKEND_MESSAGES_START_ID + 11,
+			value = "Projection of non-JSON-primitive field values is not supported: '%1$s'")
+	SearchException unsupportedProjectionOfNonJSONPrimitiveFields(JsonElement value);
```
gsmet: Minor nitpicking: we use the case "Json" everywhere as we conform to the Gson convention. I'm only talking about the method name, the message is fine.

orm/src/test/java/org/hibernate/search/test/id/withmeta/EmbeddedIdWithMetadataProvidingBridgeTest.java
```diff
@@ -0,0 +1,109 @@
...
+		assertEquals( 1, results.size() );
+		emmanuel = (Person) results.get( 0 );
+		assertEquals( "Red", emmanuel.getFavoriteColor() );
+		s.delete( results.get( 0 ) );
+		tx.commit();
+		s.close();
+	}
+
+	/**
+	 * HSEARCH-HSEARCH-306, HSEARCH-248
```
gsmet: It's probably coming from the original test but s/HSEARCH-HSEARCH/HSEARCH/.

gsmet:
Couple of typos in one of the commit message:
> HSEARCH-2071 Use correct defaults for fields provided by field bridges
>  Custom field bridges in the Lucene backend can create new fields using as default
>  the settings for property accessor,

s/,/./

>  For example, if the fieldbridge is called for an id, the custom fields won't be analyzed; when
>  used an a normal field they are going to be analyzed.

s/fieldbridge/field bridge/
s/used an a/used as a/

orm/src/test/java/org/hibernate/search/test/id/withmeta/PersonPKMetadataProviderBridge.java
```diff
@@ -0,0 +1,69 @@
...
+			.field( firstName( name ), FieldType.STRING )
+			.field( lastName( name ), FieldType.STRING );
+	}
+
+	private static String lastName(String name) {
+		return name + "_lastName";
+	}
+
+	private static String firstName(String name) {
+		return name + "_firstName";
```
gsmet: OK, so in fact, you can have fields with dots: we treat them as nested documents which sounds logical. You probably should remove this part from the commit message.

gsmet:
Finished with the review. Feel free to ping me if there's something unclear.

orm/src/test/java/org/hibernate/search/test/id/withmeta/PersonPKMetadataProviderBridge.java
DavideD: I will rephrase the sentence, what I mean is that if you create a new field in the field bridge the name does not have to contain a `.`.

engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
DavideD: I remember there was one, but I forgot now :)

I will double check

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
DavideD: ok

orm/src/test/java/org/hibernate/search/test/id/withmeta/EmbeddedIdWithMetadataProvidingBridgeTest.java
DavideD: ok

engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
DavideD: Now getBridgeContributedFieldMetadata uses the DocumentFieldMetada to make sure that the default used for the custom fields are the same as the one declare on the main field.
For this reason I moved the code after the creation of the metadata.

DavideD:
Fixed the remarks, added the analyzer name and rebased

gsmet:
@DavideD I'm not convinced it's a good idea to add the analyzer and the analyzed/not analyzed information to the BridgeDefinedField.

IIRC, in the Lucene case, the only bridge defined field inheriting from the analyzer definition is the one having the same name of the default name. The other fields are not analyzed.

If I'm correct on this, I think we should stick with it for now and have the exact same behavior in Elasticsearch.

That being said, I would really like to revisit this but in this case, we also need to add the ability to override the analyzer definition when the bridge defined field is defined.

So for now, I'll stick to the behavior we already have in Lucene and create a JIRA with the additional patch attached (or a link to one of your branches) so that we can build on it later. I had a couple of thoughts in this area, I have to dig a bit more a bug I found and I'll write something on the list.

By the way, looks like your new tests are not passing.

DavideD:
@gsmet I've force-pushed  some changes:

1.  I've removed the part related to the analyzer name
2. Added a new test for core and elasticsearch that make sure that the defaults applied are the same (there is duplication and it might be improved but I don't want to spend more time on it). The default analyzer is used as yuo says.

I think we need a new JIRA to allow to set an analyzer for custom field. WWDYT?

gsmet:
Merged! Thanks @DavideD!

PR #1071, opened 2016-04-19
HSEARCH-2213 Cryptic error

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-2213

Built on top of https://github.com/hibernate/hibernate-search/pull/1069

Not sure how to test it now that we have relaxed the status to yellow.
But the error message now explains that you should check the status of the cluster.
It also shows the status at the time of the request


DavideD:
Sending another one with a different message

PR #1072, opened 2016-04-19
HSEARCH-2213 Better error message when time out occurs

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-2213

The log error will look something like this
```
org.hibernate.search.exception.SearchException: HSEARCH400010: Elasticsearch connection time-out; check the cluster status, it should be 'green';
 Request:
========
Operation: 
Data:
null
Response:
=========
Status: 408
Error message: 408 Request Timeout
Cluster name: elasticsearch
Cluster status: yellow
```

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/JestClient.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/JestClient.java
@@ -25,6 +25,8 @@ import org.hibernate.search.spi.BuildContext;
 import org.hibernate.search.util.configuration.impl.ConfigurationParseHelper;
 import org.hibernate.search.util.logging.impl.LoggerFactory;
 
+import com.google.gson.JsonElement;
+
 import io.searchbox.action.Action;
 import io.searchbox.action.BulkableAction;
 import io.searchbox.action.DocumentTargetedAction;
@@ -45,6 +47,8 @@ import io.searchbox.params.Parameters;
 public class JestClient implements Service, Startable, Stoppable {
 
 	private static final Log LOG = LoggerFactory.make( Log.class );
+
+	private static final int TIME_OUT = 408;
 
 	private io.searchbox.client.JestClient client;
 
@@ -86,7 +90,12 @@ public class JestClient implements Service, Startable, Stoppable {
 
 			// The request failed with a status that's not ignore-able
 			if ( !result.isSucceeded() && !isIgnored( result.getResponseCode(), ignoredErrorStatuses ) ) {
-				throw LOG.elasticsearchRequestFailed( requestToString( request ), resultToString( result ), null );
+				if ( result.getResponseCode() == TIME_OUT ) {
+					throw LOG.elasticsearchRequestTimeout( requestToString( request ), resultToString( result ) );
+				}
+				else {
+					throw LOG.elasticsearchRequestFailed( requestToString( request ), resultToString( result ), null );
+				}
 			}
 
 			return result;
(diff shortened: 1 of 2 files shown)


Sanne:
I'm wondering if there might be other kind of issues to cause such a TIMEOUT. i.e. what about a "traditional" stuck TCP connection?

Might be simpler to make the message slightly more generic, like:
 
    Elasticsearch connection time-out; check the cluster status.

Good idea to print out the cluster status!

DavideD:
I think it's be  to specify that it should be 'green' as some people might think that 'yellow' is an acceptable state. I guess I could write:
`Elasticsearch connection time-out; check the connection to the cluster and the status ( it should be green).`

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/JestClient.java
```diff
@@ -46,6 +48,8 @@
 
 	private static final Log LOG = LoggerFactory.make( Log.class );
 
+	private static final int TIME_OUT = 408;
```
Sanne: optional: I'd probably rename this to make it clear that it's a response code?

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -71,4 +71,8 @@
 	@Message(id = ES_BACKEND_MESSAGES_START_ID + 9,
 			value = "'%1$s' is not a remote analyzer. It will be ignored")
 	void analyzerIsNotRemote(String name);
+
+	@Message(id = ES_BACKEND_MESSAGES_START_ID + 10,
+			value = "Elasticsearch connection time-out; check the cluster status, it should be 'green';\n Request:\n========\n%1$sResponse:\n=========\n%2$s" )
```
Sanne: BTW some people will cry at seeing multi-line logs as they make grepping very hard.

I personally like this one though.. just saying that maybe we shouldn't :)

DavideD: Yes, I'm not a big fan as well. I've been just consistent with the other messages, I think we might change them as a separate issue if someone complains

DavideD: I'm also wondering if the `\n` will work on Windows

Sanne: good point. Let's just avoid it then.

gunnarmorling: Get another editor than Notepad and it will work ;)

Kidding aside, the request and response will contain line feeds anyways (JSON structures). Also I expect people to use tools such as Splunk, Logstash etc. when analyzing production logs. To me, it's important to get out the required context in a sensible way, so I don't think we should let stop ourselves by people stuck with their tools in the 90ies.

DavideD:
* Rebased
* log message updated to
```
org.hibernate.search.exception.SearchException: HSEARCH400010: Elasticsearch connection time-out; check the connection to the cluster and the status (it should be 'green'); Request: [operation: class org.hibernate.search.backend.elasticsearch.impl.ElasticsearchIndexManager$1, data: null], Response: [error: 408 Request Timeout, response code: 408, cluster: elasticsearch, status: yellow]
```
* removed line separator char from the other error messages

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
Sanne: ok yes I have no strong opinion.. go ahead if you like it better.

Still, I'm sure that some people are stuck with their tools of the 90ths and will hate you as it might not be their choice to change that :)

Sanne:
great, merged!

PR #1073, opened 2016-04-20
HSEARCH-2233 Dates and numerics should never be analyzed

gsmet:
This does not need unit tests as it's already covered by the existing tests (tests are failing in my es-analyzer branch without it).


gunnarmorling:
Merged; Thanks, @gsmet!

PR #1074, opened 2016-04-21
HSEARCH-2164 Changing properties for ES backend so they work globally…

gunnarmorling:
… and/or per index

Only the "host" property is not "supported officially" per index. Likely it would still work if you know what you are doing (no queries across cluster boundaries), but I'd prefer to just promise a single cluster for now. I still opted to use the "default" syntax in order to avoid having users to change it down the road.


Sanne:
Looks great, thanks! Merging..
Had to fiddle a bit with conflicts, hopefully got it right.

Sanne:
P.S. conflicted with your own work on HSEARCH-2214

PR #1075, opened 2016-04-22
HSEARCH-2237 ScopedAnalyzerReference shouldn't close the underlying analyzer references

gsmet:
This one was well hidden and was triggered by my work generalizing the usage of analyzer references.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/util/impl/ScopedAnalyzerReference.java
+++ b/engine/src/main/java/org/hibernate/search/util/impl/ScopedAnalyzerReference.java
@@ -91,12 +91,9 @@ public final class ScopedAnalyzerReference implements AnalyzerReference {
 
 	@Override
 	public void close() {
-		if ( globalAnalyzer != null ) {
-			globalAnalyzer.close();
-		}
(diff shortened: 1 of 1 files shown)


Sanne:
How can you guarantee that all analyzers we created are going to be closed eventually?

Not closing one leads to critical memory leaks as several implementations use thread locals internally.

Our usual rule of thumb is "who creates it, closes it", but I'm having a hard time understanding what is going to escape from this class now.

gsmet:
Well, first, keep in mind that `ScopedAnalyzerReference` is supposed to mirror the `ScopedAnalyzer` behavior. `ScopedAnalyzer` don't close the underlying analyzers and it has been like that from the beginning.

Second, there is definitely an analyzer leaking and it's the PassThroughAnalyzer which is statically shared. It might not be a good idea but it's not something new.

The fact is that if we don't like this behavior, it's something that we should fix but it's definitely a different task and this PR just make the AnalyzerReference code works as it was working before with Analyzers.

Sanne:
thanks for the explanation. Looks better now :)

Sanne:
merged, thanks!

PR #1076, opened 2016-04-22
HSEARCH-1951 Validate sort type used for query

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-1951

### DIFF
--- a/engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
+++ b/engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
@@ -6,6 +6,8 @@
  */
 package org.hibernate.search.query.engine.impl;
 
+import static org.hibernate.search.util.impl.CollectionHelper.newHashMap;
+
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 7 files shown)


engine/src/test/java/org/hibernate/search/test/sorting/SortingTest.java
```diff
@@ -46,8 +48,13 @@
  */
 public class SortingTest {
 
+	private static final String SORT_TYPE_ERROR_CODE = "HSEARCH000306";
```
gsmet: Should be 307. Probably a merge problem.

engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
```diff
@@ -231,31 +240,185 @@ public ExtendedSearchIntegrator getExtendedSearchIntegrator() {
...
+			throw LOG.distanceSortRequiresSpatialField( sortField.getField() );
+		}
+	}
+
+	private FieldDescriptor findFieldDescriptor(ExtendedSearchIntegrator extendedIntegrator, Iterable<Class<?>> targetedEntities, String field) {
+		if ( field == null ) {
+			return null;
+		}
+		for ( Class<?> clazz : targetedEntities ) {
+			IndexedTypeDescriptor indexedTypeDescriptor = extendedIntegrator.getIndexedTypeDescriptor( clazz );
```
gsmet: Is there a reason why you used the FieldDescriptor API? From what I know, it's not recommended to use it in our code. See for instance Gunnar's comment in https://hibernate.atlassian.net/browse/HSEARCH-2070 .

engine/src/test/java/org/hibernate/search/test/sorting/SortingTest.java
DavideD: Ops, I didn't include the change in the PR

engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
DavideD: This part was already like this, I just refactored the code.
I'll check if it works using the DocumentFieldMetadata

DavideD:
@gsmet Thanks for the feedback. I've push-forced some changes. Now I'm using the DocumentFieldMetadata for spatial as well..

Sanne:
Looks great. I'm backporting it .. not trivial as:
 - the patched files didn't exist
 - some of the used interfaces didn't exist either

Sanne:
I've merged it but made one little change: 6d390a0

I realized the need for this after backporting.

PR #1078, opened 2016-04-23
HSEARCH-2240 Query of the wrong type being generated during parallel …

Sanne:
…execution

- https://hibernate.atlassian.net/browse/HSEARCH-2240

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/service/impl/StandardServiceManager.java
+++ b/engine/src/main/java/org/hibernate/search/engine/service/impl/StandardServiceManager.java
@@ -144,7 +144,16 @@ public class StandardServiceManager implements ServiceManager {
 		return Collections.unmodifiableMap( tmpServices );
 	}
 
-	private <S extends Service> ServiceWrapper<S> createAndCacheWrapper(Class<S> serviceRole) {
+	/**
+	 * The 'synchronized' is necessary to avoid loading the same service in parallel: enumerating service
+	 * implementations is not threadsafe when delegating to Hibernate ORM's org.hibernate.boot.registry.classloading.spi.ClassLoaderService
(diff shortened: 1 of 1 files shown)


Sanne:
Marked urgent as I can't essentially test reliably any other PR until this is merged

gunnarmorling:
Looks ok, but could you explain how this relates to the issue described in 2240? It's not quite clear how repeated service loading causes this error. Thx!

Sanne:
Running elasticsearch queries in parallel would attempt to lookup the Query Translator service in parallel.

Parallel service lookups would attempt to iterate on the ORM's ClassloaderService, which doesn't enumerate service implementors in a threadsafe way.

The result is that "some" threads would fail to find the translator service implementation, fall back to "null", which is then handled by the query execution code as a "no need for translation" and it eventually attempts to run the Lucene query - eventually it fails to run the Lucene query on the ES IndexManager.

Seems like an old bug, specifically in the ORM ClassLoader wrapper, although all consumers of it normally wouldn't lookup a service at runtime - using it sequentially at bootstrap as we normally do doesn't trigger the issue.

gunnarmorling:
Holy moly! Did you open a HHH bug for it?

In that light, the patch looks like a good improvement, though it ultimately seems to need fixing in ORM. Can you merge it yourself?

Sanne:
> Holy moly! Did you open a HHH bug for it?

Not yet, was wondering what you'd think of it. Clearly HHH's service management is not designed to be threadsafe - somehow we assumed it would from a Search perspective.

With this fix, it's not going to bother us anymore.

I'll open one so we can see if the ORM teams wants to fix it or add some comment/warning.

>Can you merge it yourself?

sure, thanks

PR #1079, opened 2016-04-24
HSEARCH-2241 Clarify deprecation of setFilter() method on FullTextQuery

Sanne:
Public API references to `org.apache.lucene.search.Filter` have to be removed.

This is an easy first step, as the javadoc already mentioned it to be "semi-deprecated".

Let's apply it to 5.5 as well as it just makes things clearer.

 - https://hibernate.atlassian.net/browse/HSEARCH-2241


Sanne:
@Hibernate-CI retest this please

Updated: had a mistake in javadoc which would have it fail the build.

PR #1080, opened 2016-04-25
HSEARCH-2242 Convert FacetManager implementation to no longer use Fil…

Sanne:
…ters

 - https://hibernate.atlassian.net/browse/HSEARCH-2242

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchHSQueryImpl.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchHSQueryImpl.java
@@ -363,9 +363,8 @@ public class ElasticsearchHSQueryImpl extends AbstractHSQuery {
 			filters.add( ToElasticsearch.condition( "should", typeFilters ) );
 
 			// facet filters
-			Filter facetFilter = getFacetManager().getFacetFilter();
-			if ( facetFilter != null ) {
-				filters.add( ToElasticsearch.fromLuceneFilter( getFacetManager().getFacetFilter() ) );
+			for ( Query query : getFacetManager().getFacetFilters().getFilterQueries() ) {
+				filters.add( ToElasticsearch.fromLuceneQuery( query ) );
 			}
 
 			// wrap filters into must if there is more than one
--- a/engine/src/main/java/org/hibernate/search/query/engine/impl/FacetManagerImpl.java
+++ b/engine/src/main/java/org/hibernate/search/query/engine/impl/FacetManagerImpl.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.search.query.engine.impl;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
@@ -13,9 +14,7 @@ import java.util.Map;
 
 import org.apache.lucene.search.BooleanClause;
 import org.apache.lucene.search.BooleanQuery;
-import org.apache.lucene.search.Filter;
 import org.apache.lucene.search.Query;
-import org.apache.lucene.search.QueryWrapperFilter;
 import org.hibernate.search.query.dsl.impl.FacetingRequestImpl;
 import org.hibernate.search.query.engine.spi.FacetManager;
 import org.hibernate.search.query.facet.Facet;
@@ -49,9 +48,9 @@ public class FacetManagerImpl implements FacetManager {
 	private Map<String, List<Facet>> facetResults;
 
 	/**
-	 * The combined filter for all selected facets which needs to be applied on the current query
+	 * The set of boolean clause filters for all selected facets which needs to be applied on the current query
 	 */
-	private Filter facetFilter;
+	private QueryFilters facetFilterset;
 
 	/**
 	 * The query from which this manager was retrieved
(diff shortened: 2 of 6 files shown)


engine/src/main/java/org/hibernate/search/query/engine/impl/QueryFilterset.java
```diff
@@ -0,0 +1,70 @@
...
+				boolQueryBuilder.add( bc, BooleanClause.Occur.FILTER );
+			}
+			return boolQueryBuilder.build();
+		}
+	}
+
+	/**
+	 * @return lists all Lucene query instances which should be applied as filters
+	 */
+	public List<Query> filterQueries() {
```
gsmet: Any reason why you didn't follow the getter naming convention?

Sanne: I guess that's because of a previous implementation: it didn't look like a getter at all.
This was more complex than just exposing the reference to a pre-built list.

Should I change it?

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchHSQueryImpl.java
```diff
@@ -363,9 +364,9 @@ private JsonObject getEffectiveFilter(JsonArray typeFilters) {
 			filters.add( ToElasticsearch.condition( "should", typeFilters ) );
 
 			// facet filters
-			Filter facetFilter = getFacetManager().getFacetFilter();
-			if ( facetFilter != null ) {
-				filters.add( ToElasticsearch.fromLuceneFilter( getFacetManager().getFacetFilter() ) );
+			QueryFilterset facetFilter = getFacetManager().getFacetFilterset();
+			if ( ! facetFilter.isEmpty() ) {
+				filters.add( ToElasticsearch.fromQueryFilterset( facetFilter ) );
```
gsmet: I think you can simply do something like:
```
for ( Query query : getFacetManager().getFacetFilterset().filterQueries() ) {
     filters.add( ToElasticSearch.fromLuceneQuery( query ) );
}
```

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ToElasticsearch.java
```diff
@@ -440,4 +443,28 @@ else if ( luceneFilter instanceof SpatialHashFilter ) {
 		throw LOG.cannotTransformLuceneFilterIntoEsQuery( luceneFilter );
 	}
 
+	/**
+	 * @param filterset a QueryFilterset
+	 * @return the Json representation of the filterset
+	 */
+	public static JsonObject fromQueryFilterset(QueryFilterset filterset) {
```
gsmet: Then this method becomes useless.

engine/src/main/java/org/hibernate/search/query/engine/impl/QueryFilterset.java
gsmet: Yes, I think it would be better if it's a simple getter. No need for a fluid method.

engine/src/main/java/org/hibernate/search/query/engine/impl/QueryFilterset.java
```diff
@@ -0,0 +1,70 @@
...
+		return filterQueries;
+	}
+
+	/**
+	 * @return <tt>true</tt> if this contains to filters to be applied.
```
gunnarmorling: Minor issue: Didn't we agree on using `{@code ...}`? I guess though I'd skip that comment altogether.

engine/src/main/java/org/hibernate/search/query/engine/impl/QueryFilterset.java
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Internal representation for a set of filters to be applied on a query.
+ *
+ * @author Sanne Grinovero
+ */
+public final class QueryFilterset {
+
+	public static final QueryFilterset EMPTY_FILTERSET = new QueryFilterset();
+
+	private final List<Query> filterQueries;
```
gunnarmorling: The name "...set" and this impl type (list) made me stumble. I'd align it, or better yet, change the class name into something more agnostic such as `FilteringQuery`.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchHSQueryImpl.java
Sanne: great point.

engine/src/main/java/org/hibernate/search/query/engine/impl/QueryFilterset.java
Sanne: hum it's not a `FilteringQuery` , that misses the more-than-one point.

But +1 to change the name.. easy though, it's meant an internal-only contract meant to be useful to migrate us away from using stacks of Filter(s).

I could as well have used just a `List<Query>`, but preferred to express it as an internal type so we have a good place for (future) helpers.

Sanne:
I have pushed a new version to address @gsmet 's feedback, but hadn't seen the comments from @gunnarmorling yet..

engine/src/main/java/org/hibernate/search/query/engine/impl/QueryFilterset.java
Sanne: I've copied that from the javadoc from List's `isEmpty()`.

Sanne:
re-pushed with improved names. I'll merge it if tests are green..

PR #1081, opened 2016-04-26
Services class loader cleanup

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-1724

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/service/impl/StandardServiceManager.java
+++ b/engine/src/main/java/org/hibernate/search/engine/service/impl/StandardServiceManager.java
@@ -42,6 +42,7 @@ public class StandardServiceManager implements ServiceManager {
 	private final ConcurrentHashMap<Class<?>, ServiceWrapper<?>> cachedServices = new ConcurrentHashMap<Class<?>, ServiceWrapper<?>>();
 	private final Map<Class<? extends Service>, Object> providedServices;
 	private final Map<Class<? extends Service>, String> defaultServices;
+	private final ClassLoaderService classloaderService;
 
 	private volatile boolean allServicesReleased = false;
 
@@ -56,6 +57,7 @@ public class StandardServiceManager implements ServiceManager {
 		this.properties = searchConfiguration.getProperties();
 		this.providedServices = createProvidedServices( searchConfiguration );
 		this.defaultServices = defaultServices;
+		this.classloaderService = searchConfiguration.getClassLoaderService();
 	}
 
 	@Override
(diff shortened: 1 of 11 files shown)


engine/src/main/java/org/hibernate/search/engine/service/impl/StandardServiceManager.java
```diff
@@ -44,6 +44,7 @@
 	private final Map<Class<? extends Service>, String> defaultServices;
 
 	private volatile boolean allServicesReleased = false;
+	private volatile ClassLoaderService classloaderService = null;
```
gunnarmorling: How about allocating it eagerly so to avoid the volatile reads?

gunnarmorling:
Hey, looks functionally alright, though I dislike the irregularity of this approach.

As an alternative, couldn't clients of `ClassLoaderService` keep an instance in a field so to avoid the repeated look-ups? That'd resemble a bit more how you'd do it with classic DI. It seemed nicer to me than the ugly special treatment of one of the service out of many.

engine/src/main/java/org/hibernate/search/engine/service/impl/StandardServiceManager.java
Sanne: ok, that should be doable. Remember a volatile read is not an issue on x86 though

Sanne:
Yes it's slightly irregular, but a step for a larger refactoring (explained on the mailing list).

Ultimately the `ClassLoaderService` will always be special: you can find other "special treatment code" in the `StandardServiceManager` around it, as the injection rules are different and because it's actually an essential component to load the other services...

It's not the "repeated lookups" which concern me solely. It's simply a core component and I don't see why we should have a more explicit helper for this; ultimately integrator.getClassLoaderService() seems more suited than having to handle the ServiceManager around to components which just need to load resources.

It's actually making things complex that we expose the `ServiceManager` to all these components as it is allowing circular dependencies among service(s) implementors: that's what I want to ban.

Sanne:
pushed an improved version on your suggestions

engine/src/main/java/org/hibernate/search/engine/service/impl/StandardServiceManager.java
```diff
@@ -283,4 +285,13 @@ private void stateExpectedFailure() {
 	private enum ServiceStatus {
 		RUNNING, STOPPED, STARTING, STOPPING
 	}
+
+	/* (non-Javadoc)
+	 * @see org.hibernate.search.engine.service.spi.ServiceManager#getClassLoaderService()
+	 */
+	@Override
```
gsmet: Minor nitpicking but maybe you should remove the useless javadoc?

Sanne:
> Minor nitpicking but maybe you should remove the useless javadoc?

Ok, then I'll merge it since we're getting close to nitpicking :)

PR #1082, opened 2016-04-26
HSEARCH-2243 Allow provided services to implement Startable or Stoppable

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2243


DavideD:
I think you also need to update the javadoc of `SearchConfiguration#getProvidedServices()`

engine/src/main/java/org/hibernate/search/engine/service/impl/StandardServiceManager.java
```diff
@@ -112,27 +112,9 @@ public void releaseAllServices() {
...
-				.entrySet() ) {
-			Object service = entry.getValue();
-			if ( service instanceof Startable ) {
-				throw log.providedServicesCannotImplementStartableOrStoppable(
-						service.getClass().getName(),
-						Startable.class.getName()
-				);
-			}
-			else if ( service instanceof Stoppable ) {
-				throw log.providedServicesCannotImplementStartableOrStoppable(
```
DavideD: This method in the Log interface is not used anymore, should we remove it?

engine/src/main/java/org/hibernate/search/engine/service/spi/ServiceManager.java
```diff
@@ -12,13 +12,13 @@
...
- * dependencies where a service where a service during bootstrapping could request other (uninitialized) services via
- * the {@link org.hibernate.search.engine.service.spi.Startable#start(java.util.Properties, org.hibernate.search.spi.BuildContext)}
- * callback.
+ * {@link org.hibernate.search.cfg.spi.SearchConfiguration#getClassLoaderService()}.
+ * </p>
+ * <p>
+ * It is the responsibility of the provider of these services to manage their life cycle: even if they implement
+ * {@link org.hibernate.search.engine.service.spi.Startable} or {@link org.hibernate.search.engine.service.spi.Stoppable},
+ * they will not be started or stopped automatically, while the methods on these interfaces will be invoked for
+ * the services discovered via the serviceloader.
```
DavideD: I'm not sure I understand this message, it says that they won't be started or stopped automatically but then state that the methods are going to get invoked during service discovery; am i missing something?

DavideD:
Merged https://github.com/hibernate/hibernate-search/pull/1082

I left a few rearks but I decided to merge it anyway.

PR #1083, opened 2016-04-27
HSEARCH-2046 Ignore IndexEmbedded attributes when used on collection of primitives

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-2046

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
+++ b/engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
@@ -102,6 +102,8 @@ import org.hibernate.search.util.logging.impl.LoggerFactory;
  */
 @SuppressWarnings( "deprecation" )
 public class AnnotationMetadataProvider implements MetadataProvider {
+	private static final int INFINITE_DEPTH = Integer.MAX_VALUE;
+
 	private static final Log log = LoggerFactory.make();
 	private static final StringBridge NULL_EMBEDDED_STRING_BRIDGE = DefaultStringBridge.INSTANCE;
 	private static final String UNKNOWN_MAPPED_BY_ROLE = "";
@@ -999,7 +1001,7 @@ public class AnnotationMetadataProvider implements MetadataProvider {
 	}
 
 	private void updateContainedInMetadata(ContainedInMetadataBuilder containedInMetadataBuilder, XProperty propertyWithContainedIn, String accessType) {
-		XClass memberReturnedType = propertyWithContainedIn.getElementClass();
+		XClass memberReturnedType = returnedType( propertyWithContainedIn );
 		String mappedBy = mappedBy( propertyWithContainedIn );
 		List<XProperty> returnedTypeProperties = memberReturnedType.getDeclaredProperties( accessType );
 		for ( XProperty property : returnedTypeProperties ) {
@@ -1298,7 +1300,7 @@ public class AnnotationMetadataProvider implements MetadataProvider {
 			return facetEncodingType; // encoding type explicitly set
 		}
 
-		Class<?> indexedType = reflectionManager.toClass( member.getElementClass() );
+		Class<?> indexedType = reflectionManager.toClass( returnedType( member ) );
 		if ( ReflectionHelper.isIntegerType( indexedType ) ) {
 			facetEncodingType = FacetEncodingType.LONG;
 		}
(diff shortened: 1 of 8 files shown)


engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
```diff
@@ -1772,6 +1753,68 @@ else if ( log.isTraceEnabled() ) {
...
 
+	private void logWarnings(XProperty member, IndexedEmbedded indexedEmbeddedAnnotation) {
+		if ( !isDepthNotSet( indexedEmbeddedAnnotation ) ) {
+			log.indexEmbeddedValueIgnored( member.getName(), "depth" );
+		}
+		if ( void.class != indexedEmbeddedAnnotation.targetElement() ) {
+			log.indexEmbeddedValueIgnored( member.getName(), "targetElement" );
+		}
+		if ( !isDefaultPrefix( indexedEmbeddedAnnotation ) ) {
+			log.indexEmbeddedValueIgnored( member.getName(), "prefix" );
```
Sanne: why do we ignore the prefix?

DavideD: In this context, what would you use it for?
I'm talking about this case:
```
@IndexEmbedded( prefix ="anothername" )
@ElementCollection
List<String> something
```

Sanne: I'm not sure. Just wondering: we have the @Field(name) and the prefix, so we _could_ opt to use it.

If we don't currently then +1 to not change this suddendly and have a nice warning.

DavideD: The current behaviour is odd, when the collection is null it will use the prefix as field name, otherwise it will use the field name. We could keep the prefix and add it to the field name, but I don't see any usecase where it would be useful.

Sanne: ok

gsmet:
@Sanne did you do a proper review of this one?

Thanks!

engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
```diff
@@ -1670,40 +1672,35 @@ private void checkForIndexedEmbedded(
...
-		XClass elementClass;
-		if ( void.class == indexedEmbeddedAnnotation.targetElement() ) {
-			elementClass = member.getElementClass();
-		}
-		else {
-			elementClass = reflectionManager.toXClass( indexedEmbeddedAnnotation.targetElement() );
-		}
+		XClass elementClass = elementCollection ? returnedType( member ) : elementClass( member, indexedEmbeddedAnnotation );
+		String localPrefix = elementCollection ? defaultPrefix( member ) : buildEmbeddedPrefix( prefix, indexedEmbeddedAnnotation, member );
+		boolean includeEmbeddedObjectId = elementCollection ? false : indexedEmbeddedAnnotation.includeEmbeddedObjectId();
```
gsmet: It might be just me so don't feel obliged to change it but I would have found it more readable with a global if (elementCollection) else, instead of 3 ternary operators.

gsmet:
@DavideD apart from my minor remark on the ternary operators, I think it's good to go.

Feel free to take into account this remark or not and merge it!

engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
```diff
@@ -1795,6 +1792,56 @@ else if ( log.isTraceEnabled() ) {
 		}
 	}
 
+	private void logWarnings(XProperty member, IndexedEmbedded indexedEmbeddedAnnotation) {
+		if ( !isDepthNotSet( indexedEmbeddedAnnotation ) ) {
```
gunnarmorling: Can you avoid the negated method name, making it `if ( isDepthSet(...) )`?

gunnarmorling:
I'd expect some ES tests to be unlocked by this (trying to rung {{ArrayBridgeTest}} against ES is how I originally learned about that issue)?

engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
DavideD: I prefer this way. Feel free to change it if you feel so strong about it (Maybe  after we have merged this).

DavideD: To clarify, I prefer this approach because it make it easier (for me) to figure out what's going on at a higher level.
I can see immediately that I'm trying to get the element class, the localPrefix and an object it. If you split it in two global if, I need to check first what happens in both blocks and then make sure that they are consistent and I haven't forgotten to set the default for one attribute.
Obviously, this is subjective.

DavideD:
> I'd expect some ES tests to be unlocked by this (trying to run ArrayBridgeTest against ES is how I originally learned about that issue)?

ArrayBridgeTest used to fail because of an additional field not defined in the schema, not I don't have that error anymore but hte tests fails because of an assertion. It seems it does not return the right amount of results. It's a different error that does not seem related to the issue I was trying to fix.
I would open a separte JIRA for it

DavideD:
I've renamed the method and merged this: https://github.com/hibernate/hibernate-search/pull/1083

DavideD:
I've created https://hibernate.atlassian.net/browse/HSEARCH-2271 to enable ArrayBridgeTest

PR #1084, opened 2016-04-27
HSEARCH-2244 Unify the way TermQuery for numerics are handled

gsmet:


engine/src/test/java/org/hibernate/search/test/backend/DeleteByQueryTest.java
```diff
@@ -175,24 +205,24 @@ private void testSerializationForQuery(DeletionQuery deletionQuery) {
...
+		list.add( new Work( new Book( 2, 2, 2L, 2F, 2D, "The Hobbit", "thehobbit" ), WorkType.ADD ) );
 		return list;
 	}
 
 	@Indexed(index = "books")
 	private static class Book {
 
 		@DocumentId
-		String id;
+		Integer id;
```
DavideD: Do you we have a test that checks deletion when the id is a String?
I'm a bit surprise that you changed an existing test instead of adding a new one.

DavideD:
@gsmet this looks good to me, I'm just surprised that you changed the existing tests instead of adding a new entity with an additional id. Maybe we already test that case somewhere else?

engine/src/test/java/org/hibernate/search/test/backend/DeleteByQueryTest.java
gsmet: It didn't test anything special with the string case related to id. That's why I moved the test for the string case to a standard attribute (note I added a url field for that).

gsmet:
OK. Will bring back the test and add a new one to be on the safe side. Then I'll merge it.

DavideD:
Great, thanks

gsmet:
Merged with the additional tests. Thanks for the review!

PR #1085, opened 2016-04-27
HSEARCH-2222 DSL, analyzers and Elasticsearch

gsmet:
Here is a short introduction to each commit:
### HSEARCH-2222 Clean up log message			9186802

This one is the easy one.

### 	HSEARCH-2222 Clean up API so that we know we manipulate references			d8418c7

OK, I've spent a week manipulating AnalyzerReference and Analyzer and it's definitely more clear when the API clearly states what you are manipulating. So this commit is a cleanup commit for that.

### 	HSEARCH-2222 Change the way ScopedAnalyzer are managed				72e2bcc

OK, this is probably the most important one. It changes completely the way ScopedAnalyzer are managed so that we are sure they are consistent and that, in the remote analyzer case, the API is clear when we want to get an analyzer name: it depends on the field name.

Also the lazy part is moved again to the analyzer level as I need to know if the analyzer is a Lucene one or a remote one even in the lazy case.

Last but not least, we provide sensible defaults for the remote analyzer case.

### 	HSEARCH-2222 Implement the DSL for Elasticsearch				59be211

OK, now is the time when we can use all the good stuff we coded before to get the DSL working with Elasticsearch.

### 	HSEARCH-2222 Enable DSL tests which are now working			fcf7007

Yeah, new tests passing \o/.


gsmet:
Pushed a new version taking into account Davide's comments. Renamed the analyzerIsRemote method to isAnalyzerRemote and squashed 2 commits.

After a good night of sleep, I came up to the conclusion that we can directly test if the analyzer is remote instead of relying on the index manager type (was not the case at a point in time while I was working on the patch) so the XXX are gone.

gsmet:
Rebased to latest master.

engine/src/main/java/org/hibernate/search/engine/impl/ConfigContext.java
```diff
@@ -323,23 +327,46 @@ private void addFilterDef(FullTextFilterDef defAnn) {
...
 	}
 
-	private void initRemoteAnalyzerReference(Map<String, AnalyzerReference> initializedAnalyzerReferences, LazyAnalyzerReference namedAnalyzerReference, IndexManagerHolder indexesFactory) {
+	private void initLazyRemoteAnalyzer(Map<String, AnalyzerReference> initializedAnalyzers, RemoteAnalyzerReference lazyRemoteAnalyzerReference,
+			IndexManagerHolder indexesFactory) {
+		LazyRemoteAnalyzer lazyAnalyzer = (LazyRemoteAnalyzer) lazyRemoteAnalyzerReference.getAnalyzer();
+
+		if ( initializedAnalyzers.containsKey( lazyAnalyzer.getName() ) ) {
+			AnalyzerReference analyzerReference = initializedAnalyzers.get( lazyAnalyzer.getName() );
+			if ( !( analyzerReference instanceof RemoteAnalyzerReference ) ) {
```
DavideD: No need to change it now but I was wondering if we need add this check.
Wouldn't it be possible to have a definition that applied both locally and remotely? For the simple case for example, like whitespace, simple, and so on

engine/src/main/java/org/hibernate/search/engine/metadata/impl/ParseContext.java
```diff
@@ -45,6 +46,24 @@ boolean skipFieldBridges() {
 		return indexManagerType == null;
 	}
 
+	/**
+	 * If the {@code IndexManager} type is not defined, we skip the {@code Analyzer} construction.
+	 */
+	boolean skipAnalyzers() {
+		return indexManagerType == null;
```
DavideD: I was wondering if this can really happen.
Shouldn't we have an exception in case the type is null?

But it seems related to existing code, so it is OK forthis  PR

engine/src/main/java/org/hibernate/search/util/logging/impl/Log.java
```diff
@@ -34,12 +45,6 @@
 import org.jboss.logging.annotations.Message;
 import org.jboss.logging.annotations.MessageLogger;
 
-import static org.jboss.logging.Logger.Level.DEBUG;
-import static org.jboss.logging.Logger.Level.ERROR;
-import static org.jboss.logging.Logger.Level.INFO;
-import static org.jboss.logging.Logger.Level.TRACE;
-import static org.jboss.logging.Logger.Level.WARN;
-
```
DavideD: Why I have the feeling the all our PR move the static imports up and down each time :)
(talking in general, no need to change)

gunnarmorling: We agreed to have them at the beginning, so it's good to do it, and the wrong usages will go away over time. Only if you see a move in the opposite direction, then complain :)

DavideD:
Rebased and merged https://github.com/hibernate/hibernate-search/pull/1085

Link to the issue: https://hibernate.atlassian.net/browse/HSEARCH-2222
Thanks

PR #1086, opened 2016-04-27
HSEARCH-2064 Implement filters in Elasticsearch backend

gsmet:
It's going to need some documentation but we might as well agree on the implementation before writing it.

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/ElasticsearchQueries.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/ElasticsearchQueries.java
@@ -7,10 +7,9 @@
 package org.hibernate.search.backend.elasticsearch;
 
 import org.hibernate.search.backend.elasticsearch.impl.ElasticsearchJsonQueryDescriptor;
+import org.hibernate.search.backend.elasticsearch.impl.GsonHolder;
 import org.hibernate.search.backend.elasticsearch.impl.JsonBuilder;
 import org.hibernate.search.query.engine.spi.QueryDescriptor;
-
-import com.google.gson.JsonParser;
 
 /**
  * Creates queries to be used with the Elasticsearch backend.
@@ -28,7 +27,7 @@ public class ElasticsearchQueries {
 	 * documentation</a> for the complete query syntax.
 	 */
 	public static QueryDescriptor fromJson(String jsonQuery) {
-		return new ElasticsearchJsonQueryDescriptor( new JsonParser().parse( jsonQuery ).getAsJsonObject() );
+		return new ElasticsearchJsonQueryDescriptor( GsonHolder.PARSER.parse( jsonQuery ).getAsJsonObject() );
 	}
 
 	/**
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/filter/SimpleElasticsearchFilter.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/filter/SimpleElasticsearchFilter.java
@@ -1,0 +1,29 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.elasticsearch.filter;
+
+/**
+ * Simple implementation of a native Elasticsearch full text filter.
+ *
+ * It only allows to define a static filter by calling the constructor.
+ *
+ * @author Guillaume Smet
+ */
+public class SimpleElasticsearchFilter implements ElasticsearchFilter {
+
+	private String jsonFilter;
+
+	public SimpleElasticsearchFilter(String jsonFilter) {
+		this.jsonFilter = jsonFilter;
+	}
+
+	@Override
+	public String getJsonFilter() {
+		return jsonFilter;
+	}
+
+}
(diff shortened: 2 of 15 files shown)


gunnarmorling:
What should we do about this one? Originally we meant to not support filters as they are going away in Lucene/ES soon. But now that the work has been done, we can integrate it very well, too.

Sanne:
> Originally we meant to not support filters as they are going away in Lucene/ES soon.

The concept of "filters" should remain in Hibernate Search I think. It makes sense how the API is expressed as it simplifies the way of enabling multiple filters programmatically on top of user-defined queries.
The `Filter` class within Lucene is deleted, but we can still model the same behaviour.

So what is going away in Hibernate Search is to expose the org.apache related classes but we can still expose most of it, e.g. it's nice that we have a declarative way to declare "filters".

Not least we'll need to work on the Lucene engine's internals to be able to still cache the filters, even though they are just boolean queries..

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/filter/SimpleElasticsearchFilter.java
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Simple implementation of a native Elasticsearch full text filter.
+ *
+ * It only allows to define a static filter by calling the constructor.
+ *
+ * @author Guillaume Smet
+ */
+public class SimpleElasticsearchFilter implements ElasticsearchFilter {
+
+	private String jsonFilter;
```
Sanne: I'll make this `final`

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchHSQueryImpl.java
```diff
@@ -367,6 +353,20 @@ private JsonObject getEffectiveFilter(JsonArray typeFilters) {
...
+				for ( FullTextFilterImpl fullTextFilter : filterDefinitions.values() ) {
+					JsonObject filter = buildFullTextFilter( fullTextFilter );
+					if ( filter != null ) {
+						filters.add( filter );
+					}
+				}
+			}
+
 			// wrap filters into must if there is more than one
 			return ToElasticsearch.condition( "must", filters );
```
Sanne: in the example I see here:
 - https://www.elastic.co/guide/en/elasticsearch/reference/2.0/query-filter-context.html
the various filters are not connected over a "must" element.

This doesn't seem wrong, but it might be redundant right? Just wondering if you noticed that, or if there is some additional reason.

gsmet: Indeed. In this case, we could simply return the JsonArray.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/ElasticsearchQueries.java
```diff
@@ -28,7 +27,7 @@ private ElasticsearchQueries() {
 	 * documentation</a> for the complete query syntax.
 	 */
 	public static QueryDescriptor fromJson(String jsonQuery) {
-		return new ElasticsearchJsonQueryDescriptor( new JsonParser().parse( jsonQuery ).getAsJsonObject() );
+		return new ElasticsearchJsonQueryDescriptor( GsonHolder.PARSER.parse( jsonQuery ).getAsJsonObject() );
```
gsmet: @Sanne in fact, this is where I added a reference to the GsonHolder. So maybe, we should stick with the instantiation instead?

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/filter/SimpleElasticsearchFilter.java
gsmet: Yes, thanks!

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/ElasticsearchQueries.java
Sanne: yes I noticed that, but I prefer your change (i.e. keep the static reference), it's just reassuring that in case it turns out the component will be leaking in future, then we can revert to instantiation.

BTW it's not the PARSER the one which uses tlocals, but the encoder (the "GsonHolder.GSON" )

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchHSQueryImpl.java
```diff
@@ -739,6 +739,70 @@ protected void extractFacetResults() {
...
+				throw new SearchException(
+						"Filter implementation does not implement the Filter interface or does not extend ElasticsearchFilter: "
+								+ def.getImpl().getName() + ". "
+								+ ( def.getFactoryMethod() != null ? def.getFactoryMethod().getName() : "" ) );
+			}
+		}
+
+		if ( jsonFilter != null &&
+				( FilterCacheModeType.NONE.equals( def.getCacheMode() ) || FilterCacheModeType.INSTANCE_ONLY.equals( def.getCacheMode() ) ) ) {
+			jsonFilter.entrySet().iterator().next().getValue().getAsJsonObject().addProperty( "_cache", false );
```
Sanne: Looks like the "_cache" capabilities are deprecated:
 - https://www.elastic.co/guide/en/elasticsearch/reference/2.3/breaking_20_query_dsl_changes.html#_filter_auto_caching

Sanne:
Merged!

FYI I removed adding the "_cache" attribute.

PR #1089, opened 2016-04-27
Improve date handling for Elasticsearch backend

gsmet:
Bikeshedding for the name of KeepSafeFromAnalyzerBridge very welcome!

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchBridgeProvider.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchBridgeProvider.java
@@ -6,28 +6,27 @@
  */
 package org.hibernate.search.backend.elasticsearch.impl;
 
+import java.lang.reflect.AnnotatedElement;
 import java.util.Calendar;
 import java.util.Date;
-import java.util.Locale;
-import java.util.TimeZone;
 
-import javax.xml.bind.DatatypeConverter;
-
-import org.apache.lucene.document.Document;
+import org.hibernate.search.annotations.DateBridge;
+import org.hibernate.search.annotations.Resolution;
 import org.hibernate.search.bridge.FieldBridge;
-import org.hibernate.search.bridge.LuceneOptions;
-import org.hibernate.search.bridge.TwoWayFieldBridge;
 import org.hibernate.search.bridge.impl.ExtendedBridgeProvider;
 import org.hibernate.search.bridge.spi.IndexManagerTypeSpecificBridgeProvider;
 import org.hibernate.search.indexes.spi.IndexManager;
+import org.hibernate.search.util.logging.impl.Log;
+import org.hibernate.search.util.logging.impl.LoggerFactory;
 
 /**
  * Creates bridges specific to ES.
  *
  * @author Gunnar Morling
+ * @author Guillaume Smet
  */
-// TODO Handle Calendar
 public class ElasticsearchBridgeProvider extends ExtendedBridgeProvider implements IndexManagerTypeSpecificBridgeProvider {
+	private static final Log LOG = LoggerFactory.make();
 
 	@Override
 	public Class<? extends IndexManager> getIndexManagerType() {
@@ -35,47 +34,71 @@ public class ElasticsearchBridgeProvider extends ExtendedBridgeProvider implemen
 	}
 
 	@Override
-	public FieldBridge provideFieldBridge(ExtendedBridgeProviderContext bridgeContext) {
-		if ( isDate( bridgeContext.getReturnType() ) ) {
-			return EsDateBridge.INSTANCE;
+	public FieldBridge provideFieldBridge(ExtendedBridgeProviderContext context) {
+		AnnotatedElement annotatedElement = context.getAnnotatedElement();
+		if ( Date.class.isAssignableFrom( context.getReturnType() ) ) {
+			if ( annotatedElement.isAnnotationPresent( DateBridge.class ) ) {
+				DateBridge dateBridgeAnnotation = annotatedElement.getAnnotation( DateBridge.class );
+
+				return getDateFieldBridge( dateBridgeAnnotation.resolution() );
+			}
+			else {
+				return getDateFieldBridge( Resolution.MILLISECOND );
+			}
+		}
+		else if ( Calendar.class.isAssignableFrom( context.getReturnType() ) ) {
+			if ( annotatedElement.isAnnotationPresent( DateBridge.class ) ) {
+				DateBridge dateBridgeAnnotation = annotatedElement.getAnnotation( DateBridge.class );
+
+				return getCalendarFieldBridge( dateBridgeAnnotation.resolution() );
+			}
+			else {
+				return getCalendarFieldBridge( Resolution.MILLISECOND );
+			}
 		}
 
 		return null;
 	}
 
-	private boolean isDate(Class<?> clazz) {
-		return "java.util.Date".equals( clazz.getName() );
+	private FieldBridge getDateFieldBridge(Resolution resolution) {
+		switch ( resolution ) {
+			case YEAR:
+				return ElasticsearchDateBridge.DATE_YEAR;
+			case MONTH:
+				return ElasticsearchDateBridge.DATE_MONTH;
+			case DAY:
+				return ElasticsearchDateBridge.DATE_DAY;
+			case HOUR:
+				return ElasticsearchDateBridge.DATE_HOUR;
+			case MINUTE:
+				return ElasticsearchDateBridge.DATE_MINUTE;
+			case SECOND:
+				return ElasticsearchDateBridge.DATE_SECOND;
+			case MILLISECOND:
+				return ElasticsearchDateBridge.DATE_MILLISECOND;
+			default:
+				throw LOG.unknownResolution( resolution.toString() );
+		}
 	}
 
-	// TODO Handle resolution
-	private static class EsDateBridge implements TwoWayFieldBridge {
-
-		private static EsDateBridge INSTANCE = new EsDateBridge();
-
-		@Override
-		public void set(String name, Object value, Document document, LuceneOptions luceneOptions) {
-			if ( value == null ) {
-				return;
-			}
-
-			luceneOptions.addFieldToDocument( name, convertToString( (Date) value ), document );
-		}
-
-		@Override
-		public Object get(String name, Document document) {
-			Calendar c = DatatypeConverter.parseDateTime( document.get( name ) );
-			return c.getTime();
-		}
-
-		@Override
-		public String objectToString(Object object) {
-			return convertToString( (Date) object );
-		}
-
-		private String convertToString(Date value) {
-			Calendar c = Calendar.getInstance( TimeZone.getTimeZone( "UTC" ), Locale.ENGLISH );
-			c.setTime( value );
-			return DatatypeConverter.printDateTime( c );
+	private FieldBridge getCalendarFieldBridge(Resolution resolution) {
+		switch ( resolution ) {
+			case YEAR:
+				return ElasticsearchCalendarBridge.DATE_YEAR;
+			case MONTH:
+				return ElasticsearchCalendarBridge.DATE_MONTH;
+			case DAY:
+				return ElasticsearchCalendarBridge.DATE_DAY;
+			case HOUR:
+				return ElasticsearchCalendarBridge.DATE_HOUR;
+			case MINUTE:
+				return ElasticsearchCalendarBridge.DATE_MINUTE;
+			case SECOND:
+				return ElasticsearchCalendarBridge.DATE_SECOND;
+			case MILLISECOND:
+				return ElasticsearchCalendarBridge.DATE_MILLISECOND;
+			default:
+				throw LOG.unknownResolution( resolution.toString() );
 		}
 	}
 }
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchCalendarBridge.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchCalendarBridge.java
@@ -18,6 +18,7 @@ import org.hibernate.search.bridge.LuceneOptions;
 import org.hibernate.search.bridge.ParameterizedBridge;
 import org.hibernate.search.bridge.TwoWayFieldBridge;
 import org.hibernate.search.bridge.builtin.impl.DateResolutionUtil;
+import org.hibernate.search.bridge.spi.IgnoreAnalyzerBridge;
 
 /**
  * Bridge a {@code java.util.Calendar} to a {@code String} using the ISO 8601 standard which is the default date format
@@ -27,7 +28,7 @@ import org.hibernate.search.bridge.builtin.impl.DateResolutionUtil;
  *
  * @author Guillaume Smet
  */
-public class ElasticsearchCalendarBridge implements TwoWayFieldBridge, ParameterizedBridge {
+public class ElasticsearchCalendarBridge implements TwoWayFieldBridge, ParameterizedBridge, IgnoreAnalyzerBridge {
 
 	public static final ElasticsearchCalendarBridge DATE_YEAR = new ElasticsearchCalendarBridge( Resolution.YEAR );
 	public static final ElasticsearchCalendarBridge DATE_MONTH = new ElasticsearchCalendarBridge( Resolution.MONTH );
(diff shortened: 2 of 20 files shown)


gsmet:
FYI, I force pushed a few more tests/validation for the calendar case.

DavideD:
Wouldn't it be posssible to reuse `StringEncodingDateBridge` and `StringEncodingCalendarBridge`?

gsmet:
As the date format is not the same, we would have to make the formatting function overridable.

I'm not sure the code will be better and more maintainable.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/util/impl/ElasticsearchDateHelper.java
```diff
@@ -0,0 +1,56 @@
...
+import javax.xml.bind.DatatypeConverter;
+
+/**
+ * Various utilities to manipulate dates and comply with Elasticsearch date format.
+ *
+ * @author Guillaume Smet
+ */
+public final class ElasticsearchDateHelper {
+
+	private static final TimeZone ENCODING_TIME_ZONE = TimeZone.getTimeZone( "UTC" );
```
DavideD: It probably doesn't make any difference but the @DateBridge annotation javadoc states the we use GMT as timezone and it seems that it is what we do in the Lucene backend as well.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/util/impl/ElasticsearchDateHelper.java
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	public static String dateToString(Date date) {
+		Calendar c = Calendar.getInstance( ENCODING_TIME_ZONE, Locale.ENGLISH );
+		c.setTime( date );
+		return DatatypeConverter.printDateTime( c );
+	}
+
+	public static Calendar dateToCalendar(Date date) {
+		Calendar calendar = Calendar.getInstance( ENCODING_TIME_ZONE, Locale.ROOT );
```
DavideD: in `dateToString` the Locale used is `ENGLISH` here is `ROOT`, is there a reason for it or can we stick with one?

gsmet: Probably inspired by 2 different conversion codes.

Will fix.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/util/impl/ElasticsearchDateHelper.java
gsmet: It was preexisting in Gunnar's code. Will switch to GMT though.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/util/impl/ElasticsearchDateHelper.java
gsmet: Mmmh, no, in fact, it's voluntary.

From the Es doc: "In JSON documents, dates are represented as strings. Elasticsearch uses a set of preconfigured formats to recognize and parse these strings into a long value representing milliseconds-since-the-epoch in UTC.".

In practice, it's the same but we might as well be in line with the doc.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/util/impl/ElasticsearchDateHelper.java
DavideD: I think it's because Elasticsearch uses UTC by default (if I understand correctly). I'm not sure if this is going to have an impact.

gsmet: Mmmh, no, in fact, it's voluntary.

From the Es doc: "In JSON documents, dates are represented as strings. Elasticsearch uses a set of preconfigured formats to recognize and parse these strings into a long value representing milliseconds-since-the-epoch in UTC.".

In practice, it's the same but we might as well be in line with the doc.

engine/src/main/java/org/hibernate/search/bridge/KeepSafeFromAnalyzerBridge.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+package org.hibernate.search.bridge;
+
+
+/**
+ * This interface is used to mark the bridges we want to keep safe from the analyzers.
+ *
+ * @author Guillaume Smet
+ */
+public interface KeepSafeFromAnalyzerBridge {
```
DavideD: if I understand correctly this interface is used in DSL when we shouldn't apply an analyzer (basically is the same as calling `.ignoreAnalyzer()`). Am I right?

Shouldn't this be in the package: `org.hibernate.search.query.dsl[.impl]`

DavideD: As an alternative name I would suggest `IgnoreAnalyzer` or `SkipAnalysis`

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchCalendarBridge.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.util.Calendar;
+import java.util.Date;
+
+import org.apache.lucene.document.Document;
+import org.hibernate.search.annotations.Resolution;
+import org.hibernate.search.backend.elasticsearch.util.impl.ElasticsearchDateHelper;
+import org.hibernate.search.bridge.LuceneOptions;
+
+public class ElasticsearchCalendarBridge extends ElasticsearchDateBridge {
```
DavideD: Some Javadoc with the format used to store the dates would be nice

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchDateBridge.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.lucene.document.Document;
+import org.hibernate.search.annotations.Resolution;
+import org.hibernate.search.backend.elasticsearch.util.impl.ElasticsearchDateHelper;
+import org.hibernate.search.bridge.KeepSafeFromAnalyzerBridge;
+import org.hibernate.search.bridge.LuceneOptions;
+import org.hibernate.search.bridge.ParameterizedBridge;
+import org.hibernate.search.bridge.TwoWayFieldBridge;
+import org.hibernate.search.bridge.builtin.impl.DateResolutionUtil;
+
+public class ElasticsearchDateBridge implements TwoWayFieldBridge, ParameterizedBridge, KeepSafeFromAnalyzerBridge {
```
DavideD: Javadoc

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/util/impl/ElasticsearchDateHelper.java
gsmet: In practice, there's no impact as UTC and GMT share the same current time.

Moreover, we only need to be consistent for a given backend and all the manipulations are done with the FieldBridges so we are on the safe side.

engine/src/main/java/org/hibernate/search/bridge/KeepSafeFromAnalyzerBridge.java
gsmet: Yes, you are. Yeah maybe IgnoreAnalyzerBridge is fine and mirrors the ignoreAnalyzer() method.

As for the package, I don't know what is the rule of thumb for an interface used by other public interfaces.

@Sanne ?

DavideD:
As a separate issue or commit, shouldn't we apply `KeepSafeFromAnalyzer` to the `StringEncondingDateBridge` for Lucene?

we will need to update the `@DateBridge` and `CalendarBrdige` javadoc as the timezone is now backend dependent and add some explanation in the documentation as well.

I think, that's all for my review, I'll wait for your response on the remarks about `KeepSafeFromAnalyzer`.

engine/src/main/java/org/hibernate/search/bridge/KeepSafeFromAnalyzerBridge.java
DavideD: I didn't realize that the other bridges are in a public package, `org.hibernate.search.query.dsl` is still public so it should do, but let's see what @Sanne  thinks

gsmet:
> As a separate issue or commit, shouldn't we apply KeepSafeFromAnalyzer to the StringEncondingDateBridge for Lucene?

Yes, forgot this one. Done.

> we will need to update the @DateBridge and CalendarBrdige javadoc as the timezone is now backend dependent and add some explanation in the documentation as well.

I just changed GMT to GMT/UTC. As it's in practice the same thing, I think it will be more noise to add something about the difference between backends.

> I think, that's all for my review, I'll wait for your response on the remarks about KeepSafeFromAnalyzer.

OK, fixed all your remarks.

Waiting for @Sanne to decide where we put the IgnoreAnalyzerBridge interface.

DavideD:
>> As a separate issue or commit, shouldn't we apply KeepSafeFromAnalyzer to the StringEncondingDateBridge for Lucene?

> Yes, forgot this one. Done.

In this case you should also add `StringEncondingCalendarBridge`  I guess

gsmet:
> In this case you should also add StringEncondingCalendarBridge I guess

No, it extends StringEncodingDateBridge so we're good.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchBridgeProvider.java
```diff
@@ -6,76 +6,103 @@
...
-	}
-
-	// TODO Handle resolution
-	private static class EsDateBridge implements TwoWayFieldBridge {
-
-		private static EsDateBridge INSTANCE = new EsDateBridge();
+	public FieldBridge provideFieldBridge(ExtendedBridgeProviderContext context) {
+		AnnotatedElement annotatedElement = context.getAnnotatedElement();
+		if ( Date.class.isAssignableFrom( context.getReturnType() ) ) {
+			if ( annotatedElement.isAnnotationPresent( org.hibernate.search.annotations.DateBridge.class ) ) {
```
gunnarmorling: Is there any strong need for the FQN?

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchCalendarBridge.java
```diff
@@ -0,0 +1,70 @@
...
+
+import org.apache.lucene.document.Document;
+import org.hibernate.search.annotations.Resolution;
+import org.hibernate.search.backend.elasticsearch.util.impl.ElasticsearchDateHelper;
+import org.hibernate.search.bridge.LuceneOptions;
+
+/**
+ * Bridge a {@code java.util.Calendar} to a {@code String} using the ISO 8601 standard which is the default date format
+ * of Elasticsearch.
+ * UTC is used as time zone.
```
gunnarmorling: Did you consider to store the TZ offset instead (it's my understanding that ES supports formats with a TZ offset)? That way the calendar's TZ info can e.g. be shown when projecting the source document (or just a single field). That'd more be what I'd expect as a user.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchCalendarBridge.java
```diff
@@ -0,0 +1,70 @@
...
+	@Override
+	public String objectToString(Object object) {
+		Date date = null;
+		if ( object != null ) {
+			date = ( (Calendar) object ).getTime();
+		}
+		return super.objectToString( date );
+	}
+
+}
```
gunnarmorling: Nit-picking: By convention we do a line break at EOF. Actually we used to enforce it via CS but gave up on it because that check always was brittle.

engine/src/main/java/org/hibernate/search/bridge/IgnoreAnalyzerBridge.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+package org.hibernate.search.bridge;
+
+
+/**
+ * This interface is used to mark the bridges we want to keep safe from the analyzers.
+ *
+ * @author Guillaume Smet
+ */
+public interface IgnoreAnalyzerBridge {
```
gunnarmorling: Should this be a public contract right away? If so, should it be marked as experimental? I'd expect this to be morphed into a method on `FieldBridge` a la `boolean applyAnalyzers()` down the road instead of it being a marker interface?

gunnarmorling:
@gsmet, looks good overall, see inline for some comments. Most notabily, I'd store the TZ offset in ES document fields for `Calendar` properties to facilitate projections (in the index, it will still be converted into millis-since-epoch).

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchBridgeProvider.java
gsmet: Nope, just a copy/paste from the Lucene code, will fix.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchCalendarBridge.java
gsmet: Done.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchCalendarBridge.java
gsmet: Done.

engine/src/main/java/org/hibernate/search/bridge/IgnoreAnalyzerBridge.java
gsmet: Marked as experimental, moved to spi. Is it what you had in mind?

gsmet:
@gunnarmorling I think I addressed all your comments. We should be ready to go!

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/util/impl/ElasticsearchDateHelper.java
```diff
@@ -0,0 +1,92 @@
...
+
+	public static Calendar stringToCalendar(String value) {
+		return DatatypeConverter.parseDateTime( value );
+	}
+
+	/**
+	 * Limit a calendar resolution.
+	 *
+	 * @param resolution The desired resolution of the date to be returned
+	 * @return the calendar with all values more precise than <code>resolution</code> set to 0 or 1
```
gunnarmorling: A while ago we agreed on using `{@code ...}` instead of `<code>...`. I'll change it myself before applying it, just bringing it up to make sure we all are on the same page.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/util/impl/ElasticsearchDateHelper.java
```diff
@@ -0,0 +1,92 @@
...
+		return c.getTime();
+	}
+
+	public static String dateToString(Date date) {
+		Calendar c = Calendar.getInstance( ENCODING_TIME_ZONE, Locale.ENGLISH );
+		c.setTime( date );
+		return DatatypeConverter.printDateTime( c );
+	}
+
+	public static Calendar dateToCalendar(Date date) {
```
gunnarmorling: This one seems unused. I'll remove it.

elasticsearch/src/test/java/org/hibernate/search/backend/elasticsearch/test/ElasticsearchIndexMappingIT.java
```diff
@@ -126,6 +131,7 @@ public void testMapping() throws Exception {
 				"{" +
 					"\"active\": true," +
 					"\"dateOfBirth\": \"1958-04-07T00:00:00Z\"," +
+					"\"subscriptionEndDate\": \"2016-06-07T00:00:00+02:00\"," +
```
gunnarmorling: Nice!

gunnarmorling:
Rebased, applied and added two more commits for some polishing. Let me know if you see any issues with those, @gsmet. Thanks!

PR #1090, opened 2016-04-27
HSEARCH-2247 Represent the kind of underlying technology with an expl…

Sanne:
…icit type

 - https://hibernate.atlassian.net/browse/HSEARCH-2247

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchBridgeProvider.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchBridgeProvider.java
@@ -14,12 +14,13 @@ import java.util.TimeZone;
 import javax.xml.bind.DatatypeConverter;
 
 import org.apache.lucene.document.Document;
+import org.hibernate.search.backend.elasticsearch.spi.ElasticsearchIndexManagerType;
 import org.hibernate.search.bridge.FieldBridge;
 import org.hibernate.search.bridge.LuceneOptions;
 import org.hibernate.search.bridge.TwoWayFieldBridge;
 import org.hibernate.search.bridge.impl.ExtendedBridgeProvider;
 import org.hibernate.search.bridge.spi.IndexManagerTypeSpecificBridgeProvider;
-import org.hibernate.search.indexes.spi.IndexManager;
+import org.hibernate.search.indexes.spi.IndexManagerType;
 
 /**
  * Creates bridges specific to ES.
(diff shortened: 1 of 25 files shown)


gsmet:
You know a patch is right when it looks like it should have been like that from the beginning.

@gunnarmorling no objections before me merging this?

gunnarmorling:
It looks good to me, a bit cleaner than the current approach.

Though it's not fully clear to me what the actual benefits over the current one are (apart aesthetics): Do you expect another "backend" targeting Elasticsearch which actually not extends `ElasticsearchIndexManager`? If so, that seems quite unlikely to me. In the end it all depends on how that type information is going to be used. As said before, I'd prefer to have callbacks into the specific implementation rather than doing switching logic based on this in core.

One other thing: I'd add a new annotation `@Incubating` and add this to this new type, method etc. This can be done separately if we all agree that's a good thing to do (which I hope :).

gsmet:
@gunnarmorling yes, the `@Incubating` annotation looks like a good idea.

DavideD:
Like this one? https://hibernate.atlassian.net/browse/HHH-10487

DavideD:
> One other thing: I'd add a new annotation @Incubating and add this to this new type, method etc. This can be done separately if we all agree that's a good thing to do (which I hope :).

It sounds nice, how does  it relate with the `@hsearch.experimental` tag in the javadoc?

Sanne:
Sounds ok, but I didn't see a strong need for it. Remember this is an SPI and we reserve the right to change them in any minor version..

Sanne:
> Though it's not fully clear to me what the actual benefits over the current one are (apart aesthetics): Do you expect another "backend" targeting Elasticsearch which actually not extends ElasticsearchIndexManager? If so, that seems quite unlikely to me. In the end it all depends on how that type information is going to be used. As said before, I'd prefer to have callbacks into the specific implementation rather than doing switching logic based on this in core.

I agree, but the callbacks for such things can be organized on the proposed `IndexManagerType` interface.

The interface is empty today but I'm considering it a place holder for evolution, as needed.

And yes, I expect people to want to make their own implementation of `ElasticsearchIndexManager`.. not least, I'm thinking about when we'll add at least Solr. Some things might end up being shared, i.e. we need a fined grained granularity on "capabilities" than looking at the implementation via an `instanceof`.

gunnarmorling:
> I agree, but the callbacks for such things can be organized on the proposed IndexManagerType interface.

Yes, if that's the intentention, I like it.

> Remember this is an SPI and we reserve the right to change them in any minor version..

Ah, right, though I kind of dislike this approach.

It easily leads us into discussions whether certain SPIs are good to break or not, because the limits are not precisely described (consider my refactoring PR as an example). I'd rather suggest to commit to stability in micro/minor for everything meant for public use (no matter whether API or SPI) and explicitly tag those contracts currrently evolving with `@Incubating` or similar, signalling that breakages may occur.

It's much more conscious and transparent (for users *and* us) that way.

gunnarmorling:
Superseded by #1098.

Sanne:
> It's much more conscious and transparent (for users and us) that way.

Yes that sounds better, the SPI stuff has never been very clear.

PR #1091, opened 2016-04-28
HSEARCH-2250 Improve readability and efficiency of MultiWriteDrainabl…

Sanne:
…eLinkedList

 - https://hibernate.atlassian.net/browse/HSEARCH-2250


Sanne:
to simplify the review:
I only replaced the usage of ReentrantLock with synchronized. All we needed was exclusive lock, so synchronized gives us the same semantic and might be optimized better.

The two methods which used the lock in a try/finally block are now broken up, each in  two smaller methods. That's to keep the synchonized at the smallest possible level, and in practice at the same level as it was when using the lock.

gunnarmorling:
Rebased and applied. Added one more commit for some clean-up. Thanks, @Sanne.

PR #1093, opened 2016-04-29
HSEARCH-2063 Review the pending TODOs

gsmet:


elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -665,7 +665,7 @@ public void performStreamOperation(LuceneWork singleOperation, IndexingMonitor m
 
 	@Override
 	public void optimize() {
-		// TODO Is there such thing for ES?
+		performStreamOperation( OptimizeLuceneWork.INSTANCE, null, false );
```
DavideD: Am I wrong or there is  seprate issue for this?

DavideD:
Rebased and merrged https://github.com/hibernate/hibernate-search/pull/1093

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchIndexManager.java
gsmet: Yes, there is one for ElasticsearchIndexWorkVisitor as we need to implement it. Thought I might as well fix the TODO here as it was a one liner.

DavideD: Next time, please, let's keep it separated, or at least in separate commits.

Is there a way to test this?

The issue I was talking about is this one: https://hibernate.atlassian.net/browse/HSEARCH-2092

gsmet: Yes, HSEARCH-2092 is the one I was talking about related to ElasticsearchIndexWorkVisitor.

This change doesn't do anything indeed as we still have to implement the force merge in ElasticsearchIndexWorkVisitor. That's why I considered it was more a TODO removal than adding a feature. I just put the code here so that the force merge only has to be implemented at one place.

Looks OK to you?

DavideD: ok

PR #1094, opened 2016-04-29
HSEARCH-2178 Using bulk requests during mass indexing

gunnarmorling:
Ok, that one was a true beast...

I recommend to review the commits one by one. The first and the third commit are preparatory and clean-up work, respectively. https://github.com/hibernate/hibernate-search/pull/1094/commits/fce174b50516dd1075536a2ca8bf6413180a3b56 is the interesting one.

The idea is to re-use `MultiWriteDrainableLinkedList` to maintain a queue of requests submitted via `performStreamOperation()`. A worker runnable takes all the messages which are in the queue at a given time and creates a single bulk request from that (as far as the requests are bulkable). All this happens in `BackendRequestProcessor`, which is the common facade for running sync and async requests.

I hope I got all the concurrency right, let me know if you spot any issues or room for significant improvement.


gunnarmorling:
Added one more commit to limit the number of requests per bulk. The value is not configurable for now but it can easily be done should the need arise.

Sanne:
Great work! I could honestly not grasp all of the concurrency cases when travelling on my laptop, as it was quite different than what I'd have done but it looks all good.. I had to run it and debug it though :+1: 

I found a couple of non-core issues so took the liberty to fix them in follow up commits.. simple to address leaks like not releasing the service and not shutting down the Executor.
While at it, I also changed the bulking logic applied when `RequestProcessingRunnable` is spinning to postpone index refresh.. seemed safe enough to push, but would be nice if you could post-check it.

I would not have shaped these components as `Service` as it looks like it will make it harder to change the Jest Service back from being a service too.. but we can face that later.

Merged

gunnarmorling:
> Merged

Awesome :)

> ... not shutting down the Executor.

Oops, good catch.

> While at it, I also changed the bulking logic applied when RequestProcessingRunnable is spinning to postpone index refresh

Nice idea, looks good. I didn't care about it too much as I see the refresh as an test-only option (see the other PR and its discussion), but that's a good improvement either way.

> ... but we can face that later.

Right, we can easily re-adjust this sort of thing as needed later, so +1 for moving on for now.

Thanks!

PR #1096, opened 2016-05-02
HSEARCH-2252 Document that it is not recommended to sort by id

gsmet:
You need to define another field specifically for sorting.

### DIFF
--- a/documentation/src/main/asciidoc/mapping.asciidoc
+++ b/documentation/src/main/asciidoc/mapping.asciidoc
@@ -430,6 +430,12 @@ public class Essay {
 specification. The `grade` field is annotated as Numeric with a slightly larger precisionStep than
 the default.
 
+[NOTE]
+====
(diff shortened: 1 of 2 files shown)


documentation/src/main/asciidoc/mapping.asciidoc
```diff
@@ -430,6 +430,12 @@ public class Essay {
 specification. The `grade` field is annotated as Numeric with a slightly larger precisionStep than
 the default.
 
+[NOTE]
+====
+It is not recommended to sort on the id field as it will lead to erratic behaviors. If you want to sort on your
```
gunnarmorling: Is that true for all backends?

gsmet: It's true for all the backends we support currently.

gunnarmorling:
Rebased and applied. Thanks, @gsmet.

PR #1100, opened 2016-05-03
HSEARCH-2138 Not refreshing indexes by default

gunnarmorling:
https://hibernate.atlassian.net/browse/HSEARCH-2138

~~Based on top of #1096 (only https://github.com/hibernate/hibernate-search/commit/d55e5176def9704440e8611f300f961ee0f02bf4 is the meat of this change).~~

ES refreshes indexes by default every second, making index changes visible almost immediately. That should be fine for most production purposes, but during our testing it's not enough.

Until now, we always enforced a refresh after writes towards an index in order to make the test suite run reliable. This change makes this an option, which is disabled by default and only enabled during our tests. While not perfect, it should be a good compromise between offering proper speed in practice and making our test suite run reliable.

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/cfg/ElasticsearchEnvironment.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/cfg/ElasticsearchEnvironment.java
@@ -22,6 +22,7 @@ public final class ElasticsearchEnvironment {
 		public static final IndexManagementStrategy INDEX_MANAGEMENT_STRATEGY = IndexManagementStrategy.NONE;
 		public static final int INDEX_MANAGEMENT_WAIT_TIMEOUT = 10_000;
 		public static final String REQUIRED_INDEX_STATUS = "green";
+		public static final boolean REFRESH_AFTER_WRITE = false;
 	}
 
 	/**
@@ -73,6 +74,19 @@ public final class ElasticsearchEnvironment {
 	 * specific indexes (e.g. {@code hibernate.search.someindex.elasticsearch.required_index_status=yellow}).
 	 */
 	public static final String REQUIRED_INDEX_STATUS = "elasticsearch.required_index_status";
+
+	/**
+	 * Property for specifying whether an explicit index refresh should be issued after a set of operations targeting a
+	 * given index has been executed or not.
+	 * <p>
+	 * A boolean value (true, false) is expected.
+	 * <p>
+	 * Defaults to {@link Defaults#REFRESH_AFTER_WRITE}.
+	 * <p>
+	 * Can be given globally (e.g. {@code hibernate.search.default.elasticsearch.refresh_after_write=false}) or for
+	 * specific indexes (e.g. {@code hibernate.search.someindex.elasticsearch.refresh_after_write=true}).
+	 */
+	public static final String REFRESH_AFTER_WRITE = "elasticsearch.refresh_after_write";
 
 	private ElasticsearchEnvironment() {
 	}
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/BackendRequest.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/BackendRequest.java
@@ -28,11 +28,13 @@ public class BackendRequest<T extends JestResult> {
 	private final Set<Integer> ignoredErrorStatuses;
 	private final LuceneWork luceneWork;
 	private final String indexName;
+	private final boolean refreshAfterWrite;
 
-	public BackendRequest(Action<T> action, LuceneWork luceneWork, String indexName, int... ignoredErrorStatuses) {
+	public BackendRequest(Action<T> action, LuceneWork luceneWork, String indexName, boolean refreshAfterWrite, int... ignoredErrorStatuses) {
 		this.action = action;
 		this.luceneWork = luceneWork;
 		this.indexName = indexName;
+		this.refreshAfterWrite = refreshAfterWrite;
 		this.ignoredErrorStatuses = asSet( ignoredErrorStatuses );
 	}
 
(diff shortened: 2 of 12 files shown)


elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/cfg/ElasticsearchEnvironment.java
```diff
@@ -74,6 +75,19 @@
 	 */
 	public static final String REQUIRED_INDEX_STATUS = "elasticsearch.required_index_status";
 
+	/**
+	 * Property for specifying whether and explicit index refresh should be issued after a set of operations targeting a
+	 * given index has been executed or nor not.
```
gsmet: whether and -> whether an
or nor not -> or not

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/BackendRequest.java
```diff
@@ -63,6 +67,17 @@ public LuceneWork getLuceneWork() {
 		return action;
 	}
 
+	public String getIndexName() {
+		return indexName;
+	}
+
+	/**
+	 * Whether to perform an explicit index refresh after executing this action is needed or not.
```
gsmet: Hmmm, "whether performing" would be better?

gsmet:
Looks like you also need to set `hibernate.search.default.elasticsearch.refresh_after_write` to `true` for the Wildfly integration tests.

gunnarmorling:
Thanks, @gsmet! Addressed your remarks and pushed an update.

gsmet:
Looks good to me. Waiting for https://github.com/hibernate/hibernate-search/pull/1094 to be merged.

Sanne:
It's nice to have such an option but I'm not persuaded on making this the default.. like I said on JIRA, this would be inconsistent with the semantics of the Lucene backend.

I'd rather flip the default for both (in 6) if you all think people strongly prefer async by default.

Let's talk about this? I might be persuaded.. but "I think it's better" is not good enough to beat consistency, let's try to put some more rationale in it ;)

Sanne:
P.S. I actually had this proposal in the meeting agenda: didn't expect a PR coming already.

gunnarmorling:
Hi, there are different axes of consistency. Matching typical usage patterns of ES and meeting expectations of people familiar with ES being one of them (and IMHO the more important one than trying to do all things just as with the embedded backend).

That's what the ES guide [says on this](https://www.elastic.co/guide/en/elasticsearch/guide/master/near-real-time.html):

> While a refresh is much lighter than a commit, it still has a performance cost. A manual refresh can be useful when writing tests, but don’t do a manual refresh every time you index a document in production; it will hurt your performance. Instead, your application needs to be aware of the near real-time nature of Elasticsearch and make allowances for it.

So that's why I disabled the refreshs by default: It's not how you normally work with ES. I hope we can put it into the Beta, as I fear people would be caught by suprise by the unidiomatic usage pattern and the poor performance resulting from it.

gsmet:
I agree with @gunnarmorling on this one: it's what we expect when we use Elasticsearch. We have the exact same semantic with Solr.

Maybe we should add a paragraph about it in the doc: how refreshing works, how you can configure it in Elasticsearch.

Sanne:
Let's consider what one expects when using Hibernate:
Remember that ES is not the API we expose, the programming model is the one
from JPA, with JTA et al.

I'm very well aware of the performance implications: our own Lucene backend
had to face this and deploy lots of smart tricks to not make this tradeoff
too bad.

So that warning in the ES documentation need to be put in perspective:
bring awareness of this to their users, and of course also to make a good
Out of the box impression.

I'd also be interested in exploring flush_before_query patterns (both for
ES and Lucene). Like we do some clever stuff around postponing actual
delete operations in the NRT backend.

Just saying, there's lots of interesting things which we could do without
having to break the default JPA experience.

+100 to have the option for this.

Regarding making it a default I'd like to talk about that next week.

gunnarmorling:
Good out-of-the-box experience is what we should provide, too. Sticking to recommendations and best practices of how to use the products we integrate with is a vital part of this.

Taking a step back and ignoring the clear recommendation from the ES docs for a second, what's the practical use case for requiring sub second visibility changes towards and index meant for full text search? We are not talking about actual business queries/transactions on the primary application data here.

E.g. consider a backoffice worker adding a new item to the product catalogue, do you think there is any relevant difference between that item showing up in search results one second (or even a couple of seconds) sooner or later? There might be theoretical use cases where it matters, but very clearly that appears not to be a sensible default assumption to me.

It'd be sad to provide a default which may be "safer" in some rare edge cases but comes at the prices of poor performance and need for reconfiguration in the vast majority of cases.

Maybe it'd help to provide an API to trigger explicit refreshes if needed, e.g. when having set the refresh interval to a rather high value?

Regarding flush-on-query, it might be (yet another) interesting thing, but one thing to consider is that people will also query ES bypassing HSEARCH. E.g. from a web site straight through the REST API to benefit from result highlighting etc. It's different for Lucene Embedded.

emmanuelbernard:
Can we push this PR and change the default after the discussion next week?
Personally, at least from an ES PoV, not waiting for the second is a no-go in production. I'd prefer to respect the ES decision on this..

gsmet:
@gunnarmorling this one now needs a rebase. Thanks!

gunnarmorling:
Rebased and force-pushed an update.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/BackendRequest.java
```diff
@@ -69,6 +71,13 @@ public String getIndexName() {
 		return indexName;
 	}
 
+	/**
+	 * Whether performing an explicit index refresh after executing this action is needed or not.
+	 */
+	public boolean needsRefreshAfterWrite() {
```
Sanne: This method is never used.
Wondering if you had plans, or forgot to commit some tests?

Sanne:
thanks! merged.

I love having the option, but to be clear I'm not convinced about the default yet ;)

PR #1101, opened 2016-05-03
HSEARCH-2106 Using unique name for ES projection constants interface;…

gunnarmorling:
… Only exposing constants actually supported

I made the ES constants interface not extend `ProjectionConstants` anymore and gave it a unique name. With that most wrong usages are avoided as auto-completion won't give you wrong candidates. I don't think it needs a strongly type-safe approach.


gunnarmorling:
Force-pushed an update.

engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
```diff
@@ -233,6 +242,12 @@ public ExtendedSearchIntegrator getExtendedSearchIntegrator() {
 		return extendedIntegrator;
 	}
 
+	/**
+	 * Returns the names of the projection constants (beginning with {@link #HSEARCH_PROJECTION_FIELD_PREFIX}) supported
+	 * by a specific implementation.
+	 */
+	protected abstract Set<String> getSupportedProjectionConstants();
```
gsmet: > (beginning with {@link #HSEARCH_PROJECTION_FIELD_PREFIX})

This is not true for `_hibernate_class`.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/ElasticsearchProjectionConstants.java
```diff
@@ -0,0 +1,48 @@
...
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.elasticsearch;
+
+/**
+ * Projection constants specific to the Elasticsearch backend.
+ *
```
gsmet: Maybe we should mention in the comment that they also have to be declared in `SUPPORTED_PROJECTION_CONSTANTS`.

That's where an `enum` would have been handy. Any reason why we dont switch to an enum implementing an interface (true question, interested in the answer!)? I think we could do this while keeping the backwards compatibility. It wouldn't buy us any more type safety though.

Sanne: I'm not sure I understand the enum proposal. Could you sketch it?

Ideally log ideas on https://hibernate.atlassian.net/browse/HSEARCH-2265

engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
gsmet: And we also have an issue with spatial distance. It starts with only one underscore.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/ElasticsearchProjectionConstants.java
gsmet: I sketched it in the JIRA issue.

engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
gunnarmorling: Indeed; What a mess... :/

gunnarmorling: Btw. things were still functioning correctly: We'd not consider these two as projection constants, letting them pass through, leaving it to the specific impl to deal with them.

gunnarmorling:
Rebased and added one more commit to address @gsmet's remarks. Should be good to go now.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/ElasticsearchProjectionConstants.java
gsmet: @gunnarmorling any comment on this proposal?

gunnarmorling: @gsmet, I commented on the issue; I'm wondering how that'd play together with projecting actual fields? Right now I feel we don't gain much for the price of more verbose usage.

engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
```diff
@@ -233,6 +242,13 @@ public ExtendedSearchIntegrator getExtendedSearchIntegrator() {
 		return extendedIntegrator;
 	}
 
+	/**
+	 * Returns the names of the projection constants supported by a specific implementation in addition to projecting
+	 * actual field values. If a given projection name begins with {@link #HSEARCH_PROJECTION_FIELD_PREFIX} and is not
+	 * part of the set of constants returned by an implementation, an exception will be raised.
+	 */
+	protected abstract Set<String> getSupportedProjectionConstants();
```
gsmet: I think you miss the @return here.

gunnarmorling: I'm using that style - providing contents for the "main" section, but omitting `@return` if it doesn't add any value - on purpose actually. It reads better that way in rendered JavaDoc than the other way around.

To avoid redundancy, I rarely use both. If I do it, then I have a general description in the main part (as the case here), and a description of specific values in the `@return` part (e.g. "@return null if bla bla bla, foo bar baz otherwise.").

It's not something we ever formally agreed on, just the way I came to think is providing most benefit.

gsmet:
Rebased, fixed the conflicts and merged! Thanks.

Sanne:
@gsmet whem merging stuff, remember to update JIRA! thanks

gsmet:
@Sanne hmmm, AFAICS, I marked https://hibernate.atlassian.net/browse/HSEARCH-2106 as resolved in JIRA.

Sanne:
sorry, I'm confused by the open state of https://hibernate.atlassian.net/browse/HSEARCH-2227 which is linked to this PR..

PR #1103, opened 2016-05-19
HSEARCH-2274 Making sure ES host can be specified using property "hib…

gunnarmorling:
…ernate.search.default.elasticsearch.host"

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/JestClient.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/JestClient.java
@@ -47,7 +47,17 @@ public class JestClient implements Service, Startable, Stoppable {
 
 	private static final Log LOG = LoggerFactory.make( Log.class );
 
+	/**
+	 * HTTP response code for a request timed out.
+	 */
 	private static final int TIME_OUT = 408;
+
+	/**
+	 * Prefix for accessing the {@link ElasticsearchEnvironment#SERVER_URI} variable. That's currently needed as we
+	 * don't access this one in the context of a specific index manager. The prefix is used to have the property name
+	 * inline with the other index-related property names, it cannot be overridden per moment, though.
+	 */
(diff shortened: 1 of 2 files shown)


elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/JestClient.java
```diff
@@ -47,8 +47,18 @@
...
 
+	/**
+	 * HTTP response code for a request timed out.
+	 */
 	private static final int TIME_OUT = 408;
 
+	/**
+	 * Prefix for accessing the {@link ElasticsearchEnvironment#SERVER_URI} variable. That's currently needed as we
+	 * don't access this one in the context of a specific index manager. The prefix is used to have the property name
+	 * inline with the other index-related property names, it cannot be overridden per moment, though.
```
gsmet: "inline with" -> in line with
"per moment" -> per index? at the moment?

elasticsearch/src/test/java/org/hibernate/search/backend/elasticsearch/test/HostCanBeConfiguredIT.java
```diff
@@ -0,0 +1,62 @@
...
+
+import org.hibernate.boot.Metadata;
+import org.hibernate.boot.MetadataSources;
+import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
+import org.hibernate.search.exception.SearchException;
+import org.hibernate.search.testsupport.TestForIssue;
+import org.junit.Test;
+
+/**
+ * Test to make sure that the ES host can be configured and not the default is applied unconditionally. To test it, the
```
gsmet: Looks like the "not" is misplaced.

Maybe "and the default is not applied unconditionally" would be better?

gsmet:
@gunnarmorling looks good to me apart from the minor remarks on the comments. I let you fix them and merge? Thanks!

Sanne:
merged with @gsmet 's suggestions

gunnarmorling:
Thanks, everyone!

PR #1104, opened 2016-05-19
HSEARCH-2275 Make the async indexing processor for Elasticsearch resi…

Sanne:
…lient to errors during Runnable setup and teardown

https://hibernate.atlassian.net/browse/HSEARCH-2275


elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/BackendRequestProcessor.java
```diff
@@ -224,6 +239,11 @@ public void shutdown() {
 			scheduler.shutdown();
 			try {
 				scheduler.awaitTermination( Long.MAX_VALUE, TimeUnit.SECONDS );
+				if ( lastAsyncWorkLatch != null ) {
+					//It's possible that a task was successfully scheduled but had no chance to run,
```
gunnarmorling: In which case could that happen?

gunnarmorling:
Looks good. Do you know what situation led to the stuck thread from the issue?

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/BackendRequestProcessor.java
Sanne: - you add some work, which schedules a job to run in X time. 
- you (or another thread) initiate shut down,  but X didn't start yet
- the executor is shutdown so the scheduked job is canceled
- other threads waiting for it are stuck forever

Sanne: Actually it would be nice to not skip this block on interrupted. I should move it into a finally after the catch.

gunnarmorling: But won't already scheduled tasks be executed prior to shutdown? That's how I take the description of `ExecutorService#shutdown()` / `awaitTermination()`.

Sanne:
No, TBH I don't know for sure if this resolves it but it addresses one
possible cause for that. I could reproduce it before, but not after these
changes.

I should make a proper test but that would take me days :/ I'd rather merge
this first.

elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/BackendRequestProcessor.java
Sanne: I think you're correct, the inference of my 3rd point above is not.

What I was having in mind is that "if anything goes wrong" with that Runnable not being actually run, other threads are stuck forever. In practice I think this should only happen during an interruption so I'll move that code to handle the case.

Sanne:
pushed an update (and rebased)

Sanne:
superseded by #1107 as it includes it.

PR #1106, opened 2016-05-19
HSEARCH-2110 Update the Elasticsearch documentation prior to release

gsmet:
NOTE: this takes into account the fact that the artifact and the package has been renamed to remove the "backend" part.

### DIFF
--- a/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
+++ b/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
@@ -33,7 +33,7 @@ There could be various reasons to prefer this over an "embedded Lucene" approach
 There are a couple drawbacks compared to the embedded Lucene approach though:
 
 * incur a performance penalty of remote RPCs both for index updates and to run queries
-* need to buy & manage additional servers
+* need to buy and manage additional servers
 
 Which solution is best will depend on the specific needs of your system.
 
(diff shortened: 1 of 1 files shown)


documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -73,7 +73,7 @@ docker run -p 9200:9200 -d -v "$PWD/plugin_dir":/usr/share/elasticsearch/plugins
 ==== Dependencies in your Java application
 
 In addition to the usual dependencies like Hibernate ORM and Hibernate Search,
-you will need the new `hibernate-search-backend-elasticsearch` jar.
+you will need the new `hibernate-search-elasticsearch` jar.
```
gunnarmorling: Actually the artifact id hasn't been changed yet. Do we have an alternative term for the notition of "Lucene Embedded vs. ES vs. Solr"? Or should we just remove "backend" and be done. Works for me, too.

Though it seems a bit we are lacking a term for this concept here (in general, not only in terms of artifact ids). "Indexing service"?

gsmet: I agree we need a term for this. "Indexing service" is nice, I like it.

The change in the jar name was in prevision of the new name for the artifact.

We can commit that once Sanne has changed the name.

Sanne: +1 for the need of a name, I keep calling it the "Lucene embedded vs Elasticsearch choice".

But whatever the name, I think we'll not want to add too many pre and postfixes to the Maven artifact so `hibernate-search-elasticsearch` looks good.

Sanne:
merged, thanks!

PR #1107, opened 2016-05-20
Package, artifact and module renames

Sanne:


DavideD:
Merged https://github.com/hibernate/hibernate-search/pull/1107

Thanks

backends/jgroups/pom.xml
```diff
@@ -15,7 +15,7 @@
         <relativePath>../../pom.xml</relativePath>
     </parent>
 
-    <artifactId>hibernate-search-backend-jgroups</artifactId>
+    <artifactId>hibernate-search-jgroups</artifactId>
```
DavideD: @Sanne Was this a test? Because I failed :) Sending a commit to fix this

Sanne: ouch, thanks! yes that was not intentional.

PR #1108, opened 2016-05-20
HSEARCH-2061 Update javadoc that refers to Elasticsearch as a backend

DavideD:
A follow up commit to https://github.com/hibernate/hibernate-search/pull/1107

It removes the term backend when Elasticsearch is mentioned.

Feel free to ignore this if it does not make much sense


Sanne:
makes sense! thanks. There are in fact several more similar things.. I'll incorporate.

Sanne:
merged

PR #1111, opened 2016-05-25
HSEARCH-2285 Explicit dependency on the javax.annotations to allow co…

Sanne:
…mpiling on JDK9

 - https://hibernate.atlassian.net/browse/HSEARCH-2285

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -877,6 +877,12 @@
                             <version>${jbossLoggingProcessorVersion}</version>
                             <scope>compile</scope>
                         </dependency>
+                        <dependency>
+                            <groupId>javax.annotation</groupId>
+                            <artifactId>jsr250-api</artifactId>
+                            <version>1.0</version>
+                            <scope>compile</scope>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -877,6 +877,12 @@
                             <version>${jbossLoggingProcessorVersion}</version>
                             <scope>compile</scope>
                         </dependency>
+                        <dependency>
+                            <groupId>javax.annotation</groupId>
+                            <artifactId>jsr250-api</artifactId>
+                            <version>1.0</version>
```
gsmet: Feel free to push it but I was thinking that it would be better to have it in a variable with the other versions so that we don't forget to update if one day there is a new version. Might be purely hypothetical though.

Sanne:
thanks! merged.

Good point on the version, but I think it's more likely that we'll delete this dependency soon: it's just to workaround a bug in JDK9, which is hopefully resolved soon.

PR #1112, opened 2016-05-26
HSEARCH-2287 Clear the cached results when the sort order is changed

gsmet:
https://hibernate.atlassian.net/browse/HSEARCH-2287

Clear the cache results when sort order is changed. It's especially important when using Elasticsearch because we cache much more things.


DavideD:
Merged https://github.com/hibernate/hibernate-search/pull/1112

Thanks

PR #1114, opened 2016-05-31
HSEARCH-1032 - Initial impl of mass indexer changes to support user defined identifier criteria producers.

Naros:
https://hibernate.atlassian.net/browse/HSEARCH-1032

### DIFF
--- a/orm/src/test/java/org/hibernate/search/test/batchindexing/IdentifierFilterTest.java
+++ b/orm/src/test/java/org/hibernate/search/test/batchindexing/IdentifierFilterTest.java
@@ -1,0 +1,162 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.test.batchindexing;
+
+import java.util.Map;
+
+import org.hibernate.cfg.AvailableSettings;
+import org.hibernate.criterion.Criterion;
+import org.hibernate.criterion.DetachedCriteria;
+import org.hibernate.criterion.Restrictions;
+import org.hibernate.search.FullTextSession;
+import org.hibernate.search.MassIndexer;
+import org.hibernate.search.Search;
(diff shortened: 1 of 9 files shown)


orm/src/test/java/org/hibernate/search/test/batchindexing/IdentifierFilterTest.java
```diff
@@ -0,0 +1,162 @@
...
+		fullTextSession.close();
+
+		assertEquals( 1, getNumberOfDocumentsInIndex( Clock.class ) );
+	}
+
+	@Test
+	public void testMassIndexerFilterMultipleTypes() throws InterruptedException {
+		final FullTextSession fullTextSession = openAndPopulate();
+
+		getMassIndexer( fullTextSession, Clock.class, IssueEntity.class )
```
gunnarmorling: It looks good, only I'm wondering whether we can avoid the repetition of entit types (once passed to `getMassIndexer()`, once to `registerCriteriaProvider()`.

    getMassIndexer( fullTextSession )
        .type( Clock.class )
            .limitedBy(new IdentifierCriteriaProvider() {
                public Criterion getIdentifierCriteria(Class<?> indexedType) {
                    return Restrictions.idEq( 1 );
                }
            } )
        .type( IssueEntity.class )
            .limitedBy( ... )
        .startAndWait();

Naros: So keep the original approach for backward compatibility, but provide a type-builder fashion as you described above, allowing for all 3 variants, like below?

    // original way - backward compatibility.
    fullTextSession.createIndexer( Entity1.class, Entity2.class ).startAndWait();
    
    // new way and old mixed
    fullTextSession.createIndexer( Entity2.class )
      .type( Entity1.class ) // delayed registration possible
      .startAndWait();

    // new way with identifier criteria provider
    fullTextSession.createIndexer( Entity2.class )
      .type( Entity1.class ) // delayed registration with identifier limit binder
        .limitedBy( newIdentifierCriteriaProvider() { /* things */ } )
      .startAndWait();

yrodiere:
Unfortunately, native Criteria are now deprecated in ORM and (I think) will be removed. I linked to this PR from the ticket so that we take every we can from it when we work on this again, but for now I will close this PR.
Thanks for all the work, at the very least this will be great food for thoughts!

PR #1119, opened 2016-06-15
HSEARCH-2157 Remove options of SpatialFieldBridgeByHash which are both not exposed and misleading

gsmet:
https://hibernate.atlassian.net/browse/HSEARCH-2157

The hash filter needs both the hash fields and the numeric fields to
work as the hash is used to prefilter the results and the numeric fields are
then used to return the exact results.


Sanne:
funny one, thanks! :+1:

PR #1122, opened 2016-06-18
HSEARCH-2301 Use the most specific type to build the criteria

gsmet:
https://hibernate.atlassian.net/browse/HSEARCH-2301

When building the criteria to get the entities, we used to always use
the root entity type of the hierarchy even if we only want to get
entities from one subtype.

This is especially counter productive when we are in the case of a join
inheritance with many subtypes: we then build a query with a left join
for each subtype instead of aquery only targeting the subtype.

Btw I checked visually that the query generated was what we wanted but I didn't find a way to unit test it.

### DIFF
--- a/orm/src/test/java/org/hibernate/search/test/query/initandlookup/CriteriaObjectInitializerAndHierarchyInheritanceTest.java
+++ b/orm/src/test/java/org/hibernate/search/test/query/initandlookup/CriteriaObjectInitializerAndHierarchyInheritanceTest.java
@@ -1,0 +1,207 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.test.query.initandlookup;
+
+import static org.fest.assertions.Assertions.assertThat;
+
+import java.util.List;
+import java.util.Locale;
+
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Inheritance;
+import javax.persistence.InheritanceType;
+import javax.persistence.MappedSuperclass;
+import javax.persistence.Table;
+
+import org.apache.lucene.index.Term;
+import org.apache.lucene.search.BooleanClause.Occur;
+import org.apache.lucene.search.BooleanQuery;
+import org.apache.lucene.search.Sort;
+import org.apache.lucene.search.SortField;
+import org.apache.lucene.search.TermQuery;
+import org.hibernate.Session;
+import org.hibernate.Transaction;
+import org.hibernate.search.FullTextSession;
+import org.hibernate.search.Search;
+import org.hibernate.search.annotations.DocumentId;
+import org.hibernate.search.annotations.Field;
+import org.hibernate.search.annotations.Indexed;
+import org.hibernate.search.annotations.SortableField;
+import org.hibernate.search.test.SearchTestBase;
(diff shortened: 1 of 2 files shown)


orm/src/test/java/org/hibernate/search/test/query/initandlookup/CriteriaObjectInitializerAndHierarchyInheritanceTest.java
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	@Override
+	public void configure(Map<String, Object> settings) {
+		settings.put( "hibernate.show_sql", true );
+		settings.put( "hibernate.format_sql", true );
+	}
+
+	@Test
+	@TestForIssue(jiraKey = "HSEARCH-2301") // Please fill in the JIRA key of your issue
```
Sanne: I'll remove the comment ;)

orm/src/test/java/org/hibernate/search/test/query/initandlookup/CriteriaObjectInitializerAndHierarchyInheritanceTest.java
```diff
@@ -0,0 +1,183 @@
...
+
+	@Override
+	public Class<?>[] getAnnotatedClasses() {
+		return new Class<?>[]{ BaseEntity.class, A.class, AA.class, AAA.class, AAB.class, AB.class, ABA.class, AC.class, B.class, BA.class };
+	}
+
+	@Override
+	public void configure(Map<String, Object> settings) {
+		settings.put( "hibernate.show_sql", true );
+		settings.put( "hibernate.format_sql", true );
```
Sanne: Could you remove these settings? It looks like they might have been useful when developing but not needed for the test purposes.

Sanne:
I love the idea, but I'm going to need some help to understand the unit test.

One point of concern is that if I apply only the test - without your fix - then the test passes anyway, which means that while it might be of interest it's not actually verifying your specific improvement.

Also, did you consider that you might be able to further narrow down the set of target types, by inspecting which types the query actually found?
You have the type in each `EntityInfo`..

orm/src/test/java/org/hibernate/search/test/query/initandlookup/CriteriaObjectInitializerAndHierarchyInheritanceTest.java
gsmet: Yeah sure!

gsmet:
> One point of concern is that if I apply only the test - without your fix - then the test passes anyway, which means that while it might be of interest it's not actually verifying your specific improvement.

Yes, that's what I said in the comment of the PR. I didn't find a good way to test it except visually. The test also checks that I didn't break anything.

> Also, did you consider that you might be able to further narrow down the set of target types, by inspecting which types the query actually found?

Did you read the patch? :)

Sanne:
> Did you read the patch? :)

Not completely no :) It was late and I was confused about the test, decided I needed to sleep ;)
Will polish it soon, although ideas for a real regression test would be nice..  last resort, use Byteman.

gsmet:
OK, let me try to write a working test.

gsmet:
OK, it took me a while to master the Byteman foo but here it is with a real regression test.

Sanne:
It's failing the test. ?

gsmet:
Yes, looks like the order of the Class<?> is not always the same. It was sometimes working and sometimes not. I fixed the test and repushed.

gsmet:
OK, this time, it's all good!

Sanne:
Awesome! I've added two more tests as I wanted to explore how it would behave with more full-text filtering... works beatifully!

PR #1130, opened 2016-07-06
Deprecating @ProvidedId

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2310

And adding a test insipred from a SO question, which was triggering wrong behaviour when annotated with @ProvidedId.

N.B. the test no longer uses ProvidedId so it's more of an illustration of how it should be done correctly.

### DIFF
--- a/orm/src/test/java/org/hibernate/search/test/id/RegistrationId.java
+++ b/orm/src/test/java/org/hibernate/search/test/id/RegistrationId.java
@@ -1,0 +1,71 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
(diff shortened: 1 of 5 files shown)


orm/src/test/java/org/hibernate/search/test/id/RegistrationId.java
```diff
@@ -0,0 +1,68 @@
...
+		int result = 1;
+		result = prime * result + ( ( department == null ) ? 0 : department.hashCode() );
+		result = prime * result + studentId;
+		return result;
+	}
+
+	@Override
+	public boolean equals(Object obj) {
+		if ( this == obj )
+			return true;
```
gsmet: Missing braces here, checkstyle is failing.

Sanne:
thanks @gsmet , updated!

gsmet:
Merged! Thanks!

PR #1132, opened 2016-07-26
[30%] HSEARCH-1872 Create a proper API for sorting definitions

emmanuelbernard:
This PR shows the interface needed to implement the DSL.
One can already play with it and compile the test code, not run it.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/query/dsl/sort/MultiValuedMode.java
+++ b/engine/src/main/java/org/hibernate/search/query/dsl/sort/MultiValuedMode.java
@@ -1,0 +1,42 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.query.dsl.sort;
+
+/**
+ * Mode when comparing multiple values.
+ *
+ * @author Emmanuel Bernard emmanuel@hibernate.org
+ */
+public enum MultiValuedMode {
+	/**
+	 * Pick the lowest value.
+	 */
+	MIN,
+
+	/**
+	 * Pick the highest value.
+	 */
+	MAX,
+
+	/**
+	 * Use the sum of all values as sort value.
+	 * Only applicable for number based array fields.
+	 */
+	SUM,
+
+	/**
+	 * Use the average of all values as sort value.
+	 * Only applicable for number based array fields.
+	 */
+	AVG,
+
+	/**
+	 * Use the median of all values as sort value.
+	 * Only applicable for number based array fields.
+	 */
+	MEDIAN
+}
--- a/engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
+++ b/engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
@@ -1,0 +1,52 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.query.dsl.sort;
+
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.SortField;
+
+/**
+ * @author Emmanuel Bernard emmanuel@hibernate.org
+ */
+public interface SortAdditionalSortFieldContext {
+
+	// navigation to next sort definition
+
+	/**
+	 * Order elements by a specific field
+	 */
+	SortFieldContext andByField(String field);
+
+	/**
+	 * Order elements by their relevance score in decreasing order by default.
+	 */
+	//TODO Is byScore terminal or can it be followed by other sorts?
+	SortOrderTermination andByScore();
+
+	/**
+	 * Order elements by their internal index order.
+	 * This is the fastest sorting.
+	 */
+	SortOrderTermination andByIndexOrder();
+
+	/**
+	 * Sort by distance from an already built spatial query.
+	 * For distance sort without a query, use {@code byField("location").fromCoordinates(...)}
+	 * or {@code byField("location").fromLatitude(...)}
+	 */
+	SortDistanceContext andByDistanceFromSpatialQuery(Query query);
+
+	/**
+	 * Order element using the native backend API for Lucene.
+	 */
+	SortNativeContext andByNative(SortField sortField);
+
+	/**
+	 * Order element using the native backend API for Elasticsearch.
+	 */
+	SortNativeContext andByNative(String sortField);
+}
--- a/engine/src/main/java/org/hibernate/search/query/dsl/sort/SortContext.java
+++ b/engine/src/main/java/org/hibernate/search/query/dsl/sort/SortContext.java
@@ -1,0 +1,49 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.query.dsl.sort;
+
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.SortField;
+
+/**
+ * @author Emmanuel Bernard emmanuel@hibernate.org
+ */
+public interface SortContext {
+
+	/**
+	 * Order elements by their relevance score in decreasing order by default.
+	 */
+	SortOrderTermination byScore();
+
+	/**
+	 * Order elements by their internal index order.
+	 * This is the fastest sorting.
+	 */
+	SortOrderTermination byIndexOrder();
+
+	/**
+	 * Order elements by a specific field
+	 */
+	SortFieldContext byField(String field);
+
+	/**
+	 * Sort by distance from an already built spatial query.
+	 * For distance sort without a query, use {@code byField("location").fromCoordinates(...)}
+	 * or {@code byField("location").fromLatitude(...)}
+	 */
+	SortDistanceContext byDistanceFromSpatialQuery(Query query);
+
+	/**
+	 * Order element using the native backend API for Lucene.
+	 */
+	SortNativeContext byNative(SortField sortField);
+
+	/**
+	 * Order element using the native backend API for Elasticsearch.
+	 */
+	SortNativeContext byNative(String sortField);
+}
--- a/engine/src/main/java/org/hibernate/search/query/dsl/sort/SortDistanceContext.java
+++ b/engine/src/main/java/org/hibernate/search/query/dsl/sort/SortDistanceContext.java
@@ -1,0 +1,29 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.query.dsl.sort;
+
+import org.hibernate.search.query.dsl.Unit;
+
+/**
+ * @author Emmanuel Bernard emmanuel@hibernate.org
+ */
+public interface SortDistanceContext extends SortOrder<SortDistanceContext>, SortAdditionalSortFieldContext, SortTermination {
+
+	/**
+	 * Hint for the backend to return data in the appropriate unit.
+	 */
+	SortDistanceContext in(Unit unit);
+
+	/**
+	 * Hint for the backend to use a specific distance algorithm
+	 */
+	SortDistanceContext withComputeMethod(DistanceMethod distanceMethod);
+
+	// parametrization
+
+	SortMissingValueContext<SortDistanceContext> onMissingValue();
+}
--- a/engine/src/main/java/org/hibernate/search/query/dsl/sort/SortFieldContext.java
+++ b/engine/src/main/java/org/hibernate/search/query/dsl/sort/SortFieldContext.java
@@ -1,0 +1,57 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.query.dsl.sort;
+
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+import java.util.Comparator;
+
+import org.hibernate.search.spatial.Coordinates;
+
+/**
+ * @author Emmanuel Bernard emmanuel@hibernate.org
+ */
+public interface SortFieldContext extends SortAdditionalSortFieldContext, SortOrder<SortFieldContext>, SortTermination {
+
+	// Geospatial sorting
+
+	/**
+	 * Order elements by their distance from {@code coordinates}.
+	 */
+	SortDistanceContext fromCoordinates(Coordinates coordinates);
+
+	/**
+	 * Latitude in degree
+	 */
+	SortLatLongContext fromLatitude(double latitude);
+
+	// native comparator
+
+	/**
+	 * Compare field with specific comparator implementation.
+	 */
+	SortComparatorContext withComparator(Comparator<?> comparator);
+
+	// parametrization
+
+	/**
+	 * Describe how to treat missing values when doing the sorting.
+	 */
+	SortMissingValueContext<SortFieldContext> onMissingValue();
+
+
+	/**
+	 * Describe how multi-valued fields are processed
+	 * @return
+	 */
+	SortFieldContext treatMultiValuedAs(MultiValuedMode mode);
+
+}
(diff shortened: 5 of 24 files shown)


engine/src/main/java/org/hibernate/search/query/dsl/sort/SortContext.java
```diff
@@ -0,0 +1,49 @@
...
+	 * Order elements by a specific field
+	 */
+	SortFieldContext byField(String field);
+
+	/**
+	 * Sort by distance from an already built spatial query.
+	 * For distance sort without a query, use {@code byField("location").fromCoordinates(...)}
+	 * or {@code byField("location").fromLatitude(...)}
+	 */
+	SortDistanceContext byDistanceFromSpatialQuery(Query query);
```
Sanne: here I'm wondering if we could avoid references to the Lucene type 'query'.

Assuming in a next step we'll be able to control the Sort DSL directly on the Query Builder (the existing Query DSL), then such a thing would go away right?

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortNativeContext.java
```diff
@@ -0,0 +1,13 @@
...
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.query.dsl.sort;
+
+/**
+ * @author Emmanuel Bernard emmanuel@hibernate.org
+ */
+public interface SortNativeContext extends SortAdditionalSortFieldContext, SortTermination {
```
Sanne: I guess you could persuade me to introduce a `SortNativeContext`, but I wasn't expecting us to provide such a thing.

Shouldn't we encourage people needing "native access" to avoid the DSL?

Do you think we should actually deprecate the old "setSort" and have all users - including the ones with specific native needs - move to this DSL?

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortFieldContext.java
```diff
@@ -0,0 +1,51 @@
...
+	 * Latitude in degree
+	 */
+	SortLatLongContext fromLatitude(double latitude);
+
+	// native comparator
+
+	/**
+	 * Compare field with specific comparator implementation.
+	 */
+	SortComparatorContext withComparator(Comparator<?> comparator);
```
Sanne: This is one of those things which maybe we shouldn't expose? Hard to abstract across different search engines.

Happy to leave it if you have a solid use case.

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortDistanceContext.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * @author Emmanuel Bernard emmanuel@hibernate.org
+ */
+public interface SortDistanceContext extends SortOrder<SortDistanceContext>, SortAdditionalSortFieldContext, SortTermination {
+
+	/**
+	 * Hint for the backend to return data in the appropriate unit.
+	 */
+	SortDistanceContext in(Unit unit);
```
Sanne: I'm not sure what "to return data" is referring to. Shouldn't this be an option for projections?

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
```diff
@@ -0,0 +1,52 @@
...
+	 * Sort by distance from an already built spatial query.
+	 * For distance sort without a query, use {@code byField("location").fromCoordinates(...)}
+	 * or {@code byField("location").fromLatitude(...)}
+	 */
+	SortDistanceContext andByDistanceFromSpatialQuery(Query query);
+
+	/**
+	 * Order element using the native backend API for Lucene.
+	 */
+	SortNativeContext andByNative(SortField sortField);
```
Sanne: I'd remove the following two which return `SortNativeContext` for now. Happy to re-add such things later?

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
```diff
@@ -0,0 +1,52 @@
...
+	 * This is the fastest sorting.
+	 */
+	SortOrderTermination andByIndexOrder();
+
+	/**
+	 * Sort by distance from an already built spatial query.
+	 * For distance sort without a query, use {@code byField("location").fromCoordinates(...)}
+	 * or {@code byField("location").fromLatitude(...)}
+	 */
+	SortDistanceContext andByDistanceFromSpatialQuery(Query query);
```
Sanne: I'm confused by this one. What does it mean to sort the current Query A "by distance from an already built spatial query" ?

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
```diff
@@ -0,0 +1,52 @@
...
+	/**
+	 * Order elements by a specific field
+	 */
+	SortFieldContext andByField(String field);
+
+	/**
+	 * Order elements by their relevance score in decreasing order by default.
+	 */
+	//TODO Is byScore terminal or can it be followed by other sorts?
+	SortOrderTermination andByScore();
```
Sanne: I'd remove this option for now as mixing "by score" with an additional sorting option is quite exotic.

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
```diff
@@ -0,0 +1,52 @@
...
+
+	/**
+	 * Order elements by their internal index order.
+	 * This is the fastest sorting.
+	 */
+	SortOrderTermination andByIndexOrder();
```
Sanne: Similarly to _andByScore_, I'd remove  this option too as it makes even less sense to use in practice to mix another sorting with the "by index" sort.

I think "by index" is used exclusively for fastest speed when sorting doesn't matter at all (in which case one wouldn't combine it with another sort), or if the index was specifically constructed to be laid out in a specific sort order but we don't give such control.

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortDistanceContext.java
```diff
@@ -0,0 +1,29 @@
...
+
+	/**
+	 * Hint for the backend to use a specific distance algorithm
+	 */
+	SortDistanceContext withComputeMethod(DistanceMethod distanceMethod);
```
Sanne: We currently support two modes {{SpatialMode.RANGE}} and {{SpatialMode.HASH}}, but these control indexing format, so a query needs to match the chosen type.

For sorting, we only have one strategy which is based on `org.hibernate.search.spatial.impl.DistanceComparator.DistanceComparator(Point, int, String)`.. how does this fit in, are you planning to provide more sorting implementation alternatives?

engine/src/main/java/org/hibernate/search/query/dsl/sort/MultiValuedMode.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ */
+package org.hibernate.search.query.dsl.sort;
+
+/**
+ * Mode when comparing multiple values.
+ *
+ * @author Emmanuel Bernard emmanuel@hibernate.org
+ */
+public enum MultiValuedMode {
```
Sanne: Nice, hadn't thought about multi-values. I'm afraid you'll have to pick a default, so that it can work even if the user didn't  invoke the method `org.hibernate.search.query.dsl.sort.SortFieldContext.treatMultiValuedAs(MultiValuedMode)` ?

engine/src/main/java/org/hibernate/search/query/dsl/sort/MultiValuedMode.java
```diff
@@ -0,0 +1,43 @@
...
+	/**
+	 * Pick the lowest value.
```
Sanne: I understand what you mean here, but this should probably rephrase to clarify how it would work on Strings.

engine/src/main/java/org/hibernate/search/query/dsl/sort/MultiValuedMode.java
```diff
@@ -0,0 +1,43 @@
...
+	MIN,
+
+	/**
+	 * Pick the highest value.
+	 */
+	MAX,
+
+	/**
+	 * Use the sum of all values as sort value.
+	 * Only applicable for number based array fields.
```
Sanne: "Only applicable for number based array fields"

I guess the focus should be on *number* but not strictly only an *array* ? Could work also on numeric values which have been indexed as strings via some string bridge?

Comment applies on all three of SUM, AVG, MEDIAN.

BTW as commented on JIRA, in this case it seems like you're also adding a new feature (on top of a new API).. feel free to split this up into a follow up task? I think currently our docs are pretty clear that you can't sort on multi-valued fields.

orm/src/test/java/org/hibernate/search/test/query/dsl/SortDSLTest.java
```diff
@@ -0,0 +1,251 @@
...
+		cfg.put( Environment.MODEL_MAPPING, MappingFactory.class.getName() );
+	}
+
+	public static class MappingFactory {
+		@Factory
+		public SearchMapping build() {
+			SearchMapping mapping = new SearchMapping();
+			mapping
+					.analyzerDef( "stemmer", StandardTokenizerFactory.class )
+					.filter( StandardFilterFactory.class )
```
Sanne: you'd normally indent such things nicely.. did you give up to auto-formatters? ;-)

orm/src/test/java/org/hibernate/search/test/query/dsl/SortDSLTest.java
```diff
@@ -0,0 +1,251 @@
...
+				Animal.class,
+				Day.class,
+				CoffeeBrand.class,
+				Coffee.class
+		};
+	}
+
+	@Override
+	public void configure(Map<String,Object> cfg) {
+		cfg.put( Environment.MODEL_MAPPING, MappingFactory.class.getName() );
```
Sanne: unrelated, but FYI you can put the `SearchMapping` instance directly in the cfg, so to not have to fiddle with classname, @Factory annotations, etc..

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortContext.java
emmanuelbernard: When we generalise the DSL, this would naturally be using the right generic type. I did not see a point in working on that now.

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortNativeContext.java
emmanuelbernard: In the 5.x series setSort cannot be deprecated. The DSL output is what is passed to setSort. In 6 all bets are off but I am not sure how to do it. 

I personally like how one can mix normal and native declarations. But we can remove it if people don't agree with me.

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortFieldContext.java
emmanuelbernard: It might be harder than I hoped so we will likely drop it. I pjust wanted to show all the options before removing them.

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortDistanceContext.java
emmanuelbernard: @gsmet told me it is an ES thing. It did surprise me too but they do expose unit for some reason.

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
emmanuelbernard: See comment above.

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
emmanuelbernard: Today when you do a spatial query and want to sort by it, you need to duplicate both the field name and the coordinates to distance from. This API goal is to avoid the duplication concern. Looks at the JIRA, I did show a more complete exampl in context.

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
emmanuelbernard: I,ve seen it in ES examples, I'm not sure it is exotic.

emmanuelbernard: Score and field I could be convinced of the exotism. Field then score, less so :)

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortDistanceContext.java
emmanuelbernard: Right I was thinking of offering alternative implementations.

engine/src/main/java/org/hibernate/search/query/dsl/sort/MultiValuedMode.java
emmanuelbernard: Hum, elastic search does not document their default if they even have one. I imagine they pick the first unless specified ?

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortNativeContext.java
Sanne: I guess the question is if you expect the existing "setSort" to be removed in future, and be replaced by this DSL as exclusive way?

engine/src/main/java/org/hibernate/search/query/dsl/sort/MultiValuedMode.java
emmanuelbernard: I actually copy pasted the ES doc here. These are their limitations. Maybe we should not expose this in the first round.

orm/src/test/java/org/hibernate/search/test/query/dsl/SortDSLTest.java
emmanuelbernard: This is not my code, I copied it :)

orm/src/test/java/org/hibernate/search/test/query/dsl/SortDSLTest.java
emmanuelbernard: I copied the DSLTest code really

orm/src/test/java/org/hibernate/search/test/query/dsl/SortDSLTest.java
Sanne: I knew it! your code is usually better :)

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortNativeContext.java
emmanuelbernard: if we manage to eliminate `createFullTextQuery` when a query is built from the DSL, then we should do the same for `setSort`.

Sanne: Ok that sounds compelling, but currently we can't eliminate `createFullTextQuery` as our query DSL only covers building a handful of queries.

So your suggestion is that in a next step we'd need to provide a full alternative to `createFullTextQuery` able to produce any advanced & custom query... essentially a way to include "native query definitions" in our DSL in a similar and consistent way with the Sort DSL proposal?

That sounds nice and I'm not against it, but I'm surprised that you'd start such a direction from the sort DSL.

emmanuelbernard: The DSL is already native friendly. A boolean query for example can aggregate native Lucene queries. This is very much like the native integration of native `SortField`s in sort. Where you thinking of something else?

Sanne: Ok fair enough. Just trying to gauge were this is going, as I expected a much simpler Sort API.. but I guess you're right I hadn't though about this aspect.

Let's try to make concrete preparations for deprecating the traditional `createFullTextQuery` during the 6.0 works then.

emmanuelbernard: Just to be clear, I don't have a clue how the replacement of `createFullTextQuery` would look like :D

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortDistanceContext.java
gsmet: @Sanne when you do a sort by distance, ES includes the distance calculated in the results so that you don't have to calculate it yourself. This is why they expose a way to tweak the unit. We use it in our ES work to avoid using a script to calculate the distance.

See https://github.com/hibernate/hibernate-search/blob/master/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java#L541 .

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
gsmet: I don't think score and field are so exotic. You might use a scoring which leads to equal scores and want to get stable results (and add an order by name for instance).

engine/src/main/java/org/hibernate/search/query/dsl/sort/MultiValuedMode.java
gsmet: Yup. Looks like it's orthogonal to the current work. We will need to dig in ES to see how they do it and port it to our Lucene backend. Better let it aside for now.

gsmet:
Closing for now. Yoann is working on a new PR superseding this one.

PR #1134, opened 2016-08-03
Some minor improvements

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2316
 - https://hibernate.atlassian.net/browse/HSEARCH-2317
 - https://hibernate.atlassian.net/browse/HSEARCH-2318

### DIFF
--- a/engine/src/test/java/org/hibernate/search/testsupport/BytemanHelper.java
+++ b/engine/src/test/java/org/hibernate/search/testsupport/BytemanHelper.java
@@ -77,7 +77,12 @@ public class BytemanHelper extends Helper {
 	 * @return the first
 	 */
 	public static String consumeNextRecordedEvent() {
-		return concurrentStack.removeFirst();
+		try {
+			return concurrentStack.removeFirst();
+		}
+		catch (java.util.NoSuchElementException nse) {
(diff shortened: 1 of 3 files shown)


engine/src/test/java/org/hibernate/search/testsupport/BytemanHelper.java
```diff
@@ -77,7 +77,12 @@ public void pushEvent(String message) {
 	 * @return the first
 	 */
 	public static String consumeNextRecordedEvent() {
-		return concurrentStack.removeFirst();
+		try {
+			return concurrentStack.removeFirst();
+		}
+		catch (java.util.NoSuchElementException nse) {
+			throw new IllegalStateException( "Attempted to consume an event, while no events were recorded. Check if the Byteman ruless actually triggered?" );
```
gsmet: I think the new message is misleading. I would have made it:
"Attempted to consume an event, while no events are left. If it is the first call to this method, check if the Byteman rules are actually triggered?"

The fact is that you have the same error if you have consumed too many events.

Sanne: good point, thanks I'll change it. Ok to push beyond that?

gsmet:
Merged! Thanks!

PR #1138, opened 2016-08-16
HSEARCH-2325 AbstractHSQuery.validateSortField validates SortFields even when lacking information

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2325

### DIFF
--- a/engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
+++ b/engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
@@ -405,29 +405,45 @@ public abstract class AbstractHSQuery implements HSQuery, Serializable {
 			}
 	}
 
-	private void validateNumericSortField(SortField sortField, NumericEncodingType numericEncodingType) {
+	private void validateNumericSortField(SortField sortField, NumericEncodingType indexNumericEncodingType) {
+		final NumericEncodingType sortNumericEncodingType;
 		switch ( sortField.getType() ) {
 			case BYTES:
 			case INT:
-				validateNumericEncodingType( sortField, numericEncodingType, NumericEncodingType.INTEGER );
+				sortNumericEncodingType = NumericEncodingType.INTEGER;
 				break;
 			case LONG:
-				validateNumericEncodingType( sortField, numericEncodingType, NumericEncodingType.LONG );
+				sortNumericEncodingType = NumericEncodingType.LONG;
 				break;
 			case DOUBLE:
-				validateNumericEncodingType( sortField, numericEncodingType, NumericEncodingType.DOUBLE );
+				sortNumericEncodingType = NumericEncodingType.DOUBLE;
 				break;
 			case FLOAT:
-				validateNumericEncodingType( sortField, numericEncodingType, NumericEncodingType.FLOAT );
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
```diff
@@ -405,29 +405,45 @@ private void validateSortField(SortField sortField, DocumentFieldMetadata fieldM
...
 			default:
-				throw LOG.sortTypeDoesNotMatchFieldType( String.valueOf( sortField.getType() ), String.valueOf( numericEncodingType ), sortField.getField() );
+				throw LOG.sortTypeDoesNotMatchFieldType( String.valueOf( sortField.getType() ),
+						String.valueOf( indexNumericEncodingType ), sortField.getField() );
 		}
+		if ( NumericEncodingType.UNKNOWN.equals( indexNumericEncodingType ) ) {
+			/*
+			 * The actual encoding type is unknown, so we can't validate more.
+			 * This happens most notably when using custom numeric field bridges.
+			 * We can't ask the user to add more metadata, since there is no annotation for
```
gsmet: But it works if you are using a MetadataProvidingFieldBridge, doesn't it? Probably only if the field name added in the field bridge is different from the default one?

yrodiere: Interestingly, it does not. See https://github.com/yrodiere/hibernate-search/commit/c386e678cf36e0617f744a7bd272925ce11437b4 and https://travis-ci.org/yrodiere/hibernate-search/builds/153335179
I see what you mean, though: maybe I was simply doing it wrong, and it happens that there also is an entirely different bug when doing it right.
Regarding this pull request, I think that even if remove my fix (because it's inappropriate), maybe I should still offer a more explicit error message and consider it's a fix for HSEARCH-2325.
I'll work on this tomorrow (I'll try to see where the bug lies exactly)

gsmet: Yes, I was pretty sure it wouldn't work when adding a field with the same name. I'm curious though to see if it works if you add a numeric field with a different name from the default one and sort on it.

That being said, I think your patch is probably necessary for now and we should integrate it. Let just go to the bottom of it before doing that so that it's properly documented.

yrodiere: The example I gave above used a different name from the Hibernate property. I see now you meant a different name from the one in the `@Field` annotation.
Just tried exactly that, and it does not work either: https://github.com/yrodiere/hibernate-search/commit/3a15ad43c1d500f1fec3633199a435f985ae0563 and https://travis-ci.org/yrodiere/hibernate-search/builds/153493483
It seems to be a different problem, though. Will investigate.

yrodiere: So, as we discussed on HipChat, I was simply doing things wrong.
I changed the comments as requested: https://github.com/hibernate/hibernate-search/pull/1138/commits/2cf890974551462ce4f7ddb916dbdd8d9e17641d
@gsmet Okay to merge?

PR #1139, opened 2016-08-19
HSEARCH-1872 Create a proper API for sorting definitions

yrodiere:
JIRA: https://hibernate.atlassian.net/browse/HSEARCH-1872

~~Needs to be rebased on #1141 once it's been reviewed. This will require some refactoring, most notably in `SortDSLTest`.~~ (Done)

Supersedes #1132, where Emmanuel kicked off the API and started discussing it with everyone.

This implementation should cover most needs. It's not addressing many specific needs yet (apart from distance sorts), but is a good start.

Further reading (future improvements): https://hibernate.atlassian.net/browse/HSEARCH-2320


engine/pom.xml
```diff
@@ -218,8 +218,10 @@
                             org.hibernate.search.query;version="${project.version}",
                             org.hibernate.search.query.collector;version="${project.version}",
                             org.hibernate.search.query.dsl;version="${project.version}",
+                            org.hibernate.search.query.dsl.sort;version="${project.version}",
                             org.hibernate.search.query.engine;version="${project.version}",
                             org.hibernate.search.query.engine.spi;version="${project.version}",
+                            org.hibernate.search.query.engine.spi.impl;version="${project.version}",
```
gsmet: Hmmm, looks weird to have an impl package inside a spi package?

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -295,6 +295,30 @@ you need to enable the Groovy plugin by adding the following snippet to your Ela
...
+
+.Mixing DSL-defined sorts with native Elasticsearch JSON sorts
+====
+[source, JAVA]
+----
+QueryBuilder qb = fullTextSession.getSearchFactory()
+    .buildQueryBuilder().forEntity(Book.class).get();
+Query luceneQuery = /* ... */;
+FullTextQuery query = s.createFullTextQuery( luceneQuery, Book.class );
+SortField complexSortField = /* ... */;
```
gsmet: Looks like you're not using this variable in the rest of the example?

documentation/src/main/asciidoc/query.asciidoc
```diff
@@ -775,20 +775,44 @@ It is still possible to get the total number of matching elements regardless of
...
 
-===== Sorting
+===== [[query-sorting]]Sorting
 
 Apache Lucene provides a very flexible and powerful way to sort results. While the default sorting
 (by relevance) is appropriate most of the time, it can be interesting to sort by one or several
-other properties. In order to do so set the Lucene Sort object to apply a Lucene sorting strategy.
+other properties.
 
+In order to do so, you will have to build a Lucene `Sort` object, which can be done by using the Hibernate Search sort DSL or the native Lucene `SortField` class.
```
gsmet: I would add "either" just after using. It would be clearer I think.

documentation/src/main/asciidoc/query.asciidoc
```diff
@@ -775,20 +775,44 @@ It is still possible to get the total number of matching elements regardless of
...
+  .sort()
+    .byField("author").desc() // Descending order
+    .andByField("title") // Default order (ascending)
+  .createSort();
+query.setSort(sort);
+List results = query.list();
+----
+====
+
+Alternatively, you may build your sort using the `SortField` class directly. Then you will have to always specify the sort field type manually.
```
gsmet: I would add Lucene before SortField.

documentation/src/main/asciidoc/query.asciidoc
```diff
@@ -800,6 +824,104 @@ Be aware that fields used for sorting must not be tokenized (see <<field-annotat
...
+    .byField("author")
+      .onMissingValue().sortFirst()
+    .andByField("title")
+  .createSort();
+query.setSort(sort);
+List results = query.list();
+----
+====
+
+For numeric fields (and only numeric fields), you may even provide a default value to be used in case it's missing:
```
gsmet: Is it a limit of our implementation or a limit of Lucene?

documentation/src/main/asciidoc/query.asciidoc
```diff
@@ -800,6 +824,104 @@ Be aware that fields used for sorting must not be tokenized (see <<field-annotat
...
+  .sort()
+    .byField("author")
+    .andByNative(complexSortField)
+  .createSort();
+query.setSort(sort);
+List results = query.list();
+----
+====
+
+You may achieve similar mixes with Elasticsearch sorts when using the Elasticsearch backend. See <<elasticsearch-query-sorting>> for more information.
```
gsmet: I don't think it's worth a mention here.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/client/impl/ArbitrarySort.java
```diff
@@ -0,0 +1,42 @@
...
+ * @author Yoann Rodiere
+ */
+public class ArbitrarySort extends Sort {
+
+	private String fieldName;
+
+	private Object sortDescription;
+
+	public ArbitrarySort(String fieldName, String sortDescription) {
+		super( fieldName );
```
gsmet: Any strong reason why you split the native sort into 2 pieces? Intuitively, I would have put only one JSON string.

engine/src/main/java/org/hibernate/search/query/dsl/impl/ConnectedQueryBuilder.java
```diff
@@ -69,4 +71,9 @@ public SpatialContext spatial() {
 	public MoreLikeThisContext moreLikeThis() {
 		return new ConnectedMoreLikeThisContext( context );
 	}
+
+	@Override
+	public SortContext sort() {
+		return new ConnectedSortContext(context);
```
gsmet: Formatting. (yes, our checkstyles rules are not perfect)

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
```diff
@@ -0,0 +1,99 @@
...
+	 * @see SortContext#byIndexOrder()
+	 */
+	SortOrderTermination andByIndexOrder();
+
+	/**
+	 * Order elements by the value of a specific field.
+	 *
+	 * <p>The default order is <strong>ascending</strong>.
+	 * <p>The sort field type will be determined automatically if possible, and an exception will be thrown
+	 * if guessing is not possible. Guessing is impossible in particular if a custom field bridge is
```
gsmet: Not sure I like guessing here. Would put an "it" for the first and a "Automatically determining the type" for the second.

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
```diff
@@ -0,0 +1,99 @@
...
+	 * defined on the given field.
+	 *
+	 * @param fieldName The name of the index field to sort by
+	 * @throws SearchException If the sort field type could not be guessed.
```
gsmet: s/guessed/automatically determined/

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
```diff
@@ -0,0 +1,99 @@
...
+	 * <p>The default order is <strong>ascending</strong>.
+	 * <p><strong>Note:</strong> using this method is only required when sorting on a
+	 * field on which a custom field bridge is defined. Otherwise, one may simply use
+	 * {@link #andByField(String)}.
+	 *
+	 * @param fieldName The name of the index field to sort by
+	 * @param sortFieldType The sort field type
+	 * @see #andByField(String, org.apache.lucene.search.SortField.Type)
+	 */
+	SortFieldContext andByField(String fieldName, SortField.Type sortFieldType);
```
gsmet: I'm wondering if it's such a good idea to add this method. I don't have a better solution but I don't particularly like it.

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
```diff
@@ -0,0 +1,99 @@
...
+	 * and reference coordinates.
+	 *
+	 * @see SortContext#byDistance()
+	 */
+	SortDistanceNoFieldContext andByDistance();
+
+	/**
+	 * Order element using the native backend API for Lucene.
+	 *
+	 * <p>The sort order (ascending/descending) is defined in <code>field</code>
```
gsmet: s/field/sortField/ ?

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
```diff
@@ -0,0 +1,99 @@
...
+	/**
+	 * Order element using the native backend API for Elasticsearch.
+	 *
+	 * @param fieldName The name of the index field to sort by
+	 * @param sortFieldNativeDescription The sort field description, as valid JSON. This may be a
+	 * simple quoted string (e.g. "'asc'") or a JSON map (e.g. "{'key': 'value'}"). See
+	 * Elasticsearch's documentation for information about the exact syntax.
+	 * @see SortContext#byNative(String, String)
+	 */
+	SortNativeContext andByNative(String fieldName, String sortFieldNativeDescription);
```
gsmet: See my question above about having 2 parameters.

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortContext.java
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.query.dsl.sort;
+
+import org.apache.lucene.search.SortField;
+import org.hibernate.search.exception.SearchException;
+
+/**
```
gsmet: Javadoc?

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortContext.java
```diff
@@ -0,0 +1,88 @@
...
+	 * <p>The distance is computed between the value of a field carrying coordinates
+	 * (to be provided in the {@link SortDistanceNoFieldContext next context})
+	 * and reference coordinates.
+	 */
+	SortDistanceNoFieldContext byDistance();
+
+	/**
+	 * Order element using the native backend API for Lucene.
+	 *
+	 * <p>The sort order (ascending/descending) is defined in <code>field</code>
```
gsmet: s/field/sortField/ ? (And more generally, same remarks than for the previous file)

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortDistanceFieldAndReferenceContext.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.query.dsl.sort;
+
+/**
+ * @author Emmanuel Bernard emmanuel@hibernate.org
```
gsmet: Javadoc? (I don't remember if we have Javadoc for each DSL interface in the other DSLs so feel free to ignore if we're inline with the other DSLs)

engine/src/main/java/org/hibernate/search/query/dsl/sort/impl/ConnectedSortAdditionalSortFieldContext.java
```diff
@@ -0,0 +1,77 @@
...
+	@Override
+	public SortDistanceNoFieldContext andByDistance() {
+		states.closeSortField();
+		return new ConnectedSortDistanceNoFieldContext( queryContext, states );
+	}
+
+	@Override
+	public SortNativeContext andByNative(SortField sortField) {
+		states.closeSortField();
+		states.setCurrentSortFieldNativeSortDescription(sortField);
```
gsmet: formatting

engine/src/main/java/org/hibernate/search/query/dsl/sort/impl/NativeSortField.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.apache.lucene.search.FieldComparator;
+import org.apache.lucene.search.FieldComparatorSource;
+import org.apache.lucene.search.SortField;
+import org.hibernate.search.exception.SearchException;
+
+
+/**
+ * A SortField extension that simply serves as a data holder, to be detected
+ * and interpreted by non-Lucene backend implementations (elasticsearch in particular).
```
gsmet: I would capitalize Elasticsearch. Same remarks below.

engine/src/main/java/org/hibernate/search/query/dsl/sort/impl/SortFieldStates.java
```diff
@@ -0,0 +1,268 @@
...
+		processMissingValue( sortField );
+		sortFields.add( sortField );
+		reset();
+	}
+
+	public void guessCurrentSortFieldType() {
+		this.currentType = getCurrentSortFieldTypeGuess();
+	}
+
+	private SortField.Type getCurrentSortFieldTypeGuess() {
```
gsmet: determineCurrentSortFieldType?

engine/src/main/java/org/hibernate/search/query/dsl/sort/impl/SortFieldStates.java
```diff
@@ -0,0 +1,268 @@
...
+						return SortField.Type.INT;
+					case UNKNOWN:
+						break; // Fail below
+				}
+				break;
+			case BASIC:
+				return SortField.Type.STRING;
+		}
+
+		throw new SearchException( "Cannot guess the field type for field '" + currentName
```
gsmet: s/guess/determine/

engine/src/main/java/org/hibernate/search/query/dsl/sort/impl/SortFieldStates.java
```diff
@@ -0,0 +1,268 @@
...
+		if ( currentMissingValue != null && sortField.getType() != null ) {
+			if ( sortField.getType() == SortField.Type.STRING || sortField.getType() == SortField.Type.STRING_VAL ) {
+				if ( currentMissingValue == MISSING_VALUE_LAST ) {
+					sortField.setMissingValue( SortField.STRING_LAST );
+				}
+				else if ( currentMissingValue == MISSING_VALUE_FIRST ) {
+					sortField.setMissingValue( SortField.STRING_FIRST );
+				}
+				else {
+					throw new SearchException( "Unsupported 'use(Object)' for the field type: '" + currentType + "'."
```
gsmet: Hmmm, I don't see this limitation in the Elasticsearch doc (https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-sort.html#_missing_values). Is it a Lucene one? Or it does not work with Elasticsearch too and it's not documented. If we only support numeric values, I'm wondering if we should have use(Number) instead of use(Object).

engine/src/main/java/org/hibernate/search/query/engine/spi/impl/LuceneQueryDescriptor.java
```diff
@@ -4,7 +4,7 @@
  * License: GNU Lesser General Public License (LGPL), version 2.1 or later
  * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
  */
-package org.hibernate.search.query.hibernate.impl;
+package org.hibernate.search.query.engine.spi.impl;
```
gsmet: Yeah, either choose spi or impl :)

engine/src/test/java/org/hibernate/search/test/dsl/SortDSLTest.java
```diff
@@ -0,0 +1,809 @@
...
+ */
+@TestForIssue(jiraKey = "HSEARCH-1872")
+public class SortDSLTest {
+	@Rule
+	public SearchFactoryHolder sfHolder = new SearchFactoryHolder( IndexedEntry.class );
+
+	@Before
+	public void prepareTestData() {
+		storeData(
+				new IndexedEntry(0)
```
gsmet: formatting (and same below).

engine/src/test/java/org/hibernate/search/test/dsl/SortDSLTest.java
```diff
@@ -0,0 +1,809 @@
...
+						.desc()
+				.createSort();
+		assertThat(
+				query( query, sort ),
+				returnsIDsInOrder( 3, 0, 1, 2 )
+		);
+	}
+
+	@Test
+	public void singleField_numericFieldBridge_missingValue_use() throws Exception {
```
gsmet: Not a big fan of method names with underscores. Is it something we already have elsewhere?

gsmet:
Nice :). See my comments inline, it's mostly nitpicking.

engine/pom.xml
yrodiere: It's weird indeed. In this case, it comes from the fact that the Lucene backend implementation is bundled with the Hibernate Search engine, so I can't do much to prevent this, apart from renaming the package (which would just hide the weirdness in my opinion, not solve the actual problem).
On 6.0, if we actually have an API free of Lucene dependencies, we may be able to split things so that the Lucene backend has its own separate module, and this weirdness would disappear. But that's way out of the scope of this PR, I'm afraid.

On a side note, there already is other packages with SPI implementations provided by the "engine" module. See for instance `org.hibernate.search.cfg.spi.impl` or `org.hibernate.search.spi.impl`.

documentation/src/main/asciidoc/query.asciidoc
yrodiere: Both, actually. It's a limitation of `org.apache.lucene.search.SortField.setMissingValue(Object)`, which prevents us from setting the missing value for `String` fields. Since we're using `Sort`/`SortField` for carrying information, including with the Elasticsearch backend, the Lucene limitation becomes ours.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/client/impl/ArbitrarySort.java
yrodiere: It's just easier to implement given that `org.hibernate.search.query.dsl.sort.impl.NativeSortField` is already split into 2 pieces. Upon constructing an `ArbitrarySort` I could have merged the field name and the JSON sort definition into a single JSON sort, but it seemed pointless. Especially since it's an implementation package and we're not offering `ArbitrarySort` to users as an API.

On the subject of `NativeSortField`, there was two reasons for splitting it, but not strong ones:

 1. Implementation reason: if we actually know the field name, it avoids providing the super `SortField` constructor with a dummy field name, which would make this class even more hackish than it is now.
 2. API reason: Elasticsearch seems to only use sorts in the form `'fieldName': <sortspec>`, so thought it would be easier for users to separate the target field from the sort spec.

Anyway, as I said these are not strong requirements. I can change it if you want.

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
yrodiere: The actual issue that makes this method useful is the lack of data in the metamodel is some cases. If we don't make the requirement of providing that metadata with annotations/metadataprovidingfieldbridges obvious to users (which we'll do in 6.0), we have to at least document in this API the possible lack of metadata, and how to work it around.

Now, having this method is not strictly required, we could simply tell users to use `byNative(SortField)`/`byNative(String)` if they want to sort on that kind of fields (both in `byField`/`andByField` javadoc and in exceptions). I don't think it would be much cleaner, though.

documentation/src/main/asciidoc/query.asciidoc
gsmet: OK, might be a good idea to document that in the API. I don't remember having read something about that.

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortContext.java
yrodiere: We mostly don't have javadoc on DSL interfaces, except on methods.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/client/impl/ArbitrarySort.java
gsmet: Yes, I was talking about the whole thing, not just this one.

I'm just not sure it's such a good idea to separate the field name. It feels a bit hackish to consider '_geo_location' or '_script' as a field name but I certainly could be convinced too. Anyway, we probably should ask for other opinions. @Sanne , @emmanuelbernard , thoughts about that?

engine/src/main/java/org/hibernate/search/query/dsl/sort/impl/SortFieldStates.java
gsmet: So, following your explanation, we should keep use(Object) as we will probably remove this limitation in the future. Please document the limitation in the API though.

engine/src/main/java/org/hibernate/search/query/engine/spi/impl/LuceneQueryDescriptor.java
yrodiere: Nope, see https://github.com/hibernate/hibernate-search/pull/1139#discussion_r75628156

engine/pom.xml
gsmet: My question was more: why not `org.hibernate.search.query.engine.impl`? @Sanne could you guide us on this? When does it make sense to add an impl package inside the spi package? Looks like it's not something we do on OGM for instance.

engine/src/test/java/org/hibernate/search/test/dsl/SortDSLTest.java
yrodiere: It's just for tests. If we want to express the purpose of tests in their name, we have to use exceptionally long names, and underscores help making this readable, or at least less unreadable.

You can find other places where this syntax is used:

 * In QueryDSL (which is arguably a very, very clean piece of software, and happens to only use underscores when naming test methods) : https://github.com/querydsl/querydsl/blob/master/querydsl-jpa/src/test/java/com/querydsl/jpa/CollectionTest.java
 * In Hibernate Validator: https://github.com/hibernate/hibernate-validator/blob/master/engine-jdk8-tests/src/test/java/org/hibernate/validator/test/internal/engine/typeannotationconstraint/TypeAnnotationConstraintTest.java , https://github.com/hibernate/hibernate-validator/blob/master/engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/URLValidatorTest.java
 * In this project, though it seems a bit exotic: https://github.com/hibernate/hibernate-search/blob/master/orm/src/test/java/org/hibernate/search/test/spatial/SpatialSearchSortByDistanceAndPagingTest.java

Then again, on that kind of code (tests), I'd value clarity over consistency. But let me know if I must rename these method anyway.

engine/pom.xml
yrodiere: @gsmet The reason I answered you this way is that the code in this `spi.impl` package is actually an implementation of an interface in the `spi` package. It's not just random internal code.
But yes, given it's not intended for external use anyway, we could move it to `engine.impl`.

engine/src/main/java/org/hibernate/search/query/dsl/sort/impl/SortFieldStates.java
yrodiere: It's already documented in the API:

```
	/**
	 * When documents are missing a value on the sort field, use the given value instead.
	 *
	 * <p>This is only available for numeric fields.
	 * <p>Field bridges, if any, will be ignored. Thus the actual numeric value must be provided.
	 *
	 * @throws SearchException If the field is not numeric.
	 */
	T use(Object value);
```

engine/src/test/java/org/hibernate/search/test/dsl/SortDSLTest.java
gsmet: @Sanne ?

engine/src/main/java/org/hibernate/search/query/dsl/sort/impl/SortFieldStates.java
gsmet: You documented the limitation, not the why. We need to have the why explained somewhere so that we can remove this limitation when we go away from the Lucene API.

I would go for something along the lines of:
Lucene sort API limits this feature to numeric fields. As our implementation is currently based on the Lucene API underneath, this is only available for numeric fields for all the indexing services, Elasticsearch included.

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
gsmet: Yes, I understand this case.

My question is more to @emmanuelbernard and @Sanne : should we introduce this method while waiting for a real fix in 6 or encourage people to use `MetadataProvidingFieldBridge` and let it aside as a known limitation? Note that even using `MetadataProvidingFieldBridge` does not solve all the cases: if you add a field in your `MetadataProvidingFieldBridge` with the same name as the name used in the `@Field` annotation, you still can end up in this case.

Note that we also need to fix the underlying issue for Elasticsearch.

engine/pom.xml
gsmet: It's usually how it's done in OGM (see `AssociationContext` and `AssociationContextImpl` for instance). Waiting for @Sanne 's opinion as it looks like we already have a couple of `spi.impl` in Search so the rule might not be the same.

gunnarmorling: We have just `spi` or `impl` usually, so it should rather be `org.hibernate.search.query.engine.impl`.

engine/src/test/java/org/hibernate/search/test/dsl/SortDSLTest.java
gunnarmorling: We do it sometimes in tests, I think @hferentschik was the big lover of it :) Both styles are fine with me.

engine/src/main/java/org/hibernate/search/query/dsl/sort/impl/SortFieldStates.java
yrodiere: Done. I also created [HSEARCH-2332](https://hibernate.atlassian.net/browse/HSEARCH-2332).

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
```diff
@@ -0,0 +1,99 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.query.dsl.sort;
+
+import org.apache.lucene.search.SortField;
```
gunnarmorling: I didn't follow the entire discussion, so it may have been discussed before: is it right that we don't fully abstract from Lucene here? I'd have kind of hoped we'd find a generic approach working for both Lucene, ES (and Solr in the future). I.e. should we continue to expose Lucene specifics or rather stop doing it in new APIs? I'd strongly lean towards the latter.

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
yrodiere: A complement to @gsmet 's message above: the limitation of `MetadataProvidingFieldBridge` has been filed as a JIRA ticket here: https://hibernate.atlassian.net/browse/HSEARCH-2330

engine/pom.xml
gunnarmorling: Btw. another question is why is `impl` exported here to begin with? This should be avoided for any new code (eventually we should get rid of these impl leakages, so let's not add new ones).

yrodiere: Well, in this case, it's an implementation of an SPI, so leakage is necessary. Actually that's the whole point of an SPI: providing a way for others to expose another implementation.

In this particular case, it's exported because it's used there: https://github.com/yrodiere/hibernate-search/blob/HSEARCH-1872/orm/src/main/java/org/hibernate/search/impl/FullTextSessionImpl.java#L87

Originally, the class in this `spi.impl` package was part of `ORM`. But:

 1. It has nothing to do with ORM, but rather with the Lucene backend
 2. More importantly, QueryDescriptors are the easiest way to create queries in a backend-independant way. I needed this for implementing a backend-independant test in -engine. See https://github.com/yrodiere/hibernate-search/blob/HSEARCH-1872/engine/src/test/java/org/hibernate/search/test/dsl/SortDSLTest.java#L643 in particular.

Now, while the leakage of an SPI implementation is no big deal in my opinion, a better solution certainly exists. I'd be in favor of moving the code in https://github.com/yrodiere/hibernate-search/blob/HSEARCH-1872/orm/src/main/java/org/hibernate/search/impl/FullTextSessionImpl.java#L87 up to the -engine module, since it actually implements an -engine feature: the ability to use a different backend for creating a `HSQuery`. I'm a bit lost as where to expose this feature, though... I can try to find some place if you think it's the way to go.

gunnarmorling: We use the "impl" package for anything which is *not* meant to be accessed
by outside code, be it users or integrators of the library. We reserve the
right to change it at will without previous notice. Hence "impl" should
never be exported. We use the "spi" package in order to make specific
behaviours pluggable (mainly for integrators, sometimes also for users).

So usually an SPI is not implemented by ourselves (apart from the default
implementation) but an implementation is provided by users/integrators. In
the case where we provide some base class or utilities helping SPI
implementors to do their job, this one should be part of "spi" as well
(because if the intended usage of these classes is as described, they are
meant for public usage, thus cannot easily be changed as we like and thus
should not be located in "impl").

Of course code in "impl" can refer to code in "spi" or the user API as
needed. Only public code should not expose impl code (by means of usages in
method signatures or inheritance relationships).

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
yrodiere: @gunnarmorling I asked this on the JIRA ticket, and apparently this DSL is a temporary solution and will be reworked in 6.0, which is when we'll actually abstract from Lucene.

I think the point was that the generic solution is not hard to find, it's just hard to implement without adding APIs/SPIs in other places (`FullTextQuery` in particular) or even altering some APIs (`QueryBuilder` in particular is not well-suited to provide implementation-dependent sorts).

engine/pom.xml
yrodiere: I understand (and agree with) what you're saying, and yes I might have got a bit carried away with exposing an `impl` package (even for an SPI implementation), but I do not quite follow what you're suggesting.

Anyway, I'll make a proposal that should rid us of exposing the LuceneQueryDescriptor altogether.

yrodiere: @gunnarmorling @gsmet I created a [JIRA ticket](https://hibernate.atlassian.net/browse/HSEARCH-2333) and [another PR](https://github.com/hibernate/hibernate-search/pull/1141) for this particular issue. Once it's merged, I'll rebase this PR on master, and there should no longer be a need for exporting an `impl` package.

yrodiere:
I rebased on master and addressed one of the most problematic issues (https://github.com/hibernate/hibernate-search/pull/1139#discussion-diff-75551432).

Here are the remaining discussions:

 * https://github.com/hibernate/hibernate-search/pull/1139#discussion-diff-75553498
 * https://github.com/hibernate/hibernate-search/pull/1139#discussion_r75554237
 * https://github.com/hibernate/hibernate-search/pull/1139#discussion-diff-75556628
 * https://github.com/hibernate/hibernate-search/pull/1139#discussion-diff-75641457

@Sanne, would you please shed some light on those?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/client/impl/ArbitrarySort.java
emmanuelbernard: No opinion, as I'm missing context and don't quite understand what you guys are describing.

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
emmanuelbernard: If I remember my own past self, I did prefer not to have that method and rely on native to keep the API more portable.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/client/impl/ArbitrarySort.java
yrodiere: @emmanuelbernard Guillaume is concerned about "byNative(String fieldName, String sortSpec)" requiring two arguments, because in some case the `fieldName` argument isn't really a field name, but more an Elasticsearch keyword (such is the case for `_geo_location` and `_script`).
In short, I say it's easier to use that way and seems consistent with the way to do in ElasticSearch, but Guillaume thinks it's hackish.

On second though, depending on how we want to make the DSL generic later, having two arguments instead of one may be problematic. I don't know what was discussed in Rome exactly about the DSL generification, so...

yrodiere:
@Sanne Bump! I think we need your ax for the remaining discussions :)

engine/src/test/java/org/hibernate/search/test/dsl/SortDSLTest.java
Sanne: I don't see myself committing such a crime unless forced to, but I don't care :+1:

Sanne:
@yrodiere could you rebase it please?

yrodiere:
@Sanne Sure; done.

yrodiere:
Jenkins, retest this please.

Sanne:
hi @yrodiere could you rebase it once again please? thanks!

yrodiere:
Sure, done. Just waiting for the CI to check it all, but there shouldn't be
any problem.

yrodiere:
Argh, just been had by the new checkstyle rules. Working on that.

Sanne:
> Argh, just been had by the new checkstyle rules. Working on that.

that's why I asked ;) sorry!

yrodiere:
@Sanne Fixed!

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/client/impl/ArbitrarySort.java
Sanne: I like the simplicity of approach - thinking of end users - of having both parameters.

The fact that the first one might not be a real "field" makes me think we might want to rename the parameter, more than remove it?

This deserves further discussion but I think it's time to merge this, we can amend it shortly as needed.

engine/src/main/java/org/hibernate/search/query/dsl/sort/SortAdditionalSortFieldContext.java
Sanne: I agree on avoiding this method: don't like to expose Lucene specific types as `SortField.Type` when an alternative option is available. @yrodiere are you ok if I remove this method from the public API?

Sanne:
Merged.
Two pending doubts possibly to polish:
 - should we remove andByField(String,SortField.Type) method from public Sort API [I initially thought we shouldn't expose the Lucene type, but then realized the alternative that we'd be imposing on the user is quite annoying]
 - https://github.com/hibernate/hibernate-search/pull/1139#discussion-diff-75553498

PR #1140, opened 2016-08-23
HSEARCH-2256 Don't add a tenant ID field in ES if unnecessary

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2256

The path I took may seem needlessly complicated, but it has to be this way because the Elasticsearch module does not depend on ORM,only on Engine. Thus it cannot reference the multitenancy strategy directly.

This solution may not prevent unneeded tenant ID fields in every case. In particular, when users choose to implement `SearchConfiguration` themselves, they will not benefit from multitenancy autodetection. But it's arguably their choice, as they do choose to drop our default implementation.

There is an SPI break in this solution: I had to add a method to `SearchConfiguration`. It seems the most common way of implementing `SearchConfiguration` is to extend `SearchConfigurationBase`, though, and I added a default implementation to the latter. I checked the Infinispan HS module, and they do extend `SearchConfigurationBase`, so no trouble here. As for other SPI implementors, unfortunately I cannot tell...

### DIFF
--- a/engine/src/main/java/org/hibernate/search/cfg/spi/SearchConfigurationBase.java
+++ b/engine/src/main/java/org/hibernate/search/cfg/spi/SearchConfigurationBase.java
@@ -55,4 +55,9 @@ public abstract class SearchConfigurationBase implements SearchConfiguration {
 	public boolean isIdProvidedImplicit() {
 		return false;
 	}
+
+	@Override
+	public boolean isMultitenancyEnabled() {
+		return true;
+	}
 }
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
@@ -80,6 +80,7 @@ public class ElasticsearchIndexManager implements IndexManager, RemoteAnalyzerPr
 	private boolean refreshAfterWrite;
 	private IndexSchemaManagementStrategy indexManagementStrategy;
 	private String indexManagementWaitTimeout;
+	private boolean multitenancyEnabled;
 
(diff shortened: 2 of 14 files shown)


gsmet:
@gunnarmorling @Sanne what's your take on this? I'm not sure it's worth adding this just to avoid creating a field in the Elasticsearch schema. Thoughts?

engine/src/main/java/org/hibernate/search/cfg/spi/SearchConfigurationBase.java
```diff
@@ -55,4 +55,9 @@ public InstanceInitializer getInstanceInitializer() {
 	public boolean isIdProvidedImplicit() {
 		return false;
 	}
+
+	@Override
+	public boolean isMultitenancyEnabled() {
+		return true;
```
Sanne: Why true in this case? It should probably be false, but then I'm slightly concerned that this PR actually breacks things when the tenant_id is not part of the schema, as because of this line we're often testing only the "true" case

Sanne:
Looks good but I might want to use `false` by default in our tests. Followed up by pr #1154 to see if CI thinks that's a reasonable change.. then I'll merge.

engine/src/main/java/org/hibernate/search/cfg/spi/SearchConfigurationBase.java
yrodiere: Actually, the only reason was backward compatibility. If someone was extending `SearchConfigurationBase` with its own implementation, I didn't want to change the behavior without notice. I think Infinispan does extend `SearchConfigurationBase`, in particular.

On the other hand, I see no problem regarding `WorkerBuildContextForTest`.

As you changed the default value to false, maybe we should add something to the release notes?

PR #1141, opened 2016-08-23
HSEARCH-2333 Introduced createQueryDescriptor(Query, Class<?>...)

yrodiere:
... in ExtendedSearchIntegrator.
This enables easier backend-independent querying when relying on the ORM module is not an option, e.g. in tests of the -engine module.
In particular, in #1139 we have this code for querying without relying on the ORM module:

```java
private List<EntityInfo> query(Query luceneQuery, Sort sort) {
	HSQuery hsQuery = getHSQuery( luceneQuery );
	return hsQuery
			.targetedEntities( Arrays.asList( new Class<?>[] { IndexedEntry.class } ) )
			.projection( "id" )
			.sort( sort )
			.queryEntityInfos();
}

private HSQuery getHSQuery(Query luceneQuery) {
	ExtendedSearchIntegrator sf = sfHolder.getSearchFactory();
	QueryDescriptor descriptor = null;
	try {
		ServiceReference<LuceneQueryTranslator> translator =
				sf.getServiceManager().requestReference( LuceneQueryTranslator.class );
		if ( translator.get().conversionRequired( IndexedEntry.class ) ) {
			descriptor = translator.get().convertLuceneQuery( luceneQuery );
		}
	}
	catch (SearchException e) {
		// Ignore: no translator found
	}

	if ( descriptor == null ) {
		descriptor = new LuceneQueryDescriptor( luceneQuery );
	}

	return descriptor.createHSQuery( sf );
}
```

It's a bit awkward, because we use service references simply for creating a query. But apparently that's the only way to do it in a backend-independent way.
With this PR, the code above would be much simpler:

```java
private List<EntityInfo> query(Query luceneQuery, Sort sort) {
	ExtendedSearchIntegrator sf = sfHolder.getSearchFactory();
	HSQuery hsQuery = sf.createQueryDescriptor( luceneQuery, IndexedEntry.class ).createHSQuery( sf );
	return hsQuery
			.targetedEntities( Arrays.asList( new Class<?>[] { IndexedEntry.class } ) )
			.projection( "id" )
			.sort( sort )
			.queryEntityInfos();
}
```

Now the accesses to services are hidden.
It still feels a bit weird (especially providing the `ExtendedSearchIntegrator` and the target entities twice), but we can't do much better without bringing substantial changes to QueryDescriptor/HSQuery. I tried not to add anything to these interfaces so as not to break ascending compatibility. HSQuery having been part of the SPI for a very long time, we'd better wait for HS 6.0 before changing it.

See `FullTextSessionImpl#createFullTextQuery(QueryDescriptor, Class<?>...)` for another example of an actual simplification.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/impl/ImmutableSearchFactory.java
+++ b/engine/src/main/java/org/hibernate/search/engine/impl/ImmutableSearchFactory.java
@@ -14,6 +14,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.apache.lucene.analysis.Analyzer;
+import org.apache.lucene.search.Query;
 import org.hibernate.search.analyzer.impl.AnalyzerReference;
 import org.hibernate.search.analyzer.impl.LuceneAnalyzerReference;
 import org.hibernate.search.backend.impl.batch.DefaultBatchBackend;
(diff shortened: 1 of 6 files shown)


gsmet:
Looks generally like a good idea but as there is an SPI change - especially on `SearchIntegrator` -, I'll let it to @Sanne . Note that we already introduced a new method in `SearchIntegrator` in the 5.6 cycle.

yrodiere:
@gsmet Right, sorry. I edited the original message so as to give more information.

Sanne:
looks good. Merged.. thanks!

PR #1143, opened 2016-08-23
HSEARCH-2098 Clarify default value of @IndexedEmbedded#prefix()

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2098

### DIFF
--- a/engine/src/main/java/org/hibernate/search/annotations/IndexedEmbedded.java
+++ b/engine/src/main/java/org/hibernate/search/annotations/IndexedEmbedded.java
@@ -34,9 +34,9 @@ public @interface IndexedEmbedded {
 	String DEFAULT_NULL_TOKEN = "__DEFAULT_NULL_TOKEN__";
 
 	/**
-	 * Field name prefix, defaults to {@code propertyname.}.
+	 * Field name prefix.
 	 *
-	 * @return the field name prefix. Default to "."
+	 * @return the field name prefix. Default to ".", indicating that {@code propertyname.} will be used.
 	 */
 	String prefix() default ".";
 


engine/src/main/java/org/hibernate/search/annotations/IndexedEmbedded.java
```diff
@@ -34,9 +34,9 @@
 	String DEFAULT_NULL_TOKEN = "__DEFAULT_NULL_TOKEN__";
 
 	/**
-	 * Field name prefix, defaults to {@code propertyname.}.
+	 * Field name prefix.
 	 *
-	 * @return the field name prefix. Default to "."
+	 * @return the field name prefix. Default to ".", indicating that {@code propertyname.} should be used.
```
gsmet: "should"? I don't remember the specifics but IIRC, it is a "will", isn't it?

yrodiere: Well, that's the word used elsewhere when describing the behavior of Hibernate Search.
See for instance:

```
	/**
	 * Default value for the {@link #indexNullAs} parameter. Indicates that {@code null} values should not be indexed.
	 */
	String DO_NOT_INDEX_NULL = "__DO_NOT_INDEX_NULL__";

	/**
	 * Value for the {@link #indexNullAs} parameter indicating that {@code null} values should be indexed using the null
	 * token given through the {@link org.hibernate.search.cfg.Environment#DEFAULT_NULL_TOKEN} configuration property.
	 * If no value is given for that property, the token {@code _null_} will be used.
	 */
	String DEFAULT_NULL_TOKEN = "__DEFAULT_NULL_TOKEN__";
```

```
	/**
	 * Returns {@code true}, if the id of the embedded object should be included into the index,
	 * {@code false} otherwise. The default is {@code false}.
	 *
	 * <p><b>Note</b><br>:
	 * If the id property is explicitly listed via {@link #includePaths()}, then the id is included even if this value
	 * is {@code false}.
	 * </p>
	 *
	 * @return Returns {@code true}, if the id of the embedded object should be included into the index,
	 * {@code false} otherwise.
	 */
	boolean includeEmbeddedObjectId() default false;
```

And so on.

Anyway, I'll change that.

yrodiere: @gsmet Fixed. I guess I should create a JIRA ticket about the wording in the rest of the Javadoc?

gsmet: No I think the existing wording is OK. The context where "should" is used is not the same.

But see "the token {@code _null_} will be used" in what you quoted.

yrodiere: I think we won't be able to agree on that. Anyway, is it okay with [the new wording](https://github.com/hibernate/hibernate-search/pull/1143/commits/cf2d367c7dd3c02ca21d558cf96f813e7426f080)?

gsmet:
Rebased and merged! Thanks!

engine/src/main/java/org/hibernate/search/annotations/IndexedEmbedded.java
yrodiere: Never mind, you merged when I was writing this. Thanks!

PR #1144, opened 2016-08-23
HSEARCH-2281 Omit some attributes on queries when they're not required

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2281

The ticket was mainly about "boost", but while I was at it, I also dealt with "slop" and "fuzziness". I hope it's alright; I can create another ticket if it's really necessary.

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/JsonBuilder.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/JsonBuilder.java
@@ -41,6 +41,15 @@ public class JsonBuilder {
 		return new JsonBuilder.Object( jsonObject );
 	}
 
+	/**
+	 * An interface for reusable objects that append data to a JSON object.
+	 * <p>Eases the use of conditional statements within this chainable API.
+	 * @author Yoann Rodiere
+	 */
+	public interface JsonAppender<T> {
+		void append(T appendable);
+	}
+
 	public static class Array {
 		private JsonArray jsonArray = new JsonArray();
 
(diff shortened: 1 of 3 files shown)


gsmet:
Just a suggestion. Might not be a good idea after all but as I thought about it, I might as well post it.

I'm wondering if it would be more readable with a method like:
```
private static JsonObject decorateBoost(JsonObject json, Query query) {
}
```

You could keep most of the existing code as is and just add a `decorateBoost` call. We could even imagine to simply call it `decorate` and be more future proof.

Maybe we could also have:
```
private static JsonObject decorate(JsonObject json, PhraseQuery query) {
}

private static JsonObject decorate(JsonObject json, RemotePhraseQuery query) {
}

private static JsonObject decorate(JsonObject json, RemoteMatchQuery query) {
}
```

yrodiere:
I almost did this, but on second thought I reconsidered:

 * it creates overly complicated nested statements, which some may consider as actually less readable
 * future-proofness is not a primary concern since the Lucene team is currently moving things **out** of the `Query` class and using query wrapping instead of inheritance (see the example of the `BoostQuery`, which wraps a `Query`, as opposed to the previous `boost` attribute in `Query`).

The current code is readable. It may not be flexible, but it's readable: the purpose of each statement is obvious.

If the primary concern is to always write chained code, the solution may be to simply add a `.merge(JSONBuilder.Object)` method to `JSONBuilder.Object`. This way, we could have a common `JSONBuilder.Object createQueryProperties(Query)` method and would do something like this:

```
						JsonBuilder.object().add( field, properties )
								JsonBuilder.object()		
										.addProperty( "value", termQuery.getTerm().text() )		
										.merge( createQueryProperties(termQuery) ) 
```

Then, the nested statements would at least follow the data structure, which in my opinion would be easier to read.

yrodiere:
Here, I reworked the PR so that we still use chaining. I hope it's better than my previous attempt...

yrodiere:
Jenkins, retest this please

Sanne:
Nice work! Merging

PR #1145, opened 2016-08-23
HSEARCH-2276 Provide an @Rule for ram index manager autocleanup

yrodiere:
Prevents the thread leaks mentioned in https://hibernate.atlassian.net/browse/HSEARCH-2276


engine/src/test/java/org/hibernate/search/testsupport/indexmanager/RamIndexManagerFactory.java
```diff
@@ -0,0 +1,74 @@
...
+ * 		// Here, use the managers for setting up your environment
+ * 	}
+ * </pre></code>
+ *
+ * @author Sanne Grinovero (C) 2011 Red Hat Inc.
+ * @author Yoann Rodiere
+ */
+public class RamIndexManagerFactory extends ExternalResource implements TestRule {
+
+	private final Collection<IndexManager> managersToDestroy = new ArrayList<>();
```
gsmet: We usually keep List, and don't use Collection in this case.

gsmet:
Apart from my minor comment, looks good to me. You can push it, once fixed.

yrodiere:
Merged. Thanks!

orm/src/test/java/org/hibernate/search/test/configuration/IndexManagerOverrideTest.java
```diff
@@ -27,14 +27,18 @@
...
+
 	@Test
 	public void verifyIndexExclusivity() {
 		FullTextSessionBuilder builder = new FullTextSessionBuilder();
 		FullTextSession ftSession = builder
 			.setProperty( "hibernate.search.Book.indexmanager", "near-real-time" )
 			.setProperty(
 					"hibernate.search." + Foo.class.getName() + ".indexmanager",
-					"org.hibernate.search.testsupport.indexmanager.RamIndexManager"
+					FooIndexManager.class.getName()
```
Sanne: Why did you change this?

yrodiere: `RamIndexManager`'s constructor became private because it now has to be instanciated through `RamIndexManagerFactory`, so that it can be destroyed after the tests. Thus it could not be used here anymore.
I though I could change this because, if I understood this test correctly, the actual class of the index manager does not matter here, we're just checking that Hibernate Search instantiates the type we asked for. So, `RamIndexManager` or not, it does not matter.
Besides, **if** we did rely on some feature of `RamIndexManager` here, there probably was something wrong, since Hibernate Search didn't call `RamIndexManager.makeRamDirectory` but simply the public default constructor, which skips most of the manager configuration.

Sanne: Ok, I was just wondering. My reason to use the Ram IndexManager often, is to try to keep the test execution quick.
(The default one creates stuff on disk, and it gets slow when thousand of tests do it)

PR #1146, opened 2016-08-24
HSEARCH-2168 Raise exceptions via Log API in Elasticsearch backend

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2168


Sanne:
Nice! I merged it, but changed a single message to expand the acronym "ES" and use "Elasticsearch"

PR #1147, opened 2016-08-25
HSEARCH-2257 Add Elasticsearch-specific TOOK and TIMED_OUT projections

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2257

@gsmet suggested in the original ticket that maybe only "took" and "timed_out" are strictly necessary, implying "_shards" may be unnecessary (["_shards"](https://www.elastic.co/guide/en/elasticsearch/reference/current/_the_search_api.html) tells us how many shards were searched, as well as a count of the successful/failed searched shards). I tend to agree with that as it seems like advanced usage to me, and I think HS 6.0 would be a better time for introducing advanced query metadata.
Let me know if I'm mistaken, though.


Sanne:
nice, merged!

PR #1148, opened 2016-08-25
HSEARCH-2171 Don't use analyzers by default on non-string types

yrodiere:
Here is a fix for [HSEARCH-2171](https://hibernate.atlassian.net/browse/HSEARCH-2171) , which mainly expresses concerns about how non-string values (booleans, dates, ...) are converted to ES query values.
It turns out the problem has already been addressed **for dates** as part of [HSEARCH-2246](https://hibernate.atlassian.net/browse/HSEARCH-2246) (commit d820d1f0791743c02e62366e3ee254d92c44d5ea). **But** one part of the issue, the fact that some types should not see their string value analyzed, has not been addressed for other types (booleans, longs, ...). That's the purpose of this PR.

Note that this may break some existing user code. Because we don't analyze values for these types, this kind of code won't work anymore:

```
			Query query = queryBuilder
					.keyword()
						.onField( "myNumericFieldStoredAsString" )
						.matching( "12 13" ) // Will match 12 or 13
					.createQuery();
```

That's true for the types we address in this PR as well as for dates, addressed in d820d1f0791743c02e62366e3ee254d92c44d5ea.
I'm not sure it was expected behavior, though. This requires overriding the default analyzer, in particular, since the default in Hibernate Search is to use a `LowerCaseTokenizer`, which ignores digits. 

On the other hand, users with custom default analyzers may have had surprises in the past, especially when tokenizing on "." (searching for "12.2" may have matched "12" or "2" but not "12.2").

So, the question is: is the current behavior (number, boolean and URL fields are analyzed) something we want to preserve, or is it a bug? If it's a bug, this PR should be reviewed and merged, otherwise we can put it straight to the trash bin.

### DIFF
--- a/elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/ElasticsearchDSLIT.java
+++ b/elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/ElasticsearchDSLIT.java
@@ -8,6 +8,11 @@ package org.hibernate.search.elasticsearch.test;
 
 import static org.hibernate.search.elasticsearch.testutil.JsonHelper.assertJsonEquals;
 
+import java.util.Calendar;
+import java.util.Date;
+import java.util.Locale;
+import java.util.TimeZone;
+
 import javax.persistence.Entity;
 import javax.persistence.GeneratedValue;
 import javax.persistence.Id;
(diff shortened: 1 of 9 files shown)


Sanne:
I consider this a bug, and merged this fix. Thanks!

I added one commit: cc7051b

PR #1149, opened 2016-08-25
HSEARCH-2170 Convert the GsonHolder to a Service to ensure cleanup

yrodiere:
Mainly because Gson holds references to ThreadLocals, so having it stored in a static variable is not desirable.
JsonParser has been moved to private static variables, though, since it does not contain any data.

https://hibernate.atlassian.net/browse/HSEARCH-2170


Sanne:
nice! Merging

PR #1150, opened 2016-08-26
HSEARCH-2336 Upgrade Unitils to 3.4.3

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2336

The main thing to check is: is it okay to upgrade to another major version of Easymock, even though we may expose this version through the `hibernate-search-testing` module?

The release notes of EasyMock 3 are available at the bottom of this page: http://easymock.sourceforge.net/EasyMock3_0_Documentation.html
From the release notes, it doesn't seem there's been a major overhaul on 3.x, and the build passes without changing anything other than the versions in the POM.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -547,12 +547,12 @@
             <dependency>
                 <groupId>org.easymock</groupId>
                 <artifactId>easymock</artifactId>
-                <version>2.4</version>
+                <version>3.2</version>
             </dependency>
             <dependency>
                 <groupId>org.unitils</groupId>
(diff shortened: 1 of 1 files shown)


Sanne:
> The main thing to check is: is it okay to upgrade to another major version of Easymock, even though we may expose this version through the hibernate-search-testing module?

Yes that's ok as we never really advertised usage of the hibernate-search-testing module .. we probably should in future, but it needs polishing and documentation on how to use it.

Sanne:
merged, thanks!

PR #1151, opened 2016-08-26
HSEARCH-2262 In the Elasticsearch mapping, validate the type of the null_value/indexNullAs property

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2262

~~This currenly only works when running on JVMs with UTC timezone, and will fail miserably on JVMs with GMT+1 default timezone for instance. Either we fix [HSEARCH-2335](https://hibernate.atlassian.net/browse/HSEARCH-2335) and this PR will work as is, or we decide not to, and this PR will need some work.~~ Fixed, see comments.

### DIFF
--- a/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
+++ b/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
@@ -135,6 +135,19 @@ Simply use familiar annotations from <<search-mapping>>.
 The name of the index will be the lowercased name provided to `@Indexed` (non qualified class name by default).
 Hibernate Search will map the fully qualified class name to the Elasticsearch type.
 
+==== Annotation specificities
+
+===== Field.indexNullAs
+
+The `org.hibernate.search.annotations.Field` annotation allows you to provide a replacement value for null properties through the `indexNullAs` attribute (see <<field-annotation>>), but this value must be provided as a string.
+
+In order for your value to be understood by Hibernate Search (and Elasticsearch), the provided string must follow one of those formats:
+
+ * For string values, no particular format is required.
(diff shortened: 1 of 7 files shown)


Sanne:
This isn't waiting on any other open PR right?

yrodiere:
@Sanne This currently doesn't work, as explained in the first comment. It's not waiting on a PR alright, but it's waiting on HSEARCH-2335, which hasn't been fixed yet and must be fixed before merging.

Sanne:
ok. Do you plan to take HSEARCH-2335 ?

yrodiere:
Well, it's basically a behavior change that fixes some bugs, but *will*
introduce ascending compatibility issues... Which is why I didn't address
it right away. Maybe you can give your opinion on that one? I think it
should be addressed, but someone else may think it's invalid ("works as
expected")...

If it's deemed valid, of course I can take it as soon as I'm done with the
VALIDATE/schema export stuff on Elasticsearch.

Yoann Rodière <yrodiere@redhat.com>
Software Engineer
Red Hat / Hibernate NoORM Team

yrodiere:
See the discussion in https://hibernate.atlassian.net/browse/HSEARCH-2335: I was wrong. I will fix the tests so that they don't rely on the local timezone anymore, and will tackle https://hibernate.atlassian.net/browse/HSEARCH-2355 later.

yrodiere:
The tests have been fixed. Ready for review!

Sanne:
Nice one, merging. The only thing which got me a bit confused is the usage of test Suite to make the two tests :)

Sanne:
merged

PR #1152, opened 2016-09-01
HSEARCH-2337 Async backend not properly flushed

gustavonalle:
https://hibernate.atlassian.net/browse/HSEARCH-2337


gustavonalle:
One issue I caught is after calling ```flushAndReleaseResources``` in the the async backend, the scheduler that commits at regular intervals will still be operational, when ideally is should not. I will update the PR shortly.

Sanne:
thanks @gustavonalle . We have a release today, I'll wait for your fix.

gustavonalle:
Ready for review.

Sanne:
Thanks, it looks good except an unlikely race between stopping/starting the same backend.. I already have a local fix so don't worry about it. Testing, then will merge.

Thanks!

Sanne:
merged

PR #1155, opened 2016-09-12
HSEARCH-2342 Upgrade to WildFly 10.1.0.Final

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2342


yrodiere:
Seems good, no unexpected version overrides.
Merged, thanks!

PR #1156, opened 2016-09-12
HSEARCH-2343 Possible resource leak in IndexControl MBean

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2343

### DIFF
--- a/orm/src/main/java/org/hibernate/search/jmx/impl/IndexControl.java
+++ b/orm/src/main/java/org/hibernate/search/jmx/impl/IndexControl.java
@@ -13,10 +13,12 @@ import javax.naming.InitialContext;
 import org.hibernate.CacheMode;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
+import org.hibernate.Transaction;
 import org.hibernate.engine.config.spi.ConfigurationService;
 import org.hibernate.search.FullTextSession;
 import org.hibernate.search.Search;
 import org.hibernate.search.exception.SearchException;
+import org.hibernate.search.hcore.util.impl.ContextHelper;
 import org.hibernate.search.engine.integration.impl.ExtendedSearchIntegrator;
 import org.hibernate.search.engine.service.classloading.spi.ClassLoaderService;
 import org.hibernate.search.engine.service.classloading.spi.ClassLoadingException;
@@ -83,8 +85,7 @@ public class IndexControl implements IndexControlMBean {
 
 		SessionFactory factory = getSessionFactory();
 		Session session = factory.openSession();
-		FullTextSession fulltextSession = Search.getFullTextSession( session );
-		try {
+		try ( FullTextSession fulltextSession = Search.getFullTextSession( session ) ) {
 			fulltextSession.createIndexer( clazz )
 					.batchSizeToLoadObjects( batchSize )
 					.cacheMode( CacheMode.NORMAL )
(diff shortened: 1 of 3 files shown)


orm/src/main/java/org/hibernate/search/jmx/impl/IndexControl.java
```diff
@@ -83,8 +85,7 @@ public void index(String entity) {
 
 		SessionFactory factory = getSessionFactory();
 		Session session = factory.openSession();
-		FullTextSession fulltextSession = Search.getFullTextSession( session );
-		try {
+		try ( FullTextSession fulltextSession = Search.getFullTextSession( session ) ) {
```
yrodiere: I think, to be correct, this should be:

```java

		try ( Session session = factory.openSession() ;
				FullTextSession fulltextSession = Search.getFullTextSession( session ) ) {
```

orm/src/main/java/org/hibernate/search/jmx/impl/IndexControl.java
```diff
@@ -95,33 +96,27 @@ public void index(String entity) {
...
 		Class<?> clazz = getEntityClass( entity );
 
 		SessionFactory factory = getSessionFactory();
-		Session session = factory.openSession();
-		FullTextSession fullTextSession = Search.getFullTextSession( session );
-		fullTextSession.beginTransaction();
-		fullTextSession.purgeAll( clazz );
-		fullTextSession.getTransaction().commit();
-		session.close();
+		try ( Session session = factory.openSession() ) {
```
yrodiere: Similar to the issue above, I think. Just in case the `FullTextSession` wrapper has resources of its own to close, we should do something like that:

```java
        try ( Session session = factory.openSession() ;
                FullTextSession fulltextSession = Search.getFullTextSession( session ) ) {
```

Sanne: No we only need to close the underlying `Session`, you don't want to close both.

yrodiere: I guess you mean because the `FullTextSession` does nothing more than delegating to `Session`? I hadn't noticed that.

So you're right, it is not *required* to close both. It shouldn't hurt (`Closeable#close` is expected to be idempotent as per its javadoc), but the current implementation makes it useless.

Well, never mind.

yrodiere:
@Sanne Only got one comment : https://github.com/hibernate/hibernate-search/pull/1156#discussion-diff-78393899 . The other one may go a bit too far.
If you agree with my first comment, I'll do the change and merge.

orm/src/main/java/org/hibernate/search/jmx/impl/IndexControl.java
Sanne: It is not idempotent, so it's actually dangerous to close it twice. The JPA spec requires throwing an exception on closing a closed EntityManager.

yrodiere: Ok, a bit surprising, I didn't know that. A shame the `Closeable` contract isn't enforced.
I'll keep that in mind, thanks.

Sanne: n.p. you highlighted and interesting flaw :)

yrodiere: @Sanne Interestingly enough, in the lastest versions of ORM, this seems to be fixed:

 * `SessionImpl` calls `super.close()`: https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/internal/SessionImpl.java#L400
 * The superclass' `close()` method simply returns if the session is already closed: https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/internal/AbstractSharedSessionContract.java#L289

The issue is still there in the ORM version we're using, though.

yrodiere:
Closing and following up on #1157
I will merge once the build is okay. Thanks, @Sanne!

orm/src/main/java/org/hibernate/search/jmx/impl/IndexControl.java
Sanne: sorry, I forgot to reply on this one. You're right of course!

PR #1158, opened 2016-09-13
Updating some integration test components to pave the road to ORM 5.2

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2346
 - https://hibernate.atlassian.net/browse/HSEARCH-2347

These changes are extracted from my ORM 5.2 branch but can be applied on master already.

### DIFF
--- a/integrationtest/spring/src/test/resources/persistence.xml
+++ b/integrationtest/spring/src/test/resources/persistence.xml
@@ -8,7 +8,7 @@
 <persistence xmlns="http://java.sun.com/xml/ns/persistence" version="1.0">
 
     <persistence-unit name="integration" transaction-type="JTA">
-        <provider>org.hibernate.ejb.HibernatePersistence</provider>
+        <provider>org.hibernate.jpa.HibernatePersistenceProvider</provider>
         <class>org.hibernate.search.test.integration.jtaspring.Snert</class>
         <class>org.hibernate.search.test.integration.jtaspring.Container</class>
         <class>org.hibernate.search.test.integration.jtaspring.Box</class>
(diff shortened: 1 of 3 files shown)


yrodiere:
@Sanne So I reckon it's ready for review? Just asking because the label is missing, so I'm note absolutely sure.

Sanne:
@yrodiere yes, please have a look and integrate it. I have more coming but this needs to go first ;-)

integrationtest/spring/src/test/resources/persistence.xml
```diff
@@ -17,7 +17,7 @@
     </persistence-unit>
 
     <persistence-unit name="jbossjta" transaction-type="JTA">
-        <provider>org.hibernate.ejb.HibernatePersistence</provider>
+        <provider>org.hibernate.jpa.HibernatePersistenceProvider</provider>
```
yrodiere: Seems like we could use the new class in `JBossTSIT.java`, too (see [here](https://github.com/hibernate/hibernate-search/blob/master/integrationtest/narayana/src/test/java/org/hibernate/search/test/integration/jbossjta/JBossTSIT.java#L88)).
Or was it on purpose that you skipped it?

yrodiere:
@Sanne Seems good to me, apart from the missing replacement mentioned above (maybe). Feed free to merge, or tell me if you want me to take care of that replacement.

integrationtest/spring/src/test/resources/persistence.xml
Sanne: thanks! I had actually changed that but then sent the wrong branch .. good catch! Merging the right branch..

PR #1160, opened 2016-09-14
ORM 5.2 upgrade PREVIEW

Sanne:
This is a PREVIEW not to be merged in master: master is still aiming at Hibernate ORM `5.0.x` and `5.1.x` compatibility.

Still, I'd appreciate an early review: a couple of changes in ORM made this more complex than usual.
Among these:
 - the class hierarchy: Session now implementing EntityManager, with all its consequences on Query, etc..
 - stricter on what is allowed in CMT vs non-CMT transaction modes

Since it includes a version switch to `5.7.0-SNAPSHOT`, I'll also deploy a snapshot on Nexus for the sake of testing volunteers.

Needs still:

- [ ] to avoid using the old Criteria API, as it generates logged warnings
- [ ] check for other usage of deprecated APIs from ORM
- [ ] see if we can incorporate our JPA wrappers into the Session ones (i.e. unify things such as `org.hibernate.search.jpa.FullTextQuery` and `org.hibernate.search.FullTextQuery<R>`
- [ ] Clarify this ORM update effectively gets us to require Java 8

### DIFF
--- a/engine/src/main/java/org/hibernate/search/query/engine/spi/HSQuery.java
+++ b/engine/src/main/java/org/hibernate/search/query/engine/spi/HSQuery.java
@@ -248,4 +248,10 @@ public interface HSQuery extends ProjectionConstants {
 	 * @return {@code this}  to allow for method chaining
 	 */
 	HSQuery setSpatialParameters(Coordinates center, String fieldName);
+
+	/**
+	 * @return a human readable representation of the Query in String format
+	 */
+	String getQueryString();
+
 }
--- a/integrationtest/performance/src/test/java/org/hibernate/search/test/performance/scenario/TestScenario.java
+++ b/integrationtest/performance/src/test/java/org/hibernate/search/test/performance/scenario/TestScenario.java
@@ -19,9 +19,13 @@ import java.sql.PreparedStatement;
 import java.sql.SQLException;
 import java.util.Properties;
 
+import javax.transaction.TransactionManager;
+
 import org.apache.commons.lang.time.StopWatch;
+import org.hibernate.Session;
 import org.hibernate.SessionFactory;
-import org.hibernate.Transaction;
+import org.hibernate.engine.spi.SessionImplementor;
+import org.hibernate.engine.transaction.jta.platform.spi.JtaPlatform;
 import org.hibernate.search.FullTextSession;
 import org.hibernate.search.Search;
 import org.hibernate.search.test.performance.model.Book;
@@ -170,21 +174,17 @@ public abstract class TestScenario {
 
 	protected void performCleanUp(SessionFactory sf) {
 		log( "starting clean up phase" );
+		try ( Session s = sf.openSession() ) {
+			final SessionImplementor session = (SessionImplementor) s;
+			FullTextSession fulltextSession = Search.getFullTextSession( s );
+			beginTransaction( session );
+			s.createNativeQuery( "delete from book_author where book_id < :id" ).setParameter( "id", initialOffset ).executeUpdate();
+			s.createNativeQuery( "delete from book where id < :id" ).setParameter( "id", initialOffset ).executeUpdate();
+			s.createNativeQuery( "delete from author where id < :id" ).setParameter( "id", initialOffset ).executeUpdate();
 
-		FullTextSession s = Search.getFullTextSession( sf.openSession() );
-		try {
-			Transaction tx = s.beginTransaction();
-			s.createSQLQuery( "delete from book_author where book_id < :id" ).setParameter( "id", initialOffset ).executeUpdate();
-			s.createSQLQuery( "delete from book where id < :id" ).setParameter( "id", initialOffset ).executeUpdate();
-			s.createSQLQuery( "delete from author where id < :id" ).setParameter( "id", initialOffset ).executeUpdate();
-
-			s.purgeAll( Book.class );
-			s.flush();
-			s.flushToIndexes();
-			tx.commit();
-		}
-		finally {
-			s.close();
+			fulltextSession.purgeAll( Book.class );
+			fulltextSession.flushToIndexes();
+			commitTransaction( session );
 		}
 	}
 
@@ -228,6 +228,44 @@ public abstract class TestScenario {
 		}
 
 		ctx.startAndWait();
+	}
+
+	private void commitTransaction(SessionImplementor session) {
+		if ( session.getTransactionCoordinator().getTransactionCoordinatorBuilder().isJta() ) {
+			TransactionManager transactionManager = lookupTransactionManager( session );
+			try {
+				transactionManager.commit();
+			}
+			catch (Exception e) {
+				throw new RuntimeException( e );
+			}
+		}
+		else {
+			session.getTransaction().commit();
+		}
+	}
+
+	private void beginTransaction(SessionImplementor session) {
+		if ( session.getTransactionCoordinator().getTransactionCoordinatorBuilder().isJta() ) {
+			TransactionManager transactionManager = lookupTransactionManager( session );
+			try {
+				transactionManager.begin();
+			}
+			catch (Exception e) {
+				throw new RuntimeException( e );
+			}
+		}
+		else {
+			session.getTransaction().begin();
+		}
+	}
+
+	private static TransactionManager lookupTransactionManager(SessionImplementor session) {
+		return session
+				.getSessionFactory()
+				.getServiceRegistry()
+				.getService( JtaPlatform.class )
+				.retrieveTransactionManager();
 	}
 
 }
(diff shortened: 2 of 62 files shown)


integrationtest/performance/src/test/java/org/hibernate/search/test/performance/scenario/TestScenario.java
```diff
@@ -170,21 +174,17 @@ protected void performWarmUp(SessionFactory sf) {
...
-			s.purgeAll( Book.class );
-			s.flush();
-			s.flushToIndexes();
-			tx.commit();
-		}
-		finally {
-			s.close();
+		try ( Session s = sf.openSession() ) {
+			final SessionImplementor session = (SessionImplementor) s;
+			FullTextSession fulltextSession = Search.getFullTextSession( sf.openSession() );
```
gunnarmorling: Is it intentional to open a second session here?

orm/src/main/java/org/hibernate/search/query/hibernate/impl/FullTextQueryImpl.java
```diff
@@ -173,7 +181,7 @@ private Loader getProjectionLoader(ObjectLoaderBuilder loaderBuilder) {
 	}
 
 	@Override
-	public ScrollableResults scroll() {
+	public ScrollableResultsImpl scroll() {
```
gunnarmorling: What's the reason for returning the impl type?

gunnarmorling:
> unify things such as org.hibernate.search.jpa.FullTextQuery and org.hibernate.search.FullTextQuery<R>

That seems reasonable given the merge in ORM itself.

orm/src/main/java/org/hibernate/search/query/hibernate/impl/FullTextQueryImpl.java
```diff
@@ -176,7 +181,7 @@ private Loader getProjectionLoader(ObjectLoaderBuilder loaderBuilder) {
 	}
 
 	@Override
-	public ScrollableResultsImplementor scroll() {
+	public ScrollableResultsImpl scroll() {
```
gsmet: I'm a bit surprised we return an Impl, here?

engine/src/main/java/org/hibernate/search/query/engine/spi/HSQuery.java
```diff
@@ -248,4 +248,9 @@
 	 * @return {@code this}  to allow for method chaining
 	 */
 	HSQuery setSpatialParameters(Coordinates center, String fieldName);
+
+	/**
+	 * @return
```
gsmet: Very minor, Javadoc.

orm/src/test/java/org/hibernate/search/test/interceptor/IndexingActionInterceptorTest.java
```diff
@@ -46,7 +47,7 @@ public void setUp() throws Exception {
 	@Override
 	@After
 	public void tearDown() throws Exception {
-		if ( fullTextSession.getTransaction().getStatus() != TransactionStatus.ACTIVE ) {
+		if ( getTransactionStatus( fullTextSession) != TransactionStatus.ACTIVE ) {
```
gsmet: Formatting (missing space after fullTextSession).

orm/src/main/java/org/hibernate/search/batchindexing/impl/IdentifierProducer.java
```diff
@@ -105,9 +105,18 @@ private void inTransactionWrapper(StatelessSessionImpl upperSession) throws Exce
 		}
 		try {
 			Transaction transaction = session.accessTransaction();
-			transaction.begin();
-			loadAllIdentifiers( session );
-			transaction.commit();
+			final boolean controlTransactions = ! transaction.isActive();
```
gsmet: Very minor but the space after the ! looks weird. I had to read it twice.

orm/src/main/java/org/hibernate/search/jmx/impl/IndexControl.java
```diff
@@ -109,13 +110,22 @@ public void optimize(String entity) {
...
 		Class<?> clazz = getEntityClass( entity );
-
 		SessionFactory factory = getSessionFactory();
 		try ( Session session = factory.openSession() ) {
 			FullTextSession fullTextSession = Search.getFullTextSession( session );
-			Transaction transaction = fullTextSession.beginTransaction();
-			fullTextSession.purgeAll( clazz );
-			transaction.commit();
+			Transaction transaction = ( (SessionImplementor) session ).accessTransaction();
+			final boolean controlTransactions = ! transaction.isActive();
```
gsmet: Same here.

gsmet:
@Sanne looks good to me overall apart from some minor comments and questions. The transaction stuff doesn't look pretty but I suppose we don't have much choice.

orm/src/main/java/org/hibernate/search/batchindexing/impl/IdentifierProducer.java
```diff
@@ -93,21 +93,30 @@ public void run(StatelessSession upperSession) throws Exception {
...
+				session = (StatelessSessionImpl) sessionFactory.withStatelessOptions().tenantIdentifier( tenantId ).openStatelessSession();
 			}
 		}
 		try {
-			Transaction transaction = session.getTransaction();
-			transaction.begin();
-			loadAllIdentifiers( session );
-			transaction.commit();
+			Transaction transaction = session.accessTransaction();
+			final boolean controlTransactions = ! transaction.isActive();
```
gunnarmorling: If you need this pattern several times, you could extract in a separate method, using Lambda style:

    runInTransactionIfNeeded( session, session -> { loadAllIdentifiers( session ); } );

Sanne: we're still supposed to be JDK 7 compatible ;)

orm/src/main/java/org/hibernate/search/batchindexing/impl/IdentifierProducer.java
Sanne: We do this regularly, to make sure the `!` is visible.
We simply decided to not enforce it on checkstyle but I still prefer it.

Sometimes to highlight it I might even write `( x == false )`

orm/src/test/java/org/hibernate/search/test/interceptor/IndexingActionInterceptorTest.java
Sanne: thanks. Checkstyle bug!

gunnarmorling:
Why is that? ORM 5.2 itself moved to Java 8.

orm/src/main/java/org/hibernate/search/query/hibernate/impl/FullTextQueryImpl.java
Sanne: this was required when I originally made this for ORM 5.2.0, seems I can switch back to `ScrollableResultsImplementor` today with 5.2.1.. thanks.

orm/src/main/java/org/hibernate/search/query/hibernate/impl/FullTextQueryImpl.java
Sanne: The API requires it; although as noticed in the next commit it seems the SPI `ScrollableResultsImplementor` is good enough now.. changing, thanks!

orm/src/main/java/org/hibernate/search/batchindexing/impl/IdentifierProducer.java
gunnarmorling: Why is that? ORM 5.2 itself moved to Java 8.

integrationtest/performance/src/test/java/org/hibernate/search/test/performance/scenario/TestScenario.java
Sanne: awful! good catch

Sanne:
> Why is that? ORM 5.2 itself moved to Java 8.

Ok, I stand corrected. Adding to the other tasks then: to update the build and docs to explicitly require it.

orm/src/main/java/org/hibernate/search/batchindexing/impl/IdentifierProducer.java
Sanne: BTW this is pretty much what I proposed on the dev list, but there it's about making it a public API of Hibernate.
I'll not introduce such an helper here though to avoid reuse in a different context, as this specific check of `isActive()` isn't the universally correct strategy, it only is in this context as this block is wrapped by an `org.hibernate.search.batchindexing.impl.OptionallyWrapInJTATransaction` :-/

I'll agree that this detail doesn't look good, we'll come back on how to deal with CMT / JTA and all that. Will have to do for now though as I don't want to refactor this area heavily with 5.6 still in progress.

Sanne:
this has been polished, rebased and promoted as upstream branch 5.7

PR #1162, opened 2016-09-21
HSEARCH-2352 Index.NO is being ignored for numeric fields (and also boolean or date fields for Elasticearch)

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2352

Necessary for HSEARCH-2260.

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
@@ -74,6 +74,15 @@ public class ElasticsearchIndexManager implements IndexManager, RemoteAnalyzerPr
 
 	private static final String ANALYZED = "analyzed";
 	private static final String NOT_ANALYZED = "not_analyzed";
+	private static final String NOT_INDEXED = "no";
+
+	private static final String TYPE_STRING = "string";
+	private static final String TYPE_BOOLEAN = "boolean";
+	private static final String TYPE_DATE = "date";
+	private static final String TYPE_INTEGER = "integer";
+	private static final String TYPE_LONG = "long";
+	private static final String TYPE_FLOAT = "float";
+	private static final String TYPE_DOUBLE = "double";
 
 	private String indexName;
 	private String actualIndexName;
@@ -298,7 +307,7 @@ public class ElasticsearchIndexManager implements IndexManager, RemoteAnalyzerPr
 
 			if ( multitenancyEnabled ) {
 				JsonObject field = new JsonObject();
-				field.addProperty( "type", "string" );
+				field.addProperty( "type", TYPE_STRING );
 				field.addProperty( "index", NOT_ANALYZED );
 				properties.add( DocumentBuilderIndexedEntity.TENANT_ID_FIELDNAME, field );
 			}
@@ -419,7 +428,7 @@ public class ElasticsearchIndexManager implements IndexManager, RemoteAnalyzerPr
 			else {
 				// the fields potentially created for the spatial hash queries
 				JsonObject field = new JsonObject();
-				field.addProperty( "type", "string" );
+				field.addProperty( "type", TYPE_STRING );
 				field.addProperty( "index", NOT_ANALYZED );
 
 				getOrCreateProperties( payload, fieldName ).add( fieldName, field );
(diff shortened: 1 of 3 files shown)


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -361,6 +361,10 @@ private void addFieldMapping(JsonObject payload, EntityIndexBinding descriptor,
 
 		String index = getIndex( descriptor, fieldMetadata );
 		field.addProperty( "index", index );
+		if ( "no".equals( index ) && FieldHelper.isSortableField(descriptor, fieldMetadata.getFieldName()) ) {
```
gsmet: Formatting (padding in parenthesis).

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -458,28 +462,34 @@ private JsonObject addFieldMapping(JsonObject payload, FacetMetadata facetMetada
...
+				return NOT_ANALYZED;
+			}
 		}
 
-		Field.Index index = bridgeDefinedField.getIndex();
-		return getIndexValue( index );
+		return index;
 	}
 
 	private String getIndexValue(Field.Index index) {
```
gsmet: Note: this comment is mostly orthogonal to your patch but I think it could benefit from it.

You will probably do that when you'll fix the norms issue but I'm wondering if we should introduce a private `IndexOptions` class with all these information.

I think I would prefer if we pass an additional `canBeAnalyzed` boolean to `getIndexValue` and keep all the logic in getIndexValue instead of changing the value in the respective `getIndex`. This would allow to return an immutable `IndexOptions` from `getIndexValue` (which would be renamed `getIndexOptions`).

In `IndexOptions`, we would have `getIndexValue()`, `isIndexed()` and in the future `hasNorms()`.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
yrodiere: Fixed, thanks.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
yrodiere: Right! I just amended the last commit, implementing what you suggested.

I changed it a bit, though: in order to avoid having to write `field.addProperty("xx", indexOptions.getXX())` multiple times in the code, I didn't use a IndexOptions structure, but rather transformed the `getIndexValue` method into a `addIndexOptions` method. I can't explain it well, but the code should be a bit clearer (I hope).

Also, I used this opportunity to simplify the "canBeAnalyzed" condition: if you look at `getFieldType()`, you'll see that the previous condition was listing every type except "string". So comparing the type to "string" is simpler.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -457,44 +440,44 @@ private JsonObject addFieldMapping(JsonObject payload, FacetMetadata facetMetada
...
+			String analyzerName = analyzerName( binding.getDocumentBuilder().getBeanClass(), fieldName, analyzerReference );
+			if ( analyzerName != null ) {
+				field.addProperty( "analyzer", analyzerName );
+			}
+		}
+	}
+
+	private boolean canTypeBeAnalyzed(String fieldType) {
+		// Only strings can be analyzed
+		return "string".equals( fieldType );
```
gsmet: Time to add constants for the types.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -457,44 +440,44 @@ private JsonObject addFieldMapping(JsonObject payload, FacetMetadata facetMetada
...
+				break;
 			case NO:
-				return "no";
+				elasticsearchIndex = "no";
 			default:
 				throw new AssertionFailure( "Unexpected index type: " + index );
 		}
+		field.addProperty( "index", elasticsearchIndex );
+
+		if ( "no".equals( elasticsearchIndex ) && FieldHelper.isSortableField( binding, fieldName ) ) {
```
gsmet: Maybe add a constant too for "no"?

gsmet:
Yes, I prefer this approach, it's far clearer. I think you should add a couple of constants to make it less fragile - see my comments inline. Apart from that, it looks good.

yrodiere:
Done. I'll merge once the build has passed.

yrodiere:
Merged. Thanks!

PR #1163, opened 2016-09-22
HSEARCH-2357 Make FieldBridges in tests implement MetadataProvidingFieldBridge

gsmet:
I found this one in one of my work trees so I thought I might as well push it.

 * https://hibernate.atlassian.net/browse/HSEARCH-2357

### DIFF
--- a/elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/AgeBridge.java
+++ b/elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/AgeBridge.java
@@ -7,13 +7,15 @@
 package org.hibernate.search.elasticsearch.test;
 
 import org.apache.lucene.document.Document;
-import org.hibernate.search.bridge.FieldBridge;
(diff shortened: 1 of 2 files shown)


yrodiere:
So it's required to implement this interface even when using the original field name? The documentation states:

> For class bridges and field bridges creating multiple fields, make sure to make your bridge implementation also implement the MetadataProvidingFieldBridge contract.

Anyway, since the main tests (in -engine and -orm) are actually executed on the Elasticsearch backend as well, shouldn't we also implement `MetadataProvidingFieldBridge` in the main tests, for field bridges creating multiple fields?

Also, it seems the following `FieldBridge`s in the Elasticsearch tests does not implement `MetadataProvidingFieldBridge`: `org.hibernate.search.elasticsearch.test.ElasticsearchExceptionHandlingIT.ErroneousFieldBridge`. Given its name, maybe it's intentional, though...

gsmet:
OK, it's not worth the space it takes in the PR page. Let's close it and revisit it with FieldBridge 2.0.

PR #1164, opened 2016-09-26
HSEARCH-2189 Elasticsearch Scroll API

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2189

There were features I didn't want to provide because they would be no efficient way to implement them. I implemented them anyway, because of -engine/-orm tests that were failing when ran against Elasticsearch. The details about inefficiencies are provided in the documentation.

The commits should probably be squashed together before merging. I just kept them separate in case we end up not providing those features.

### DIFF
--- a/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
+++ b/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
@@ -313,7 +313,7 @@ If your result set is bigger, you may take advantage of scrolling by using the `
 Limitations:
 
 * This method is not available in `org.hibernate.search.jpa.FullTextQuery`.
-* The Elasticsearch implementation does not allow defining an offset (i.e. calling `setFirstResult(int)` on the query before calling `scroll()`): the scrolling will always start with the first result, and using `setFirstResult(int)` will result in a `UnsupportedOperationException` being thrown.
+* The Elasticsearch implementation has poor performance when an offset has been defined (i.e. `setFirstResult(int)` has been called on the query before calling `scroll()`). This is because Elasticsearch does not provide such feature, thus Hibernate Search has to scroll through every previous result under the hood.
 * The Elasticsearch implementation allows only limited backtracking. Calling `scrollableResylts.setRowNumber(4)` when currently positioned at index `1006`, for example, may result in a `SearchException` being thrown, because only 1000 previous elements had been kept in memory. You may work this around by tweaking the property: `hibernate.search.elasticsearch.scroll_backtracking_window_size` (see <<elasticsearch-integration-configuration>>).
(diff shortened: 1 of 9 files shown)


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java
```diff
@@ -503,12 +495,108 @@ private void addScriptFields(JsonBuilder.Object query) {
...
+					builder.addSort( getSort( sortField, idFieldName ) );
+				}
+			}
+			Search search = builder.build();
 			try ( ServiceReference<JestClient> client = getExtendedSearchIntegrator().getServiceManager().requestReference( JestClient.class ) ) {
 				return client.get().executeRequest( search );
 			}
 		}
 
+		private String getScrollTimeout() {
```
Sanne: I've opened https://hibernate.atlassian.net/browse/HSEARCH-2361 as a follow up as I'd like us to parse (and validate) such parameters upfront, but it doesn't seem worth to delay integrating this PR so I'd rather open a follow up.

Targeted at 5.7 (low priority)

Sanne: P.S. the real reason - which I forgot to mention on JIRA - is that we hope to eventually be able to drop the reference to the configuration Properties instance, as it has been a source of leaks and is sometimes much larger than expected.

yrodiere: Right, I totally agree. I did it this way because I couldn't see any central place where I could provide accessors for these properties. So sure, I'll see to that later :)

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -295,6 +300,23 @@ you need to enable the Groovy plugin by adding the following snippet to your Ela
...
+* The Elasticsearch implementation allows only limited backtracking. Calling `scrollableResylts.setRowNumber(4)` when currently positioned at index `1006`, for example, may result in a `SearchException` being thrown, because only 1000 previous elements had been kept in memory. You may work this around by tweaking the property: `hibernate.search.elasticsearch.scroll_backtracking_window_size` (see <<elasticsearch-integration-configuration>>).
+* The `ScrollableResults` will become stale and unusable after a given period of time spent without fetching results from Elasticsearch. You may work this around by tweaking two properties: `hibernate.search.elasticsearch.scroll_timeout` and `hibernate.search.elasticsearch.scroll_fetch_size` (see <<elasticsearch-integration-configuration>>). Typically, you will solve timeout issues by reducing the fetch size and/or increasing the timeout limit, but this will also increase the performance hit on Elasticsearch.
+
```
Sanne: I've opened https://hibernate.atlassian.net/browse/HSEARCH-2362 to discuss potential improvements on this, but there are downsides to the ideas so I'm not sure if we'll want to do it.

Good to have the first implementation like you did, so we can see to get some feedback. :+1:

Sanne:
merged

PR #1166, opened 2016-09-27
HSEARCH-2269 Added more appropriate storage type/format for some Java 8 Time types in Elasticsearch

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2269

Note: I didn't do anything for some date/time types that have no adequate equivalent within Elasticsearch type system. Those will use the default bridges that are already used in the Lucene backend.
This includes:

 * ~~`OffsetDateTime`~~
 * ~~`OffsetTime`~~
 * ~~`ZonedDateTime`~~
 * `ZoneOffset`
 * `ZoneId`
 * `Period`
 * `Duration`

~~The offset/zoned date/time types, in particular, cannot be stored as dates because Elasticsearch converts them to UTC when storing (the offset/timezone is not stored, just the actual timestamp). So, of course, we *could* use the "_source" attribute for retrieving the offset/zone when needed, but that feels hackish to me, and is unreliable if some data is added externally with a different format, as pointed out in https://hibernate.atlassian.net/browse/HSEARCH-2358.~~

EDIT: Offset/zoned date/time types now have specific Elasticsearch bridges, following conversations in this PR.


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -571,6 +624,29 @@ private String getFieldType(FacetMetadata facetMetadata) {
...
+
+		return type;
+	}
+
+	private void addTypeOptions(JsonObject field, String type, List<String> formats) {
+		field.addProperty( "type", type );
+
+		if ( formats != null && ! formats.isEmpty() ) {
+			// TODO HSEARCH-2350 Use String.join when JDK8 becomes mandatory
+			StringBuilder builder = new StringBuilder();
```
gsmet: See `StringHelper.join`.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/FieldHelper.java
```diff
@@ -101,6 +101,76 @@ static boolean isCalendar(EntityIndexBinding indexBinding, String fieldName) {
 		return Calendar.class.isAssignableFrom( propertyClass );
 	}
 
+	static boolean isInstant(EntityIndexBinding indexBinding, String fieldName) {
+		Class<?> propertyClass = getPropertyClass( indexBinding, fieldName );
```
gsmet: Maybe we could remove some of the duplicated code?

yrodiere: It feels to me that a method returning an enum representing the actual type would make much more sense (and be more concise), especially given how those methods are used, but I didn't want to change the way it was done. I'll give it a try.

gsmet:
As for the projection stuff, we also rely on _source for `ElasticsearchDateCalendarBridgeIT#testProjectionOfCalendarValueRetrievesCorrectTimeZoneOffset`.

So, I think I would be consistent with this for the time being. Or if we decide the other way, we need to rethink what we did for Calendar.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/FieldHelper.java
gsmet: Yes, it's always difficult to decide when to change an existing pattern. I think the fact that you added a lot more instances and that we can now rely on an enum for the types is a good reason to change that. That being said, maybe you should wait for https://github.com/hibernate/hibernate-search/pull/1151/commits to be merged to change that?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
yrodiere: Fixed, thanks.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/FieldHelper.java
yrodiere: Thanks, I fixed it. And yeah, I'll probably have a hard time rebasing :)

yrodiere:
@gsmet I addressed your remarks, except the projection stuff. I guess we'll have to discuss whether https://hibernate.atlassian.net/browse/HSEARCH-2358 is invalid before we go on with this PR. I'll expose the issue on the mailing list soon.

yrodiere:
As seen on HSEARCH-2358, I'll need to implement zoned/offset date/time types bridges with an actual date storage type.

yrodiere:
And here it is. There are now Elasticsearch-specific zoned/offset date/time types bridges.

To be noted: I had to work around a JDK bug in order to make things work (JDK-8066982), which is the main reason why it took so long. See 9d4dbeb for more information.

yrodiere:
@gsmet Now all your concerns should have been addressed. Does it seem okay to you?

engine/src/main/java/org/hibernate/search/util/impl/TimeHelper.java
```diff
@@ -0,0 +1,60 @@
...
+			// No need for a workaround
+			return ZonedDateTime.from( temporal );
+		}
+		/*
+		 * Rationale
+		 *
+		 * The bug lies in the way epoch seconds are retrieved: the date is interpreted as being
+		 * expressed in the default offset for the timezone, instead of using the given offset.
+		 * ZonedDateTime.from uses these epoch seconds in priority when retrieving the local date/time,
+		 * and falls back to creating a
```
gsmet: a...

Please, please, the suspense is killing me :)

yrodiere:
@gsmet @Sanne Rebased and fixed. Thanks!

PR #1167, opened 2016-09-29
HSEARCH-2363 Test case for ambiguous string representation in ZonedDateTimeBridge

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2363


DavideD:
Can we close this one and add the test to the JIRA?

yrodiere:
@DavideD Ok, done.

PR #1168, opened 2016-09-30
HSEARCH-2260 Deal with index mappings creation/upgrade/concurrency in the Elasticsearch case

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2260

~~This PR is work in progress: it's based on top of #1166 (HSEARCH-2269), which must be merged first.~~ => OK, #1166 has been merged and this PR has been rebased.

This PR:
- Adds a VALIDATE index management strategy
- Cleans up some of the code for HSEARCH-2366. The feature was originally planned in HSEARCH-2260, so I made the work, but then we canceled it. We might as well merge it since it's only some cleanup...
- Adds some tests for the MERGE index management strategy
- Adds some documentation regarding index management strategies

I mainly added in the documentation some warnings about using `MERGE` in a production environment, and why one shouldn't.
I didn't address the concerns about concurrency when putting mappings: isn't it an Elasticsearch issue? I mean, replicas and clustering are managed internally in Elasticsearch, so I don't get why this should affect us. Plus, it's less of an issue once we warned user not to use `MERGE` in a production environment.


yrodiere:
Rebased following the recent merges to master. This PR is now ready for review.

Sanne:
@yrodiere could you rebase this please?

yrodiere:
@Sanne I'll see when I have some time, but it's probably better to merge the other ones first (HSEARCH-2396 and HSEARCH-2408) because there are some many conflicts in the index manager.

yrodiere:
@Sanne I rebased on master: ready for review! I also tried to split some commits and to merge others, in order to make them more readable. You'll probably want to review commit per commit...

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/IndexSchemaManagementStrategy.java
```diff
@@ -14,10 +14,19 @@
 public enum IndexSchemaManagementStrategy {
 
 	/**
-	 * Indexes will never be created or deleted. The index schema is not managed by Hibernate Search.
+	 * Indexes will never be created or deleted. The index schema is not managed by Hibernate Search and is never checked.
```
Sanne: I'd simply say "is not checked" rather than "is never checked". Even better "is not validated".

The "never" is confusing as it suggests that we do some kind of operation continuously in time.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/IndexSchemaManagementStrategy.java
```diff
@@ -14,10 +14,19 @@
...
 	 */
 	NONE,
 
+
+	/**
+	 * Upon session factory initialization, existing index mappings will be checked by Hibernate Search, causing an
+	 * exception if a required mapping does not exist or exists but differs in a non-compatible way (more strict type
+	 * constraints, for instance).
+	 * This strategy will never bring any change to the mappings, nor create or delete any index.
```
Sanne: same here: s/never/not/

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/logging/impl/Log.java
```diff
@@ -189,4 +190,45 @@
 	)
 	void unsupportedDynamicBoost(Class<?> boostStrategyType, Class<?> entityType, String fieldPath);
 
+	@Message(id = ES_BACKEND_MESSAGES_START_ID + 33,
+			value = "Validation failed for schema of index '%1$s'"
+	)
+	SearchException schemaValidationFailed(String indexName, @Cause Exception cause);
+
+	@Message(id = ES_BACKEND_MESSAGES_START_ID + 34,
+			value = "No mappings available from Elasticsearch for validation of index '%1$s'"
```
Sanne: When would this happen? I assume it might mean that the index was not defined; could we suggest a possible cause in the error?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/logging/impl/Log.java
```diff
@@ -189,4 +190,45 @@
 	)
 	void unsupportedDynamicBoost(Class<?> boostStrategyType, Class<?> entityType, String fieldPath);
 
+	@Message(id = ES_BACKEND_MESSAGES_START_ID + 33,
+			value = "Validation failed for schema of index '%1$s'"
+	)
+	SearchException schemaValidationFailed(String indexName, @Cause Exception cause);
```
Sanne: Are you sure this is going to explain what's wrong with the schema? I'd rather not rely on nested (cause) as people hate deep stack traces.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/schema/impl/DefaultElasticsearchSchemaValidator.java
```diff
@@ -0,0 +1,235 @@
...
+	@Override
+	public void validate(IndexMetadata expectedIndexMetadata, IndexMetadata actualIndexMetadata) {
+		// Unknown mappings are ignored, we only validate expected mappings
+		for ( Map.Entry<String, TypeMapping> entry : expectedIndexMetadata.getMappings().entrySet() ) {
+			String mappingName = entry.getKey();
+			TypeMapping expectedTypeMapping = entry.getValue();
+			TypeMapping actualTypeMapping = actualIndexMetadata.getMappings().get( mappingName );
+
+			if ( actualTypeMapping == null ) {
+				throw LOG.mappingMissing( mappingName );
```
Sanne: It is a bit strange that we abort on the first encountered mismatch.

Also, I find it strange that we throw an exception only to catch it in the client method.

Wouldn't it be nicer to collect all validation issues in one list, then throw a single clear exception with an explanation?

elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/ElasticsearchSchemaValidationIT.java
```diff
@@ -0,0 +1,429 @@
...
+		init( SimpleDateEntity.class, SimpleBooleanEntity.class );
+
+		assertSuccessfulInitialization();
+	}
+
+	private void assertSuccessfulInitialization() {
+		/*
+		 * If we manage to get a session, we assume initialization worked fine.
+		 * This check is not strictly required, since a failing initialization should
+		 * normally throw exceptions before this assertion anyway.
```
Sanne: I'd remove this code.

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -110,12 +110,30 @@ Let's see the options for the `index_schema_management_strategy` property:
...
 |`RECREATE`|Indexed will be deleted if existing and then created. This will delete all content from the index!
 |`RECREATE_DELETE`|Similarly to `RECREATE` but will also delete the index at shutdown. Commonly used for tests.
 |===============
++
+[CAUTION]
+.Strategies in production environments
+====
+It is strongly recommended to use either `NONE` or `VALIDATE` in a production environment. `CREATE` and `CREATE_DELETE` are obviously unsuitable in this context (unless you want to reindex everything upon every startup), and `MERGE` may leave your mapping half-merged in case of conflict.
+
+To be precise, if your mapping changed in an incompatible way, such as a field having its type changed, merging may be impossible. In this case, the `MERGE` strategy will prevent Hibernate Search from starting, but it may already have successfully merged another index, making a rollback difficult at best.
```
Sanne: Couldn't we validate our capability to do a merge on all indexes first, and only start the merge process if it all looks good? This option for an half-processed MERGE operation looks nasty.

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -110,12 +110,30 @@ Let's see the options for the `index_schema_management_strategy` property:
...
+
+For this reason, migrating your mapping should be considered a part of your deployment process and be planned cautiously.
+====
++
+[NOTE]
+--
+Mapping validation is as permissive as possible. Fields or mappings that are unknown to Hibernate Search will be ignored, and settings that are more powerful than required (e.g. a field annotated with `@Field(store = Store.NO)` in Search but marked as `"store": true` in Elasticsearch) will be deemed valid.
```
Sanne: Given the Store.NO example (good one!), should we clarify what will happen in that case? I haven't tested it, but I'd assume that the storage operation would happen (as ES would still do it), but we wouldn't allow a projection on this field.

yrodiere:
Jenkins, retest this please

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -267,7 +270,12 @@ private void waitForIndexCreation() {
 		Health health = new Health( healthBuilder ) {
 			@Override
 			protected String buildURI() {
-				return super.buildURI() + actualIndexName;
+				try {
+					return super.buildURI() + URLEncoder.encode(actualIndexName, AbstractAction.CHARSET);
+				}
+				catch (UnsupportedEncodingException e) {
+					throw new AssertionFailure("Unexpectedly unsupported charset", e);
+				}
```
Sanne: I'm wondering why do we do the index name encoding only in this case? Do we need a follow-up issue to URL-encode the index name only once (at initialization of the IndexManager) and then make sure we use that consistently?

Secondarily.. this code isn't respecting the formatting conventions around for whitespacing. Wondering how you managed to fool checkstyle :)

Sanne:
ah, right. Thanks for the explanation, I forgot it's a "custom" command.

Sanne:
merged. Great work!

PR #1169, opened 2016-09-30
HSEARCH-2355 Document default timezone in ES query

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2355

### DIFF
--- a/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
+++ b/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
@@ -299,6 +299,14 @@ List<?> result = session.createFullTextQuery(query, GolfPlayer.class).list();
 ----
 ====
 
+[CAUTION]
+.Date/time in native Elasticsearch queries
+====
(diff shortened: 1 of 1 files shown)


documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -299,6 +299,14 @@ List<?> result = session.createFullTextQuery(query, GolfPlayer.class).list();
 ----
 ====
 
+[CAUTION]
+.Date/time in native Elasticsearch queries
+====
+Elasticsearch, by default, understands date/time strings without time zone as UTC. If overlooked, this can cause your native Elasticsearch queries to be completely off.
```
Sanne: s/understands/treats/ ? Or "interprets" ?

Also the section "treats date/time strings without time zone as UTC" is a bit ambiguous on the "without", it would be clearer as:

> By default Elasticsearch interprets the date/time strings lacking the time zone as if they were represented using the UTC timezone.

yrodiere: Fixed and rebased. Thanks!

Sanne:
merged, thanks!

PR #1170, opened 2016-09-30
HSEARCH-2367 LuceneQueryDescriptor ignores targetEntities

gustavonalle:
https://hibernate.atlassian.net/browse/HSEARCH-2367

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/impl/ImmutableSearchFactory.java
+++ b/engine/src/main/java/org/hibernate/search/engine/impl/ImmutableSearchFactory.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.search.engine.impl;
 
+import java.util.Arrays;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Properties;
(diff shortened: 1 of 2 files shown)


yrodiere:
@gustavonalle Thanks for your report and PR. I answered on the JIRA ticket with a possible solution to your issue, and I'm going to investigate to try to fix this and other issues at the same time.

Sanne:
N.B. `org.hibernate.search.elasticsearch.impl.ElasticsearchJsonQueryDescriptor` also ignores the types. We can discuss improving the approach but let's keep the two consistent?

yrodiere:
I opened PR containing an alternative solution: #1172
That one does address the `ElasticsearchJsonQueryDescriptor` issue too, but requires some SPI changes.
I'm keeping this PR opened for now, since I'm not sure #1172 will make it to master.

yrodiere:
Fixed in #1172 ; closing this PR. Thanks for pointing out the issue!

gustavonalle:
cool, thanks for the quick fix!

PR #1171, opened 2016-09-30
HSEARCH-2368 Avoid failing query if the index does not exist yet

gustavonalle:
https://hibernate.atlassian.net/browse/HSEARCH-2368


yrodiere:
Closing this as the issue reported in HSEARCH-2368 seems to stem from a configuration mistake.

PR #1172, opened 2016-10-03
HSEARCH-2367 (+ others) QueryDescriptor API/SPI improvements

yrodiere:
This is an attempt to fix:

 * [HSEARCH-2367 QueryDescriptors ignore targetEntities](https://hibernate.atlassian.net/browse/HSEARCH-2367) (first and foremost)
 * [HSEARCH-2372 QueryDescriptor creates an implementation leak](https://hibernate.atlassian.net/browse/HSEARCH-2372) (a related, easy-to-fix issue since we're at it)
 * [HSEARCH-2370 Document how to use QueryDescriptor](https://hibernate.atlassian.net/browse/HSEARCH-2370) (a related, easy-to-fix issue since we're at it)

This also happens to fix [HSEARCH-2371 Using a SearchIntegrator-provided QueryDescriptor will require passing the targeted entities twice](https://hibernate.atlassian.net/browse/HSEARCH-2371)

On the bright side, this fixes the issues with only a few changes. Most additions/deletions in this PR are not strictly required, but were added to ease up test maintenance in 59e14ea .

On the not-so-bright side, I had to change the `SearchIntegrator` interface and to add a method to `HSQuery`. **But** I think it's okay for the following reasons:

 * The changes to `SearchIntegrator` only affect methods added in version 5.6, which is still in beta.
 * The new method in `HSQuery` is fairly simple (`getQueryString`), and I don't think we provide a way for service providers to provide their own `HSQuery`. This contract is about a service we provide to service provider, more than about a service that service providers provide to us. So adding a requirement mainly affects us.

I checked the Infinispan integration and didn't find an implementation of `HSQuery`, so I think we're fine on this side. Plus, HSEARCH-2367 was opened by an Infinispan maintainer trying to port it to the next version of Hibernate Search, so I think they at least might understand.

To reviewers: I'd recommend reviewing the commits individually, since 59e14ea , the commit with the most additions/deletions, also happen to be the least relevant to this PR (mostly boilerplate code).

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/impl/LuceneQueryDescriptor.java
+++ b/engine/src/main/java/org/hibernate/search/engine/impl/LuceneQueryDescriptor.java
@@ -7,9 +7,9 @@
 package org.hibernate.search.engine.impl;
 
 import org.apache.lucene.search.Query;
-import org.hibernate.search.engine.integration.impl.ExtendedSearchIntegrator;
 import org.hibernate.search.query.engine.spi.HSQuery;
 import org.hibernate.search.query.engine.spi.QueryDescriptor;
+import org.hibernate.search.spi.SearchIntegrator;
 
 /**
  * A {@link QueryDescriptor} for a Lucene query.
@@ -25,7 +25,7 @@ public class LuceneQueryDescriptor implements QueryDescriptor {
 	}
 
 	@Override
-	public HSQuery createHSQuery(ExtendedSearchIntegrator extendedIntegrator) {
+	public HSQuery createHSQuery(SearchIntegrator extendedIntegrator) {
 		HSQuery hsQuery = extendedIntegrator.createHSQuery();
 		hsQuery.luceneQuery( luceneQuery );
 
--- a/engine/src/main/java/org/hibernate/search/query/engine/spi/QueryDescriptor.java
+++ b/engine/src/main/java/org/hibernate/search/query/engine/spi/QueryDescriptor.java
@@ -6,7 +6,7 @@
  */
 package org.hibernate.search.query.engine.spi;
 
-import org.hibernate.search.engine.integration.impl.ExtendedSearchIntegrator;
+import org.hibernate.search.spi.SearchIntegrator;
 
 /**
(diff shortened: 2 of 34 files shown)


engine/src/main/java/org/hibernate/search/engine/impl/LuceneQueryDescriptor.java
```diff
@@ -25,7 +25,7 @@ public LuceneQueryDescriptor(Query luceneQuery) {
 	}
 
 	@Override
-	public HSQuery createHSQuery(ExtendedSearchIntegrator extendedIntegrator) {
+	public HSQuery createHSQuery(SearchIntegrator extendedIntegrator) {
```
gsmet: You forgot to rename the variable here (all the other spots are OK).

Sanne:
Looks good, I'd go even further with this line of though ant deprecate the old parameter-less createHSQuery:
 - https://github.com/hibernate/hibernate-search/commit/1dc17ee26dc9e660b3f835a7df3f55d966af0c30

Ok to apply?

engine/src/main/java/org/hibernate/search/query/engine/spi/QueryDescriptor.java
```diff
@@ -11,6 +11,10 @@
 /**
  * An index query which is able to create a corresponding {@link HSQuery} object.
  *
+ * <p><strong>Note:</strong>In the Hibernate-Search-ORM module, you may pass query descriptors to
```
gsmet: The capitalization of the module name is weird. AFAICT, the module is named orm.

gsmet:
@Sanne I added a few proofreading comments. Nothing serious but it would be nice to fix them before merging.

Sanne:
thanks @gsmet ! one is already changed ;)

yrodiere:
@Sanne @gsmet Thanks!
@Sanne Your addition is fine to me! Want me to fix the issues spotted by @gsmet, or are you currently doing it?

Sanne:
@yrodiere already done, I was merely waiting for your permission to push ;)

Sanne:
applied! thanks

PR #1173, opened 2016-10-03
HSEARCH-2373 Some ES-related services are not properly released

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2373

This seems to fix the most obvious issues with "un-released" services (those one can see in any Elasticsearch test).


Sanne:
seems solved! thanks

PR #1174, opened 2016-10-04
HSEARCH-2374 Make the default value of hibernate.search.default.elasticsearch.index_schema_management_strategy less confusing	

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2374

After a bit of thought, I think the `CREATE` strategy added in this PR would probably be a better default than `NONE`. As @gsmet pointed out, users would probably expect a similar behavior irrespective of the backend they use, and the `CREATE` strategy is similar to the behavior when using a local Lucene backend. And it does not harm if you overlook the strategy setting, since this `CREATE` strategy will never alter existing indexes: worst case, you got new, unwanted indexes.

The pre-existing strategies were renamed in order to avoid confusions. It's not strictly necessary, but I think it's better than naming the new strategy `CREATE_ONLY` to distinguish from the pre-existing `CREATE` that also drops the schema...
The breaking of existing configuration should not be bery bad, since the `(RE)CREATE` and `(RE)CREATE_DELETE` can hardly be considered in a production environment. So warning users when we announce the release should do the trick...

### DIFF
--- a/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
+++ b/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
@@ -95,7 +95,7 @@ Add them to your `persistence.xml` or where you put the rest of your Hibernate S
 
 Select Elasticsearch as the backend:: `hibernate.search.default.indexmanager elasticsearch`
 Hostname and port for Elasticsearch:: `hibernate.search.default.elasticsearch.host \http://127.0.0.1:9200`
-Selects the index creation strategy:: `hibernate.search.default.elasticsearch.index_schema_management_strategy CREATE_DELETE`
+Selects the index creation strategy:: `hibernate.search.default.elasticsearch.index_schema_management_strategy RECREATE_DELETE`
 Maximum time to wait for the indexes to become available before failing (in ms):: `hibernate.search.default.elasticsearch.index_management_wait_timeout 10000`
 Status an index must at least have in order for Hibernate Search to work with it (one of "green", "yellow" or "red"):: `hibernate.search.default.elasticsearch.required_index_status green`
 Whether to perform an explicit refresh after a set of operations has been executed against a specific index (`true` or `false`):: `hibernate.search.default.elasticsearch.refresh_after_write false`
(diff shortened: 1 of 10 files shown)


Sanne:
As discussed on Hipchat: the CREATE case needs to deal with a concurrent CREATE which might have been triggered by another Hibernate Search application node attempting the same in parallel.

ES should return an error to one of the two (or more) nodes attempting the creation, and this is fine.

yrodiere:
@Sanne the later commit should do the trick.

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -95,7 +95,7 @@ Add them to your `persistence.xml` or where you put the rest of your Hibernate S
 
 Select Elasticsearch as the backend:: `hibernate.search.default.indexmanager elasticsearch`
 Hostname and port for Elasticsearch:: `hibernate.search.default.elasticsearch.host \http://127.0.0.1:9200`
-Selects the index creation strategy:: `hibernate.search.default.elasticsearch.index_schema_management_strategy CREATE_DELETE`
+Selects the index creation strategy:: `hibernate.search.default.elasticsearch.index_schema_management_strategy RECREATE_DELETE`
```
gsmet: I don't think it's a good idea to advertise `RECREATE_DELETE` in this example: it seems to be a bit risky - especially because other values looks sane. I would stick to the default for the example.

yrodiere: Sure. I just translated the old value, but I can change this.

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -109,10 +109,11 @@ Let's see the options for the `index_schema_management_strategy` property:
...
 |Value|Definition
-|NONE|Indexes will not be created nor deleted.
-|MERGE|Missing indexes and mappings will be created, mappings will be updated if there are no conflicts.
-|CREATE|Indexed will be deleted if existing and then created. This will delete all content from the index!
-|CREATE_DELETE|Similarly to 'CREATE' but will also delete the index at shutdown. Commonly used for tests.
+|`NONE`|Indexes will not be created nor deleted.
+|`MERGE`|Missing indexes and mappings will be created, mappings will be updated if there are no conflicts.
+|`CREATE`|**The default**: existing indexes will not be altered, missing indexes will be created.
+|`RECREATE`|Indexed will be deleted if existing and then created. This will delete all content from the index!
+|`RECREATE_DELETE`|Similarly to `RECREATE` but will also delete the index at shutdown. Commonly used for tests.
```
yrodiere: I think Hibernate's wording is prior to the JPA wording, and they're not consistent.
If I had to stick to an existing wording, I'd rather choose JPA's, which is a standard and is much more explicit ("create", "drop-and-create", etc.).

I can switch to the JPA wording if you wish, but I fear that the one we have now will harm some careless users, which is why I'd really like those terms to change before the release.

yrodiere: Note: see org.hibernate.tool.schema.Action for a list of `hbm2ddl.auto` values and their JPA equivalent.

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
yrodiere: @gsmet I just changed the example to use `CREATE`.

Sanne:
nice! merged

PR #1175, opened 2016-10-04
HSEARCH-2092 Map the optimize() operation to Elasticsearch 'force merge' requests

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2092

### DIFF
--- a/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
+++ b/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
@@ -464,6 +464,13 @@ ftQuery.enableFullTextFilter( "namedDriver" ).setParameter( "name", "liz" );
 
 For static filters, you can simply extend the `SimpleElasticsearchFilter` and provide an Elasticsearch filter in JSON form.
 
+=== Index optimization
+
+The optimization features documented in <<search-optimize>> are only partially implemented. That kind of optimization is rarely needed with recent versions of Lucene (on which Elasticsearch is based), but some of it is still provided for the very specific case of indexes meant to stay read-only for a long period of time:
+
+ * The automatic optimization is not implemented and most probably never will be.
+ * The manual optimization (`searchFactory.optimize()`) is implemented.
+
 === Limitations
 
 Not everything is implemented yet.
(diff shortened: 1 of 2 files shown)


documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -464,6 +464,13 @@ ftQuery.enableFullTextFilter( "namedDriver" ).setParameter( "name", "liz" );
 
 For static filters, you can simply extend the `SimpleElasticsearchFilter` and provide an Elasticsearch filter in JSON form.
 
+=== Index optimization
+
+The optimization features documented in <<search-optimize>> are only partially implemented. That kind of optimization is rarely needed with recent versions of Lucene (on which Elasticsearch is based), but some of it is still provided for the very specific case of indexes meant to stay read-only for a long period of time:
+
+ * The automatic optimization is not implemented and most probably never will be.
+ * The manual optimization (`searchFactory.optimize()`) is implemented.
+
```
Sanne: Nice, well said.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
```diff
@@ -101,9 +99,17 @@ public ElasticsearchIndexWorkVisitor(String indexName, boolean refreshAfterWrite
...
 	@Override
 	public BackendRequest<?> visitOptimizeWork(OptimizeLuceneWork work, Void p) {
-		// TODO HSEARCH-2092 implement
-		LOG.warn( "Optimize work is not yet supported for Elasticsearch, ignoring it" );
-		return null;
+		/*
+		 * As of ES 2.1, the Optimize API has been renamed to ForceMerge,
+		 * but Jest still does not provide commands for the ForceMerge API as of
+		 * version 2.0.3
+		 * See https://github.com/searchbox-io/Jest/issues/292
```
Sanne: Do you know if the deprecated "Optimize" command is applies as a ForceMerge on a newer version of ES ?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
```diff
@@ -101,9 +99,17 @@ public ElasticsearchIndexWorkVisitor(String indexName, boolean refreshAfterWrite
...
+		 */
+		Optimize optimize = new Optimize.Builder()
+				.addIndex( indexName )
+				.build();
+
+		return new BackendRequest<JestResult>( optimize, work, indexName, refreshAfterWrite );
```
Sanne: That seems simple enough :-) :+1:

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
yrodiere: Currently no, Jest really doesn't even know about the newer API.

But to Elasticsearch, it seems both APIs are the same: [the docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-optimize.html) say the Optimize API has been "renamed", and the descriptions are really close (almost copy/paste).

PR #1177, opened 2016-10-05
HSEARCH-2376 Analyze SortedDocValuesField values

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2376

To sum up, `SortedDocValuesField` are not analyzed automatically by lucene as other `Field`s are, so we have to do it manually.

Note: we don't do anything for remote analyzers, because the Elasticsearch module doesn't use docvalues and thus is unaffected by this bug. If one day it does, it will be affected, because fields with docvalues in ES are not analyzed. And then we're gonna have a hard time analyzing values with an analyzer that's not even here...

I'll submit PR for backports once this one is approved.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
+++ b/engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.search.engine.spi;
 
+import java.io.IOException;
 import java.io.Serializable;
 import java.util.Calendar;
 import java.util.Collection;
@@ -13,10 +14,12 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.document.DateTools;
 import org.apache.lucene.document.Document;
 import org.apache.lucene.document.DoubleDocValuesField;
@@ -34,6 +37,8 @@ import org.hibernate.annotations.common.reflection.ReflectionManager;
 import org.hibernate.annotations.common.reflection.XClass;
 import org.hibernate.annotations.common.reflection.XMember;
 import org.hibernate.search.analyzer.Discriminator;
+import org.hibernate.search.analyzer.impl.AnalyzerReference;
+import org.hibernate.search.analyzer.impl.LuceneAnalyzerReference;
 import org.hibernate.search.annotations.ProvidedId;
 import org.hibernate.search.annotations.Store;
 import org.hibernate.search.backend.AddLuceneWork;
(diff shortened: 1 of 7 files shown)


engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
```diff
@@ -803,12 +809,46 @@ else if ( numericValue instanceof Float ) {
 					}
 				}
 				else {
-					document.add( new SortedDocValuesField( sortField.getFieldName(), new BytesRef( field.stringValue() ) ) );
+					String value = field.stringValue();
+					AnalyzerReference analyzerReference = fieldMetaData.getAnalyzerReference();
+					if ( fieldMetaData.getIndex().isAnalyzed() && analyzerReference.is( LuceneAnalyzerReference.class ) ) {
+						// Analysis is skipped altogether when the analyzer is remote. It's up to the backend to handle it.
+						Analyzer analyzer = analyzerReference.unwrap( LuceneAnalyzerReference.class ).getAnalyzer();
+						value = analyzeSortFieldValue( sortField.getFieldName(), field.stringValue(), analyzer );
```
gsmet: s/field.stringValue()/value/

engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
```diff
@@ -803,12 +809,46 @@ else if ( numericValue instanceof Float ) {
...
+			else {
+				return null;
+			}
+		}
+		catch (SearchException | IOException e) {
+			throw log.couldNotAnalyzeSortableField( fieldName, e );
 		}
 	}
 
+
```
gsmet: Unnecessary new line.

engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
```diff
@@ -803,12 +809,46 @@ else if ( numericValue instanceof Float ) {
...
+					if ( value != null ) {
+						document.add( new SortedDocValuesField( sortField.getFieldName(), new BytesRef( value ) ) );
+					}
+				}
+			}
+		}
+	}
+
+	/*
+	 * Necessary because Lucene doesn't automatically analyze SortedDocValuesFields.
```
gsmet: I don't think it's the right wording as we don't even provide the analyzer to the `SortedDocValuesField`s so we can't expect Lucene to deal with this.

engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
```diff
@@ -803,12 +809,46 @@ else if ( numericValue instanceof Float ) {
...
+	 * See https://hibernate.atlassian.net/browse/HSEARCH-2376
+	 */
+	private String analyzeSortFieldValue(String fieldName, String fieldValue, Analyzer analyzer) {
+		try {
+			Iterator<String> analyzedTokens = AnalyzerUtils.tokenizedTermValues( analyzer, fieldName, fieldValue ).iterator();
+
+			if ( analyzedTokens.hasNext() ) {
```
gsmet: I wouldn't have created an iterator and would have worked on the list here but I suppose it's a matter of taste :).

engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
yrodiere: I fixed the variable reference issue and used a new util to implement the optimization suggested by @Sanne.

Sanne:
merged and backported. Great!

PR #1181, opened 2016-10-06
HSEARCH-2375 Document the supported Elasticsearch versions

yrodiere:
... in the reference documentation

https://hibernate.atlassian.net/browse/HSEARCH-2375


Sanne:
merged

PR #1182, opened 2016-10-06
HSEARCH-2314 TokenStream contract violation during serialization

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2314


Sanne:
nice, merged!

PR #1183, opened 2016-10-07
HSEARCH-2271 Enable ArrayBridgeTest (and similar) for Elasticsearch

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2271

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
@@ -66,6 +66,14 @@ import io.searchbox.indices.Optimize;
  */
 class ElasticsearchIndexWorkVisitor implements IndexWorkVisitor<Void, BackendRequest<?>> {
 
+	/*
+	 * Nesting marker field name: fields with this name have the parent path as their value,
+	 * and are used for denoting the parent element of the subsequent fields
+	 *
+	 * Use a very specific String to avoid naming conflict with user-defined field names
+	 */
+	public static final String NESTING_MARKER = "__HSearch_NestingMarker";
+
 	private static final Pattern DOT = Pattern.compile( "\\." );
 	private static final Pattern NAME_AND_INDEX = Pattern.compile( "(.+?)(\\[([0-9]+)\\])?" );
 
@@ -200,8 +208,7 @@ class ElasticsearchIndexWorkVisitor implements IndexWorkVisitor<Void, BackendReq
 
 		String parentPath = null;
 		for ( IndexableField field : document.getFields() ) {
-			// marker field for denoting the parent element of the subsequent fields
-			if ( "$nesting".equals( field.name() ) ) {
+			if ( NESTING_MARKER.equals( field.name() ) ) {
 				parentPath = field.stringValue();
 				continue;
 			}
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
@@ -39,9 +39,11 @@ import org.hibernate.search.engine.ProjectionConstants;
 import org.hibernate.search.engine.integration.impl.ExtendedSearchIntegrator;
 import org.hibernate.search.engine.metadata.impl.DocumentFieldMetadata;
 import org.hibernate.search.engine.metadata.impl.EmbeddedTypeMetadata;
+import org.hibernate.search.engine.metadata.impl.EmbeddedTypeMetadata.Container;
 import org.hibernate.search.engine.metadata.impl.TypeMetadata;
 import org.hibernate.search.engine.spi.DocumentBuilderIndexedEntity;
 import org.hibernate.search.engine.spi.EntityIndexBinding;
+import org.hibernate.search.exception.AssertionFailure;
 import org.hibernate.search.spatial.impl.SpatialHelper;
 
 import com.google.gson.JsonArray;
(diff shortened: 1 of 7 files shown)


gsmet:
@yoann could you rebase this one now that we merged HSEARCH-2390?

Thanks!

yrodiere:
@gsmet I already did, actually. Strangely, Github is showing commits e019aef and c7cc705 as being part of this PR, but these commits are already on master...

yrodiere:
@gsmet I amended my last commit (with no change, just to change the dates and hash) and force-pushed. It seems to have solved the issue in GitHub's interface.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
```diff
@@ -66,6 +68,25 @@
  */
 class ElasticsearchIndexWorkVisitor implements IndexWorkVisitor<Void, BackendRequest<?>> {
 
+	/*
+	 * Nesting marker field name: fields with this name have the parent path as their value,
+	 * and are used for denoting the parent element of the subsequent fields
+	 *
+	 * Use a very specific String to avoid naming conflict with user-defined field names
+	 */
+	public static final String NESTING_MARKER = "__Hibernate_nesting_marker";
```
gsmet: Maybe we should follow the convention used in `ProjectionConstants`? Something like `__HSearch_NestingMarker` and `__HSearch_ThisMarker` - not sure about the camel casing, depends on what you want to convey? I know `ProjectionConstants` is not entirely consistent but it would follow the rule of the majority.

gsmet:
In f591e867b70e5c99790b7b3837a8c406e482a4dc commit message, s/BuiltinArayBridge/BuiltinArrayBridge/.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
```diff
@@ -66,6 +68,25 @@
...
+
+	/*
+	 * "this" marker field name: fields with this name are used for denoting that the
+	 * subsequent field represents an embedded as a whole, as opposed to simply a field
+	 * in an embedded.
+	 * The marker is necessary because in some cases, the embedded and its elements
+	 * share a common field name. For instance it is the case of embedded collections.
```
gsmet: s/is the case of/is the case for/?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
```diff
@@ -66,6 +68,25 @@
...
+	 *
+	 * Use a very specific String to avoid naming conflict with user-defined field names
+	 */
+	public static final String THIS_MARKER = "__Hibernate_this_marker";
```
gsmet: What is the rationale behind the choice of "this" for this marker? I don't find the name very explicit.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
yrodiere: Ok, done.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
yrodiere: Right. Fixed, thanks

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
yrodiere: This marker is useful when `@IndexedEmbedded` collections are indexed.

For instance:

```java
@Indexed
public class MyEntity {
@IndexedEmbedded(indexNullAs = "NULL")
private List<String> myCollection;
}
```

When `myCollection` contains `["foo", "bar"]`, the following fields will appear in the Lucene document:

```
myCollection:foo
myCollection:bar
```

When `myCollection` is `null`, the following field will appear:
```
myCollection:NULL
```

When you're in the second case you cannot know only from the Lucene document whether the field represents a collection element (say, `myCollection[1]` in pseudo-code) or the collection itself (still in pseudo-code, `myCollection`, or `myCollection.this` to be more explicit).

"This" or "self" are the only words that came to my mind and seemed to fit the meaning of this marker: it states that the field value is not the value of a collection element, but really the value of the embedded itself.

Be aware, though, that in order to fix HSEARCH-2397 I'm already refactoring this: just using a string for the nesting path is not enough, the marker has to convey more information. So it might not be worth a long discussion...

yrodiere:
@gmset Thanks! I fixed the issues and answered your question.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
gsmet: I think "object value" would be better (I don't know if you kept the marker in the follow-up of this work).

yrodiere: ObjectValue it is, then. I changed that and force-pushed.

Yes, the marker is dropped as part of HSEARCH-2397/#1188 (we don't need it anymore because the nesting marker carries more information).

gsmet:
Merged! Thanks!

PR #1185, opened 2016-10-11
HSEARCH-2390 Investigate and label test exclusions in the ES module

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2390

Just a PR to reorganize test exclusions and match them with the tickets I just created. Most of those tickets identify the actual issue.
I also enabled some tests that happened to pass anyway.

The patch of the pom file is a bit unreadable, sorry about that... But I guess there's no other way.

### DIFF
--- a/engine/src/test/java/org/hibernate/search/test/backend/lucene/ScheduledCommitPolicyTest.java
+++ b/engine/src/test/java/org/hibernate/search/test/backend/lucene/ScheduledCommitPolicyTest.java
@@ -15,12 +15,14 @@ import org.hibernate.search.backend.spi.WorkType;
 import org.hibernate.search.engine.integration.impl.ExtendedSearchIntegrator;
 import org.hibernate.search.query.engine.spi.HSQuery;
 import org.hibernate.search.testsupport.junit.SearchFactoryHolder;
+import org.hibernate.search.testsupport.junit.SkipOnElasticsearch;
 import org.hibernate.search.testsupport.setup.CountingErrorHandler;
 import org.hibernate.search.testsupport.setup.TransactionContextForTest;
 import org.jboss.byteman.contrib.bmunit.BMRule;
 import org.jboss.byteman.contrib.bmunit.BMUnitRunner;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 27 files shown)


yrodiere:
Jenkins, retest this please

orm/src/test/java/org/hibernate/search/test/query/validation/QueryValidationTest.java
```diff
@@ -42,7 +42,9 @@ public void setUp() throws Exception {
 		super.setUp();
 
 		Transaction tx = openSession().beginTransaction();
-		getSession().save( new A() );
+		A a = new A();
```
gsmet: Why do you need to have a value for Elasticsearch?

engine/src/test/java/org/hibernate/search/test/backend/lucene/ScheduledCommitPolicyTest.java
```diff
@@ -35,6 +37,7 @@
  * @author gustavonalle
  */
 @RunWith(BMUnitRunner.class)
+@Category(SkipOnElasticsearch.class)
```
gsmet: We usually add a comment for this. Just a comment stating that this test is specific to Lucene would do.

gsmet:
Not directly related to this PR but be aware that a few tests are skipped using `ElasticsearchSupportInProgress` and they probably should be reevaluated too.

yrodiere:
Jenkins, retest this please

orm/src/test/java/org/hibernate/search/test/query/validation/QueryValidationTest.java
yrodiere: Well spotted; I shouldn't have committed this, I'll revert that.

yrodiere:
@gsmet Thanks for your feedback. I just force-pushed, your concerns should have been addressed. I also added comments to the various uses `ElasticsearchSupportInProgress` mentioning the JIRA ticket.

gsmet:
Looks good, merging.

gsmet:
Merged! Thanks!

PR #1187, opened 2016-10-12
Remove some "Experimental" flags

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2313
 - https://hibernate.atlassian.net/browse/HSEARCH-2407


gsmet:
@Sanne could we wait for the Elasticsearch part to be implemented for MoreLikeThis queries before removing the experimental flag? Or do we consider that we can break it in 6 anyway?

Sanne:
@gsmet we can change it in 6 anyway.

yrodiere:
Merged, thanks!

I also created [HSEARCH-2411](https://hibernate.atlassian.net/browse/HSEARCH-2411) to consider dropping other flags, at the very least when we release 6.0.

PR #1188, opened 2016-10-13
HSEARCH-2397 and HSEARCH-2409: fixes around @IndexedEmbeddeds

yrodiere:
~~Must wait for HSEARCH-2271 (#1183) to be merged first.~~ => Done

Content of this PR:

 * https://hibernate.atlassian.net/browse/HSEARCH-2409
 * https://hibernate.atlassian.net/browse/HSEARCH-2397

### DIFF
--- a/elasticsearch/pom.xml
+++ b/elasticsearch/pom.xml
@@ -270,10 +270,6 @@
                                 <exclude>**/CollectionInitializeTest.java</exclude>
                                 <exclude>**/EmbeddedSortableIdFieldTest.java</exclude>
                                 
-                                <!-- HSEARCH-2397 The Elasticsearch backend doesn't support embeddeds prefixes properly -->
-                                <exclude>**/DslEmbeddedSearchTest.java</exclude>
-                                <exclude>**/PrefixedEmbeddedCaseInPathTest.java</exclude>
-                                
                                 <!-- HSEARCH-2398 Improve field name/type validation when querying the Elasticsearch backend -->
                                 <exclude>**/FacetUnknownFieldFailureTest.java</exclude><!-- The elasticsearch backend seems too lenient for unknown field names. -->
                                 <exclude>**/QueryValidationTest.java</exclude><!-- testTargetNumericEncodedFieldWithStringQueryThrowsException: The Elasticsearch backend throws an exception, but not exactly the one we expect -->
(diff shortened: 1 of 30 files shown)


gsmet:
Just a quick comment on this one: whatever the representation of embeddeds is in the index, we need to also make sure that we can sort on them (sorting on a nested path requires an extra parameter). I don't know if you took care of this (and it might be better to tacke this once we agree on the representation) but I wanted to mention it.

yrodiere:
@gsmet I don't know what extra parameter you're talking about, so I guess I didn't take care of this.
Are you talking about the "nested" datatype in Elasticsearch? Because I didn't tackle this issue at all.

I simply made sure that when we index fields, we use the field path (concatenated `@IndexedEmbedded` prefixes) instead of using the property path (Java properties) like we used to, which failed miserably as soon as an `@IndexedEmbedded` had a custom prefix. So in fact, I just made sure that the JSON we index to ES for indexing has a structure that matches mapping we defined.

gsmet:
I might be wrong but reading the commit comments, I understood that you might have used nested fields to deal with embedded.

It's probably the way to go but it has consequences. This was an open issue in our current implementation, we haven't taken good care of it (see `ToElasticSearch#isNested` for instance).

See https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-sort.html#nested-sorting for information on nested sorting.

yrodiere:
Thanks for the piece of information. From what I read, this applies to the nested datatype ("In the below example offer is a field of type nested").

One thing to note is that when indexing, you can pass arrays for any field (including `object` fields), but they get flattened internally. So you can nest without using the `nested` datatype. The `nested` datatype brings more subtle behavior, which (I guess) is why special care should be taken when sorting.

Currently every `IndexedEmbedded` path component is mapped to the `object` datatype: see `org.hibernate.search.elasticsearch.impl.ElasticsearchIndexManager.getOrCreateProperties(JsonObject, String)`, the JsonObject have no `type` properties, and the default is `object`. I did not change anything to the mapping, I just made sure we produce correct JSON documents when indexing. 

From what I can see, https://hibernate.atlassian.net/browse/HSEARCH-2263 relates more closely to the issue you're mentioning. But I agree we'll have to address that one, too.

yrodiere:
Adding the "Needs work" label as I just noticed there is some incorrect magic in `org.hibernate.search.elasticsearch.impl.FieldHelper.getPropertyMetadata(EntityIndexBinding, String)` that I did not address. I will add tests demonstrating the issues and try to solve them.

gsmet:
@yrodiere you should be able to rebase this one!

yrodiere:
@gsmet Thanks, I rebased on master.

Sanne:
merged, thanks!

PR #1191, opened 2016-10-17
HSEARCH-2401 Boosts are (erroneously) not applied to IndexedEmbeddeds

yrodiere:
~~This PR is based on HSEARCH-2397, we'll have to review and merge the PR ( #1188) first.~~ => #1188 has been merged and I rebased this PR.

Related JIRA ticket: https://hibernate.atlassian.net/browse/HSEARCH-2401

PR #1192, opened 2016-10-19
HSEARCH-2396 Support custom field bridges with the Elasticsearch backend (for EmbeddedIds for instance)

yrodiere:
~~This PR is based on HSEARCH-2401, we'll have to review and merge the PR ( #1191) first.~~ => Done and rebased.

Related JIRA ticket: https://hibernate.atlassian.net/browse/HSEARCH-2396

The purpose of this PR is to make the following tests work when ran against Elasticsearch:

```
EmbeddedIdTest.java
EmbeddedIdWithDocumentIdTest.java
ProgrammaticEmbeddedItTest.java
CollectionInitializeTest.java
EmbeddedSortableIdFieldTest.java
```

There were multiple issues with these tests:
1. We didn't support creating nested field from custom field bridges, so the field bridges used in the tests failed because they created fields such as `id.lastName` or `id.firstName`
2. We didn't include the IDs in the indexed documents, so searches on the ID used to fail (the field was missing)


gsmet:
> We didn't include the IDs in the indexed documents, so searches on the ID used to fail (the field was missing)

@yrodiere we might want to revisit https://hibernate.atlassian.net/browse/HSEARCH-2253 (see https://github.com/hibernate/hibernate-search/blob/master/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java#L316 and https://github.com/hibernate/hibernate-search/blob/master/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java#L453 ). In the documentation, we recommend to not sort on the id field (especially for the Lucene backend where it leads to erratic behaviors) but I think it will work with Es and the new mapping.

yrodiere:
Right, thanks for spotting this. I self-assigned and added this ticket to
the current sprint.

Yoann Rodière <yoann@hibernate.org>
Hibernate NoORM Team

gsmet:
Jenkins, retest this please

yrodiere:
@gsmet Yeah, I seem to have confused the CI with my rebases again... It's only building the second-but-last commit :/
For what it's worth, it passes on Travis: https://travis-ci.org/yrodiere/hibernate-search/builds/169196158

yrodiere:
Jenkins, retest this please... ?

Sanne:
merged, just added a single character to fix a javadoc typo.

PR #1193, opened 2016-10-19
HSEARCH-2394 Fix dependency issues for tests that are re-used in Elasearch

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2394


gsmet:
Merged, thanks!

PR #1194, opened 2016-10-19
HSEARCH-2403 TwoWayFieldBridge exceptions are not wrapped properly when retrieving ES projection values

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2403


gsmet:
Merged, thanks!

PR #1196, opened 2016-10-24
New "async" IndexReader strategy, with performance tests

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2427
 - https://hibernate.atlassian.net/browse/HSEARCH-402

These are my results:
```

Benchmark                                                  (directorytype)  (indexSize)  (maxResults)  (readerStrategy)   Mode  Cnt     Score     Error  Units
JMHBenchmarks.concurrentReadWriteTest                                   fs      5000000            10             async  thrpt   10  2021.166 ± 203.769  ops/s
JMHBenchmarks.concurrentReadWriteTest:readWriteTestReader               fs      5000000            10             async  thrpt   10   259.620 ±  16.380  ops/s
JMHBenchmarks.concurrentReadWriteTest:readWriteTestWriter               fs      5000000            10             async  thrpt   10  1761.546 ± 206.186  ops/s
JMHBenchmarks.concurrentReadWriteTest                                   fs      5000000            10            shared  thrpt   10  1804.948 ± 132.665  ops/s
JMHBenchmarks.concurrentReadWriteTest:readWriteTestReader               fs      5000000            10            shared  thrpt   10   159.672 ±  13.547  ops/s
JMHBenchmarks.concurrentReadWriteTest:readWriteTestWriter               fs      5000000            10            shared  thrpt   10  1645.276 ± 139.887  ops/s
```

The benchmark runs queries and index writes concurrently; the query is very simple so that it doesn't dominate the CPU usage; the high amount of concurrent writes will stress the `ReaderProvider`: as you can see in this scenario the new "async" reader strategy gives a good 60% performance boost.

I actually expected more, but at least it proofs the point; I guess that with higher latency hard drives or larger indexes the benefit might get stronger.


integrationtest/engine-performance/pom.xml
```diff
@@ -47,6 +47,57 @@
...
+                                 </bundledSignatures>
+                             </configuration>
+                             <phase>test-compile</phase>
+                             <goals>
+                                 <goal>testCheck</goal>
+                             </goals>
+                        </execution>
+                    </executions>
+                </plugin>
+            <plugin>
```
Sanne: nice trick, thanks I'll try that.

integrationtest/engine-performance/src/main/java/org/hibernate/search/engineperformance/JMHBenchmarks.java
```diff
@@ -33,4 +42,31 @@ public void simpleIndexing(EngineHolder eh) {
...
+				.forEntity( BookEntity.class )
+				.get()
+				.all()
+				.createQuery();
+
+		long expectedIndexSize = eh.getExpectedIndexSize();
+		int maxResults = eh.getMaxResults();
+
+		HSQuery hsQuery = searchIntegrator.createHSQuery().luceneQuery( luceneQuery );
+		hsQuery.targetedEntities( Arrays.<Class<?>>asList( BookEntity.class ) );
```
Sanne: it wasn't on purpose, just that this code is actually quite old :) I'll fix that.

engine/src/main/java/org/hibernate/search/indexes/impl/PeriodicRegreshingReaderProvider.java
```diff
@@ -0,0 +1,301 @@
+/*
```
Sanne: :+1:

engine/src/main/java/org/hibernate/search/indexes/impl/PeriodicRegreshingReaderProvider.java
```diff
@@ -0,0 +1,301 @@
...
+	 */
+	private synchronized PerDirectoryLatestReader createReader(Directory directory) {
+		PerDirectoryLatestReader reader = currentReaders.get( directory );
+		if ( reader != null ) {
+			return reader;
+		}
+
+		try {
+			reader = new PerDirectoryLatestReader( directory );
+			currentReaders.put( directory, reader );
```
Sanne: Right, in this case if there is a concurrent request it's preferrable to have a thread wait on the synchronization's lock rather than doing multiple IO operations to attempt to open multiple readers (and then have to close and discard one of them).

Think of it in terms of spent resources: if you lock, the thread will be parked and CPU time assigned to do something useful on a different thread. The cost is if course that your current thread will have to wait N milliseconds, were N is the time for the other thread to complete the indexreader refresh.

In the alternative strategy, using a putIfAbsent, the thread will not hit the lock but it will have to perform the same IO operations, which will take it N milliseconds to do (as it's the same operation as in the previous case).

So if you compare the two scenarios, while the lock might looks like "contention", it actually frees the CPU but does less IO, and takes the take time for the client to actually get an IndexReader.

So that's why I choose for the synchronized block.

engine/src/main/java/org/hibernate/search/indexes/impl/PeriodicRegreshingReaderProvider.java
```diff
@@ -0,0 +1,301 @@
...
+			int refCount = usageCounter.decrementAndGet();
+			if ( refCount == 0 ) {
+				ReaderUsagePair removed = allReaders.remove( reader );//remove ourself
+				try {
+					reader.close();
+				}
+				catch (IOException e) {
+					log.unableToCloseLuceneIndexReader( e );
+				}
+				assert removed != null;
```
Sanne: Any "current" IndexReader has an additional token in the `usageCounter`, so if close() method is invoked on the current IR, it won't actually be closed.

See also the comment on `usageCounter`.

Also on the method `refreshIfNeeded()` you'll see that the no-longer-current IndexReader is closed - yet this simply means that its usage counter is decremented, so that if any query is still using it it won't actually be closed, but it will only be closed when the last query which is using it will be closed.

Sanne:
That's right, the focus is on `JMHBenchmarks.concurrentReadWriteTest:readWriteTestReader`; it's good to see that the other metrics didn't change.

Sanne:
fixed as suggested and rebased.

Sanne:
@gustavonalle might be in good enough shape for you to give it a try as well?

Sanne:
to try it out:
 - https://github.com/hibernate/hibernate-search/pull/1196/files#diff-40176244800f79a0f62639216680e695R906

yrodiere:
Thanks for the explanations! Ok to merge on my side, I'll let you see with @gustavonalle .

Sanne:
cool, merged it.

PR #1199, opened 2016-10-26
HSEARCH-2432 Entity Ids containing underscore lost when querying

gustavonalle:
https://hibernate.atlassian.net/browse/HSEARCH-2432

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/DocumentIdHelper.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/DocumentIdHelper.java
@@ -21,8 +21,8 @@ public class DocumentIdHelper {
 		return work.getTenantId() == null ? work.getIdInString() : work.getTenantId() + "_" + work.getIdInString();
 	}
 
-	static String getEntityId(String documentId) {
-		if ( documentId.contains( "_" ) ) {
+	static String getEntityId(String documentId, String tenantId) {
+		if ( tenantId != null && documentId.contains( "_" ) ) {
 			documentId = documentId.substring( documentId.indexOf( "_" ) + 1 );
 		}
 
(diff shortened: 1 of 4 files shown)


yrodiere:
Thanks for the PR. I've thought a bit about it, and I'm not sure about that fix, though:

 * The condition probably should be using `org.hibernate.search.engine.impl.ImmutableSearchFactory.isMultitenancyEnabled()` instead of checking the tenant ID. Granted, the effect would be the same, since tenant IDs are mandatory once multi-tenancy is enabled, but this would be easier to read.
 * It seems to me that, if the tenant ID contains an underscore, there will still be a problem (https://hibernate.atlassian.net/browse/HSEARCH-2420)

The last one in particular make me thinks we're really shoving the issue under the carpet here. I think we should rather never try to extract the ID from the Elasticsearch ID, and simply use an in-document ID field. By chance, one pending PR added an ID field: https://github.com/hibernate/hibernate-search/pull/1192 .

I'll see if I can get your test working this way, and will submit a PR.

gustavonalle:
|  I think we should rather never try to extract the ID from the Elasticsearch ID, and simply use an in-document ID field

Sounds good. Out of curiosity, what will go in the ES `_id` field?

yrodiere:
> Out of curiosity, what will go in the ES _id field?
Same as now: the actual document ID (as opposed to the entity ID), which is the concatenation of the tenant ID and the document ID. Its main purpose is to allow ES to know when an update should erase an existing document.
It would only be of use to Hibernate Search when querying by entity ID (ES's "_id" queries), but this type of query is not implemented yet. Right know, users can only query on the document fields (which include the tenant ID and the entity ID in separate fields anyway).

yrodiere:
Closing this PR because of my follow-up: https://github.com/hibernate/hibernate-search/pull/1201

Thanks!

PR #1200, opened 2016-10-27
HSEARCH-2433 AssertionFailure with providedId in embeddeds

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2433

Details in the commit comments.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
+++ b/engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
@@ -240,7 +240,8 @@ public class AnnotationMetadataProvider implements MetadataProvider {
 		}
 		else {
 			if ( parseContext.includeEmbeddedObjectId() || pathsContext.containsPath( path ) ) {
-				createPropertyMetadataForEmbeddedId( member, typeMetadataBuilder, propertyMetadataBuilder, numericFields, configContext, parseContext, unprefixedAttributeName, path );
+				createPropertyMetadataForEmbeddedId( member, typeMetadataBuilder, propertyMetadataBuilder,
+						numericFields, configContext, parseContext, unprefixedAttributeName, path );
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
```diff
@@ -427,37 +430,57 @@ private Annotation getIdAnnotation(XProperty member, TypeMetadata.Builder typeMe
...
+		}
+
+		DocumentFieldMetadata fieldMetadata = fieldMetadataBuilder.build();
 
 		propertyMetadataBuilder.addDocumentField( fieldMetadata );
 		propertyMetadata = propertyMetadataBuilder.build();
-		typeMetadataBuilder.idProperty( propertyMetadata );
+
+		if ( isRoot ) {
+			typeMetadataBuilder.idProperty( propertyMetadata );
```
gsmet: Mmmmh, did I miss something or if `!isRoot`, the whole work we did above was totally useless?

Sanne:
@yrodiere you mention on the JIRA comments that this problem affects the Lucene backend as well.

Worth moving the test to `-engine` ?

yrodiere:
@Sanne Yes, I'll see if I can highlight the issue in an -engine test.

Sanne:
I meant to suggest to move this one, not to add an additional one. We don't want to have too many redundant tests checking for the same things ;)

yrodiere:
@Sanne I'll move to -orm, then. -engine tests are not yet executed on Elasticsearch. Not even sure yet when they will be...

Sanne:
> I'll move to -orm, then. -engine tests are not yet executed on Elasticsearch. 

Right, forgot about that. Let's merge this then, the missing test for the embedded mode is an orthogonal issue.

yrodiere:
@Sanne I rebased on master: the build was failing because of https://hibernate.atlassian.net/browse/HSEARCH-2438 , which was merged to master a few minutes ago

yrodiere:
@gsmet I addressed your concerns. Thanks!

Sanne:
merged, thanks all!

PR #1197, opened 2016-10-25
HSEARCH-2393 Remove dependency to Lucene in tests for features common to Lucene and Elasticsearch

yrodiere:
This PR is based on HSEARCH-2296, we'll have to review and merge the PR ( #1192) first.

Related JIRA ticket: https://hibernate.atlassian.net/browse/HSEARCH-2393

The purpose of this PR is to make tests currently ignored in `elasticsearch/pom.xml` work with Elasticsearch, so that we can enable them.

PR #1201, opened 2016-10-27
HSEARCH-2432 and HSEARCH-2420: issues with underscores in IDs

yrodiere:
~~This PR is based on HSEARCH-2393, we'll have to merge #1197 first.~~ => Done
This PR is based on HSEARCH-2450, we'll have to merge #1216 first.

Relevant JIRA tickets:
- https://hibernate.atlassian.net/browse/HSEARCH-2432
- https://hibernate.atlassian.net/browse/HSEARCH-2420

This fixes both issues by extracting the ID from the in-document ID field, and not from the "_id" in the Elasticsearch query hits.


elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/ElasticsearchMultitenancyIT.java
```diff
@@ -0,0 +1,89 @@
...
+
+	@Rule
+	public SearchFactoryHolder sfHolder = new SearchFactoryHolder( User.class )
+			.withProperty( "hibernate.search.default." + Environment.INDEX_MANAGER_IMPL_NAME, "elasticsearch" )
+			.withProperty( "hibernate.search.default." + ElasticsearchEnvironment.REFRESH_AFTER_WRITE, "true" )
+			.withProperty( "hibernate.search.default." + ElasticsearchEnvironment.INDEX_SCHEMA_MANAGEMENT_STRATEGY,
+					IndexSchemaManagementStrategy.RECREATE_DELETE.name() )
+			.withMultitenancyEnabled( true );
+
+	private void indexUser(String tenantId, User user, Serializable id, SearchIntegrator searchIntegrator) {
```
gsmet: Same remark about the order!

elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/ElasticsearchMultitenancyIT.java
```diff
@@ -0,0 +1,89 @@
...
+		Query query = queryBuilder.keyword().onField( "surname" ).matching( "Lee" ).createQuery();
+
+		List<EntityInfo> entityInfoList = searchIntegrator.createHSQuery( query, User.class ).queryEntityInfos();
+
+		Assert.assertEquals( 1, entityInfoList.size() );
+		Assert.assertEquals( customId, entityInfoList.iterator().next().getId() );
+	}
+
+	@Indexed
+	public static class User {
```
gsmet: Make it a private class maybe?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
```diff
@@ -77,15 +77,19 @@ public ElasticsearchIndexWorkVisitor(String indexName, boolean refreshAfterWrite
 		this.searchIntegrator = searchIntegrator;
 	}
 
+	static String getDocumentId(LuceneWork work) {
```
gsmet: Is it used anywhere else? If not, let's make it private (I suppose it was the purpose of dropping `DocumentIdHelper`).

gsmet:
So... I think I would have liked the last commit to be more step by step. There are a lot of changes which could have been made separately (and, I think, be more understandable).

Overall I like the changes, apart from the original issue, they fix a few things that's been bothering me for a while.

Btw, I don't know if it's GitHub bugging but it looks like there is a conflict (but let's wait for the other PR to be merged before rebasing).

yrodiere:
Thanks; I rebased, addressed your concerned and split the last commit in two. We can now wait for the other PR to be merged.

yrodiere:
@gsmet I just rebase this on #1197 (HSEARCH-2393), because it seems some tests enabled in #1197 are affected by these changes. So I'd rather have the CI execute those tests.
I noticed that while working on https://hibernate.atlassian.net/browse/HSEARCH-2425, which happens to have been fixed by accident in the last commit in this PR...

yrodiere:
So, as we discussed with @gsmet  yesterday, I had to add some fixes because of the rebase on top of HSEARCH-2393: the latter enabled more tests, and some of those are failing due to this PR. It's basically a pre-existing bugs that only surfaced because I changed the way we generate the ES mapping (we crawl through the document fields directly instead of going through the properties).

Anyway, @gsmet: I just updated the last two commits following your comments. I didn't rename the `documentFieldMetadata`/`documentFieldNameToFieldMetadata` properties to make it more obvious that they don't include non-embedded fields, because if I did, I should rename almost every property in `TypeMetadata`: none of them includes embedded information along with the main type's information. If we do something, I thing it should be part of a broader refactoring, but then the relevance to HSEARC-2420 and 2432 becomes less and less clear.

I'll wait for you to come back to me before merging, since I'm not sure you had time to express all of your concerns.

yrodiere:
@gsmet So... Actually, the more I fix it, the more tests begin to fail because previous bugs were hiding other ones. I'll just make another PR for all this metadata refactoring, because it's looking more and more like a major issue. Sorry for the noise...

yrodiere:
Alright, I extracted the problematic code to #1216 . We should be able to merge this PR as soon as the other one has been approved and merged.

PR #1203, opened 2016-10-27
HSEARCH-2414 and HSEARCH-2413: work around buggy ZonedDateTime format support in ES 2.4.1 and upgrade ES test version

yrodiere:
This PR changes ElasticsearchZonedDateTimeBridge to use a date format that ES 2.4 accepts and upgrades ES test version to 2.4

The downside is we don't use the standard format for ZonedDateTime on ES anymore. But right now ES is buggy (https://github.com/elastic/elasticsearch/issues/20911) so we can't do otherwise.

So we have to choose: either we wait for Joda to release the fix and ES to upgrade on both the 2.4 and 5.0 branches, or we use a non-standard format (and that will probably stay that way a long time because of backward compatibility)


gsmet:
I asked Stephen from Joda to know when he plans to release 2.9.5: https://github.com/JodaOrg/joda-time/issues/373 .

yrodiere:
Closing this PR following our discussion. A PR is coming for [HSEARCH-2436](https://hibernate.atlassian.net/browse/HSEARCH-2436) (document the bug)

PR #1204, opened 2016-10-28
HSEARCH-2435 Improve Elasticsearch configuration section

emmanuelbernard:
Describe the default value.
Explain when some of these options are useful.


emmanuelbernard:
BTW, I dont understand the following paragraph in the documentation

>    This will work as long as there is not a local analyzer defined with the same name. If that’s the case, Hibernate Search will ignore the analyzer when Elasticsearch is used as backend and log a warning. This might sound complicated but we are looking for ways to ease the experience.

What is a local analyzer? Do you mean that if there is a `@AnalyzerDef` with the same name, this one cancels the use of the Elasticsearch analyzer of the same name but does not make use of the local analyzer either?
Is that worth "just" a warning in that case?

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -122,6 +116,24 @@ Let's see the options for the `index_schema_management_strategy` property:
 |`RECREATE`|Indexed will be deleted if existing and then created. This will delete all content from the index!
 |`RECREATE_DELETE`|Similarly to `RECREATE` but will also delete the index at shutdown. Commonly used for tests.
 |===============
+Maximum time to wait for the indexes to become available before failing (in ms):: `hibernate.search.default.elasticsearch.index_management_wait_timeout 10000` (default)
+Status an index must at least have in order for Hibernate Search to work with it (one of "green", "yellow" or "red")::
+`hibernate.search.default.elasticsearch.required_index_status green` (default)
++
+Only operate if the index is at this level or safer.
+In development, set this value to `yellow` if the number of nodes started is below the number of expected replicas.
```
Sanne: Why would you suggest this? Seems quite a bad practice, you have the option to start your ES node as a single node, so it will be green, and faster.

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -284,19 +284,20 @@ If your metadata processors create fields with a different name from the one pas
...
+and all but one the query generated by the Hibernate Search DSL (see below).
 
 [NOTE]
 .Unsupported Query DSL features
 --
 Queries written via the DSL work.
 Open a JIRA otherwise.
 
-Notable exceptions are:
+The notable exception is more like this queries.
```
Sanne: should we quote or italicize "more like this" queries? The sentence seems weird, if you don't already know in advance what you mean.

Sanne:
Thanks, some important improvements in here. I highlighted a couple minor expressions I found confusing, but they are not blockers.

I can push this and send a follow-up if you prefer?

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
yrodiere: I just noticed that in ES 5.0, if you have no replica, your index will never be green. It wasn't the case in ES 2.x, I think.

yrodiere: Actually I'm wrong. I checked a second time, and it seems I messed up the configuration somewhere.

emmanuelbernard: @Sanne I am not sure I am following you but my understanding is that:
- a developer will only start a single node
- the default settings do require more than one nodes to be green

Hence the option to run at yellow. Am I missing something?

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -284,19 +284,20 @@ If your metadata processors create fields with a different name from the one pas
 === Queries
 
 You can write queries like you usually do in Hibernate Search: native Lucene queries and DSL queries (see <<search-query>>).
-We do automatically translate the most common types of Apache Lucene queries and many of the queries generated by the Hibernate Search DSL.
+We do automatically translate the most common types of Apache Lucene queries
+and all but one the query generated by the Hibernate Search DSL (see below).
```
emmanuelbernard: Fixed and pushed a new version.

emmanuelbernard:
What is the `+1 just push it` flag used for in general?

emmanuelbernard:
@Sanne I've fixed the sentence issue around the DSL. On the Elasticsearch yellow status, I wrote down my understanding but maybe I'm missing something.
But if you are correct, then we did not need the readiness option in the first place so I'm confused.

I think we can push as is and do a follow up.

Sanne:
> What is the +1 just push it flag used for in general?

We use it in two cases:
 - Reviewer thinks it's ok to integrate, but can't do that himself right away
 - Reviewer suggested minor improvements, but leaves it up to the requestor to choose to follow up on the suggestions or decide to push it as-is

I guess the first case is getting obsolete now that we have the option to merge from the web gui..

Sanne:
cool, merged! Thanks

PR #1206, opened 2016-10-28
HSEARCH-2437 Prepare the support for Elasticsearch 5 (+ HSEARCH-2439)

yrodiere:
Relevant tickets:

 * https://hibernate.atlassian.net/browse/HSEARCH-2437
 * https://hibernate.atlassian.net/browse/HSEARCH-2439 (it's linked because I only spotted the bug when trying out ES 5)

A few changes that would facilitate the ES 5 support, and yet do not break 2.x compatibility.

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java
@@ -328,13 +328,10 @@ public class ElasticsearchHSQueryImpl extends AbstractHSQuery {
 			this.idFieldName = queriedEntityIdFieldName;
 
 			// Query filters; always a type filter, possibly a tenant id filter;
-			JsonObject effectiveFilter = getEffectiveFilter( typeFilters );
+			JsonObject filteredQuery = getFilteredQuery( jsonQuery.get( "query" ), typeFilters );
 
 			JsonBuilder.Object completeQuery = JsonBuilder.object();
-
-			completeQuery.add( "query",
-					JsonBuilder.object()
-							.add( "filtered", JsonBuilder.object( jsonQuery ).add( "filter", effectiveFilter ) ) );
+			completeQuery.add( "query", filteredQuery );
 
 			if ( !getFacetManager().getFacetRequests().isEmpty() ) {
 				JsonBuilder.Object facets = JsonBuilder.object();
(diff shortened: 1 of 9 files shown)


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java
```diff
@@ -328,13 +328,15 @@ private IndexSearcher() {
...
-			JsonObject effectiveFilter = getEffectiveFilter( typeFilters );
+			JsonObject filteredQuery = getFilteredQuery( jsonQuery.get( "query" ), typeFilters );
 
 			JsonBuilder.Object completeQuery = JsonBuilder.object();
-
-			completeQuery.add( "query",
-					JsonBuilder.object()
-							.add( "filtered", JsonBuilder.object( jsonQuery ).add( "filter", effectiveFilter ) ) );
+			// Copy the properties from the initial query
+			for ( Entry<String, JsonElement> entry : jsonQuery.entrySet() ) {
```
gsmet: Do we agree this particular thing is not related to the fix at hand? (just to be sure I understand correctly)

gsmet:
Looks good overall.

I think ba86523077d30ffa0a198b01446f6ff2387bd44b would merit its own JIRA as it's a bug not directly related to Es 5 (you can keep it in this PR but a specific JIRA would be nice).

As stated on IRC, I have my concerns about the Integer.MAX_VALUE thing. Waiting for your feedback on this.

yrodiere:
By the way, here's the commit that removed the size=0 feature: https://github.com/elastic/elasticsearch/commit/dcc598c414184ffa6cf06b3a69eb15767e274330 (for the record)

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java
gsmet: I would appreciate another JIRA. It changes the way we interpret the user query so I would prefer having a specific issue mentioning it.

gsmet:
OK, looks like we're on the safe side for the Integer.MAX_VALUE thing, thanks for checking. I let you split the other commit in 2 with a different JIRA number and merge?

Thanks!

yrodiere:
Merged, thanks!

PR #1207, opened 2016-10-28
HSEARCH-2384 Make the Elasticsearch version used for integration tests a build parameter

yrodiere:
~~This PR is based on HSEARCH-2437, so we need to merge #1206 first.~~ => Merged

Relevant JIRA ticket: https://hibernate.atlassian.net/browse/HSEARCH-2384

Not sure about this one...
Actually, there already was a parameter to test against another Elasticsearch version (kinda): we only need to pass `-DtestElasticsearchVersion=<myVersion>`.
Now, I polished it a bit, notably by including the parameter in the generated doc, but there's not much to do as long as we stay on ES 2.x.

On ES 5.x though, things get a bit tricky.
First thing: we have to use maven profiles, because the plugin configuration itself will change. Most notably, some dependencies are not available in ES 5 (the DeleteQuery plugin), and the elasticsearch-maven-plugin will be in a different version. Also, we need additional dependencies because ES 5 uses log4j2.
But the worst is, there's no support from elasticsearch-maven-plugin for ES 5 yet. So I had to work it around, forking the plugin and building it if needed in the `.travis.yml`. From Jenkins, we can't even build yet, and locally we have to first build the elasticsearch-maven-plugin from my fork and install it in the local Maven repository.

We could keep it that way, temporarily, though I have my doubts about licensing (since we're building a modified version of one of our compile-time dependencies, there may be some legal requirements on our side). Or we can wait and see what happens to my PR (https://github.com/alexcojocaru/elasticsearch-maven-plugin/pull/19). But if it doesn't move, we'll have to do _something_ (either maintain a fork, or setup our build environments differently).

*Update*: I got additional pieces of information, and it seems the future isn't very bright for elasticsearch-maven-plugin, mainly because the Elasticsearch team doesn't support embedded mode and will not publish plugins/modules JARs anymore. So the way to go is probably this: http://david.pilato.fr/blog/2016/10/18/elasticsearch-real-integration-tests-updated-for-ga/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -181,11 +181,6 @@
         <commonsLangVersion>2.6</commonsLangVersion>
 
         <!-- Elasticsearch -->
-        <testElasticsearchMavenPluginVersion>2.2</testElasticsearchMavenPluginVersion>
-        <!-- In theory we're not coupled to a specific version of Elasticsearch; this property
-             controls which version is being used to run integration tests.
-        -->
-        <testElasticsearchVersion>2.3.1</testElasticsearchVersion>
         <elasticsearchJestVersion>2.0.0</elasticsearchJestVersion>
         <elasticsearchCommonsLang3Version>3.4</elasticsearchCommonsLang3Version>
         <elasticsearchGsonVersion>2.4</elasticsearchGsonVersion>
@@ -577,6 +572,7 @@
                 <artifactId>slf4j-log4j12</artifactId>
                 <version>${slf4jVersion}</version>
             </dependency>
+            
             <dependency>
                 <groupId>log4j</groupId>
                 <artifactId>log4j</artifactId>
@@ -1259,18 +1255,8 @@ org.jboss.logging.BasicLogger#debugf(java.lang.Throwable, java.lang.String, long
                             <artifactId>elasticsearch</artifactId>
                             <version>${testElasticsearchVersion}</version>
                         </dependency>
-
-                        <!-- The next three are used to get ES log output into the Maven console -->
-                        <dependency>
-                            <groupId>org.slf4j</groupId>
-                            <artifactId>slf4j-api</artifactId>
-                            <version>${slf4jVersion}</version>
-                        </dependency>
-                        <dependency>
-                            <groupId>org.slf4j</groupId>
-                            <artifactId>slf4j-log4j12</artifactId>
-                            <version>${slf4jVersion}</version>
-                        </dependency>
+                        
+                        <!-- The next is used to get ES log output into the Maven console -->
                         <dependency>
                             <groupId>org.hibernate</groupId>
                             <artifactId>hibernate-search-sharedtestresources</artifactId>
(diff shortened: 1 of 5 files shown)


gsmet:
Changed the status to "On ice" waiting for a global decision on the Es 5 support.

Sanne:
> Now, I polished it a bit, notably by including the parameter in the generated doc, but there's not much to do as long as we stay on ES 2.x.

I'd be happy to apply this then, ES5 is out of scope for the specific issue. We also need this to be able to document that (if?) it works fine with ES 2.3, 1.5, ... and possibly setup CI jobs.

pom.xml
```diff
@@ -1278,14 +1277,6 @@ org.jboss.logging.BasicLogger#debugf(java.lang.Throwable, java.lang.String, long
                             <artifactId>hibernate-search-sharedtestresources</artifactId>
                             <version>${project.version}</version>
                         </dependency>
-
-                        <!-- JNA isn't part of the ES fat JAR; Adding it so ES doesn't complain 
-                             upon start-up -->
-                        <dependency>
```
gsmet: So, this issue does not exist anymore?

yrodiere:
@gsmet @Sanne I refactored the PR to remove any ES 5.0-related stuff, and to support ES 2.0/2.1 (which happen to require a different configuration). Ready for review!

(FYI, the ES 5.0 stuff has been moved to my ES 5.0 branch: https://github.com/yrodiere/hibernate-search/compare/HSEARCH-2384...yrodiere:HSEARCH-2434)

Sanne:
cool, thanks! Reviewing..

Sanne:
don't we need the groovy plugin on ES 2.1 ?

yrodiere:
It doesn't exist below version 2.2 :/ Maybe with different coordinates, I'll have a look.

yrodiere:
Okay, it's org.elasticsearch:elasticsearch-groovy. I'll try that and force-push.

gsmet:
Stupid question but do we really want to spend time supporting such older versions?

IIRC, Sanne, you told me AWS supported 2.3. It would make us support 2.3, 2.4 (and hopefully 5) but wouldn't care about older releases of Es.

Sanne:
indeed, I wouldn't spend too much time supporting older versions.

If it doesn't work, fine. The PR is still valid as it does start the tests using the older version ;)

But I wondered, as the `README` seemed to suggest that 2.1 is supposed to work.

gsmet:
Yeah, I think we should get rid of that and support:
- the last version supported by AWS (provided it's not too old)
- the last 2.x version
- eventually 5.x

And that's all. I don't think it has any value to support Elasticsearch versions which are probably not supported anymore.

yrodiere:
@gsmet @Sanne Fixed. Or at least the server now starts with groovy support even for 2.0/2.1. May we at least keep the settings, even if we advertise we don't support it in the documentation? Please? :/

Sanne:
yes, happy to keep it for now. Thanks!

Sanne:
merged

PR #1208, opened 2016-10-31
HSEARCH-2438 Fix erroneous test failing with DST change

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2438


Sanne:
thanks, merged

PR #1210, opened 2016-10-31
HSEARCH-2253 Remove ID sort workaround using "_uid" in Elasticsearch

yrodiere:
This PR is based on #1192 (HSEARCH-2396), which will have to be merged first.

Relevant JIRA ticket: https://hibernate.atlassian.net/browse/HSEARCH-2253


gsmet:
Nice! Good to go once the pending one is merged.

PR #1211, opened 2016-11-02
HSEARCH-2412 Add the target URL in Elasticsearch error messages

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2412


gsmet:
Merged! Thanks!

PR #1212, opened 2016-11-02
HSEARCH-2330 (+	HSEARCH-2261) Allow MetadataProvidingFieldBridges to override field type 

yrodiere:
This PR is based on #1197 (HSEARCH-2393), which will have to be merged first.

Relevant tickets:

 * https://hibernate.atlassian.net/browse/HSEARCH-2330
 * https://hibernate.atlassian.net/browse/HSEARCH-2261

I implemented the solution suggested in my last comment on HSEARCH-2330. By allowing `MetadataProvidingFieldBridge`s to override the field type, we end up removing the need for users to use workarounds, such as adding a @Numeric annotation on non-numeric properties that have a custom field bridge that will generate a numeric field. Now the whole knowledge about the type resides in the custom field bridge, which seems a bit cleaner.

### DIFF
--- a/engine/src/test/java/org/hibernate/search/test/dsl/SortDSLTest.java
+++ b/engine/src/test/java/org/hibernate/search/test/dsl/SortDSLTest.java
@@ -13,6 +13,7 @@ import java.util.Arrays;
 import java.util.List;
 
 import org.apache.lucene.document.Document;
+import org.apache.lucene.document.DoubleDocValuesField;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.Sort;
 import org.apache.lucene.search.SortField;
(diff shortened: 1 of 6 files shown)


gsmet:
Apart from my question, it looks good to me and it's nice to see this one tackled!

engine/src/test/java/org/hibernate/search/test/dsl/SortDSLTest.java
```diff
@@ -673,7 +682,7 @@ public void set(String name, Object value, Document document, LuceneOptions luce
 				return;
 			}
 
-			luceneOptions.addNumericFieldToDocument( name, doubleValue, document );
+			document.add( new DoubleDocValuesField( name, doubleValue ) );
```
yrodiere: Simply because there's no way to add a doc values field using luceneOptions. It seems to be designed that way, seeing the examples in the documentation.
I agree it's a bit weird, though. Another ticket, maybe?

gsmet: Silly me. I think I would prefer to keep the original field in addition with the sortable one. It's more in line with what we do elsewhere. WDYT?

About adding something in LuceneOptions, I wonder if it's worth it, given the FieldBridge 2.0 plan but I agree it's a bit harsh for the user to have to go in this sort of details while we usually hide all that machinery from him. If I had to do something, I think I would add a method to add a sortable field which adds the original field and the doc values one in one go and hide this logic from the user. Of course, it is a bit redundant with the fact that we declare the field sortable a few lines above but I don't think we can do better given the current API. Anyway, I agree this part is for another day!

yrodiere: It's not strictly necessary, but as you wish. I amended the commit.

About making it easier, I created the ticket: https://hibernate.atlassian.net/browse/HSEARCH-2447

gsmet:
OK, this one is good to go too.

PR #1213, opened 2016-11-03
HSEARCH-2422 Always compute scores when projecting on the score in Elasticsearch

yrodiere:
This PR is based on #1197 (HSEARCH-2393), so we'll have to merge that one first.

Relevant JIRA ticket: https://hibernate.atlassian.net/browse/HSEARCH-2422


gsmet:
Good to go, once the other PRs are merged!

PR #1214, opened 2016-11-03
HSEARCH-2425 Projection on spatial coordinates returns a String instead of a number with Elasticsearch

yrodiere:
This PR is based on #1201 (HSEARCH-2332 and HSEARCH-2420), so we'll have to merge that one first.

Relevant JIRA ticket: https://hibernate.atlassian.net/browse/HSEARCH-2425

The full story is that one commit fixed in another PR ("HSEARCH-2420 and HSEARCH-2432 Rationalize data extraction from ES query"), which was intended to fix another issue, happened to also fix HSEARCH-2425. So we only have to enable the test again and profit.


gsmet:
OK, well, I like this one :).

PR #1216, opened 2016-11-04
HSEARCH-2450 In internal metadata, provide access to BridgeDefinedFields through their source document field

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2450

Note: the first commit was originally part of #1201 , but has been moved to a specific ticket/PR due to the amount of additional work it required (most notably fixing previously hidden bugs).

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/metadata/impl/TypeMetadata.java
+++ b/engine/src/main/java/org/hibernate/search/engine/metadata/impl/TypeMetadata.java
@@ -80,6 +80,11 @@ public class TypeMetadata {
 	private final Map<String, DocumentFieldMetadata> documentFieldNameToFieldMetadata;
 
 	/**
+	 * Metadata for a bridge-defined field keyed against the field name
+	 */
+	private final Map<String, BridgeDefinedField> bridgeDefinedFieldNameToFieldMetadata;
+
+	/**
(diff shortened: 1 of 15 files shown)


engine/src/main/java/org/hibernate/search/engine/metadata/impl/TypeMetadata.java
```diff
@@ -228,22 +241,22 @@ public BridgeDefinedField getBridgeDefinedFieldMetadataFor(String fieldName) {
...
 			Collection<DocumentFieldMetadata> allMetadata = new ArrayList<DocumentFieldMetadata>(
-					documentFieldNameToFieldMetadata.size()
+					documentFieldMetadata.size()
 			);
 			for ( EmbeddedTypeMetadata element : embeddedTypeMetadata ) {
 				allMetadata.addAll( element.getAllDocumentFieldMetadata() );
 			}
-			allMetadata.addAll( documentFieldNameToFieldMetadata.values() );
+			allMetadata.addAll( documentFieldMetadata );
 			return allMetadata;
```
gsmet: I would have expected a `Collections.unmodifiableCollection` here.

gsmet:
@yrodiere looks good to me. I let you merge it with the others!

PR #1217, opened 2016-11-08
HSEARCH-2451 Use Elasticsearch "fields" feature to implement faceting (+ HSEARCH-2457)

yrodiere:
~~This PR is based on #1168 (HSEARCH-2260), so we'll have to merge that one first.~~ => Done!

Relevant JIRA tickets:

 * https://hibernate.atlassian.net/browse/HSEARCH-2451
 * https://hibernate.atlassian.net/browse/HSEARCH-2457

HSEARCH-2457 is addressed here because solving it helps to understand changes to the metadata in HSEARCH-2451.

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java
@@ -681,11 +681,11 @@ public class ElasticsearchHSQueryImpl extends AbstractHSQuery {
 				}
 			}
 
-			return new EntityInfoImpl( clazz, binding.getDocumentBuilder().getIdentifierName(), (Serializable) id, projections );
+			return new EntityInfoImpl( clazz, binding.getDocumentBuilder().getIdPropertyName(), (Serializable) id, projections );
 		}
 
 		private Object getId(JsonObject hit, EntityIndexBinding binding, ConversionContext conversionContext) {
-			return getFieldValue( binding, hit, binding.getDocumentBuilder().getIdKeywordName(), conversionContext );
+			return getFieldValue( binding, hit, binding.getDocumentBuilder().getIdFieldName(), conversionContext );
 		}
 
 		/**
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java
@@ -696,7 +696,7 @@ public class ElasticsearchHSQueryImpl extends AbstractHSQuery {
 			DocumentFieldMetadata field = binding.getDocumentBuilder().getTypeMetadata().getDocumentFieldMetadataFor( projectedField );
 
 			if ( field != null ) {
-				conversionContext.pushProperty( field.getFieldName() );
+				conversionContext.pushProperty( field.getAbsoluteName() );
 				try {
 					return convertFieldValue( hit, binding, field, conversionContext );
 				}
@@ -717,7 +717,7 @@ public class ElasticsearchHSQueryImpl extends AbstractHSQuery {
 		}
 
 		private Object convertFieldValue(JsonObject hit, EntityIndexBinding binding, DocumentFieldMetadata field, ConversionContext conversionContext) {
-			String fieldPath = field.getFieldName();
+			String fieldPath = field.getAbsoluteName();
 			FieldBridge fieldBridge = field.getFieldBridge();
 			JsonElement jsonValue = extractFieldValue( hit.get( "_source" ).getAsJsonObject(), fieldPath );
 
@@ -732,14 +732,14 @@ public class ElasticsearchHSQueryImpl extends AbstractHSQuery {
 
 				// Add to the document the additional fields created when indexing the value
 				for ( BridgeDefinedField bridgeDefinedField : field.getBridgeDefinedFields().values() ) {
-					String bridgeDefinedFieldPath = bridgeDefinedField.getName();
+					String bridgeDefinedFieldPath = bridgeDefinedField.getAbsoluteName();
 					ExtendedFieldType bridgeDefinedFieldType = FieldHelper.getType( bridgeDefinedField );
 
 					JsonElement bridgeDefinedFieldJsonValue = extractFieldValue( hit.get( "_source" ).getAsJsonObject(), bridgeDefinedFieldPath );
 					addDocumentField( tmp, binding, bridgeDefinedFieldType, bridgeDefinedFieldPath, bridgeDefinedFieldJsonValue );
 				}
 
-				return conversionContext.twoWayConversionContext( (TwoWayFieldBridge) fieldBridge ).get( field.getName(), tmp );
+				return conversionContext.twoWayConversionContext( (TwoWayFieldBridge) fieldBridge ).get( field.getAbsoluteName(), tmp );
 			}
 			// Should only be the case for custom bridges
 			else {
@@ -775,7 +775,7 @@ public class ElasticsearchHSQueryImpl extends AbstractHSQuery {
 		}
 
 		private Object convertFieldValue(JsonObject hit, EntityIndexBinding binding, BridgeDefinedField bridgeDefinedField) {
-			String fieldPath = bridgeDefinedField.getName();
+			String fieldPath = bridgeDefinedField.getAbsoluteName();
 			ExtendedFieldType fieldType = FieldHelper.getType( bridgeDefinedField );
 			JsonElement jsonValue = extractFieldValue( hit.get( "_source" ).getAsJsonObject(), fieldPath );
 			switch ( fieldType ) {
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java
@@ -50,6 +50,8 @@ import org.hibernate.search.engine.impl.FilterDef;
 import org.hibernate.search.engine.integration.impl.ExtendedSearchIntegrator;
 import org.hibernate.search.engine.metadata.impl.BridgeDefinedField;
 import org.hibernate.search.engine.metadata.impl.DocumentFieldMetadata;
+import org.hibernate.search.engine.metadata.impl.FacetMetadata;
+import org.hibernate.search.engine.metadata.impl.TypeMetadata;
 import org.hibernate.search.engine.service.spi.ServiceReference;
 import org.hibernate.search.engine.spi.DocumentBuilderIndexedEntity;
 import org.hibernate.search.engine.spi.EntityIndexBinding;
@@ -290,6 +292,7 @@ public class ElasticsearchHSQueryImpl extends AbstractHSQuery {
 	private class IndexSearcher {
 
 		private final Map<String, Class<?>> entityTypesByName = new HashMap<>();
+		private final Map<String, Class<?>> documentBuilderIndexedEntityByName = new HashMap<>();
 		private final Set<String> indexNames = new HashSet<>();
 		private final String completeQueryAsString;
 
@@ -328,7 +331,7 @@ public class ElasticsearchHSQueryImpl extends AbstractHSQuery {
 				JsonBuilder.Object facets = JsonBuilder.object();
 
 				for ( Entry<String, FacetingRequest> facetRequestEntry : getFacetManager().getFacetRequests().entrySet() ) {
-					ToElasticsearch.addFacetingRequest( facets, facetRequestEntry.getValue() );
+					addFacetingRequest( facets, facetRequestEntry.getValue() );
 				}
 
 				completeQuery.add( "aggregations", facets );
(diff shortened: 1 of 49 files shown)


orm/src/main/java/org/hibernate/search/batchindexing/impl/IdentifierConsumerDocumentProducer.java
```diff
@@ -268,9 +268,7 @@ private void index(Object entity, Session session, InstanceInitializer sessionIn
 		DocumentBuilderIndexedEntity docBuilder = entityIndexBinding.getDocumentBuilder();
 		TwoWayFieldBridge idBridge = docBuilder.getIdBridge();
 		String idPropertyName = docBuilder.getIdPropertyName();
-		if ( idPropertyName != null ) {
-			conversionContext.pushProperty( docBuilder.getIdFieldName() );
-		}
+		conversionContext.pushProperty( docBuilder.getIdFieldName() );
```
gsmet: Care to explain this change? (maybe a separate commit would be in order)

gsmet: Looks like you roll this back in the following commit. Maybe you should remove this change from this commit?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java
```diff
@@ -431,6 +435,28 @@ private JsonObject getTenantIdFilter() {
...
+				EntityIndexBinding binding = extendedIntegrator.getIndexBinding( entityType );
+				TypeMetadata typeMetadata = binding.getDocumentBuilder().getTypeMetadata();
+				facetMetadata = typeMetadata.getFacetMetadataFor( facetFieldAbsoluteName );
+				if ( facetMetadata != null ) {
+					break;
+				}
+			}
+
+			if ( facetMetadata == null ) {
+				throw new AssertionFailure( "Unknown facet field name:" + facetFieldAbsoluteName );
```
gsmet: Didn't look at the path to arrive here but we are sure the user can't get this error? If he can, we should move this the logger.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ToElasticsearch.java
```diff
@@ -59,22 +64,24 @@
 	private ToElasticsearch() {
 	}
 
-	public static void addFacetingRequest(JsonBuilder.Object jsonQuery, FacetingRequest facetingRequest) {
-		String fieldName = facetingRequest.getFieldName();
+	public static void addFacetingRequest(JsonBuilder.Object jsonQuery, FacetingRequest facetingRequest,
+			String sourceFieldName, String facetRelativedName) {
```
gsmet: Is the sourceFieldName absolute or relative? As you wanted to be clear everywhere, it might be a good idea to be more precise.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/schema/impl/DefaultElasticsearchSchemaTranslator.java
```diff
@@ -151,13 +152,11 @@ private void addPropertyMapping(ElasticsearchMappingBuilder mappingBuilder, Docu
 		// Create facet fields if needed: if the facet has the same name as the field, we don't need to create an
 		// extra field for it
```
gsmet: The comment looks outdated.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/schema/impl/DefaultElasticsearchSchemaTranslator.java
```diff
@@ -214,17 +213,17 @@ else if ( SpatialHelper.isSpatialFieldLatitude( propertyPath ) ) {
 		}
 	}
 
-	private void addPropertyMapping(ElasticsearchMappingBuilder mappingBuilder, FacetMetadata facetMetadata) {
-		String propertyPath = facetMetadata.getAbsoluteName();
+	private void addFieldMapping(PropertyMapping propertyMapping, ElasticsearchMappingBuilder mappingBuilder, FacetMetadata facetMetadata) {
```
gsmet: Maybe rename the method `addFacetFieldMapping`?

gsmet: Or you wanted to make it more general? If so maybe Subfield?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/schema/impl/model/PropertyMapping.java
```diff
@@ -7,6 +7,8 @@
 package org.hibernate.search.elasticsearch.schema.impl.model;
 
 import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
```
gsmet: A few lines below (does not appear in the diff but just expand the lines), s/representic/representing/

engine/src/main/java/org/hibernate/search/engine/metadata/impl/DocumentFieldMetadata.java
```diff
@@ -216,13 +226,14 @@ public String toString() {
 
 		public Builder(BackReference<TypeMetadata> sourceType,
 				BackReference<PropertyMetadata> sourceProperty,
-				String absoluteName,
+				String absoluteName, String relativeName,
```
gsmet: If feels really weird to have the `absoluteName` and the `relativeName`. Maybe it's time to introduce a `FieldPath` enclosing the prefix and the relative name and having a method building the absolute name?

engine/src/main/java/org/hibernate/search/engine/metadata/impl/TypeMetadata.java
```diff
@@ -234,6 +240,20 @@ public BridgeDefinedField getBridgeDefinedFieldMetadataFor(String fieldName) {
 		return null;
 	}
 
+	public FacetMetadata getFacetMetadataFor(String facetFieldName) {
+		FacetMetadata result = facetFieldNameToFacetMetadata.get( facetFieldName );
+		if ( result != null ) {
+			return result;
+		}
+		for ( EmbeddedTypeMetadata element : embeddedTypeMetadata ) {
```
gsmet: Is there a reason why we do not include the embedded metadata in the various maps?

engine/src/main/java/org/hibernate/search/engine/metadata/impl/FacetMetadata.java
```diff
@@ -63,6 +70,7 @@ public FacetEncodingType getEncoding() {
 
 		// optional parameters
 		private FacetEncodingType encoding = FacetEncodingType.AUTO;
```
gsmet: Not related to your patch but I dislike the fact that the default value of encoding is AUTO while this value is not supported when exploiting `FacetMetadata` later on. It's quite misleading.

So defining the encoding type is mandatory, not optional.

Not sure what to do about it but I thought I might as well mention it.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/schema/impl/DefaultElasticsearchSchemaTranslator.java
```diff
@@ -281,6 +281,15 @@ private void addTypeOptions(PropertyMapping propertyMapping, BridgeDefinedField
...
 
+		if ( facetMetadata.isEncodingAuto() ) {
+			/*
+			 * If the user didn't ask for a specific encoding, just use the same datatype
+			 * as the source field.
+			 */
+			addTypeOptions( fieldMapping, facetMetadata.getSourceField() );
+			return;
+		}
+
```
gsmet: OK, I see how it was the easiest way to fix the issue, considering the current state of the FacetMetadata thing and the fact that dates are mapped to long, which is not very convenient for Es.

But I don't like it very much.

I didn't look at the Lucene part or at which changes would be needed to do that but couldn't we keep the `AUTO` value in the `FacetMetadata` and determine the encoding type in the Lucene specific part, such as what we do here, considering we have the sourceField at hand?

Without looking at the code, I'd say the issue is that for Lucene we probably need to determine the encoding type several times.

Just thinking out loud.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/schema/impl/DefaultElasticsearchSchemaTranslator.java
```diff
@@ -151,13 +152,11 @@ private void addPropertyMapping(ElasticsearchMappingBuilder mappingBuilder, Docu
...
-					addPropertyMapping( mappingBuilder, facetMetadata );
-				}
-				catch (IncompleteDataException e) {
-					LOG.debug( "Not adding a mapping for facet " + facetMetadata.getAbsoluteName() + " because of incomplete data", e );
-				}
+			try {
+				addFieldMapping( propertyMapping, mappingBuilder, facetMetadata );
+			}
+			catch (IncompleteDataException e) {
+				LOG.debug( "Not adding a mapping for facet " + facetMetadata.getAbsoluteName() + " because of incomplete data", e );
```
yrodiere: That code goes back to long before this PR (it's even been refactored at least twice), and similar debug messages are generated for other types of mappings (regular fields, bridge-defined fields). Probably not something to address in this PR.
I created a separate JIRA: https://hibernate.atlassian.net/browse/HSEARCH-2458

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/schema/impl/DefaultElasticsearchSchemaTranslator.java
yrodiere: Yes, that's really annoying: what we call "fields" in Lucene are called "properties" in ES, and ES gives "fields" a totally different meaning.

I renamed to `addSubfieldMetadata`. Not very correct, but at least it's not ambiguous.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/schema/impl/model/PropertyMapping.java
yrodiere: I fixed it in the other PR ( #1168 ). Thanks.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/schema/impl/model/PropertyMapping.java
```diff
@@ -119,6 +126,25 @@ public void setNullValue(JsonPrimitive nullValue) {
 		this.nullValue = nullValue;
 	}
 
+	public Map<String, PropertyMapping> getFields() {
+		return fields;
+	}
+
+	private Map<String, PropertyMapping> getInitializedFields() {
```
yrodiere: We use Gson, not Jackson. Gson seems only serializes fields, and not methods (unless we use custom serializers).

The reflective type adapter (the default) seems to really only consider fields: https://github.com/google/gson/blob/gson-parent-2.4/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java#L147

And anyway, the `Expose` annotation (used to ignore data when we cannot mark fields as "transient") can only be applied to fields, not to methods:

```
@Target(ElementType.FIELD)
public @interface Expose {
```

Also, *if* the result of this method was being serialized, a good load of tests would be failing, because "initializedFields" would not be understood by Elasticsearch.

engine/src/main/java/org/hibernate/search/engine/metadata/impl/DocumentFieldMetadata.java
yrodiere: It doesn't feel weird to me, but I agree enclosing it in a separate object would be cleaner. I think we could leverage the `PathContext` to generate those paths, too. 
I'll work on that.

engine/src/main/java/org/hibernate/search/engine/metadata/impl/TypeMetadata.java
yrodiere: For other maps there is, as we sometime want to include only non-embedded metadata (especially when generating the ES mapping). Here, we never want, but we might as well be consistent.
Anyway, we'll all agree that something has to change in TypeMetadata. Nothing urgent, though: I'd say it's 6.0 stuff.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/schema/impl/DefaultElasticsearchSchemaTranslator.java
yrodiere: The issue is mainly that we don't have Lucene-specific metadata, so we're putting this lucene-specific encoding in the general metadata. Like we do for a lot of other things.

If we wanted to move the automatic encoding resolution around, there's no Lucene-specific, central place to store the resolved encoding, so the only thing we could do would be to do type resolution every time we need the actual encoding, which would be in `org.hibernate.search.engine.spi.DocumentBuilderIndexedEntity.addFacetDocValues(Document, DocumentFieldMetadata, FacetMetadata, Object)`. Which I find far more unappealing than adding that flag in the current metadata.
Note that it's not an issue of duplicating code (we only use the encoding in one place, it seems), but really more an issue of re-executing bootstrap code again and again at runtime.

Anyway, you're telling me "we need to break free of Lucene dependency", and that's definitely not the purpose of this PR, but more of 6.0. Once we have Lucene-independent APIs and (perhaps more importantly) engine, we'll add Lucene-specific metadata on top of engine metadata, as we will for Elasticsearch, and everything will be fine. Until then, well... we'll do as best we can.

engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
```diff
@@ -853,18 +858,12 @@ private void buildDocumentFieldForClassBridges(Document doc,
...
-						fieldName,
-						unproxiedInstance,
-						doc,
-						typeMetadata.getClassLuceneOptions( fieldMetadata, documentBoost )
-				);
-			}
-			finally {
-				conversionContext.popProperty();
-			}
+			oneWayConversionContext.set(
```
yrodiere: Here is the commit: https://github.com/hibernate/hibernate-search/pull/1217/commits/966d90a45940f9edddbd8149263fb742971a83ac
(`HSEARCH-2457 Make sure ContextualExceptionBridgeHelper provides as much information as possible in exception messages`)

yrodiere:
@gsmet I added the DocumentFieldPath type so it's a bit less "weird": https://github.com/hibernate/hibernate-search/pull/1217/commits/e4d013b4ce5ed2124f792fdc221ee82def04280e (HSEARCH-2451 Introduced DocumentFieldPath). I kept the `getAbsoluteName` methods on `DocumentFieldMetadata` / `FacetMetadata` for now, because it seemed a bit easier to use to me, but I can remove it if you feel it's not necessary.

engine/src/main/java/org/hibernate/search/engine/metadata/impl/TypeMetadata.java
yrodiere: Follow-up: https://hibernate.atlassian.net/browse/HSEARCH-2463

yrodiere:
Jenkins, retest this please.

gsmet:
OK, merged, thanks!

PR #1218, opened 2016-11-09
HSEARCH-2445 Improve error handling

gustavonalle:
https://hibernate.atlassian.net/browse/HSEARCH-2445

As described in the JIRA, this PR introduces two changes:
* Expose the ```IndexManager``` in the ```ErrorContext```
* Relocate the ```ErrorHandler``` creation to ```Configuration``` level so that it's easier to eventually wrap the default one

### DIFF
--- a/engine/src/main/java/org/hibernate/search/spi/SearchIntegratorBuilder.java
+++ b/engine/src/main/java/org/hibernate/search/spi/SearchIntegratorBuilder.java
@@ -48,7 +48,6 @@ import org.hibernate.search.engine.impl.ReflectionReplacingSearchConfiguration;
 import org.hibernate.search.engine.integration.impl.ExtendedSearchIntegrator;
 import org.hibernate.search.engine.metadata.impl.AnnotationMetadataProvider;
 import org.hibernate.search.engine.metadata.impl.TypeMetadata;
-import org.hibernate.search.engine.service.classloading.spi.ClassLoaderService;
 import org.hibernate.search.engine.service.impl.StandardServiceManager;
 import org.hibernate.search.engine.service.spi.ServiceManager;
 import org.hibernate.search.engine.spi.DocumentBuilderContainedEntity;
@@ -58,7 +57,6 @@ import org.hibernate.search.engine.spi.EntityState;
 import org.hibernate.search.engine.spi.SearchMappingHelper;
 import org.hibernate.search.exception.ErrorHandler;
 import org.hibernate.search.exception.SearchException;
-import org.hibernate.search.exception.impl.LogErrorHandler;
 import org.hibernate.search.filter.FilterCachingStrategy;
 import org.hibernate.search.filter.impl.CachingWrapperFilter;
 import org.hibernate.search.filter.impl.MRUFilterCachingStrategy;
@@ -91,6 +89,7 @@ public class SearchIntegratorBuilder {
 	private SearchConfiguration cfg;
 	private MutableSearchFactory rootFactory;
 	private final List<Class<?>> classes = new ArrayList<Class<?>>();
+	private final ErrorHandlerFactory errorHandlerFactory = new ErrorHandlerFactory();
 
 	public SearchIntegratorBuilder configuration(SearchConfiguration configuration) {
 		this.cfg = configuration;
@@ -287,7 +286,7 @@ public class SearchIntegratorBuilder {
 					)
 			);
 			factoryState.setAllIndexesManager( new IndexManagerHolder() );
-			factoryState.setErrorHandler( createErrorHandler( cfg ) );
+			factoryState.setErrorHandler( errorHandlerFactory.createErrorHandler( cfg ) );
 			factoryState.setInstanceInitializer( cfg.getInstanceInitializer() );
 			factoryState.setTimingSource( new DefaultTimingSource() );
 			factoryState.setIndexMetadataComplete( cfg.isIndexMetadataComplete() );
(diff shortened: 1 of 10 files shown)


engine/src/main/java/org/hibernate/search/cfg/spi/SearchConfiguration.java
```diff
@@ -130,4 +131,6 @@
 	 * {@link org.hibernate.search.engine.service.spi.ServiceManager}
 	 */
 	ClassLoaderService getClassLoaderService();
+
+	ErrorHandler getErrorHandler();
```
Sanne: This will need a little bit of javadoc.

engine/src/main/java/org/hibernate/search/cfg/spi/SearchConfigurationBase.java
```diff
@@ -60,4 +72,46 @@ public boolean isIdProvidedImplicit() {
...
+
+	@Override
+	public ErrorHandler getErrorHandler() {
+		if ( errorHandler == null ) {
+			errorHandler = createErrorHandler();
+		}
+		return errorHandler;
+	}
+
+	private ErrorHandler createErrorHandler() {
```
Sanne: I'd prefer to keep the logic in `SearchConfiguration` implementations to the bare minimal, so that it's easy to create a different implementation.

Could you undo moving these methods, and simply have the `getErrorHandler()` of this implementation return null?

engine/src/main/java/org/hibernate/search/spi/SearchIntegratorBuilder.java
```diff
@@ -499,37 +497,6 @@ private static IndexingMode defineIndexingMode(SearchConfiguration cfg) {
 		return IndexingMode.fromExternalRepresentation( indexingStrategy );
 	}
 
-	private ErrorHandler createErrorHandler(SearchConfiguration searchConfiguration) {
-		Object configuredErrorHandler = searchConfiguration.getProperties().get( Environment.ERROR_HANDLER );
```
Sanne: we could simply invoke the `searchConfiguration.getErrorHandler()` here, and if it's not-null give priority to that one.

Ideally, if the ERROR_HANDLER property is also defined log a warning to let people know that it's going to be ignored.

As an alternative strategy you could convert it into a Service, which means you can treat it as a ProvidedService when you need to inject one from Infinispan.

engine/src/main/java/org/hibernate/search/cfg/spi/SearchConfiguration.java
gustavonalle: It's self explanatory, what do you have in mind?

engine/src/main/java/org/hibernate/search/spi/SearchIntegratorBuilder.java
gustavonalle: Let me detail my use case:

```
class MyErrorHandler {

  public MyErrorHandler(ErrorHandler configuredErrorHandler) {
      ...
  }
  ...
}
```

By moving the creation to the Configuration level, it allows me to decorate the configured ErrorHandler and return it.

Not sure about Services: would it create a single ErrorHandler per JVM? One ErrorHandler per ```SearchIntegrator``` is enough.

What about undoing everything and exposing the creation logic at SPI level ```org.hibernate.search.spi.ErrorHandlerCreator``` instead? I can simply re-use it as a building block for my custom ErrorHandler

engine/src/main/java/org/hibernate/search/cfg/spi/SearchConfiguration.java
Sanne: I _had_ in mind something like "return null unless you want to override it", but that depends on how we address your other points.

I know we don't document much, especially self-explanatory methods, but when it comes to SPI a little guidance is needed.

engine/src/main/java/org/hibernate/search/spi/SearchIntegratorBuilder.java
Sanne: Ok so you want to wrap whatever is being configured by the user. I had missed that.
[Services won't fit this either then - although a Service is 1 per integrator, unless you want to provide the same instance to each]

Do you need to handle some error internally in Infinispan, while still forward to the user other errors?

Sanne: Exposing the ErrorHandler creation as an SPI might be acceptable if you really must, but I'd prefer to avoid it.

gustavonalle: > although a Service is 1 per integrator

I was with the impression hsearch services were based on ```java.util.ServiceLoader``` and thus 1 instance only for across all SearchIntegrator...

> Do you need to handle some error internally in Infinispan, while still forward to the user other errors?

Yes. On Infinispan, during topology changes, there is a small time window where a lock obtention fails due to transient ownership changes, and my custom error handler catches that and redirect the work to the rightful owner.

Sanne:
Hi @gustavonalle sorry for the delay. I think we could go with your idea of exposing the method which builds the "standard" ErrorHandler; could you make a proposal about that?

Also, did you notice that you can inject the ErrorHandler instance in the configuration Properties?

gustavonalle:
@Sanne Updated

Sanne:
rebased and merged. Thanks!

PR #1219, opened 2016-11-15
HSEARCH-2468 method getShardIdentifiersForDeletion in ShardIdentifier…

Sanne:
…ProviderTemplate should not be final

 - https://hibernate.atlassian.net/browse/HSEARCH-2468


yrodiere:
Loogs good to me; merged. Thanks!

PR #1220, opened 2016-11-15
HSEARCH-2235 Added the ability to configure multiple Elasticsearch hosts

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2235

Unfortunately Jest does not handle failover out of the box. So I created [HSEARCH-2469](https://hibernate.atlassian.net/browse/HSEARCH-2469) as a follow-up.

The tests are far from perfect, but right now the elasticsearch-maven-plugin doesn't allow launching multiple Elasticsearch nodes in a single cluster, so we can't do much better.

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/ElasticsearchEnvironment.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/ElasticsearchEnvironment.java
@@ -29,12 +29,14 @@ public final class ElasticsearchEnvironment {
 	}
 
 	/**
-	 * Property for specifying the host name of the Elasticsearch server to connect to. Only a single server (which may
-	 * be part of a cluster) is supported at this point.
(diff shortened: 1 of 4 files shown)


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/ElasticsearchEnvironment.java
```diff
@@ -29,13 +29,15 @@
 	}
 
 	/**
-	 * Property for specifying the host name of the Elasticsearch server to connect to. Only a single server (which may
-	 * be part of a cluster) is supported at this point.
+	 * Property for specifying the host name of the Elasticsearch server to connect to.
```
Sanne: should we change these to plural forms? host nameS .. serverS

yrodiere: Right. Done.

Sanne:
thanks, merged!

PR #1221, opened 2016-11-16
HSEARCH-2324 Use Double.MAX_VALUE as the default distance in spatial queries

yrodiere:
…  (when missing the latitude and/or longitude)

https://hibernate.atlassian.net/browse/HSEARCH-2324

The issue was that distance computation worked differently with our custom spatial layer on Lucene than with Elasticsearch. It was solved by aligning our spatial layer on Elasticsearch's behavior, which seems more sensible. That required to detect missing longitudes/latitudes, which were previously silently replaced by `0`.

I also replaced some sentences in the docs: I had to re-read that several time before I understood it was about the behavior when doing distance projection or distance sort on non-spatial fields. Which essentially won't work. Which isn't surprising, so is it really worth mentioning... ?


engine/src/main/java/org/hibernate/search/spatial/impl/DistanceCollector.java
```diff
@@ -147,8 +152,8 @@ public void setScorer(Scorer scorer) throws IOException {
 		@Override
 		public void collect(int doc) throws IOException {
 			final int absolute = docBase + doc;
-			double lat = coordinate( latitudeValues, doc );
-			double lon = coordinate( longitudeValues, doc );
+			Double lat = coordinate( docsWithLatitude, latitudeValues, doc );
```
Sanne: rather than changing the `coordinate` method, which is forcing you to return a `Double` rather than a `double`, you could check for this document to actually have both coordinates in advance, and if not skip the collection.

engine/src/main/java/org/hibernate/search/spatial/impl/DistanceFilter.java
```diff
@@ -106,14 +106,16 @@ public DocIdSet getDocIdSet(LeafReaderContext context, Bits acceptDocs) throws I
...
 		final NumericDocValues latitudeValues = DocValues.getNumeric( atomicReader, getLatitudeField() );
 		final NumericDocValues longitudeValues = DocValues.getNumeric( atomicReader, getLongitudeField() );
 
 		return new FilteredDocIdSet( docs ) {
 			@Override
 			protected boolean match(int documentIndex) {
-				double lat = coordinate( latitudeValues, documentIndex );
-				double lon = coordinate( longitudeValues, documentIndex );
+				Double lat = coordinate( docsWithLatitude, latitudeValues, documentIndex );
+				Double lon = coordinate( docsWithLongitude, longitudeValues, documentIndex );
```
Sanne: As in the `Collector`, I think we need to keep using primitives for each lat/lon value, and you can do so by checking in advance for the fields to exist in the `Bits` instance. If not, eagerly return false.

engine/src/main/java/org/hibernate/search/spatial/impl/Point.java
```diff
@@ -175,7 +175,12 @@ public double getDistanceTo(Point other) {
 	 * @return the distance between the points
 	 * @see <a href="http://www.movable-type.co.uk/scripts/latlong.html">Distance haversine formula</a>
 	 */
-	public double getDistanceTo(final double latitude,final double longitude) {
+	public double getDistanceTo(final Double latitude,final Double longitude) {
+		if ( latitude == null || longitude == null ) {
```
Sanne: In light of the other suggestions, maybe you can keep the original signature?
I don't like this one, as the invoker of the method should be responsible to not invoke this with null values, rather than the method implementation being responsibl to work around for such a case.

documentation/src/main/asciidoc/spatial.asciidoc
```diff
@@ -314,13 +314,8 @@ The default (field) name in case `@Spatial` is placed on the entity level is `or
...
 ====
-Using distance projection on non `@Spatial` enabled entities and/or with a non spatial Query will have
-unexpected results as entities not spatially indexed and/or having `null` values for latitude or
-longitude will be considered to be at (0,0)/(lat,0)/(0,long).
-
-Using distance projection with a spatial query on spatially indexed entities having, eventually,
-`null` values for latitude and/or longitude is safe as they will not be found by the spatial query
-and won't have distance calculated.
+When a spatial field on an entity has a `null` value for either its latitude or longitude (or both),
+the resulting projected distance will always be `Double.MAX_VALUE` (`1.7976931348623157e+308`).
```
yrodiere: Maybe... But we'd have to convert the `Double.MAX_VALUE` returned by Elasticsearch to `null`, in that case. I'll see if it can be done reliably. One thing that bothers me is I'm not sure I will be able to do the same with sorts (consider the value is null), and I'd rather have the projection and sorts be consistent.

documentation/src/main/asciidoc/spatial.asciidoc
```diff
@@ -367,13 +362,10 @@ repetition is needed to allow you to define Queries with any tool.
...
-unexpected results as entities non spatially indexed and/or with null values for latitude or
-longitude will be considered to be at (0,0)/(lat,0)/(0,long)
-
-Using distance sort with a spatial query on spatially indexed entities having, potentially, `null`
-values for latitude and/or longitude is safe as they will not be found by the spatial query and so
-won't be sorted
+When a spatial field on an entity has a `null` value for either its latitude or longitude (or both),
+the resulting distance will always be `Double.MAX_VALUE` (`1.7976931348623157e+308`),
+and the entity will be ranked accordingly:
+last if the sort is ascending, or first if the sort is descending.
```
yrodiere: Two things:

 1. The fact that a value is listed or not has nothing to do with sorts. It's related to distance constraints (distance queries), which are orthogonal to the sorts.
 2. *If* you also have a constraint on distance, the entities with null latitude/longitude will probably not appear in the search results, since `Double.MAX_VALUE` is larger than any max distance you might have set (except ``Double.MAX_VALUE`).

I'll update the doc to make that more obvious.

engine/src/main/java/org/hibernate/search/spatial/impl/Point.java
yrodiere: I'll see what I can do regarding those `Double`s.

yrodiere:
Here! All your concerns should have been addressed:

 * I only do as little double boxing as possible (see the last commit)
 * I added a commit (the last one) that changes the projection result for missing coordinates from `Double.MAX_VALUE` to `null`.

The sort behavior is unchanged: entities without coordinates are considered furthest by default. It seems to me that it should stay that way for now: it's more or less consistent with the previous behavior (which used the distance to (0,0), which for most coordinates is a large distance).
By the way, the fact that we now return null for projections is likely to trigger NPEs in some applications... Maybe we should postpone that last commit to Hibernate Search 6.0?

Sanne:
merged

PR #1222, opened 2016-11-17
HSEARCH-2360 and others: fixes around projections

yrodiere:
~This PR is based on #1217 (HSEARCH-2451), so we'll have to wait for that one to be merged first.~ => Done!

The purpose of this PR is:

 1. to use source filtering when projecting: [HSEARCH-2360](https://hibernate.atlassian.net/browse/HSEARCH-2360)
 2. to add support for projection on fields that are not mapped in Elasticsearch (aligning the behavior on the Lucene indexing service): [HSEARCH-2471](https://hibernate.atlassian.net/browse/HSEARCH-2471)
 3. to drop support for projection on fields that have only one-way bridges (aligning the behavior on the Lucene indexing service): [HSEARCH-2470](https://hibernate.atlassian.net/browse/HSEARCH-2470)

The main change is that, because of 1., metadata inspection for projections is now executed when building the query, and is not executed anymore when converting JSON hits to Java objects: the initial metadata inspection produces objects responsible for later data conversion that already take metadata into account.
The exact reason for this change is to avoid code duplication: since we now have to inspect metadata before running the query (to build the source filtering parameter), we may as well make the most of it.
One side-effect is that now, we only inspect metadata once for each field and for each queried type, whereas we used to inspect metadata once for each field and for each result. Which is not perfect, but arguably better.

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java
@@ -10,6 +10,7 @@ import java.io.IOException;
 import java.io.Serializable;
 import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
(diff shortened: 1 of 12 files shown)


yrodiere:
#1217 has just been merged, and I rebased this one. Ready for review!

Sanne:
Very nice, great job!

PR #1223, opened 2016-11-17
HSEARCH-2400 Support IndexingMonitor for the Elasticsearch backend

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2400

If you feel like method parameters in `BackendRequestSuccessHandler` are too much, I can remove them: they're not strictly necessary right now, but they're cheap and I think they would come handy if we make the `IndexMonitor` more complex one day.

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/client/impl/BulkRequest.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/client/impl/BulkRequest.java
@@ -8,12 +8,18 @@ package org.hibernate.search.elasticsearch.client.impl;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 
+import org.hibernate.search.backend.IndexingMonitor;
 import org.hibernate.search.backend.LuceneWork;
 import org.hibernate.search.exception.ErrorHandler;
 import org.hibernate.search.exception.impl.ErrorContextBuilder;
+
+import io.searchbox.core.BulkResult.BulkResultItem;
 
 /**
  * A request group backed by an actual bulk request.
@@ -38,7 +44,8 @@ public class BulkRequest implements ExecutableRequest {
 	 */
 	private final Set<String> indexesNeedingRefresh;
 
-	public BulkRequest(JestClient jestClient, ErrorHandler errorHandler, List<BackendRequest<?>> requests, Set<String> indexNames, Set<String> indexesNeedingRefresh, boolean refresh) {
+	public BulkRequest(JestClient jestClient, ErrorHandler errorHandler, List<BackendRequest<?>> requests,
+			Set<String> indexNames, Set<String> indexesNeedingRefresh, boolean refresh) {
 		this.jestClient = jestClient;
 		this.errorHandler = errorHandler;
 		this.requests = requests;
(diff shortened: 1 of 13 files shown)


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/client/impl/BulkRequest.java
```diff
@@ -78,6 +83,45 @@ public void execute() {
...
+	 * locally caught exception. Every locally caught exception that is not the "main" exception will be
+	 * added to the "main" exception.
+	 *
+	 * @param successfulItems The requests and their results, for calling result handler.
+	 * @param preexistingMainException The exception to use as a "main" exception (see above), or {@code null}.
+	 * @return The "main" exception (see above), or {@code null} if there isn't any.
+	 */
+	private RuntimeException callHandlers(Map<BackendRequest<?>, BulkResultItem> successfulItems, RuntimeException preexistingMainException) {
+		RuntimeException mainException = preexistingMainException;
+		for ( Map.Entry<BackendRequest<?>, BulkResultItem> entry : successfulItems.entrySet() ) {
```
Sanne: The `IndexingMonitor` can become a contention point when it's invoked too frequently, for example when running on servers with many CPUs, one `MassIndexer` job will use them all in parallel, and the pattern is relatively effective because it attempts to not share much; the monitor is one of the exceptions.

For this reason it takes a `long` argument, so that its counters can be incremented by values larger than one; so I would expect a Bulk Request not cause invoking the `IndexingMonitor#documentsAdded( 1L )` method N times, but invoke it with a paremeter N.

On the other hand, I like your idea of having the `handle` method to accept the `BulkResultItem` argument. 
Could you move this loop into the `BackendRequestSuccessHandler` implementation? There it can inspect the bulk result, and invoke the `IndexingMonitor` once as necessary.

yrodiere: I'll see what I can do...
But I'd like to point out that the only other uses of `IndexingMonitor` only increment it by 1. Which is why I implemented it that way: it sure didn't seem to be a concern with the Lucene indexing service, so I figured there was not reason it should be a concern with the Elasticsearch indexing service. Maybe we should have a look at the Lucene side too (`org.hibernate.search.backend.impl.lucene.works.AddWorkExecutor.performWork(LuceneWork, IndexWriterDelegate, IndexingMonitor)` for instance).

Sanne: > But I'd like to point out that the only other uses of IndexingMonitor only increment it by 1. 

ah, interesting! It's not how I originally intended it. Another thing to fix :)

yrodiere:
I changed my commit and rebased: this should be better.

This solution is a bit more twisted when it comes to calling the monitor, but otherwise it seems fine (it's even ready for when we'll add methods to `IndexMonitor`) and it should work just as well.

### Details:
I didn't move the loop to the handler, because the handler was specific to each backend request, and even the monitor it delegated to was specific to each backend request (theoretically, you may have two backend requests in the same bulk that have a different monitor).

Even if that wasn't an issue, tying the handler to the bulk instead of tying it to the backend request would have forced me to include action-specific code in the handler (to distinguish actions that require incrementing the monitor, like `Update` from those that don't, like `Delete`).

Sanne:
nice, thanks for going the extra mile!

Sanne:
merged

PR #1228, opened 2016-11-23
HSEARCH-2426 Fixed Dynamic sharding on Elasticsearch

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2426

This is a minimal fix.

Eventually, we'll have to introduce the concept of `IndexingService`, but it'll be a big change, so I planned that in another ticket: https://hibernate.atlassian.net/browse/HSEARCH-2485


Sanne:
great! merged this, after re-disabling the `DirectoryProviderForQueryTest` test as it depends on HSEARCH-2477 as well.

PR #1229, opened 2016-11-24
AsyncExecutor blocks forever on shutdown

Sanne:
https://hibernate.atlassian.net/browse/HSEARCH-2442

I could not reproduce the exact issue as reported, but there clearly was some racy code here, it wasn't taking into account concurrent lifecycle events of the Executor, in parallel with usage of it.

Introduced some order, I hope this resolves the issue.

N.B. we could have done some more research but it doesn't seem worth it in light of https://hibernate.atlassian.net/browse/HSEARCH-2340 , which I expect to require a rewrite of how these Executor's lifecycle is handled.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/backend/impl/lucene/LuceneBackendResources.java
+++ b/engine/src/main/java/org/hibernate/search/backend/impl/lucene/LuceneBackendResources.java
@@ -7,8 +7,6 @@
 package org.hibernate.search.backend.impl.lucene;
 
 import java.util.Properties;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock.ReadLock;
@@ -22,9 +20,6 @@ import org.hibernate.search.indexes.impl.PropertiesParseHelper;
 import org.hibernate.search.indexes.spi.DirectoryBasedIndexManager;
 import org.hibernate.search.indexes.spi.IndexManager;
 import org.hibernate.search.spi.WorkerBuildContext;
-import org.hibernate.search.util.impl.Executors;
-import org.hibernate.search.util.logging.impl.Log;
-import org.hibernate.search.util.logging.impl.LoggerFactory;
 
 /**
  * Collects all resources needed to apply changes to one index,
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/hibernate/search/backend/impl/lucene/LuceneBackendResources.java
```diff
@@ -66,28 +87,42 @@ private LuceneBackendResources(LuceneBackendResources previous) {
 		this.workspace = previous.workspace;
 		this.maxQueueLength = previous.maxQueueLength;
 		this.asyncIndexingExecutor = previous.asyncIndexingExecutor;
```
yrodiere: Shouldn't this be protected by locking `previous.executorStateReadLock`? Since the attribute is no longer `volatile`, that seems necessary... ?

Also, I wonder if the whole executor + readLock + writeLock stuff shouldn't be encapsulated in a separate object... What I find dangerous is that just after we copied the executor, one could call `shutdown` on a previous `LuceneBackendResources`, which would shut down the executor but would not clean up the `asyncIndexingExecutor` attribute on any new `LuceneBackendResources` that happened to share the same executor. And then we would have tasks being submitted but never handled.

Having a single, shared object hold the references to the locks would *and* to the executor would probably solve the issue. Then, when starting or shutting down the executor, we wouldn't change an attribute of `LuceneBackendResources`, but an attribute of that single, shared object that happens to be referenced from `LuceneBackendResources`, which is slightly different.

Or maybe there's another mechanism that guarantees that we'll never call a method on a previous version of the resources after having created a new version?

engine/src/main/java/org/hibernate/search/backend/impl/lucene/LuceneBackendResources.java
```diff
@@ -43,20 +43,41 @@
...
 	private final IndexManager indexManager;
 
-	private final ReadLock readLock;
-	private final WriteLock writeLock;
+	/**
+	 * Externally exposed Read/Write locks used by FSMasterDirectoryProvider
+	 * to be able to make copies of a locked (immutable) index.
+	 * TODO: explore if that could use a better snapshotting technique
+	 */
+	private final ReadLock indexmodificationReadLock;
```
yrodiere: Maybe `indexModificationReadLock` would be better (with a capital 'M')?

Sanne:
pushed a new version.

yrodiere:
Merged. I simply removed an obsolete commit in `LazyExecutorHolder`:

```
			//N.B. this method needs to be invoked within
			//ownership of the executorStateWriteLock.
```

(the code changed and the lock is now acquired inside the method)

Thanks!

Sanne:
thanks @yrodiere ! don't forget to mark the JIRA issues resolved ;)

PR #1224, opened 2016-11-22
HSEARCH-2406 Execute -engine tests against Elasticsearch too

yrodiere:
~This PR is based on  #1222 (HSEARCH-2360), which will have to be merged first.~ => Done!

Relevant tickets:
 * https://hibernate.atlassian.net/browse/HSEARCH-2406
 * https://hibernate.atlassian.net/browse/HSEARCH-2392 (because the fix was also needed for HSEARCH-2406)

Nothing much to say, except that some bugs were uncovered and will have to be fixed (see the linked tickets in [HSEARCH-2406](https://hibernate.atlassian.net/browse/HSEARCH-2406)). There weren't many of them, though :)

PR #1230, opened 2016-11-24
HSEARCH-2480 Adding classes to the SearchIntegrator concurrently fails with Elasticsearch

yrodiere:
~~This PR is based on #1224 (HSEARCH-2406), which will have to be merged first.~~ => Done

Relevant ticket: https://hibernate.atlassian.net/browse/HSEARCH-2480

It appears that the issue was valid, but my analysis was completely wrong.

The actual issue was that whenever the search factory changed (due to a new class mapping being added), every existing `ElasticsearchIndexManager` would reinitialize as if we were just starting Hibernate Search. Which means in particular that if we had chosen the `RECREATE_DELETE` schema management strategy, every Elasticsearch index would be deleted and re-created. Which includes indexes containing pre-existing entities whose mapping didn't change at all. Which is quite annoying...

### DIFF
--- a/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
+++ b/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
@@ -118,9 +118,9 @@ Let's see the options for the `index_schema_management_strategy` property:
 |Value|Definition
 |`NONE`|Indexes and their mappings will not be created, deleted nor altered.
 |`VALIDATE`|Existing indexes and mappings will be checked for conflicts with Hibernate Search's metamodel. Indexes and their mappings will not be created, deleted nor altered.
-|`MERGE`|Missing indexes and mappings will be created, mappings will be updated if there are no conflicts.
-|`CREATE`|**The default**: existing indexes will not be altered, missing indexes will be created.
-|`RECREATE`|Indexes will be deleted if existing and then created. This will delete all content from the indexes!
+|`MERGE`|Missing indexes and mappings will be created, existing mappings will be updated if there are no conflicts.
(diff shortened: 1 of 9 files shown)


documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -121,6 +121,9 @@ Let's see the options for the `index_schema_management_strategy` property:
 |`MERGE`|Missing indexes and mappings will be created, existing mappings will be updated if there are no conflicts.
 |`CREATE`|**The default**: existing indexes will not be altered, missing indexes will be created along with their mappings.
 |`RECREATE`|Indexes will be deleted if existing and then created along with their mappings. This will delete all content from the indexes!
++
+Note that whenever a search factory is altered after initialization (i.e. new entities are mapped),
+the index will *not* be deleted again: new mappings will simply be added to the index.
```
Sanne: N.B. that for the typical Hibernate Search user there is no such thing as "adding new entities at runtime". This capability is an implementation detail of the Infinispan integration, and possibly of other integrators (beyond the Hibernate ORM integration).

This documentation is specific for the Hibernate Search / Hibernate ORM integration, so to not make it all too confusing for them I'd avoid such notes.

I'll edit this commit to remove this specific change, then merge it all.. I hope you don't mind?

Sanne:
merged, thanks!

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
yrodiere: Not at all :) Thanks!

PR #1231, opened 2016-11-24
HSEARCH-2239 Update project tagline and first-impact project descript…

Sanne:
…ions

This addresses the README.
Next we'll have:
 - website - will need to be a subsequent PR based on this
 - documentation - I'll review that as a separate issue.

https://hibernate.atlassian.net/browse/HSEARCH-2239

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,16 +4,30 @@
 
 ## Description
 
-Full text search engines like Apache Lucene are very powerful technologies to add efficient free
-text search capabilities to applications. However, Lucene suffers several mismatches when dealing
-with object domain models. Amongst other things indexes have to be kept up to date and mismatches
-between index structure and domain model as well as query mismatches have to be avoided.
+Full text search for Java objects
 
-Hibernate Search addresses these shortcomings - it indexes your domain model with the help of a few
-annotations, takes care of database/index synchronization and brings back regular managed objects
-from free text queries.
+This project provides synchronization between entities managed by Hibernate ORM and full-text
+indexing services like Apache Lucene and Elasticsearch.
 
-Hibernate Search is using [Apache Lucene](http://lucene.apache.org/) under the cover.
+It will automatically apply changes to indexes, which is tedious and error prone coding work,
+while leaving you full control on the query aspects. The development community constantly
+researches and refines the index writing techniques to improve performance.
+
+Mapping your objects to the indexes is declarative, using a combination of Hibernate Search
+specific annotations and the knowledge it can gather from your existing Hibernate/JPA
+mapping.
+
+Queries can be defined by any combination of:
+ - "native" Apache Lucene queries
+ - writing "native" Elasticsearch queries in Json format (if using Elasticsearch, which is optional)
+ - using our DSL which abstracts the previous two generating optimal backend specific queries
+
+Query results can include projections to be loaded directly from the index, or can materialize
+fully managed Hibernate entities loaded from the database within the current transactional scope.
+
+Hibernate Search is using [Apache Lucene](http://lucene.apache.org/) under the cover; this
+can be used directly (running embedded in the same JVM) or remotely provided by Elasticsearch
+over its REST API.
 
 ## Requirements
 
@@ -48,8 +62,8 @@ installation directory. Then read the documentation available in *docs/reference
 
 #### Build options (profiles and properties)
 
-The documentation is based on [AsciiDoctor](http://asciidoctor.org/). Per default only the html
-output is enabled. To also generate the docbok output and build the documentation from there use:
+The documentation is based on [AsciiDoctor](http://asciidoctor.org/). By default only the html
+output is enabled; to also generate the docbok output use:
 
     > mvn clean install -Pdocbook -s settings-example.xml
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,16 +4,30 @@
...
-Hibernate Search addresses these shortcomings - it indexes your domain model with the help of a few
-annotations, takes care of database/index synchronization and brings back regular managed objects
-from free text queries.
+This project provides synchronization between entities managed by Hibernate ORM and full-text
+indexing services like Apache Lucene and Elasticsearch.
 
-Hibernate Search is using [Apache Lucene](http://lucene.apache.org/) under the cover.
+It will automatically apply changes to indexes, which is tedious and error prone coding work,
+while leaving you full control on the query aspects. The development community constantly
+researches and refines performance improvements to apply the writes at optimal performance.
```
yrodiere: There's a repeatition of "performance". Maybe "apply the writes optimally"?

README.md
```diff
@@ -4,16 +4,30 @@
...
+
+Mapping your objects to the indexes is declarative, using a combination of Hibernate Search
+specific annotations and the knowledge it can gather from your existing Hibernate/JPA
+mapping.
+
+Queries can be defined by any combination of:
+ - "native" Apache Lucene queries
+ - writing "native" Elasticsearch queries in Json format
```
yrodiere: Should state "for the Elasticsearch indexing service only", maybe?

README.md
```diff
@@ -4,16 +4,30 @@
...
+ - using our DSL which abstracts the previous two generating optimal backend specific queries
+
+Query results can use projections to be loaded directly from the index, or can materialize
```
yrodiere: I'd say "Queries can use...", or rephrase the whole sentence to basically state "Query results can include ...".

But "Query results can use" seems wrong: results don't "use" anything...

README.md
```diff
@@ -4,16 +4,30 @@
...
+fully managed Hibernate entities loaded from the database within the current transactional scope.
+
+Hibernate Search is using [Apache Lucene](http://lucene.apache.org/) under the cover; this
+can be used directly (running embedded in the same JVM) or remotely provided by Elasticsearch
+over its REST API.
```
yrodiere: Maybe this should be merged to the first paragraph? It seems we're repeating ourselves here.

Also, we can add a link to Elasticsearch: https://www.elastic.co/products/elasticsearch

README.md
```diff
@@ -76,8 +87,8 @@ A list of available versions for `testElasticsearchVersion` can be found on
 
 ### Contributing
 
-If you want to contribute, you find all you need to know in
-[Contributing to Hibernate Search](http://community.jboss.org/wiki/ContributingtoHibernateSearch)
+New contributors are always welcome. We collected some helpful hints on how to get started
+on our website at [Contribute to Hibernate Search](http://hibernate.org/search/contribute/)
 
 ### Source code structure
```
yrodiere: The modules "sharedtestresources", "legacy" and "infinispan" are not mentioned. Shouldn't we at least mention that those modules are private and/or obsolete and shouldn't be referenced directly from user projects?

README.md
```diff
@@ -91,14 +102,16 @@ The project is split in several Maven modules:
...
-* _integrationtest_: Integration tests with various technologies like WildFly, Spring or Karaf.
+* _integrationtest_: Integration tests with various technologies like WildFly, Spring and Karaf.
 Also includes performance tests.
 
 * _modules_: Integration with [WildFly](http://www.wildfly.org/) using JBoss Modules.
 
-* _orm_: Native integration for [Hibernate ORM](http://hibernate.org/orm/).
+* _orm_: Native integration for [Hibernate ORM](http://hibernate.org/orm/), and also home of most public API code.
 
 * serialization: Serialization code used by remote backends.
```
yrodiere: Not caused by your changes, I think this should be in italic; same with "test" below.

README.md
Sanne: thanks. rewrote this paragraph.

README.md
Sanne: I was hoping to keep it short, but ok I see that this might be necessary to be qualified..

README.md
Sanne: I need the first paragraph to be short. Here I'm indeed repeating the concept, as a closing point.. but  the real message is to make people aware that even if they opt for using ES it's still Lucene under the covers.

README.md
Sanne: There's a lot of stuff not being explained in the README, and whatever we write gets obsolete quickly. I'd rather not add details about stuff that occasional contributors are safe to ignore, the longer readme becomes unnecessarily scary.

Sanne:
Updated

yrodiere:
Merged, thanks!

README.md
```diff
@@ -4,16 +4,30 @@
...
+Full text search for Java objects
 
-Hibernate Search addresses these shortcomings - it indexes your domain model with the help of a few
-annotations, takes care of database/index synchronization and brings back regular managed objects
-from free text queries.
+This project provides synchronization between entities managed by Hibernate ORM and full-text
+indexing services like Apache Lucene and Elasticsearch.
 
-Hibernate Search is using [Apache Lucene](http://lucene.apache.org/) under the cover.
+It will automatically apply changes to indexes, which is tedious and error prone coding work,
```
gunnarmorling: "which is"="saving you from"

README.md
gunnarmorling: Why is it though that you are stressing this point (using Lucene under the covers)? Seems this should be more an internal detail of ES itself?

PR #1232, opened 2016-11-24
HSEARCH-2423 Raise an exception when trying to project on an unstored field

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2423

Opening this PR for advice... After solving HSEARCH-2423, I'm not 100% sure we should do it. Not even 10%, actually. Could someone confirm it's not a good idea?

The issue: with Lucene we require fields to be stored in order to be able to project them. With Elasticsearch, we don't.

We don't in Elasticsearch because we don't really need it: we retrieve the source JSON from the index to project values, so we don't really care that Elasticsearch also stores the field value besides that.

The argument I made in the ticket was that we don't know if we'll keep projecting that way indefinitely. If we release 5.6 and _then_ notice that there's a big issue that will force us to use field storage for projections, we'll have to forbid projections on non-stored fields. Which is an API break...

But then, forcing users to have their fields stored whenever they project them means forcing them to use more disk space for their indexes, for no practical reason since field storage is not used in Hibernate Search projections.

Reminder: the only reason we use source JSON for projections right now is that Elasticsearch does not store any other information than the timestamp for dates, so projecting ZonedDateTime for instance is impossible with just stored fields. See https://hibernate.atlassian.net/browse/HSEARCH-2358

If we reject this PR, we'll have to update the documentation, though:

 * It  states that we retrieve the whole source document when projecting, which is not true anymore since [HSEARCH-2360](https://hibernate.atlassian.net/browse/HSEARCH-2360) (we use source filtering)
 * It implies that adding a `Store.YES` marker will change the projection strategy to use field storage, which is not true (we can't use it, anyway, for the reasons I gave above).


Sanne:
Good analysis, thanks for that. I agree with you on not supporting to project a field which wasn't explicitly marked for "projection capabilities" (Store=yes for the lucene backend).

Storing the source document in ES is optional. I'm not sure if we want to keep indexing it, at the very least we should not rely on it to be stored so that we might introduce an option to skip it, as the storage and performance penalty might be unbearable for some.

With Lucene we also allow to project unknown fields as strings, but we already agreed that this legacy behaviour should not apply to ES.

You mention needing docValues as alternative. Why? I lost you from there onwards.

yrodiere:
> You mention needing docValues as alternative. Why? I lost you from there onwards.

It was a mistake, I meant field storage. I updated my original post.

> Good analysis, thanks for that. I agree with you on not supporting to project a field which wasn't explicitly marked for "projection capabilities" (Store=yes for the lucene backend).
> Storing the source document in ES is optional. I'm not sure if we want to keep indexing it, at the very least we should not rely on it to be stored so that we might introduce an option to skip it, as the storage and performance penalty might be unbearable for some.

I agree not relying on the source document is far from ideal, but right now it's the only way to support projections properly (as I explained in the paragraph starting with "Reminder:"). But yes, we could introduce a global parameter allowing users to use field storage for projections, and document the limitations.

> With Lucene we also allow to project unknown fields as strings, but we already agreed that this legacy behaviour should not apply to ES.

Right. Though the case at hand is slightly different: we're talking of known fields, with known metadata, but with storage disabled (store = NO). Currently those are being projected anyway (because we don't use field storage), while on Lucene an exception is raised.

Sanne:
I'll close this as it needs at least to update the PR metadata.

PR #1234, opened 2016-11-28
HSEARCH-2448 (+ HSEARCH-2419) Throw exceptions with clear messages when there are field type conflicts with Elasticsearch	

yrodiere:
Relevant tickets:

 1. https://hibernate.atlassian.net/browse/HSEARCH-2448: Throw exceptions with clear messages when there are field type conflicts with Elasticsearch
 2. https://hibernate.atlassian.net/browse/HSEARCH-2419 Using array/iterable/map field bridges requires the @IndexedEmbedded annotation

About the second ticket: we used to force users to use `@IndexedEmbedded` along with `@Field` to correctly index containers (iterables, maps, arrays) of primitive types. On top of being weird (we're not really embedding another `@Indexed` type in that case), this started to not work at all once we fixed the first ticket, which was (more or less) about preventing that kind of use (`@IndexedEmbedded` + `@Field` on the same property).

Please note that the second commit is not just here to ease up debugging: without it, some tests cannot be written at all, because metadata order is not predictable.

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
@@ -28,6 +28,7 @@ import org.hibernate.search.bridge.FieldBridge;
 import org.hibernate.search.bridge.TwoWayFieldBridge;
 import org.hibernate.search.elasticsearch.client.impl.BackendRequest;
 import org.hibernate.search.elasticsearch.impl.NestingMarker.NestingPathComponent;
+import org.hibernate.search.elasticsearch.logging.impl.Log;
 import org.hibernate.search.elasticsearch.util.impl.FieldHelper;
 import org.hibernate.search.elasticsearch.util.impl.FieldHelper.ExtendedFieldType;
 import org.hibernate.search.engine.ProjectionConstants;
@@ -39,7 +40,9 @@ import org.hibernate.search.engine.spi.DocumentBuilderIndexedEntity;
 import org.hibernate.search.engine.spi.EntityIndexBinding;
 import org.hibernate.search.exception.AssertionFailure;
 import org.hibernate.search.spatial.impl.SpatialHelper;
+import org.hibernate.search.util.logging.impl.LoggerFactory;
 
+import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonPrimitive;
 
@@ -57,6 +60,8 @@ import io.searchbox.indices.Optimize;
  * @author Gunnar Morling
  */
 class ElasticsearchIndexWorkVisitor implements IndexWorkVisitor<IndexingMonitor, BackendRequest<?>> {
+
+	private static final Log LOG = LoggerFactory.make( Log.class );
 
 	private static final String DELETE_ALL_QUERY = "{ \"query\" : { \"constant_score\" : { \"filter\" : { \"match_all\" : { } } } } }";
 	private static final String DELETE_ALL_FOR_TENANT_QUERY = "{ \"query\" : { \"constant_score\" : { \"filter\" : { \"term\" : { \"" + DocumentBuilderIndexedEntity.TENANT_ID_FIELDNAME + "\" : \"%s\" } } } } }";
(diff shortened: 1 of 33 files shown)


DavideD:
Merged https://github.com/hibernate/hibernate-search/pull/1234#issue-192005213

Thanks

DavideD:
I suppose we need to update the documentation as well in regards to the `@IndexEmbedded` not need anymore

Sanne:
I had to add a commit, probably because of the different merged PRs now interacting:
 - https://github.com/hibernate/hibernate-search/commit/efd2830bb838d4f763975667b36769bbe40db705

yrodiere:
@DavideD The container fields are not documented. There's a ticket to add the documentation: https://hibernate.atlassian.net/browse/HSEARCH-1574

@Sanne Right... Probably because of HSEARCH-2406, which enabled -engine tests on Elasticsearch

yrodiere:
@Sanne @Davide About the maps/sets implementations in the metadata ( https://github.com/hibernate/hibernate-search/commit/133c54294cd6cc0c07101ca357478cbb39f8781f ).

The thing is, we used to use HashMap/HashSet, whose iteration order is notoriously unstable (it might change when you add new elements). But this iteration order is important: it determines in which order fields are added to a document, or in which order fields are added to the Elasticsearch mapping. And in some cases (most notably indexing in Elasticsearch), the order in which fields are added to the JSON document may change the code that is executed.

And so it happens that, in https://github.com/hibernate/hibernate-search/commit/e5382aaced0d9d9e4ba0361e3473a58428f88936, I fixed two aspects of the same bug. As a matter of fact, there is two places where we might change the JSON document when indexing:

 * [explicitly](https://github.com/hibernate/hibernate-search/commit/e5382aaced0d9d9e4ba0361e3473a58428f88936#diff-5c25139bd711d1dae0c48d6d1fa00536R65): when we add a property value to the JSON document. The `add` method seems a bit odd, but it's only so that we can support multi-valued fields,even though this information that a field is multi-valued is not part of our metadata.
 * [implicitly](https://github.com/hibernate/hibernate-search/commit/e5382aaced0d9d9e4ba0361e3473a58428f88936#diff-5c25139bd711d1dae0c48d6d1fa00536R79): when we implicitly create parent properties. For instance when we add a value for the property `foo.bar`, then `foo` has to be added to the JSON document as an object property, so we can set `bar` on it. This is done under the hood, but it is a document modification. And it does not follow the same rules as the "explicit" change, since multi-values are not at all supported in this case.

Thus, since there are two different places in the code where we change a document, there are two places where a field naming conflict could be detected. Which is why we have two error cases:

 1. when, for the the same field name, first an explicit change is performed (`add` sets a "concrete" value such as an integer, a date, etc. for field `foo`), and then an implicit one (`getOrCreate` tries to set a composite value, i.e. a JSON object for field `foo` in order to add the field `foo.bar`): https://github.com/hibernate/hibernate-search/commit/36cee0230910c124192c983b5e68e4215bb71ffd#diff-92dcdf2c51f954af5389442b38ba3de7R176
 2. when the changes happen in reverse order (first the implicit `getOrCreate`, then the explicit `add`): https://github.com/hibernate/hibernate-search/commit/36cee0230910c124192c983b5e68e4215bb71ffd#diff-92dcdf2c51f954af5389442b38ba3de7R208

Now, one can understand that if the iteration order over the fields is unpredictable, I can hardly test those two cases, because I won't have any guarantee on which field is added to the JSON document first, and so I won't be able to ensure I am in case 1 or in case 2.

That's where the map/set implementation come in play: by using `LinkedHashSet` and `LinkedHashMap`, I ensure that the metadata order is the same as the processing order in the `AnnotationMetadataProcessor`. Thus the metadata order becomes predictable to some extent. Not totally predictable, since we still rely on the order of the fields/methods returned by the JVM (which is undefined and not even guaranteed to be deterministic), but partially predictable: I can be sure, for instance, that `@Fields` on methods will appear in the metadata before `@Field` on attributes, because methods are processed first.

On a related subject, I don't find the reliance on a particular map/set implementation _that_ smelly. At least, not anymore than Java is smelly: the contract of `Set` and `Map` does not mention that the iteration order should be stable, so it's up to the implementation. I would gladly use the `StableSet` and `StableMap` interfaces, but there is no such thing.
`SortedSet` and `SortedMap` are not exactly what we want, because their only implementations (`TreeSet` and `TreeMap`) require a total order on the elements we add to these containers, and in this case we don't have one. Given two fields, we could order them by their field name, but what if two different `DocumentFieldMetadata` have the same name? They're not identical, but there's no way to say which one is "greater". And yes, this can happen, most notably [when class bridges are defined without a field name](https://github.com/hibernate/hibernate-search/commit/fc304a32133d07829eb542a7e85e4fe3b1c438bb).

PR #1235, opened 2016-11-28
HSEARCH-2491 Flush per entity ignored for ElasticsearchIndexManager

gustavonalle:
https://hibernate.atlassian.net/browse/HSEARCH-2491

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
@@ -8,13 +8,12 @@ package org.hibernate.search.elasticsearch.impl;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Properties;
 import java.util.Set;
 
-import org.apache.lucene.analysis.Analyzer;
-import org.apache.lucene.search.similarities.Similarity;
 import org.hibernate.search.backend.FlushLuceneWork;
 import org.hibernate.search.backend.IndexingMonitor;
 import org.hibernate.search.backend.LuceneWork;
@@ -45,6 +44,9 @@ import org.hibernate.search.indexes.spi.ReaderProvider;
 import org.hibernate.search.spi.WorkerBuildContext;
 import org.hibernate.search.util.configuration.impl.ConfigurationParseHelper;
 import org.hibernate.search.util.logging.impl.LoggerFactory;
+
+import org.apache.lucene.analysis.Analyzer;
+import org.apache.lucene.search.similarities.Similarity;
 
 /**
  * An {@link IndexManager} applying indexing work to an Elasticsearch server.
@@ -378,14 +380,23 @@ public class ElasticsearchIndexManager implements IndexManager {
 
 	@Override
 	public void performStreamOperation(LuceneWork singleOperation, IndexingMonitor monitor, boolean forceAsync) {
-		if ( singleOperation == FlushLuceneWork.INSTANCE ) {
+		BackendRequest<?> request = singleOperation.acceptIndexWorkVisitor( visitor, monitor );
+		if ( singleOperation instanceof FlushLuceneWork ) {
 			requestProcessor.awaitAsyncProcessingCompletion();
+			executeBackendRequest( request, true );
 		}
 		else {
-			BackendRequest<?> request = singleOperation.acceptIndexWorkVisitor( visitor, monitor );
+			executeBackendRequest( request, false );
+		}
+	}
 
-			if ( request != null ) {
-				requestProcessor.executeAsync( request );
+	private void executeBackendRequest(BackendRequest<?> backendRequest, boolean sync) {
+		if ( backendRequest != null ) {
+			if ( sync ) {
+				requestProcessor.executeSync( Collections.<BackendRequest<?>>singletonList( backendRequest ) );
+			}
+			else {
+				requestProcessor.executeAsync( backendRequest );
 			}
 		}
 	}
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
@@ -10,6 +10,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Set;
 
+import io.searchbox.indices.Flush;
 import org.apache.lucene.document.Document;
 import org.apache.lucene.facet.FacetsConfig;
 import org.apache.lucene.index.DocValuesType;
@@ -137,8 +138,18 @@ class ElasticsearchIndexWorkVisitor implements IndexWorkVisitor<IndexingMonitor,
 
 	@Override
 	public BackendRequest<?> visitFlushWork(FlushLuceneWork work, IndexingMonitor monitor) {
-		// Nothing to do
-		return null;
+		Flush flush = new Flush.Builder().setParameter( "wait_if_ongoing", "true" )
+				.addIndex( indexName )
+				.refresh( true )
+				.build();
+		return new BackendRequest<>(
+				flush,
+				work,
+				indexName,
+				monitor,
+				NoopBackendRequestSuccessHandler.INSTANCE,
+				false
+		);
 	}
 
 	@Override
--- a/elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/ElasticsearchFlushIT.java
+++ b/elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/ElasticsearchFlushIT.java
@@ -1,0 +1,145 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.elasticsearch.test;
+
+import java.io.Serializable;
+import java.util.Collections;
+import java.util.Locale;
+
+import org.hibernate.search.annotations.Field;
+import org.hibernate.search.annotations.Indexed;
+import org.hibernate.search.backend.LuceneWork;
+import org.hibernate.search.bridge.util.impl.ContextualExceptionBridgeHelper;
+import org.hibernate.search.cfg.Environment;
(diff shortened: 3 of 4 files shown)


elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/ElasticSearchFlushIT.java
```diff
@@ -0,0 +1,149 @@
...
+import org.hibernate.search.testsupport.junit.SearchFactoryHolder;
+
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+
+import io.searchbox.indices.settings.UpdateSettings;
+import org.apache.lucene.search.MatchAllDocsQuery;
+
+public class ElasticSearchFlushIT {
```
yrodiere: Could you please use `Elasticsearch` (no capital "S"), just to be consistent with other tests?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
```diff
@@ -132,8 +133,15 @@ public ElasticsearchIndexWorkVisitor(String indexName, boolean refreshAfterWrite
 
 	@Override
 	public BackendRequest<?> visitFlushWork(FlushLuceneWork work, IndexingMonitor monitor) {
-		// Nothing to do
-		return null;
+		Flush flush = new Flush.Builder().addIndex( indexName ).refresh( true ).build();
```
yrodiere: From what I can read in [the documentation](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/indices-flush.html), since we're not setting `wait_if_ongoing` to `true` here, the operation will fail if a flush operation is in progress. I'd say it's not what we want, especially given that we're already executing the operation asynchronously and thus waiting a bit more is no big deal.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -293,15 +297,19 @@ public void performOperations(List<LuceneWork> workList, IndexingMonitor monitor
...
 
-			if ( request != null ) {
+		BackendRequest<?> request = singleOperation.acceptIndexWorkVisitor( visitor, monitor );
+
+		if ( request != null ) {
+			if ( forceAsync ) {
 				requestProcessor.executeAsync( request );
 			}
+			else {
+				requestProcessor.executeSync( Collections.<BackendRequest<?>>singletonList( request ) );
```
yrodiere: So, here, we're switching the default behavior (when `forceAsync` is `false`) from asynchronous to synchronous.

From what I can see in `org.hibernate.search.indexes.spi.IndexManager.performStreamOperation(LuceneWork, IndexingMonitor, boolean)`, the behavior when `forceAsync` is `false` "will depend on the backend configuration". So I guess we could make it async by default if we wanted to.

I must admit I don't quite grasp the implications of that change, though. The whole concept of being able to execute "stream" operations asynchronously or synchronously seems rather blurry to me, so I must be missing something.

@Sanne , what do you think?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -293,15 +297,19 @@ public void performOperations(List<LuceneWork> workList, IndexingMonitor monitor
 
 	@Override
 	public void performStreamOperation(LuceneWork singleOperation, IndexingMonitor monitor, boolean forceAsync) {
-		if ( singleOperation == FlushLuceneWork.INSTANCE ) {
+		if ( singleOperation instanceof FlushLuceneWork ) {
```
yrodiere: Nice catch!

gustavonalle:
Rebased and addressed (most of the) reviews.

Sanne:
Looks like I have to explain a bit of the intent behind _performStreamOperation_, sorry for the lack of clear documentation.

Once upon a time, the MassIndexer would use its own dedicated IndexWriter. This implies that when the MassIndexer was running, any other change to the index would be blocked in wait for the index lock to be released.
The IndexWriter backend of the MassIndexer was different than the one being used by the "transactional backend": it would apply changes in parallel (no ordering guarantees) and no flushing between writes, while the traditional one would apply a "block" (a batch?) of changes at a time, to maintain strict order, and commit the IW at the end of each block.

We introduced the _performStreamOperation_ as a way to "sneak by" the changes required by the MassIndexer on the same IW already (possibly) open by the standard IW backend, bypassing the "one change at a time" requirement of the normal backend while reusing the same IW instance.
In practice this needed (and needs!) some care, as now we had two incoming streams of changes on the same IW, with subtly different semantics. N.B. it is not an option to simply apply the strongest required semantics, as otherwise the performance penalty is unacceptable (like taking 1 month rather than 1 minute to do certain operations).

The main flaw of the current design, is that the conceptual "streams" have no starting point and no end; the Flush command for example is needed by the MassIndexer to signal for the phase of "I've sent a bunch of stream commands, now I'm done so now I want you to finally flush to disk.." This is actually a very tricky point, as the previous "stream ops" being applied in parallel need to be strictly ordered all before the flush.

I hope that clarifies the *original* intent of the _performStreamOperation_ method. We should likely rethink such things now, as it all looks quite obsolete when combined with many other changes we applied to the backend. For example I wonder if we should explicitly open/close "stream tasks", making the explicit Flush command unnecessary, or rather separate the high level flush concept from actual index commits, which should be a backend implementation detail.

On this PR now: 

My main question is why are you testing semantics of _performStreamOperation_ ? Is Infinispan using this for anything else than the MassIndexer?

Keep in mind that the MassIndexer is the only one which is meant to use this method, and it's assumed to never produce more than one entry for the same ID. Your test intentionally violates this property .. in other words, the _performStreamOperation_ might have an ill name and I'm sorry for that but it doesn't guarantee ordering within the "stream", as it expects a stream of *unrelated* data points.

Let's also keep https://hibernate.atlassian.net/browse/HSEARCH-2282 in mind!

gustavonalle:
| why are you testing semantics of performStreamOperation

Infinispan's MassIndexer uses a ```org.hibernate.search.backend.spi.BatchBackend``` that in turn ends up calling this method. The test just does the same.

| but it doesn't guarantee ordering within the "stream",

That's fine, currently we don't expect the stream to be ordered

gustavonalle:
| never produce more than one entry for the same ID. Your test intentionally violates this property

This is not a requirement of the test, I pushed a new version with different ids

gustavonalle:
@yrodiere removed the "sync by default", it sync only for the flush.
@Sanne changed the test to go through the ```BatchBackend``` instead of calling performStreamOperation directly

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -378,14 +380,23 @@ public void performOperations(List<LuceneWork> workList, IndexingMonitor monitor
...
 	public void performStreamOperation(LuceneWork singleOperation, IndexingMonitor monitor, boolean forceAsync) {
-		if ( singleOperation == FlushLuceneWork.INSTANCE ) {
+		BackendRequest<?> request = singleOperation.acceptIndexWorkVisitor( visitor, monitor );
+		if ( singleOperation instanceof FlushLuceneWork ) {
 			requestProcessor.awaitAsyncProcessingCompletion();
+			executeBackendRequest( request, true );
 		}
 		else {
-			BackendRequest<?> request = singleOperation.acceptIndexWorkVisitor( visitor, monitor );
+			executeBackendRequest( request, false );
```
yrodiere: The second parameter should be `!forceAsync` instead of `false`, I think. Otherwise `forceAsync` is simply ignored.

gustavonalle: Short story:

I left it ```false``` to maintain the behaviour currently in place: all work is async by default, so ```forceAsync``` is not used as before.

Long Story:

This is what the contract says about this flag:
```
@param forceAsync if true, the invocation will not block to wait for it being applied. 
When false this will depend on the backend configuration.
```

In the ```DirectotyBasedIndexManager``` there is a configuration ```worker.execution``` which assumes values ```{sync,async}``` that defines how works are applied to the index. In theory, the ```forceAsync``` flag should have no effect if the backend is already async. So assuming ```worker.execution=async``` and ```forceAsync=false```, what should happen? Having ```!forceAsync``` would make it sync... :disappointed: 

You may be asking why does it matter since the ```ElasticsearchIndexManager``` does not honor ```worker.execution```?  Should it honor it? I realized answering those question and implementing missing pieces were not directly related to this PR (which is about flush per entity) and maybe should be addressed in another JIRA, wdyt?

yrodiere: You're right, I got confused. Obviously if it's async by default, then forcing async shouldn't do anything... Sorry about that.

About `worker.execution`:

 1. It only applies to non-stream works (from what I can see in the code), so in our case it's not relevant
 2. Support is indeed missing for Elasticsearch. I added a ticket on JIRA so we can at least discuss it: https://hibernate.atlassian.net/browse/HSEARCH-2502

elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/ElasticsearchFlushIT.java
```diff
@@ -0,0 +1,145 @@
...
+		Entity1 entity1 = new Entity1( 1 );
+		Entity2 entity2 = new Entity2( 1 );
+
+		indexAsStream( "E1:1", entity1 );
+		indexAsStream( "E2:1", entity2 );
+
+		assertIndexed( Entity1.class, 0 );
+		assertIndexed( Entity2.class, 0 );
+
+		flush( Entity1.class );
```
Sanne: I think there's a race here: the _indexAsStream_ operations are merely scheduling work into a queue so that they can be batched with periodic writes. The _flush_ operation is issuing a flush request on the ES node, but it's not flushing this queue first.

I'll merge it anyway as it doesn't seem to affect this test in practice, and there's good progress here..

Sanne:
Merged! Thanks.

As commented above, I still suspect the reproducer might eventually fail with the wrong timings, but we'll see if it comes to that.

Also, I don't think raising the refresh timeout makes any difference?

elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/ElasticsearchFlushIT.java
gustavonalle: @Sanne  ```ElasticsearchIndexManager#performStreamOperation``` is waiting for the async works to be applied to the index

PR #1236, opened 2016-11-30
HSEARCH-2486 Enable @ContainedIn on classes whose only link to Search…

yrodiere:
… is one of their superclasses

https://hibernate.atlassian.net/browse/HSEARCH-2486

Those classes were being ignored until now, because we were checking
whether classes are relevant to Hibernate Search without taking into
account their superclasses, thereby skipping metadata building entirely.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/util/impl/ReflectionHelper.java
+++ b/engine/src/main/java/org/hibernate/search/util/impl/ReflectionHelper.java
@@ -115,13 +115,41 @@ public abstract class ReflectionHelper {
 	}
 
 	/**
-	 * Checks whether the specified class contains any Search specific annotations.
+	 * Checks whether the specified class contains any Search-specific annotations.
+	 *
+	 * <p>This method will return {@code true} if such an annotation is detected
+	 * in the class itself or in one of its superclass, either applied directly on the class,
+	 * on a field, or on a method.
 	 *
 	 * @param mappedClass the {@code XClass} to check for Search annotations
 	 *
 	 * @return Returns {@code true} if the class contains at least one Search annotation, {@code false} otherwise
 	 */
 	public static boolean containsSearchAnnotations(XClass mappedClass) {
+		List<XClass> hierarchy = createXClassHierarchy( mappedClass );
+
+		for ( XClass clazz : hierarchy ) {
+			if ( containsLocalSearchAnnotation( clazz ) ) {
+				return true;
+			}
+		}
+
+		return false;
+	}
+
+	/**
+	 * Checks whether the specified class contains any locally-declared, Search-specific annotations.
+	 *
+	 * <p>This method will return {@code true} if such an annotation is detected
+	 * in the class itself, either applied directly on the class, on a field, or on a method.
+	 * <p>Only fields and methods declared by the specified class are taken into account:
+	 * inherited fields and methods are ignored.
+	 *
+	 * @param mappedClass the {@code XClass} to check for Search annotations
+	 *
+	 * @return Returns {@code true} if the class contains at least one Search annotation, {@code false} otherwise
+	 */
+	private static boolean containsLocalSearchAnnotation(XClass mappedClass) {
 		// check the type annotations
 		if ( containsSearchAnnotation( mappedClass.getAnnotations() ) ) {
 			return true;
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/hibernate/search/util/impl/ReflectionHelper.java
```diff
@@ -115,13 +115,34 @@ public static Object getMemberValue(Object bean, XMember getter) {
...
-	 * Checks whether the specified class contains any Search specific annotations.
+	 * Checks whether the specified class contains any Search-specific annotations, i.e. if
+	 * the class itself or one of its superclasses carries such an annotation.
 	 *
 	 * @param mappedClass the {@code XClass} to check for Search annotations
 	 *
 	 * @return Returns {@code true} if the class contains at least one Search annotation, {@code false} otherwise
 	 */
 	public static boolean containsSearchAnnotations(XClass mappedClass) {
+		List<XClass> hierarchy = ReflectionHelper.createXClassHierarchy( mappedClass );
```
gsmet: Any reason why you added `ReflectionHelper.`? Might as well remove it so that we see it's a method of this very class?

engine/src/main/java/org/hibernate/search/util/impl/ReflectionHelper.java
```diff
@@ -115,13 +115,34 @@ public static Object getMemberValue(Object bean, XMember getter) {
...
+
+		for ( XClass clazz : hierarchy ) {
```
gsmet: I'm wondering if it would be worth it to exclude `java.lang.Object`? Sounds like some useless work. Using the programmatic API to declare Search annotations on it would be weird to say the least.

But maybe, it's safer to also take it into account.

WDYT?

engine/src/main/java/org/hibernate/search/util/impl/ReflectionHelper.java
```diff
@@ -115,13 +115,34 @@ public static Object getMemberValue(Object bean, XMember getter) {
 	}
 
 	/**
-	 * Checks whether the specified class contains any Search specific annotations.
+	 * Checks whether the specified class contains any Search-specific annotations, i.e. if
+	 * the class itself or one of its superclasses carries such an annotation.
```
gsmet: Wondering if `carries` is the right word here. As stated here, I wouldn't be sure the annotations on methods and properties are considered. It might be just me so feel free to ignore :).

engine/src/main/java/org/hibernate/search/util/impl/ReflectionHelper.java
yrodiere: I agree it would be weird (total nonsense, even) to index something on `Object`, at least in production. I guess we could imagine someone indexing `toString` for debugging purposes in a development environment. Though it'd still be weird.

Anyway, I don't think the number of classes inspected here is that important. We're talking about one execution per class in the hierarchy of each non-`@Indexed` Hibernate classes, and only when boostrapping the Search factory. Which is really not much work, especially since it does not involve I/O.

If we were talking about runtime code, why not, but in this case I find the weirdness of excluding `Object` more annoying than the performance hit (!) of inspecting the `Object` class.

engine/src/main/java/org/hibernate/search/util/impl/ReflectionHelper.java
yrodiere: No particular reason, it was copy/paste I think. I'll remove it.

engine/src/main/java/org/hibernate/search/util/impl/ReflectionHelper.java
yrodiere: "Contains" seemed especially fuzzy to me. It's okay for a method name, but here in the javadoc I wouldn't want any ambiguity. I'll try something else...

engine/src/main/java/org/hibernate/search/util/impl/ReflectionHelper.java
gsmet: Yup, that was my overall feeling as well but I felt better mentioning it.

gsmet:
Thanks, merged!

Backporting it to 5.5.

PR #1237, opened 2016-11-30
Hsearch 2494 tika bridge bug

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-2494

The issue is caused by to problems:

1. ExtendedBridgeProviderContext#getReturnType returns only the contained type for arrays and collections but the TikaBridgeProvider expects the complete type.

2. TikaBridge provider should overrides the default bridge we use for arrays.

I've also added a test to check the support for URI.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/bridge/impl/ExtendedBridgeProvider.java
+++ b/engine/src/main/java/org/hibernate/search/bridge/impl/ExtendedBridgeProvider.java
@@ -58,5 +58,13 @@ public abstract class ExtendedBridgeProvider implements BridgeProvider {
 		 * not.
 		 */
 		boolean isExplicitlyMarkedAsNumeric();
+
+		/**
+		 * Returns the type of the indexed member/property; it works for arrays and collections too.
+		 *
+		 * @see #getReturnType()
+		 * @return the type of the indexed member
+		 */
+		Class<?> getElementOrContainerReturnType();
 	}
 }
--- a/engine/src/main/java/org/hibernate/search/bridge/impl/BridgeFactory.java
+++ b/engine/src/main/java/org/hibernate/search/bridge/impl/BridgeFactory.java
@@ -26,6 +26,7 @@ import org.hibernate.search.bridge.FieldBridge;
 import org.hibernate.search.bridge.ParameterizedBridge;
 import org.hibernate.search.bridge.TwoWayFieldBridge;
 import org.hibernate.search.bridge.TwoWayStringBridge;
+import org.hibernate.search.bridge.builtin.TikaBridge;
 import org.hibernate.search.bridge.builtin.impl.BuiltinArrayBridge;
 import org.hibernate.search.bridge.builtin.impl.BuiltinIterableBridge;
 import org.hibernate.search.bridge.builtin.impl.BuiltinMapBridge;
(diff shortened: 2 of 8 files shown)


yrodiere:
@DavideD It seems checkstyle is not happy. Could you please fix the [code formatting errors](http://ci.hibernate.org/job/hibernate-search-PR/1760/org.hibernate$hibernate-search-orm/console)?

DavideD:
@yrodiere I think it is fixed now

engine/src/main/java/org/hibernate/search/bridge/impl/ExtendedBridgeProvider.java
```diff
@@ -58,5 +58,13 @@ public FieldBridge provideFieldBridge(BridgeProviderContext bridgeProviderContex
...
 		 */
 		boolean isExplicitlyMarkedAsNumeric();
+
+		/**
+		 * Returns the type of the indexed member/property; it works for arrays and collections too.
+		 *
+		 * @see #getReturnType()
+		 * @return the type of the indexed member
+		 */
+		Class<?> getExactReturnType();
```
yrodiere: I know BridgeProviderContext is SPI and can't be changed, but this naming scheme scares me. It feels like `getReturnType` is somewhat random (not "exact")...

I see two solutions:

 * we could deprecate the old `getReturnType` and use more appropriate naming, like for instance `getPropertyType` and `getPropertyElementType`
 * or we could change the behavior of this method slightly, rename it `getContainerType` and make it return null when the property is not a container.

What do you think?

orm/src/test/java/org/hibernate/search/test/bridge/tika/TikaBridgeSupportTest.java
```diff
@@ -38,7 +38,7 @@
 /**
  * @author Hardy Ferentschik
  */
-public class TikaBridgeBlobSupportTest extends SearchTestBase {
+public class TikaBridgeSupportTest extends SearchTestBase {
```
yrodiere: Did you see there already are *two* classes named `TikaBridgeTest`? Maybe introducing a third class with a name this generic is not a good idea...
What would you think of "TikaBridgeInputFormatTest"?

yrodiere:
@DavideD I have some doubts related to naming, see above. Apart from that, it looks good!

orm/src/test/java/org/hibernate/search/test/bridge/tika/TikaBridgeSupportTest.java
DavideD: This class is called `TikaBridge[Support]Test`,  but I'll change it

engine/src/main/java/org/hibernate/search/bridge/impl/ExtendedBridgeProvider.java
DavideD: I prefer the `getProperty[Element]Type` approach, I don't like the `getContainerType` one because sometimes it returns null and force me to check.

But I wanted to keep the patch simple without renaming other existing SPI.

Somewhere in the code I think I saw something like `getElementOrContainerType()`, would that be better?

yrodiere: It's probably a good compromise until 6.0, yes. `getElementOrContainerReturnType`, then?

DavideD:
Renamed the way you suggested

yrodiere:
Rebased and merged. Thanks!

yrodiere:
Backported to 5.5.

PR #1238, opened 2016-12-01
HSEARCH-2258 Normalized names conflict

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-2258

### DIFF
--- a/engine/src/main/java/org/hibernate/search/util/logging/impl/Log.java
+++ b/engine/src/main/java/org/hibernate/search/util/logging/impl/Log.java
@@ -1000,4 +1000,7 @@ public interface Log extends BasicLogger {
 			+ " Could not parse '%1$s'." )
 	SearchException nullMarkerInvalidFormat(String proposedTokenValue, String fieldName, String invalidFormatMessage, @Cause Exception e);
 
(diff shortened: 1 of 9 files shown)


engine/src/main/java/org/hibernate/search/util/logging/impl/Log.java
```diff
@@ -1000,4 +1000,7 @@
 	@Message(id = 324, value = "The fieldBridge for field '%1$s' is an instance of '%2$s', which does not implement TwoWayFieldBridge. Projected fields must have a TwoWayFieldBridge.")
 	SearchException projectingFieldWithoutTwoWayFieldBridge(String fieldName, Class<?> fieldBridgeClass);
 
+	@Message(id = 325, value = "Index names collision detected. Different logical indexes will point to the same actual index: %1$s")
```
yrodiere: I couldn't find any place where we use the terms "logical indexes". Ok with you if I replace that with "index managers"?

DavideD:
Updated and rebased

yrodiere:
Jenkins, test this please

yrodiere:
Rebased and merged. Thanks!

PR #1239, opened 2016-12-02
Upgrading ORM version, and deadlock with MariaDB tests resolved

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2379
 - https://hibernate.atlassian.net/browse/HSEARCH-2499


Sanne:
N.B. the ORM upgrade is not going to the 5.7 branch (neither now nor later).

For the test fix, I have a better solution too coming which requires Java 8 (also restricted to 5.7 only)

A preview is here:
 96ad5f727b189ecc3d2b89cae61589d805847703

yrodiere:
Rebased and merged, thanks!

Sanne:
thanks, I'll re-enable the mariadb tests on ci.hibernate.org

Sanne:
thanks, I'll re-enable the mariadb tests on ci.hibernate.org

PR #1240, opened 2016-12-02
HSEARCH-2415 Extend indexNullAs type checking to potentially any field bridge

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2415

The issue is that, with Elasticsearch, some indexed string values must have a specific format in order to be indexed. For instance, dates must follow the ISO-8601 format.

We don't check if that format is followed on `Field.indexNullAs` for string fields, so badly formatted indexNullAs strings are hard to debug. Which is why we added type checking in [HSEARCH-2262](https://hibernate.atlassian.net/browse/HSEARCH-2262).

But I missed java 8 time types... And those are not easy to implement, because each one has its own very specific format.

This PR fixes the issue by making indexNullAs parsing a bridge responsibility. It makes sense because, currently, it's the field bridge that is responsible for encoding values in the exact format required by the indexing service. This responsibility [might move away from field bridges in the future](https://hibernate.atlassian.net/browse/HSEARCH-2462), but probably not before 6.0.


engine/src/main/java/org/hibernate/search/bridge/builtin/impl/NullEncodingTwoWayFieldBridge.java
```diff
@@ -58,6 +59,7 @@ public String objectToString(Object object) {
 		}
 	}
 
+	@Override
 	public TwoWayFieldBridge unwrap() {
```
Sanne: The "unwrap" method is typically implemented by letting the method return some internal component which matches the type of the parameter. That's for example the contract on `EntityManager`.

An example implementation is here:
 - org.hibernate.search.jpa.impl.FullTextEntityManagerImpl.unwrap(Class<T>)

Just wondering.. did you consider that? Did it not fit?

Your solution doesn't look bad, it's just that given the specific name I would expect it to follow the well known pattern.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/logging/impl/Log.java
```diff
@@ -159,14 +159,13 @@ BulkRequestFailedException elasticsearchBulkRequestFailed(String request, String
 	SearchException facetingRequestHasUnsupportedType(String facetingRequestType);
 
 	@Message(id = ES_BACKEND_MESSAGES_START_ID + 27,
-			value = "The 'indexNullAs' property for field '%2$s' needs to represent a Boolean to match the field type of the index. "
-					+ "Please change value from '%1$s' to represent a Boolean." )
-	SearchException nullMarkerNeedsToRepresentABoolean(String proposedTokenValue, String fieldName);
+			value = "The 'indexNullAs' property for Boolean fields must represent a Boolean ('true' or 'false')." )
+	IllegalArgumentException invalidNullMarkerForBoolean();
```
Sanne: When it comes to these exceptions I think that quoting the value which the user attempted to use is helpful to understand the error. I see the previous version of the error had it, why remove this?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/schema/impl/DefaultElasticsearchSchemaTranslator.java
```diff
@@ -56,17 +61,14 @@
...
 	@Override
 	public void start(Properties properties, BuildContext context) {
 		serviceManager = context.getServiceManager();
-		gsonService = serviceManager.requestService( GsonService.class );
 	}
 
 	@Override
 	public void stop() {
-		gsonService = null;
 		serviceManager.releaseService( GsonService.class );
```
Sanne: Since you no longer requested the service, you should not release it either.
Might as well not keep the reference to _serviceManager_ ...

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/schema/impl/DefaultElasticsearchSchemaTranslator.java
```diff
@@ -56,17 +61,14 @@
 	private static final Log LOG = LoggerFactory.make( Log.class );
 
 	private ServiceManager serviceManager;
-	private GsonService gsonService;
 
 	@Override
 	public void start(Properties properties, BuildContext context) {
 		serviceManager = context.getServiceManager();
```
Sanne: no longer needed?

engine/src/main/java/org/hibernate/search/util/logging/impl/Log.java
```diff
@@ -1000,4 +996,8 @@
 	@Message(id = 324, value = "The fieldBridge for field '%1$s' is an instance of '%2$s', which does not implement TwoWayFieldBridge. Projected fields must have a TwoWayFieldBridge.")
 	SearchException projectingFieldWithoutTwoWayFieldBridge(String fieldName, Class<?> fieldBridgeClass);
 
+	@Message(id = 325, value = "The 'indexNullAs' property for field '%2$s', with value '%1$s', has invalid format: %3$s"
+			+ " Could not parse '%1$s'." )
+	SearchException nullMarkerInvalidFormat(String proposedTokenValue, String fieldName, String invalidFormatMessage, @Cause Exception e);
```
Sanne: Ok, I understand now what you're doing with those messages..

engine/src/main/java/org/hibernate/search/bridge/builtin/impl/NullEncodingTwoWayFieldBridge.java
yrodiere: I did not consider it, you're right. The `unwrap()` method in `NullEncodingTwoWayFieldBridge` was already there and I just generalized its use. Maybe following the `unwrap(Class<?>)` pattern would be better.

I'll see what I can do, but it probably won't look perfect: we actually have two ways of unwrapping: one which goes through container bridges, the other that doesn't.

yrodiere:
@Sanne  I think I addressed your concerns.

I'm really not sure the `unwrap(Class<?>)` pattern fits here, though. It feels rather clumsy in our case, mostly because bridges have no common superinterface exposing the unwrap(?)` method, so we must rely on utils anyway whenever we want to unwrap bridges. This is true for client code (metadata processor and such), but also for the implementation of the `unwrap(Class<?>)` method itself... 

So if we have to use utils, we might as well make the `unwrap` code in bridges as simple as possible and stick to its previous form.

I'll let you judge of that: I added the `unwrap` method change in the very last commit, you can remove it easily if you agree with me.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/schema/impl/DefaultElasticsearchSchemaTranslator.java
yrodiere: Right, I missed that. Fixed, thanks!

yrodiere:
Jenkins, test this please

Sanne:
thanks for that research @yrodiere ! I see your point, yet I'd prefer to take the change so that other people maintaining the code won't be surprised by the contract.

I've added one commit to make your function simpler, I hope it makes you feel better about that ;-) 87afa4f

Sanne:
(rebased and merged)

PR #1241, opened 2016-12-02
HSEARCH-2172 Non-null properties persisted as null in ES if value equals indexNullAs()

yrodiere:
~~This PR is based on #1240 , which we'll have to merge first.~~ => Done!

Relevant ticket: https://hibernate.atlassian.net/browse/HSEARCH-2172


yrodiere:
Rebased just now. Ready for review!

Sanne:
merged as well! Thanks

PR #1242, opened 2016-12-02
HSEARCH-2499 fix version using lambdas

Sanne:


orm/src/test/java/org/hibernate/search/test/util/ResourceCleanupFunctions.java
```diff
@@ -0,0 +1,71 @@
...
+	public static void withinTransaction(EntityManager em, Runnable codeblock) {
+		EntityTransaction transaction = em.getTransaction();
+		transaction.begin();
+		try {
+			codeblock.run();
+			transaction.commit();
+		}
+		catch (Throwable t) {
+			if ( transaction.isActive() ) {
+				transaction.rollback();
```
yrodiere: Same as above, I think this should something like that:

```
catch (Throwable t1) {
	try {
		if ( transaction.isActive() ) {
			transaction.rollback();
		}
	}
	catch(Throwable t2) {
		someLogger.warn("Error while automatically rolling back a transaction", e); // Maybe we could remove that
		t1.addSuppressed(t2);
	}
	throw t1;
}
```

orm/src/test/java/org/hibernate/search/test/util/ResourceCleanupFunctions.java
```diff
@@ -0,0 +1,71 @@
...
+	 */
+	public static void withinEntityManager(EntityManagerFactory emf, Consumer<FullTextEntityManager> consumer) {
+		EntityManager entityManager = emf.createEntityManager();
+		try {
+			FullTextEntityManager fem = Search.getFullTextEntityManager( entityManager );
+			consumer.accept( fem );
+		}
+		finally {
+			entityManager.close();
+		}
```
yrodiere: I think we should handle exceptions with more caution. Instead of `finally`, this would be:

```
catch (Throwable t1) {
	try {
		entityManager.close();
	}
	catch(Throwable t2) {
		someLogger.warn("Error while automatically closing an entity manager", e); // Maybe we could remove that
		t1.addSuppressed(t2);
	}
	throw t1;
}
entityManager.close(); // Let any exception when closing propagate, since we didn't catch any exception before
```

orm/src/test/java/org/hibernate/search/test/util/ResourceCleanupFunctions.java
Sanne: it's a test helper. Either way, they will end up on our logs?

orm/src/test/java/org/hibernate/search/test/util/ResourceCleanupFunctions.java
Sanne: What do you gain with that? If anything fails, the test fails and we'll have both exceptions logged.

yrodiere: Maybe in that case... Though I'm not sure that exceptions get logged when they are bypassed due to another exception being thrown in a finally block.
Before Java 7, people advised to explicitly log those, but it might be because the default behavior simply sends the stacktrace to `System.err`.
Since Java 7, I must admit I don't know what happens in that case.

In the helper below, though, I'm positive the Throwable t gets swallowed and never appears in logs if an exception arises when rolling back.

Sorry to be pestering, but exceptions being swallowed in tests, especially assertion failures, are a real pain. And since we get to solve the issue in a single, widely used helper, I don't see why we wouldn't...

Sanne: ok, pushed an update. I checked how the "suppressed exception" looks like, it's quite nice and  readable under the original exception so we shouldn't need the log.

yrodiere:
Thanks, merged!

PR #1243, opened 2016-12-02
HSEARCH-2348 Improve logging for elasticsearch integration

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2348

The pretty-printing requires an ugly hack, but I guess it's not that bad, since we only execute it when query logging is active.

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java
@@ -74,8 +74,10 @@ import org.hibernate.search.spatial.DistanceSortField;
 import org.hibernate.search.util.configuration.impl.ConfigurationParseHelper;
 import org.hibernate.search.util.impl.CollectionHelper;
 import org.hibernate.search.util.impl.ReflectionHelper;
+import org.hibernate.search.util.logging.impl.LogCategory;
 import org.hibernate.search.util.logging.impl.LoggerFactory;
 
+import com.google.gson.Gson;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
@@ -100,6 +102,7 @@ public class ElasticsearchHSQueryImpl extends AbstractHSQuery {
 	private static final JsonParser JSON_PARSER = new JsonParser();
 
 	private static final Log LOG = LoggerFactory.make( Log.class );
+	private static final Log QUERY_LOG = LoggerFactory.make( Log.class, LogCategory.QUERY );
 
 	private static final Pattern DOT = Pattern.compile( "\\." );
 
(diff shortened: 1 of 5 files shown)


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java
```diff
@@ -694,6 +696,16 @@ private SearchResult search(boolean enableScrolling) {
 				}
 			}
 			Search search = builder.build();
+
+			if ( QUERY_LOG.isDebugEnabled() ) {
+				try ( ServiceReference<GsonService> gsonService =
+						extendedIntegrator.getServiceManager().requestReference( GsonService.class ) ) {
```
yrodiere: I don't think I can:

 * those queries are created in `org.hibernate.search.elasticsearch.impl.ElasticsearchJsonQueryDescriptor.createHSQuery(SearchIntegrator)`, and that method is generic SPI. We could change it, but we certainly couldn't (shouldn't) add Elasticsearch-specific parameters.
 * the query descriptors themselves may be created from public static methods (such as `org.hibernate.search.elasticsearch.ElasticsearchQueries.fromJson(String)`), making it impossible to pass a reference to the service when creating the query descriptor.

What I can do is:

 * request the reference only once in the `ElasticsearchHSQuery` constructor, which is not much better since queries are generally only executed once or twice. 
 * replace the GsonService by static methods, but I did exactly the opposite a few months ago, so that would be a shame.
 * maybe see why requesting a reference to a service is such a performance issue? Especially given we have a cache...

Also, I'd like to mention two facts:

 1. We already request a reference to a service each time we create a query through `org.hibernate.search.engine.impl.ImmutableSearchFactory.createHSQuery(Query, Class<?>...)` (provided Elasticsearch is being used)
 2. This code is only executed in debug mode, which I'd expect to be less efficient than normal mode.

What do you think?

yrodiere: @Sanne I just had a look at the service manager code, and I guess the efficiency issue you hinted at was with those synchronized start/stop methods in `org.hibernate.search.engine.service.impl.StandardServiceManager.ServiceWrapper`.

Couldn't we use plain `java.util.concurrent.locks.Lock`s under the hood instead of our own counters here? That way, we'd have a read lock (that would ensure nobody is going to start/stop a service while it's being used) and a write lock (that would ensure we're free to start/stop a service without blocking anyone else). I was thinking of something similar to what you did in `org.hibernate.search.backend.impl.lucene.LazyExecutorHolder`.

yrodiere: > Think in terms of limits: what would happen on a single threaded app? For every single query, we'd be starting the service, then stopping it again.

Ok, that not what I had in mind. It's true in the general case, though even in a single threaded app, `GsonService` is likely to be referenced elsewhere, so it likely wouldn't be started for each query. It is for instance referenced by `org.hibernate.search.elasticsearch.client.impl.JestClient`, which is referenced (indirectly) by `ElasticsearchIndexManager`, whose lifespan is large enough to make this start/stop dance quite rare.

Anyway... We're both willing to avoid this potential performance issue here. I'll simply repeat what I said yesterday: in that particular case, there's no single, non-service object created when bootstrapping that could hold the reference to this service and pass it along to queries at runtime. I would _like_ to do so, but it's simply not possible in the current state of the code. Or at least I'm unable to see how it's possible. Which is why I'm trying to think of alternative solutions...

I created a ticket to address potential performance issues in the `ServiceManager` for those few cases where we are forced to repeatedly acquire reference to services. We should avoid it, I agree, but there may be cases where we can't do otherwise. See https://hibernate.atlassian.net/browse/HSEARCH-2504

yrodiere:
Rebased on latest master.

Sanne:
merged! thanks

PR #1244, opened 2016-12-02
HSEARCH-2500 TikaBridge not working with container types

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-2500

I sort of solved this issue while thinking how to remove the 'instanceof TikaBridge' from `BridgeFactory`

The code should be simple but some bridges might need a way to tell the factory when byte[] should be treated as an array of byte elements or as a whole. So I added an interface.

I hope it makes sense


yrodiere:
@DavideD Hi! I thought about the TikaBridge a bit this week-end, and I think we have another way out for the container type issues with TikaBridge, without adding yet another bridge interface.
My concern with adding new bridge interfaces is that for every new interface we add, the extent of what a bridge can implement becomes less and less obvious to a newcomer. I know type introspection is useful (and I've done it myself in other PRs), but if we can avoid it, maybe we should?

So, the solution: currently when a return type is a container, we only do one pass on bridge providers, passing a context with getReturnType() returning the element type. What if we did *two* passes? One where the context would hold the container type (a getReturnType method that returns the container type, and a getReturnElementType method that returns the element type), and a second pass that where we would act as usual, with getReturnType returning the element type. We would only do bridge-wrapping with our container wrappers in the second case. This would probably be backward-compatible,  because pre-existing bridge providers would give up on the first pass ("I don't support Collection"), and everything would be like it used to be on the second pass.

What do you think?

yrodiere:
@DavideD Actually the interface has the added benefit of also working on field bridges created because of explicit annotations (`@FieldBridge(impl = MyContainerBridge.class)`). So it's probably better. I'll wait for your opinion before merging...

By the way, I think that fixing [HSEARCH-2419](https://hibernate.atlassian.net/browse/HSEARCH-2419) has probably made the problem experienced with `TikaBridge` even broader.
The `TikaBridge` issue was with bridge providers getting the wrong metadata, but with HSEARCH-2419 fixed, the issue will also be there at runtime: people that designed custom field bridges for collections will have their bridge wrapped in an `IterableBridge` (for instance) without them knowing, which will probably result in their field throwing a `ClassCastException` when indexing (they'll be passed an element instead of the whole collection).
Introducing the new `ContainerAwareBridge` interface offers a solution to users experiencing this issue, but I wonder how we should advertise this interface, especially because it'll be needed in 5.6 while it wasn't in 5.5...

DavideD:
>  wonder how we should advertise this interface

I'm not sure I understand your remarks...

This interface overrides the default behaviours for containers bridges. The default behaviour is to index the elements of a container and if you instead want to index the whole container you can use this interface.

> especially because it'll be needed in 5.6 while it wasn't in 5.5...

That's not true, TikaBridge does not work in 5.5 with byte[] arrays anyway. You just have a different error. Adding this interface doesn't affect what was working on 5.5...unless I'm missing something.

DavideD:
To be fair, I didn't check if it works with 5.5 but given the nature of the problem I assume it doesn't work with it. I might be wrong

yrodiere:
@DavideD Let me clarify: there are two different issues.

The first is the one you're addressing, with bridge providers. Bridge providers were being passed a context that returns the element type, and gives no information about the actual type of objects that will be passed to the bridge in `set(String name, Object value, Document document, LuceneOptions options)`. That made the `TikaBridgeProvider` fail for `byte[]` because `byte` was passed, and it didn't support `byte`, only `byte[]`.

But there is another, nastier issue here.

We used to pass the element type instead of the actual property type because we usually wrap provided bridges within an `ArrayBridge`, `IterableBridge` and so on. That's what is done in `BridgeFactory#getFieldBridgeFromBridgeProvider`. So the bridge providers need not know about the actual container type, we handle it for them and only pass them the elements, one after another.

That the theory, but in practice it didn't work until we fixed HSEARCH-2419 (or at least not in every case: I'll spare you the details). So until we fixed HSEARCH-2419 , bridges applied on container types (arrays, iterables, maps) were actually being passed the container when indexing with `set(String name, Object value, Document document, LuceneOptions options)`. So there are users out there (I know some) that built field bridges that supported containers well before, and it used to work.

So... now that we fixed `HSEARCH-2419`, those bridges will now be wrapped in our container bridges (as they should have been from the beginning), and will now get passed container *elements* in `set(String name, Object value, Document document, LuceneOptions options)`. Since they're expecting containers, they'll fail when trying to cast the value to a container.

This means that their code that used in 5.5 (because HSEARCH-2419 hadn't been fixed) won't work in 5.6 anymore.

Your interface would solve this second issue, because custom bridges supporting containers would only have to implement this interface. But still, it requires users to do something, and that's why I wondered "how we should advertise this interface".

Is it clearer?

DavideD:
The FieldBridge contract is very vague about which value is going to be index, we should probably take the chance to decide the correct behaviour and add some tests for it. Users will need to update their bridges this time if they used a different behaviour.

That said, unless we still require the @IndexEmbedded annotation I don't see how we can know if the user wants to index the container or the element and still keep back-compatibility.  This would mean revert some of the changes and post-pone them after 5.6.

I would treat all this as a follow up issue, though. As it requires some additional discussion how to address it. In particular, I would expect to receive the whole object when I create a custom field bridge and not the element (as default).

I would treat all this as a follow up issue, though. As it requires some additional discussion how to address it. In particular, I would expect to have to specify when I want to index the elements instead of the

DavideD:
This is my 0.02$, @Sanne WDYT?

yrodiere:
Jenkins, test this please

yrodiere:
For your information, I'm working on a solution to https://hibernate.atlassian.net/browse/HSEARCH-2508 that may remove the need for this PR, because the `if (bridge instanceof TikaBridge)` may disappear.

Please do not merge until I get the chance to open a PR.

DavideD:
@yrodiere Can't you rebase your current work on top of this one? We still need to add the the missing test. Also probably the changes in the TikaBridgeProvider will still be needed.

Then you can send a PR that solves both issues

yrodiere:
@DavideD Right, there's more to this PR than just `ContainerAwareBridge`. Thanks for the tip. I'll take as much from this PR as possible. I only expect to remove the `ContainerAwareBridge` interface, which won't make sense anymore because the default will be not to wrap bridges in container bridges.
I'll keep you updated.

yrodiere:
@DavideD I cherry-picked most commits in my PR for HSEARCH-2508: #1253 , thanks!

I'm closing this one for now, since HSEARCH-2508 must be fixed for other reasons and makes the `ContainerAwareBridge` obsolete (container bridges are now only applied when explicitly requested by users). Feel free to reopen this PR if you disagree.

PR #1245, opened 2016-12-03
Removing outdated references to Solr Analyzer Framework

Sanne:
https://hibernate.atlassian.net/browse/HSEARCH-2223

This was actually much harder than it looks, thanks to HSEARCH-2501 ;-)


yrodiere:
Nice. Merged, thanks!

PR #1246, opened 2016-12-05
HSEARCH-2492 Added method to wait for async processing to finish

gustavonalle:
https://hibernate.atlassian.net/browse/HSEARCH-2492

PR #1247, opened 2016-12-05
HSEARCH-2492 Use an exclusive lock on the index writer when executing…

yrodiere:
… Flush works

Alternative solution to https://hibernate.atlassian.net/browse/HSEARCH-2492

See also https://github.com/hibernate/hibernate-search/pull/1246: we'll have to merge one or the other.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/backend/impl/lucene/LuceneBackendQueueTask.java
+++ b/engine/src/main/java/org/hibernate/search/backend/impl/lucene/LuceneBackendQueueTask.java
@@ -9,6 +9,7 @@ package org.hibernate.search.backend.impl.lucene;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.locks.Lock;
 
+import org.hibernate.search.backend.FlushLuceneWork;
 import org.hibernate.search.backend.IndexingMonitor;
 import org.hibernate.search.backend.LuceneWork;
 import org.hibernate.search.exception.impl.ErrorContextBuilder;
@@ -27,7 +28,8 @@ final class LuceneBackendQueueTask implements Runnable {
 
 	private static final Log log = LoggerFactory.make();
 
-	private final Lock modificationLock;
+	private final Lock exclusiveModificationLock;
+	private final Lock parallelModificationLock;
 	private final LuceneBackendResources resources;
 	private final Iterable<LuceneWork> workList;
 	private final IndexingMonitor monitor;
@@ -36,11 +38,23 @@ final class LuceneBackendQueueTask implements Runnable {
 		this.workList = workList;
 		this.resources = resources;
 		this.monitor = monitor;
-		this.modificationLock = resources.getParallelModificationLock();
+		this.exclusiveModificationLock = resources.getExclusiveModificationLock();
+		this.parallelModificationLock = resources.getParallelModificationLock();
 	}
 
 	@Override
 	public void run() {
+		boolean exclusiveLockRequired = false;
+		for ( LuceneWork work : workList ) {
+			if ( work instanceof FlushLuceneWork ) {
+				exclusiveLockRequired = true;
+				break;
+			}
+		}
+
+		Lock modificationLock =
+				exclusiveLockRequired ? exclusiveModificationLock : parallelModificationLock;
+
 		modificationLock.lock();
 		try {
 			applyUpdates();
--- a/engine/src/main/java/org/hibernate/search/backend/impl/lucene/LuceneBackendTaskStreamer.java
+++ b/engine/src/main/java/org/hibernate/search/backend/impl/lucene/LuceneBackendTaskStreamer.java
@@ -8,6 +8,7 @@ package org.hibernate.search.backend.impl.lucene;
 
 import java.util.concurrent.locks.Lock;
 
+import org.hibernate.search.backend.FlushLuceneWork;
 import org.hibernate.search.backend.IndexingMonitor;
 import org.hibernate.search.backend.LuceneWork;
 import org.hibernate.search.backend.impl.lucene.works.LuceneWorkExecutor;
@@ -29,16 +30,26 @@ final class LuceneBackendTaskStreamer {
 	private static final Log log = LoggerFactory.make();
 
 	private final LuceneBackendResources resources;
-	private final Lock modificationLock;
+	private final Lock exclusiveModificationLock;
+	private final Lock parallelModificationLock;
 	private final AbstractWorkspaceImpl workspace;
 
 	public LuceneBackendTaskStreamer(LuceneBackendResources resources) {
 		this.workspace = resources.getWorkspace();
-		this.modificationLock = resources.getParallelModificationLock();
+		this.exclusiveModificationLock = resources.getExclusiveModificationLock();
+		this.parallelModificationLock = resources.getParallelModificationLock();
 		this.resources = resources;
 	}
 
 	public void doWork(final LuceneWork work, final IndexingMonitor monitor) {
+		Lock modificationLock;
+		if ( work instanceof FlushLuceneWork ) {
+			modificationLock = exclusiveModificationLock;
+		}
+		else {
+			modificationLock = parallelModificationLock;
+		}
+
 		modificationLock.lock();
 		try {
 			IndexWriterDelegate delegate = workspace.getIndexWriterDelegate();


gustavonalle:
| See also #1246: we'll have to merge one or the other.

If I understood it correctly, those changes only affect the ```DirectoryBasedIndexManager```, right? 

At least from Infinispan point of view, there are no issues with the ```DirectoryBasedIndexManager``` at the moment since all the stream work is applied synchronously to the index

So even if this PR is merged, the issue of missing documents because the flush was called while async works were not applied to the index when using the ES would still remain, that's what #1246 addresses.

yrodiere:
> If I understood it correctly, those changes only affect the DirectoryBasedIndexManager, right?

@gustavonalle They do, but the approach is different.

You only have synchronization issues because you send indexing works from multiple nodes, but only flush from one node, so the flush may happen before some indexing works.
You said you had to do this because, with `DirectoryBasedIndexManager`, the flushes would mess with works being executed. I gathered there must be a synchronization issue then, and this PR solves that issue.

Theoretically, after this PR is merged, you'll be able to send a final flush from every node which added asynchronous indexing works, removing the need to explicitly wait for the works to finish executing before flushing.

Mind that the final flushes may have to be executed through `performOperations(List<LuceneWork>, IndexingMonitor`, according to the javadoc:

```
	 * Operations might be applied out-of-order! To mark two series of operations which need to be applied
	 * in order, use a transactional operation between them: a transactional operation will always flush
	 * all streaming operations first, and be applied before subsequent streaming operations.
```

gustavonalle:
| You said you had to do this because, with DirectoryBasedIndexManager, the flushes would mess with 
| works being executed

No, I wrote earlier the issue is only with the ```ElasticsearchIndexManager```. The "flush-once-after-indexing-from-multiple-nodes in stream fashion" strategy works fine for the ```DirectoryBasedIndexManager```, because it does not support async stream indexing.

yrodiere:
I'm talking about another issue we mentioned on the same ticket, the one that prevented you from sending a flush from every node.

Quoting:

My question:

> I suppose flushing from every node that used enqueueAsyncWork is out of the question?

Your answer: 

> This does not work well for the DirectoryIndexManager: it could lead to cascade failures since flush can be very expensive after inserting a large amount of data. That's why flush is carefully done once. For the Elasticsearch I haven't tested it, but I suppose it could go better since they prevent multiple flushes in parallel.

So what I propose is we prevent multiple flushes in parallel on the same index writer and be done with it.

gustavonalle:
| So what I propose is we prevent multiple flushes in parallel on the same index writer and be done with it.

Right, I overlooked the part where you suggested that on Infinispan we start flushing from every node instead of once after all nodes are done submitting their data.

Thinking about it, this can lead to other undesired issues, let me explain: your changes may prevent multiple expensive flushes at the same time (and thus bringing the whole cluster to its knees, which it's good) but still it'll execute multiple flushes. Imagine a cluster of 100 nodes, each node finishes submitting data at a different point in time (reasonable to assume that, data distribution may not be exactly identical across nodes) and then flushes: the flushes will not be executed in parallel but will happens nonetheless, 100 times. A node that finishes early and flushes will effectively slow down everyone else while it does its (in theory) non required flush, which can decrease throughput of the mass indexing processing.

yrodiere:
> A node that finishes early and flushes will effectively slow down everyone else while it does its (in theory) non required flush, which can decrease throughput of the mass indexing processing.

How I see it, this flush is not required now, but the data it flushes will have to be flushed eventually. And once it's been flushed, well it's that much less to flush at the end.
However, I understand what you mean about performance: worst case we will have 100 flushes, each for 1/100th of the data, which may prevent Lucene to perform some optimizations and may require some of its internal operations to be repeated needlessly. I'd expect a flush with less data to be faster, but 100 consecutive flushes with each 100 times less data may indeed be sub-optimal.

Bottom line: I honestly can't tell how badly this would impact performance without testing it. Let's see what @Sanne thinks about it. The fix might be relevant anyway, even if Infinispan doesn't switch to a "one-flush-per-node" strategy, because there really seems to be a synchronization issue (from what you said).

gustavonalle:
@yrodiere Agreed on the synchronization issue, so let's promote this PR as an addition to #1246 rather than a replacement for it.

Sanne:
Looks like I have to catch up reading on several fronts, so forgive me if I start with some dumb questions on this one:

I don't understand what problem is being resolved here. A test might be necessary?

There are patterns to prevent the 100 sequential flushes, see for example org.hibernate.search.indexes.impl.SharingBufferReaderProvider.refreshOperationId
It's unlikely to fold 100 flushes into one, but it should result into 2 to 3 operations (assuming all 100 flush operations are incoming in the start/finish period of the first flush).

gustavonalle:
@Sanne as long as Infinispan is concerned, there is only 1 flush being done at the moment, and there are no plans to start making more than one due to the reasons I mentioned above.

yrodiere:
> I don't understand what problem is being resolved here. A test might be necessary?

I'll try to do that today or tomorrow, but I'm not sure the issue will be easily reproductible.

Technically, this commit makes sure that while doing a flush, we don't try to write to the index. Which could be an issue since the "flush" actually commits the index writer, and after that we have to use a new index writer.

The link with HSEARCH-2492 is that HSEARCH-2492 is only relevant in the following scenario:

 * Purge indexes (once)
 * BatchBackend.enqueueAsyncWork (several times)
 * BatchBackend.flush (once)

... And the reason for the last step to be executed only once in a centralized fashion is (presumably) that if one node flushes while another is still indexing, there will be failures because the flush changes the index writer (sets it to null) and the indexing works use that same index writer.

Here, I'm making sure that even if one node flushes while another is still indexing, everything will go smoothly. Which will enable Infinispan to change its scenario to the following:

 * Purge indexes (once)
 * BatchBackend.enqueueAsyncWork (several times in parallel on several nodes)
 * BatchBackend.flush (once at the end, for each node)

And then HSEARCH-2492 is not relevant anymore. 

> There are patterns to prevent the 100 sequential flushes, see for example org.hibernate.search.indexes.impl.SharingBufferReaderProvider.refreshOperationId

I don't see how it's related to flushes. You mean we could apply the same pattern to fold multiple flushes into one?

Sanne:
> Technically, this commit makes sure that while doing a flush, we don't try to write to the index.

That's not a problem, a Lucene IndexWriter is threadsafe. Worst case we'll be including in the flush some changes which didn't *require* the flush: we don't guarantee that changes which don't have an explicit flush will not be flushed yet. For example the write buffers are bounded: when the buffers are full, a flush will be triggered automatically.

> Which could be an issue since the "flush" actually commits the index writer, and after that we have to use a new index writer.

We don't have to use a new IndexWriter after a flush - only some configuration actually demand this - yet I don't see the problem as that specific case is driven by +org.hibernate.search.backend.impl.lucene.SharedIndexCommitPolicy+ , which ensures the IW is not closed when it's being used by using reference counting. The one exception I can think of is how OOM exceptions, out of disk, and other critical exceptions are handled, in this case IW might need to be forcefully closed.

yrodiere:
@Sanne You're right, I didn't pay enough attention to the commit policies and especially to the fact that only `SharedIndexCommitPolicy` is used whenever an index writer is shared.

This means I don't know where the issue mentioned by @gustavonalle (the issue with flushing from every node) could come from. Which means the easiest way to solve HSEARCH-2492 is propably #1246.

Closing.

gustavonalle:
> This means I don't know where the issue mentioned by @gustavonalle (the issue with flushing from every node) could come from

The main issue is doing several flushes that can potentially affect performance of the whole mass indexing process. We do 1 because only 1 is needed really.

gustavonalle:
> @yrodiere Agreed on the synchronization issue, so let's promote this PR as an addition to #1246 rather than a replacement for it.

I am not sure as well if there is a synchronization issue, we never tried to flush at the same time

Sanne:
np @yrodiere, it's great to see you exploring this complex area :+1: Only a few of us know this section, and  it's not particularly readable. It's been quite battle-tested though, including in highly demanding environments from a perspective of performance, so I'll be conversvative on making any changes without a reproducer or at least having observed a problem.

Sanne:
> I am not sure as well if there is a synchronization issue, we never tried to flush at the same time

Ultimately that's synchronized within the IW. The only potential drawback, is the cost of doing many flushes in sequence but that shouldn't be a situation which we're likely to trigger.

PR #1249, opened 2016-12-05
HSEARCH-2503 Execute ExpectedLog4jLog properly even if log expectations are added *during* the test

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2503

This reverts commit b1867179159fa2784344d922cf142524c180ac61. That
commit added an optimization that turned out bypassing assertions
completely when expectations are defined during the test (and not inline
just after calling ExpectedLog4jLog.create()).


Sanne:
interesting one! merged

PR #1252, opened 2016-12-06
HSEARCH-2506 Add an acknowledgment section to the Elasticsearch integration docs

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2506


Sanne:
thanks!

PR #1253, opened 2016-12-07
HSEARCH-2508 Introduce @IndexedContainer to index multi-valued properties

yrodiere:
This PR is based on #1241 (HSEARCH-2172), which should be merged first.

Relevant tickets:

 * https://hibernate.atlassian.net/browse/HSEARCH-2508
 * https://hibernate.atlassian.net/browse/HSEARCH-1574
 * https://hibernate.atlassian.net/browse/HSEARCH-2500

The main ticket (HSEARCH-2508) states we should revert to the previous behavior of not applying our container bridge wrappers by default, for reasons that are explained in that ticket.

I did revert the default behavior, but I changed how to make Hibernate Search apply those container bridge wrappers: it now requires a dedicated annotation, `@IndexedContainer`, whereas it used to require an `@IndexedEmbedded`, which caused a lot of trouble recently (see in particular [HSEARCH-2419](https://hibernate.atlassian.net/browse/HSEARCH-2419), which was required for  [HSEARCH-2448](https://hibernate.atlassian.net/browse/HSEARCH-2448)).

As explained in the first commit, this is a breaking change, but the feature that we break (container indexing using element field bridges) was not documented and arguably already broken,
since we required to use @IndexedEmbedded to enable it, which in turn
enables another feature: the embedding of fields carried by the
elements in the root document.

I took this opportunity to finally document the feature (HSEARCH-1574), and I cherry-picked the tests by @DavideD for HSEARCH-2500, which happens to be fixed by this PR.

I'm not entirely satisfied of the name of `@IndexedContainer` and am totally open to suggestions. I thought about things like `@IndexElement`, `@ContainerElementField`, `@LoopingField`, `@ApplyOnElement`, but I'm not sure those would be better... On thing to keep in mind is that we can't have a trailing "s" because we need a repeatable annotation (`@IndexedContainers` in this PR)


yrodiere:
Please note that for an unknown reason, some commits appear jumbled on GitHub. For instance `	HSEARCH-2508 Revert "HSEARCH-2419 Split container bridges tests to te…  …` should be just before `HSEARCH-2508 Update (Array|Iterable|Map)BridgeTest following the introduction of @IndexedContainer`...
This seems to be a GitHub issue though, since a local `git log` gives me something more sensible.

engine/src/main/java/org/hibernate/search/bridge/impl/TikaBridgeProvider.java
```diff
@@ -36,19 +36,28 @@
...
+		Class<?> elementType = context.getReturnType();
+		// Example: URI, List<URI>, List<byte[]>
+		if ( ! Blob.class.isAssignableFrom( elementType )
+				&& ! String.class.isAssignableFrom( elementType )
+				&& ! byte[].class.isAssignableFrom( elementType )
+				&& ! URI.class.isAssignableFrom( elementType ) ) {
+			// byte[] is actually an array but we want to treat it as a simple element
+			Class<?> returnType = context.getElementOrContainerReturnType();
+			if ( ! byte[].class.isAssignableFrom( returnType ) ) {
+				throw LOG.unsupportedTikaBridgeType( elementType );
```
DavideD: I noticed it now, but I think the log should refer to `returnType` and not `elementType`

yrodiere: Right. I changed that on the new PR I'm preparing, thanks.

yrodiere:
PR rejected, closing.
I'm going to submit another PR soon that will not introduce a new annotation (but the behavior for Elasticsearch will be buggy again, which will have to do until 6.0).

PR #1255, opened 2016-12-08
HSEARCH-2456 Wait for index status does not raise an exception

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2456

It appears we were not checking that the index at least existed in some situations, and more importantly we were not waiting for the index status to reach the expected level in some situations. This PR fixes that.


Sanne:
great, merged!

Sanne:
I had only one comment - which can be done later - about the logging message org.hibernate.search.elasticsearch.logging.impl.Log.unexpectedIndexStatus(String, String, String)

do you think we should rephrase it to point out that we waited some n milliseconds before failing?

yrodiere:
@Sanne Sure: #1260

PR #1256, opened 2016-12-08
HSEARCH-2488 Throw an exception with a clear message when field bridges bypass @IndexedEmbedded prefixes with Elasticsearch

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2488


Sanne:
merged

PR #1258, opened 2016-12-08
HSEARCH-2508 Restore the former default for container bridges

yrodiere:
Relevant tickets:

 * https://hibernate.atlassian.net/browse/HSEARCH-2508
 * https://hibernate.atlassian.net/browse/HSEARCH-2500

This PR supersedes #1253. It:

 * reverts the default behavior for container bridges
 * mitigates the incompatibility between Elasticsearch and this container bridges features, making it work as long as `@IndexedEmbedded.indexNullAs` isn't used and container elements are not mapped using Hibernate Search annotations (e.g. Integer or String would work)
 * makes TikaBridge work again with these container bridges


yrodiere:
Rebased just now on master.

DavideD:
Merged https://github.com/hibernate/hibernate-search/pull/1258

PR #1259, opened 2016-12-08
HSEARCH-2441 ElasticsearchQueries.fromJson expects top-level JSON but only the "query" parameter is supported

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2441

Given that users cannot retrieve the results from aggregations (yet), I chose to disallow any attribute other than 'query'.

Other attributes seemed to be either:

 * query parameters, hence not in the JSON (`search_type`, `request_cache`)
 * related to aggregations (`aggs`)
 * already used by our implementations (`from`, `size`, `sort`, `timeout`, `_source`, `script_fields`)
 * or quite specific ( `terminate_after`, `post_filter`, ...)

All in all, since attributes other than `query` never worked anyway in our betas, I think we should disallow everything for the moment and take the time to think about how we want to provide access to features that are not directly supported by Hibernate Search.
In particular, it seems a shame that in order to use aggregations, for instance, one would have to write the query's JSON by hand and can't use the DSL anymore.
Maybe some crude mechanism could be added to `FullTextQuery` in order to allow passing additional, implementation-specific parameters? Something along the lines of `myQuery.setHint(ElasticsearchHints.POST_FILTER, queryBuilder.....createQuery()`?

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/ElasticsearchQueries.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/ElasticsearchQueries.java
@@ -21,6 +21,7 @@ import org.hibernate.search.util.logging.impl.LoggerFactory;
 
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
+import com.google.gson.JsonSyntaxException;
 
 /**
  * Creates queries to be used with Elasticsearch.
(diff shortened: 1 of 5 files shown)


Sanne:
great, thanks! Merged.

Made a minor change, as after rebase the Log IDs had to be bumped up, and tests needed to match.

PR #1260, opened 2016-12-09
HSEARCH-2456 When timing out waiting for an Elasticsearch index status, show the timeout value

yrodiere:
... in the error message

This is a follow-up for #1255


Sanne:
merged! thanks

PR #1261, opened 2016-12-09
HSEARCH-2500 backport to branch 5.5

yrodiere:
This PR is a partial backport of #1258, so we'll have to merge that one into master first.

Relevant ticket: https://hibernate.atlassian.net/browse/HSEARCH-2500

This is a backport of HSEARCH-2500 for branch 5.5. Branch 5.5 doesn't actually need a fix, but the "instanceof TikaBridge" hack wasn't needed on this branch in the first place. So we might as well remove it and add the tests to make sure it isn't needed.


DavideD:
Merged https://github.com/hibernate/hibernate-search/pull/1261

PR #1262, opened 2016-12-10
HSEARCH-2387 Dynamic mapping option

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-2387

This patch will add the property `hibernate.search.default.elasticsearch.dynamic_mapping`.
User can now set the dynamic mapping approach for the whole index.

It will also make it possible for a user to specify the dynamic mapping for a field using the MetadataProvidingFieldBridge. It will look like this;

```
	@Override
	public void configureFieldMetadata(String name, FieldMetadataBuilder builder) {
		builder.field( "dynamicField", FieldType.OBJECT )
				.mappedOn( Elasticsearch.class )
				     .dynamic( DynamicType.TRUE );

		builder.field( "dynamicStrict", FieldType.OBJECT );
	}
```

Note that I also added FieldType.OBJECT, this is required in elasticsearch because we cannot set a `dynamic: true` on a field of type primitive.

@Sanne I haven't implemented the .overrideSchemaJson(...) funciton we discussed yet, but it shouldn't be too hard to add it as a separate PR.


DavideD:
jenkins, retest this please

DavideD:
Jenkins retest this please

yrodiere:
@DavideD FYI, I just rebased this on master. The conflicts were just on imports.
Reviewing now.

engine/src/test/java/org/hibernate/search/test/bridge/MultiFieldMapBridge.java
```diff
@@ -34,4 +36,8 @@ private void setField(String fieldPrefix, String key, String value, Document doc
 		luceneOptions.addFieldToDocument( fieldPrefix + "." + key, value, document );
 	}
 
+	@Override
+	public void configureFieldMetadata(String name, FieldMetadataBuilder builder) {
+		builder.field( "value", FieldType.OBJECT );
```
yrodiere: This should not be a hardcoded constant (`"value"`), but simply `name`. I know this won't change anything in practice, but it's just in case people copy/paste this code.

elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/bridge/MapAsInnerObjectFieldBridge.java
```diff
@@ -0,0 +1,49 @@
...
+		luceneOptions.addFieldToDocument( fieldPrefix + "." + key, value, document );
+	}
+
+	@Override
+	public void configureFieldMetadata(String name, FieldMetadataBuilder builder) {
+		builder.field( "dynamicField", FieldType.OBJECT )
+				.mappedOn( Elasticsearch.class )
+				.dynamic( DynamicType.TRUE );
+
+		builder.field( "dynamicStrict", FieldType.OBJECT );
```
yrodiere: I'm a bit perplexed: "dynamicStrict" does not match any field name... I wonder why this works.

Anyway... Maybe there should be two different bridges here? It feels a bit strange that the `name` parameter in `configureFieldMetadata` is ignored.

Same as before: I agree it can work as is, but it's more in case someone copy/pastes this code.

engine/src/main/java/org/hibernate/search/engine/metadata/impl/BridgeDefinedField.java
```diff
@@ -57,4 +67,12 @@ public FieldType getType() {
 	public String toString() {
 		return "BridgeDefinedField [name=" + absoluteName + ", type=" + type + "]";
 	}
+
+	public void add(Class class1, Object elasticsearch) {
```
yrodiere: This could use generics, so that we ensure (more or less) that the cast in `mappedOn` (below) will always work.

```
public <T> void add(Class<T> class1, T elasticsearch) {
  extra.put( class1, elasticsearch );
}
```

Also, "elasticsearch" should probably be renamed :)

engine/src/main/java/org/hibernate/search/engine/metadata/impl/BridgeDefinedField.java
```diff
@@ -24,6 +27,13 @@
 	 */
 	private final String absoluteName;
 	private final FieldType type;
+	private Map<Class, Object> extra = new HashMap<>();
```
yrodiere: Shouldn't this be `Map<Class<?>, Object>` instead? It will save us one compiler warning.

engine/src/main/java/org/hibernate/search/engine/metadata/impl/BridgeDefinedField.java
```diff
@@ -57,4 +67,12 @@ public FieldType getType() {
 	public String toString() {
 		return "BridgeDefinedField [name=" + absoluteName + ", type=" + type + "]";
 	}
+
+	public void add(Class class1, Object elasticsearch) {
+		extra.put( class1, elasticsearch );
+	}
+
+	public <T> T mappedOn( Class<T> class1 ) {
+		return (T) extra.get( class1 );
```
yrodiere: Unless I'm mistaken, there should probably be a `@SuppressWarnings("unchecked")` on this method, to avoid a warning. I know there already are tons of them, but we could at least refrain from adding new ones...

Another (maybe better) solution would be to use the class to do the cast:

```
  return class1.cast( extra.get( class1 ) );
```

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/ElasticsearchEnvironment.java
```diff
@@ -223,6 +224,21 @@
 	 */
 	public static final String SCROLL_TIMEOUT = "elasticsearch.scroll_timeout";
 
+	/**
+	 * Property for specifying the behaviour when Elasticsearch encounter a previously unknown field in a document.
```
yrodiere: Maybe "encounters", with a trailing "s"?

DavideD:
I rebased the project and applied all the remarks.

Instead of creating 2 field bridges for `DynamicMappingIT` I've created a parameterized filed bridge. I've also added some javadoc

gunnarmorling:
What are actual use cases for enabling dynamic mappings? Reading through the issue, it seems the reporter actually hit an issue which would rather be solved by properly mapping indexed embeddables (using the correct strict schema).

It seems like a big deviation from the current approach to support dynamically added fields and I'm wondering whether we've considered all the consequences. How does field validation in the query DSL work for instance?

yrodiere:
@gunnarmorling I must admit I thought the same before this PR, but @DavideD enabled a test that demonstrate its usefulness when you want to index a map. This test previously only worked on Lucene, see the commit "HSEARCH-2387 Update existing tests".

So it may be useful. Now, do we want to support that on Elasticsearch, that's another question. Regarding the type detection in the query DSL, it doesn't seem any different from Lucene to me (there also is a difference between querying strings and numbers). Maybe that's not an issue with Elasticsearch, but a more general one?

As a side note, we do have tickets relative to "dynamic" fields, but not directly related to Elasticsearch: [HSEARCH-1379](https://hibernate.atlassian.net/browse/HSEARCH-1379) and maybe also [HSEARCH-1401](https://hibernate.atlassian.net/browse/HSEARCH-1401)

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/ElasticsearchEnvironment.java
```diff
@@ -18,6 +18,7 @@
 	 */
 	public static final class Defaults {
 
+		public static final String DYNAMIC_MAPPING = "strict";
```
gunnarmorling: Can we have a public enum (as we e.g. do for `ElasticsearchIndexStatus`)?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/schema/impl/DefaultElasticsearchSchemaTranslator.java
```diff
@@ -145,6 +148,12 @@ private void addPropertyMapping(ElasticsearchMappingBuilder mappingBuilder, Docu
 		mappingBuilder.setPropertyAbsolute( propertyPath, propertyMapping );
 	}
 
+	private void addStoreOption(DocumentFieldMetadata fieldMetadata, PropertyMapping propertyMapping) {
```
gunnarmorling: What's the reason for having this in a separate method but not doing the same for boost?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/bridge/builtin/impl/Elasticsearch.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.hibernate.search.bridge.spi.FieldMetadataCreationContext;
+import org.hibernate.search.bridge.spi.FieldType;
+import org.hibernate.search.elasticsearch.schema.impl.model.DynamicType;
+import org.hibernate.search.engine.metadata.impl.BridgeDefinedField;
+
+/**
+ * @author Davide D'Alto
+ */
+public class Elasticsearch implements FieldMetadataCreationContext {
```
yrodiere: Sorry about that, but I just noticed: this class is part of an implementation package. If we want to make it available for use in custom field bridges, shouldn't this be API?

Also, maybe a more specific name is necessary. Or we consider that this class will be used everywhere someone needs an Elasticsearch-specific API, and in that case there should be a slightly more complex design (one central interface with multiple `Class<? extends XXX> getXXXAPI()` methods).

engine/src/main/java/org/hibernate/search/bridge/spi/FieldType.java
```diff
@@ -48,5 +48,10 @@
 	/**
 	 * A double field.
 	 */
-	DOUBLE
+	DOUBLE,
+
+	/**
+	 * A container for othe fields, for example an embedded object or a map.
```
gunnarmorling: "othe"="other".

I haven't fully groked yet why this is needed now. Could you explain a bit?

DavideD: In Elasticsearch you cannot set dynamic:true on a field that it's not an object (or nested). FieldType
does not have a way to define a similar type. At the moment FIeldTYpe allows only primitive types.

I guess the javadoc shouldn't talk about containers, I'll update it

DavideD:
@gunnarmorling @yrodiere 
One last attempt of keeping the API part:

* `FieldMetadataCreationContext` is an experimental API and we specify that it is currently under development
* `Elasticsearch` is the equivalent of `FieldMetadataCreationContextImpl`  that it is currently under an impl package. The name should probably be `ElasticsearchFieldMetadataCreationContext` (but it seems too long to me). `ElasticsearchFieldMetadata`?
* FieldMetadataCreationContext is an SPI and not a public API, it also does not mention Elasticsearch in the API but only that mappedOn should return a FieldCreationContext

@Sanne better if you give a look as well.

I'm OK postponing this part, but I would prefer to keep it because the user can startto experiment with it and we have an initial POC before committing to an API

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/ElasticsearchEnvironment.java
DavideD: We alredy have, I'll update it

yrodiere: @DavideD Caution with the existing enum: it's part of an implementation package, and for a good reason. It's used for building our internal representation of the Elasticsearch mapping, and as such it shouldn't be exposed.

We can always add another enum, if it's really necessary.

DavideD:
> Do we have strong use case for this? If there are issues with the emitted strict mapping, we should rather fix those.

What do you mean by issues? Elasticsearch allow to define different field with different dynamic type properties. So you could specify that all the fields are defined but one of the field is an inner object and the fields in it can change. Maybe this is not the best use case for hibernate search, but there might be other clients that need to access the same index. There is an example in the Elasticsearch documentation: https://www.elastic.co/guide/en/elasticsearch/guide/current/dynamic-mapping.html

> The notion of "object fields" seems to have a huge impact. It may be justified to add, but it's still a bit blurry to me.

I suspect that the more backends we have the more fileld types we will need, for example Elastic also support `nested` types. I'm not sure how we want to address this in the future. For this specific case, INNER_OBJECT might be a better name, the javadoc will specify that the mapping can change based on the backend capabilities.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/ElasticsearchEnvironment.java
DavideD: RIght, I will add a new one to be consistent with the other options

DavideD:
FYI I'm applying the remarks

Sanne:
> @Sanne better if you give a look as well.

Sure, looking forward to it but I'll wait for your revised version! Thanks

DavideD:
I've applied all the remarks and kept the API for now:
- I've created a new cfg.DynamivTYpe for the configuration
- I've moved the `bridge.builtin.impl.Elasticsearch` class under `bridge.spi.Elasticsearch` 
- Added/Updated javadoc
- Fiixed an error in the MapAsInnerObjectFieldBridge bridge used by the test

The API is still the same (I've only changed the package), so the questions about keeping it are still relevant.
 
@Sanne You can have a look now,

DavideD:
Jenkins, retest this please

yrodiere:
@DavideD While working on something else, I just came across a section in the documentation that stated "Creating sub-fields in custom field bridges is not supported." and explain at length why it isn't supported.

Thing is, with your new `OBJECT` field type, it actually is supported (you only have to override the type of the default field to `OBJECT`). Maybe this part of the documentation should be rewritten?

Sanne:
please don't use the "@see" Javadoc tag to point to external URLs. that will have the release process fail.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/ElasticsearchEnvironment.java
```diff
@@ -223,6 +224,21 @@
 	 */
 	public static final String SCROLL_TIMEOUT = "elasticsearch.scroll_timeout";
 
+	/**
+	 * Property for specifying the behaviour when Elasticsearch encounters a previously unknown field in a document.
```
Sanne: Could you clarify who is the subject of "the behaviour"?

Also "unknown field" isn't very clear IMO. An "unexpected field name"? Or better, a "field name which was not declared in the index schema" ?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/ElasticsearchEnvironment.java
```diff
@@ -223,6 +224,21 @@
...
+	 * <p>
+	 * Possible values are:
+	 * <p>
+	 * <ul>
+	 *   <li>{@code true}: Add new fields dynamically</li>
+	 *   <li>{@code false}: Ignore new fields</li>
+	 *   <li>{@code strict}: Throw an exception if an unknown field is encountered</li>
```
Sanne: Same comment on "unknown field"; I might be wrong but I suspect it's not very clear.

engine/src/main/java/org/hibernate/search/bridge/spi/FieldMetadataCreationContext.java
```diff
@@ -33,4 +33,12 @@
 	 * @return this context for fluent API invocations
 	 */
 	FieldMetadataCreationContext sortable(boolean sortable);
+
+	/**
+	 * Give access to properties that are backend specific
```
Sanne: Please don't call it "backend".. I feel your pain, we are lacking a good designation for this but "backend" is an overloaded term in this context, which already has a (different) meaning.

engine/src/main/java/org/hibernate/search/engine/metadata/impl/BridgeDefinedField.java
```diff
@@ -19,12 +22,24 @@
...
-	 * only provide the absolute name, without telling the prefix from the relative name.
+	 * Here we cannot use DocumentFieldPath because metadata-providing field bridges only provide the absolute name,
+	 * without telling the prefix from the relative name.
 	 */
 	private final String absoluteName;
 	private final FieldType type;
 
+	/*
+	 * Backends have specific field properties (Elastic search dynamic mapping for example). this map will group the
+	 * properties per backend.
```
Sanne: I'm not sure what you mean here. Specifically what is  "backend" meaning in this context?

engine/src/main/java/org/hibernate/search/engine/metadata/impl/FieldMetadataBuilderImpl.java
```diff
@@ -57,8 +57,21 @@ public FieldMetadataCreationContext field(String name, FieldType type) {
...
+		public <T extends FieldMetadataCreationContext> T mappedOn(Class<T> backend) {
+			return create( backend );
+		}
+
+		private <T extends FieldMetadataCreationContext> T create(Class<T> backend) {
+			try {
+				return backend.getDeclaredConstructor( BridgeDefinedField.class, FieldMetadataCreationContext.class ).newInstance( field, this );
+			}
+			catch (Exception e) {
+				throw new RuntimeException( e );
```
Sanne: Could you use the Logger, and have it create a SearchException ?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/bridge/spi/Elasticsearch.java
```diff
@@ -0,0 +1,60 @@
...
+ * Extends {@link FieldMetadataCreationContext} allowing the user to define additional properties for a field specific
+ * to Elasticsearch.
+ *
+ * @hsearch.experimental This contract is currently under active development and may be altered in future releases,
+ * @author Davide D'Alto
+ */
+public class Elasticsearch implements FieldMetadataCreationContext {
+
+	private final FieldMetadataCreationContext creationContext;
+	private ElasticsearchBridgeDefinedField elasticsearchDefinedField;
```
Sanne: `final` ?

engine/src/main/java/org/hibernate/search/bridge/spi/FieldType.java
```diff
@@ -48,5 +48,13 @@
 	/**
 	 * A double field.
 	 */
-	DOUBLE
+	DOUBLE,
+
+	/**
+	 * An object field, a generic object that doesn't have a unique definition in all the backends.
```
Sanne: The "object" being repeated is confusing. "backends" at the end of the sentence is as well.

alternative proposal:
"A composite object that will benefit from different encoding on each indexing technology."

Sanne:
API looks great, thanks for exploring all those options.

For the record, I expect to be ableto "chain" such metadata definitions eventually, when we'll work on supporting multiple indexing technologies.

e.g.

    builder.field( "dynamicField", FieldType.OBJECT )
				.mappedOn( Elasticsearch.class )
				     .dynamic( DynamicType.TRUE ) // Not terminating yet..
				.mappedOn( Solr.class )
				     .useDocValuesAsStored( true ); // A Solr specific option

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/ElasticsearchEnvironment.java
DavideD: Isn't it explained in details one line after this one?

engine/src/main/java/org/hibernate/search/bridge/spi/FieldMetadataCreationContext.java
DavideD: I'll think a differrent word, but we need a term to define these. Should we just refer to index managers?

engine/src/main/java/org/hibernate/search/engine/metadata/impl/FieldMetadataBuilderImpl.java
DavideD: Sure

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/ElasticsearchEnvironment.java
Sanne: I mean: it's not clear if this is to configure the "behaviour of Elasticsearch" or of Hibernate Search

engine/src/main/java/org/hibernate/search/bridge/spi/FieldMetadataCreationContext.java
Sanne: "Give access to properties that are specific to an indexing technology" ? Not particularly fond of that, I'm open to better ideas.

DavideD:
Rebased and remarks applied.

I realized that I made a mistake and applied the PR for HSEARCH-2508 and HSEARCH-2500 (https://github.com/hibernate/hibernate-search/pull/1258) on master on my repository and not on upstream. The end result is that  I had to to rebase the patches made by @yrodiere  again and they are included in this PR making it bigger than it actually is. I'm sorry about it.

@yrodiere I don't feel confident about removing that part of the documentation (the one related about the  sub-fields) until we have a test covering the use-case. What it is describing seems different to what I've done and I don't feel confident removing or rewriting the section until we tested it (as a separate issue).

Sanne:
@DavideD I'm confused. There are several other commits included now, some of them are reverts. Intentional?

DavideD:
@Sanne that's my fault, I made a mistake when I applied the the  https://github.com/hibernate/hibernate-search/pull/1258. I didn't realized I applied that on my remote repository and not on upstream.

I noticed it today when other PR were already included. This branch was already build on top of those changes.

At least, that's what I think happened since master does not have those commits and I remember pushing them.

Anyway, this changes related to HSEARCH-2500 and HSEARCH-2508 should be on master already.

The reverts are related to add back the `@IndexEmbedded` as requirement for collections like discussed elsewhere.

Sanne:
Ok I understand you had forgotten to push some changes. But why include them in this PR? Just push them?

DavideD:
Yes, I could have done that :)

Sanne:
thanks! merged

PR #1263, opened 2016-12-16
HSEARCH-2219 Define analyzers via the REST API

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2219

This PR adds automatic registration of analyzer definitions to Elasticsearch upon index creation.

There are a few gotchas, and those are documented. The main ones being that the MERGE strategy doesn't update the analyzer definitions yet (because it would require closing/opening the index, and I'm still pondering the implications) and the VALIDATE strategy doesn't validate analyzer definitions yet (because I didn't have the time to add that). I created follow-up tickets: https://hibernate.atlassian.net/browse/HSEARCH-2519 and https://hibernate.atlassian.net/browse/HSEARCH-2520.


engine/src/main/java/org/hibernate/search/analyzer/impl/LuceneEmbeddedAnalyzerStrategy.java
```diff
@@ -0,0 +1,154 @@
...
+
+	@Override
+	public void initializeReferences(Collection<LuceneAnalyzerReference> references, Map<String, AnalyzerDef> analyzerDefinitions) {
+		Map<String, Analyzer> initializedAnalyzers = new HashMap<>();
+		for ( LuceneAnalyzerReference reference : references ) {
+			initializeReference( initializedAnalyzers, reference, analyzerDefinitions );
+		}
+	}
+
+
```
gsmet: Unneeded white space.

engine/src/main/java/org/hibernate/search/engine/impl/ConfigContext.java
```diff
@@ -302,86 +253,57 @@ private void addFilterDef(FullTextFilterDef defAnn) {
...
-		initLazyRemoteAnalyzer( initializedAnalyzers, RemoteAnalyzerReference.PASS_THROUGH, indexesFactory );
-
-		// initialize the remaining definitions
-		for ( Map.Entry<String, AnalyzerDef> entry : analyzerDefs.entrySet() ) {
-			if ( !initializedAnalyzers.containsKey( entry.getKey() ) ) {
-				final Analyzer analyzer = buildAnalyzer( entry.getValue() );
-				final AnalyzerReference reference = new LuceneAnalyzerReference( analyzer );
-				initializedAnalyzers.put( entry.getKey(), reference );
+		// Ensure every analyzer definition will be initialized
+		for ( String name : analyzerDefs.keySet() ) {
```
gsmet: As discussed on the phone, could you explain a bit more the sequence of operations. Typically here, it would be nice to explain that analyzers not used in the mapping are registered as Lucene analyzers.

engine/src/main/java/org/hibernate/search/engine/impl/ConfigContext.java
```diff
@@ -302,86 +253,57 @@ private void addFilterDef(FullTextFilterDef defAnn) {
...
+			AnalyzerReferenceRegistry<?> registry = getAnalyzerReferenceRegistry( indexManagerType );
+			registry.initialize( analyzerDefs );
+			Map<String, ? extends AnalyzerReference> referencesByNameForType =
+					registry.getAnalyzerReferencesByName();
+
+			// Check for naming conflicts between different index manager types
+			if ( !referencesByName.isEmpty() ) {
+				for ( Map.Entry<String, ? extends AnalyzerReference> entry : referencesByNameForType.entrySet() ) {
+					String referenceName = entry.getKey();
+					if ( referencesByName.containsKey( referenceName ) ) {
```
gsmet: A comment explaining the limitation we discussed about the order would be nice.

engine/src/main/java/org/hibernate/search/engine/metadata/impl/TypeMetadata.java
```diff
@@ -504,13 +505,24 @@ public boolean isJpaIdUsedAsDocumentId() {
...
+			else {
+				IndexManagerType indexManagerType = parseContext.getIndexManagerType();
+				this.analyzerReferenceRegistry = configContext.getAnalyzerReferenceRegistry( indexManagerType );
+				this.scopedAnalyzerReferenceBuilder = new ScopedAnalyzerReference.Builder(
+						analyzerReferenceRegistry.getDefaultAnalyzerReference() );
+			}
 		}
 
-		public Builder(Class<?> indexedType, ScopedAnalyzerReference.Builder scopedAnalyzerReferenceBuilder) {
+		public Builder(Class<?> indexedType, Builder builderToClone) {
```
gsmet: `builderToClone` is misleading here. You don't clone the `Builder`. I think `containerBuilder` would be better.

engine/src/main/java/org/hibernate/search/engine/metadata/impl/TypeMetadata.java
```diff
@@ -594,15 +606,13 @@ public AnalyzerReference addToScopedAnalyzerReference(DocumentFieldPath fieldPat
...
 			}
 
-			if ( Field.Index.ANALYZED.equals( index ) || Field.Index.ANALYZED_NO_NORMS.equals( index ) ) {
-				if ( analyzerReference != null ) {
-					scopedAnalyzerReferenceBuilder.addAnalyzerReference( fieldPath.getAbsoluteName(), analyzerReference );
-				}
-			}
-			else {
+			if ( !index.isAnalyzed() ) {
 				// no analyzer is used, add a fake one for queries
```
gsmet: Not related to your patch, but while at it, "fake" is misleading here, we use the pass through analyzer.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/analyzer/impl/ElasticsearchAnalyzerStrategy.java
```diff
@@ -7,12 +7,14 @@
 package org.hibernate.search.elasticsearch.analyzer.impl;
 
 import java.util.Collection;
+import java.util.Collections;
```
gsmet: Mmmmh, the commit message does not really state the content of this commit. It's far more extensive AFAICS.

engine/src/main/java/org/hibernate/search/analyzer/impl/LuceneAnalyzerReference.java
```diff
@@ -33,7 +33,7 @@ public void close() {
 
 	@Override
 	public <T extends AnalyzerReference> boolean is(Class<T> analyzerType) {
-		return LuceneAnalyzerReference.class.isAssignableFrom( analyzerType );
+		return analyzerType.isAssignableFrom( LuceneAnalyzerReference.class );
```
gsmet: Shouldn't we use getClass() instead of using a static class reference?

engine/src/main/java/org/hibernate/search/analyzer/impl/RemoteAnalyzerReference.java
```diff
@@ -35,7 +35,7 @@ public void close() {
 
 	@Override
 	public <T extends AnalyzerReference> boolean is(Class<T> analyzerType) {
-		return RemoteAnalyzerReference.class.isAssignableFrom( analyzerType );
+		return analyzerType.isAssignableFrom( RemoteAnalyzerReference.class );
```
gsmet: Shouldn't we use getClass() instead of using a static class reference?

engine/src/main/java/org/hibernate/search/analyzer/impl/ScopedLuceneAnalyzerReference.java
```diff
@@ -0,0 +1,32 @@
...
+	}
+
+	@Override
+	public ScopedLuceneAnalyzer getAnalyzer() {
+		return (ScopedLuceneAnalyzer) super.getAnalyzer();
+	}
+
+	@Override
+	public <T extends AnalyzerReference> boolean is(Class<T> analyzerType) {
+		return analyzerType.isAssignableFrom( ScopedLuceneAnalyzerReference.class );
```
gsmet: Shouldn't we use getClass() instead of using a static class reference?

engine/src/main/java/org/hibernate/search/analyzer/impl/ScopedRemoteAnalyzer.java
```diff
@@ -26,53 +26,74 @@
 
 	private static final Log log = LoggerFactory.make();
 
-	private RemoteAnalyzer globalAnalyzer;
+	private final RemoteAnalyzerReference globalAnalyzerReference;
```
gsmet: Having the analyzer here was not a mistake. For me the analyzer didn't really need to be aware of the notion of reference. Why exactly did you change that?

engine/src/main/java/org/hibernate/search/backend/impl/lucene/works/AddWorkExecutor.java
```diff
@@ -84,7 +85,7 @@ static ScopedAnalyzerReference updateAnalyzerMappings(Workspace workspace, Scope
 			return scopedAnalyzerReference;
 		}
 
-		ScopedAnalyzerReference.Builder copyBuilder = new ScopedAnalyzerReference.Builder( scopedAnalyzerReference );
+		ScopedAnalyzer.Builder copyBuilder = scopedAnalyzerReference.getAnalyzer().startCopy();
```
gsmet: It's weird. It looks like we are copying an analyzer while in fact we build an AnalyzerReference. Maybe the name `Builder` is misleading and `ReferenceBuilder` would be better?

All in all, I'm not convinced this is an improvement. What's the exact reason for this change (the commit message does not really help).

engine/src/main/java/org/hibernate/search/analyzer/impl/LuceneAnalyzerReference.java
```diff
@@ -16,20 +19,51 @@
...
 		return analyzer;
 	}
 
+	public boolean isInitialized() {
+		return analyzer != null;
+	}
+
+	public void initialize(Analyzer analyzer) {
+		if ( this.analyzer != null ) {
+			throw new AssertionFailure( "An analyzer reference has been initialized more than once." );
```
gsmet: Maybe include the name of the analyzer or the analyzer if defined?

gsmet: I think we should include the name for the builtin analyzers. That would be more clear. (I know this code has moved so it's probably something to apply on top of the other changes)

engine/src/main/java/org/hibernate/search/util/logging/impl/Log.java
```diff
@@ -964,7 +963,7 @@
 	SearchException remoteAnalyzerAlreadyDefinedAsLuceneAnalyzer(String name);
 
 	@Message(id = 315, value = "Lazy remote analyzer %1$s is not initialized.")
-	SearchException lazyRemoteAnalyzerNotInitialized(LazyRemoteAnalyzer analyzer);
+	SearchException lazyRemoteAnalyzerReferenceNotInitialized(RemoteAnalyzerReference reference);
```
gsmet: Looks like you forgot to update the message as you did above.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/analyzer/impl/ElasticsearchAnalyzerReference.java
```diff
@@ -0,0 +1,86 @@
...
+
+	private ElasticsearchAnalyzer analyzer;
+
+	public ElasticsearchAnalyzerReference(String name) {
+		this.name = name;
+		this.analyzer = null; // Not initialized
+	}
+
+	public ElasticsearchAnalyzerReference(ElasticsearchAnalyzer analyzer) {
+		this.name = null;
```
gsmet: I think it would be nice to initialize the name.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/settings/impl/LuceneAnalysisDefinitionTranslationMapBuilder.java
```diff
@@ -0,0 +1,144 @@
...
+ *
+ * @author Yoann Rodiere
+ */
+class LuceneAnalysisDefinitionTranslationMapBuilder<D extends AnalysisDefinition> {
+
+	private final Class<D> targetClass;
+	private final Map<String, AnalysisDefinitionFactory<D>> result = new HashMap<>();
+
+	public LuceneAnalysisDefinitionTranslationMapBuilder(Class<D> targetClass) {
+		super();
```
gsmet: No `super();` needed.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/settings/impl/LuceneAnalysisDefinitionTranslationMapBuilder.java
```diff
@@ -0,0 +1,144 @@
...
+			translations.put( luceneValue, elasticsearchValue );
+			return this;
+		}
+
+		public SimpleAnalysisDefinitionFactoryBuilder<D> end() {
+			return parent.transform( parameterName, new MapParameterValueTransformer( factoryClass, parameterName, translations ) );
+		}
+
+	}
+}
```
gsmet: IIRC, for new files, we add an empty line at the end.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/settings/impl/SimpleAnalysisDefinitionFactory.java
```diff
@@ -0,0 +1,95 @@
...
+	private final Class<D> targetClass;
+	private final String type;
+	private final Map<String, String> parameterNameTranslations;
+	private final Map<String, ParameterValueTransformer> parameterValueTranslations;
+	private final Map<String, JsonElement> staticParameters;
+
+	public SimpleAnalysisDefinitionFactory(Class<D> targetClazz, String type, Map<String, String> parameterNameTranslations,
+			Map<String, ParameterValueTransformer> parameterValueTranslations,
+			Map<String, JsonElement> staticParameters) {
+		super();
```
gsmet: Mmmh, OK, any reason why you add a `super();` everywhere? I can see a pattern here :).

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/settings/impl/SimpleAnalysisDefinitionFactory.java
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	@Override
+	public D create(Parameter[] parameters) {
+		D result;
+		try {
+			result = targetClass.newInstance();
+		}
+		catch (InstantiationException | IllegalAccessException e) {
+			throw new AssertionFailure( "Unexpected failure while instanciating a definition", e );
```
gsmet: instan**t**iating

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/settings/impl/SimpleAnalysisDefinitionFactory.java
```diff
@@ -0,0 +1,95 @@
...
+			result.setParameters( parameterMap );
+		}
+
+		return result;
+	}
+
+	protected void addParameter(Map<String, JsonElement> parameterMap, String luceneParameterName, String value) {
+		String elasticsearchParameterName = parameterNameTranslations.get( luceneParameterName );
+		if ( elasticsearchParameterName == null ) {
+			// Assume the parameter actually exist in Elasticsearch
```
gsmet: exists

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/settings/impl/model/AnalyzerDefinitionJsonAdapterFactory.java
```diff
@@ -0,0 +1,30 @@
...
+package org.hibernate.search.elasticsearch.settings.impl.model;
+
+import java.util.List;
+
+import com.google.gson.reflect.TypeToken;
+
+/**
+ * @author Yoann Rodiere
+ */
+public class AnalyzerDefinitionJsonAdapterFactory extends AnalysisDefinitionJsonAdapterFactory {
```
gsmet: Wondering if it's worth introducing this infrastructure rather than simply use a JsonElement/JsonArray to serialize the elements.

orm/src/test/java/org/hibernate/search/test/analyzer/inheritance/AnalyzerInheritanceTest.java
```diff
@@ -53,12 +56,12 @@ public void testBySearch() throws Exception {
...
 
-		QueryParser parser = new QueryParser( "name", s.getSearchFactory().getAnalyzer( SubClass.class ) );
-		org.apache.lucene.search.Query luceneQuery = parser.parse( "name:Proca\u00EFne" );
+		QueryParser parser = new QueryParser( "name", TestConstants.keywordAnalyzer );
+		org.apache.lucene.search.Query luceneQuery = parser.parse( "name:John" );
 		FullTextQuery query = s.createFullTextQuery( luceneQuery, SubClass.class );
 		assertEquals( 1, query.getResultSize() );
 
-		luceneQuery = parser.parse( "name:Procaine" );
+		luceneQuery = parser.parse( "name:Doe" );
```
gsmet: I'm a bit worried about this change. The StandardAnalyzer that might be used as default would have the exact same behavior. I really think using an analyzer filtering the non ASCII characters is better. I think it's feasible, isn't it?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/logging/impl/Log.java
```diff
@@ -339,4 +339,9 @@ BulkRequestFailedException elasticsearchBulkRequestFailed(String request, String
 			value = "The parameter '%2$s' for the factory '%1$s' has an unsupported value: '%3$s' is unsupported with Elasticsearch." )
 	SearchException unsupportedAnalysisDefinitionParameterValue(Class<?> factoryClass, String parameterName, String parameterValue);
 
+	@Message(id = ES_BACKEND_MESSAGES_START_ID + 63,
+			value = "The analyzer implementation '%1$s' is not supported with Elasticsearch."
+					+ " Please only use builtin Lucene analyzers that have a builtin equivalent in Elasticsearch.")
+	SearchException unsupportedAnalyzerImplementation(Class<?> luceneClass);
```
gsmet: Use `@FormatWith`.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/analyzer/ElasticsearchCharFilterFactory.java
```diff
@@ -0,0 +1,42 @@
...
+ * Example:
+ * <pre><code>
+ * {@literal @}CharFilterDef(
+ * 		name = "custom-pattern-replace",
+ * 		factory = ElasticsearchCharFilterFactory.class,
+ * 		params = {
+ * 				{@literal @}Parameter(name = "type", value = "pattern_replace"),
+ * 				{@literal @}Parameter(name = "pattern", value = "'[^0-9]'"),
+ * 				{@literal @}Parameter(name = "replacement", value = "'0'"),
+ * 				{@literal @}Parameter(name = "tags", value = "'CASE_INSENSITIVE|COMMENTS'")
```
gsmet: I wonder if we should simply pass a json object instead of separate parameters?

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -286,11 +286,208 @@ public static class Tweet {
...
+https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-custom-analyzer.html[custom analyzer]
+in  the index settings.
+
+[WARNING]
+====
+Analyzer definitions will only be created automatically when Hibernate Search creates the index.
+Hibernate Search will never update analyzer definitions on an existing index
+(no support for the `MERGE` <<elasticsearch-schema-management-strategy,schema management strategy>>)
+and will never validate that pre-existing definitions actually match your Hibernate Search mapping
+(no support for the `VALIDATE` <<elasticsearch-schema-management-strategy,schema management strategy>>).
```
gsmet: Is it a work in progress? It seems so as you created issues for it.

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -286,11 +286,208 @@ public static class Tweet {
...
+	@AnalyzerDef(
+		name = "tweet_analyzer",
+		charFilters = {
+			@CharFilterDef(
+				name = "custom_html_strip",
+				factory = ElasticsearchCharFilterFactory.class,
+				params = {
+					@Parameter(name = "type", value = "'html_strip'"),
+					// One can use Json arrays
+					@Parameter(name = "escaped_tags", value = "['br' 'p']")
```
gsmet: I know the parser is lenient but wouldn't it be better with a , between the array elements?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/settings/impl/model/AnalyzerDefinitionJsonAdapterFactory.java
yrodiere: Actually, it occurred to me only *after* I spent hours on this crap :-(
Anyway... Using `JsonElement`/`JsonArray` may hurt when we'll implement validation for analyzer definitions, so I'd like to avoid that as much as possible. I know it was a real pain when I implemented mapping validation.

orm/src/test/java/org/hibernate/search/test/analyzer/inheritance/AnalyzerInheritanceTest.java
yrodiere: Thing is, we can only use a built-in Lucene analyzer, unless we want to alter the test another way, by using analyzer definitions instead of an analyzer implementation.
I changed this to use Turkish diacritics. `StandardAnalyzer` seems to fail at removing those and the turkish analyzer is built-in.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/logging/impl/Log.java
yrodiere: Fixed

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/analyzer/ElasticsearchCharFilterFactory.java
yrodiere: I agree it could do, but I'm not sure it would be better.

1. We'd have some magic parameter name (not really intuitive)
2. We'd have more error cases to handle (what happens if someone gives a parameter name other than the magic one?)
3. It would not be much more readable, thanks to Java lacking multi-line string constants (a real pain, that one...)
4. On top of that, I'm not even sure there are analyzers with "object" parameters in Elasticsearch. I sure didn't come across any. The parameters are essentially flat, with arrays sometimes. So we wouldn't gain much...

To be honest, I'd say "yes" if we were introducing Elasticsearch-specific annotations, but in that case it doesn't seem to fit the annotation very well...

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
yrodiere: It sure would :) Fixed

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
yrodiere: Yes it is. I added details about that.

engine/src/main/java/org/hibernate/search/analyzer/impl/LuceneEmbeddedAnalyzerStrategy.java
yrodiere: Fixed.

engine/src/main/java/org/hibernate/search/engine/impl/AnalyzerReferenceRegistry.java
```diff
@@ -0,0 +1,73 @@
...
+
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import org.hibernate.search.analyzer.spi.AnalyzerReference;
+import org.hibernate.search.analyzer.spi.AnalyzerStrategy;
+import org.hibernate.search.annotations.AnalyzerDef;
+
+public class AnalyzerReferenceRegistry<T extends AnalyzerReference> {
```
yrodiere: Fixed.

engine/src/main/java/org/hibernate/search/engine/impl/AnalyzerReferenceRegistry.java
```diff
@@ -0,0 +1,73 @@
...
+
+	private final AnalyzerStrategy<T> strategy;
+
+	private final T defaultReference;
+
+	private final T passThroughtReference;
```
yrodiere: Fixed.

engine/src/main/java/org/hibernate/search/engine/impl/ConfigContext.java
yrodiere: Fixed.

engine/src/main/java/org/hibernate/search/engine/impl/ConfigContext.java
yrodiere: Fixed.

engine/src/main/java/org/hibernate/search/engine/metadata/impl/TypeMetadata.java
yrodiere: Fixed.

engine/src/main/java/org/hibernate/search/engine/metadata/impl/TypeMetadata.java
yrodiere: Fixed.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/analyzer/impl/CustomElasticsearchAnalyzerImpl.java
```diff
@@ -0,0 +1,53 @@
+/*
```
yrodiere: Fixed.

yrodiere:
To sum up, there are three main topics remaining:

 1. We have analyzer/analyzerReference mixups: https://github.com/hibernate/hibernate-search/pull/1263#discussion_r92823889 . @gsmet , my last commit should solve that issue. Does it seem ok to you? 
 2. Whether file parameters should be forwarded as is (and files should exist on the Elasticsearch instance), or we should parse those files and forward their content, adequately formatted in JSON: https://github.com/hibernate/hibernate-search/pull/1263#discussion_r92822304
 3. Whether we pass parameters to Elasticsearch-specific analysis factories using a) multiple, small parameters with the same name as in Elasticsearch or b) one big JSON object parameter with a magic name: https://github.com/hibernate/hibernate-search/pull/1263#discussion_r92823889 . My opnion is we should leave it as is, but we could also provide both to users and let them choose.

The first one should be addressed in my latest commit, and I'll try to address the second right now (but it's getting late, I probably won't have time to finish today).

@Sanne, any opinion on this?

Sanne:
Thanks!
My quick opinion on your questions, bearing in mind I didn't review the PR yet:
1. Not sure what you're talking about, I guess I'll see during the review.
2. That's a nice idea, but I'd classify it as "future && maybe" improvement. No need for JIRAs, we can see how it plays out.
3. I'd go with the most intuitive approach from an end user perspective, but I don't have enough experience with ES Analyzer definitions to judge. Happy to follow your suggestion; I'll merely check to see how maintainable this looks like.

onward with some hands-on tests..

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/settings/impl/model/AnalyzerDefinitionJsonAdapterFactory.java
gsmet: OK.

engine/src/main/java/org/hibernate/search/analyzer/impl/LuceneAnalyzerReference.java
yrodiere: That's what I would expect, yes: if a subclass wants to expose more of its implemented interfaces, it should do it explicitly.

That's the point of this whole `is()/`/`unwrap` pattern in this case: we don't use adaptors, so simple `instanceof` and client-side casts would have done the trick. The only reason to implement that pattern is to be a bit cleaner and to be more strict about encapsulation. See `org.hibernate.jpa.spi.AbstractEntityManagerImpl.unwrap(Class<T>)` for instance, it's even more strict than what I propose (it only accepts interface types).

Anyway... that's what I think, but it's more nitpicking than anything else, really. I don't think it'll make a difference. I'll use `getClass`, so we'll be safe if someone adds a subclass.

yrodiere:
@Sanne FYI I'm still working on adding better support for file parameters for analysis factories in HSEARCH-2219.
I just discovered that some analysis-related tests did not actually test indexing, and it seems the syntax for those files is slightly different for Elasticsearch (especially char mapping rules, where ES doesn't expect surrounding double-quotes, while Lucene does).
All in all, @gsmet's solution of parsing files on the Hibernate Search side and sending the content as JSON arrays seems to be more and more appropriate, especially given the amount of work required for users to convert their files.

Sanne:
@yrodiere you mean more improvement commits are expected to appear on this PR soon right?

Or shall we look into merging this, and then open follow-up improvement JIRAs as needed?

yrodiere:
@Sanne I just pushed the improvements. I kept them in separate commits to make that readable.
We could handle this separately, but probably not after the CR: the behavior we are discussing here (the one regarding files) could not be changed without breaking users.

Other matters are less critical. I'll open separate PRs. I'll move the commit about analyzer/reference mixups to another PR, too, since Guillaume is "not particularly happy with the solution" and I'll have to change it anyway.

Sanne:
merged. great job all! I was quite worried about us attempting this "translation" but it's looking much better than what I was expecting.

PR #1264, opened 2016-12-19
Race condition fixes and optimisations for concurrent lifecycle changes on Service instances

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-1589
- https://hibernate.atlassian.net/browse/HSEARCH-2484

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/service/impl/StandardServiceManager.java
+++ b/engine/src/main/java/org/hibernate/search/engine/service/impl/StandardServiceManager.java
@@ -72,8 +72,9 @@ public class StandardServiceManager implements ServiceManager {
 		}
 
 		// provided services have priority over managed services
-		if ( providedServices.containsKey( serviceRole ) ) {
-			return (S) providedServices.get( serviceRole );
+		final Object providedService = providedServices.get( serviceRole );
+		if ( providedService != null ) {
+			return (S) providedService;
 		}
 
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/hibernate/search/engine/service/impl/StandardServiceManager.java
```diff
@@ -221,7 +221,7 @@ synchronized void stopVirtual() {
 				if ( status != ServiceStatus.RUNNING ) {
 					stateExpectedFailure();
 				}
-				stopAndRemoveFromCache();
+				stopReal();
```
yrodiere: If I understand correctly, this commit also make the wrappers remain in the cache when they are stopped. But I suppose it is on purpose?

Sanne: Yes, in fact that was the problem: one thread A would perform a stop operation and then remove the Service from the cache, while another thread B had already got a reference to it, yet had not started it. It will then start it (virtually), however by doing so it would increment the usage counter on a service wrapper which is no longer tracked by the registry.
Subsequent operations would initialize a new service wrapper, as the one in the registry was gone, and so the +1 and -1 operations end up being unbalanced.

Sanne: Take it as a testimony that all this stuff was really not meant to do service start/stop frequently at runtime ;-)

Sanne:
thanks, merged.

PR #1265, opened 2016-12-19
HSEARCH-2095 LuceneQueryTranslator is forcing to export the whole imp…

Sanne:
…l package to OSGi headers

Looks like procrastination paid off in this case: I could simply remove the OSGi export statement, this was probably resolved by some of our other changes.

@gunnarmorling am I missing something maybe?

https://hibernate.atlassian.net/browse/HSEARCH-2095

### DIFF
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -226,7 +226,6 @@
                             org.hibernate.search.query.dsl.sort;version="${project.version}",
                             org.hibernate.search.query.engine;version="${project.version}",
                             org.hibernate.search.query.engine.spi;version="${project.version}",
-                            org.hibernate.search.query.engine.impl;version="${project.version}",
                             org.hibernate.search.query.facet;version="${project.version}",
                             org.hibernate.search.query.fieldcache;version="${project.version}",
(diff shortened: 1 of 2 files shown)


yrodiere:
@Sanne It's a bit odd, because `LuceneQueryTranslator` is definitely referenced from the Elasticsearch module: it is implemented in `org.hibernate.search.elasticsearch.impl.ElasticsearchLuceneQueryTranslator`.
Are we sure the Elasticsearch module is used in OSGi tests?

Sanne:
> Are we sure the Elasticsearch module is used in OSGi tests?

I'm sure the Elasticsearch module is NOT tested within OSGi.

Previously though, the LuceneQueryTranslator class was needed to be exposed (initialized? I don't remember) even when not using Elasticsearch.

yrodiere:
@Sanne That's probably related to 4067e6c9daa8626e2d378259a41a2dc4d5db0c88, where I moved code using this interface from the -orm module to the -engine module.

Sanne:
right, that's probably the reason this works now. Good job ;-)

yrodiere:
Merged, thanks!

PR #1268, opened 2016-12-21
HSEARCH-2527 Avoid mixups of analyzers and analyzer references in analyzer-related code

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2527

@gsmet So, this should address most of your remarks:

 * There isn't any `AnalyzerReference` in the scoped analyzers anymore
 * Copied analyzer references are now initialized immediately when they are built
 * There isn't any synchronization in the scoped analyzer references anymore (as a result of the previous item: it allows to transform the lazy initialization into an explicit initialization when boostrapping)

I could not move the copy builder to the scoped analyzers, because in places where we copy the analyzer, we actually want a reference and not directly an analyzer. References have the additional benefit of offering the `is` method that allows to distinguish between lucene and remote analyzers.


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/analyzer/impl/ScopedElasticsearchAnalyzerReference.java
```diff
@@ -6,26 +6,161 @@
...
+		return analyzer;
+	}
+
+	public boolean isInitialized() {
+		return analyzer != null;
+	}
+
+	public void initialize() {
+		if ( this.analyzer != null ) {
+			throw new AssertionFailure( "An Elasticsearch analyzer reference has been initialized more than once:" + this );
```
gsmet: A space after the dot would be nice.

(Yes, this is the one for which I expect crazy eyes)

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/analyzer/impl/ScopedElasticsearchAnalyzerReference.java
```diff
@@ -6,26 +6,161 @@
...
+			}
+
+			return new ScopedElasticsearchAnalyzer( globalAnalyzer, scopedAnalyzers );
+		}
+	}
+
+	public static class CopyBuilder implements ScopedAnalyzerReference.CopyBuilder {
+
+		private final ScopedElasticsearchAnalyzer baseAnalyzer;
+		private final Map<String, ElasticsearchAnalyzerReference> scopedAnalyzerReferences = new HashMap<>();
```
gsmet: Let's work directly on the analyzers here. We don't need to be on the reference level.

yrodiere:
@gsmet Thanks! It should be ok now.
So we'll just wait for @Sanne to give his opinion on whether we can merge this in a CR1 after the holidays.

engine/src/main/java/org/hibernate/search/engine/impl/AnalyzerReferenceRegistry.java
```diff
@@ -82,11 +80,59 @@ public T getAnalyzerReference(Class<?> analyzerClazz) {
 	}
 
 	public void initialize(Map<String, AnalyzerDef> analyzerDefinitions) {
-		strategy.initializeNamedAnalyzerReferences( referencesByName, analyzerDefinitions );
+		List<AnalyzerReference> references = new ArrayList<>();
```
Sanne: This looks like fine after checking other parts but I found it a bit suspicious:

you seem to put some special effort to collect all references into a single collection, to then invoke the initialize method on all references at once.
I found that suspicious, as relying on that would mean we could have some issues with dynamic discovery of new entities (and hence new analyzers). Still, it doesn't look like the initialization method actually requires to have them all in one single collection.

yrodiere: As far as I can tell, the actual problem is we create new instances of analyzers when dynamically adding new entities, even for analyzers we already discovered. So we end up having multiple references to the same analyzer, pointing to different instances of this analyzer. It's similar for Elasticsearch.
I guess it's not critical, unless you have dozens of analyzers and you dynamically add new entities every two seconds.

This is a pre-existing bug, though, not really the point of this PR. Solving it would require to initialize the `ConfigContext` analyzer registries with the content of the previous version of the search integrator.
Actually I did solve this bug in another branch I'm working on (for other reasons), but that branch is currently blocked because of [an unrelated, ongoing debate over what to do next](http://lists.jboss.org/pipermail/hibernate-dev/2017-January/015704.html).
You can see the solution on [my HSEARCH-2534 branch](https://github.com/yrodiere/hibernate-search/commits/HSEARCH-2534), the commit https://github.com/yrodiere/hibernate-search/commit/af1642f06de212ed4ab130cff7dafd2786cf49c6 in particular, and [here](https://github.com/yrodiere/hibernate-search/commit/af1642f06de212ed4ab130cff7dafd2786cf49c6#diff-a85290d5c89683cd208360a89c065a51R122) to be very precise.

Sanne:
nice, merged it. Thanks!

PR #1269, opened 2016-12-22
HSEARCH-2531 Cannot override index names through configuration properties with Elasticsearch

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2531


Sanne:
I merged this earlier, sorry forgot to close.

PR #1270, opened 2016-12-23
HSEARCH-2519 Support analyzer definitions with the Elasticsearch VALIDATE strategy

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2519

Some remarks:

 * Whether we should merge that during a CR still has to be determined. It's probably too much change, but the PR can still be merged later (5.6.0.CR2? 5.7.0 ? 5.6.1? 5.8?)
 * The commit `HSEARCH-2519 Reduce code duplication in DefaultElasticsearchSchemaValidator` is not strictly necessary; I thought it would be cleaner that way, but we can remove this commit without affecting the purpose of this PR.
 * I chose not to handle default values in parameters at all, which means there will be "false negatives" when a parameter is not mentioned in the Hibernate Search definition but is explicitly set to its default value in the current Elasticsearch definition.
   To avoid these false negatives, we'd have to rely on an extensive catalog of analysis types (analyzer types, char filter types, tokenizer types, ...), their expected parameters and the default value for each parameter.
   I don't think that would be wise, considering that different Elasticsearch versions will probably have different analysis types  and different parameters, and also considering that Elasticsearch plugins can add analysis types. We can't seriously hope to have an accurate catalog, so we may as well not even try.
 * I chose to support unordered array for some select analysis types, because there wasn't many of them (on contrary to default values which are ubiquitous).


Sanne:
excellent, seems way more comprehensive than what I'd have expected. Love the good test coverage too!
merged

PR #1271, opened 2016-12-23
HSEARCH-2520 Support analyzer definitions with the Elasticsearch MERGE strategy

yrodiere:
~~This PR is based on #1270 (HSEARCH-2519), which should be merged first.~~ => Done

Relevant JIRA ticket: https://hibernate.atlassian.net/browse/HSEARCH-2520

Whether we should merge that during a CR still has to be determined. It's probably too much change, but the PR can still be merged later (5.6.0.CR2? 5.7.0 ? 5.6.1? 5.8?)

The merge strategy now closes indexes when updating analyzer definitions. This is mandatory, we can't update (or even add) analyzer definitions without closing indexes.

This means in particular that, in environments where multiple clients access a same Elasticsearch index, users will have to set up some sort of synchronization mechanisms to avoid other clients to access the index while Hibernate Search is updating the analyzer definitions.

Handling this synchronization ourselves seems like a bad idea for multiple reasons. This would basically boil down to making the JGroup clustering mandatory when using multiple Hibernate Search instances even with Elasticsearch, which defeats one of the purposes of using Elasticsearch in the first place.

I'd say it's not that bad, because altering analyzer definitions is no small change anyway: in most cases it would require reindexing.

The only thing that bothers me is that we have to stop the index even when simply *adding* an analyzer, which is a shame. Maybe I should submit a feature request to the Elasticsearch team? I'd have to test ES 5 first, though.

EDIT: @gsmet, I looked into throwing specially-crafted exceptions when we have an error because of a closed index. It appears we can't do it with bulk requests, because those may target multiple indexes, which means even if one operation fails because of a closed index, this wouldn't always be true for all operations. So we can't decently make the whole `BulkRequestFailedException` state that the issue was with a closed index.
Knowing that we can only do it for single requests, I wonder if it's worth it? Mind that Elasticsearch error message are really obvious (`"type": "index_closed_exception", "reason": "closed"`).


yrodiere:
Jenkins, test this please

DavideD:
jenkins retest this please

Sanne:
merged!

PR #1273, opened 2017-01-02
HSEARCH-2535 @Facet on container (Iterable, Array, ...) properties indexes the container's toString()

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2535

I added code to support both collections of Strings and collections of numeric values, but only containers of Strings work right now, due to [HSEARCH-1927](https://hibernate.atlassian.net/browse/HSEARCH-1927). This one will have to be fixed in another PR.

I think it's safe to merge this PR during the CR period, because the previous behavior was clearly suboptimal:

 * it simply failed for multi-valued numeric fields, which it still will unless you only have arrays/collections/maps with one element
 * and more importantly it used to index the result of calling `toString()` on arrays/collections/maps for multi-valued string fields, which may work in some cases but clearly isn't a good solution

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
+++ b/engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
@@ -444,7 +444,7 @@ public class DocumentBuilderIndexedEntity extends AbstractDocumentBuilder {
 			ConversionContext conversionContext,
 			InstanceInitializer objectInitializer,
 			final float inheritedBoost,
-			boolean multiValued,
+			boolean isParentPropertyMultiValued,
 			NestingContext nestingContext) {
 
 		// needed for field access: I cannot work in the proxied version
@@ -459,7 +459,7 @@ public class DocumentBuilderIndexedEntity extends AbstractDocumentBuilder {
 				objectInitializer,
 				inheritedBoost,
 				unproxiedInstance,
-				multiValued
+				isParentPropertyMultiValued
 		);
 
 		// allow analyzer override for the fields added by the class and field bridges
@@ -477,7 +477,7 @@ public class DocumentBuilderIndexedEntity extends AbstractDocumentBuilder {
 				objectInitializer,
 				inheritedBoost,
 				unproxiedInstance,
-				multiValued,
+				isParentPropertyMultiValued,
 				nestingContext
 		);
 	}
@@ -491,7 +491,7 @@ public class DocumentBuilderIndexedEntity extends AbstractDocumentBuilder {
 			InstanceInitializer objectInitializer,
 			float inheritedBoost,
 			Object unproxiedInstance,
-			boolean multiValued,
+			boolean isParentPropertyMultiValued,
 			NestingContext nestingContext) {
 		for ( EmbeddedTypeMetadata embeddedTypeMetadata : typeMetadata.getEmbeddedTypeMetadata() ) {
 			XMember member = embeddedTypeMetadata.getEmbeddedGetter();
@@ -585,7 +585,7 @@ public class DocumentBuilderIndexedEntity extends AbstractDocumentBuilder {
 								conversionContext,
 								objectInitializer,
 								embeddedBoost,
-								multiValued,
+								isParentPropertyMultiValued,
 								nestingContext
 						);
 						break;
@@ -604,7 +604,7 @@ public class DocumentBuilderIndexedEntity extends AbstractDocumentBuilder {
 	}
 
 	/**
-	 * @param multiValued Whether the type whose properties should be added may appear more than once (within the same
+	 * @param isParentPropertyMultiValued Whether the type whose properties should be added may appear more than once (within the same
 	 * role) in a document or not. That's the case if the type is (directly or indirectly) contained within an embedded
 	 * to-many association.
 	 */
@@ -615,7 +615,7 @@ public class DocumentBuilderIndexedEntity extends AbstractDocumentBuilder {
 			InstanceInitializer objectInitializer,
 			float documentBoost,
 			Object unproxiedInstance,
-			boolean multiValued) {
+			boolean isParentPropertyMultiValued) {
 		XMember previousMember = null;
 		Object currentFieldValue = null;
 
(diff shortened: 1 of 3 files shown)


yrodiere:
Jenkins, test this please

engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
```diff
@@ -444,7 +444,7 @@ private void buildDocumentFields(Object instance,
 			ConversionContext conversionContext,
 			InstanceInitializer objectInitializer,
 			final float inheritedBoost,
-			boolean multiValued,
+			boolean multiValuedParent,
```
gsmet: I'm not really convinced it's clearer. What is "Parent"?

engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
```diff
@@ -661,10 +661,33 @@ else if ( currentFieldValue instanceof Map ) {
 					if ( fieldMetadata.hasFacets() ) {
 						faceting.enableFacetProcessing();
 						for ( FacetMetadata facetMetadata : fieldMetadata.getFacetMetadata() ) {
-							if ( multiValuedParent ) {
```
gsmet: s/facetting/faceting/ in the commit comment

orm/src/test/java/org/hibernate/search/test/query/facet/MultiValuedFacetingTest.java
```diff
@@ -0,0 +1,354 @@
+/*
```
gsmet: s/facetting/faceting/ in the commit comment

orm/src/test/java/org/hibernate/search/test/query/facet/MultiValuedFacetingTest.java
```diff
@@ -0,0 +1,354 @@
...
+		tc.end();
+	}
+
+	private void assertFacet(Facet facet, String expectedValue, int expectedCount) {
+		assertEquals( "Wrong facet value", expectedValue, facet.getValue() );
+		assertEquals( "Wrong facet count", expectedCount, facet.getCount() );
+	}
+
+	@Indexed
+	static class StringArrayFacetEntity {
```
gsmet: Any reason why they are not private? I wouldn't expect them to be used elsewhere in the package.

orm/src/test/java/org/hibernate/search/test/query/facet/MultiValuedFacetingTest.java
```diff
@@ -124,6 +125,7 @@ public void stringMap() throws Exception {
 	}
 
 	@Test
+	@Ignore // HSEARCH-1927 : Range faceting on multiple numeric values does not work
```
gsmet: Yeah, it's not a good idea to commit tests not working then ignore them. It's especially painful when you try to bisect something.

engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
```diff
@@ -661,10 +661,33 @@ else if ( currentFieldValue instanceof Map ) {
...
+							}
+							else if ( member.isCollection() && currentFieldValue instanceof Map ) {
+								multivalued = true;
+								for ( Object element : ((Map<?,?>) currentFieldValue).values() ) {
+									addFacetDocValues( document, fieldMetadata, facetMetadata, element );
+								}
+							}
+							else if ( member.isArray() ) {
+								multivalued = true;
+								for ( Object element : (Object[]) currentFieldValue ) {
```
gsmet: So we don't expect array of primitives here?

engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
yrodiere: Feel free to suggest a better name. I just meant to mention that it's not the property we're processing that is multi-valued, but its parent property.
What's important is that even though the parent is not multi-valued, the property we're processing still can be.

engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
yrodiere: Done

engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
yrodiere: See my other PR: https://github.com/hibernate/hibernate-search/pull/1274

`@Facet` does not work if we don't also have a `@Field`. Which means supporting arrays of primitives for facets makes little sense if we have no support for arrays of primitives in the field bridges.

Plus, I wanted to discuss the whole solution of `Array.*` in a single PR, because it's likely to raise some objections.

orm/src/test/java/org/hibernate/search/test/query/facet/MultiValuedFacetingTest.java
yrodiere: Done.

orm/src/test/java/org/hibernate/search/test/query/facet/MultiValuedFacetingTest.java
yrodiere: I originally used Hibernate ORM in the test, and Hibernate ORM cannot instantiate private classes.
But you're right, since the test doesn't use ORM anymore, there's no reason for making the class package-scoped.
Fixed.

orm/src/test/java/org/hibernate/search/test/query/facet/MultiValuedFacetingTest.java
yrodiere: Ok. I'll squash that into the previous commit.

gsmet:
I fixed a variable name for consistency and merged (s/multivalued/multiValued/).

Thanks.

PR #1274, opened 2017-01-02
HSEARCH-2510 ArrayBridge doesn't support arrays of primitive types

yrodiere:
This PR is based on #1273 (HSEARCH-2535), which will have to be merged first.

Relevant ticket: https://hibernate.atlassian.net/browse/HSEARCH-2510

Not sure it's the right time to merge this, but I had to investigate and found a solution while working on HSEARCH-2535, so I though we might as well discuss it.

~~The main drawback of the solution is performance: it appears `Array.*` static methods are notoriously slow, as [discussed there](http://stackoverflow.com/questions/30306160/performance-of-java-lang-reflect-array) and [reported there](https://bugs.openjdk.java.net/browse/JDK-8051447]. If it's an issue, I can set up our own utils in Java: it appears that would be faster. Not sure that's worth it, though: yes, these methods are slower than using the built-in equivalents (`[]` operator and `array.length`) but I suppose the latter are pretty fast to begin with.~~ => Addressed by adding utils to `CollectionHelper`.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/bridge/builtin/ArrayBridge.java
+++ b/engine/src/main/java/org/hibernate/search/bridge/builtin/ArrayBridge.java
@@ -11,6 +11,7 @@ import org.apache.lucene.document.Document;
 import org.hibernate.search.bridge.ContainerBridge;
 import org.hibernate.search.bridge.FieldBridge;
 import org.hibernate.search.bridge.LuceneOptions;
+import org.hibernate.search.util.impl.CollectionHelper;
 
 /**
  * Each entry ({@code null included}) of an array is indexed using the specified {@link org.hibernate.search.bridge.FieldBridge}.
@@ -43,8 +44,8 @@ public class ArrayBridge implements FieldBridge, ContainerBridge {
 	}
 
 	private void indexNotNullArray(String name, Object value, Document document, LuceneOptions luceneOptions) {
-		Object[] collection = (Object[]) value;
-		for ( Object entry : collection ) {
+		// Use CollectionHelper.iterableFromArray to also support arrays of primitive values
+		for ( Object entry : CollectionHelper.iterableFromArray( value ) ) {
 			indexEntry( name, entry, document, luceneOptions );
 		}
 	}
--- a/engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
+++ b/engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
@@ -74,6 +74,7 @@ import org.hibernate.search.engine.nesting.impl.NestingContextFactoryProvider;
 import org.hibernate.search.exception.AssertionFailure;
 import org.hibernate.search.exception.SearchException;
 import org.hibernate.search.spi.InstanceInitializer;
+import org.hibernate.search.util.impl.CollectionHelper;
 import org.hibernate.search.util.impl.InternalAnalyzerUtils;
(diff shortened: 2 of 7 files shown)


engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
```diff
@@ -677,7 +685,9 @@ else if ( member.isCollection() && currentFieldValue instanceof Map ) {
 							}
 							else if ( member.isArray() ) {
 								multivalued = true;
-								for ( Object element : (Object[]) currentFieldValue ) {
+								// Use Array.* to also support arrays of primitive values
+								for ( int i = 0, length = Array.getLength( currentFieldValue ); i < length; ++i ) {
```
gsmet: I think we usually use `i++`

engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
```diff
@@ -635,6 +640,9 @@ else if ( currentFieldValue instanceof Map ) {
 						objectInitializer.initializeMap( (Map) currentFieldValue );
 					}
 				}
+				else if ( member.isArray() && currentFieldValue instanceof Object[] ) {
```
gsmet: Doesn't look related to this commit, does it?

engine/src/main/java/org/hibernate/search/bridge/builtin/ArrayBridge.java
```diff
@@ -43,8 +45,9 @@ public FieldBridge getElementBridge() {
 	}
 
 	private void indexNotNullArray(String name, Object value, Document document, LuceneOptions luceneOptions) {
-		Object[] collection = (Object[]) value;
-		for ( Object entry : collection ) {
+		// Use Array.* to also support arrays of primitive values
+		for ( int i = 0, length = Array.getLength( value ) ; i < length ; ++i ) {
```
gsmet: I would vote for not using `Array` and introduce what's needed in `CollectionHelper` as we did for HV.

engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
yrodiere: It's related, since the commit makes it more likely that arrays will be used, but yeah, it addresses a different issue.

It's just nitpicking anyway: `initializeArray` is a no-op.

engine/src/main/java/org/hibernate/search/engine/spi/DocumentBuilderIndexedEntity.java
yrodiere: Seriously?

yrodiere:
@gsmet About using what's in [HV's CollectionHelper](https://github.com/hibernate/hibernate-validator/blob/master/engine/src/main/java/org/hibernate/validator/internal/util/CollectionHelper.java#L98), ok. I'll introduce another method to create an iterable, though, so we can use foreach loops.

Sanne:
@Hibernate-CI retest this

DavideD:
Jenkins, retest this, please

engine/src/main/java/org/hibernate/search/bridge/builtin/ArrayBridge.java
yrodiere: Done.

gsmet:
Rebased and merged, thanks!

PR #1282, opened 2017-01-05
HSEARCH-2501 Elasticsearch calendar resolution handling is not consistent with Lucene's

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2501

Please note that with Lucene, we always truncate date/calendars with the GMT timezone, regardless of the timezone that has been set on the calendar.

If we are to solve this issue, I think we should solve it in a way that is consistent between ES and Lucene. For instance we could add a configuration option to set the timezone to use when truncating dates ('GMT', '+01:00', 'default' for the JVM's default, ...) and another for calendars (same options + another to instruct to use the calendar's timezone).

See also the thread I started on the hibernate-dev mailing list: http://lists.jboss.org/pipermail/hibernate-dev/2017-January/015712.html

For now, I made the Elasticsearch integration behave the same as Lucene.


Sanne:
Great, it even worked in the late evening :+1: 
Merged

PR #1283, opened 2017-01-10
HSEARCH-2540 Fix a misleading piece of code

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2540


Sanne:
frankly I think this change is silly but since you all find it confusing I'll apply ;)

PR #1284, opened 2017-01-10
HSEARCH-2522 Elasticsearch backend in the classpath causes eager initialization of non related IndexManagers

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2522

I had to add one method to `org.hibernate.search.engine.spi.EntityIndexBinding` in order to avoid eager initialization of index managers when checking whether they are related to ES or not.
Thus, this commit sort of breaks an SPI. The important question is: are custom implementations of `EntityIndexBinding` actually possible/useful? And that I don't know.
Maybe @gustavonalle (who reported the bug) could tell us whether this interface is implemented somewhere in Infinispan, and whether it's acceptable to add this method or not?

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/ElasticsearchEntityHelper.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/ElasticsearchEntityHelper.java
@@ -9,10 +9,10 @@ package org.hibernate.search.elasticsearch.util.impl;
 import java.util.Collection;
 import java.util.Set;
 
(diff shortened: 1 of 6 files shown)


gustavonalle:
@yrodiere Infinispan does not implement this interface

yrodiere:
Thanks @gustavonalle . I'll try to see whether it's actually possible for people to integrate a custom implementation of `EntityIndexBinding` within Hibernate Search, then.

gsmet:
Cool! That's something I wanted to do from the beginning. Happy we need to do it to fix a real issue :).

@Sanne any objection against the SPI change? If not, it's good to go!

Sanne:
Looks good! Merging..

PR #1285, opened 2017-01-13
HSEARCH-2545 Validation of IndexedEmbedded targeting a numeric document ID fails inside a non-indexed entity

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2545

See the commit comments for details about the solution.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
+++ b/engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
@@ -1040,7 +1040,19 @@ public class AnnotationMetadataProvider implements MetadataProvider {
 		);
 		checkForContainedIn( member, typeMetadataBuilder, parseContext );
 
-		numericFields.validate();
+		/*
+		 * When we skip field bridges, the numeric fields configuration may not
+		 * be aware of all the processed fields, because in this case we actually
+		 * don't care about which field is numeric.
+		 * Thus the validation cannot be performed reliably when we skip field bridges.
+		 * It's not important anyway, because field bridges are only skipped when
+		 * building contained-in metadata, and when we build contained-in metadata
+		 * we always have another pass on the same annotations to build the indexed
+		 * entity metadata.
+		 */
+		if ( !parseContext.skipFieldBridges() ) {
+			numericFields.validate();
+		}
 
 		PropertyMetadata property = propertyMetadataBuilder.build();
 		if ( !property.getFieldMetadataSet().isEmpty() ) {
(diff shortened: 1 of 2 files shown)


orm/src/test/java/org/hibernate/search/test/engine/NumericDocumentIdIndexedEmbeddedTest.java
```diff
@@ -0,0 +1,122 @@
...
+			// Range Queries including lower and upper bounds
+			assertEquals( "Query id ", 1, numericQueryFor( fts, C.class, "b.a.id", a.id, a.id ).size() );
+			tx.commit();
+			fts.clear();
+		}
+	}
+
+	@After
+	public void cleanData() {
+		if ( getTestResourceManager() == null ) {
```
gsmet: Totally unrelated but couldn't we factorize this somehow? I mean a method taking the classes as parameters you directly call in your `@After`.

engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
```diff
@@ -1040,7 +1040,9 @@ private void initializeMemberLevelAnnotations(String prefix,
 		);
 		checkForContainedIn( member, typeMetadataBuilder, parseContext );
 
-		numericFields.validate();
+		if ( !parseContext.skipFieldBridges() ) {
```
Sanne: Puzzling check. I guess I could find out the reasons by debugging the tests, but this means some comments here could be helpful?

yrodiere: It's definitely not the only puzzling thing in this part of the code :D But yes, you're right.  I'll add a comment :)

gsmet: @Sanne it's probably something we should discuss for 6. The annotation metadata provider magic became a little too magic and obscure lately with Es and obscure bug fixing.

orm/src/test/java/org/hibernate/search/test/engine/NumericDocumentIdIndexedEmbeddedTest.java
yrodiere: Actually, this code is probably not even needed, since the default HBM2DDL strategy drops the schema when closing the factory. I'll remove the code and check that we really drop the schema.

yrodiere: @gsmet I just checked: the tables are dropped automatically after the test. So I removed this code. 
I should take more care when I copy/paste...

yrodiere:
@Sanne @gsmet Thanks for the reviews. I think I addressed your concerns, is it ok to merge as is?

The additional comment is there: https://github.com/hibernate/hibernate-search/pull/1285/files#diff-0a572dae381a036d0fc969a5b45a2ac7R1043

gsmet:
@yrodiere +1

yrodiere:
Jenkins, retest this please

yrodiere:
Jenkins, retest this please

PR #1286, opened 2017-01-13
Upgrade to Hibernate ORM 5.2.6

Sanne:
This is for 5.7 only; we could backport the first two commits (HSEARCH-2549, HSEARCH-2550).

 - https://hibernate.atlassian.net/browse/HSEARCH-2380
 - https://hibernate.atlassian.net/browse/HSEARCH-2549
 - https://hibernate.atlassian.net/browse/HSEARCH-2550

Related:
 - https://hibernate.atlassian.net/browse/HHH-11309

### DIFF
--- a/integrationtest/spring/src/test/resources/beans.xml
+++ b/integrationtest/spring/src/test/resources/beans.xml
@@ -69,8 +69,8 @@
                 <entry key="hibernate.generate_statistics" value="true"/>
                 <!-- 2 is TRANSACTION_READ_COMMITTED -->
                 <entry key="hibernate.connection.isolation" value="2"/>
-                <!-- http://docs.jboss.org/hibernate/core/3.3/reference/en/html/architecture.html#architecture-current-session -->
-                <entry key="hibernate.current_session_context_class" value="org.hibernate.context.JTASessionContext"/>
+                <!-- http://docs.jboss.org/hibernate/orm/5.2/userguide/html_single/Hibernate_User_Guide.html#configurations-general -->
+                <entry key="hibernate.current_session_context_class" value="jta"/>
 
                 <!--  trying things to make Oracle happy -->
                 <entry key="hibernate.jdbc.batch_size" value="0"/>
--- a/integrationtest/wildfly/pom.xml
+++ b/integrationtest/wildfly/pom.xml
@@ -374,6 +374,39 @@
                             </resources>
                         </configuration>
                     </execution>
+                    <!-- Temporary workaround for HHH-11394 -->
+                    <execution>
+                        <id>orm-module-patch-node1</id>
+                        <phase>pre-integration-test</phase>
+                        <goals>
+                            <goal>copy-resources</goal>
+                        </goals>
+                        <configuration>
+                            <outputDirectory>${project.build.directory}/node1/wildfly-${wildflyVersion}/modules/org/hibernate/${hibernateVersion}/</outputDirectory>
+                            <overwrite>true</overwrite>
+                            <resources>
+                                <resource>
+                                    <directory>${basedir}/src/orm-module-override</directory>
+                                </resource>
+                            </resources>
+                        </configuration>
+                    </execution>
+                    <execution>
+                        <id>orm-module-patch-node2</id>
+                        <phase>pre-integration-test</phase>
+                        <goals>
+                            <goal>copy-resources</goal>
+                        </goals>
+                        <configuration>
+                            <outputDirectory>${project.build.directory}/node2/wildfly-${wildflyVersion}/modules/org/hibernate/${hibernateVersion}/</outputDirectory>
+                            <overwrite>true</overwrite>
+                            <resources>
+                                <resource>
+                                    <directory>${basedir}/src/orm-module-override</directory>
+                                </resource>
+                            </resources>
+                        </configuration>
+                    </execution>
                 </executions>
             </plugin>
         </plugins>
--- a/orm/src/main/java/org/hibernate/search/FullTextQuery.java
+++ b/orm/src/main/java/org/hibernate/search/FullTextQuery.java
@@ -13,7 +13,6 @@ import org.apache.lucene.search.Filter;
 import org.apache.lucene.search.Sort;
 
 import org.hibernate.Criteria;
-import org.hibernate.Query;
 import org.hibernate.query.spi.QueryImplementor;
 import org.hibernate.search.engine.ProjectionConstants;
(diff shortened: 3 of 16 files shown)


Sanne:
contains commits from September & October last year :-/

orm/src/test/java/org/hibernate/search/test/util/DdlTransactionIsolatorTestingImpl.java
```diff
@@ -0,0 +1,85 @@
...
+import org.hibernate.resource.transaction.backend.jdbc.internal.DdlTransactionIsolatorNonJtaImpl;
+import org.hibernate.service.ServiceRegistry;
+import org.hibernate.tool.schema.internal.exec.JdbcConnectionAccessConnectionProviderImpl;
+import org.hibernate.tool.schema.internal.exec.JdbcConnectionAccessProvidedConnectionImpl;
+import org.hibernate.tool.schema.internal.exec.JdbcContext;
+
+/**
+ * Copied from Hibernate ORM: being originally part of the `org.hibernate.test.util` package,
+ * this is not published in a jar which could be consumed by Hibernate Search.
+ * This copy is not meant to stay here for long.
```
Sanne: I agreed with Steve that we should move the `DdlTransactionIsolatorTestingImpl.java` file to ORM testing utilities, so we'll be able to depend on it. Not least, it means they'll fix the file in case there's need for further changes ;)

So I prefer to use a plain copy of their class, to that now I know which class exactly I need to move.

orm/src/main/java/org/hibernate/search/query/hibernate/impl/FullTextQueryImpl.java
```diff
@@ -390,12 +390,12 @@ protected boolean isNativeQuery() {
 	}
 
 	@Override
-	public Query setEntity(int position, Object val) {
+	public org.hibernate.query.Query<R> setEntity(int position, Object val) {
```
Sanne: Why would that be simpler?

I didn't put a great deal of thought here, as both alternatives are deprecated, and we don't even implement it ;)

integrationtest/spring/src/test/resources/beans.xml
```diff
@@ -70,7 +70,7 @@
                 <!-- 2 is TRANSACTION_READ_COMMITTED -->
                 <entry key="hibernate.connection.isolation" value="2"/>
                 <!-- http://docs.jboss.org/hibernate/core/3.3/reference/en/html/architecture.html#architecture-current-session -->
-                <entry key="hibernate.current_session_context_class" value="org.hibernate.context.JTASessionContext"/>
+                <entry key="hibernate.current_session_context_class" value="org.hibernate.context.internal.JTASessionContext"/>
```
Sanne: > Or is there a particular reason to reference the class explicitly?

Yes, me not knowing about the `jta` configuration option ;-)

Thanks, I'll change it.

integrationtest/wildfly/pom.xml
```diff
@@ -374,6 +374,39 @@
                             </resources>
                         </configuration>
                     </execution>
+                    <!-- Temporary workaround for HHH-11394 -->
```
Sanne: right, thanks!

Sanne: I didn't mean it as a strict rule, more like something to keep in mind.

New rule:
 - there never are strict rules.

I typically see value in keeping `git bisect` working fine; this is more valuable in maintenance branches as the likelyhood for it to be broken anyway during a large feature sprint is high.

There's also value in keeping changes indipendent. In this case I felt keeping each step of the ORM upgrade independent was making it clearer both for you to review, and for future reference.

If you feel strongly about it I can squash them all; what do you prefer?

orm/src/main/java/org/hibernate/search/query/hibernate/impl/FullTextQueryImpl.java
yrodiere: It would be simpler because `FullTextQuery` is what we want to implement in this file, while `org.hibernate.query.Query` is something we happen to inherit from indirectly. And that has the same name as another class we already reference in the same file.

Anyway, it's no important enough to argue about it... Let's keep it that way.

integrationtest/wildfly/pom.xml
yrodiere: I don't mind either solution. I just thought you may have overlooked this potential issue with `git bisect`, but obviously you haven't. So let's keep it that way :)

Sanne:
@yrodiere thanks for the review. I addressed your remarks and added one more commit to solve 

 - https://hibernate.atlassian.net/browse/HSEARCH-2551

orm/src/main/java/org/hibernate/search/FullTextQuery.java
```diff
@@ -13,7 +13,6 @@
 import org.apache.lucene.search.Sort;
 
 import org.hibernate.Criteria;
-import org.hibernate.Query;
```
yrodiere: It seems this import was being used in the javadocs, and the build now fails.
See lines 153 and 159.

orm/src/main/java/org/hibernate/search/FullTextQuery.java
```diff
@@ -29,7 +28,7 @@
  * @author Hardy Ferentschik
  * @author Emmanuel Bernard
  */
-public interface FullTextQuery<R> extends Query<R>, ProjectionConstants, QueryImplementor<R> {
```
yrodiere: It's a small detail really, but since there's already an issue with this commit: `QueryImplementor` already extends `org.hibernate.query.Query` (and `org.hibernate.Query`, indirectly), so we could remove the explicit ` extends org.hibernate.query.Query<R>`. What do you think?

orm/src/main/java/org/hibernate/search/FullTextQuery.java
Sanne: damn :) I really need to figure out why such things don't fail to me locally.

orm/src/main/java/org/hibernate/search/FullTextQuery.java
Sanne: Interesting, I'm going to try that as it looks promising. FYI this wasn't possible in earlier versions of ORM 5.2.x, but those SPIs evolved a bit also based on our earlier attempts to migrate Search..

yrodiere:
Merged. Thanks!

PR #1287, opened 2017-01-13
HSEARCH-2547 Nesting @IndexedEmbeddeds with "includePaths" results in invalid path restrictions

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2547

So... I think the first two commits should not raise many objections, since it's basically about allowing users to do something that previously failed miserably for no apparent reason.

The two last commits, on the other hand, will probably be less consensual.
Basically, we used to allow an `@IndexedEmbedded` on a property `b` of type `B` to include paths that were not included in type `B`. See the test for an example.
I think this should be forbidden, mainly because it feels like a violation of encapsulation to me, but also because the current behavior of `@ContainedIn` will not work well with such setups.

You'll note I had to fix a test to make it pass after my "fix", so breakage for existing users is to be expected.
Maybe we should fix this in 6.0, and not take the risk to break existing users?

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/metadata/impl/PathsContext.java
+++ b/engine/src/main/java/org/hibernate/search/engine/metadata/impl/PathsContext.java
@@ -24,12 +24,20 @@ public class PathsContext {
 		pathsEncounteredState.put( path, Boolean.FALSE );
 	}
 
+	public boolean isIncluded(String path) {
+		return pathsEncounteredState.keySet().contains( path );
+	}
+
 	public boolean isIncluded(DocumentFieldPath path) {
-		return pathsEncounteredState.keySet().contains( path.getAbsoluteName() );
+		return isIncluded( path.getAbsoluteName() );
+	}
+
+	public void markEncounteredPath(String path) {
+		pathsEncounteredState.put( path, Boolean.TRUE );
 	}
 
 	public void markEncounteredPath(DocumentFieldPath path) {
-		pathsEncounteredState.put( path.getAbsoluteName(), Boolean.TRUE );
+		markEncounteredPath( path.getAbsoluteName() );
 	}
 
 	public Set<String> getEncounteredPaths() {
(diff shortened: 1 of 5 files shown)


Sanne:
> we used to allow an @IndexedEmbedded on a property b of type B to include paths that were not included in type B. See the test for an example.
I think this should be forbidden, mainly because it feels like a violation of encapsulation to me, but also because the current behavior of @ContainedIn will not work well with such setups.

That's a very interesting point. From one side, I agree with you that this shouldn't happen. On the other, I can see how someone needing to "fine tune" things might want to do that, so if we "fix" it by not doing it, we might need to introduce a specific feature for such a mapping.

You guessed right I don't think we should merge the second change at this stage; could you split that in its own issue, so then we can decide if it's material for 6?

Sanne:
I guess the culprit is to clarify if `@IndexedEmbedded(includePaths)` is a _narrowing_ filter, or indicative.

When creating it I strictly had the _narrowing_ idea in mind, but I can see how the _include_ Paths terminology might have misled some people. For what it's worth, I realize this feature has been around for some time but I still feel like it's "bolted on" as we never revisited it from the first POC.. it's been a long time that I want to rethink the whole DocumentBuilder process from scratch.

engine/src/main/java/org/hibernate/search/engine/metadata/impl/PathsContext.java
```diff
@@ -24,12 +24,20 @@ public void addIncludedPath(String path) {
 		pathsEncounteredState.put( path, Boolean.FALSE );
 	}
 
+	public boolean isIncluded(String path) {
+		return pathsEncounteredState.keySet().contains( path );
```
Sanne: This could be:

    return pathsEncounteredState.containsKey( path );

Sanne:
I've merged the first two commits, which resolve the ticket. 

@yrodiere don't throw the useful work of the other two commits away :)

PR #1288, opened 2017-01-18
HSEARCH-2552 @IndexedEmbedded.includePaths allows to pick paths that were excluded from the embedded entity

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2552

Please do not merge yet, we'll probably have to wait until 6.0 because this changes the behavior of `@IndexedEmbedded`.
Submitting this PR anyway so that we don't lose the commits.


gsmet:
@yrodiere is this something you have already fixed in 6? If so, maybe we can close this PR?

yrodiere:
@gsmet Right, closed.
It's probably solved in Search 6, but not tested explicitly yet, so I'll leave the ticket open for now.

PR #1289, opened 2017-01-20
[WIP] HSEARCH-2358 Use field storage instead of source document for projections

yrodiere:
**This is work in progress**, and the tests currently fail. I'm just opening this PR for discussion.

Relevant tickets:

 * https://hibernate.atlassian.net/browse/HSEARCH-2423
 * https://hibernate.atlassian.net/browse/HSEARCH-2358

So, here we change how projections work with Elasticsearch by using Elasticsearch's [field storage](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/search-request-fields.html) feature instead of [source documents](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/search-request-source-filtering.html) as a data source.
Those features are quite similar, with the main difference being the field storage is configurable per property in the mapping, while source document storage is configurable per index only (or maybe only globally).
Presumably, the data structures behind field storage may be a bit more advanced and efficient, since they are probably strongly typed, but I haven't checked that.

Anyway... as shown on [this build](https://travis-ci.org/yrodiere/hibernate-search/builds/193696440), switching to field storage brings a few challenges:

 * Elasticsearch doesn't store time zone information for dates, so we lose some information when storing/retrieving some types of fields, such as `Calendar` or `ZonedDateTime`. This could be solved by adding bridge-defined fields containing only the time zone/offset for such types.
  See how `ElasticsearchDateCalendarBridgeIT` or `JavaTimeTest.testZonedDateTime` fail, for instance.
 * We use the `indexNullAs` feature in Hibernate Search to populate the `null_value` attribute in the ES mapping, which has the effect of making documents with a null in a field match some search queries on this field, by assuming that null is in fact some pre-defined value.
  Unfortunately, this pre-defined value is also returned for null fields when retrieving the stored value. So `ElasticsearchIndexNullAsIT` now fail, because when we store `null`, we expect to retrieve `null` when projecting, but we get the `indexNullAs` instead.
  See how `ElasticsearchIndexNullAsIT` fails, for instance.
 * Elasticsearch seems to store timestamps with millisecond resolution, while date/time objects in Java 8 have nanosecond resolution.
  See how `JavaTimeTest.testLocalTime` fails, for instance.
 * For some reasons spatial projections seem to fail with this patch, but I'm fairly confident it's related to my changes and not to Elasticsearch. I'll look into it if we decide to go this way.
  See how `SpatialIndexingTest` fails, for instance.

### DIFF
--- a/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
+++ b/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
@@ -671,10 +671,7 @@ List results = query.list();
 
 ==== Projections
 
-All fields are stored by Elasticsearch in the JSON document it indexes,
-there is no specific need to mark fields as stored when you want to project them.
-The downside is that to project a field, Elasticsearch needs to read the whole JSON document.
-If you want to avoid that, use the `Store.YES` marker.
+As with Lucene, fields must be marked as stored (`store = Store.YES`) in order to be able to project them.
 
 You can also retrieve the full JSON document by using `org.hibernate.search.elasticsearch.ElasticsearchProjectionConstants.SOURCE`.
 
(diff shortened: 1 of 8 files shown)


yrodiere:
This PR is completely obsolete now that we've overhauled the Elasticsearch backend in Search 6. The ticket is still relevant, though. Closing the PR only.

PR #1290, opened 2017-01-20
HSEARCH-2554 Using array/iterable/map field bridges requires the @IndexedEmbedded annotation (second attempt)

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2554

Please do not merge yet, we'll probably have to wait until 6.0 because this changes the behavior of Hibernate Search for `@IndexedEmbedded`-annotated fields.
Submitting this PR anyway so that we don't lose the commits.


gsmet:
@yrodiere IIRC you already implemented at least part of it if not all of it in Search 6. If so we could close this PR too.

yrodiere:
@gsmet Done.
As for #1288 , this is probably solved in Search 6 but not tested explicitly yet, so I'll leave the ticket open for now.

PR #1291, opened 2017-01-20
HSEARCH-2548 Upgrade to Hibernate ORM 5.2.7.Final

Sanne:
it's missing documentation, I can't do that today but it would be nice to warn people attempting the same about needing to use a `jboss-deployment-structure.xml` to exclude javassist. Opening a follow-up jira.


yrodiere:
Nice! Merged.

yrodiere:
@Sanne By the way, the ticket states "When doing this, consider removing the workaround for issue HHH-11309: we added ByteBuddy to the OSGi dependencies in the features file as it was not listed optional.".
Did you decide against removing the workaround?

Sanne:
right, I decided to not remove that. Sorry should have commented.

By keeping bytebuddy people can choose if they want to use it ;)

PR #1292, opened 2017-01-20
HSEARCH-2555 Ensure the test suite fail if any service isn't properly released

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2555

This also fixes various issues with services not being released properly.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/service/impl/StandardServiceManager.java
+++ b/engine/src/main/java/org/hibernate/search/engine/service/impl/StandardServiceManager.java
@@ -112,19 +112,38 @@ public class StandardServiceManager implements ServiceManager {
 	}
 
 	@Override
-	public void releaseAllServices() {
-		List<String> unreleasedServicesToReport = failOnUnreleasedService ? new ArrayList<String>() : null;
-
+	public synchronized void releaseAllServices() {
 		for ( ServiceWrapper<?> wrapper : cachedServices.values() ) {
-			boolean stoppedProperly = wrapper.ensureStopped();
-			if ( !stoppedProperly ) {
+			/*
+			 * Perform an additional stopVirtual, to remove the extra usage token granted at first initialization,
+			 * which keeps the service to be really stopped when it's released by the service client, yet we're not shutting down
+			 * the Search engine yet.
+			 */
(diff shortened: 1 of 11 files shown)


yrodiere:
Jenkins, test this please

engine/src/main/java/org/hibernate/search/engine/service/impl/StandardServiceManager.java
```diff
@@ -58,6 +62,7 @@ public StandardServiceManager(SearchConfiguration searchConfiguration,
 		this.providedServices = createProvidedServices( searchConfiguration );
 		this.defaultServices = defaultServices;
 		this.classloaderService = searchConfiguration.getClassLoaderService();
+		this.failOnUnreleasedService = Boolean.parseBoolean( System.getProperty( "org.hibernate.search.fail_on_unreleased_service" ) );
```
Sanne: there's a short version for that:

    Boolean.getBoolean( "org.hibernate.search.fail_on_unreleased_service" );

engine/src/main/java/org/hibernate/search/engine/service/impl/StandardServiceManager.java
```diff
@@ -108,10 +113,24 @@ public StandardServiceManager(SearchConfiguration searchConfiguration,
 
 	@Override
 	public void releaseAllServices() {
-		for ( ServiceWrapper wrapper : cachedServices.values() ) {
-			wrapper.ensureStopped();
+		List<String> unreleasedServicesToReport = failOnUnreleasedService ? new ArrayList<String>() : null;
+
+		for ( ServiceWrapper<?> wrapper : cachedServices.values() ) {
+			boolean stoppedProperly = wrapper.ensureStopped();
```
Sanne: the wrapper needs to be synchronized here, like the original method which you're inlining now.
Failing to do so might expose you to a service instance to increment its usage counter during this stop operation for a legitimate usage, and yet throw the exception (or log the warning).

It's extremely unlikely as we'd be shutting down anyway.. but it's in theory possible if e.g. a service is needed to perform a clean teardown.
I'll fix that.

Sanne:
merged, thanks! Just added two trivial commits

PR #1294, opened 2017-01-25
Fix the changelog for 5.7.0.CR1

yrodiere:
Some tickets were not moved to the correct "Fix version" before the release, resulting in incorrect changelogs.


Sanne:
thanks! merged

PR #1295, opened 2017-01-26
HSEARCH-2447 Make it easier to add sortable fields from custom field bridges

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2447


Sanne:
merged, in 5.7 exclusively.

PR #1296, opened 2017-01-30
HSEARCH-2559 Switch to AsciiDoctor for the final documentation output

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-2559

This PR is for the master branch.

```
mvn -Pdocumentation-pdf -DskipTests clean install
```

It will build the HTML and PDF output.


Sanne:
I tried it locally: works fine and looks great :+1: 

Looking forward to be able to merge it.

gsmet:
@Sanne it's ready for review!

documentation/pom.xml
```diff
@@ -22,202 +22,219 @@
     <description>The Hibernate Search reference manual</description>
 
     <properties>
+        <asciidoctor.base-outputdir>${basedir}/target/asciidoctor/en-US</asciidoctor.base-outputdir>
```
yrodiere: It's so secondary I feel bad to mention it, but the other property names below have a hyphen before the `dir` keyword, so maybe this should be `base-output-dir`?

yrodiere: Also, about the value of this property, didn't you say you were going to remove the `/en-US` part in because we don't translate the doc anyway? For consistency?

gsmet: Yeah, it's mostly because I don't want to change the hierarchy on the website and the stable URL of the documentations.

I fixed the dir issue (and a small other one I didn't notice before) and force pushed an update.

Sanne:
looks great! One doubt, I didn't double-check.. is the output going to be in the same location, especially for e.g. the release scripts to upload them still correctly?

Sanne:
merged, thanks!

gsmet:
@Sanne yes, everything is already taken care of, in the assembly and in the release scripts. We already did one release of HV with the new documentation.

PR #1297, opened 2017-01-30
HSEARCH-2566 Stop deploying integration testing artifacts to the maven repositories

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2566


yrodiere:
To be discussed. The main argument against this change, in my opinion, is that we already released those artifacts in version 5.7.0.x, for instance 5.7.0.CR1.
But, really, why would someone depend on these artifacts?

Sanne:
These modules already have the "skip deploy" property enabled. Did you find them in the Maven central repositories?

yrodiere:
You're right... I saw [this folder](http://repo1.maven.org/maven2/org/hibernate/hibernate-search-integrationtest-osgi/) and failed to check inside, but this is actually only a remnant of some alpha releases.
Sorry about that, closing.

PR #1298, opened 2017-01-31
HSEARCH-2568 Do no check that the index exists with the NONE schema management strategy

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2568

To be discussed, but it feels a bit strange that the `NONE` strategy does in fact do something (check if the index exists). This PR reverts the `NONE` strategy to the pre-HSEARCH-2456 behavior of not doing anything, not even checking whether the index exists.

I'm not sure we should backport this to 5.6, though, since it's a behavior change.

### DIFF
--- a/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
+++ b/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
@@ -118,7 +118,7 @@ Let's see the options for the `index_schema_management_strategy` property:
 |===============
 |Value|Definition
 |`NONE`|Indexes, their mappings and their analyzer definitions will not be created, deleted nor altered.
-Hibernate Search will only check that the indexes already exist.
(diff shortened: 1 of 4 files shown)


Sanne:
What is the value in "not checking" at all - even not for index existance?

It's not that Hibernate ORM can start when there's no database..

yrodiere:
> What is the value in "not checking" at all - even not for index existance?

The only use case I see is mentioned in the ticket: a user wanting to create the index himself, but after HS started. Strange, but then I guess using `NONE` is also quite strange to begin with...

> It's not that Hibernate ORM can start when there's no database..
True, but that's not particularly because of HBM2DDL. So if we want to mirror that behavior, we shouldn't try to check the index existence inside our strategy.

Sanne:
ok, yes I had merged it after reading Alex Laptseu's comments, but .. I agree it's strange.

PR #1301, opened 2017-02-01
HSEARCH-2573 Upgrade the Elasticsearch version used for tests to 2.4.4

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2573


Sanne:
backported as well.

PR #1302, opened 2017-02-02
HSEARCH-2569 Make the 'delete-by-query' plugin requirement for ES clearer

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2569

This PR brings two changes:

 1. The documentation is clearer about the fact that the 'delete-by-query' plugin is required when using the "purgeAll" option on the mass indexer, and this part of the documentation is more visible.
 2. Whenever a user tries to purge indexes and a 404 error occurs, the error message now mentions that the 'delete-by-query' plugin is required.

We'll probably want to backport at least the first commit to 5.6.

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/DeleteByQueryResultAssessor.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/DeleteByQueryResultAssessor.java
@@ -1,0 +1,52 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.elasticsearch.impl;
+
+import org.hibernate.search.elasticsearch.client.impl.BackendRequestResultAssessor;
+import org.hibernate.search.elasticsearch.logging.impl.Log;
+import org.hibernate.search.exception.AssertionFailure;
+import org.hibernate.search.exception.SearchException;
+import org.hibernate.search.util.logging.impl.LoggerFactory;
+
+import io.searchbox.action.Action;
+import io.searchbox.client.JestResult;
+import io.searchbox.core.BulkResult.BulkResultItem;
+
+
+/**
+ * @author Yoann Rodiere
+ */
+public class DeleteByQueryResultAssessor implements BackendRequestResultAssessor<JestResult> {
+
+	private static final Log LOG = LoggerFactory.make( Log.class );
+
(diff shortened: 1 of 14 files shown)


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/DeleteByQueryResultAssessor.java
```diff
@@ -0,0 +1,53 @@
...
+
+
+/**
+ * @author Yoann Rodiere
+ */
+public class DeleteByQueryResultAssessor implements BackendRequestResultAssessor<JestResult> {
+
+	private static final Log LOG = LoggerFactory.make( Log.class );
+
+	private static final int NOT_FOUNT_HTTP_STATUS_CODE = 404;
```
gsmet: s/FOUNT/FOUND/

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/DeleteByQueryResultAssessor.java
```diff
@@ -0,0 +1,53 @@
...
+
+	private final JestAPIFormatter formatter;
+
+	private final DefaultBackendRequestResultAssessor delegate;
+
+	public DeleteByQueryResultAssessor(JestAPIFormatter formatter) {
+		super();
```
gsmet: No need for the `super()` here.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/DeleteByQueryResultAssessor.java
yrodiere: Fixed.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/DeleteByQueryResultAssessor.java
yrodiere: I'll remove it, but I seriously don't know what's the problem with this.

Sanne:
Looks nice, and +1 to backport: I see no significant risk.

Minor question: why did you opt for `org.hibernate.search.elasticsearch.impl.JestAPIFormatter` to be a Service?

(I couldn't test & build it yet, still figuring out what's wrong with my machine..)

yrodiere:
> Minor question: why did you opt for org.hibernate.search.elasticsearch.impl.JestAPIFormatter to be a Service?

Because it needs access to `GsonService`. I guess I could have made it a constructor parameter, but users would still have had to request/release an instance of `GsonService` (instead of `JestAPIFormatter` right now), so it wouldn't have been simpler in the end.

Sanne:
merged on master and 5.6

PR #1303, opened 2017-02-03
Improve reliability of JGroups integration tests

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2141


yrodiere:
@Sanne If understand correctly, we were trying to detect failures over a loopback interface?
If so, then yes, this is probably a bit too much for tests.
Merged, thanks!

Sanne:
yes that's right. Even worse, it was using the "actual" network interface for that, while we are using the LOOPBACK protocol which is just a static hashmap to avoid trouble with setting up the network interfaces correctly for it..

PR #1304, opened 2017-02-03
Allow repeatable annotations

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2574

### DIFF
--- a/documentation/src/main/asciidoc/mapping.asciidoc
+++ b/documentation/src/main/asciidoc/mapping.asciidoc
@@ -297,7 +297,8 @@ public class Book {
 
 If there is a single `@Field` declared for a given property, `@SortableField` implicitly applies to this field.
 In case several fields exist for a single property, the `@Field` to be marked as sortable can be specified
-via `@SortableField#forField()`. Several sortable fields can be defined with help of the `@SortableFields` annotation.
+via `@SortableField#forField()`.
+Several sortable fields can be defined as the `@SortableField` annotation is repeatable.
 
 The field to be marked as sortable must not be analyzed.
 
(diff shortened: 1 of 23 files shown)


Sanne:
I pushed an improvement.

documentation/src/main/asciidoc/mapping.asciidoc
```diff
@@ -440,20 +441,21 @@ document id, it is recommended to declare another field specifically for sorting
...
 ----
 @Entity
 @Indexed(index = "Book")
 public class Book {
-    @Fields( {
-            @Field,
-            @Field(name = "summary_forSort", analyze = Analyze.NO, store = Store.YES)
-    } )
+
+    @Field,
```
yrodiere: There's an extra comma here.

Sanne: thanks, fixed!

Sanne:
> Here's an error I noticed while reading through the changes.

From now on, you shall be remembered as "eagle eye Yoann" :+1:

Sanne:
I noticed as well, but came to the same conclusions. `TikaBridge` has also an uncertain future IMO: I like the functionality it provides but it brings too many dependencies; I'd like to use it as an example of how we might be able to tightly integrate with "advanced bridges", including possibly having their own annotations, yet have these live in a separate module so that we can isolate the dependencies properly. This might be a strong requirement for producing Jigsaw modules..

In short, I'd not see the benefit on opening new tickets for now.

yrodiere:
Merged! Thanks.
And ok to look at TikaBridge again for HSearch 6.0.

PR #1306, opened 2017-02-06
HSEARCH-2576 FacetingManager should allocate maps lazily

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2576

### DIFF
--- a/engine/src/main/java/org/hibernate/search/query/engine/impl/FacetManagerImpl.java
+++ b/engine/src/main/java/org/hibernate/search/query/engine/impl/FacetManagerImpl.java
@@ -34,12 +34,12 @@ public class FacetManagerImpl implements FacetManager {
 	/**
 	 * The map of currently active/enabled facet requests.
 	 */
-	private final Map<String, FacetingRequest> facetRequests = newHashMap();
+	private Map<String, FacetingRequest> facetRequests;
 
 	/**
 	 * Keep track of the current facet selection groups.
 	 */
-	private final Map<String, FacetSelectionImpl> facetSelection = newHashMap();
+	private Map<String, FacetSelectionImpl> facetSelection;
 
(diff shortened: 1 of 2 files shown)


yrodiere:
Jenkins, test this please

Sanne:
> I suppose you created this ticket following performance testing in Infinispan?

I have been busy with performance testing, but actually not (yet) with Infinispan..

It's actually amazing what a difference this small change makes: we have a flaw in our container based performance tests, which triggers an occasional deadlock in H2 when we're "too fast". After this change, that test fails much more frequently (from "almost never" to "almost all the time" .. on my pc at least).

yrodiere:
> I have been busy with performance testing, but actually not (yet) with Infinispan..

So you used the existing performance testing project from our own code? Great :)

> After this change, that test fails much more frequently (from "almost never" to "almost all the time" .. on my pc at least).

I suppose there's no way to fix this? Maybe upgrade H2? It would be a shame if, after merging this PR, our CI became useless because it would always fail...

Sanne:
> I suppose there's no way to fix this? Maybe upgrade H2? It would be a shame if, after merging this PR, our CI became useless because it would always fail...

Yes, let's keep an eye on this.. I hope not as CI is quite slow. let's not upgrade it! :D More seriously, yes fixing the test is on my watchlist..

> So you used the existing performance testing project from our own code? Great :)

I have some more perf tests as well, being developed. Will need to tidy them up eventually and share, it's currently a mess of experiments and various branches trying different things, e.g. different combinations of dependencies.

Moving ahead we'd want to have nightly performance regression reports built by CI.

yrodiere:
Merged, thanks!

We'll see how well the CI behaves and prioritize this H2 locking bug accordingly.

PR #1307, opened 2017-02-06
HSEARCH-2381 Avoid logging warnings caused by internal usage of the legacy Criteria API

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2381

### DIFF
--- a/orm/src/main/java/org/hibernate/search/batchindexing/impl/IdentifierConsumerDocumentProducer.java
+++ b/orm/src/main/java/org/hibernate/search/batchindexing/impl/IdentifierConsumerDocumentProducer.java
@@ -24,6 +24,7 @@ import org.hibernate.criterion.Restrictions;
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.engine.spi.SessionImplementor;
 import org.hibernate.engine.transaction.jta.platform.spi.JtaPlatform;
+import org.hibernate.internal.CriteriaImpl;
 import org.hibernate.search.backend.AddLuceneWork;
 import org.hibernate.search.backend.spi.BatchBackend;
 import org.hibernate.search.batchindexing.MassIndexerProgressMonitor;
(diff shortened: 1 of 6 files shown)


orm/src/main/java/org/hibernate/search/batchindexing/impl/IdentifierConsumerDocumentProducer.java
```diff
@@ -24,6 +24,7 @@
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.engine.spi.SessionImplementor;
 import org.hibernate.engine.transaction.jta.platform.spi.JtaPlatform;
+import org.hibernate.internal.CriteriaImpl;
```
yrodiere: Are we absolutely sure that this class won't change or (more importantly) get repackaged somewhere else before ORM 6.0? Just asking because of what happened in ORM 5.2...

It would be a shame that some previous version of Search ends up non-compatible with a new (possibly micro) version of ORM just because we wanted to get rid of a warning...

Sanne: no we're not absolutely sure, but this class has been there for ~10 years.

As soon as the 5.7 final is out we'll set up CI jobs, to discuss with the ORM team any breaking changes.

Getting rid of the warning is quite important, both from performance and user confusion perspective, and I couldn't find a better solution.

One alternative would be to patch ORM to introduce an explicit SPI level contract for this, but then you'd need to bump the requirement to > 5.2.8 ... I prefer this solution.

Sanne:
test this please

Sanne:
> test this please

nevermind, this had been triggered normally.

orm/src/main/java/org/hibernate/search/batchindexing/impl/IdentifierConsumerDocumentProducer.java
yrodiere: Ok then, merging. Thanks!

PR #1308, opened 2017-02-07
HSEARCH-2418 (+HSEARCH-2577) Introduce *AnalyzerDefinitionProvider SPIs

yrodiere:
Relevant JIRA tickets:

 * https://hibernate.atlassian.net/browse/HSEARCH-2418
 * https://hibernate.atlassian.net/browse/HSEARCH-2577

As mentioned in the ticket description for HSEARCH-2577, it may not be 100% required to separate the Lucene/ES namespaces in order to fix HSEARCH-2418, but it definitely makes things easier, and it's probably easier on users too. Also, I happened to have already implemented this change for another ticket (which is still in standby), so we may as well merge it.

About the PR itself:

 * I introduced two `*AnalyzerDefinitionProvider` SPIs, one for Lucene and one for Elasticsearch
 * Both definition providers take a builder as a parameter, and simply push their definitions through this builder
 * The SPIs are undocumented for now. Do we want to advertise those to users? I think they could be useful, but on the other hand they are still rather primitive (no support for multiple providers, for index-scoped providers, etc.)
 * In order to integrate the SPI cleanly in ES, I had to move the definition translation sooner, when building the analyzer references, instead of when starting the index managers. It shouldn't have any practical impact on users, though. Maybe a slight slowdown when using the `NONE` index management strategy?
 * Once again, I think it's better to review one commit at a time.


engine/src/main/java/org/hibernate/search/analyzer/impl/LuceneEmbeddedAnalyzerStrategy.java
```diff
@@ -35,7 +35,7 @@
  */
 public class LuceneEmbeddedAnalyzerStrategy implements AnalyzerStrategy {
 
-	private static final Log log = LoggerFactory.make();
+	private static final Log LOG = LoggerFactory.make();
```
Sanne: I'm ok to change this, but TBH I always thought it was weird that you would create `LOG` instances... traditionally we've always just called them `Log`, as it's not really a constant (it's a mutable object).

No need to change anything..

yrodiere:
It seems there's a conflict. I'll rebase.

engine/src/main/java/org/hibernate/search/analyzer/impl/LuceneEmbeddedAnalyzerStrategy.java
yrodiere: True, we could consider that loggers are output streams, and as such they can be considered "mutable" as `System.out` for instance.
I've always followed the rule "static final => capital case", but there's no real reason beyond consistency. And it seems I was wrong in thinking "LOG" is more common: we have 163 occurrences of "static final Log log" vs. 72 occurrences of "static final Log LOG".
Maybe I should try and rename all those loggers to "log"... one day... In the meantime, I'll remember to name them "log".

yrodiere:
Rebased on master just now.

engine/src/main/java/org/hibernate/search/analyzer/impl/LuceneEmbeddedAnalyzerStrategy.java
gsmet: FYI, we're renaming them to `LOG` in HV.

Sanne: I have no strong opinion, for the sake of sharing my view:

- the new LOG ones are the ones you created, I simply didn't complain before but they looked odd ;-)
- I definitely prefer smaller PRs when possible, as my main problem is not to find 3 hours to review PRs, but to find 3 contiguous uninterrupted hours.

Sanne:
Could you add javadoc to the interfaces in the public API?

- org.hibernate.search.elasticsearch.analyzer.definition.ElasticsearchAnalysisComponentDefinitionContext
- org.hibernate.search.elasticsearch.analyzer.definition.ElasticsearchAnalysisDefinitionRegistryBuilder
- org.hibernate.search.elasticsearch.analyzer.definition.ElasticsearchAnalyzerDefinitionContext
- org.hibernate.search.elasticsearch.analyzer.definition.ElasticsearchAnalyzerDefinitionWithTokenizerContext
- org.hibernate.search.elasticsearch.analyzer.definition.ElasticsearchTypedAnalysisComponentDefinitionContext

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/analyzer/definition/ElasticsearchTypedAnalysisComponentDefinitionContext.java
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * @author Yoann Rodiere
+ */
+public interface ElasticsearchTypedAnalysisComponentDefinitionContext {
+
+	ElasticsearchTypedAnalysisComponentDefinitionContext param(String name, String value);
+
+	ElasticsearchTypedAnalysisComponentDefinitionContext param(String name, String... values);
+
+	ElasticsearchTypedAnalysisComponentDefinitionContext param(String name, Boolean value);
```
Sanne: why not using `boolean` (lowercase) ?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/analyzer/definition/ElasticsearchTypedAnalysisComponentDefinitionContext.java
```diff
@@ -0,0 +1,23 @@
...
+
+
+/**
+ * @author Yoann Rodiere
+ */
+public interface ElasticsearchTypedAnalysisComponentDefinitionContext {
+
+	ElasticsearchTypedAnalysisComponentDefinitionContext param(String name, String value);
+
+	ElasticsearchTypedAnalysisComponentDefinitionContext param(String name, String... values);
```
Sanne: I was wondering why the String parameter has a vararg option, and the others don't. I guess you have a specific analyzer need example in mind?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/analyzer/impl/ElasticsearchAnalyzerStrategy.java
```diff
@@ -14,24 +14,66 @@
...
+
+	private ElasticsearchAnalysisDefinitionRegistry createDefaultDefinitionRegistry() {
+		ElasticsearchAnalysisDefinitionRegistryBuilderImpl builder =
+				new ElasticsearchAnalysisDefinitionRegistryBuilderImpl();
+
+		String providerClassName = cfg.getProperty( ElasticsearchEnvironment.ANALYZER_DEFINITION_PROVIDER );
+		if ( providerClassName != null ) {
+			ElasticsearchAnalysisDefinitionProvider provider;
+			try {
+				Class<?> providerClazz = ClassLoaderHelper.classForName( providerClassName, serviceManager );
```
Sanne: could you use `ClassLoaderHelper.instanceFromName` instead ? It would provide more sanity checks and errors consistent with other services.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/logging/impl/Log.java
```diff
@@ -307,18 +307,21 @@ BulkRequestFailedException elasticsearchBulkRequestFailed(String request, String
 	SearchException indexedEmbeddedPrefixBypass(Class<?> entityType, String fieldPath, String expectedParent);
 
 	@Message(id = ES_BACKEND_MESSAGES_START_ID + 55,
-			value = "The same tokenizer name '%1$s' is used multiple times. The tokenizer names must be unique."
-					+ " If this name was automatically generated, you may override it by using @TokenizerDef.name." )
+			value = "The same tokenizer name '%1$s' is assigned to multiple definitions. The tokenizer names must be unique."
+					+ " If your used the @TokenizerDef annotation and this name was automatically generated,"
```
Sanne: "if your" seems a typo. "If you" ?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/logging/impl/Log.java
```diff
@@ -307,18 +307,21 @@ BulkRequestFailedException elasticsearchBulkRequestFailed(String request, String
...
+					+ " you may override this name by using @TokenizerDef.name." )
 	SearchException tokenizerNamingConflict(String remoteName);
 
 	@Message(id = ES_BACKEND_MESSAGES_START_ID + 56,
-			value = "The same char filter name '%1$s' is used multiple times. The char filter names must be unique."
-					+ " If this name was automatically generated, you may override it by using @CharFilterDef.name." )
+			value = "The same char filter name '%1$s' is assigned to multiple definitions. The char filter names must be unique."
+					+ " If your used the @CharFilterDef annotation and this name was automatically generated,"
```
Sanne: same typo

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/logging/impl/Log.java
```diff
@@ -307,18 +307,21 @@ BulkRequestFailedException elasticsearchBulkRequestFailed(String request, String
...
+					+ " you may override this name by using @CharFilterDef.name." )
 	SearchException charFilterNamingConflict(String remoteName);
 
 	@Message(id = ES_BACKEND_MESSAGES_START_ID + 57,
-			value = "The same token filter name '%1$s' is used multiple times. The token filter names must be unique."
-					+ " If this name was automatically generated, you may override it by using @TokenFilterDef.name." )
+			value = "The same token filter name '%1$s' is assigned to multiple definitions. The token filter names must be unique."
+					+ " If your used the @TokenFilterDef annotation and this name was automatically generated,"
```
Sanne: same typo

engine/src/main/java/org/hibernate/search/analyzer/impl/LuceneEmbeddedAnalyzerStrategy.java
```diff
@@ -73,6 +76,33 @@ private Version getLuceneMatchVersion(SearchConfiguration cfg) {
...
 	}
 
+	private Map<String, AnalyzerDef> createDefaultAnalyzerDefinitions() {
+		LuceneAnalyzerDefinitionRegistryBuilderImpl builder = new LuceneAnalyzerDefinitionRegistryBuilderImpl();
+
+		String providerClassName = cfg.getProperty( Environment.ANALYZER_DEFINITION_PROVIDER );
+		if ( providerClassName != null ) {
+			LuceneAnalyzerDefinitionProvider provider;
+			try {
+				Class<?> providerClazz = ClassLoaderHelper.classForName( providerClassName, serviceManager );
```
Sanne: same here. Could you use `ClassLoaderHelper.instanceFromName` instead?

engine/src/main/java/org/hibernate/search/util/logging/impl/Log.java
```diff
@@ -1004,4 +1004,12 @@
 
 	@Message(id = 328, value = "Cannot create context for class: %1$s" )
 	SearchException cannotCreateBridgeDefinedField(@FormatWith(ClassFormatter.class) Class<?> backend, @Cause Exception e);
+
+	@Message(id = 329, value = "Property " + Environment.ANALYZER_DEFINITION_PROVIDER + "set to value '%1$s' is invalid."
```
Sanne: - you need a space just before `set to value`.
- maybe better to quote the Environment.ANALYZER_DEFINITION_PROVIDER with single quotes?
- I'd not request "must not throw any exception", seems redundant. Also a bit scary.. will the data center blow up if an OOM happens in there?
- optional alternative: may be more useful to remind that a public, no-argument constructor is expected.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/analyzer/impl/ElasticsearchAnalyzerStrategy.java
yrodiere: `ReflectionHelper.createInstance` below allows the use of factory classes with the `@Factory` annotation. It seems useful in this particular case, but it won't work well with `ClassLoaderHelper.instanceFromName`.
Did you mean I should drop the `@Factory` support entirely?

Sanne: interesting idea, I wasn't thinking of `@Factory`. The checks I was thinking of are the ones provided by 

- org.hibernate.search.util.impl.ClassLoaderHelper.verifySuperTypeCompatibility(Class<T>, Object, Class<?>, String)
- org.hibernate.search.util.impl.ClassLoaderHelper.checkHasNoArgConstructor(Class<?>, String)
- org.hibernate.search.util.impl.ClassLoaderHelper.checkClassType(Class<?>, String)

I guess you can leave the `@Factory`, it looks good to have it. We probably should unify the two approaches in a follow-up improvement.

engine/src/main/java/org/hibernate/search/util/logging/impl/Log.java
yrodiere: Fixed. I also fixed the similar message for Elasticsearch.
New version:

```
	@Message(id = 329, value = "Property '" + Environment.ANALYZER_DEFINITION_PROVIDER + "' set to value '%1$s' is invalid."
			+ " The value must be the fully-qualified name of a class with a public, no-arg constructor in your classpath."
			+ " Also, the class must either implement LuceneAnalyzerDefinitionProvider or expose a public,"
			+ " @Factory-annotated method returning a LuceneAnalyzerDefinitionProvider.")
```

elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/ElasticsearchAnalysisDefinitionProviderIT.java
```diff
@@ -0,0 +1,342 @@
...
+						.getAnalyzerReference( CUSTOM_ANALYZER_NAME ) )
+				.as( "Analyzer reference for '" + CUSTOM_ANALYZER_NAME + "' fetched from the integrator" )
+				.isNotNull();
+
+		AnalyzerDefAnnotationEntity entity = new AnalyzerDefAnnotationEntity();
+		entity.id = 0;
+		entity.field = "charFilterShouldReplace|foo";
+		index( integrator, entity );
+		// Expecting a keyword tokenizer (as per the @AnalyzerDef annotation)
+		assertMatchesExactly( integrator, entity, "field", "charFilterShouldReplace|foo" );
```
Sanne: It looks like - if I'm interpreting this right - that you expect the annotations to override the configuration.
Could you explain the rationale?

Sanne: BTW the comment on line 80 mentioned "fetched from the integrator" so that's not consistent.

Sanne: (same question applies to the override() test in LuceneAnalyzerDefinitionProviderTest)

yrodiere: About the comment: the code actually fetches the analyzer reference from the integrator, so technically the comment is correct. The integrator is not the analyzer provider. Maybe you have a better wording in mind?

The rationale behind override is you mentioned (initially) that we could have a "default" provider that would add some common analyzers. If we don't make `@AnalyzerDef` override definitions from the provider, on existing applications the introduction of those default definitions might change analyzer definitions without the users knowing about it. For instance if we introduce a definition for "standard" and the user already has an `@AnalyzerDef` named "standard".

The other solution would be to forbid analyzer overrides completely (throw an exception), but then again that would mean breaking compatibility with existing code if we have default analyzer definitions.
Also, I suspect we'll need to support overrides at some point, especially when we'll handle index-scoped analyzer providers.

Sanne:
great! merged.

I did two last changes:
 - marked the new DSL as experimental, so we can change it in the next minor with more flexibility.
 - changed the JIRA description as the text implied we'd introduce a `Service` (literally) but it's not one.

PR #1309, opened 2017-02-09
HSEARCH-2453 Add dedicated configuration options for Elasticsearch authentication

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2453

**To be backported to 5.6.1.**

Authentication was already possible, but a bit painful, since you had to
include the username and password in the host URLs. Which meant
repeating it over and over again when using multiple URLs, and also
meant you had to URL-encode the username and password yourself.

Tests are not included, because it's a pain to automate the setup of
authentication on the ES instance used for tests. Also, I'm quite sure
we can't do it with an embedded version of ES, and I don't even know if
the licensing terms would allow it (the Shield plugin is subject to a
commercial license).

To test manually, see:
https://www.elastic.co/blog/getting-started-with-elasticsearch-ssl-native-authentication

### DIFF
--- a/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
+++ b/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
@@ -124,6 +124,14 @@ You may also select multiple hosts (separated by whitespace characters), so that
 `hibernate.search.default.elasticsearch.host \http://es1.mycompany.com:9200 \http://es2.mycompany.com:9200`
 +
 In the example above, the first request will go to `es1`, the second to `es2`, the third to `es1`, and so on.
+Username for Elasticsearch connection:: `hibernate.search.default.elasticsearch.username ironman` (default is empty, meaning anonymous access)
+Password for Elasticsearch connection:: `hibernate.search.default.elasticsearch.password j@rV1s` (default is empty)
++
+[CAUTION]
+====
+If you used HTTP instead of HTTPS in any of the Elasticsearch host URLs (see above),
+your password will be transmitted in clear text over the network.
+====
 [[elasticsearch-schema-management-strategy]]Select the index creation strategy::
 `hibernate.search.default.elasticsearch.index_schema_management_strategy CREATE` (default)
 +
@@ -207,7 +215,7 @@ Properties prefixed with `hibernate.search.default` can be given globally as sho
 `hibernate.search.someindex.elasticsearch.index_schema_management_strategy MERGE`
 
 This excludes properties related to the internal Elasticsearch client, which at the moment is common to every index manager (but this will change in a future version).
-Excluded properties are `host`, `read_timeout`, `connection_timeout`, `max_total_connection`, `max_total_connection_per_route`, `discovery.enabled` and `discovery.refresh_interval`.
+Excluded properties are `host`, `username`, `password`, `read_timeout`, `connection_timeout`, `max_total_connection`, `max_total_connection_per_route`, `discovery.enabled` and `discovery.refresh_interval`.
 --
 
 === Mapping and indexing
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/ElasticsearchEnvironment.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/ElasticsearchEnvironment.java
@@ -49,9 +49,36 @@ public final class ElasticsearchEnvironment {
 	 * {@code hibernate.search.default.elasticsearch.host=http://myeshost.com:9200}), i.e. only a single Elasticsearch
 	 * cluster is supported for all the indexed entities. This limitation will be removed in a future version of
 	 * Hibernate Search.
-	 * <p>
 	 */
 	public static final String SERVER_URI = "elasticsearch.host";
+
+	/**
+	 * Property for specifying the username to send when connecting to the Elasticsearch servers.
+	 * <p>
+	 * A string is expected.
+	 * <p>
+	 * Defaults to no username (anonymous access).
+	 * <p>
+	 * Can only be given <b>globally</b> (e.g.
+	 * {@code hibernate.search.default.elasticsearch.username=ironman}), because. only a single Elasticsearch
+	 * cluster is supported for all the indexed entities. This limitation will be removed in a future version of
+	 * Hibernate Search.
+	 */
+	public static final String SERVER_USERNAME = "elasticsearch.username";
+
+	/**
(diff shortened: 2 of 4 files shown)


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/ElasticsearchEnvironment.java
```diff
@@ -49,11 +49,38 @@
 	 * {@code hibernate.search.default.elasticsearch.host=http://myeshost.com:9200}), i.e. only a single Elasticsearch
 	 * cluster is supported for all the indexed entities. This limitation will be removed in a future version of
 	 * Hibernate Search.
-	 * <p>
 	 */
 	public static final String SERVER_URI = "elasticsearch.host";
 
 	/**
+	 * Property for specifying the host names and HTTP ports of the Elasticsearch servers to connect to.
```
Sanne: hostname and port? Looka like a copy/paste error

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/ElasticsearchEnvironment.java
```diff
@@ -49,11 +49,38 @@
...
+	 * <p>
+	 * A string is expected.
+	 * <p>
+	 * Defaults to no username (anonymous access).
+	 * <p>
+	 * Can only be given <b>globally</b> (e.g.
+	 * {@code hibernate.search.default.elasticsearch.username=ironman}), because. only a single Elasticsearch
```
Sanne: I'm confused about the "e.g." usage here. You say it has to be global, but then you make an example which uses an index-specific property.

(I understand the reasons, just pointing out the explanation is oddly phrased)

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/ElasticsearchEnvironment.java
```diff
@@ -49,11 +49,38 @@
...
+	public static final String SERVER_USERNAME = "elasticsearch.username";
+
+	/**
+	 * Property for specifying the password to use when connecting to the Elasticsearch servers.
+	 * <p>
+	 * A string is expected.
+	 * <p>
+	 * Defaults to no password at all.
+	 * <p>
+	 * Can only be given <b>globally</b> (e.g.
```
Sanne: same issue with "e.g." here

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/client/impl/JestClient.java
```diff
@@ -119,9 +119,23 @@ public void start(Properties properties, BuildContext context) {
...
-				.build()
+				.gson( gsonService.getGson() );
+
+		String username = ConfigurationParseHelper.getString(
+				properties,
+				CLIENT_PROP_PREFIX + ElasticsearchEnvironment.SERVER_USERNAME,
+				null
 		);
+		if ( !StringHelper.isEmpty( username ) ) {
+			String password = ConfigurationParseHelper.getString(
```
Sanne: what if the username is non-empty, but the password is. Would that be valid or should we warn / ignore?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/client/impl/JestClient.java
```diff
@@ -119,9 +119,23 @@ public void start(Properties properties, BuildContext context) {
...
-				.gson( gsonService.getGson() )
-				.build()
+				.gson( gsonService.getGson() );
+
+		String username = ConfigurationParseHelper.getString(
+				properties,
+				CLIENT_PROP_PREFIX + ElasticsearchEnvironment.SERVER_USERNAME,
+				null
 		);
+		if ( !StringHelper.isEmpty( username ) ) {
```
Sanne: so a zero-length string would disable it? Should we rather check for not-null, or is that going to be all the same for ES?

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -124,6 +124,14 @@ You may also select multiple hosts (separated by whitespace characters), so that
 `hibernate.search.default.elasticsearch.host \http://es1.mycompany.com:9200 \http://es2.mycompany.com:9200`
 +
 In the example above, the first request will go to `es1`, the second to `es2`, the third to `es1`, and so on.
+Username for Elasticsearch connection:: `hibernate.search.default.elasticsearch.username ironman` (default is empty, meaning anonymous access)
+Password for Elasticsearch connection:: `hibernate.search.default.elasticsearch.password j@rV1s` (default is empty)
++
+[CAUTION]
+====
+If you used HTTP instead of HTTPS in any of the Elasticsearch host URLs (see above),
+your password will be transmitted in clear text over the network.
```
Sanne: Would be nice to log a WARN at boostrap about this?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/client/impl/JestClient.java
Sanne: +1

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/client/impl/JestClient.java
Sanne: It was meant as a question, not a demand :)

yrodiere: Ok, sorry. There is no particular check in Jest or the Apache HTTP Client, so I guess an empty string is propagated to ES as is. From my tests, ES rejects empty usernames:

```
$ curl -XGET ':password@localhost:9200/_cat?pretty'
{
  "error" : {
    "root_cause" : [ {
      "type" : "security_exception",
      "reason" : "missing authentication token for REST request [/_cat?pretty]",
      "header" : {
        "WWW-Authenticate" : "Basic realm=\"shield\" charset=\"UTF-8\""
      }
    } ],
    "type" : "security_exception",
    "reason" : "missing authentication token for REST request [/_cat?pretty]",
    "header" : {
      "WWW-Authenticate" : "Basic realm=\"shield\" charset=\"UTF-8\""
    }
  },
  "status" : 401
}
```

Sanne:
thanks! merged it

Sanne:
and backported

PR #1310, opened 2017-02-09
HSEARCH-2580 Update of the "previous.stable" property following the 5.6.0.Final release

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2580

**To be backported to 5.6**. Backport for 5.6 available there: https://github.com/yrodiere/hibernate-search/tree/HSEARCH-2580-5.6

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -235,7 +235,7 @@
         <puppycrawl.checkstyle.version>7.4</puppycrawl.checkstyle.version>
 
         <!-- Version to be used as baseline for API/SPI change reports -->
-        <previous.stable>5.5.2.Final</previous.stable>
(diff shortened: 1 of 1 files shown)


yrodiere:
Actually, I'm not sure whether we should backport this to 5.6 or not... Against what version do we want to check version x.y.z? The latest `x.(y-1).*.Final`? Or simply the previous `*.*.*.Final`?
In https://developer.jboss.org/wiki/ReleasingHibernateSearch I stated "the latest `x.(y-1).*.Final`", which would mean we don't want to backport this to 5.6.

Sanne:
I'll not backport it. Those ships have sailed..

Sanne:
merged, thanks!

PR #1311, opened 2017-02-10
HSEARCH-2559 Use a local copy of Font Awesome

gsmet:
This is a request of Gunnar so that the doc is in good shape even if you don't have an Internet connection.


gsmet:
Jenkins, retest this please

Sanne:
thanks, merged!

PR #1312, opened 2017-02-13
HSEARCH-2453 Add unit tests for JestClient (authentication in particular)

yrodiere:
Follow-up on #1309 : https://hibernate.atlassian.net/browse/HSEARCH-2453

Here are the test you asked for, @Sanne ;)
We can also backport this if you think it's worth it.

### DIFF
--- a/elasticsearch/pom.xml
+++ b/elasticsearch/pom.xml
@@ -94,6 +94,22 @@
             <artifactId>unitils-easymock</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.github.tomakehurst</groupId>
+            <artifactId>wiremock</artifactId>
+            <scope>test</scope>
+            <exclusions>
+                <!-- Exclude dependencies that conflict with io.searchbox:jest -->
+                <exclusion>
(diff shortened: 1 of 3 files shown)


yrodiere:
I just force-pushed an update to solve a dependency convergence issue. Hope this will fix the build...

Sanne:
looks great! Thanks for researching this, feels much better.

Testing now..

Sanne:
there's some problem with the dependencies, I guess caused by the exclusions you had to add.

```
Tests in error: 
  JestClientTest.authentication:163 » NoClassDefFound com/fasterxml/jackson/data...
  JestClientTest.multipleHosts:122 » NoClassDefFound com/fasterxml/jackson/datab...
  JestClientTest.simple:69 » NoClassDefFound com/fasterxml/jackson/databind/Json...

```

yrodiere:
@Sanne Solved, sorry about that.
So in the end, there were two issues:

 * a dependency convergence between wiremock and jest
 * a dependency convergence within jackson itself (jackson-databind 2.6.1 depends on jackson-annotations 2.6.0, go figure...), which wiremock depends on

Sanne:
merged! thanks. I'll backport as well.

PR #1313, opened 2017-02-13
HSEARCH-2564 Generic type signature of FullTextQuery should not be as flexible as the JPA Query

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2564

I chose to make `FullTextQuery` extend `Query` (the raw type) instead of `Query<T>`. This solves our immediate problems and doesn't involve breaking any API.
Given that we can't do much with the generic type parameter without breaking APIs anyway, I think it's our safest option.

### DIFF
--- a/orm/src/main/java/org/hibernate/search/query/hibernate/impl/FullTextQueryImpl.java
+++ b/orm/src/main/java/org/hibernate/search/query/hibernate/impl/FullTextQueryImpl.java
@@ -50,7 +50,8 @@ import org.hibernate.type.Type;
  * @author Emmanuel Bernard
  * @author Hardy Ferentschik
  */
-public class FullTextQueryImpl<R> extends AbstractProducedQuery<R> implements FullTextQuery<R> {
+@SuppressWarnings("rawtypes") // We extend the raw version of AbstractProducedQuery on purpose, see HSEARCH-2564
+public class FullTextQueryImpl extends AbstractProducedQuery implements FullTextQuery {
 
 	private static final Log log = LoggerFactory.make();
 
@@ -100,7 +101,7 @@ public class FullTextQueryImpl<R> extends AbstractProducedQuery<R> implements Fu
 	}
 
 	@Override
-	public List<R> getResultList() {
+	public List getResultList() {
 		return list();
 	}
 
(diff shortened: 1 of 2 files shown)


yrodiere:
Jenkins, retest this please

orm/src/main/java/org/hibernate/search/query/hibernate/impl/FullTextQueryImpl.java
```diff
@@ -289,11 +290,20 @@ public FullTextQuery setResultTransformer(ResultTransformer transformer) {
...
 
+	/*
+	 * Implementation note: this method is defined as generic in the interface,
+	 * but we must implement it without generics (otherwise it won't compile).
+	 *
+	 * The actual reason is a bit hard to explain: basically we implement
+	 * javax.persistence.Query as a raw type at some point, and our superclass
+	 * (also extended as a raw type) also implements this interface, but as a non-raw type.
+	 * This seems to confuse the compiler, which thinks there are two different methods.
+	 */
```
Sanne: I've been trying to understand this but couldn't figure out why this is necessary :-/

@emmanuelbernard @hferentschik @gunnarmorling ? You all are the generics gurus..

In case we have to integrate this, will it cause runtime compatbility issues with our previous signature?
(I think not, as our API on the interface is still the generic one... but that just makes it even more odd)

yrodiere: @Sanne From what I understand of type erasure, my code should be strictly equivalent to the previous code.
I can have a try if you want, compile a small application against 5.6 and then try to run it against 5.7?

Sanne: thanks @yrodiere, no need. I'll soon test this and other integrations with a previous build of Infinispan and I'm quite optimistic.

I'll merge this for now so we can move on, but if someone knows I'd be interested.. looks like a Java compiler bug, but then again it would be quite a coincidence that both javac and the Eclipse compiler have the same quirk.

Sanne:
merged

PR #1314, opened 2017-02-16
HSEARCH-2587 Sort DSL on embedded fields throws NPE

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2587

### DIFF
--- a/engine/src/main/java/org/hibernate/search/query/dsl/sort/impl/SortFieldStates.java
+++ b/engine/src/main/java/org/hibernate/search/query/dsl/sort/impl/SortFieldStates.java
@@ -14,12 +14,13 @@ import java.util.Map;
 import org.apache.lucene.search.Sort;
 import org.apache.lucene.search.SortField;
 import org.apache.lucene.search.SortField.Type;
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/hibernate/search/query/dsl/sort/impl/SortFieldStates.java
```diff
@@ -150,30 +151,32 @@ public void determineCurrentSortFieldTypeAutomaticaly() {
...
-					case INTEGER:
-						return SortField.Type.INT;
-					case UNKNOWN:
-						break; // Fail below
-				}
-				break;
-			case BASIC:
-				return SortField.Type.STRING;
+		TypeMetadata typeMetadata = queryContext.getFactory()
+				.unwrap( SearchIntegrator.class )
```
Sanne: there's no need to unwrap ?
I guess the `getFactory` method should be renamed.

Sanne:
merged, thanks! I rename that method and removed the unwrapping.

PR #1315, opened 2017-02-16
HSEARCH-2588 Make byField(String) in the sort DSL automatically determine the sort type for metadata providing field bridges	

yrodiere:
This PR is based on #1314 , which will have to be merged first.

Related JIRA ticket: https://hibernate.atlassian.net/browse/HSEARCH-2588

Note: I added a commit to deprecated `byField(String, SortField.Type)` and `andByField(String, SortField.Type)`, so we incite users to implement `MetadataProvidingFieldBridge` as much as possible. We can merge the PR without this last commit if you think it's too harsh, though.


yrodiere:
Jenkins, retest this please

Sanne:
thanks, merged it. Including the deprecation, I agree with that.

PR #1317, opened 2017-02-16
HSEARCH-2590 Fix an over-allocation of memory when paging on Elasticsearch

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2590


yrodiere:
Note: this should be backported to 5.6.

gsmet:
@yrodiere looks good, just push it.

Sanne:
merged, great find!

PR #1318, opened 2017-02-17
HSEARCH-2678 Add a simple query string entry point to the DSL

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-2678

So, this has a very long history as I started to work on this in 2011 :). This patch is very different from what I envisioned at the start.

Why revisit it now? Got kicked out last week of $oldJob GitHub organization and this branch was hosted in the hibernate-search repository of this organization so it was high time to host it in a safe place. While at it, I decided to rebase it, polish it and well... here we are...

## Some history

The idea of this patch came from the impossibility to execute a plain text query with a AND operator with the `keyword()` DSL.

Basically, we had our users entering "word1 word2" as a search query and they wanted only the results containing word1 and word2. We also wanted this to work with multiple fields. Meaning I would want to look for results containing word1 and word2 but either in name or description.

It looks like a simple issue but it's not that simple, especially when analyzers come into play.

Our users might also want to use simple "advanced" query features like phrase queries or negative queries.

I tried several times to arrive to something working well, thought about using DisMax...

The current version of the patch became possible when Lucene introduced the `SimpleQueryParser` (which is also available on Elasticsearch) which is a very simple lenient parser that can be used exactly for this purpose.

## How it works

You can see a couple of examples in the doc added at the start of this patch. It should give you a good overview of how it works.

I also added a couple of tests that demonstrates the features.

While quite simple by itself, I think it's a great feature for the end user.


gsmet:
Jenkins, retest this please

documentation/src/main/asciidoc/query.asciidoc
```diff
@@ -289,6 +289,74 @@ Query luceneQuery = mythQB.keyword()
...
+If you prefer using the AND operator as default, you can add the `defaultOperatorIsAnd` flag.
+The following query returns results containing storm AND tree.
+
+====
+[source, JAVA]
+----
+Query luceneQuery = mythQB
+    .text()
+    .onField("history")
+    .defaultOperatorIsAnd()
```
yrodiere: This is a bit clumsy.

So...:

 1. `defaultOperator(Operator.AND)` isn't much better
 2. `or()`/`and()` lack some context
 3. I guess `.conjunctive`/`.disjunctive` are a tad too obscure

If nobody finds a better option, I think I'd say we go for 1... Even though it's far from ideal.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ToElasticsearch.java
```diff
@@ -418,6 +424,32 @@ private static JsonObject convertRemoteMatchQuery(RemoteMatchQuery query) {
...
+			StringBuilder sb = new StringBuilder( field.getName() );
+			if ( field.getBoost() != DEFAULT_BOOST ) {
+				sb.append( BOOST_OPERATOR ).append( field.getBoost() );
+			}
+			fieldArray.add( sb.toString() );
+		}
+
+		JsonBuilder.Object queryBuilder = JsonBuilder.object()
+				.addProperty( "query", query.getQuery() )
+				.addProperty( "defaultOperator", query.isMatchAll() ? "and" : "or" );
```
yrodiere: It's `default_operator`, not `defaultOperator`. It may work in ES2, but I've seen cases where ES5 dropped such undocumented syntax, so we should stay away from such syntax.

engine/src/main/java/org/hibernate/search/query/dsl/PlainTextQueryDefinitionTermination.java
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+package org.hibernate.search.query.dsl;
+
+
+public interface PlainTextQueryDefinitionTermination {
+
+	/**
+	 * Value searched in the fields.
+	 * The value is passed to the respective fields analyzers.
```
yrodiere: This javadoc should be rephrased: the value is parsed, it's not analyzed by the field analyzers. Only part of it is actually analyzed.

engine/src/main/java/org/hibernate/search/query/dsl/PlainTextContext.java
```diff
@@ -0,0 +1,30 @@
...
+	/**
+	 * @param field The field name the plain text query is executed on
+	 *
+	 * @return {@code PlainTextMatchingContext} to continue the term query
+	 */
+	PlainTextMatchingContext onField(String field);
+
+	/**
+	 * @param fields The field names the term query is executed on. The underlying properties for the specified
+	 * fields need to be of the same type. For example, it is not possible to use this method with a mixture of
```
yrodiere: This javadoc seems wrong; we can target any field, regardless of its type. We may want to warn users that queries apply on the bridged data, though (so dates may have a weird format, for instance).

engine/src/main/java/org/hibernate/search/query/dsl/impl/ConnectedMultiFieldsPlainTextQueryBuilder.java
```diff
@@ -0,0 +1,85 @@
...
+			throw LOG.simpleQueryParserDoesNotSupportEmptyQueries();
+		}
+
+		Query query;
+
+		if ( queryContext.getQueryAnalyzerReference().is( RemoteAnalyzerReference.class ) ) {
+			// we use the analyzers defined in the schema as it is not possible to define per field schema here
+			RemoteSimpleQueryStringQuery.Builder builder = new RemoteSimpleQueryStringQuery.Builder()
+					.query( searchText )
+					.defaultOperatorIsAnd( defaultOperatorIsAnd );
```
yrodiere: We don't pass the reference to the analyzer here... If I understand correctly, this is because we can't define per-field analyzers in Elasticsearch's `simple_query_string`. We only have one global `analyzer` parameter.

Maybe this should be documented? Users may be surprised when they write:

```
searchFactory.buildQueryBuilder().forEntity( MyEntity.class ).overridesForField( "foo", "someQueryAnalyzer" ).get()
    .text().forField( "foo" ).matching( "blah blah" ).createQuery();
```

... and the resulting query doesn't use the overridden analyzer.

Maybe we should also support the case where there's only one field, and throw an error if there are multiple fields with different analyzers?

documentation/src/main/asciidoc/query.asciidoc
```diff
@@ -289,6 +289,74 @@ Query luceneQuery = mythQB.keyword()
 In the previous example, only field name is boosted to 5.
 
 
+===== Plain text queries
+
+Plain text queries use the Lucene SimpleQueryParser. The query text is parsed and it exposes more
```
yrodiere: Maybe we should add something to the Elasticsearch integration doc, to mention that the syntax there is the one for `simple_query_string` (which may be slightly different, at least in the future).

gsmet:
@yrodiere I think I addressed all your comments and we should be on the same page regarding this feature. The first 2 commits should probably be squashed after review.

As for the build status, the failure looks unrelated.

yrodiere:
Jenkins, retest this please

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ToElasticsearch.java
```diff
@@ -421,6 +432,58 @@ private static JsonObject convertRemoteMatchQuery(RemoteMatchQuery query) {
...
+		queryBuilder.add( "fields", fieldArray );
+
+		if ( overridingRemoteAnalyzerName != null ) {
+			if ( analyzers.size() == 1 ) {
+				queryBuilder.addProperty( "analyzer", overridingRemoteAnalyzerName );
+			}
+			else {
+				List<String> overridingAnalyzers = new ArrayList<>();
+				overridingAnalyzers.addAll( analyzers );
+				overridingAnalyzers.add( overridingRemoteAnalyzerName );
```
yrodiere: From what I can see above, `overridingRemoteAnalyzerName` is already in the `analyzers` set, so we could just pass the set to the logger method (if you made the parameter a `Collection`, that is :grin: )

```java
			analyzers.add( queryRemoteAnalyzerName ); // Value added to the set here
 			if ( !queryRemoteAnalyzerName.equals( originalRemoteAnalyzerName ) ) {
 				if ( overridingRemoteAnalyzerName == null ) {
 					overridingRemoteAnalyzerName = queryRemoteAnalyzerName; // Same value put into overridingRemoteAnalyzerName here
 				}
```

engine/src/main/java/org/hibernate/search/query/dsl/QueryBuilder.java
```diff
@@ -44,6 +44,13 @@
 	PhraseContext phrase();
 
 	/**
+	 * Build a plain text query.
```
yrodiere: Should probably be "Build a query from a simple query string"... ? Same for the `@return` below.

engine/src/main/java/org/hibernate/search/query/dsl/SimpleQueryStringContext.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package org.hibernate.search.query.dsl;
+
+/**
+ * @author Guillaume Smet
+ */
+public interface SimpleQueryStringContext extends QueryCustomization<SimpleQueryStringContext> {
+	/**
+	 * @param field The field name the plain text query is executed on
```
yrodiere: Same here, should be about a "simple query string" rather than a "plain text query", since we decided to use that terminology.

engine/src/main/java/org/hibernate/search/query/dsl/SimpleQueryStringContext.java
```diff
@@ -0,0 +1,28 @@
...
+	 *
+	 * @return {@code SimpleQueryStringMatchingContext} to continue the term query
```
yrodiere: Copy/paste error here, it's not a term query

engine/src/main/java/org/hibernate/search/query/dsl/SimpleQueryStringContext.java
```diff
@@ -0,0 +1,28 @@
...
+	 */
+	SimpleQueryStringMatchingContext onField(String field);
+
+	/**
+	 * @param field The first field added to the list of fields (follows the same rules described below for fields)
+	 * @param fields The field names the term query is executed on
```
yrodiere: Copy/paste error here, it's not a term query

engine/src/main/java/org/hibernate/search/query/dsl/SimpleQueryStringMatchingContext.java
```diff
@@ -0,0 +1,39 @@
...
+
+	/**
+	 * Fields the query is executed on.
+	 */
+	SimpleQueryStringMatchingContext andFields(String... field);
+
+	/**
+	 * Define the default operator as AND.
+	 */
+	SimpleQueryStringDefinitionTermination useAndAsDefaultOperator();
```
yrodiere: `withAndAsDefaultOperator`, maybe? This would make the DSL feel a bit more natural: "simple query string, on field 'foo', with 'and' as default operator, matching 'bar'"
It's just a suggestion, I don't mean to argue about this. Ignore it if you don't like it.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ToElasticsearch.java
gsmet: Well spotted, fixed.

engine/src/main/java/org/hibernate/search/query/dsl/QueryBuilder.java
gsmet: Yup, fixed.

engine/src/main/java/org/hibernate/search/query/dsl/SimpleQueryStringContext.java
gsmet: Fixed.

engine/src/main/java/org/hibernate/search/query/dsl/SimpleQueryStringContext.java
gsmet: Fixed, sorry about that.

engine/src/main/java/org/hibernate/search/query/dsl/SimpleQueryStringContext.java
gsmet: Fixed, sorry about that.

engine/src/main/java/org/hibernate/search/query/dsl/SimpleQueryStringMatchingContext.java
gsmet: Done.

gsmet:
@yrodiere I pushed a new version:
* squashed the 2 first commits
* added one commit to rename useAnd...
* added one commit to add lucene-queryparser to OSGi and distribution. It was already present in the WildFly modules.

Sanne:
Great work!

Looks very good. The only detail I'm not convinced of is the starting point of the DSL: 

`qb.simpleQueryString()`

It feels a bit at odds with the other methods, but not having a better idea I merged it.

Proposal: let's post a dedicated blog to present / explain the feature and see if someone has a better name to offer.

Merged!

PR #1320, opened 2017-02-20
HSEARCH-2578 Include the lucene-grouping artifact in the WildFly modules

Sanne:
This is to keep things easier until we don't actually move to use the separately packaged Lucene modules.

I'm including no tests as it's a temporary measure.

- https://hibernate.atlassian.net/browse/HSEARCH-2578


yrodiere:
Thanks; merged.

PR #1322, opened 2017-02-21
HSEARCH-2597 Do not rely on automatic detection of JPA annotations fr…

Sanne:
…om the classpath

Resolve some inconsistencies around `@Id` handling, and polish a couple of tricky tests.

See JIRA for the full rationale.

https://hibernate.atlassian.net/browse/HSEARCH-2597

### DIFF
--- a/engine/src/test/java/org/hibernate/search/testsupport/junit/SearchFactoryHolder.java
+++ b/engine/src/test/java/org/hibernate/search/testsupport/junit/SearchFactoryHolder.java
@@ -48,6 +48,7 @@ public class SearchFactoryHolder extends ExternalResource {
 	private int numberOfSessionFactories = 1;
 	private boolean idProvidedImplicit = false;
 	private boolean multitenancyEnabled = false;
+	private boolean enableJPAAnnotationsProcessing = false;
 	private DefaultBatchBackend batchBackend;
 
 	public SearchFactoryHolder(Class<?>... entities) {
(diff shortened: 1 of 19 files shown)


engine/src/test/java/org/hibernate/search/testsupport/junit/SearchFactoryHolder.java
```diff
@@ -152,4 +154,14 @@ public BatchBackend getBatchBackend() {
 		return batchBackend;
 	}
 
+	/**
+	 * The default for the tests in hibernate-search-engine is to not rely on JPA
+	 * specific annotations. These are typically used though when integrating
+	 * with Hibernate, so some tests might want to explicitly enable this.
+	 */
+	public SearchFactoryHolder enableJPAAnnotationsProcessing(boolean enableJPAAnnotationsProcessing) {
```
yrodiere: Not related to this PR, but it seems we're introducing two different naming schemes in this class for option setters:

 * `withMultitenancyEnabled`, `withIdProvidedImplicit`, ...
 * `multipleInstances`, `enableJPAAnnotationsProcessing`, ...

No need to change this PR, but just so I know, do you have a preference?

engine/src/test/java/org/hibernate/search/testsupport/junit/SearchFactoryHolder.java
```diff
@@ -86,6 +87,7 @@ private SearchIntegrator createSearchFactory() {
 		}
 		cfg.setIdProvidedImplicit( idProvidedImplicit );
 		cfg.setMultitenancyEnabled( multitenancyEnabled );
+		cfg.setEnableJPAAnnotationsPRocessing( enableJPAAnnotationsProcessing );
```
Sanne: thanks

engine/src/test/java/org/hibernate/search/testsupport/junit/SearchFactoryHolder.java
Sanne: I find the new name more suited for this specific case, as it has an imperative effect on a "toggle".

You're right it's inconsistent though. I think the only (weak) justification I can think of is that the multi-tenancy and provided-id-strategies need to be plugged in additional strategies to work.

Sanne:
fixed the camel-casing and merged

PR #1323, opened 2017-02-21
HSEARCH-2593 Node discovery uses the HTTP (not HTTPS) scheme by default and isn't configurable

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2593

I decided to keep the "http" scheme as default, since it's in line with our policy to make it easy to test Hibernate Search, but I must admit I'm not really happy with that (it should be secure by default...).

I tried to add a unit test, but I only got so far as testing the discovery with a HTTP scheme, because we'd need some support for self-signed certificates to be able to test easily, and we don't have any (see [HSEARCH-2599](https://hibernate.atlassian.net/browse/HSEARCH-2599)). I left the test case anyway, since it's useful, but it definitely doesn't test HSEARCH-2593.

### DIFF
--- a/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
+++ b/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
@@ -186,6 +186,9 @@ and will remove them from the server list if they don't.
 Time interval between two executions of the automatic discovery (in seconds):: `hibernate.search.default.elasticsearch.discovery.refresh_interval 10` (default)
 +
 This setting will only be taken into account if automatic discovery is enabled (see above).
+Scheme to use when connecting to automatically discovered nodes (`http` or `https`):: `hibernate.search.default.elasticsearch.discovery.scheme http` (default)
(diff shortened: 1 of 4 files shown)


Sanne:
There was an inconsistency between the property name being documented vs. the one being used:
- hibernate.search.default.elasticsearch.discovery.scheme
- hibernate.search.default.elasticsearch.discovery.default_scheme

I picked the second one. Hope it's good?

Sanne:
merged

yrodiere:
> There was an inconsistency between the property name being documented vs. the one being used:

Ha, I had trouble making my mind about this. I guess my subconscious didn't agree with me :) Thanks for spotting this!

> I picked the second one. Hope it's good?

It's currently a bit less appropriate, since currently there's no way to define a per-node scheme, but it's also more future-proof. So let's say it's the best one :)

Sanne:
Good to hear! Because it's released ;)

PR #1324, opened 2017-02-21
HSEARCH-2585 Allow injecting Analyzer definitions using a Service

Sanne:
https://hibernate.atlassian.net/browse/HSEARCH-2585

### DIFF
--- a/engine/src/main/java/org/hibernate/search/analyzer/definition/impl/PropertiesBasedLuceneAnalyzerDefinitionSourceService.java
+++ b/engine/src/main/java/org/hibernate/search/analyzer/definition/impl/PropertiesBasedLuceneAnalyzerDefinitionSourceService.java
@@ -1,0 +1,57 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.analyzer.definition.impl;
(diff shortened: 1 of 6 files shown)


engine/src/main/java/org/hibernate/search/analyzer/definition/impl/PropertiesbasedLuceneAnalyzerDefinitionSourceService.java
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * This service allows to set the {@link Environment#ANALYZER_DEFINITION_PROVIDER configuration properties} to
+ * point to a custom implementation of {@link org.hibernate.search.analyzer.definition.spi.LuceneAnalyzerDefinitionProvider}.
+ *
+ * Integrators which prefer to inject an alternative service by reference rather than setting a configuration
+ * property can provide an alternative Service implementation by overriding {@link org.hibernate.search.cfg.spi.SearchConfiguration.getProvidedServices()}.
+ *
+ * @author Sanne Grinovero
+ */
+public class PropertiesbasedLuceneAnalyzerDefinitionSourceService implements LuceneAnalyzerDefinitionSourceService, Startable {
```
yrodiere: Should be `PropertiesBased` instead of `Propertiesbased`?

yrodiere: Don't forget to also rename in the `META-INF/services` file :)

yrodiere:
Merged, thanks!

PR #1325, opened 2017-02-21
HSEARCH-2600 Update javadoc configuration

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-2600

@yrodiere @Sanne waiting for your feedback on whether we should disable or not the individual javadocs.

The first commit should be backported to 5.6 and IIRC 5.6 depends on 5.1 so we should probably update the ORM javadoc link here too.

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -101,6 +101,7 @@
                         ${basedir}/../serialization/avro/src/main/java;
                         ${basedir}/../backends/jgroups/src/main/java;
                         ${basedir}/../backends/jms/src/main/java;
+                        ${basedir}/../elasticsearch/src/main/java;
                     </sourcepath>
                     <docfilessubdirs>true</docfilessubdirs>
                     <stylesheetfile>${basedir}/src/main/javadoc/stylesheet.css</stylesheetfile>
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -106,8 +106,8 @@
                     <docfilessubdirs>true</docfilessubdirs>
                     <stylesheetfile>${basedir}/src/main/javadoc/stylesheet.css</stylesheetfile>
                     <links>
-                        <link>http://download.oracle.com/javase/7/docs/api/</link>
-                        <link>http://docs.jboss.org/hibernate/orm/5.0/javadocs/</link>
+                        <link>http://download.oracle.com/javase/8/docs/api/</link>
+                        <link>http://docs.jboss.org/hibernate/orm/5.2/javadocs/</link>
                         <link>http://lucene.apache.org/core/5_4_0/core/</link>
                         <link>http://lucene.apache.org/core/5_4_0/analyzers-common/</link>
                         <link>http://lucene.apache.org/core/5_4_0/queryparser/</link>
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -108,9 +108,9 @@
                     <links>
                         <link>http://download.oracle.com/javase/8/docs/api/</link>
                         <link>http://docs.jboss.org/hibernate/orm/5.2/javadocs/</link>
-                        <link>http://lucene.apache.org/core/5_4_0/core/</link>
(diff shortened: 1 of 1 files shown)


distribution/pom.xml
```diff
@@ -101,12 +101,13 @@
...
                         ${basedir}/../backends/jgroups/src/main/java;
                         ${basedir}/../backends/jms/src/main/java;
+                        ${basedir}/../elasticsearch/src/main/java;
                     </sourcepath>
                     <docfilessubdirs>true</docfilessubdirs>
                     <stylesheetfile>${basedir}/src/main/javadoc/stylesheet.css</stylesheetfile>
                     <links>
-                        <link>http://download.oracle.com/javase/7/docs/api/</link>
-                        <link>http://docs.jboss.org/hibernate/orm/5.0/javadocs/</link>
+                        <link>http://download.oracle.com/javase/8/docs/api/</link>
```
yrodiere: Well spotted. Maybe we should also update the Lucene doc link to `5_4_1`? It seems there isn't any javadoc for later `5.x` versions, though...

gsmet: Yeah, I tried 5.4.4 and it wasn't there so I left it alone. To be honest, I find it less disturbing to have 5.4.0 rather than 5.4.1 if we don't have 5.4.4. But I dont really care :).

gsmet:
@yrodiere we also use the dist profile in the regular master job.

The dist profile only generates aggregated javadoc so you won't have any issue if this particular javadoc is not included. To be completely thorough, when the aggregated javadoc is built, it builds the javadoc of all the modules but it only reports errors for the modules included in the final output. In the log file of the CI job, you can see there was an error generating the elasticsearch javadoc but as it wasn't included in the aggregated output, it wasn't reported.

In the release job, we use `-DperformRelease=true` to generate the sources jars and the javadoc jars (in HV, they are disabled with an explicit `maven.javadoc.skip` `true` directive).

It should be tested but I'm not sure generating the individual javadocs + the aggregated one will take much more time as it looks like the aggregate task looks for already generated javadoc before generating them. If it's confirmed, we would only have to add `-DperformRelease=true` to the CI job to be on the safe side.

distribution/pom.xml
Sanne: http://lucene.apache.org/core/5_5_4/core/

gsmet: @Sanne nice detective job :).

yrodiere: > http://lucene.apache.org/core/5_5_4/core/

Right, it's 5.*5*.4, not 5.*4*.4...

Sanne:
Thanks!

+1 to keep building javadocs. Just correctly ;)

I'm interested in keeping the builds fast locally - during dev - as that's the moment we all might need quick experiments.
I see no reason to disable such validations on CI jobs? We're talking about a couple of seconds, right?

I'd prefer to document a switch for local builds (default builds) so that I remember how I can *opt in* to check for verifications.

distribution/pom.xml
gsmet: Yeah, they were really *way* outdated :) I'll push another commit.

gsmet:
@Sanne it's more like one minute more for the entire build. No biggie IMHO. I'll update the CI jobs. BTW, you don't have a CI job for branch 5.6, I'll create one too.

Sanne:
@gsmet nice. While on it, create one for 5.7 as well? Master is now 5.8 ..

gsmet:
@Sanne right.

So, I created CI jobs for 5.6 and 5.7.

 * master, 5.6 and 5.7 runs with -DperformRelease=true (to be sure the javadoc is OK) and -Pdocumentation-pdf (to be sure we can generate the PDF doc - we might have issues with special characters or things like that)
 * PR job runs with -DperformRelease=true

JIRA also needs some love on the fixVersion front but I'll let you do that as you like it :).

This PR should be:
 * included in master
 * backported to 5.7 as is
 * adapted for 5.6 but it shouldn't be too hard ;)

gsmet:
Pushed to 5.6, 5.7 and master.

PR #1326, opened 2017-02-22
HSEARCH-2602 Recommend using MetadataProvidingFieldBridge instead of byField(String, SortField.Type) in the Sort DSL

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2602

### DIFF
--- a/documentation/src/main/asciidoc/query.asciidoc
+++ b/documentation/src/main/asciidoc/query.asciidoc
@@ -798,6 +798,13 @@ List results = query.list();
 ----
 ====
 
+[WARNING]
+====
+If you use the Sort DSL (like in the example above) to target fields indexed through *custom* ``FieldBridge``s,
+then those field bridges must implement <<metadata-providing-field-bridge,`MetadataProvidingFieldBridge`>>,
+so as to define the type of those fields and to mention that they are sortable.
+====
+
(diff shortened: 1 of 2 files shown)


documentation/src/main/asciidoc/query.asciidoc
```diff
@@ -819,27 +819,10 @@ Be aware that fields used for sorting must not be tokenized (see <<field-annotat
...
-  .sort()
-    .byField("authorId", SortField.Type.LONG)//Defining Sort type
-    .andByField("title")
-  .createSort();
-query.setSort(sort);
-List results = query.list();
------
-=====
+If you target fields indexed through *custom* ``FieldBridge``s,
+those field bridges must implement <<metadata-providing-field-bridge,`MetadataProvidingFieldBridge`>>.
```
Sanne: they don't have to ? Can you qualify? I guess it's missing something like "to be sorteable"

yrodiere: I was trying to say that if, through the Sort DSL, you target fields that are populated by custom field bridges, then those field bridges must implement `MetadataProvidingFieldBridge`.

But I see now that my warning isn't right after the Sort DSL example, like it should be. I pushed an update and rephrased the warning a bit. Is it better?

PR #1328, opened 2017-02-27
HSEARCH-2581 Add the ability to support multiple incompatible Elasticsearch versions (dialects)

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2581

This is the first step toward ES5 integration: add more abstraction so that we can implement different behaviors depending on the version of Elasticsearch.

In the process I had to replace Jest with the official [Elasticsearch REST client](https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/index.html). The reasons are:

 1. Jest doesn't support ES5, at least not officially
 2. We don't need all of Jest, only a small part
 3. Using Jest to target multiple ES versions would be hard, because we'd need to mitigate classpath conflicts (each major version of Jest targets a specific major version of Elasticsearch)
 4. The official Elasticsearch REST client is on par with Jest on the lower level (async, discovery, ...) and even has some features that Jest is currently lacking (like [failover](https://hibernate.atlassian.net/browse/HSEARCH-2469) or the ability to ~~[set per-request timeouts](https://hibernate.atlassian.net/browse/HSEARCH-2399)~~ => [Actually, no](https://github.com/elastic/elasticsearch/issues/21789))

You'll notice that, in essence, the new `ElasticsearchWork` interface replaces the `Action` interface from Jest. The main difference being that these works are built by an `ElasticsearchWorkFactory` whose implementation can change depending on the target ES version.

I tried to organize the changes in atomic commits, so that they can be more easily reviewed. If you think I should reorganize some of it so that it's easier to review, please say so, I'll see what I can do (I'll need a hint though ;) ).


Sanne:
> I tried to organize the changes in atomic commits, so that they can be more easily reviewed. If you think I should reorganize some of it so that it's easier to review, please say so, I'll see what I can do (I'll need a hint though ;) ).

Nothing important, but I was confused by the change of directions. In the second commit you introduce an `NoOpBackendRequestResultAssessor` Service, but then it's removed again.

Not worth spending time to polish the commit history; I just mention it in as when I notice such things on my own PRs it typically suggests I've been experimenting but didn't re-start with a clear purpose.

engine/src/main/java/org/hibernate/search/backend/impl/lucene/WorkProcessor.java
```diff
@@ -19,7 +19,7 @@
  * @author Sanne Grinovero (C) 2014 Red Hat Inc.
  * @since 5.0
  */
-interface WorkProcessor {
+public interface WorkProcessor {
```
Sanne: It looks like this wasn't needed?

yrodiere:
> Nothing important, but I was confused by the change of directions. In the second commit you introduce an NoOpBackendRequestResultAssessor Service, but then it's removed again.

Yes, it was either that, or I had to squash this commit and the following one, but they have slightly different purposes. Maybe next time I should squash them anyway?

Sanne:
> Maybe next time I should squash them anyway?

Maybe. It's a tradeoff, I understand it's not always worth to re-work the commits extensively.. I trust you on that. I'm only highlighting it to make sure it's your choice and not by accident ;)

For the record, I'm having massive changes locally which I never sent a PR for as I wanted to polish them more; the reality is I'll have to do the work over as I waited too long and it's easier to re-do the work than fixing conflicts. So, lesson learned .. better to merge fast unless it's easy.

I see other similar cases now, like `ElasticsearchDialectProvider`.. I guess you changed direction after we talked more about Services :+1:

Sanne:
`org.elasticsearch.client.sniff.Sniffer` : why is there support to create these things?
I don't see it ever being used, I'm probably missing something.

yrodiere:
> org.elasticsearch.client.sniff.Sniffer : why is there support to create these things? I don't see it ever being used, I'm probably missing something.

It spawns a thread that does all the work. You don't need to "use" it, which is why we don't: we just create it and close it when it's not needed anymore.

engine/src/main/java/org/hibernate/search/backend/impl/lucene/WorkProcessor.java
yrodiere: Right; I added a fixup commit. Will `rebase --autosquash` when you're done reviewing, I don't want to mess with the commit hashs until then.

Sanne:
merged !

PR #1329, opened 2017-02-27
HSEARCH-2434 Add support for Elasticsearch 5

yrodiere:
~~This PR is based on #1343 , which will have to be merged first.~~ => Done

Relevant JIRA ticket: https://hibernate.atlassian.net/browse/HSEARCH-2434

Compared to #1328, this PR adds the specific code needed to support ES5. To sum up #1328 introduced abstraction layers and interfaces, and this PR introduces alternative implementations of these interfaces.

Don't get fooled by the very high number of additions: I had to copy/paste 4 or 5 large test cases in order to adapt them to ES5. I tried to make them compatible with both ES2 and ES5 first, but the tests ended up almost as complex as the code being tested, so I figured it wasn't a good idea.


yrodiere:
Note: I just pushed an update which added the commit "When validating Elasticsearch schemas, use "default" as a default value for analyzer references" and the next one.
I noticed the issue while working on something else, and since the PR hasn't been merged yet... :]

yrodiere:
Jenkins, retest this please

yrodiere:
Jenkins, retest this please

yrodiere:
@Sanne This should be ready for review.

About the builds: the previous CI build passed, but the one before failed for no apparent reason, so I'm launching another one to see if it was a transient failure that can be blamed on the slave (concurrent build or whatever) or if there is a randomly occurring problem in Hibernate Search itself.

yrodiere:
Jenkins, retest this please

yrodiere:
Jenkins, retest this please

yrodiere:
Jenkins, retest this please

Sanne:
There are some problems with the Maven profiles:
if I run with profile `elasticsearch-2.2` or `elasticsearch-2.0` the "start" goal is no longer valid. I tried to change it to the new goal "runforked" as in the new ES5 profile but then there's another failure relating to not being able to find "org.elasticsearch.plugin:delete-by-query:zip:5.2.1".

The plugin versions and tasks need to have stricter separation among profiles?

yrodiere:
@Sanne So... This situation is not new,  it's been here since I introduced the elasticsearch profiles.

The elasticsearch-5.0 profile is enabled automatically when the `testElasticsearchVersion` is not defined. I had to do this, because just making this a "default" profile means the profile will only be enabled when no other profile is... So, enable "docs" and you lose the elasticsearch5 profile :/

Unfortunately, "a property is not defined" for Maven means "not defined explicitly", and definitions in explicitly enabled profiles do not count. So in your case, you had both the `elasticsearch-2.2` and the `elasticsearch-5.0` profiles enabled, and I suppose the `elasticsearch-5.0` profile defined the `testElasticsearchVersion` property, but `elasticsearch-2.2` defined its own plugin configuration. Which won't work for several reasons.

So, if you want to run 2.0/2.2 tests, you must either:

 * Stick to the instructions in README.md and provide the `testElasticsearchVersion` explicitly:
  ```
   mvn clean install -Pelasticsearch-2.0 -DtestElasticsearchVersion=2.1.0
  ```
 * Or disable the default profile (`elasticsearch-5.0`) explicitly:
  ```
   mvn clean install -P'!elasticsearch-5.0,elasticsearch-2.0'
  ```

I'll see if I can change the poms to make things easier, but I doubt it...

Sanne:
ah right! Sorry I forgot about that. No rush with improvements, I'd like to integrate this first and the README instructions are good enough for now.

Sanne:
awesome work!
It's merged. 

I added one commit to address a couple of super trivial javadoc copy/paste errors.

PR #1330, opened 2017-02-27
HSEARCH-2469 Ability to fail over to the next host when a request to an Elasticsearch host fails

yrodiere:
~~This PR is based on #1329 , which will have to be merged first.~~ => Done

https://hibernate.atlassian.net/browse/HSEARCH-2469

Luckily I didn't have to implement anything: failover is already implemented in the official REST client, which we switched to in #1328. So this is only tests and documentation.


yrodiere:
Jenkins, retest this please

Sanne:
Looks good.

I don't think we'll want to test too many implementation details of the Rest Client but it's good to have the basics.

So the failed nodes are no longer contacted. Do you know for how long? I expect the client to eventually try to see if they recover..

Merging.

yrodiere:
@Sanne The code handling the blacklisting time is located in `DeadHostState`. It seems to start with 1 minute, and then increase with the number of failures to up to 30 minutes.

```
    private static final long MIN_CONNECTION_TIMEOUT_NANOS = TimeUnit.MINUTES.toNanos(1);
    private static final long MAX_CONNECTION_TIMEOUT_NANOS = TimeUnit.MINUTES.toNanos(30);
```

PR #1332, opened 2017-02-28
HSEARCH-2606 Disallow conflicting usage of @Parameter

Sanne:
https://hibernate.atlassian.net/browse/HSEARCH-2606

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/settings/impl/translation/AnalysisDefinitionFactory.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/settings/impl/translation/AnalysisDefinitionFactory.java
@@ -6,12 +6,12 @@
  */
 package org.hibernate.search.elasticsearch.settings.impl.translation;
 
-import org.hibernate.search.annotations.Parameter;
+import java.util.Map;
 
 public interface AnalysisDefinitionFactory<D> {
 
 	String getType();
 
-	D create(Parameter[] parameters);
+	D create(Map<String,String> parameters);
 
 }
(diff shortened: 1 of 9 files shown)


yrodiere:
@Sanne I think the fix is missing on the Elasticsearch side.

Currently the test passes because the analyzer isn't used in your schema, so the analyzer is instantiated as a Lucene analyzer, and then it reaches your check.
But if you assign the analyzer to a field in the `Sample` entity, the analyzer will be correctly detected as an Elasticsearch one and the check will be bypassed.

I added a commit to demonstrate the issue, feel free to squash it.

I suspect that `org.hibernate.search.engine.impl.ConfigContext.addAnalyzerDef(AnalyzerDef, XAnnotatedElement)` might be a better place to put the check, but I suppose you can also duplicate the check on the Elasticsearch side.

Sanne:
pushed a new version, applying this validation more generally across the code.

It also changes an SPI, but I think that's for the better as I like to keep the annotations out of SPI signatures.

Sanne:
re-pushed as I had a mistake in javadoc.

engine/src/main/java/org/hibernate/search/cfg/ParameterAnnotationsReader.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.hibernate.search.annotations.Parameter;
+import org.hibernate.search.util.logging.impl.Log;
+import org.hibernate.search.util.logging.impl.LoggerFactory;
+
+/**
+ * Helper to convert the org.hibernate.search.annotations.Parameter
+ * annotations.
+ */
+public final class ParameterAnnotationsReader {
```
yrodiere: Since this is mainly intended to be used by service providers, maybe we should move this class to `org.hibernate.search.cfg.spi`?

Sanne: Ok I didn't mean to add is an an API. Actually .impl even better?

yrodiere: You can, but that's one more OSGi problem, because this must be used in the Elasticsearch module :)

Sanne:
Moved to SPI

yrodiere:
Jenkins, retest this please

yrodiere:
Merged, thanks!

PR #1333, opened 2017-03-01
magic code removed

vcharmcaster:
I have moved strings to constant variables.
its minor change which improves code quality

### DIFF
--- a/engine/src/main/java/org/hibernate/search/analyzer/definition/impl/LuceneTokenizerDefinitionContextImpl.java
+++ b/engine/src/main/java/org/hibernate/search/analyzer/definition/impl/LuceneTokenizerDefinitionContextImpl.java
@@ -17,6 +17,10 @@ import org.hibernate.search.annotations.TokenizerDef;
  */
 public class LuceneTokenizerDefinitionContextImpl implements LuceneAnalysisDefinitionBuilder<TokenizerDef> {
 
+	private static final String PARAMS2 = "params";
+
+	private static final String FACTORY = "factory";
+
 	private Class<? extends TokenizerFactory> factoryClass;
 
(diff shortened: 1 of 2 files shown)


Sanne:
ok to test

engine/src/main/java/org/hibernate/search/analyzer/definition/impl/LuceneTokenizerDefinitionContextImpl.java
```diff
@@ -17,6 +17,10 @@
  */
 public class LuceneTokenizerDefinitionContextImpl implements LuceneAnalysisDefinitionBuilder<TokenizerDef> {
 
+	private static final String PARAMS2 = "params";
```
yrodiere: Any reason to use the name `PARAMS2` instead of just `PARAMS`? Maybe you used an automated tool?

vcharmcaster: there is params variable present, used eclipse to refactor code.
params2 matches more closely to "params" value.

Sanne:
Hi @vcharmcaster , thanks for sending a patch!

I appreciate you have good intentions to help, but I won't apply this change as it's not necessary: 

- it doesn't fix any problem
- the code quality improvement is questionable.  We prefer the code to be readable.

If you want to help, may I suggest to try having a look at some of these issues?
https://hibernate.atlassian.net/issues/?filter=11250

Some of these might be simple enough for someone new to the project, but it's possible I'm wrong and they might be actually hard. Feel free to try any of these, and change issue if you get stuck.

Also, pick something you're most familiar with. For example I wouldn't try to upgrade JGroups if you're not familiar with using JGroups already.

PR #1334, opened 2017-03-06
HSEARCH-2630 and HSEARCH-2631 Some uses of full text queries will corrupt object loading statistics

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2630
https://hibernate.atlassian.net/browse/HSEARCH-2631


yrodiere:
Note: we should backported this to 5.7, and maybe also to 5.6.

yrodiere:
I just spotted another case where we have a similar problem; will push an update.

yrodiere:
Done; I added a test and fix for HSEARCH-2631.

Sanne:
nice. Merged in master, backporting now..

Sanne:
backported to both 5.6 and 5.7 branches. Branch 5.6 needed a little change as `Session` didn't implement the JPA flavoured `getResultList()` method yet.

Thanks!

PR #1335, opened 2017-03-06
HSEARCH-2541 FullTextQuery with single result does not populate fields or ID in entity results

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2541

This PR doesn't really "solve" the root problem, which is a wrong configuration of Hibernate ORM ([HHH-11550](https://hibernate.atlassian.net/browse/HHH-11550)), but at least it mitigates the issue by avoiding the use of proxies when they are not strictly necessary.


Sanne:
merged! thanks

PR #1336, opened 2017-03-07
HSEARCH-2502 Support the *.worker.execution configuration option on Elasticsearch

yrodiere:
~~This PR is based on #1343, which should be merged first.~~ => Done

Relevant JIRA ticket: https://hibernate.atlassian.net/browse/HSEARCH-2502

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
@@ -17,6 +17,7 @@ import java.util.Set;
 
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.search.similarities.Similarity;
+import org.hibernate.search.backend.BackendFactory;
 import org.hibernate.search.backend.FlushLuceneWork;
 import org.hibernate.search.backend.IndexingMonitor;
 import org.hibernate.search.backend.LuceneWork;
@@ -79,6 +80,7 @@ public class ElasticsearchIndexManager implements IndexManager, IndexNameNormali
 	private String actualIndexName;
 
 	private boolean refreshAfterWrite;
+	private boolean sync;
 	private IndexSchemaManagementStrategy schemaManagementStrategy;
 	private ExecutionOptions schemaManagementExecutionOptions;
 
@@ -135,6 +137,7 @@ public class ElasticsearchIndexManager implements IndexManager, IndexNameNormali
 		String overriddenIndexName = getOverriddenIndexName( indexName, properties );
 		this.actualIndexName = ElasticsearchIndexNameNormalizer.getElasticsearchIndexName( overriddenIndexName );
 		this.refreshAfterWrite = getRefreshAfterWrite( properties );
+		this.sync = BackendFactory.isConfiguredAsSync( properties );
 
 		this.similarity = similarity;
 
@@ -396,7 +399,11 @@ public class ElasticsearchIndexManager implements IndexManager, IndexNameNormali
 
 	@Override
 	public void flushAndReleaseResources() {
-		// no-op
+		ElasticsearchWork<?> flushWork = elasticsearchService.getWorkFactory().flush()
+				.index( actualIndexName )
+				.build();
+		workProcessor.awaitAsyncProcessingCompletion();
+		workProcessor.executeSyncSafe( Collections.singletonList( flushWork ) );
 	}
 
 	@Override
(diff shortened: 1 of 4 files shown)


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexManager.java
```diff
@@ -417,7 +424,14 @@ public void performOperations(List<LuceneWork> workList, IndexingMonitor monitor
 			elasticsearchWorks.add( luceneWork.acceptIndexWorkVisitor( visitor, monitor ) );
 		}
 
-		workProcessor.executeSyncSafe( elasticsearchWorks );
+		if ( sync ) {
+			workProcessor.executeSyncSafe( elasticsearchWorks );
+		}
+		else {
+			for ( ElasticsearchWork<?> work : elasticsearchWorks ) {
```
Sanne: I suspect this isn't dealing with the `ErrorHandler` in the usual way?

The `org.hibernate.search.exception.ErrorContext` is able to highlight a single failure our of a group, where the group usually was meant to be "all operations from the current transaction".

It's probably not easy to apply the `ErrorHandler` in the usual way as it normally would handle a failure as something like "out of operations [A, B, C] we failed on B " somehow suggesting that C wasn't executed.

In this case the various operations might have been applied out of order so you should collect all success/fail operations to forge a new kind of error state.

Sanne: I wouldn't consider this a blocker and not inclined to invest too much time on this as we want to evolve the ErrorHandler API.

Since errors are handled individually, users aren't going to miss them.

Just add a comment?

Sanne:
Usability question: which combinations of _refreshAfterWrite_ and _sync_ (fields of *ElasticsearchIndexManager*) belong to which use case?

Do we need to clarify in the docs which combinations are suggested?

Finally, do you think any of these combinations are nonsense and should issue a log warning?

yrodiere:
>Usability question: which combinations of refreshAfterWrite and sync (fields of ElasticsearchIndexManager) belong to which use case?

I would say that:

 * sync is about ensuring all works have been submitted to Elasticsearch before returning from a transaction
 * refreshAfterWrite is about ensuring changes to the index are visible before returning from a transaction

So with that in mind:

 * sync + refreshAfterWrite => Fully synchronous: works have been executed, errors have been reported, and results are visible upon returning from a transaction.
 * sync + NOT refreshAfterWrite => Execution-synchronous: works have been executed, errors have been reported, and but results may not be visible upon returning from a transaction.
 * async + NOT refreshAfterWrite => Asynchronous: no guarantee as to whether works have been executed, or even transmitted to Elasticsearch
 * async + refreshAfterWrite => Asynchronous with forced near real-time : no guarantee as to whether works have been executed, or even transmitted to Elasticsearch, but guarantees to make results visible "soon" after works have been executed (see below)

>Do we need to clarify in the docs which combinations are suggested?
> Finally, do you think any of these combinations are nonsense and should issue a log warning?

So, after digging up a little:

 * every combination will _work_, even sync = false and refreshAfterWrite = true
 * in synchronous mode, every combination makes sense
 * if choosing asynchronous mode, "refreshAfterWrite = true" makes little sense in my opinion, since it basically means "refresh the index at some unknown point in the future, after you completed the given works". Which is already something Elasticsearch guarantees: periodical index refreshes.
   So in general I don't see the point of such a configuration, and I think it could only hurt if enabled without thinking about it.
   That being said... I'm no Elasticsearch expert, but the [ES docs do mention a way to customize the refresh interval](https://www.elastic.co/guide/en/elasticsearch/guide/master/near-real-time.html). So I guess that, in async mode, `refreshAfterWrite` could be used to set the refresh interval in Elasticsearch to some high value, and to force more frequent refreshes when works are requested by Hibernate Search.
 * in asynchronous mode, the "refresh after write" will occur whenever the `RequestProcessingRunnable` will run out of works. Due to the fact we add asynchronous works to the queue one by one, this basically means the refresh could happen anytime.
   I think I could change the code so that when asked to execute 10 works asynchronously, those 10 works are added to the queue atomically. Then we'd be sure that "refreshes after write" would only be executed before or after this work set, but never between works in the set. This isn't strictly necessary, but it could provide a (probably very slight) performance gain.

So to sum up... I guess the async + refreshAfterWrite use case could make sense, even if it sounds a bit far-fetched. So I would vote against a warning log; maybe an info-level log. But we definitely need some documentation.

What do you think if I add something similar to my list of use cases above in the documentation about `refresh_after_write false`? Would this be enough?

Sanne:
Thanks for the summary, it's good to think about these combinations.

I agree with most of your thoughts, but to make sure we're on the same page:

 - remember that by default we send operations to the backend _after_ the database changes are committed. So your first case (sync + refreshAfterWrite) is described correctly as you state, but remember that even if an error gets synhcornously reported it won't abort the database changes. Essentially there's a third parameter for these combinations: happens within the transaction vs happens after commit.

 - for the case "async + refreshAfterWrite" - which you agree makes less sense - I believe that ES won't actually perform any periodic refresh on its own if no changes have been applied so even that minor benefit is not really there. Essentially we should agree that this mode is quite pointless, should we warn against it (by logging a warning)? I'm tempted to not to, as it's quite harmless other than being less efficient - but who knows is someone finds better reasons to do this.

Sanne:
merged!

Opened https://hibernate.atlassian.net/browse/HSEARCH-2693 as an optionally follow-up. I'm happy to wait with it, as I suspect our own experience is about to improve in this area.

PR #1337, opened 2017-03-07
HSEARCH-2521 Clarify if error HSEARCH000278 could be caused by the entity not being a managed entity

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2521


Sanne:
merged

PR #1338, opened 2017-03-07
HSEARCH-2565 Boolean QueryBuilder clauses should accept a 'null' clause

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2565


Sanne:
merged

PR #1339, opened 2017-03-07
HSEARCH-2353 Add support for the 'norms' mapping attribute with Elasticsearch 5

yrodiere:
~~This PR is based on #1329, which will have to be merged first.~~ => Done

Relevant JIRA ticket: https://hibernate.atlassian.net/browse/HSEARCH-2353

I didn't add support for this on ES2 because the syntax is not the same as in ES5, it's more complex with more options (lazy/eager loading for instance). See [the docs for ES2](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/norms.html) and [the equivalent for ES5](https://www.elastic.co/guide/en/elasticsearch/reference/5.2/norms.html).

I'm not convinced there would be much value, but if you want I can have a look at how to support norms on ES2.


Sanne:
merged! thanks

PR #1341, opened 2017-03-08
HSEARCH-2628 JNDI support should reuse the JNDIService from Hibernate…

Sanne:
… ORM

A small enhancement, for multiple reasons:
 - *duplication* : delete duplicate code with ORM
 - *usability* : don't force people to set those initial context et al properties twice
 - *modularity* : would be nice to remove our direct dependency on _javax.naming_

The configuration property change will need notes in the migration guide; I considered introducing a "deprecation warning" but that would introduce too much complexity. I'll prefer the clean cut this time, it's unlikely people won't notice, and it's just a configuration property.

 - https://hibernate.atlassian.net/browse/HSEARCH-2628


engine/src/main/java/org/hibernate/search/engine/service/named/impl/NoopNamedResolver.java
```diff
@@ -0,0 +1,26 @@
...
+import org.hibernate.search.exception.SearchException;
+
+/**
+ * This implementation of NamedResolver is not functional:
+ * integrator code is expected to provide a custom implementation
+ * to replace this.
+ * This implementation is provided to generate an appropriate
+ * error when no replacement is provided.
+ */
+public class NoopNamedResolver implements NamedResolver {
```
yrodiere: Shouldn't we add this to `org.hibernate.search.cfg.Environment.DEFAULT_SERVICES_MAP`? `NoopNamedResolver` is not used anywhere, so I don't think it is serving its purpose right now...

Sanne: good catch! forgot about that.. not that missing or having it would make much of a difference ;)

PR #1343, opened 2017-03-17
HSEARCH-2581 follow-up to fix performance issues

yrodiere:
~~This PR is based on #1345 , which we will have to merge first.~~ => Done

https://hibernate.atlassian.net/browse/HSEARCH-2581

Follow-up on #1328.

I'm creating this PR to trigger a build on our CI.
This should fix the problems we've been experiencing on CI lately.


yrodiere:
Jenkins, retest this please

yrodiere:
Jenkins, retest this please.

yrodiere:
Jenkins, retest this please.

Sanne:
Jenkins, retest this please.

Sanne:
The commits look good, I would like to merge this.

But the build is still failing with a similar error:

```
Caused by: java.io.IOException: Too many open files
	at sun.nio.ch.IOUtil.makePipe(Native Method)
	at sun.nio.ch.EPollSelectorImpl.<init>(EPollSelectorImpl.java:65)
	at sun.nio.ch.EPollSelectorProvider.openSelector(EPollSelectorProvider.java:36)
	at java.nio.channels.Selector.open(Selector.java:227)
	at org.apache.http.impl.nio.reactor.AbstractMultiworkerIOReactor.<init>(AbstractMultiworkerIOReactor.java:142)
	... 65 more
```

Shall I leave the PR open?

It's possible that we simply have to raise the limits.

Sanne:
I have a theory - I couldn't verify yet - but I suspect the problem is simply caused by the fact that the reactive threads of the async http client aren't shut down aggressively but just "scheduled to close gracefully".

This graceful shutdown happens rather quickly, but it should take at least a full second as the selector needs to wake up first (and has a 1s timeout by default), so we don't notice any leak when debugging a single test yet I see ~2500+ threads over JConsole during the testsuite.

The testsuite takes longer than 1s but since the amount of threads to be shut down quickly raises in the hundreds, the scheduler isn't going to wake them all up as aggressively so this 1s wakeup time becomes longer.. coumpounding the problem.

I think we either find a way to block and wait for the shutdown to have happened, or we just avoid the explosive scheduling issue by setting a lower setting in the testsuite:


`elasticsearch/src/test/resources/hibernate.properties` :

    hibernate.search.default.elasticsearch.max_total_connection 2

yrodiere:
@Sanne Interesting.
You commit actually reduces the number of leaked file descriptors for each search factory from 64 to 10. It's better, though I don't understand why 10 times less threads translates into 6.4 times less leaked file descriptors.
On the other hand, setting max_total_connection to 2 means we end up with ours tests only being able to perform 2 concurrent calls to Elasticsearch. Which supposedly was what made performance tests fail on the CI, and is what I tried to solve in my first commit...

So I'm a bit torn here: it seems we end up exactly where we started from, and had some luck with the CI... I'll try to see what those 10 remaining threads are, maybe it can shed some light on the issue...

yrodiere:
I just rebased to also implement the improvements in `TestElasticsearchClient`. Still investigating, but one intersting thing is that the `TestElasticsearchClient` does not leak threads for some reason: the threads of its `RestClient` are correctly terminated, while those of the main `RestClient` (the one used by the `SearchFactory`) never are. `RestClient.close` is invoked in both cases, though...

yrodiere:
@Sanne Here, everything should be fine now. The build seems to run fine (no timeout) on ci-slave-4, even after I removed my initial commit that raised the number of I/O reactor threads. So I guess the two remaining commits should be enough...

Sanne:
looks good! Testing one last time..

Sanne:
thanks! merged

PR #1344, opened 2017-03-19
Hsearch 2581

Sanne:
Duplicate of #1343 : just want to have Jenkins test with a slightly different configuration.


yrodiere:
Closing, since I added your commit to #1343

PR #1345, opened 2017-03-20
HSEARCH-2277 Some resources may not be cleaned on bootstrap failure

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2277


Sanne:
I'm having some failures in the ES module:


    Failed tests: 
    SearchFactoryTest.testIndexManagerStoppedAfterIndexInitializationException:147 expected:<1> but was:<0>
    SearchFactoryTest.testIndexManagerStoppedAfterIndexManagerSearchFactorySetupException:207 expected:<1> but was:<0>
    SearchFactoryTest.testServicesStoppedAfterIndexInitializationException:105 expected:<1> but was:<0>
    SearchFactoryTest.testServicesStoppedAfterIndexManagerSearchFactorySetupException:177 expected:<1> but was:<0>
    
    Tests run: 1196, Failures: 4, Errors: 0, Skipped: 27

engine/src/main/java/org/hibernate/search/engine/impl/ImmutableAnalyzerRegistry.java
```diff
@@ -67,13 +77,19 @@ public AnalyzerReference getAnalyzerReference(Class<?> analyzerClazz) {
...
-	}
-
-	private void close(Collection<AnalyzerReference> references) {
-		for ( AnalyzerReference reference : references ) {
+		for ( AnalyzerReference reference : getAllReferences() ) {
 			reference.close();
 		}
 	}
 
+	private List<AnalyzerReference> getAllReferences() {
```
Sanne: not important as it's not an hot code area, but it looks a bit wasteful to add them all to a new collection ?
You could just close them all.

BTW should we invoke each close() method in a try/catch "log & ignore" block?

yrodiere: @Sanne I wanted to use a Stream, but decided against it in order to be able to backport.
I guess I'll add a good old closeAll() method.

> BTW should we invoke each close() method in a try/catch "log & ignore" block?
I'll try to do that. Ideally we should throw an exception if anything goes wrong and still close every reference, but in order to do that, we should do it everywhere, and unfortunately we don't: https://hibernate.atlassian.net/browse/HSEARCH-2640

yrodiere:
Jenkins, test this please

engine/src/test/java/org/hibernate/search/test/searchfactory/SearchFactoryTest.java
```diff
@@ -49,6 +67,158 @@ public void testTypeWithNoDocumentIdThrowsException() {
 	}
 
 	@Test
+	@Category(ElasticsearchSupportInProgress.class) // HSEARCH-2481 Byteman-based tests re-executed in the Elasticsearch module won't work
+	@TestForIssue(jiraKey = "HSEARCH-2277")
+	@BMRules(rules = {
+			@BMRule(
+					name = "Skip failure on first index manager initialization",
```
Sanne: What's the need for this rule? It doesn't look like you need this Flag, you can just omit the condition in the second rule?

engine/src/test/java/org/hibernate/search/test/searchfactory/SearchFactoryTest.java
```diff
@@ -49,6 +67,158 @@ public void testTypeWithNoDocumentIdThrowsException() {
...
+	} )
+	public void testServicesStoppedAfterIndexInitializationException() {
+		SearchConfigurationForTest cfg = new SearchConfigurationForTest()
+				.addClass( AnnotatedClass.class ).addClass( SecondAnnotatedClass.class );
+
+		try {
+			new SearchIntegratorBuilder().configuration( cfg ).buildSearchIntegrator();
+		}
+		catch (SearchException e) {
+			// Ignore the exception
```
Sanne: I would actually do something here so that the test can assert that the `SearchException` was thrown.

e.g. flip a boolean, so that you can assert on the boolean after the catch block?

Otherwise you have the risk that the Services are indeed released, but just because the failures didn't happen.

I'm being extra paranoid here because the Byteman rules tend to break without us noticing when they rely on specific signatures...

Sanne:
merged

PR #1347, opened 2017-03-24
Refactor JBoss modules for new ES client requirements

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2645
 - https://hibernate.atlassian.net/browse/HSEARCH-2646

### DIFF
--- a/modules/src/main/assembly/dist.xml
+++ b/modules/src/main/assembly/dist.xml
@@ -65,6 +65,21 @@
             <outputDirectory>/org/hibernate/search/elasticsearch/${hibernate.search.version}</outputDirectory>
             <filtered>true</filtered>
         </file>
+        <file>
+            <source>${module.xml.basedir}/es-client/module.xml</source>
+            <outputDirectory>/org/hibernate/search/elasticsearch-client/${hibernate.search.version}</outputDirectory>
+            <filtered>true</filtered>
+        </file>
+        <file>
+            <source>${module.xml.basedir}/es-sniffer/module.xml</source>
+            <outputDirectory>/org/hibernate/search/elasticsearch-sniffer/${hibernate.search.version}</outputDirectory>
+            <filtered>true</filtered>
+        </file>
+        <file>
+            <source>${module.xml.basedir}/gson/module.xml</source>
+            <outputDirectory>/org/hibernate/search/elasticsearch-gson/${hibernate.search.version}</outputDirectory>
+            <filtered>true</filtered>
+        </file>
     </files>
 
     <dependencySets>
@@ -140,9 +155,36 @@
             <unpack>false</unpack>
             <includes>
                 <include>org.hibernate:hibernate-search-elasticsearch</include>
-                <include>com.google.code.gson:gson</include>
+            </includes>
+        </dependencySet>
+
+        <dependencySet>
+            <useProjectArtifact>false</useProjectArtifact>
+            <outputDirectory>org/hibernate/search/elasticsearch-client/${hibernate.search.version}</outputDirectory>
+            <useTransitiveFiltering>false</useTransitiveFiltering>
+            <unpack>false</unpack>
+            <includes>
                 <include>org.elasticsearch.client:rest</include>
+            </includes>
+        </dependencySet>
+
+        <dependencySet>
+            <useProjectArtifact>false</useProjectArtifact>
+            <outputDirectory>org/hibernate/search/elasticsearch-sniffer/${hibernate.search.version}</outputDirectory>
+            <useTransitiveFiltering>false</useTransitiveFiltering>
+            <unpack>false</unpack>
+            <includes>
                 <include>org.elasticsearch.client:sniffer</include>
+            </includes>
+        </dependencySet>
+
+        <dependencySet>
+            <useProjectArtifact>false</useProjectArtifact>
+            <outputDirectory>org/hibernate/search/elasticsearch-gson/${hibernate.search.version}</outputDirectory>
+            <useTransitiveFiltering>false</useTransitiveFiltering>
+            <unpack>false</unpack>
+            <includes>
+                <include>com.google.code.gson:gson</include>
             </includes>
         </dependencySet>
 
--- a/modules/src/main/modules/search/elasticsearch/module.xml
+++ b/modules/src/main/modules/search/elasticsearch/module.xml
@@ -8,17 +8,15 @@
 <module xmlns="urn:jboss:module:1.1" name="org.hibernate.search.elasticsearch" slot="${hibernate.search.version}">
     <resources>
         <resource-root path="hibernate-search-elasticsearch-${hibernate.search.version}.jar" />
-        <resource-root path="rest-${elasticsearchClientVersion}.jar" />
-        <resource-root path="sniffer-${elasticsearchClientVersion}.jar" />
-        <resource-root path="gson-${elasticsearchGsonVersion}.jar" />
     </resources>
     <dependencies>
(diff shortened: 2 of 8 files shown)


modules/src/main/assembly/dist.xml
```diff
@@ -140,6 +145,15 @@
...
+            </includes>
+        </dependencySet>
+
+        <dependencySet>
+            <useProjectArtifact>false</useProjectArtifact>
+            <outputDirectory>org/hibernate/search/elasticsearch-client/${hibernate.search.version}</outputDirectory>
+            <useTransitiveFiltering>false</useTransitiveFiltering>
+            <unpack>false</unpack>
+            <includes>
                 <include>com.google.code.gson:gson</include>
```
yrodiere: The rest client doesn't depend on gson (unlike Jest), so this should be moved to `org.hibernate:hibernate-search-elasticsearch` just above.

modules/src/main/modules/elasticsearch/module.xml
```diff
@@ -0,0 +1,18 @@
...
+ ~ Hibernate Search, full-text search for your domain model
+ ~
+ ~ License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ ~ See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+  -->
+<module xmlns="urn:jboss:module:1.1" name="org.hibernate.search.elasticsearch-client" slot="${hibernate.search.version}">
+    <resources>
+        <resource-root path="rest-${elasticsearchClientVersion}.jar"/>
+        <resource-root path="sniffer-${elasticsearchClientVersion}.jar"/>
+        <resource-root path="gson-${elasticsearchGsonVersion}.jar"/>
```
yrodiere: Same here, gson should not appear here.

modules/src/main/modules/elasticsearch/module.xml
```diff
@@ -0,0 +1,18 @@
...
+<!--
+ ~ Hibernate Search, full-text search for your domain model
+ ~
+ ~ License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ ~ See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+  -->
+<module xmlns="urn:jboss:module:1.1" name="org.hibernate.search.elasticsearch-client" slot="${hibernate.search.version}">
+    <resources>
+        <resource-root path="rest-${elasticsearchClientVersion}.jar"/>
+        <resource-root path="sniffer-${elasticsearchClientVersion}.jar"/>
```
yrodiere: Actually the property should be `elasticsearchSnifferVersion`. Or we could remove `elasticsearchSnifferVersion` from the pom if you want, but I'm not 100% sure the release cycles are synchronized.

modules/src/main/modules/search/elasticsearch/module.xml
```diff
@@ -8,17 +8,13 @@
 <module xmlns="urn:jboss:module:1.1" name="org.hibernate.search.elasticsearch" slot="${hibernate.search.version}">
     <resources>
         <resource-root path="hibernate-search-elasticsearch-${hibernate.search.version}.jar" />
-        <resource-root path="rest-${elasticsearchClientVersion}.jar" />
-        <resource-root path="sniffer-${elasticsearchClientVersion}.jar" />
-        <resource-root path="gson-${elasticsearchGsonVersion}.jar" />
```
yrodiere: gson should have remained here, *we* use it but the rest client does not.

modules/src/main/modules/elasticsearch/module.xml
```diff
@@ -0,0 +1,18 @@
...
+        <resource-root path="gson-${elasticsearchGsonVersion}.jar"/>
+    </resources>
+    <dependencies>
+        <module name="org.apache.httpcomponents"/>
+        <module name="org.apache.commons.logging"/>
```
yrodiere: The dependency to `javax.xml.bind.api` should be moved here, since only the sniffer uses that.

modules/src/main/modules/search/elasticsearch/module.xml
```diff
@@ -8,17 +8,13 @@
...
     </resources>
     <dependencies>
-        <module name="org.hibernate.search.engine" slot="${hibernate.search.version}" />
-        <module name="org.hibernate.commons-annotations" />
+        <module name="org.hibernate.search.engine" slot="${hibernate.search.version}"/>
+        <module name="org.hibernate.search.elasticsearch-client" slot="${hibernate.search.version}"/>
         <module name="javax.xml.bind.api"/>
```
yrodiere: I don't think we depend on `javax.xml.bind.api`, only the rest client does (the sniffer, actually).

modules/src/main/modules/elasticsearch/module.xml
Sanne: No our own hibernate-search-elasticsearch requires it too.

modules/src/main/modules/search/elasticsearch/module.xml
Sanne: That's not how the modules work :)
Since *we* use it, it should be in a module *we depend on*.

The intent of the isolation is to avoid transitive dependencies: we do export all of `hibernate-search-elasticsearch` to the end user's classpath so that people can use our public API contained in there, but we don't want them to be able to use our `gson` jar as it might conflict with their own copy.

modules/src/main/assembly/dist.xml
Sanne: Good point, but I don't want to expose it to our users.

I'll move gson to its own module then..

modules/src/main/modules/search/elasticsearch/module.xml
yrodiere: Hum, ok. I thought this was about laying the groundwork to contribute a Wildfly module to Elastic, but it seems I was confused.

modules/src/main/modules/elasticsearch/module.xml
yrodiere: Right, in a single class: ElasticsearchDateHelper.

modules/src/main/modules/search/elasticsearch/module.xml
Sanne: Yes I'll separate gson as I mentioned in the above comment: it doesn't belong in either module.

modules/src/main/modules/elasticsearch/module.xml
yrodiere: But I think we still need to add a dependency to a Jackson module here (even if we leave it in `hibernate-search-elasticsearch`). The class `org.elasticsearch.client.sniff.ElasticsearchHostsSniffer` depends on Jackson classes to parse JSON.

... unless modules visible by the `hibernate-search-elasticsearch` module are automatically visible by its dependencies?

Now which module to depend on, I'm not sure. I think `org.jboss.resteasy.resteasy-jackson-provider` would do the trick, but maybe you'll want to be more precise and just depend on `org.codehaus.jackson.jackson-core-asl`, which should be enough.

Sanne: I was just looking into that :)
Unfortunately the Wildfly provided module uses Jackson `1.9.13` which is way older than what the sniffer expects.

And that's our fault, as we depend on an old version of Avro which requires that version of Jackson :-/

I'm trying.

Sanne:
I've updated the modules structure.

I think this reflects now what we want; although to capture issues with the Sniffer module I've opened the follow up HSEARCH-2655

`commons-codec` was missing as well.

yrodiere:
Merged, thanks!

PR #1348, opened 2017-03-24
HSEARCH-2648 Add debug/trace logs showing execution time of Elasticsearch requests

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2648


yrodiere:
@Sanne Rebased, it should work fine now :)

Sanne:
Are you sure that making LuceneLogCategories extensible (rather than an enum) isn't a bit too much?

Was it because you didn't want to add Elasticsearch specific categories to the enum in hibernate-search-engine?

I see it as a nice move as other extension points - like Infinispan - will be able to create their own categories but I don't think an external integrator will actually do this, as they'll have their own logging conventions.

On the other hand, the previous `enum` approach had the nice effect of giving us (and users) a single place to look to have a list of all categories.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/client/impl/DefaultElasticsearchClient.java
```diff
@@ -59,6 +67,10 @@ public Response execute(ElasticsearchRequest request) throws IOException {
 			 */
 			return e.getResponse();
 		}
+		finally {
+			long executionTime = System.currentTimeMillis() - start;
```
Sanne: To measure durations don't use `System.currentTimeMillis()`, use System.nanoTime()

yrodiere: Right; fixed.

yrodiere:
> Are you sure that making LuceneLogCategories extensible (rather than an enum) isn't a bit too much?
> 
> Was it because you didn't want to add Elasticsearch specific categories to the enum in hibernate-search-engine?

Yes.

> I see it as a nice move as other extension points - like Infinispan - will be able to create their own categories but I don't think an external integrator will actually do this, as they'll have their own logging conventions.

I did not expect them to do this either, to be honest. Just wanted to avoid leaky abstractions internally.

> On the other hand, the previous enum approach had the nice effect of giving us (and users) a single place to look to have a list of all categories.

It doesn't seems _very_ hard to get a listing of all categories; Ctrl + Shift + G with my cursor on the constructor of `LogCategory` does the trick in Eclipse.

Do you want me to revert the change? I find this cleaner, but either way it will work.

Sanne:
> Do you want me to revert the change? I find this cleaner, but either way it will work.

No, it's ok.

I just wanted to explore the reasoning with you. BTW I don't expect all users to always work with sources open, so we'll have to create a section in the documentation.

I'll merge this when back from lunch.. didn't go yet ;)

yrodiere:
> so we'll have to create a section in the documentation

Pre-existing loggers are already documented [here](https://docs.jboss.org/hibernate/search//5.7/reference/en-US/html_single/#_optimizing_the_query_process) and [here](https://docs.jboss.org/hibernate/search//5.7/reference/en-US/html_single/#lucene-infostream). And I added a section in the Elasticsearch docs in this PR.
But maybe we should have a dedicated section, and make those I just mentioned redirect there?

Sanne:
> But maybe we should have a dedicated section, and make those I just mentioned redirect there?

Yes that's what I meant. No rush though, just adding ideas for eventually reorganizing the docs.

Sanne:
merged! thanks

PR #1349, opened 2017-03-27
HSEARCH-2651 Calls to numericField in the programmatic API create duplicate fields

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2651


Sanne:
merged & backported.. thanks!

PR #1351, opened 2017-03-29
HSEARCH-1316 Support @Inject in ClassBridges and FieldBridge instances

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-1316

So... This will be a self-service PR :) I expect some commits not to make it, but I added the whole lot so we can discuss it. Mainly:

 * "HSEARCH-1316 Allow integrators to defer the initialization of the search factory using a provided EnvironmentSynchronizer" is absolutely necessary for Wildfly, and yet I'm not comfortable with the extent of the changes. It adds multithreaded logic where we previously had none, and in critical areas on top of that. It's definitely worth a second (and maybe a third) look, and we may want to defer this to 6.0. Also, we may want to add some timeouts in the calls to `CompletableFuture.join`, or at least provide a way for integrators to cancel the boot.
 * I don't know if I should document this feature. This is very clearly an SPI, but let's face it, it is not likely to be used by integrators anytime soon. So users will probably want to do the integration themselves, especially since it's not very complex (see the Spring ITs for instance). Maybe I should add some documentation in [10.7. Using external services](https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#section-services)?
 * @gsmet I made sure it is *possible* for integrators to only handle beans with a specific annotation, but I did not mandate it. It's up to integrators to decide whether having only part of the beans in their injection context makes sense, and it's one less oddity in our code. And like I said, when the user does not specify a bridge implementation explicitly (through `@FieldBridge.impl`), we don't use injection anyway. I think it should be satisfying?
 * @Sanne you mentioned some whitelist to avoid trying to instantiate our own classes through an external `BeanProvider`, for performance reasons. But... I cannot rely on the package name, because obviously our own integration tests are in the `org.hibernate.search` package, and having an actual exhaustive whitelist seems a sure way to end up with an outdated list. Any ideas?


Sanne:
I'd like to see if we can simplify this PR a bit. I'll start by pushing afe4bf9 to master already.  [done]

I probably would have sent a preamble PR "Upgrade Spring integration tests to use a recent Spring Boot version", but ok don't change it now as I'm almost done understanding it all :)

yrodiere:
@Sanne Rebased just now.

Sanne:
Thanks, let's see if I can merge this today

integrationtest/spring/src/test/java/org/hibernate/search/test/integration/spring/injection/SpringInjectionIT.java
```diff
@@ -0,0 +1,121 @@
...
+		assertThat( search.apply( NonSpringBridge.PREFIX + InternationalizedValue.HELLO.name() ) ).onProperty( "id" ).isEmpty();
+
+		EntityWithSpringAwareBridges entity = new EntityWithSpringAwareBridges();
+		entity.setInternationalizedValue( InternationalizedValue.HELLO );
+		dao.create( entity );
+		assertThat( search.apply( NonSpringBridge.PREFIX + InternationalizedValue.HELLO.name() ) ).onProperty( "id" ).containsOnly( entity.getId() );
+	}
+
+	@Test
+	public void explicitNonInjectedFieldBridge() {
```
yrodiere: @Sanne I think there's something missing here... I'll dive back into it soon and will try to add a proper test if necessary.

yrodiere:
@Sanne I just rebased to remove the [empty test I noticed](https://github.com/hibernate/hibernate-search/pull/1351#pullrequestreview-38127235) (it was not needed) and to resolve a conflict in the logger. Nothing changed apart from that!

Sanne:
cool, thanks

Sanne:
superseded by #1406

PR #1352, opened 2017-03-30
HSEARCH-2644 Mention what needs to be done to switch Elasticsearch version in docs

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2644


Sanne:
merged, thanks!

PR #1355, opened 2017-03-31
HSEARCH-2009 Document supported types for @FieldBridge#impl()

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2009


Sanne:
merged!

PR #1356, opened 2017-03-31
HSEARCH-2656 Correctly detect numeric fields when they are default fields with their type overridden by a custom MetadataProvidingFieldBridge

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2656

Note: given there is a behavior change, I'm not sure we should backport this. The new behavior is definitely better, though.


Sanne:
I'll follow your suggestion and change the JIRA to flag "fixed for" only on 5.8. We can always backport later if need be.

Sanne:
merged! thanks

PR #1357, opened 2017-03-31
Various bugfixes and improvements in ElasticsearchHSQueryImpl

yrodiere:
Grouping these as a single PR, because they touch the same area of code, and I'd rather not do a hundred separate rebases :)

Relevant tickets:

 * https://hibernate.atlassian.net/browse/HSEARCH-2361 : Make sure Elasticsearch related configuration properties are parsed and validated at boot time
 * https://hibernate.atlassian.net/browse/HSEARCH-2662: Querying an empty, dynamically sharded index won't work (Lucene + Elasticsearch)
   About this one: it actually "worked" with Elasticsearch, but we used to send a search query without specifying the index to search on, with type filters that reduced the results to nothing. So it was more a performance issue than anything else for the Elasticsearch part.
 * https://hibernate.atlassian.net/browse/HSEARCH-2661: Avoid the recurring calls to serviceManager.requestReference in ElasticsearchHSQueryImpl

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/DefaultElasticsearchService.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/DefaultElasticsearchService.java
@@ -9,6 +9,7 @@ package org.hibernate.search.elasticsearch.impl;
 import java.io.IOException;
 import java.util.Properties;
 
+import org.hibernate.search.elasticsearch.cfg.ElasticsearchEnvironment;
 import org.hibernate.search.elasticsearch.client.impl.ElasticsearchClient;
 import org.hibernate.search.elasticsearch.client.impl.ElasticsearchClientFactory;
 import org.hibernate.search.elasticsearch.client.impl.ElasticsearchClientImplementor;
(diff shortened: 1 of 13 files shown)


yrodiere:
Jenkins, retest this please

yrodiere:
Jenkins, retest this please

yrodiere:
The build seems to fail on Jenkins, but not on Travis for some reason... Having a look right now.

yrodiere:
Jenkins, retest this please

yrodiere:
After a rebase, we only have the same timeout issues as on other branches (including master). Marking as "Ready for review".

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/DefaultElasticsearchService.java
```diff
@@ -48,6 +50,10 @@
 	 */
 	private static final String CLIENT_SCOPE_NAME = "default";
 
+	private static final String QUERY_PROPERTIES_PREFIX = "hibernate.search.";
```
Sanne: There's no need to use this prefix, we automatically pre-mask all read properties with an appropriate prefix.

More importantly, we should never force this specific prefix usage as other integrators use a custom prefix for the Search component.

But since I see this was an existing problem I'll take the PR and send a follow-up fix.

Sanne:
nice one. merged! thanks

PR #1358, opened 2017-04-03
HSEARCH-2382 Merge the JPA and Hibernate native components like ORM did

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2382

The main thing to discuss here is: what do we do when two implementations of a given method are different on the JPA and the native side?
I chose to be consistent with what was done in Hibernate ORM when possible, and to use the JPA behavior for our own methods. See the commit comments for more details.

Note that whatever we do, we will have to change the behavior in some cases. Unless we don't merge `FullTextEntityManagerImpl` with `FullTextSessionImpl`, in which case we could pass a flag to `FullTextQueryImpl` indicating which behavior to use (JPA or native).


orm/src/main/java/org/hibernate/search/jpa/FullTextEntityManager.java
```diff
@@ -42,10 +42,13 @@
 
 	/**
 	 * Force the (re)indexing of a given <b>managed</b> object.
-	 * Indexing is batched per transaction: if a transaction is active, the operation
+	 * Indexation is batched per transaction: if a transaction is active, the operation
```
Sanne: why are you changing this? `Indexation` is not the correct term. It needs to be `Indexing`. I now it's wrong in some other places, but don't make it worse :)

orm/src/main/java/org/hibernate/search/jpa/Search.java
```diff
@@ -43,7 +44,41 @@ else if ( em instanceof FullTextEntityManager ) {
...
+		}
+		else if ( EntityManager.class.isAssignableFrom( delegate.getClass() ) ) {
+			//Some app servers wrap the EM twice
+			delegate = ( (EntityManager) delegate ).getDelegate();
+			if ( delegate == null ) {
+				throw new SearchException(
+						"Trying to use Hibernate Search without an Hibernate EntityManager (no delegate)"
+				);
+			}
+			else if ( Session.class.isAssignableFrom( delegate.getClass() ) ) {
```
Sanne: `instanceof` is démodé ?

Sanne: ah, it's just moved code. Ok...

Sanne:
very nice! thanks, merged

orm/src/main/java/org/hibernate/search/jpa/FullTextEntityManager.java
yrodiere: I think I just copy/pasted the comments from FullTextSession, which I guess was more recent. I failed to see this difference.

PR #1360, opened 2017-04-04
HSEARCH-2670 Facet drill-down will never return any result when the source field has a different name

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2670

There are a few behavior changes that may prevent us from backporting this PR:

 * Faceting will no longer work when there is no faceting metadata (but I suppose we don't support creating facets from custom bridges?)
 * Errors when selecting field bridge that don't have faceting enabled will now happen sooner (when querying, instead of when retrieving the facets).

### DIFF
--- a/engine/src/main/java/org/hibernate/search/query/dsl/impl/AbstractFacet.java
+++ b/engine/src/main/java/org/hibernate/search/query/dsl/impl/AbstractFacet.java
@@ -18,13 +18,15 @@ import org.hibernate.search.query.facet.Facet;
  */
 public abstract class AbstractFacet implements Facet {
 	private final String facetingName;
-	private final String fieldName;
+	private final String facetFieldName;
+	private final String sourceFieldName;
 	private final String value;
 	private final int count;
 
-	public AbstractFacet(String facetingName, String fieldName, String value, int count) {
+	public AbstractFacet(String facetingName, String facetFieldName, String sourceFieldName, String value, int count) {
 		this.facetingName = facetingName;
-		this.fieldName = fieldName;
+		this.facetFieldName = facetFieldName;
+		this.sourceFieldName = sourceFieldName;
 		this.count = count;
 		this.value = value;
 	}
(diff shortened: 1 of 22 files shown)


yrodiere:
@Sanne rebased just now

Sanne:
> There are a few behavior changes that may prevent us from backporting this PR

Good points. Neither change seems too bad, but in such cases we can simply agree to not backport it, and re-evaluate if there's demand for it.

yrodiere:
> in such cases we can simply agree to not backport it, and re-evaluate if there's demand for it.

I certainly agree with that. Especially since this fix is the best I could do when [trying to fix another bug](https://hibernate.atlassian.net/browse/HSEARCH-2668?focusedCommentId=92608&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-92608), which unfortunately cannot be fixed without breaking the APIs (I think). So basically nobody asked for this specific fix, but since it was already done...

engine/src/main/java/org/hibernate/search/query/dsl/impl/AbstractFacet.java
```diff
@@ -39,9 +41,21 @@ public String getValue() {
...
 	public String getFieldName() {
-		return fieldName;
+		return getFacetFieldName();
+	}
+
+	public String getFacetFieldName() {
+		return facetFieldName;
+	}
+
+	public String getSourceFieldName() {
```
Sanne: shall we call this _getSourcePropertyName()_ ?

I tend to try disambiguate the entity properties from the Lucene fields by calling them respectivelly "property" and "field" in such cases.

Sanne: ah sorry, I'm definitely distracted. this is an index field :)

yrodiere: In this specific case, we are actually speaking about a field... See [here](https://github.com/hibernate/hibernate-search/pull/1360/files/8a1cb838838a7435b2e2787e5766ea2d788884bb#diff-2bb3b66e5fbcd04002af53e6f647d57bR33) and the `getFacetQuery` method a few lines below to see why.

Maybe "Source" isn't the right term, though. "Attached"? "Origin"? I'm not sure how to call it. It's basically whatever we pass to `@Facet.forField`...

yrodiere: > ah sorry, I'm definitely distracted. this is an index field :)

Ah, you were too quick. Yes :)

Sanne:
very nice cleanup! merged

PR #1361, opened 2017-04-06
HSEARCH-2660 Make index schema management strategy options more consistent with JPA

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2660

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/IndexSchemaManagementStrategy.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/IndexSchemaManagementStrategy.java
@@ -8,6 +8,7 @@ package org.hibernate.search.elasticsearch.cfg;
 
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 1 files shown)


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/cfg/IndexSchemaManagementStrategy.java
```diff
@@ -49,11 +53,42 @@
...
+	static {
+		Map<String, IndexSchemaManagementStrategy> tmpMap = new HashMap<>();
+		for ( IndexSchemaManagementStrategy strategy : values() ) {
+			tmpMap.put( strategy.externalName, strategy );
+		}
+		VALUES_BY_EXTERNAL_NAME = Collections.unmodifiableMap( tmpMap );
+	}
+
+	public static IndexSchemaManagementStrategy interpretPropertyValue(String propertyValue) {
+		IndexSchemaManagementStrategy strategy = VALUES_BY_EXTERNAL_NAME.get( propertyValue );
```
Sanne: Traditionally we've been more forgiving about case preferences and trailing whitespaces.

Could you change this strategy from using a map to attempt a ignore-case equals on each enum name, and trim the propertyValue in advance as well?

Sanne: Ideally we'd want some of the configurations to use an intentionally inconsistent case in property value, to make sure it works ;)

yrodiere: I can just trim and lowercase the property value before passing it to `.get()`, it will achieve the same effect. Will do.

PR #1362, opened 2017-04-06
HSEARCH-1886 Master indexing nodes fail to lookup lazily initialized IndexManager instances

yrodiere:
~~This PR is based on #1369 (HSEARCH-2684), which will have to be merged first.~~ => Done

https://hibernate.atlassian.net/browse/HSEARCH-1886

So, let's be honest: that one was hard. Really. And it's only the beginning.

I didn't follow the suggestion in the ticket (initialize all the indexes at boot time), because this is _dynamic_ sharding, so even if we initialize something at boot time, ultimately there will be a new shard and we'll end up with the exact same issue.

There is a good deal of changes, but each of them was necessary for one reason or another. The comments in the commits should be self-explanatory; if not, please ask.

To sum up:

 * The first three commits solve one issue: the fact that index managers were not lazily created/updated when a shard appeared on a slave first (only the sharding strategy on the slave was aware of the change, so the master never created the index manager). I suspect that we'll still have issues with other slaves not being aware of the shard addition, but hey, maybe that's worth [another ticket](https://hibernate.atlassian.net/browse/HSEARCH-2676) :)
 * The following commits solve the other issues: the fact that message consumers on the master node are not started until an index manager (shard) is actually created. So if a slave happened to create a shard first, its messages were simply ignored...


yrodiere:
Note: I just rebased this on #1357, since one commit was already present on that other PR (the fix about queries on dynamically sharded indexes that don't have any shard yet).

yrodiere:
Builds are failing because of some timeouts because the CI is slow (https://hibernate.atlassian.net/browse/HSEARCH-2684). I'll send a PR to fix that first.

yrodiere:
I sent a PR to fix the random test failures (#1369) and rebased this PR on the other. The build should be a bit more stable now.

yrodiere:
Rebased on master following the merging of #1369.

Sanne:
Why do you say that the validation on consistent `Similarity` is useless?

It can be configured in many ways - some are legacy - and we had several cases in which people would get themselves into trouble (before introducing this check).

Commits like 5dd7f80 should best be handled in separate PRs, so that it doesn't slow down / distract from the issue we're supposed to solve.

yrodiere:
> Why do you say that the validation on consistent Similarity is useless?
> It can be configured in many ways - some are legacy - and we had several cases in which people would get themselves into trouble (before introducing this check).

I only see two ways to configure it (see `createSimilarity`): the index-level configuration, and the default configuration. Since entities are not taken into account when retrieving the `Similarity` (only the index and the more global `SearchConfiguration`), I don't see how we could have two entities in the same index with a different `Similarity`... ?
Maybe there _was_ a legacy way to configure similarity, and it got removed?

> Commits like 5dd7f80 should best be handled in separate PRs, so that it doesn't slow down / distract from the issue we're supposed to solve.

Hey, this PR depended on two other fixes, which I did address in separate PRs, so I don't *always* put irrelevant commits in my PRs :) I'll try to be more careful though.

yrodiere:
Jenkins, retest this please

yrodiere:
Jenkins, retest this please

Sanne:
About the similarity: you're right, sorry I got confused. There were several other options with a focus on the entity, but like you suggested they have been removed already. I remembered deprecating them but forgot we got rid of them.

Sanne:
It's taking me a long time to wrap my head around the changes, but I like it so far. Reminiscent of various things that I wished to do in the past but it never was a good time in terms of branch maintenance.

On the kind of Dispatchers: I'm not seeing any specific problem, but remember that there's more going on than simple "streams of changes" and "transactional changes".

We need to keep in mind also that in different circumstances there are strong expectations about ordering of fired events: order needs to be strictly maintained even in "async mode", which means we can't just delegate to a parallel executor at any stage.

To the contrary when doing batch rebuilding (MassIndexer) the strict ordering is a burden which prevents fast index rebuilds. However even in that case, there are some operations which need to be kept in order:
 - the purgeAll before the MassIndexer runs need to happen strictly before the subsequent tasks
 - any final optimisation needs to happen after the last (less critical but still desirable)
 
Finally we need to allow for the "transactional operations" and MassIndexer operations to happen concurrently: say the MassIndexer is processing, and in parallel some database changes are performed. Our listeners will have to stack up the events fired from the transactional changes making sure that their relative order is maintained, and also that they overwrite any in-flight work from the mass-indexer. 

In the past it seems I had a clear picture in mind of the backend design, as it was simple enough to reason about this and see that it was correct. I can no longer load the whole picture in mind by simply reading the source code, so I think we should eventually add some tests at least to ensure the ordering properties.

Adding such tests is unrelated with this PR but I felt like sharing the concern already as this new design is making some subtle changes in the area.

yrodiere:
About the dispatchers, to be honest I just moved some code around so I could reuse it someplace else :) I didn't intend to make any change to work orchestration.> To the contrary when doing batch rebuilding (MassIndexer) the strict ordering is a burden which prevents fast index rebuilds. However even in that case, there are some operations which need to be kept in order:

But I agree work orchestration is quite complex, and thus might need some (more?) tests.

> To the contrary when doing batch rebuilding (MassIndexer) the strict ordering is a burden which prevents fast index rebuilds. However even in that case, there are some operations which need to be kept in order:

It seems to me that such "synchronisations" in the work stream are what @gustavonalle added recently, with `org.hibernate.search.backend.impl.batch.DefaultBatchBackend.awaitAsyncProcessingCompletion()`. After sending a `purgeAll` command, we would call this method, and so would we before final optimizations.

Sanne:
great work, I've merged it. I applied two additional trivial commits.

PR #1365, opened 2017-04-07
HSEARCH-2001 By default enable a "strict mode" validation for sorting

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2001


Sanne:
merged!

PR #1369, opened 2017-04-13
HSEARCH-2684 Improve timeout handling in tests

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2684

This should put an end to the random JMS test failures on the slower CI nodes...


integrationtest/jms/src/test/java/org/hibernate/search/test/jms/master/JMSMasterTest.java
```diff
@@ -69,6 +73,10 @@
 	 */
 	private static final String CONNECTION_FACTORY_NAME = "java:/ConnectionFactory";
 
+	public static final Poller POLLER = Poller.milliseconds( 10_000, 100 );
+
+	private final QueryParser parser = new QueryParser( "id", TestConstants.stopAnalyzer );
```
Sanne: In this case it's fine, but generally speaking remember that such a `QueryParser` is not threadsafe. People will have to be careful when reusing helper methods from this class.

Sanne:
Lovely pattern. In Infinispan we have a similar one, I believe that's what had inspired Gustavo to add that `Condition` which you now removed (but it wasn't the full thing - I believe Gustavo tried to keep it simple).

I'm not super convinced of the "step" parameter: it's not simple to guess an adequate value when writing the test, so it would be even better by using auto-detection. For example hard code an initial step of 100 nanoseconds and double it at each failed attempt.

None of the above are blocking problems - I'd have merged it already if it weren't for the fact that all integration tests are failing.

integrationtest/wildfly/src/test/java/org/hibernate/search/test/integration/jms/transaction/TransactionalJmsMasterSlave.java
```diff
@@ -181,42 +184,21 @@ public void searchRollbackedMemberAfterSynchronizationOnSlave2() throws Exceptio
...
-		assertEquals( "Richard Mayhew", findAtLeastOneName( "richard" ).get( 0 ) );
+		POLLER.pollAssertion( () -> assertAtLeastOneName( "Davide D'Alto", "davide" ) );
+		POLLER.pollAssertion( () -> assertAtLeastOneName( "Peter O'Tall", "peter" ) );
+		POLLER.pollAssertion( () -> assertAtLeastOneName( "Richard Mayhew", "richard" ) );
 	}
 
-	private void assertSearchResult(String expectedResult, List<RegisteredMember> results) {
+	private void assertAtLeastOneEntity(String expectedResult, String searchString) {
+		List<RegisteredMember> results = memberRegistration.search( searchString );
 		assertEquals( "Unexpected number of results from search", 1, results.size() );
```
Sanne: why is this check called "at least one" but verifies "exactly one" ?

integrationtest/jms/src/test/java/org/hibernate/search/test/jms/master/JMSMasterTest.java
yrodiere: Right... I'm afraid the problem is widespread though, because I've already seen this pattern (the query parser as a test attribute) multiple times...

integrationtest/wildfly/src/test/java/org/hibernate/search/test/integration/jms/transaction/TransactionalJmsMasterSlave.java
yrodiere: Right, I got confused with the other "assert" method (which I didn't implement right, either. I fixed both methods.

yrodiere:
@Sanne Thanks, I fixed the assert methods and the deployments. Everything runs fine on Travis now, let's wait for Jenkins to agree.

Sanne:
nice, merged

PR #1370, opened 2017-04-18
HSEARCH-2687 Upgrade Hibernate ORM to 5.2.10.Final

Sanne:
The dependency convergence problem was avoided by ORM, just in time before the release, by changing the pom structure.

 - https://hibernate.atlassian.net/browse/HSEARCH-2687


yrodiere:
Merged; thanks!

PR #1371, opened 2017-04-19
HSEARCH-2688 Clearing scroll IDs (when closing ScrollableResults) fails systematically with Elasticsearch

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2688

Note: there's no need to backport this after all, as explained in comments on the JIRA ticket.


Sanne:
merged, thanks!

PR #1372, opened 2017-04-21
HSEARCH-2690 FileHelper based index replication attempts to delete lo…

Sanne:
…cked files on Windows

I decided to change the semantics of the method, even though we occasionally use it for other purposes as those are just testing. Adding a new method for this seemed overkill.

 - https://hibernate.atlassian.net/browse/HSEARCH-2690

### DIFF
--- a/engine/src/main/java/org/hibernate/search/util/impl/FileHelper.java
+++ b/engine/src/main/java/org/hibernate/search/util/impl/FileHelper.java
@@ -124,7 +124,7 @@ public class FileHelper {
 		}
 		else {
 			if ( destination.exists() && destination.isDirectory() ) {
-				delete( destination );
+				tryDelete( destination.toPath() );
 			}
 			if ( destination.exists() ) {
 				long sts = source.lastModified() / FAT_PRECISION;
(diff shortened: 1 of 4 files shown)


Sanne:
@Hibernate-CI re test this please

yrodiere:
Jenkins, retest this please

engine/src/main/java/org/hibernate/search/util/logging/impl/Log.java
```diff
@@ -1029,4 +1029,8 @@
 	@Message(id = 335, value = "Unable to build a Lucene query from the query string '%1$s'.")
 	EmptyQueryException unableToBuildLuceneQueryFromQueryString(String query);
 
+	@LogMessage(level = Level.DEBUG)
+	@Message(id = 336, value = "A file could not be deleted: likely lock contention. Not a problem for index replications as it will be attempted again in the future.")
+	void fileDeleFailureIgnored(@Cause IOException e);
```
yrodiere: Shouldn't this be "fileDeleteFailureIgnored"? I don't get why we would use an abbreviation here.

engine/src/main/java/org/hibernate/search/util/impl/FileHelper.java
```diff
@@ -184,10 +184,11 @@ public static void delete(File file) throws IOException {
...
 	/**
 	 * Attempts to delete a file. If the file is a directory delete recursively all content.
+	 * Any IOException preventing a file to be deleted will be swallowed.
 	 *
 	 * @param path the file or directory to be deleted
 	 *
-	 * @throws IOException if it wasn't possible to delete all content which is a common problem on Windows systems.
+	 * @throws IOException on unexpected io errors
 	 */
 	public static void delete(Path path) throws IOException {
```
yrodiere: Given the  new behavior, maybe renaming this to "tryDelete" would make sense?

engine/src/main/java/org/hibernate/search/util/logging/impl/Log.java
Sanne: thanks yes that's just a typo

engine/src/main/java/org/hibernate/search/util/impl/FileHelper.java
Sanne: ok; In that case I would deprecate it.

Sanne:
thanks, good ideas. Applied them, then actually differentiated the methods to make sure there's no mistake and merged.

PR #1374, opened 2017-04-23
HSEARCH-2691 Custom filter implementations must provide proper implementations of the #equals() and #hashCode() methods.

golovnin:
https://hibernate.atlassian.net/browse/HSEARCH-2691


Sanne:
Thanks @golovnin ! Applied and backported to all 4 actively maintained branches: 5.5 5.6 5.7 (and master, 5.8)

PR #1375, opened 2017-04-25
Reduce test verbosity and warnings

Sanne:
This also improves build execution times a bit.

 - https://hibernate.atlassian.net/browse/HSEARCH-2694

### DIFF
--- a/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
+++ b/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
@@ -677,7 +677,7 @@ you need to enable the Groovy plugin by adding the following snippet to your Ela
 
 .Enabling Groovy support in your elasticsearch.yml
 ----
-script.engine.groovy.inline.search: on
+script.engine.groovy.inline.search: true
 ----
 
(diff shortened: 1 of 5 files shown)


yrodiere:
@Sanne Don't you want to do the same for ES2? As it's an embedded instance, it uses the same configuration as the rest of our tests, though. Maybe the configuration for ES2 should be in `sharedtestresources/src/main/resources/log4j.properties`?

Sanne:
@yrodiere no I'm not too interested in spending time polishing the ES2 tests :)

Speaking of resources, I actually think they shouldn't be in `src/[test|main]/resources` at all as we don't load them from our testsuite.
In fact I had them in the root before, but you moved them in the resources. Having them in `sharedtestresources` seems interesting but that's a task for another day.. not sure if the ES plugin can deal with that.

yrodiere:
> Speaking of resources, I actually think they shouldn't be in src/[test|main]/resources at all as we don't load them from our testsuite.

Created https://hibernate.atlassian.net/browse/HSEARCH-2696.

yrodiere:
Looks good, merged!

PR #1376, opened 2017-04-25
HSEARCH-2695 Speedup initialization of Elasticsearch client by skippi…

Sanne:
…ng SSL engine initialization when not required

 - https://hibernate.atlassian.net/browse/HSEARCH-2695


yrodiere:
Merged, thanks!

PR #1377, opened 2017-04-25
HSEARCH-2697 Only use setAccessible when strictly necessary

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2697

This fixes the last remaining problem in the -engine build on JDK9. The -orm build still seems to fail on JDK9, mainly due to Hibernate ORM issues, but that's another story :)


Sanne:
nice solution, merged!

PR #1378, opened 2017-04-25
HSEARCH-2698 Solve some warnings from the build logs

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2698


Sanne:
merged, thanks!

PR #1379, opened 2017-04-25
HSEARCH-2698 Move the empty Jar requirements to the parent pom

Sanne:
Follow up on #1378 : I already merged it but then realized that the new `skipIffEmpty` requirement fails the build when building with `-Pdist` (as it's missing in the distribution module).

Let's move the attribute to the parent pom?

### DIFF
--- a/infinispan/pom.xml
+++ b/infinispan/pom.xml
@@ -34,14 +34,6 @@
             <plugin>
                 <artifactId>maven-deploy-plugin</artifactId>
             </plugin>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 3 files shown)


yrodiere:
@Sanne Sorry about that... I think we'd better fix this by declaring the distribution module as a POM module, though. This won't change anything for potential consumers, as the assembly will still produce tar.gz and zip artifacts; we just won't have an empty JAR artifact anymore. Which seems to make sense.
I'm on it, it's currently building and I'll submit a PR as soon as it succeeded.

yrodiere:
Closing for now; superseded by #1381. Feel free to reopen if necessary.

PR #1381, opened 2017-04-26
HSEARCH-2698 Make the distribution module a POM artifact, instead of a JAR artifact

yrodiere:
Follow-up on #1378 , supersedes #1379

Relevant ticket (**I reopened it**): https://hibernate.atlassian.net/browse/HSEARCH-2698

We don't produce any JAR for this module, and we produce *multiple*
artifacts, so the easier solution is to just not declare anything.

The travis build is in progress: https://travis-ci.org/yrodiere/hibernate-search/builds/225892334
I ran a quick build locally without tests and it went fine.

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -19,6 +19,7 @@
 
     <name>Hibernate Search Distribution</name>
     <description>Builds the distribution bundles</description>
+    <packaging>pom</packaging>
 
     <properties>
         <!-- Skip artifact deployment -->


yrodiere:
The CI is down right now (because of the upgrade to maven 3.5, I think), but the travis build succeeded: https://travis-ci.org/yrodiere/hibernate-search/builds/225892334

Sanne:
> The CI is down right now (because of the upgrade to maven 3.5, I think)

FYI upgrading Maven took 2 seconds and doesn't require any downtime. To clarify: we didn't update all builds to use the new Maven but it's an opt-in so most builds are still using whatever versions they where using before.

Sanne:
This PR looks good and since it's urgent I'll merge it already, but I'm confused about why we need those empty jars in the other modules now.

Assuming that indeed having any module not produce an empty jar will fail the build, then why not declare this in the parent pom as in my other PR?

Finally, do we need to review the "skipDeploy" attributes? We don't want to release these empty jars which were added. with the previous PR.

Sanne:
merged!

yrodiere:
> This PR looks good and since it's urgent I'll merge it already, but I'm confused about why we need those empty jars in the other modules now.

Actually that's the opposite: what I did in the previous PR was to skip the production of JARs when there is nothing to put in it. When skipIfEmpty is true, no JAR will be generated if there is no classes/resources to put in it.

This is an improvement for most modules, since we won't publish empty, useless JARs anymore. We mostly used to for pom-packaging modules and for test-jar artifacts in modules without tests.

But in the case of the "distribution" module, this module had no classes and still had a JAR packaging. Maven doesn't like when a module has a JAR packaging and doesn't produce any JAR, so the build started to fail. This PR solves the issue by switching the packaging of this module to POM, which seems to make sense.

Sanne:
yes I agree it makes sense to switch to POM in this specific case.

I'm confused about why it's necessary to skip the jar building of the other modules though, and if so why not declare it on the parent pom rather than re-defining the rule in each individual pom ?

Incidentally several other modules are broken now, I'm sending another follow-up PR.

yrodiere:
> I'm confused about why it's necessary to skip the jar building of the other modules though
Well because we don't want to publish empty JARs, especially empty tests-jars in modules where there are no tests... do we?

> and if so why not declare it on the parent pom rather than re-defining the rule in each individual pom ?
Well that's what I did... ? I added a default rule to not produce any JAR when there nothing to put in it, and overrode this rule where we need to produce empty JARs for some reason.

Sanne:
> Because we don't want to publish empty JARs, especially empty tests-jars in modules where there are no tests... do we?

Fair enough, it's indeed pointless but remember we only publish what gets deployed to Maven central: we can "skip deploy" on case by case and still use the jars for other purposes during the build.

> Well that's what I did... ? I added a default rule to not produce any JAR when there nothing to put in it, and overrode this rule where we need to produce empty JARs for some reason.

oh right :) Sorry I got confused by some other patch I've been reviewing doing the opposite.

PR #1384, opened 2017-04-27
HSEARCH-2014 Expose size of the indexes among statistics

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2014


Sanne:
merged! thanks

PR #1385, opened 2017-04-28
HSEARCH-2703 Improve boost handling when translating Lucene queries to Elasticsearch

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2703


Sanne:
merged! thanks

PR #1386, opened 2017-04-28
HSEARCH-2704

Sanne:
https://hibernate.atlassian.net/browse/HSEARCH-2704

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/client/impl/Paths.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/client/impl/Paths.java
@@ -1,0 +1,34 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.elasticsearch.client.impl;
+
+/**
+ * Useful paths to compose Elasticsearch URLs.
+ */
+public interface Paths {
+
+	URLEncodedString _EXPLAIN = URLEncodedString.fromString( "_explain" );
+	URLEncodedString _QUERY = URLEncodedString.fromString( "_query" );
+	URLEncodedString _FLUSH = URLEncodedString.fromString( "_flush" );
+	URLEncodedString _OPTIMIZE = URLEncodedString.fromString( "_optimize" );
+	URLEncodedString _DELETE_BY_QUERY = URLEncodedString.fromString( "_delete_by_query" );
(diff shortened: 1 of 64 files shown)


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/schema/impl/model/IndexMetadata.java
```diff
@@ -20,17 +21,17 @@
  */
 public class IndexMetadata {
 
-	private String name;
+	private URLEncodedString name;
```
yrodiere: It would be a bit clearer if we used a dedicated `IndexIdentifier` type, able to return a simple string or an encoded one. Mainly because it's really not obvious why the name is URL-encoded, both here and in the IndexManager.

... but it would once again require two hours of find and replace, so I think we'll stick with this :p

Sanne: yes I agree with that and I actually was tempted to do it.. but for sake of getting this done I stopped polishing :)

+1 to do such a thing at another round

Sanne:
> More seriously, great that you found this. I guess this will was right in your way when trying to benchmark the ES integration?

Yes. It wasn't a big deal as we have other bottlenecks to solve first, but this was a low hanging fruit and  it would become a problem eventually.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/client/impl/Paths.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.elasticsearch.client.impl;
+
+/**
+ * Useful paths to compose Elasticsearch URLs.
+ */
+public interface Paths {
```
Sanne: what's the worst thing that will happen if someone implements it? ;)

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/client/impl/URLEncodedString.java
```diff
@@ -0,0 +1,90 @@
...
+	 * @param string a string to encode
+	 * @return
+	 */
+	public static URLEncodedString fromString(String string) {
+		Objects.requireNonNull( string );
+		return new URLEncodedString( string );
+	}
+
+	/**
+	 * @param jsonElement
```
Sanne: I'll remove the javadoc. I feel it's easier to read the one line of code than understanding the explanations I tried to draft...

Sanne:
thanks, applied your suggestions and merged. I didn't change the interface, it's in the .impl package anyway.

PR #1387, opened 2017-05-02
HSEARCH-2709 Add support for converting MatchNoDocsQuery to Elasticsearch

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2709

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ToElasticsearch.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ToElasticsearch.java
@@ -22,6 +22,7 @@ import org.apache.lucene.search.Filter;
 import org.apache.lucene.search.FilteredQuery;
 import org.apache.lucene.search.FuzzyQuery;
 import org.apache.lucene.search.MatchAllDocsQuery;
+import org.apache.lucene.search.MatchNoDocsQuery;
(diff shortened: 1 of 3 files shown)


gsmet:
Jenkins, retest this please

gsmet:
Jenkins, retest this please

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ToElasticsearch.java
```diff
@@ -255,6 +259,20 @@ private static JsonObject convertMatchAllDocsQuery(MatchAllDocsQuery matchAllDoc
 		return JsonBuilder.object().add( "match_all", new JsonObject() ).build();
 	}
 
+	private static JsonObject convertMatchNoDocsQuery(MatchNoDocsQuery matchNoDocsQuery) {
```
gsmet: Probably not worth it for this but at one point I suppose we will need to inject the dialect in the conversion mechanism.

yrodiere: You're right. We managed to avoid it until now though, and if we can avoid it until 6.0 we may not even have to do it at all: I'm not sure we'll want to support Lucene => ES conversion once Hibernate Search APIs don't depend on Lucene anymore.

gsmet:
Merged, thanks!

PR #1388, opened 2017-05-03
HSEARCH-2208 Avoid the org.apache.lucene.search.Filter API

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2208

So, that wasn't a "low-hanging fruit"... I should have guessed. Anyway, it's done.
We should probably wait for Sanne to give it a look before we merge this PR. @Sanne: don't worry, it can wait a few weeks.

Some context: `org.apache.lucene.search.Filter` has been deprecated in favor of `org.apache.lucene.search.Query`, since `Query` is now immutable and offers as much possibility for filtering than `Filter`. Thus, we need to move away from `Filter`, which will be removed in a future version of Lucene.

Main changes:

 * I replaced references to the Filter type with references to the Query type everywhere I could in internal code
 * I added support to use a `Query` instance instead of a `Filter` instance for named filters (`@FullTextFilterDef`). I updated the documentation so as to promote using the `Query` type, and warn against `Filter` being deprecated.
 * I added default methods to `FilterCachingStrategy`, which is an API using the Filter type, to make it possible to implement it without relying on the `Filter` type. It should make the transition a bit smoother for users.
 * I deprecated `HSQuery#filter`
 * I duplicated some filter-related tests so that we test both approach: filtering through a Query and filtering through a Filter (see the `org.hibernate.search.test.filter.deprecated` package introduced in "	HSEARCH-2208 Test filtering with Query objects instead of Filter".


Sanne:
@yrodiere could you rebase it and trigger a new build please?

yrodiere:
@Sanne Done, build in progress

Sanne:
very nice! Merged it.

I've added one additional change - it's trivial so I don't believe it needed review but feel free to have a look: 61540cce72db5ca2f6243a58494075af0f393370

PR #1389, opened 2017-05-03
HSEARCH-2680 Investigate alerts from lgtm

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2680

Turns out it was mostly false positives.


Sanne:
Looks good, but didn't we have more warnings?

yrodiere:
@Sanne The others are either questionable or false positives:

 * Suspicious reference equality test of boxed types => We're just checking whether both variables are null, so it's okay
 * Possible confusion of local and field => The first one is legit, but solved in another PR, and the other ones reflect a standard practice in our codebase, so I don't think it's a problem
 * Self assignment => Was already solved in a previous PR (seems the analysis is out-of-date)
 * Result of integer multiplication cast to long => Doesn't seem that critical to me, but I can solve it if you want
 * Non-synchronized override of synchronized method => False positive, we only override methods to throw UnsupportedOperationException so we don't need synchronization

Sanne:
nice. Merged

PR #1390, opened 2017-05-03
HSEARCH-2667 Fulltext Query timeout methods should no longer be flagged as experimental

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2667

Apparently that's all there is to do: the fact that these methods were
experimental wasn't mentioned in the documentation.


Sanne:
merged! thanks

PR #1391, opened 2017-05-04
HSEARCH-2458 Throw an exception when an Elasticsearch field has incomplete metadata

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2458

To be perfectly clear, this situation only arises with Elasticsearch when applying a `@NumericField` annotation on a field that uses a non-numeric default bridge, or on a field that uses a custom bridge that doesn't provide metadata. So only when users misconfigured their mapping.


gsmet:
Merged, thanks!

Sanne:
resolving

PR #1392, opened 2017-05-04
HSEARCH-2675 + HSEARCH-2710 Multiple JGroups fixes

yrodiere:
Relevant JIRA tickets:

 * [HSEARCH-2710](https://hibernate.atlassian.net/browse/HSEARCH-2710): Restore master checking code in JGroupsMasterMessageListener. Note that the fix is not a simple revert of e62b27eeadb73c3c867866f4aeaed14fb77baa98, since I tried to retain the benefits of [HSEARCH-1886](https://hibernate.atlassian.net/browse/HSEARCH-1886) (which added partial support for dynamic sharding on JGroups)
 * [HSEARCH-2675](https://hibernate.atlassian.net/browse/HSEARCH-2675): JGroupsBackendQueueProcessor doesn't switch from slave to master dynamically. It turns out the problem was only visible with 3 or more nodes, which may explain why we ever thought it worked.


Sanne:
merged!

PR #1394, opened 2017-05-04
HSEARCH-2636 Throw a clear exception when an entity identifier/class could not be retrieved from the indexed document

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2636


gsmet:
Merged, thanks!

PR #1395, opened 2017-05-08
added hint to .travis.yml about deactivated state

krichter722:
added hint that the main CI is http://ci.hibernate.org/ and that Travis is supposed to be used for github forks (see https://hibernate.atlassian.net/browse/HSEARCH-2712)


Sanne:
thanks, good idea!

PR #1396, opened 2017-05-10
HSEARCH-2477 Shard filtering doesn't work with Elasticsearch queries

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2477

~~Not to be merged immediately, I am going to do a release.~~ => Done

### DIFF
--- a/engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
+++ b/engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
@@ -11,6 +11,7 @@ import static org.hibernate.search.util.impl.CollectionHelper.newHashMap;
 
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
@@ -33,8 +34,10 @@ import org.hibernate.search.engine.spi.DocumentBuilderIndexedEntity;
 import org.hibernate.search.engine.spi.EntityIndexBinding;
 import org.hibernate.search.exception.SearchException;
 import org.hibernate.search.filter.FullTextFilter;
+import org.hibernate.search.filter.FullTextFilterImplementor;
 import org.hibernate.search.filter.ShardSensitiveOnlyFilter;
 import org.hibernate.search.filter.impl.FullTextFilterImpl;
+import org.hibernate.search.indexes.spi.IndexManager;
 import org.hibernate.search.metadata.NumericFieldSettingsDescriptor.NumericEncodingType;
 import org.hibernate.search.query.engine.spi.HSQuery;
 import org.hibernate.search.query.engine.spi.TimeoutExceptionFactory;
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/hibernate/search/query/engine/impl/LuceneHSQuery.java
```diff
@@ -282,6 +279,25 @@ protected void extractFacetResults() {
...
 
+	@Override
+	protected List<IndexManager> getIndexManagers(EntityIndexBinding binding) {
+		List<IndexManager> indexManagers = super.getIndexManagers( binding );
+
+		for ( IndexManager indexManager : indexManagers ) {
+			if ( !( indexManager instanceof DirectoryBasedIndexManager ) ) {
+				//TODO clarify intent:
+				// A) improve the error message so that people understand what they should do
+				// B) Is the point really to not support "directory-based" or rather non-Lucene native based ones?
```
gsmet: Hmmm, it looks weird to limit this to directory based indexes. I would expect this method to work even for RAM based indexes.

I then realized that it was already existing code but I thought I would mention it anyway.

engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
```diff
@@ -530,6 +535,25 @@ private FacetMetadata findFacetMetadata(FacetingRequest facetingRequest, Iterabl
 		return facetMetadata;
 	}
 
+	protected List<IndexManager> getIndexManagers(EntityIndexBinding binding) {
+		FullTextFilterImplementor[] fullTextFilters = getFullTextFilterImplementorsAsArray();
+		List<IndexManager> indexManagers = Arrays.asList( binding.getSelectionStrategy().getIndexManagersForQuery( fullTextFilters ) );
+		return indexManagers;
+	}
+
+	private FullTextFilterImplementor[] getFullTextFilterImplementorsAsArray() {
```
gsmet: Is the fact that it returns an Array really important? I wouldn't include it in the method name.

And I must admit I preferred the old method name altogether (`getFullTextFilter()`). It's more to the point and doesn't add useless information.

engine/src/main/java/org/hibernate/search/query/engine/impl/LuceneHSQuery.java
yrodiere: RAM-based indexes are directory-based: they use a `RAMDirectoryProvider`. See `BaseDirectoryProviderService`.

yrodiere: But you're right: this comment should disappear eventually (why the hell create a TODO in the code without a corresponding ticket???). I created [HSEARCH-2724](https://hibernate.atlassian.net/browse/HSEARCH-2724).

engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
yrodiere: FullTextFilter is much too vague, we could be returning half a dozen different things, and none of them would be handled the same way:

 * `Query`
 * `Filter`
 * `FullTextFilterImplementor`
 * `FullTextFilterDef` (!)
 * `FullTextFilter` (!)
 * etc.

I'll remove the "asArray" suffix, though.

yrodiere:
@gsmet Thanks, I removed the "asArray" suffix. I will merge and backport as soon as the CI build succeeds.

yrodiere:
Merged and backported to 5.6 and 5.7.

PR #1397, opened 2017-05-11
HSEARCH-2718 Make the analyzer definition provider interfaces API instead of SPI

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2718


Sanne:
merged

PR #1400, opened 2017-05-12
HSEARCH-2721 Fix transient failures in DefaultElasticsearchClientFactoryTest.discovery

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2721

Below is a description of the method I followed to check that the fix is effective.

* run this shell code before making any change:
```
function countFailures() {
	FAILURE=0
	SUCCESS=0;
	while true
	do
		if eval "${@}" 1>/dev/null 2>&1
		then
			(( ++ SUCCESS ))
		else
			(( ++ FAILURE ))
		fi
		echo "Success: $SUCCESS, failure: $FAILURE"
	done
}
mvn clean install -DskipTests=true
countFailures mvn test -pl elasticsearch -rf elasticsearch " -Dtest=DefaultElasticsearchClientFactoryTest#discovery"
```
* notice that for 50 runs, you get approximately 5 failures
* apply the changes in this commit
* run the same shell code again
* notice that you don't get any failures even after 50 runs

### DIFF
--- a/elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/DefaultElasticsearchClientFactoryTest.java
+++ b/elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/DefaultElasticsearchClientFactoryTest.java
@@ -28,6 +28,7 @@ import org.hibernate.search.elasticsearch.client.impl.URLEncodedString;
 import org.hibernate.search.elasticsearch.impl.JsonBuilder;
 import org.hibernate.search.test.util.impl.ExpectedLog4jLog;
 import org.hibernate.search.testsupport.TestForIssue;
+import org.hibernate.search.testsupport.concurrency.Poller;
 import org.hibernate.search.testsupport.setup.SearchConfigurationForTest;
 import org.junit.Rule;
 import org.junit.Test;
@@ -45,6 +46,8 @@ import com.google.gson.JsonParser;
  * @author Yoann Rodiere
  */
 public class DefaultElasticsearchClientFactoryTest {
+
+	private static final Poller POLLER = Poller.milliseconds( 10_000, 500 );
 
 	private static final JsonParser JSON_PARSER = new JsonParser();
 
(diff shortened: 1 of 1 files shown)


gsmet:
Funny we have another transient failure when testing this one :).

Jenkins retest this please

elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/DefaultElasticsearchClientFactoryTest.java
```diff
@@ -310,7 +310,7 @@ public void discovery() throws Exception {
 		SearchConfigurationForTest configuration = new SearchConfigurationForTest()
 				.addProperty( CLIENT_PROPERTY_PREFIX + ElasticsearchEnvironment.SERVER_URI, httpUrlFor( wireMockRule1 ) )
 				.addProperty( CLIENT_PROPERTY_PREFIX + ElasticsearchEnvironment.DISCOVERY_ENABLED, "true" )
-				.addProperty( CLIENT_PROPERTY_PREFIX + ElasticsearchEnvironment.DISCOVERY_REFRESH_INTERVAL, "1" );
+				.addProperty( CLIENT_PROPERTY_PREFIX + ElasticsearchEnvironment.DISCOVERY_REFRESH_INTERVAL, "2" );
```
Sanne: Why do you think raising the refresh time would help?

elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/DefaultElasticsearchClientFactoryTest.java
```diff
@@ -331,12 +331,18 @@ public void discovery() throws Exception {
 			Response result = doPost( client, "/myIndex/myType", payload );
 			assertThat( result.getStatusLine().getStatusCode() ).as( "status code" ).isEqualTo( 200 );
 
-			Thread.sleep( 2000 ); // Wait for the refresh to occur
+			Thread.sleep( 4000 ); // Wait for the refresh to occur
```
Sanne: Needing this test to wait `4s` is annoying and I would accept that if it was the only way, but we all know that these figures are going to fail in some other slightly different condition.

Is there no way with this mock library to "wait until" the condition is satisfied?

I'm confident you could code that yourself but I'd also expect a library like this to have some out of the box support for such kind of tests?

elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/DefaultElasticsearchClientFactoryTest.java
yrodiere: Because if the refresh occurs less often, it will be less likely to occur between our two requests (the two calls to `doPost`), which will still execute just as fast.
Anyway, that was just to be safe, but you're right, polling will be more efficient.

yrodiere:
> I'd also expect a library like this to have some out of the box support for such kind of tests?

It doesn't seem so. Fortunately, it's easy to do ourselves: we already have a utility just for that.

I removed the increase in the refresh interval in my commit and amended the commit to use polling. On my machine, executing only the `#discovery()` test takes about 5s, while executing the `#simple()` test takes about 2.5s. So we wait for approximately 2.5s.

Knowing that the sniffer cannot be configured to sniff more often than every second (which is quite healthy if you ask me), I don't think we can improve much on that.

Side note: using `mvn surefire:test` instead of `mvn test` in the commands I gave in the PR description will make it run faster, since it won't compile the tests each time.

Sanne:
merged. Thanks!

PR #1401, opened 2017-05-12
HSEARCH-2723 Clarify the documentation about built-in bridges

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2723

### DIFF
--- a/documentation/src/main/asciidoc/mapping.asciidoc
+++ b/documentation/src/main/asciidoc/mapping.asciidoc
@@ -1427,7 +1427,7 @@ Are converted into their string representation and indexed.
 `char`, `Character`::
 Are converted into their string representation and indexed.
 
-`short`, `Short`, `integer`, `Integer`, `long`, `Long`, `float`, `Float`, `double`, `Double`::
+`integer`, `Integer`, `long`, `Long`, `float`, `Float`, `double`, `Double`::
 Are per default indexed numerically using a link:http://en.wikipedia.org/wiki/Trie[Trie structure] which
 allows for efficient range queries and sorting, resulting in better query response times.
 You need to use a `NumericRangeQuery` to search for values. See also <<field-annotation>> and
--- a/documentation/src/main/asciidoc/mapping.asciidoc
+++ b/documentation/src/main/asciidoc/mapping.asciidoc
@@ -1455,21 +1455,6 @@ internal format. It is important, however, to query a numerically indexed date v
 Usually, storing the date up to the millisecond is not necessary.
 `@DateBridge` and `@CalendarBridge` allow to define the appropriate resolution you are willing to store in the index.
 +
-
-`java.time.Year`:: converts the year to the integer representation.
-`java.time.Duration`:: converts the duration to the total length in nanoseconds.
-`java.time.Instant`:: converts the instant to the number of milliseconds from Epoch.
-Note that these values are indexed with a precision to the millisecond.
-
-[IMPORTANT]
-====
-Note that it must be possible to convert the `Instant` or the `Duration` to a `Long`.
-If these values are too big or too small an exception is thrown.
-====
-
-`LocalDate`, `LocalTime`, `LocalDateTime`, `LocalTime`, `MonthDay`, `OffsetDateTime`, `OffsetTime`, `Period`, `YearMonth`, `ZoneDateTime`, `ZoneId`, `ZoneOffset`:: the bridges
-for these classes in the `java.time` package store the values as string padded with 0 when required to allow sorting.
-
 ====
 [source, JAVA]
 ----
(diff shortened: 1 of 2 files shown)


documentation/src/main/asciidoc/mapping.asciidoc
```diff
@@ -1428,41 +1419,48 @@ full text representation.
...
+`short`, `Short`, `integer`, `Integer`, `long`, `Long`, `float`, `Float`, `double`, `Double`::
+Are per default indexed numerically using a link:http://en.wikipedia.org/wiki/Trie[Trie structure] which
+allows for efficient range queries and sorting, resulting in better query response times.
+You need to use a `NumericRangeQuery` to search for values. See also <<field-annotation>> and
+<<numeric-field-annotation>>.
+
+[CAUTION]
+====
+Prior to Search 5, numeric field encoding was only chosen if explicitly requested via `@NumericField`.
+As of Search 5 this encoding is automatically chosen for the numeric types mentioned above.
```
gsmet: It was preexisting, but I think we should use `Hibernate Search` not `Search`.

Maybe a comma after `As of (Hibernate) Search 5` would be in order.

documentation/src/main/asciidoc/mapping.asciidoc
```diff
@@ -1428,41 +1419,48 @@ full text representation.
...
+
+To avoid numeric encoding you can explicitly specify a non numeric field bridge via `@Field.bridge` or `@FieldBridge`.
+The package `org.hibernate.search.bridge.builtin` contains a set of bridges which encode numbers as strings,
+for example `org.hibernate.search.bridge.builtin.IntegerBridge`.
+====
 
+`byte`, `Byte`, `short`, `Short`::
```
gsmet: Looks like `short` and `Short` were already mentioned above in the numeric fields?

documentation/src/main/asciidoc/mapping.asciidoc
```diff
@@ -1428,41 +1419,48 @@ full text representation.
...
 with a potential loss in precision can be seen in <<example-custom-numericfieldbridge>>.
 
 `java.util.Date`, `java.util.Calendar`:: Dates are indexed as `long` value representing the number
 of milliseconds since _January 1, 1970, 00:00:00 GMT_. You shouldn't really bother with the
 internal format. It is important, however, to query a numerically indexed date via a `NumericRangeQuery`.
 +
-Usually, storing the date up to the millisecond is not necessary. `@DateBridge` defines the
-appropriate resolution you are willing to store in the index.
+Usually, storing the date up to the millisecond is not necessary.
+`@DateBridge` and `@CalendarBridge` define the appropriate resolution you are willing to store in the index.
```
gsmet: Maybe `allows to define`?

documentation/src/main/asciidoc/mapping.asciidoc
```diff
@@ -1475,22 +1473,36 @@ public class Meeting {
...
-
++
 [IMPORTANT]
 ====
 A Date whose resolution is lower than `MILLISECOND` cannot be a `@DocumentId`
 ====
-
++
+You can also choose to encode the date as string using the `encoding=EncodingType.STRING` of
+`DateBridge`. In this case the dates are stored in the format _yyyyMMddHHmmssSSS_ (using GMT time).
```
gsmet: `**@**DateBridge` to be consistent with above. And maybe mention `@CalendarBridge` too.

documentation/src/main/asciidoc/mapping.asciidoc
```diff
@@ -1475,22 +1473,36 @@ public class Meeting {
...
 ----
 ====
-
-You can also choose to encode the date as string using the `encoding=EncodingType.STRING` of
-`DateBridge`. In this case the dates are stored in the format _yyyyMMddHHmmssSSS_ (using GMT time).
-
++
 [IMPORTANT]
 ====
 A Date whose resolution is lower than `MILLISECOND` cannot be a `@DocumentId`
```
gsmet: This looks pretty arbitrary and weird. Do we really have a safeguard for this?

documentation/src/main/asciidoc/mapping.asciidoc
yrodiere: Done.

documentation/src/main/asciidoc/mapping.asciidoc
yrodiere: You're right: it was a pre-existing mistake. I'll remove those types from the list above.

documentation/src/main/asciidoc/mapping.asciidoc
yrodiere: Done.

documentation/src/main/asciidoc/mapping.asciidoc
yrodiere: Done.

documentation/src/main/asciidoc/mapping.asciidoc
yrodiere: I can see why we would enforce this: truncating IDs in any way is bad idea.
But no, this doesn't seem to be enforced anywhere: the only places we call `org.hibernate.search.annotations.DateBridge.resolution()` from are the date/calendar bridge providers, and we don't  check anything about the ID.

Maybe I should simply remove this warning... ? After all, you can also shoot yourself in the foot by assigning any custom field bridge to the document ID.

gsmet: Yeah, if we don't have specific code to check for this, let's remove this.

gsmet:
Rebased and merged. With one additional commit s/integer/int/ I missed on the first review.

PR #1405, opened 2017-05-15
Warmup: Two trivial ones

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2716
 - https://hibernate.atlassian.net/browse/HSEARCH-2727

### DIFF
--- a/elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/ElasticsearchSchemaNamingErrorsIT.java
+++ b/elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/ElasticsearchSchemaNamingErrorsIT.java
@@ -16,6 +16,7 @@ import javax.persistence.Embedded;
 import javax.persistence.Entity;
 import javax.persistence.GeneratedValue;
 import javax.persistence.Id;
+import javax.persistence.Table;
 
 import org.apache.lucene.document.Document;
 import org.hibernate.AssertionFailure;
@@ -267,6 +268,7 @@ public class ElasticsearchSchemaNamingErrorsIT extends SearchInitializationTestB
 	 */
 	@Entity
 	@Indexed
+	@Table(name = "ConcreteOnCompositeEntity")
 	static class ConcreteOnCompositeEntity {
 		@DocumentId
 		@Id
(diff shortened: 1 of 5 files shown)


elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/ElasticsearchSchemaNamingErrorsIT.java
```diff
@@ -267,6 +268,7 @@ public void configureFieldMetadata(String name, FieldMetadataBuilder builder) {
 	 */
 	@Entity
 	@Indexed
+	@Table(name = "ConcreteOnCompositeEntity")
```
yrodiere: This is not strictly necessary, since the resulting name for this entity will be under 64 characters anyway. But I guess it won't hurt :)
On a side note, we have quite a few other nested classes in other tests, and they seem to be doing just fine without explicit table names. We've probably been lucky not to reach the 64 character limit until now.

integrationtest/performance/src/test/java/org/hibernate/search/test/performance/scenario/TestScenario.java
```diff
@@ -90,7 +90,6 @@ public Properties getHibernateProperties() {
 		Properties properties = new Properties();
 		properties.setProperty( "hibernate.hbm2ddl.auto", "create" );
 		properties.setProperty( "hibernate.search.default.lucene_version", "LUCENE_CURRENT" );
-		properties.setProperty( "hibernate.search.index_uninverting_allowed", "false" );
```
yrodiere: We have a similar setup `org.hibernate.search.test.query.sorting.SortOnFieldsFromCustomBridgeTest.configure(Map<String, Object>)`. Did you leave it on purpose?

Sanne: I'm not finding that. Are you sure?

Sanne: Looks like that's the one I had already modified in the same commit.

yrodiere: Right. Sorry :x

PR #1406, opened 2017-05-16
HSEARCH-1316 Support @Inject in ClassBridges and FieldBridge instances

Sanne:
This is building on @yrodiere 's Spring and CDI work pr #1351

I had some concerns related to the mutability and the "delayed initialization" introduced in some core classes but rather than commenting on the original PR I think it's simpler if I write the follow-ups I had in mind directly.

https://hibernate.atlassian.net/browse/HSEARCH-1316


orm/src/main/java/org/hibernate/search/hcore/impl/SearchFactoryReference.java
```diff
@@ -22,16 +20,16 @@
...
+	public void initialize(ExtendedSearchIntegrator extendedIntegrator) {
+		this.extendedIntegrator = extendedIntegrator;
 	}
 
 	public ExtendedSearchIntegrator getSearchIntegrator() {
-		final CompletableFuture<ExtendedSearchIntegrator> value = this.extendedIntegratorFuture;
+		ExtendedSearchIntegrator value = this.extendedIntegrator;
 		if ( value != null ) {
-			return value.join();
+			return value;
```
yrodiere: If I understand correctly, this does not just "avoid redundant delegation".

This has two side effects. We'll take CDI as an example:

1. Before this change, if someone tried to use Hibernate Search during CDI beans initialization, the call would block because Hibernate Search wasn't initialized yet, which would result in a deadlock since Hibernate Search waits for CDI to be initialized. After this change, such calls will simply fail. So in that respect, I guess this change is a good thing.
2. Before this change, if someone tried to use Hibernate Search *just after* CDI beans initialization, before Hibernate Search had the time to fully initialize, the call would block, and resume once Hibernate Search is fully initialized. After this change, such calls will simply fail. And **that** is not a good thing...

So in the end, I'd rather add a (configurable) timeout to the `join()`, so we can get the benefits of your patch (1) without its downsides (2).

Sanne: Sounds like you're assume that the two frameworks (Hibernate, CDI) are booting in two different threads running fully in parallel, and in parallel with the end user's app. Is that the case?

With the added synchronizations I've been addressing _visibility_ concerns, in case the two frameworks are initialized _by_ different threads - potentially something some containers will do - but I don't expect the application to be considered "deployed" until all frameworks have been initialized.

Let me try explain that in other words: I've addressed cross-thread visibility concerns, not timing. The "happens-before" rules should apply and I think we'll need to have faith in the containers?

yrodiere: > we'll need to have faith in the containers?
Having experienced the thrilling task of adding a Wildfly/CDI integration test, I have mixed feelings about this statement :grin:

> I don't expect the application to be considered "deployed" until all frameworks have been initialized
If that's true, we shouldn't bother with blocking calls anywhere, not even in `InitializingIntegratorState` for instance... should we?

Also, you seem to assume that the container itself will take care of integration. As we know, this will probably take years, if it ever happens. So I suspect users will have to take care of that for a time, and they may not have access to as much internals as Wildfly developpers, preventing them to control the execution order as delicately.

Anyway... I must admit I don't clearly see why this commit is useful, I thought it was simply code cleanup. If you think it's necessary, then let's go with it. We'll wait for a ticket confirming there is a problem before we try to fix this problem :)

Sanne: yes I think that's a reasonable tradeoff.

It's not my typical style to push for integrating a problem and "let the users report issues" but in this case I think I'm quite confident on taking the risk, as I believe the timing issue is really not a problem in practice as this won't be run in parallel.

The reasoning is that the variable will be read very often, and beyond a volatile read it's all getting expensive. We'll revisit this (of course) if someone managed to proof I'm wrong on the parallel timings..

Sanne: The `InitializingIntegratorState` is different as it is more likely triggered by our own integration points relying on ORM's listeners. This specific case however is meant to serve user threads, which should really not be free to run before we're initialized. In case that happens, I think having the error message is clearer than a deadlock to defend against "dodgy integration patterns"

PR #1407, opened 2017-05-17
HSEARCH-2733 FullTextIndexEventListener doesn't need to be Serializable

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2733


yrodiere:
Merged, thanks!

PR #1416, opened 2017-05-23
Fixing a couple of problems in the Sandbox / Performance tests

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2717
 - https://hibernate.atlassian.net/browse/HSEARCH-2740

### DIFF
--- a/integrationtest/sandbox/pom.xml
+++ b/integrationtest/sandbox/pom.xml
@@ -38,12 +38,6 @@
             <scope>test</scope>
         </dependency>
 
-        <!-- Avoid warnings during build -->
-        <dependency>
-            <groupId>org.jboss.logging</groupId>
-            <artifactId>jboss-logging-annotations</artifactId>
-        </dependency>
-
         <!-- Test -->
         <dependency>
             <groupId>junit</groupId>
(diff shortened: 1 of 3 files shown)


yrodiere:
Nice: it fixed the problem on my laptop.

On the error handling side, though, I couldn't help but notice that this particular test pattern appears at least three times in our tests:

 * `org.hibernate.search.test.engine.worker.WorkerTestCase`
 * `org.hibernate.search.test.performance.optimizer.OptimizerPerformanceTest`
 * `org.hibernate.search.test.performance.reader.ReaderPerformanceTestCase`

I think the last one won't behave well in case of failure, just as `OptimizerPerformanceTest` before your commits.

Perhaps we should fix `ReaderPerformanceTestCase` too? Or maybe we could use `ConcurrentRunner` in all of those tests. It should behave well when errors occur, plus IIRC we spent some time on it in order to make sure it produces usable error messages (with suppressed exceptions and such).

If you disagree, I'll just merge it.

Sanne:
Thanks for merging. To note our resolution regarding the other problems: yes they should be fixed as well, or maybe even evaluated for removal.

But not related with this specific issue and for now we need to focus on the 5.8 tasks, we'll cross that bridge when we'll have a problem.

PR #1418, opened 2017-05-24
HSEARCH-2744 Improve the Simple Query String documentation

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-2744

### DIFF
--- a/documentation/src/main/asciidoc/query.asciidoc
+++ b/documentation/src/main/asciidoc/query.asciidoc
@@ -289,10 +289,39 @@ Query luceneQuery = mythQB.keyword()
 In the previous example, only field name is boosted to 5.
 
 
-===== Simple query string queries
+===== Simple Query String queries
 
-Simple query string queries use the Lucene SimpleQueryParser. The query text is parsed and it exposes more
-search features to the end user such as boolean operators and phrase queries.
+Simple Query String queries use the Lucene `SimpleQueryParser`, presented as a "parser for
+human-entered queries". The point of this parser is to be a very simple lenient state machine
+to parse queries entered by your end users.
+
+The parser is capable to transform `keyword "some phrase" -keywordidontwant fuzzy~ prefix*`
+into a Lucene query, giving your users a little more power (phrase queries, fuzzy queries,
+boolean operators...).
+
+The lenient part is important as it will try to build the best possible query without throwing
+a parsing exception, even if the query is not what we would consider syntactically correct.
+
+Another nice feature is that it allows to search on multiple fields. You basically end up
+establishing the following contract with Lucene:
+
+ * users will enter a search query (more or less syntactically correct)
+ * it will search on the fields you have specified (and you can also specify a specific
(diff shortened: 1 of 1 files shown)


documentation/src/main/asciidoc/query.asciidoc
```diff
@@ -289,25 +289,54 @@ Query luceneQuery = mythQB.keyword()
...
     .simpleQueryString()
     .onField("history")
-    .matching("(storm OR hurricane) AND \"falling branches\"")
+    .matching("storm")
     .createQuery();
 ----
 ====
 
-By default, a simple query string query uses the OR operator if the operator is not explicitly defined.
+By default, a Simple Query String query uses the OR operator if the operator is not explicitly defined.
```
Sanne: +1 let's keep it consistently capitalized to clarify that it's the proper name of the feature.

documentation/src/main/asciidoc/query.asciidoc
```diff
@@ -339,12 +368,12 @@ Query luceneQuery = mythQB
 ----
 ====
 
-The above query returns results containing "storm" and "tree" either in "name", "history" or "description".
+The above query returns results containing `storm` and `tree` either in `name`, `history` or `description`.
 
-The syntax supported by the SimpleQueryParser includes the following operators: boolean (AND, OR, NOT),
+The syntax supported by the `SimpleQueryParser` includes the following operators: boolean (AND `+`, OR `|`, NOT `-`),
 prefix, phrase, precedence, fuzzy and near operator for phrase queries.
```
Sanne: could you list also how to do "prefix, phrase, precedence, fuzzy " ?

documentation/src/main/asciidoc/query.asciidoc
```diff
@@ -339,12 +368,12 @@ Query luceneQuery = mythQB
 ----
 ====
 
-The above query returns results containing "storm" and "tree" either in "name", "history" or "description".
+The above query returns results containing `storm` and `tree` either in `name`, `history` or `description`.
 
-The syntax supported by the SimpleQueryParser includes the following operators: boolean (AND, OR, NOT),
+The syntax supported by the `SimpleQueryParser` includes the following operators: boolean (AND `+`, OR `|`, NOT `-`),
```
yrodiere: It's definitely better, I'm a bit worried that users will understand that `AND`, `OR`, etc. can be used as operators in queries, while in fact they cannot.
Maybe something like "(AND using `+`, OR using `|`, NOT using `-`)" would avoid any risk of confusion?

gsmet: I was just changing it following the above remark of Sanne (and the use of a list to present each option). :)

gsmet:
@Sanne @yrodiere force pushed an update addressing your comments on the operator list. It should be more clear now.

yrodiere:
Nice! Since @Sanne's remarks have been addressed too, I merged it.

PR #1420, opened 2017-05-24
HSEARCH-2735 Deprecate index-time boosting capabilities

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2735

### DIFF
--- a/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
+++ b/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
@@ -299,9 +299,13 @@ the provided string must not include the zone offset (`+01:00`) or the zone ID (
 Even when they make sense for the type of your field, the time and time zone may be omitted
  (if omitted, the time zone will be interpreted as the default JVM time zone).
 
-===== Dynamic boosting
+===== Index-time boosting
 
-The `org.hibernate.search.annotations.DynamicBoost` annotation is not (and cannot be) supported with Elasticsearch, because the platform lacks per-document, index-time boosting capabilities. Static boosts (`@Boost`) are, however, supported.
+The `org.hibernate.search.annotations.DynamicBoost` annotation is not (and cannot be)
+supported with Elasticsearch, because the platform lacks per-document, index-time boosting capabilities.
+
+The `@Boost` annotation will work, but since it's deprecated
+we suggest to use query-time boosting instead: see <<search-query>>>.
 
 ==== [[elasticsearch-mapping-analyzer]] Analyzers
 
--- a/documentation/src/main/asciidoc/mapping.asciidoc
+++ b/documentation/src/main/asciidoc/mapping.asciidoc
@@ -102,9 +102,9 @@ field.
 Fields used for sorting or faceting _must not_ be analyzed.
 ====
 
-* [[field-annotation-norms]]`norms`: describes whether index time boosting information should be stored (`Norms.YES`) or
+* [[field-annotation-norms]]`norms`: describes whether index time scoring information should be stored (`Norms.YES`) or
 not (`Norms.NO`). Not storing the norms can save a considerable amount of memory, but index time
-boosting will not be available in this case. The default value is `Norms.YES`.
+boosting (in particular) will not be available in this case. The default value is `Norms.YES`.
 
 * `termVector`: describes collections of term-frequency pairs. This attribute enables the storing
 of the term vectors within the documents during indexing. The default value is `TermVector.NO`.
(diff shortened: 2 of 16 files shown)


documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -299,9 +299,13 @@ the provided string must not include the zone offset (`+01:00`) or the zone ID (
...
  (if omitted, the time zone will be interpreted as the default JVM time zone).
 
-===== Dynamic boosting
+===== Index-time boosting
 
-The `org.hibernate.search.annotations.DynamicBoost` annotation is not (and cannot be) supported with Elasticsearch, because the platform lacks per-document, index-time boosting capabilities. Static boosts (`@Boost`) are, however, supported.
+The `org.hibernate.search.annotations.DynamicBoost` annotation is not (and cannot be)
+supported with Elasticsearch, because the platform lacks per-document, index-time boosting capabilities.
+
+Static boosts (`@Boost`) are supported, but as in Lucene embedded mode,
```
Sanne: I don't think it's clear what "static boosts" mean. What about we simply remove  this paragraph? I think we should stop highlighting what actually works in Elasticsearch, and shift to a mode in which the user is reassured that pretty much everything will work the same - with excetions. We should highlight the exceptions.

Or maybe just a simple "The `@Boost` annotation will work, but since it's deprecated we suggest to use query time boosting".

documentation/src/main/asciidoc/mapping.asciidoc
```diff
@@ -866,6 +866,13 @@ following sections show you how you can achieve index time boosting using Hibern
 [[section-boost-annotation]]
 ==== Static index time boosting
 
+[WARNING]
+====
+Index-time boosting is deprecated: it will be removed starting from Lucene 7.
```
Sanne: s/it will be removed starting from Lucene 7/support for this approach will be removed in Apache Lucene since version 7.

documentation/src/main/asciidoc/mapping.asciidoc
```diff
@@ -866,6 +866,13 @@ following sections show you how you can achieve index time boosting using Hibern
...
+
+To implement static boosting, you should use query-time boosting instead: see <<search-query>>>.
```
Sanne: "static" ? Why not simply say "We recommend to use query-time boosting instead : .." ?

documentation/src/main/asciidoc/mapping.asciidoc
```diff
@@ -909,6 +916,16 @@ Lucene In Action from Otis Gospodnetic and Erik Hatcher.
 [[section-dynamic-boost]]
 ==== Dynamic index time boosting
 
+[WARNING]
+====
+Index-time boosting is deprecated: it will be removed starting from Lucene 7.
+
+To implement static boosting, you can use query-time boosting instead: see <<search-query>>.
```
Sanne: static? Same as above.

documentation/src/main/asciidoc/mapping.asciidoc
```diff
@@ -2935,6 +2952,16 @@ public class Departments {
 
 ==== Mapping dynamic boost
 
+[WARNING]
+====
+Index-time boosting is deprecated: it will be removed starting from Lucene 7.
+
+To implement static boosting, you can use query-time boosting instead: see <<search-query>>.
```
Sanne: static? Same as above.

documentation/src/main/asciidoc/mapping.asciidoc
yrodiere: Done.

documentation/src/main/asciidoc/mapping.asciidoc
yrodiere: As a matter of fact, query-time boosting through our DSL is "static" (at least if we give "static" the same meaning as in the "Static index time boosting" section above): a given field will have the same boost for every document, regardless of the content of each document.

Anyway... I rephrased this so that I don't mention "static" at all.

documentation/src/main/asciidoc/mapping.asciidoc
yrodiere: Will replace with the same warning as in the "Dynamic index time boosting" section.

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
yrodiere: > I don't think it's clear what "static boosts" mean

It's the exact same wording as in the "mapping" section.

>  think we should stop highlighting what actually works in Elasticsearch

In general I'd say yes, but here the point was that `@Boost` and `@DynamicBoost` are closely related, so I didn't want one to be confused with the other.

> Or maybe just a simple "The `@Boost` annotation will work, but since it's deprecated we suggest to use query time boosting".

Done.

documentation/src/main/asciidoc/mapping.asciidoc
yrodiere: Done.

Sanne:
merged! thanks

PR #1421, opened 2017-05-24
HSEARCH-2746 Translate parameters of WordDelimiterFilterFactory from 0/1 to booleans on Elasticsearch

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2746


elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/ElasticsearchAnalyzerDefinitionCreationIT.java
```diff
@@ -164,25 +164,25 @@ public void success_simple() throws Exception {
...
 									+ "'type': 'word_delimiter',"
-									+ "'generate_word_parts': '1',"
-									+ "'generate_number_parts': '1',"
-									+ "'catenate_words': '0',"
-									+ "'catenate_numbers': '0',"
-									+ "'catenate_all': '0',"
-									+ "'split_on_case_change': '0',"
-									+ "'split_on_numerics': '0',"
-									+ "'preserve_original': '1'"
+									+ "'generate_word_parts': 'true',"
```
gsmet: Shouldn't it be a boolean and not a string?

yrodiere: Yes it should, but it's out of my control. Elasticsearch just sends back a string, even though I originally sent a boolean.

Sanne:
merged!

From a performance perspective I'm not particularly fond of us creating a Map of parameters to then later convert them all and create a new Map of parameters.

But that's having sub-zero priority right now so let's not spend time on it, we'll revisit if we see it's actually a problem during perf tests.

Thanks!

yrodiere:
>merged!

Thanks!

> But that's having sub-zero priority right now so let's not spend time on it, we'll revisit if we see it's actually a problem during perf tests.

It's especially low priority given that we're talking about bootstrapping code executed at most once per analyzer definition... Granted, bootstrapping code shouldn't be inefficient, but copying a dozen maps is a drop in the ocean.

Sanne:
> It's especially low priority given that we're talking about bootstrapping code executed at most once per analyzer definition... Granted, bootstrapping code shouldn't be inefficient, but copying a dozen maps is a drop in the ocean.

Ooops, ok I missed it was in the context of bootstrap. Good thing I merged then ;)

PR #1422, opened 2017-05-26
HSEARCH-2742 NPE when clearing the index in Elasticsearch

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2742

I took this opportunity to make the JSON extraction code a bit safer.


Sanne:
merged! thanks

PR #1424, opened 2017-05-30
HSEARCH-2603 Switch to the Painless language for Elasticsearch inline scripts

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2603


Sanne:
great, merged this.

PR #1425, opened 2017-05-30
HSEARCH-2751 SimpleQueryStringDSLTest fails on Elasticsearch 5.1

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2751


gsmet:
@yrodiere looks good to me. I already had to tune it before that when we upgraded Es.

yrodiere:
@gsmet Merged, thanks!

PR #1423, opened 2017-05-30
HSEARCH-2749 Make sure every SearchIntegrator built in test is closed properly

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2749

PR #1426, opened 2017-05-30
HSEARCH-2714 Allow finer-grained configuration of Elasticsearch authentication

yrodiere:
~~This PR is based on #1424, #1423 and #1425, which we'll have to merge first.~~ => Done

Relevant JIRA ticket: https://hibernate.atlassian.net//browse/HSEARCH-2714

So, in this PR:

 * The first commit ("HSEARCH-2714 Allow finer-grained configuration of Elasticsearch authe…") introduces a way for users to finely tune the HTTP client configuration when using Elasticsearch
 * All the other commits are about testing this fine-tuning on Amazon Web Services using a new profile

The main problem with AWS was the lack of quality libraries for request signing:

 * The AWS SDK only works if you use its own HTTP client, which we can't (the Elasticsearch client uses its own instance of an Apache HTTP client)
 * The most popular community project, https://github.com/inreachventures/aws-signing-request-interceptor, has a bug with POST requests whose content-length is 0 (it removes the content-length from the hashed string for some reason, and thus computes an invalid hash). The bug cannot be worked around.
 * The project I used, https://github.com/lucasweb78/aws-v4-signer-java, also has quite a few bugs, but I was able to find a workaround for each bug.

The new profile allowing to test on AWS is named "aws". Be warned: it only works on the elasticsearch project, I didn't try to make it work on Wildfly for instance, mainly because it would requires additional classes in the Wildfly module, and, well, it didn't seem worthwhile.

I managed to get our CI job to pass: http://ci.hibernate.org/view/Search/job/hibernate-search-master-elasticsearchAWS/ .
It only builds modules up to the elasticsearch module, and doesn't run any other integration tests, but it passes.
For now it targets my dev branch on my own Hibernate Search repo, so we should update it once this PR has been merged.


yrodiere:
Actually there's an issue with URL encoding I discovered when trying to create a patch for https://github.com/lucasweb78/aws-v4-signer-java. I'll try to solve that issue first, since it may make this PR a bit simpler.

yrodiere:
... and I got excited about nothing, once again. See https://hibernate.atlassian.net/browse/HSEARCH-2755. We can safely review this PR.

Sanne:
# API Encapsulation

I think we have an encapsulation challenge when exposing `org.hibernate.search.elasticsearch.client.ElasticsearchHttpClientConfigurer` as an API.

We have not tested this within WildFly modules but I'm confident it would fail as the client code would be unable to load the types from the HTTP client module.
Besides, we're trying to remove strong coupling to other libraries from the API, a step necessary to guarantee longer term API compatibility.

In particular both of these seem essential to implement a meaningful `ElasticsearchHttpClientConfigurer`, but many more are likely to be needed in practice:
 - `org.apache.http.impl.nio.client.HttpAsyncClientBuilder`
 - `org.apache.http.HttpRequestInterceptor`

## Wrap it all (bad idea)
One option would be to not expose these types but some surrogate interfaces which we adapt; I'd rather avoid this for various practical reasons.

## Explicit dependency to WildFly modules

To address the WildFly modules specific issue we could just suggest that the end user will need to have his application to explicitly depend on the `org.apache.httpcomponents` module.

I personally like this solution better as it offers good flexibility to users and is practical for us to maintain.

Having the user need to specify the dependency explicitly is IMO ok for an SPI, and I like explicit dependencies better than some obscure rule.

So my proposal: let's make this an _SPI_ rather than an _API_?

# Ship the AWS integration component ?

I wasn't planning to ship this, but considering:
 - you already implemented it
 - we can mark it experimental
 - it has just one dependency, which seems reasonable code and is licensed ASL2

I think it's useful and we'd be otherwise hiding this great feature in our testsuite.. seems selfish ;)

Proposal: _as an independent, follow-up JIRA,_ we move the AWS signing component in a dedicated Maven module and include it in our WildFly - packaged modules.

As a side effect this would imply that users on AWS specifically will not need to explicitly depend on the  `org.apache.httpcomponents` module as we'll ship the correct module configuration already.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/client/ElasticsearchHttpClientConfigurer.java
```diff
@@ -0,0 +1,19 @@
...
+
+import org.apache.http.impl.nio.client.HttpAsyncClientBuilder;
+
+/**
+ * @author Yoann Rodiere
+ * @hsearch.experimental This contract is currently under active development and may be altered in future releases,
+ */
+public interface ElasticsearchHttpClientConfigurer {
+
+	HttpAsyncClientBuilder configure(HttpAsyncClientBuilder builder);
```
Sanne: It's not clear to me why the method doesn't return `void` ?

I don't think we should allow users to return a different instance.

Sanne: This API (SPI?) will need some proper javadoc.

yrodiere: > It's not clear to me why the method doesn't return void ?
> I don't think we should allow users to return a different instance.

I generally take care to always use the latest builder returned from the builder methods, but it's just to be safe: the builder could, in theory, choose to return a different instance at some point. In which case we need the users to return the latest instance.
This builder doesn't do that currently, so if you think it's unlikely it ever will, then yes, we can return void instead.  

> This API (SPI?) will need some proper javadoc.

True, sorry about that :x I will add it.

Sanne: Yes I understand what you wanted to do, but it might be confusing. I'd rather return void to make it clear what you expect people to do.

Methods which mutate parameters are not commonly used either but I think this is an example in which it's ok.

Sanne:
> Give the configurers access to (masked) configuration properties, so that we avoid passing properties through environment variables as we do now. It would in particular allow a single application to target multiple Elasticsearch clusters in the future, which is something we kinda promised in our documentation.

+1

> Allow multiple configurers to be set up at the same time, so that for instance a user can both use an AWS authentication configurer and a custom one for performance tuning or whatever.

I wouldn't do this. Sure it's cool but we can't spend too much time on this and people who really have such needs will have the skills to create an adaptor which decorates as many customizers they want.

Be careful on what you spend your time on, there's high value in other tasks and this isn't one of them.

> Allow extensions to be independent from each other. For instance if a third-party adds support for an Azure cloud or whatever, I wouldn't want them to have to re-implement some kind "configurer registry" to return they own registry (a bit like DirectoryProviderService); that would prevent multiple third-party extensions to work together.

I'd treat this like the previous point. Don't go over board, we won't know the right level of abstraction until you have a concrete request anyway.

> Ok, but I'll work on this PR to turn the feature to an SPI first.

Yes definitely. Any further improvements should not block 5.8,

yrodiere:
> I wouldn't do this. Sure it's cool but we can't spend too much time on this and people who really have such needs will have the skills to create an adaptor which decorates as many customizers they want.

Luckily it doesn't require much time :)

```
		Iterable<ElasticsearchHttpClientConfigurer> configurers =
				serviceManager.getClassLoaderService().loadJavaServices( ElasticsearchHttpClientConfigurer.class );
		for ( ElasticsearchHttpClientConfigurer configurer : configurers ) {
			configurer.configure( builder, properties );
		}
```

I'll submit a full PR soon.

yrodiere:
(Note that we already have similar code in BridgeFactory to handle IndexManagerType-specific bridge providers, so I'm not doing anything new)

yrodiere:
@Sanne Done: I added a commit and rebased on master.

yrodiere:
Also, the build passes on AWS: http://ci.hibernate.org/job/hibernate-search-master-elasticsearchAWS/28/

elasticsearch/pom.xml
```diff
@@ -559,6 +541,10 @@
                 <test.elasticsearch.client.http_client_configurer>
                     org.hibernate.search.elasticsearch.testutil.aws.AWSElasticsearchHttpClientConfigurer
                 </test.elasticsearch.client.http_client_configurer>
+                <test.elasticsearch.failsafe.excludedGroups.aws>
+                    ,
```
Sanne: is this comma intentional?

yrodiere: Yes, see how we use the variable (we concatenate it to another).

Sanne: ok thanks, didn't get so far yet :)

yrodiere:
@Sanne Rebased, build in progress: http://ci.hibernate.org/job/hibernate-search-master-elasticsearchAWS/29/

yrodiere:
... and the same, but with proper job configuration (I didn't account for the fact that we now expect an Elasticsearch 5.2+ instance in tests by default, while our AWS instance is 5.1): http://ci.hibernate.org/job/hibernate-search-master-elasticsearchAWS/30/

yrodiere:
Sorry about the mess; we're good now: http://ci.hibernate.org/job/hibernate-search-master-elasticsearchAWS/33/

Sanne:
merged!

PR #1427, opened 2017-05-30
HSEARCH-2752

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2752

### DIFF
--- a/engine/src/main/java/org/hibernate/search/query/engine/impl/LuceneHSQuery.java
+++ b/engine/src/main/java/org/hibernate/search/query/engine/impl/LuceneHSQuery.java
@@ -137,18 +137,16 @@ public class LuceneHSQuery extends AbstractHSQuery implements HSQuery {
 		}
 		try {
 			QueryHits queryHits = getQueryHits( searcher, calculateTopDocsRetrievalSize() );
-			int first = firstResult;
-			int max = max( first, queryHits.getTotalHits() );
-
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/hibernate/search/query/engine/impl/LuceneHSQuery.java
```diff
@@ -147,7 +147,8 @@ public Query getLuceneQuery() {
 			DocumentExtractor extractor = buildDocumentExtractor( searcher, queryHits, first, max );
 			for ( int index = first; index <= max; index++ ) {
 				infos.add( extractor.extract( index ) );
-				if ( index % 10 == 0 ) {
+				//Check for timeout each 16 elements:
+				if ( (index & 0x000F) == 0 ) {
```
yrodiere: `% 16` would be clearer... Wouldn't it be optimized anyway?

Sanne: same question as I had myself a couple of years ago :) no it seems JIT doesn't do that specific kind of optimisations. "%" is one of the operations we should always avoid according to our friends on OpenJDK.

yrodiere: Weird... but good to know :)

PR #1429, opened 2017-05-31
HSEARCH-2481 Byteman-based tests executed in the Elasticsearch module won't work

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2481

Note that only the last two commits (in topological order, so "HSEARCH-2481 Do not allow the maven-failsafe-plugin to scan the whole ..." and "HSEARCH-2481 Enable Byteman-based tests with Elasticsearch") are strictly necessary.
The other commits were my first attempts at solving the issue, but since they also clean up our poms, we may as well keep them.

### DIFF
--- a/backends/jgroups/pom.xml
+++ b/backends/jgroups/pom.xml
@@ -179,10 +179,6 @@
                             <goal>integration-test</goal>
                             <goal>verify</goal>
                         </goals>
-                        <configuration>
-                            <argLine>${additionalRuntimeArgLine}</argLine>
-                            <redirectTestOutputToFile>true</redirectTestOutputToFile>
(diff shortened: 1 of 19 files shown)


gsmet:
@yrodiere looks good to me. Did you check that you execute the exact same number of tests with the new inclusions?

yrodiere:
> Did you check that you execute the exact same number of tests with the new inclusions?

I didn't, because we had the exact same naming constraints in the -engine and -orm module, but you're right, there was one less test executed. Turns out it wasn't executed in the -engine module either, because the name was incorrect. I fixed that.

> If so, it's good to merge!

Merging, thanks!

PR #1430, opened 2017-06-01
HSEARCH-2748 Just log a warning instead of throwing an exception when connecting to an ES server with a higher version than supported

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2748


Sanne:
merged!

PR #1431, opened 2017-06-01
HSEARCH-2741 Fix dependency convergence issue between MariaDB and integration tests

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2741

A build with this patch applied can be seen here: http://ci.hibernate.org/job/hibernate-search-yoann/1/

The build still has failing tests, but it's unrelated (seems to be a charset issue). I'm working on it and will submit another PR.


Sanne:
Nice. I'll merge this from the web ui :)

PR #1432, opened 2017-06-02
HSEARCH-2652 Create one error context per set of works, instead of one per work, when processing Elasticsearch works

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2652

Note: the purpose of this PR is to solves issues originally spotted there: https://github.com/hibernate/hibernate-search/pull/1336#pullrequestreview-29142895


engine/src/main/java/org/hibernate/search/backend/impl/lucene/MultiWriteDrainableLinkedList.java
```diff
@@ -37,17 +37,39 @@
...
 	}
 
-	private synchronized void addNode(Node<T> newnode) {
+	/**
+	 * Adds multiple new entries to this list, in the given order,
+	 * making sure no entry is added between the given elements.
+	 *
+	 * @param elements The elements to add.
+	 */
+	public void addAll(Iterable<T> elements) {
```
Sanne: This looks like very tricky, and there's a simpler approach ;)

See `org.hibernate.search.backend.impl.lucene.Changeset`

yrodiere: I was trying to avoid instantiating an object for each set of works (which could be one object _per work_ in the case of `org.hibernate.search.indexes.spi.IndexManager.performStreamOperation(LuceneWork, IndexingMonitor, boolean)`), but in retrospect it seems a bit stupid given that this forces us to instantiate dozens of nodes in the non-stream case... I'll change that.

On a related note, the `monitor` object in `Changeset` is never used. There may be something wrong there... ? I'll remove it.

Sanne: yes the `monitor` looks odd. I wouldn't remove it though, I suspect it's needed for Infinispan - and seems handy for us as well.

Feel free to create a new, similar class for Elasticsearch if you feel it doesn't fit as-is.

Sanne: > this forces us to instantiate dozens of nodes in the non-stream case.

Yes that's what I was thinking too.

yrodiere:
@Sanne I applied the suggested changes.

Sanne:
merged

PR #1433, opened 2017-06-02
HSEARCH-2683 Avoid explicit usage of hibernate.search. prefix for new Elasticsearch configuration properties

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2683


yrodiere:
There seems to be a problem even though the test passed... Investigating.

Sanne:
merged!

PR #1436, opened 2017-06-06
HSEARCH-2754 SpatialFieldBridge uses inefficient Reflection access

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2754

We probably could do more ([HSEARCH-2759](https://hibernate.atlassian.net/browse/HSEARCH-2759) in particular), but unfortunately the `SpatialFieldBridge` class and its subclasses are API.
I chose to only fix the performance issue for now, so as not to break user code.

### DIFF
--- a/engine/src/test/java/org/hibernate/search/test/metadata/Snafu.java
+++ b/engine/src/test/java/org/hibernate/search/test/metadata/Snafu.java
@@ -65,4 +65,15 @@ public class Snafu {
 	@Field
 	@org.hibernate.search.annotations.Analyzer(impl = FooAnalyzer.class)
 	private String custom;
+
+
+	// Just so that @Spatial works; see HSEARCH-2759
+	public double getLatitude() {
+		return latitude;
+	}
+
+	// Just so that @Spatial works; see HSEARCH-2759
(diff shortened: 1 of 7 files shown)


engine/src/test/java/org/hibernate/search/test/metadata/Snafu.java
```diff
@@ -65,4 +65,15 @@
 	@Field
 	@org.hibernate.search.annotations.Analyzer(impl = FooAnalyzer.class)
 	private String custom;
+
+
+	// Just so that @Spatial works; see HSEARCH-2759
+	public double getLatitude() {
```
Sanne: It looks like this workaround wasn't needed before. Do we have a regression?

yrodiere: Not really; the tests using this class do not attempt to index an instance, so previously the reflection code was never executed.

Now that we execute the reflection on startup, the bug (HSEARCH-2759) is more visible. It's always been here, though, it's just that we didn't really use the `@Spatial` feature (no indexing).

Sanne:
merged!

PR #1438, opened 2017-06-06
HSEARCH-2734 Re-brand the "ram" DirectoryProvider as "local-heap"

yrodiere:
This PR is based on #1445 , which should be merged first.

https://hibernate.atlassian.net//browse/HSEARCH-2734


yrodiere:
Maybe we should also continue to support the "ram" name, but log a warning when it's used? Otherwise it looks a lot like a breaking change for users...

Sanne:
> Maybe we should also continue to support the "ram" name, but log a warning when it's used? Otherwise it looks a lot like a breaking change for users...

Yes, definitely!

Sanne:
merged, thanks!

PR #1439, opened 2017-06-06
HSEARCH-2696 Move integration test environment resources to a more fitting location

yrodiere:
This PR is based on #1445 , which should be merged first.

https://hibernate.atlassian.net//browse/HSEARCH-2696


yrodiere:
I'm working on JVM options for Elasticsearch right now; I may have to move the configuration around so as to avoid duplicating it. I'll update this PR accordingly.

PR #1442, opened 2017-06-07
Milestone 1 for "Free Form Indexing"

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-1404

The new utility and collection classes have more functionality than what is going to be needed in future; in part that's to keep many of the new-deprecated methods standing. They are not cast in stone, I expect we'll polish them later.

In particular I don't want this double-type system to become a burden on performance, so part of the design is driven to make it possible to (eventually) avoid allocating any of these objects on hot paths.

Not least, I hope to reuse some of these structures: while previously we would create Sets of classes as needed, for example to compute the polymorphic hiearchy of a specific root - of only the known indexed types - such sets can be pre-computed and reused for each known type.
To get there this includes seeds towards a type registry, incidentally if we manage to push the no-allocation to the limits this will allow reference based comparisons and to optimise for fast, uncontended lookups in read-only.

The reason I avoided Java standard collections is to evolve into immutable structures and aggressively trim down on the exposed methods so that we can understand how to tune these; to keep some of the deprecated methods "afloat" for now they are not immutable yet.

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/schema/impl/Elasticsearch2SchemaTranslator.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/schema/impl/Elasticsearch2SchemaTranslator.java
@@ -503,7 +503,7 @@ public class Elasticsearch2SchemaTranslator implements ElasticsearchSchemaTransl
 			return new JsonPrimitive( (Boolean) indexedNullToken );
 		}
 		else {
-			throw LOG.unsupportedNullTokenType( mappingBuilder.getBeanClass(), fieldPath.getAbsoluteName(),
+			throw LOG.unsupportedNullTokenType( mappingBuilder.getBeanClass().getName(), fieldPath.getAbsoluteName(),
 					indexedNullToken.getClass() );
 		}
 	}
--- a/engine/src/main/java/org/hibernate/search/engine/impl/ImmutableSearchFactory.java
+++ b/engine/src/main/java/org/hibernate/search/engine/impl/ImmutableSearchFactory.java
@@ -12,7 +12,6 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Predicate;
 
@@ -65,11 +64,17 @@ import org.hibernate.search.query.engine.spi.HSQuery;
 import org.hibernate.search.query.engine.spi.QueryDescriptor;
 import org.hibernate.search.query.engine.spi.TimeoutExceptionFactory;
 import org.hibernate.search.spi.CustomTypeMetadata;
+import org.hibernate.search.spi.IndexedTypeIdentifier;
+import org.hibernate.search.spi.IndexedTypeSet;
 import org.hibernate.search.spi.IndexingMode;
 import org.hibernate.search.spi.InstanceInitializer;
 import org.hibernate.search.spi.SearchIntegrator;
+import org.hibernate.search.spi.IndexedTypeMap;
 import org.hibernate.search.spi.WorkerBuildContext;
+import org.hibernate.search.spi.impl.ConcurrentIndexedTypeMap;
 import org.hibernate.search.spi.impl.ExtendedSearchIntegratorWithShareableState;
+import org.hibernate.search.spi.impl.IndexedTypesSets;
+import org.hibernate.search.spi.impl.PojoIndexedTypeIdentifier;
 import org.hibernate.search.spi.impl.TypeHierarchy;
 import org.hibernate.search.spi.impl.SearchFactoryState;
 import org.hibernate.search.stat.Statistics;
@@ -92,12 +97,12 @@ public class ImmutableSearchFactory implements ExtendedSearchIntegratorWithShare
 
 	private static final Log log = LoggerFactory.make();
 
-	private final Map<Class<?>, EntityIndexBinding> indexBindingForEntities;
-	private final Map<Class<?>, DocumentBuilderContainedEntity> documentBuildersContainedEntities;
+	private final IndexedTypeMap<EntityIndexBinding> indexBindingForEntities;
+	private final IndexedTypeMap<DocumentBuilderContainedEntity> documentBuildersContainedEntities;
 	/**
 	 * Lazily populated map of type descriptors
 	 */
-	private final ConcurrentHashMap<Class, IndexedTypeDescriptor> indexedTypeDescriptors;
+	private final IndexedTypeMap<IndexedTypeDescriptor> indexedTypeDescriptors;
 	private final Worker worker;
 	private final Map<String, FilterDef> filterDefinitions;
 	private final FilterCachingStrategy filterCachingStrategy;
@@ -185,7 +190,7 @@ public class ImmutableSearchFactory implements ExtendedSearchIntegratorWithShare
 		}
 
 		this.indexReaderAccessor = new DefaultIndexReaderAccessor( this );
-		this.indexedTypeDescriptors = new ConcurrentHashMap<>();
+		this.indexedTypeDescriptors = new ConcurrentIndexedTypeMap<>();
 
 		this.defaultObjectLookupMethod = determineDefaultObjectLookupMethod();
 		this.defaultDatabaseRetrievalMethod = determineDefaultDatabaseRetrievalMethod();
@@ -287,24 +292,22 @@ public class ImmutableSearchFactory implements ExtendedSearchIntegratorWithShare
 
 	@Override
 	public HSQuery createHSQuery(Query luceneQuery, Class<?>... entityTypes) {
-		QueryDescriptor descriptor = createQueryDescriptor( luceneQuery, entityTypes );
-
-		return descriptor.createHSQuery( this ).targetedEntities( Arrays.asList( entityTypes ) );
+		IndexedTypeSet newtypes = IndexedTypesSets.fromClasses( entityTypes );
+		QueryDescriptor descriptor = createQueryDescriptor( luceneQuery, newtypes );
+		return descriptor.createHSQuery( this ).targetedEntities( newtypes );
 	}
 
 	@Override
 	public HSQuery createHSQuery(Query luceneQuery, CustomTypeMetadata... types) {
 		List<CustomTypeMetadata> typeList = Arrays.asList( types );
-		Class<?>[] entityTypes = typeList.stream()
+		IndexedTypeSet entityTypes = typeList.stream()
 				.map( CustomTypeMetadata::getEntityType )
-				.toArray( Class<?>[]::new );
-
+				.collect( IndexedTypesSets.streamCollector() );
 		QueryDescriptor descriptor = createQueryDescriptor( luceneQuery, entityTypes );
-
 		return descriptor.createHSQuery( this ).targetedTypes( typeList );
 	}
 
-	private QueryDescriptor createQueryDescriptor(Query luceneQuery, Class<?>[] entityTypes) {
+	private QueryDescriptor createQueryDescriptor(Query luceneQuery, IndexedTypeSet entityTypes) {
 		QueryDescriptor descriptor = null;
 
 		if ( queryTranslatorPresent ) {
@@ -324,22 +327,34 @@ public class ImmutableSearchFactory implements ExtendedSearchIntegratorWithShare
 	}
 
 	@Override
-	public Map<Class<?>, DocumentBuilderContainedEntity> getDocumentBuildersContainedEntities() {
+	public IndexedTypeMap<DocumentBuilderContainedEntity> getDocumentBuildersContainedEntities() {
 		return documentBuildersContainedEntities;
 	}
 
 	@Override
-	public Map<Class<?>, EntityIndexBinding> getIndexBindings() {
+	public IndexedTypeMap<EntityIndexBinding> getIndexBindings() {
 		return indexBindingForEntities;
 	}
 
 	@Override
+	@Deprecated
 	public EntityIndexBinding getIndexBinding(Class<?> entityType) {
+		return indexBindingForEntities.get( new PojoIndexedTypeIdentifier( entityType ) );
+	}
+
+	@Override
+	public EntityIndexBinding getIndexBinding(IndexedTypeIdentifier entityType) {
 		return indexBindingForEntities.get( entityType );
 	}
 
 	@Override
+	@Deprecated
 	public DocumentBuilderContainedEntity getDocumentBuilderContainedEntity(Class entityType) {
+		return documentBuildersContainedEntities.get( entityType );
+	}
+
+	@Override
+	public DocumentBuilderContainedEntity getDocumentBuilderContainedEntity(IndexedTypeIdentifier entityType) {
 		return documentBuildersContainedEntities.get( entityType );
 	}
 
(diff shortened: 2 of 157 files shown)


Sanne:
failed because of IndexingActionInterceptorTest.testInterceptorWithMassIndexer

@Hibernate-CI retest please

engine/src/main/java/org/hibernate/search/indexes/spi/DirectoryBasedIndexManager.java
```diff
@@ -136,8 +136,11 @@ public void setSearchFactory(ExtendedSearchIntegrator boundSearchIntegrator) {
 	}
 
 	@Override
-	public void addContainedEntity(Class<?> entity) {
-		if ( containedEntityTypes.add( entity ) ) {
+	public void addContainedEntity(IndexedTypeIdentifier entity) {
+		final IndexedTypesSet oldSet = containedEntityTypes;
+		final IndexedTypesSet newSet = IndexedTypesSets.composite( oldSet, entity );
+		if ( ! oldSet.equals( newSet ) ) {
```
yrodiere: Why not simply `if ( !oldSet.contains( entity ) )`?

engine/src/main/java/org/hibernate/search/query/engine/impl/EntityInfoImpl.java
```diff
@@ -77,12 +83,12 @@ public Serializable getId() {
...
 			 * because in some cases we don't need the identifier at all (e.g. projections).
 			 */
-			if ( clazz == null ) {
+			if ( type == null ) {
 				//FIXME this is horrible, but a temporary refactoring as we move to IndexedTypeIdentifier
 				throw log.incompleteEntityInfo( null, null );
 			}
 			else {
-				throw log.incompleteEntityInfo( new PojoIndexedTypeIdentifier( clazz ), id );
+				throw log.incompleteEntityInfo( type, id );
```
yrodiere: I think you can remove the surrounding if, it's quite safe :]

engine/src/main/java/org/hibernate/search/spi/SearchIntegratorBuilder.java
```diff
@@ -423,10 +423,11 @@ else if ( metadataProvider.containsSearchMetadata( mappedClassIdentifier ) ) {
 
 		// Create all IndexManagers, configure and start them:
 		for ( XClass mappedXClass : rootIndexedEntities ) {
-			Class mappedClass = classMappings.get( mappedXClass );
+			final Class mappedClass = classMappings.get( mappedXClass );
```
yrodiere: I think you could go a step further and make `classMappings` a `Map<XClass, IndexedTypeIdentifier>`. Not that it matters much for now, but that's probably where we're going, so...

engine/src/main/java/org/hibernate/search/spi/impl/HashSetIndexedTypesSet.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.hibernate.search.spi.IndexedTypeIdentifier;
+import org.hibernate.search.spi.IndexedTypesSet;
+
+final class HashSetIndexedTypesSet implements IndexedTypesSet, Serializable {
```
yrodiere: Same remark as in `ConcurrentIndexedTypeMap`:

Maybe we'd better have an implementation simply accept a java.util.function.Supplier<? extends Set<IndexedTypeIdentifier>> in its main constructor? Then you'd only have to write something like new IndexedTypesSet<>(HashSet::new). We'd avoid duplicating the whole inheritance tree of java.util.Map...

engine/src/main/java/org/hibernate/search/spi/impl/HashSetIndexedTypesSet.java
```diff
@@ -0,0 +1,78 @@
...
+
+	public static final IndexedTypesSet EMPTY = new HashSetIndexedTypesSet( Collections.EMPTY_SET );
+
+	private final Set<IndexedTypeIdentifier> set;
+
+	HashSetIndexedTypesSet(Set<IndexedTypeIdentifier> set) {
```
yrodiere: If you don't want to change the implementation (see my comment above), this should only accept `HashSet<IndexedTypeIdentifier>`.

engine/src/main/java/org/hibernate/search/spi/impl/HashSetIndexedTypesSet.java
```diff
@@ -48,6 +49,14 @@ public boolean isEmpty() {
 		return set.stream().map( IndexedTypeIdentifier::getPojoType ).collect( Collectors.toSet() );
 	}
 
+	boolean contains(IndexedTypeIdentifier id) {
+		return set.contains( id );
+	}
+
+	Set<IndexedTypeIdentifier> cloneMap() {
```
yrodiere: Er... What? `cloneMap` returns a `Set`? :]

engine/src/main/java/org/hibernate/search/spi/impl/IndexedTypesSets.java
```diff
@@ -81,4 +81,67 @@ else if ( types.length == 1 ) {
 		}
 	}
 
+	public static IndexedTypesSet composite(final IndexedTypesSet set, final IndexedTypeIdentifier additionalId) {
```
yrodiere: Maybe this should be an instance method? Since it isn't null-safe anyway...

engine/src/main/java/org/hibernate/search/spi/impl/IndexedTypesSets.java
```diff
@@ -81,4 +81,67 @@ else if ( types.length == 1 ) {
...
+			HashSet<IndexedTypeIdentifier> newSet = new HashSet<>( set.size() + 1 );
+			for ( IndexedTypeIdentifier existingId : set ) {
+				newSet.add( existingId );
+			}
+			newSet.add( additionalId );
+			return new HashSetIndexedTypesSet( newSet );
+		}
+	}
+
+	public static IndexedTypesSet composite(final IndexedTypesSet setA, final IndexedTypesSet setB) {
```
yrodiere: Same as above:

> Maybe this should be an instance method? Since it isn't null-safe anyway...

engine/src/main/java/org/hibernate/search/spi/impl/IndexedTypesSets.java
```diff
@@ -81,4 +81,67 @@ else if ( types.length == 1 ) {
...
+			return new HashSetIndexedTypesSet( newSet );
+		}
+	}
+
+	/**
+	 * @param containedEntityTypes
+	 * @param initializedContainedEntityTypes
+	 * @return
+	 */
+	public static IndexedTypesSet subtraction(IndexedTypesSet referenceSet, IndexedTypesSet subtraend) {
```
yrodiere: Same as above:

> Maybe this should be an instance method? Since it isn't null-safe anyway...

engine/src/main/java/org/hibernate/search/spi/SearchIntegrator.java
```diff
@@ -57,21 +57,6 @@
...
-	 * Offers a few execution approaches:
-	 * - return the list of results eagerly
-	 * - return the list of results lazily
-	 * - get the number of results
-	 *
-	 * @return an Hibernate Search query object
-	 * @deprecated This method will be removed. Use {@link #createHSQuery(Query, Class...)}
-	 */
-	@Deprecated
-	HSQuery createHSQuery();
```
Sanne: moved back to take it into consideration.. we'll move it out as needed.

engine/src/main/java/org/hibernate/search/engine/impl/ImmutableSearchFactory.java
```diff
@@ -324,17 +331,22 @@ private QueryDescriptor createQueryDescriptor(Query luceneQuery, Class<?>[] enti
...
 	}
 
 	@Override
-	public Map<Class<?>, EntityIndexBinding> getIndexBindings() {
+	public IndexedTypeMap<EntityIndexBinding> getIndexBindings() {
 		return indexBindingForEntities;
 	}
 
-	@Override
+	@Override @Deprecated
```
Sanne: since you insist..

engine/src/main/java/org/hibernate/search/spi/impl/ConcurrentIndexedTypeMap.java
```diff
@@ -0,0 +1,76 @@
...
+
+import java.util.Collection;
+import java.util.Map.Entry;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.hibernate.search.spi.IndexedTypeIdentifier;
+import org.hibernate.search.spi.IndexedTypeMap;
+import org.hibernate.search.spi.IndexedTypesSet;
+
+public class ConcurrentIndexedTypeMap<V> implements IndexedTypeMap<V> {
```
Sanne: This whole thing is temporary, I already have several other candidate prototypes for later..

It's a placeholder and I'd prefer to just avoid conflicts for now.

Once it's in, you'll also be able to propose changes as you prefer.

engine/src/main/java/org/hibernate/search/query/dsl/impl/ConnectedQueryContextBuilder.java
```diff
@@ -36,34 +37,35 @@ public ConnectedQueryContextBuilder(ExtendedSearchIntegrator factory) {
...
 			if ( indexBoundType == null ) {
-				Set<Class<?>> configuredSubTypes = factory.getConfiguredTypesPolymorphic( new Class<?>[] { entityType } );
+				IndexedTypesSet configuredSubTypes = factory.getConfiguredTypesPolymorphic( new Class<?>[] { entityType.getPojoType() } );
 				if ( configuredSubTypes.isEmpty() ) {
-					throw log.cantQueryUnconfiguredType( entityType.getCanonicalName() );
+					throw log.cantQueryUnconfiguredType( entityType.getName() );
 				}
 				else {
-					throw log.cantQueryUnindexedType( entityType.getCanonicalName() );
+					throw log.cantQueryUnindexedType( entityType.getName());
```
Sanne: I'm surprised checkstyle didn't fail ?!

engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
```diff
@@ -149,17 +153,18 @@ public final HSQuery targetedTypes(List<CustomTypeMetadata> types) {
 			return targetedEntities( null );
 		}
 
-		List<Class<?>> entityTypes = types.stream()
+		List<IndexedTypeIdentifier> entityTypes = types.stream()
 				.map( CustomTypeMetadata::getEntityType ).collect( toList() );
```
Sanne: I agree but let's do after the next phase: I'm pretty confident that with another general pass we wouldn't even have a `List<CustomTypeMetadata>` as method input ;)

engine/src/main/java/org/hibernate/search/spi/impl/IndexedTypesSets.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Objects;
+import java.util.Set;
+
+import org.hibernate.search.spi.IndexedTypeIdentifier;
+import org.hibernate.search.spi.IndexedTypesSet;
+
+public final class IndexedTypesSets {
```
Sanne: Since it's an `.impl` we can definitely explore such things.

I think we need a type system to be defined first, e.g. a POC not using Class<>, to see if we can totally hide creation of new type-sets. At that point we can see how set operations can be created optimally, possibly even allocating most things only at bootstrap?

engine/src/main/java/org/hibernate/search/spi/impl/IndexedTypesSets.java
Sanne: `IndexedTypesSet` is an SPI exposed type, while IndexedTypesSets helper is currently not.

I guess we might end up realising that we need to expose such helpers too, but I'd hope to try avoiding that, and then when we have more stuff cleaned up see if it's really useful to integrators to access such helpers.

engine/src/main/java/org/hibernate/search/spi/IndexedTypesSet.java
```diff
@@ -0,0 +1,24 @@
...
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.spi;
+
+import java.util.Set;
+
+public interface IndexedTypesSet extends Iterable<IndexedTypeIdentifier> {
```
Sanne: there are deep thoughts here.. let's talk about it if your'e interested.

My whole point is to strictly avoid - or at least discourage - people to use their own collections to aggregate somehow a "type set" concept from multiple type-ids.

I definitely don't like throwing UnsupportedOperationException on the other methods, when there's no real reason to have a standard Set here.

By  exposing a very small API we can control how ti make it efficient.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/schema/impl/Elasticsearch2SchemaTranslator.java
```diff
@@ -503,7 +503,7 @@ else if ( indexedNullToken instanceof Boolean ) {
 			return new JsonPrimitive( (Boolean) indexedNullToken );
 		}
 		else {
-			throw LOG.unsupportedNullTokenType( mappingBuilder.getBeanClass(), fieldPath.getAbsoluteName(),
+			throw LOG.unsupportedNullTokenType( mappingBuilder.getBeanClass().getName(), fieldPath.getAbsoluteName(),
```
Sanne: oh.. you're right. I'm trying but it feels like I just found a new mining vein... let's see how far it goes :)

engine/src/main/java/org/hibernate/search/engine/impl/WorkPlan.java
```diff
@@ -145,7 +147,8 @@ public void processContainedInAndPrepareExecution() {
 	 */
 	public <T> void recurseContainedIn(T value, ContainedInRecursionContext context, String tenantId) {
 		Class<T> entityClass = instanceInitializer.getClass( value );
-		PerClassWork classWork = getClassWork( tenantId, entityClass );
+		//TODO separate the ContainedIn processing in its own registry of types
```
Sanne: You can't forget about it. If you ever try to figure out how to encode non-pojo types you'll have to rewrite this... it's useful to have a TODO here to help with "what do I do here" at that point.

engine/src/main/java/org/hibernate/search/indexes/spi/DirectoryBasedIndexManager.java
Sanne: Conceptually it's the same?

engine/src/main/java/org/hibernate/search/query/engine/impl/EntityInfoImpl.java
Sanne: ah, yes. This was actually required, but in intermediate stages which are long gone.

engine/src/main/java/org/hibernate/search/spi/impl/HashSetIndexedTypesSet.java
Sanne: well it was a Map originally :)

engine/src/main/java/org/hibernate/search/spi/IndexedTypesSet.java
Sanne: ok doing this but it feels like exotic english..

engine/src/main/java/org/hibernate/search/spi/impl/HashSetIndexedTypesSet.java
yrodiere: You mean it's a copy/paste mistake? Because the thing being cloned is not a map, either...

Sanne: Yes I mean it's a silly mistake. The whole class was representing Maps, a while back. And cloning them..

Sanne:
should be good now. I'll let CI finish.. feel free to merge it as I guess you'll start earlier ;)

yrodiere:
Jenkins, retest this please

yrodiere:
The CI is down, but the build passes locally. Merged!

PR #1443, opened 2017-06-08
HSEARCH-2761 IndexingActionInterceptorTest failing randomly

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2761


Sanne:
Please explain :) Why is this a good solution? It feels like inherently unsafe..

yrodiere:
> Please explain :) Why is this a good solution? It feels like inherently unsafe..

As was mentioned in the commit message:

> This should avoid *some* cases where delete-by-query failed due to a
> version conflict: those cases where the update was performed
> strictly before the delete-by-query (for instance by the same thread).

But thinking about it, there should have been a refresh after those updates, so this explanation doesn't hold up... I'll continue working on it.

yrodiere:
Closing as invalid.

PR #1444, opened 2017-06-08
HSEARCH-2767 Make ElasticsearchWorkProcessor use one execution context per runnable once again

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2767


Sanne:
merged

PR #1445, opened 2017-06-08
HSEARCH-2761 Elasticsearch purges may fail when purging multiple classes of the same inheritance tree in the same transaction

yrodiere:
This PR is based on #1444, which should be merged first.

https://hibernate.atlassian.net//browse/HSEARCH-2761

@Sanne I added non-regression tests. Stricly speaking, only "HSEARCH-2761 Always refresh indexes before performing a delete-by-query" is needed to fix the issue, but the other commits fix hidden issues (discussed in the ticket).


Sanne:
Looks very good! Doing some more testing here to understand it better.

One downside: those type related collections are not just conflicting with my PR but I'll need to re-implement some similar approach. The good part is that with the immutable, pre-computed collection I should be able to guarantee a stable order w/o having a performance penalty.

Sanne:
merged 🥇

PR #1450, opened 2017-06-14
HSEARCH-2775 Migrate internal file helpers and SPIs from File to NIO …

Sanne:
…APIs

A bit of a mindless, boring translation task.. but IMO the code looks much more readable now, and I could get rid of several deprecated methods.

https://hibernate.atlassian.net/browse/HSEARCH-2775

### DIFF
--- a/engine/src/main/java/org/hibernate/search/store/impl/DefaultLockFactoryCreator.java
+++ b/engine/src/main/java/org/hibernate/search/store/impl/DefaultLockFactoryCreator.java
@@ -7,6 +7,7 @@
 package org.hibernate.search.store.impl;
 
 import java.io.File;
+import java.nio.file.Path;
 import java.util.Properties;
 
 import org.apache.lucene.store.LockFactory;
@@ -43,7 +44,7 @@ public class DefaultLockFactoryCreator implements LockFactoryCreator, Startable 
 	}
 
 	@Override
-	public LockFactory createLockFactory(File indexDir, Properties dirConfiguration) {
+	public LockFactory createLockFactory(Path indexDir, Properties dirConfiguration) {
 		//For FS-based indexes default to "native", default to "single" otherwise.
 		String defaultStrategy = indexDir == null ? "single" : "native";
 		String lockFactoryName = dirConfiguration.getProperty( Environment.LOCKING_STRATEGY, defaultStrategy );
@@ -72,7 +73,8 @@ public class DefaultLockFactoryCreator implements LockFactoryCreator, Startable 
 					Environment.LOCKING_STRATEGY,
 					serviceManager
 			);
-			return lockFactoryFactory.createLockFactory( indexDir, dirConfiguration );
+			File legacy = indexDir == null ? null : indexDir.toFile();
+			return lockFactoryFactory.createLockFactory( legacy, dirConfiguration );
 		}
 	}
 }
--- a/engine/src/main/java/org/hibernate/search/store/impl/FSDirectoryProvider.java
+++ b/engine/src/main/java/org/hibernate/search/store/impl/FSDirectoryProvider.java
@@ -6,11 +6,12 @@
  */
 package org.hibernate.search.store.impl;
 
-import java.io.File;
 import java.io.IOException;
+import java.nio.file.Path;
 import java.util.Properties;
 
 import org.apache.lucene.store.FSDirectory;
+import org.hibernate.search.exception.AssertionFailure;
 import org.hibernate.search.exception.SearchException;
 import org.hibernate.search.indexes.spi.DirectoryBasedIndexManager;
 import org.hibernate.search.spi.BuildContext;
@@ -40,15 +41,13 @@ public class FSDirectoryProvider implements DirectoryProvider<FSDirectory> {
 	private static final Log log = LoggerFactory.make();
 
 	private FSDirectory directory;
-	private String indexName;
 
 	@Override
 	public void initialize(String directoryProviderName, Properties properties, BuildContext context) {
 		// on "manual" indexing skip read-write check on index directory
 		boolean manual = IndexingMode.MANUAL == context.getIndexingMode();
-		File indexDir = DirectoryHelper.getVerifiedIndexDir( directoryProviderName, properties, !manual );
+		Path indexDir = DirectoryHelper.getVerifiedIndexPath( directoryProviderName, properties, !manual );
 		try {
-			indexName = indexDir.getCanonicalPath();
 			//this is cheap so it's not done in start()
 			directory = DirectoryProviderHelper.createFSIndex( indexDir, properties, context.getServiceManager() );
 		}
@@ -79,24 +78,12 @@ public class FSDirectoryProvider implements DirectoryProvider<FSDirectory> {
 
 	@Override
 	public boolean equals(Object obj) {
-		// this code is actually broken since the value change after initialize call
-		// but from a practical POV this is fine since we only call this method
-		// after initialize call
-		if ( obj == this ) {
-			return true;
-		}
-		if ( obj == null || !( obj instanceof FSDirectoryProvider ) ) {
-			return false;
-		}
-		return indexName.equals( ( (FSDirectoryProvider) obj ).indexName );
+		throw new AssertionFailure( "this type can not be compared reliably" );
 	}
 
 	@Override
 	public int hashCode() {
-		// this code is actually broken since the value change after initialize call
-		// but from a practical POV this is fine since we only call this method
-		// after initialize call
-		int hash = 11;
-		return 37 * hash + indexName.hashCode();
+		throw new AssertionFailure( "this type can not be compared reliably" );
 	}
+
 }
--- a/engine/src/main/java/org/hibernate/search/store/impl/FSMasterDirectoryProvider.java
+++ b/engine/src/main/java/org/hibernate/search/store/impl/FSMasterDirectoryProvider.java
@@ -55,7 +55,6 @@ public class FSMasterDirectoryProvider implements DirectoryProvider<FSDirectory>
 
 	//variables having visibility granted by a read of "current"
 	private FSDirectory directory;
-	private long copyChunkSize;
 
 	//variables needed between initialize and start (used by same thread: no special care needed)
 	private Path sourceDir;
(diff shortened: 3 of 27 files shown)


engine/src/main/java/org/hibernate/search/util/impl/FileHelper.java
```diff
@@ -84,58 +84,50 @@ else if ( !destination.isDirectory() ) {
...
-				);
-			}
-			String[] sources = source.list();
-			Set<String> srcNames = new HashSet<String>( Arrays.asList( sources ) );
-			String[] dests = destination.list();
+		if ( Files.isDirectory( source ) ) {
+			Files.createDirectories( destination );
+			Set<Path> sources = Files.list( source ).collect( Collectors.toSet() );
+			Set<String> sourceFilenames = sources.stream().map( v -> v.getFileName().toString() ).collect( Collectors.toSet() );
+			Set<Path> dests = Files.list( destination ).collect( Collectors.toSet() );
```
yrodiere: I'm pretty sure you don't need this set. Just using `Files.list( destination ).map` instead of `dests.stream().map` when building `destFilenames` below would work equally well.

And while we're at it, I don't think you need to collect destination files at all. This would work just fine for deletion, and destination files are not used anywhere else:

```
			//delete files not present in source
			try ( DirectoryStream<Path> dests = Files.newDirectoryStream( source ) ) {
				for ( Path dest : dests ) {
					if ( !sourceFilenames.contains( dest.getFileName().toString() ) ) {
						delete( dest );
					}
				}
			}
```

You could also avoid collecting source file names and just resolve each dest file name against the source, but I'm not sure there's a benefit in that.

engine/src/main/java/org/hibernate/search/util/impl/FileHelper.java
```diff
@@ -84,58 +84,50 @@ else if ( !destination.isDirectory() ) {
...
-						"Source and Destination not of the same type:"
-								+ source.getCanonicalPath() + " , " + destination.getCanonicalPath()
-				);
-			}
-			String[] sources = source.list();
-			Set<String> srcNames = new HashSet<String>( Arrays.asList( sources ) );
-			String[] dests = destination.list();
+		if ( Files.isDirectory( source ) ) {
+			Files.createDirectories( destination );
+			Set<Path> sources = Files.list( source ).collect( Collectors.toSet() );
```
yrodiere: I must admit I'm not very familiar with nio, but shouldn't we explicitly close the streams? (`stream.close()`)

From the javadoc of `Files.list`:

```
     * <p> The returned stream encapsulates a {@link DirectoryStream}.
     * If timely disposal of file system resources is required, the
     * {@code try}-with-resources construct should be used to ensure that the
     * stream's {@link Stream#close close} method is invoked after the stream
     * operations are completed.
```

engine/src/test/java/org/hibernate/search/test/util/impl/FileHelperTest.java
```diff
@@ -48,19 +50,19 @@ public static void prepareRootDirectory() {
...
+		Path file = dir.resolve( name );
+		Files.createFile( file );
 		writeDummyDataToFile( file );
 		return file;
 	}
 
-	private void writeDummyDataToFile(File file) throws IOException {
-		FileOutputStream os = new FileOutputStream( file, true );
+	private void writeDummyDataToFile(Path file) throws IOException {
+		FileOutputStream os = new FileOutputStream( file.toFile(), true );
```
yrodiere: I think the nio way would be: `Files.newOutputStream( file, StandardOpenOption.APPEND )`;

engine/src/main/java/org/hibernate/search/store/impl/FSMasterDirectoryProvider.java
```diff
@@ -69,12 +71,12 @@ public void initialize(String directoryProviderName, Properties properties, Buil
 		this.properties = properties;
 		this.directoryProviderName = directoryProviderName;
 		//source guessing
-		sourceDir = DirectoryProviderHelper.getSourceDirectory( directoryProviderName, properties, true );
-		log.debugf( "Source directory: %s", sourceDir.getPath() );
-		indexDir = DirectoryHelper.getVerifiedIndexDir( directoryProviderName, properties, true );
-		log.debugf( "Index directory: %s", indexDir.getPath() );
+		sourceDir = DirectoryProviderHelper.getSourceDirectoryPath( directoryProviderName, properties, true );
+		log.debugf( "Source directory: %s", sourceDir );
+		indexDir = DirectoryHelper.getVerifiedIndexPath( directoryProviderName, properties, true );
```
yrodiere: You may want to normalize the path at some point (`indexDir = indexDir.normalize()`), since previously we used `indexDir.getCanonicalPath()` to set `indexName`. Or maybe it's not important now that we don't use it for `equals` anymore?

engine/src/main/java/org/hibernate/search/store/impl/FSSlaveDirectoryProvider.java
```diff
@@ -76,16 +78,11 @@ public void initialize(String directoryProviderName, Properties properties, Buil
...
-		log.debugf( "Index directory: %s", indexDir.getPath() );
-		try {
-			indexName = indexDir.getCanonicalPath();
-		}
-		catch (IOException e) {
-			throw new SearchException( "Unable to initialize index: " + directoryProviderName, e );
-		}
+		sourceIndexDir = DirectoryProviderHelper.getSourceDirectoryPath( directoryProviderName, properties, false );
+		log.debugf( "Source directory: %s", sourceIndexDir );
+		indexDir = DirectoryHelper.getVerifiedIndexPath( directoryProviderName, properties, true );
```
yrodiere: Same remark as in `FSMasterDirectoryProvider.java` about normalizing `indexDir`.

engine/src/main/java/org/hibernate/search/store/impl/FSDirectoryProvider.java
```diff
@@ -79,24 +78,12 @@ public FSDirectory getDirectory() {
...
-		// but from a practical POV this is fine since we only call this method
-		// after initialize call
-		if ( obj == this ) {
-			return true;
-		}
-		if ( obj == null || !( obj instanceof FSDirectoryProvider ) ) {
-			return false;
-		}
-		return indexName.equals( ( (FSDirectoryProvider) obj ).indexName );
+		throw new AssertionFailure( "this type can not be compared reliably" );
```
Sanne: We used to need this, so for that reason they had "reasonable" implementations.

That requirement is gone - since many years a I suspect - and since it's damn complex to do it "right" I prefer to introduce the assertion to make sure we never rely on it.

engine/src/main/java/org/hibernate/search/store/impl/FSMasterDirectoryProvider.java
Sanne: I think it's no longer crucial, but yes I'll do it. Will at least look better when logging the dir..

engine/src/main/java/org/hibernate/search/store/impl/FSMasterDirectoryProvider.java
```diff
@@ -223,11 +205,9 @@ public void run() {
 					log.unableToSynchronizeSource( indexName, e );
 					return;
 				}
-				if ( ! new File( destination, CURRENT_DIR_NAME[oldIndex] ).delete() ) {
-					log.unableToRemovePreviousMarket( indexName );
-				}
 				try {
-					new File( destination, CURRENT_DIR_NAME[index] ).createNewFile();
+					Files.delete( destination.resolve( CURRENT_DIR_NAME[oldIndex] ) );
```
Sanne: that's right. The File API doesn't throw an error when it fails, it would just return false.
The NIO API throws appropriate exceptions.

engine/src/main/java/org/hibernate/search/util/impl/FileHelper.java
Sanne: Good catch. I assumed this was handled by the Stream/Collector framework but apparently it will only do that if you flatmap the stream (?!)

engine/src/main/java/org/hibernate/search/util/impl/FileHelper.java
```diff
@@ -84,58 +84,50 @@ else if ( !destination.isDirectory() ) {
...
-				tryDelete( destination.toPath() );
-			}
-			if ( destination.exists() ) {
-				long sts = source.lastModified() / FAT_PRECISION;
-				long dts = destination.lastModified() / FAT_PRECISION;
+			if ( Files.exists( destination ) && Files.isDirectory( destination ) ) {
+				tryDelete( destination );
+			}
+			final File sourceFile = source.toFile();
+			final File destinationFile = destination.toFile();
```
Sanne: ok, but I just wanted to remove a `File` reference from a deprecated SPI and look what it got me into :)

engine/src/main/java/org/hibernate/search/store/impl/DefaultLockFactoryCreator.java
```diff
@@ -72,7 +73,8 @@ else if ( "none".equals( lockFactoryName ) ) {
 					Environment.LOCKING_STRATEGY,
 					serviceManager
 			);
-			return lockFactoryFactory.createLockFactory( indexDir, dirConfiguration );
+			File legacy = indexDir == null ? null : indexDir.toFile();
+			return lockFactoryFactory.createLockFactory( legacy, dirConfiguration );
```
Sanne: https://hibernate.atlassian.net/browse/HSEARCH-2779

Sanne:
rebased and updated

engine/src/main/java/org/hibernate/search/util/impl/FileHelper.java
yrodiere: @Sanne Unless I'm mistaken, you didn't address this... ?

Sanne: @yrodiere you're right. How embarrassing. There are two blocks of code which are very similar, I fixed one and missed the other.

Sanne:
Fixed it an added a bonus commit for:
 - - https://hibernate.atlassian.net/browse/HSEARCH-2783

yrodiere:
Merged! Thanks

PR #1452, opened 2017-06-14
HSEARCH-2774 The default build should include documentation and javad…

Sanne:
…oc validation

 - https://hibernate.atlassian.net/browse/HSEARCH-2774

### DIFF
--- a/documentation/pom.xml
+++ b/documentation/pom.xml
@@ -181,6 +181,11 @@
     <profiles>
         <profile>
             <id>documentation-pdf</id>
+            <activation>
+                <property>
+                    <name>!skipPdfDocumentation</name>
+                </property>
+            </activation>
             <build>
                 <plugins>
                     <plugin>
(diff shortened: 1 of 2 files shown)


documentation/pom.xml
```diff
@@ -181,6 +181,9 @@
     <profiles>
         <profile>
             <id>documentation-pdf</id>
+            <activation>
+                <activeByDefault>!skipPdfDocumentation</activeByDefault>
```
yrodiere: This is not a good idea. `activeByDefault` means "active if no other profile is enabled" (don't ask me why, just another maven weirdness).
In our case, we have other profiles, so this change didn't make the documentation-pdf profile activate by default.

You may want to use this instead:

```
            <activation>
                <property>
                    <name>!skipPdfDocumentation</name>
                </property>
            </activation>
```

Sanne: oops yes I realized that when testing as I initially used just `activeByDefault`. In fact what I committed later is illegal, activeByDefault doesn't expect a property..

Sanne:
pushed a new version, and changed the scope as commented on JIRA.

yrodiere:
Jenkins, retest this please

PR #1454, opened 2017-06-15
HSEARCH-2772 Add an elasticsearch-aws module

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2772

Since it's now possible, I also made sure one can execute JGroups/WildFly tests against AWS.

Proof it works: http://ci.hibernate.org/view/Search/job/hibernate-search-master-elasticsearchAWS/40/


yrodiere:
Jenkins, retest this please...

Sanne:
Could you briefly explain why it had to copy verbatim test resources from the aws-signer project?

I'm uncomfortable with copying the ASL2 License file as we would need to make it clear to what portion it applies, _and which not_ ...

yrodiere:
> Could you briefly explain why it had to copy verbatim test resources from the aws-signer project?

I did not. The test suite is provided by AWS to test signer implementations. It's a TCK, kinda. You can get it there: http://docs.aws.amazon.com/general/latest/gr/signature-v4-test-suite.html 

In any case, we need the official AWS test suite to make sure signing works as expected. Running our own test suite against AWS is not enough in my opinion, because some cases that could happen in the future are not covered. And in fact, when I introduced the AWS test suite, only about half of it passed even though all of our integration tests were already passing.

Regarding the license file, the easier option may be to move the whole signing stuff to a separate, ASL2-licensed project... ?

I feel some explanation is needed regarding the implementation:

* The dependency I'm using is a community project available here: https://github.com/lucasweb78/aws-v4-signer-java ; crucially, this project is ASL2-licensed.
* This implementation is not well tested. In particular it didn't go through the official AWS test suite.
* Probably as a result, this implementation has quite a lot of bugs, that I had to circumvent. That's why we have our own `AWSSigner` class wrapping the other.
* I searched a lot, and there is no better alternative. The official AWS SDK only provides signing for requests that were built using the SDK; we can't do that, so if we wanted to use it to sign our requests we'd have to use internals. Other community projects like https://github.com/inreachventures/aws-signing-request-interceptor are more mature, but have bugs that cannot be worked around, and they are not well tested either (they don't use the AWS test suite).

When we were only using `aws-v4-signer-java` for test, I didn't care enough to do much about. But now that we intend to use it in Hibernate Search, we may want to address its problems...
In the future, I see three options:

1. Either I contribute my fixes and tests to the `aws-v4-signer-java` project, and when it's released we use it.
2. Or we re-implement what is currently in the `aws-v4-signer-java` project in Hibernate Search, and get rid of the dependency. It wouldn't be that bad, really; there's only about 200 lines of meaningful code in `aws-v4-signer-java`, all of which are really only about following instructions published in the AWS documentation.
3. Or we do option 2 (re-implement AWS signing), but in a separate, ASL2-licensed project of our own.

The second and third options may be easier, because I had to use some Apache dependencies, and `aws-v4-signer-java` makes it very clear that "no dependencies" is an important feature it provides.

yrodiere:
So, after our discussion, I opened a pull request here: https://github.com/lucasweb78/aws-v4-signer-java/pull/4
On our side, I created a branch that uses the hypothetical aws-v4-signer-java v1.3: https://github.com/yrodiere/hibernate-search/tree/HSEARCH-2772-aws-v4-signer-1.3 . Of course, you have to build and install aws-v4-signer-java for this to work.

Let's wait and see if the PR is accepted.

yrodiere:
@Sanne My changes have been merged into aws-v4-signer-java and released. I updated this PR to simply use the new version of aws-v4-signer-java. On the bright side, this PR is now much smaller ;)

Sanne:
Excellent! Thanks

Sanne:
There is now a duplicated class `org/hibernate/search/elasticsearch/testutil/JsonHelper.java` .

I didn't check yet if/why this is necessary, but in case it's necessary to not depend on one (or the the other) might be best to change the name of one of them to make sure there's no confusion in the future?

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -139,6 +139,24 @@ you will need the new `hibernate-search-elasticsearch` jar.
 ----
 ====
 
+If your Elasticsearch instance is hosted on Amazon Web Services,
```
Sanne: We need to clarify this, in particular what you mean by "hosted"

I talked with David Pilato about this feature and it turns out the Elasticsearch is not involved with the AWS-managed services, so they generally recommend running it "unmanaged" on your own hardware (even though possibly run on EC2) or via their own cloud offering - which is also hosted on AWS.

So I think:
 - "running it on AWS" is not the right premise
 - also referring to "and if your instance is not public" -> nobody will do that right?

What about "If you use Amazon's proprietary `IAM` authentication mechanism to access Elasticsearch..." ?

modules/src/main/modules/search/elasticsearch/module.xml
```diff
@@ -18,5 +18,9 @@
         <module name="org.hibernate.commons-annotations"/>
         <module name="org.jboss.logging"/>
         <module name="org.apache.httpcomponents"/>
+
+        <!-- AWS authentication support -->
+        <!-- Note: we must export the services (hence the classes), because services are loaded using the -engine classloader -->
+        <module name="org.hibernate.search.elasticsearch-aws" export="true" services="export" optional="true" slot="${hibernate.search.version}" />
```
Sanne: I understand the need for `services="export"`. But are you sure that `export="true"` was also necessary?

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -139,6 +139,24 @@ you will need the new `hibernate-search-elasticsearch` jar.
...
+and if your instance is not public,
+you will need an additional dependency to handle authentication:
+`hibernate-search-elasticsearch-aws`.
+
+.Maven dependencies for Hibernate Search with AWS-hosted Elasticsearch
```
Sanne: Should we clarify these are not necessary when using our modules pack on WildFly ?

yrodiere: I don't think so.
First, because there is no such explanation anywhere else in the documentation, so if we start talking about Wildfly (and I'm not sure we should), it should be in a dedicated section.
Second, because if consume a Wildfly module, I'd be really surprised to have to include part of the library as modules, and part of it as JARs in my deployment... It would just be weird, wouldn't it?

modules/src/main/modules/search/elasticsearch/module.xml
yrodiere: Unfortunately I am... If I don't export classes too, the service file (`META-INF/services/org.hibernate.search.elasticsearch.client.spi.ElasticsearchHttpClientConfigurer`) will be found, but the implementation mentioned in this file will not, and the deployment will fail: https://gist.github.com/yrodiere/420424979363cc07f3adc82cd8bf55b7
I find it weird, but I suppose it makes sense in some way.

yrodiere:
Jenkins, retest this please

Sanne:
merged! Great work. I've added a follow up PR to propose two very minor changes: pr #1473

PR #1455, opened 2017-06-16
HSEARCH-2781 Rework internals of sharding to make it possible for backends not to rely on Class<?>

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2781

After this patch, integrators will be able to write `entityIndexBinding.getIndexManagerSelector().forExisting(typeId, null, null)` to get all index managers containing a document for the given type. No use of `Class<?>` required.


engine/src/main/java/org/hibernate/search/indexes/impl/IndexShardingStrategyIndexManagerSelector.java
```diff
@@ -0,0 +1,74 @@
...
+	private final IndexManager[] allIndexManagers;
+
+	public IndexShardingStrategyIndexManagerSelector(IndexShardingStrategy shardingStrategy, IndexManager[] allIndexManagers) {
+		super();
+		this.shardingStrategy = shardingStrategy;
+		this.allIndexManagers = allIndexManagers;
+	}
+
+	@Override
+	public Set<IndexManager> all() {
```
Sanne: The `Set` returned by this method could be initialized during constructor?

Sanne:
merged. Thanks!

engine/src/main/java/org/hibernate/search/indexes/impl/IndexShardingStrategyIndexManagerSelector.java
yrodiere: I see you did it; thanks.
It should work, yes, but I wasn't sure whether strategies were allowed to add/remove index managers at runtime or not. I guess they're not.

Sanne: Sure. That's the main difference between the dynamic sharding strategies and the "legacy" (this one)

PR #1456, opened 2017-06-16
HSEARCH-2369 HSQuery.targetedEntities(List<Class<?>>) won't always work when using multiple backends 

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2369

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchHSQueryImpl.java
@@ -73,7 +73,9 @@ import org.hibernate.search.query.facet.Facet;
 import org.hibernate.search.query.facet.FacetSortOrder;
 import org.hibernate.search.query.facet.FacetingRequest;
 import org.hibernate.search.spatial.DistanceSortField;
+import org.hibernate.search.spi.CustomTypeMetadata;
 import org.hibernate.search.spi.IndexedTypeIdentifier;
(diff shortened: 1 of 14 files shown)


Sanne:
I merged it as it's a step in the right direction, but I'm not 100% happy about exposing two methods on the `org.hibernate.search.query.engine.spi.QueryDescriptor` SPI ?

Specifically, I don't see why it should be a `List<CustomTypeMetadata>` but refrained from changing it, especially as to not add an equal/hashcode requirement to the `CustomTypeMetadata` implementors.

yrodiere:
> I merged it as it's a step in the right direction, but I'm not 100% happy about exposing two methods on the org.hibernate.search.query.engine.spi.QueryDescriptor SPI ?

We basically have two ways of creating queries: using only boot-time metadata (passing type IDs), and using those but overriding some metadata at query time (passing a `CustomTypeMetadata` for each targeted type). So I'm not sure how you'd do that with only one method... ? You'd pass both an `IndexedTypeSet` and a `List<CustomTypeMetadata>` in a single method, and expect all types referred to in the `CustomTypeMetadata` to be present in the `IndexedTypeIdentifierSet`? Or maybe use an `IndexedTypeIdentifierMap`?

> Specifically, I don't see why it should be a List<CustomTypeMetadata> but refrained from changing it, especially as to not add an equal/hashcode requirement to the CustomTypeMetadata implementors.

It was basically a copy/paste from `targetedEntities`, which used to accept a `List<Class<?>>`. But you're right, we'll have to change this.

I opened https://hibernate.atlassian.net/browse/HSEARCH-2801 to fix this.

PR #1458, opened 2017-06-19
HSEARCH-2787 Replace INFO blocks in the documentation with valid styles

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2787

### DIFF
--- a/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
+++ b/documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
@@ -334,7 +334,7 @@ we suggest to use query-time boosting instead: see <<search-query>>.
 
 ==== [[elasticsearch-mapping-analyzer]] Analyzers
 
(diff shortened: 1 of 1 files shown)


Sanne:
`INFO:`
this was indeed wrong

However:
`[INFO]`
was showing a nice yellow "info" box. You didn't like these?

yrodiere:
> However:
> [INFO]
> was showing a nice yellow "info" box. 
> You didn't like these?

As mentioned in the ticket, I think the INFO style doesn't even exist: http://asciidoctor.org/docs/user-manual/#style

In our case it was showing as a "warning" box, but I suspect that's the default behavior for unknown styles.
I replaced INFO with NOTE because it seemed to be the closest to what you wanted, but WARNING is fine by me, as long as it's really the meaning you want to convey.

Sanne:
ok let's apply this then. I kinda liked the "fallback style" though :)

PR #1462, opened 2017-06-22
HSEARCH-2789 Test that the CREATE strategy actually doesn't create mappings when the index already exists

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2789


Sanne:
merged

PR #1466, opened 2017-06-26
HSEARCH-2791 Fix integration tests failing on non-H2 databases due to mapping specifics

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2791

Internal CI:
 * Before this PR: https://jenkins.hosts.mwqe.eng.bos.redhat.com/hudson/view/Hibernate/view/Hibernate%20Community/job/hibernate-search-master-dbmatrix/
 * After this PR:  https://jenkins.hosts.mwqe.eng.bos.redhat.com/hudson/view/Hibernate/view/Hibernate%20Community/job/hibernate-search-master-dbmatrix-temp-yrodiere/

As you can see it's not exactly an absolute victory, but I think it solves most of the problems and I already spent way too much time on this.

Remaining issues:

* Tests won't work at all on DB2. The error messages are quite cryptic, but after a bit of googling it may be caused by a failure in schema generation.
* Envers integration testing fails on Oracle.
* Tests involving heavily concurrent DB access fail on MS SQL Server
* `DeleteByQueryMultiTenancyIT.deleteByQueryOnlyAffectsEntitiesOfCurrentTenant` fails on MS SQL Server when executing `drop table hibernate_sequence` (???)
* Many tests won't work on Sybase, seemingly because of schema initialization failures.

I'm pretty sure all those are configuration problems, rather than actual bugs (the errors seem way too specific, and we don't do anything exotic with Hibernate ORM). But I don't understand the CI jobs very well; in particular several hibernate properties seem to be overridden programmatically by the CI job at some point, and I fail to see the reasoning behind that (beyond the JDBC URL and credentials).


Sanne:
nice cleanup! Thanks.

Opened a follow-up issue for each of your bullet points. They are marked for 5.8 but aren't urgent, we'll work on these only in case someone happens to have some free time.

PR #1467, opened 2017-06-28
[Preview-Do not merge] Request for comments on Elasticsearch performance tests

yrodiere:
Not to be merged: work in progress.

New performance tests for the Elasticsearch module. Their purpose is to give us an idea of how our code performs in the real world, so I tried to use scenarios involving multiple work types, and most importantly multiple works in each transaction.

With these tests, we should be able to compare different implementations of the Elasticsearch work processor (https://hibernate.atlassian.net/browse/HSEARCH-2764).

@Sanne This is largely based on your work on the engine tests, but my additions may overlook some micro-benchmarking pitfalls. Does anything seem incorrect to you?


Sanne:
Looks like a great start!

One pitfall might be the usage of randoms.
Two different angles:
 - it's good to have a randomized source but make sure that when running the same benchmark multiple times, the same sequence is generated. You want to be able to compare them, and reproduce them when investigating issues.
 - do not generate random numbers *during* the test run, try to create all data you need upfront. The reason is that the host running the test has only a limited amount of entropy: let's say it contains enough data to feed 5,000 iterations of your test, and your test consumes about 5,000. You'll run the benchmark twice and the second time it will be significantly slower. Entropy builds up over time but this process can be surprisingly slow, especially on VMs and headless servers. Servers also consume a lot of it for all the secure connections (assuming they have no dedicated hardware to boost it).

yrodiere:
Closing to avoid CI builds while I'm working on this branch.

PR #1468, opened 2017-06-28
Enable Apache HTTP and Elasticsearch RestClient logs in tests

yrodiere:
This should help debug [HSEARCH-2682](https://hibernate.atlassian.net//browse/HSEARCH-2682) (not fixed yet), which I cannot reproduce on my machine, and which I'm having trouble to reproduce on the CI through repeated builds...


Sanne:
merged

PR #1469, opened 2017-06-28
HSEARCH-2682 Malformed HTTP responses from Elasticsearch are not handled consistently

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2682

It seems removing the (useless) timeout does the trick: I ran the test approximatively 70 times on the CI and got no failure (see http://ci.hibernate.org/job/hibernate-search-yoann/99/ and the previous ones).


Sanne:
merged

PR #1470, opened 2017-06-30
HSEARCH-2802 Make sure the default elasticsearch profile will only be enabled if the ES version is not provided

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2802

@Sanne I didn't remove your modifications to the README, it's probably safer that way.


Sanne:
merged

PR #1476, opened 2017-07-06
Cleanup deprecated methods from SPI

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2770

I verified each step foollowing along with the Infinispan project, so I know that these are safe now :)


engine/src/main/java/org/hibernate/search/spi/SearchFactoryBuilder.java
```diff
@@ -1,40 +0,0 @@
...
-package org.hibernate.search.spi;
-
-import org.hibernate.search.cfg.spi.SearchConfiguration;
-import org.hibernate.search.engine.integration.impl.ExtendedSearchIntegrator;
-
-/**
- * @deprecated Use SearchIntegrationBuilder instead. This class will be removed!
- */
-@Deprecated
-public class SearchFactoryBuilder extends SearchIntegratorBuilder {
```
yrodiere: I don't know if it was related, but maybe you could also remove the deprecated `SearchFactoryIntegrator`?

Sanne:
pushed a new version

- fixed the javadoc
- removed `SearchFactoryIntegrator`

No idea why the javadoc didn't fail locally?

engine/src/main/java/org/hibernate/search/bridge/util/impl/ContextualExceptionBridgeHelper.java
```diff
@@ -46,12 +45,6 @@
 	private final StringConversionContextImpl stringAdapter = new StringConversionContextImpl();
 
 	@Override
-	@Deprecated
-	public ConversionContext setClass(Class<?> clazz) {
-		return setConvertedTypeId( new PojoIndexedTypeIdentifier( clazz ) );
-	}
-
-	@Override
 	public ConversionContext setConvertedTypeId(IndexedTypeIdentifier type) {
```
Sanne: it's a good point but I'll do that later:
I'll send some more PRs later with more deprecation changes and will include a proposal for this.

Sanne:
ok, CI confirmed it's green: I'll merge it. Thanks for the review!

PR #1483, opened 2017-07-11
HSEARCH-2813 Add Elasticsearch performance tests

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2813

The test themselves consist of two method (mainly): write and query. They are executed concurrently.

What's particular about the write (and different from the Lucene performance tests) is that instead of using only "add" works, I use changesets, with a parameterized proportion of add/update/deletes. Currently in the JenkinsFile, I use 10 adds, 20 updates, 10 deletes for each changeset. Not sure it will change anything to the results, but I wanted to at least try to simulate an real-life workload.

Also, the write method performs a (parameterized) amount of changesets followed by a flush. This allows to also test the async mode, which otherwise would have very high scores because the actual work would not be finished when we return from the method.

As a result, the default operation counters do not make much sense. I added a "changeset" counter, which I think is the only sensible way we have to measure throughput.

Here are the other parameters:

 * client: "default" or "blackhole-<elasticsearchVersion>". Allows to test throughput without actually sending works to an Elasticsearch cluster. Obviously scores are much higher.
 * dataset: currently I use only one, "hibernate-dev-ml-2016-01", with test from the hibernate-dev mailing list.
 * indexSize
 * maxResult: max results for the query method
 * refreshAfterWrite: controls the Hibernate Search parameter with the same name
 * workerExecution: sync/async

Other parameters are set to their defaults. For instance the connection pool size is 20 (and max 2 simultaneous connections per server).

Results of running the tests on AWS are available here: http://ci.hibernate.org/view/Search/job/hibernate-search-performance-elasticsearch/19/artifact/output/elasticsearch-benchmark-results.csv

Here are the number for the "changeset" counter only:

Score	|Score Error (99.9%)	|Unit	|Param: changesetsPerFlush	|Param: client	|Param: dataset	|Param: indexSize	|Param: maxResults	|Param: refreshAfterWrite	|Param: workerExecution	|Param: worksPerChangeset
--- | --- | --- | --- | --- | --- | --- | --- | --- | --- | ---
16.316362	|1.088212	|ops/s	|100	|default	|hibernate-dev-ml-2016-01	|1000	|100	|true	|sync	|10;20
30.526409	|1.996321	|ops/s	|100	|default	|hibernate-dev-ml-2016-01	|1000	|100	|true	|async	|10;20
15.93716	|0.92353	|ops/s	|100	|default	|hibernate-dev-ml-2016-01	|1000	|100	|false	|sync	|10;20
30.311185	|2.208205	|ops/s	|100	|default	|hibernate-dev-ml-2016-01	|1000	|100	|false	|async	|10;20
7332.029931	|1651.904326	|ops/s	|100	|blackhole-5.4.0	|hibernate-dev-ml-2016-01	|1000	|100	|true	|sync	|10;20
2750.966185	|599.785469	|ops/s	|100	|blackhole-5.4.0	|hibernate-dev-ml-2016-01	|1000	|100	|true	|async	|10;20
6731.977235	|1867.217065	|ops/s	|100	|blackhole-5.4.0	|hibernate-dev-ml-2016-01	|1000	|100	|false	|sync	|10;20
2639.952722	|339.338806	|ops/s	|100	|blackhole-5.4.0	|hibernate-dev-ml-2016-01	|1000	|100	|false	|async	|10;20

First thing we can notice is we have a very low throughput (15-30 changesets per second, wich is 450-900 add/update/delete per second), and I'm wondering whether it's my tests, the AWS setup, or if Hibernate Search is actually that slow... Note that blackhole numbers are much higher (between 2500 and 7000 changesets per second, which is between 75,000 and 210,000 add/update/delete per second), so our internal processing (entity/document mapping, work dispatching) is not the bottleneck. We may not be taking advantage of parallism as much as we should, though... We'll see once my work on HSEARCH-2764 is complete.

Oddly enough, the "refreshAfterWrite" parameter doesn't seem to affect performance.

Also a bit odd, we have a twofold increase in throughput when using async processing with the "default" client (which I suppose is good news, and may be explained by better bulking), but when using the blackhole client it's the reverse: the sync throughput is twice as much as the async one. It may be related to the fact that currently, async processing is fully serial and thus will only ever use one connection, while sync processing doesn't suffer from this limitation.


yrodiere:
Jenkins, retest this please

yrodiere:
An example run (it takes quite a bit of time): http://ci.hibernate.org/view/Search/job/hibernate-search-performance-elasticsearch/20/execution/node/20/log/

Sanne:
merged! great job

PR #1485, opened 2017-07-11
HSEARCH-2818

Sanne:
The first commit is trivial.

The second one is extremely interesting: see the screenshots I attached on JIRA.

It reduced the "out of TLAB allocations" from 4,20**GB** to 2,7**MB** of the whole system. Running the blackhole/ES benchmark using its default settings.

https://hibernate.atlassian.net/browse/HSEARCH-2818

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/ElasticsearchClientUtils.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/ElasticsearchClientUtils.java
@@ -49,31 +49,30 @@ public class ElasticsearchClientUtils {
 	}
 
 	public static String formatRequest(GsonProvider gsonProvider, ElasticsearchRequest request) {
-		StringBuilder sb = new StringBuilder();
+		//Wild guess for some tuning. The only certainty is that the default (16) is too small.
+		StringBuilder sb = new StringBuilder( 180 );
 
-		sb.append( "Method: " ).append( request.getMethod() ).append( "\n" );
-		sb.append( "Path: " ).append( request.getPath() ).append( "\n" );
-
-		sb.append( "Data:\n" );
-		sb.append( formatRequestData( gsonProvider, request ) );
+		sb.append( "Method: " ).append( request.getMethod() );
+		sb.append( "\nPath: " ).append( request.getPath() );
+		sb.append( "\nData:\n" );
+		appendRequestData( sb, gsonProvider, request );
 		sb.append( "\n" );
 		return sb.toString();
 	}
 
-	public static String formatRequestData(GsonProvider gsonProvider, ElasticsearchRequest request) {
+	private static void appendRequestData(StringBuilder sb, GsonProvider gsonProvider, ElasticsearchRequest request) {
 		List<JsonObject> bodyParts = request.getBodyParts();
 		Gson gson = gsonProvider.getGsonPrettyPrinting();
-		StringBuilder builder = new StringBuilder();
 		for ( JsonObject bodyPart : bodyParts ) {
-			gson.toJson( bodyPart, builder );
-			builder.append("\n");
+			gson.toJson( bodyPart, sb );
+			sb.append("\n");
 		}
-		return builder.toString();
 	}
 
-	public static String formatRequestData(GsonProvider gsonProvider, JsonObject body) {
-		Gson gson = gsonProvider.getGsonPrettyPrinting();
-		return gson.toJson( body );
+	public static String formatRequestData(GsonProvider gsonProvider, ElasticsearchRequest request) {
+		StringBuilder sb = new StringBuilder( 180 );
+		appendRequestData( sb, gsonProvider, request );
+		return sb.toString();
 	}
 
 	public static String formatResponse(GsonProvider gsonProvider, ElasticsearchResponse response) {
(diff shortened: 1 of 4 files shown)


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
```diff
@@ -0,0 +1,205 @@
...
+import com.google.gson.Gson;
+import com.google.gson.JsonIOException;
+import com.google.gson.JsonObject;
+import com.google.gson.internal.Streams;
+import com.google.gson.stream.JsonWriter;
+
+/**
+ * Optimised adapter to encode GSON objects into HttpEntity instances.
+ * The naive approach was using various StringBuilders and the kind
+ * of objects we need to serialize into JSON where causing the internal
```
gsmet: s/where/were/

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
```diff
@@ -0,0 +1,205 @@
...
+ * StringBuilder buffers to need frequent resizing.
+ *
+ * Rather than try guessing reasonable default sizes for these buffers,
```
gsmet: `trying to guess`, maybe?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
```diff
@@ -0,0 +1,205 @@
...
+ * of the HTTP client. By doing so we can use buffer pages of fixed size.
+ *
+ * To implement this trick we have to implement both interfaces.
+ *
+ * @author Sanne Grinovero
+ */
+final class GsonHttpEntity implements HttpEntity, HttpAsyncContentProducer {
+
+	/**
+	 * A conservative quess at the size of the buffer in characters
```
gsmet: s/quess/guess/

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	@Override
+	public boolean isChunked() {
+		return false;
+	}
+
+	@Override
+	public long getContentLength() {
+		//Can't computer the size in advance w/o it being very expensive
```
gsmet: s/computer/compute/

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
```diff
@@ -0,0 +1,205 @@
...
+					// Done
+					return;
+				}
+			}
+		}
+
+		@Override
+		public void flush() throws IOException {
+			buffer.flip();
+			ByteBuffer byteBuffer = java.nio.charset.StandardCharsets.UTF_8.encode( buffer );
```
gsmet: Any reason to use the FQCN here?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
```diff
@@ -0,0 +1,205 @@
...
+import org.apache.http.message.BasicHeader;
+import org.apache.http.nio.ContentEncoder;
+import org.apache.http.nio.IOControl;
+import org.apache.http.nio.entity.HttpAsyncContentProducer;
+import org.apache.http.protocol.HTTP;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonIOException;
+import com.google.gson.JsonObject;
+import com.google.gson.internal.Streams;
```
yrodiere: This is internal API, we'd better avoid it. One solution would be to use `com.google.gson.Gson.toJson(JsonElement, JsonWriter)`; it delegates to `Streams.write`.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
```diff
@@ -0,0 +1,205 @@
...
+	@Override
+	public Header getContentEncoding() {
+		return null;
+	}
+
+	@Override
+	public InputStream getContent() {
+		//This could be implemented but would be inefficient.
+		//I therefore prefer throwing the exception so that we can easily spot unintended usage via tests.
+		throw new UnsupportedOperationException( "Not implemented! Expected to produce content only over produceContent()" );
```
yrodiere: Did you test this with elasticsearch-aws? I think we use this method to get the request body and hash it... Obviously we may want to also optimize AWS request hashing (maybe add a wrapper to the content encoder to hash on the fly), but my point is this may break AWS support.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/ElasticsearchClientUtils.java
```diff
@@ -36,69 +32,66 @@ public static boolean isSuccessCode(int code) {
...
-					sb.append( "  Type: " ).append( propertyAsString( value, "_type" ) ).append( "\n" );
-					sb.append( "  Id: " ).append( propertyAsString( value, "_id" ) ).append( "\n" );
-					sb.append( "  Status: " ).append( propertyAsString( value, "status" ) ).append( "\n" );
-					sb.append( "  Error: " ).append( propertyAsString( value, "error" ) ).append( "\n" );
+					sb.append( "\n  Index: " ).append( propertyAsString( value, "_index" ) );
+					sb.append( "\n  Type: " ).append( propertyAsString( value, "_type" ) );
+					sb.append( "\n  Id: " ).append( propertyAsString( value, "_id" ) );
+					sb.append( "\n  Status: " ).append( propertyAsString( value, "status" ) );
+					sb.append( "\n  Error: " ).append( propertyAsString( value, "error" ) )
+					.append( "\n" );
```
yrodiere: Either indentation is wrong here or we should prefix the call with "sb".

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
Sanne: ouch that's embarassing! I'm starting to become as grammar aware as native brits :P
Thanks

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
Sanne: woah, thanks for spotting that. No I didn't try it.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
Sanne: Thanks you nailed it. I was deep in the internals yesterday night to figure out how to connect GSON vs HTTP client and somewhat got convinced internals were going to be needed.

Looks much better after this change!

Sanne:
> I'm a bit torn about merging it before we established a baseline for our performance tests :) But I guess we could always create a branch from a commit "before performance improvements" and continue tuning our performance tests there.

For investigations like that, my preference is to use "git revert". Makes it easy to temporarily undo some improvement so that you can compare figures even with updated benchmarks.

Either way, this specific work has a great impact on GC pressure and potential GC induced pauses but the main priority to fix throughput is your work - I intentionally stay clear of the batching and async operations to avoid conflicts with you. I simply expect this to have paved the road to measure the other areas with less swings in the numbers ;)

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
Sanne: If I have to find a reason: "because it was 1am" :P Thanks!

Sanne:
I've updated the PR to address all minor concerns.
Still need to fix & test the AWS integration.

Sanne:
Waiting for CI build http://ci.hibernate.org/view/Search/job/hibernate-search-elasticsearchAWS-customBranch/1/

Sanne:
Closing this for now: the digest signature requirement will need a different design

Sanne:
Replaced by PR #1490

PR #1486, opened 2017-07-12
HSEARCH-2816 Make sure Elasticsearch performance tests apply update/delete works to existing documents and add works to non-existing documents

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2816


integrationtest/performance/engine-elasticsearch/src/main/java/org/hibernate/search/engineperformance/elasticsearch/ChangesetGenerator.java
```diff
@@ -0,0 +1,191 @@
...
+		int updatesShuffledIdsOffset = deletesShuffledIdsOffset + addDeletesPerInvocation;
+		for ( int i = 0 ; i < addDeletesPerInvocation ; ++i ) {
+			toAdd.set( shuffledIds.get( i ) );
+		}
+		for ( int i = deletesShuffledIdsOffset ; i < updatesShuffledIdsOffset ; ++i ) {
+			toDelete.set( shuffledIds.get( i ) );
+		}
+		for ( int i = updatesShuffledIdsOffset ; i < threadIdIntervalSize ; ++i ) {
+			toUpdate.set( shuffledIds.get( i ) );
+		}
```
Sanne: I'm not sure I understand the goal here. How do you guarantee that `toAdd` and `toDelete` are disjoint sets ?

yrodiere: shuffledIds only contains unique indices, see how it's generated

Sanne: oh, right. Looks good.

Sanne:
merged this

PR #1488, opened 2017-07-13
[PREVIEW] HSEARCH-2764 Improve orchestration of Elasticsearch works

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2764

Here is what I did so far.

@Sanne I know some things are still unacceptable (like the fact that non-stream async works may be processed out of order), and I will work on those. Also, I should work a bit on splitting the last commit into smaller commits, because right now the purpose of my changes is far from clear.
The thing is, I already spent a great deal of time on this, and while I'm confident I can ultimately submit a solution matching most requirements, I'd really like your input before I continue down this path.
Specifically: when working at the sources at the HEAD of this branch (don't bother with the history), and looking at how orchestration works (see ElasticsearchWorkProcessor.java), do you think that technically, the solution would acceptable, or could become acceptable with some adjustments?

Thanks a lot...


yrodiere:
Closing, I'll do a proper PR later

PR #1489, opened 2017-07-14
HSEARCH-2822 Automatic activation of the JBoss Logging AP on IDEA pro…

Sanne:
…ject import

https://hibernate.atlassian.net/browse/HSEARCH-2822


Sanne:
@DavideD I suspect this trick would be useful for OGM too..

yrodiere:
Nice, it works :) Merged, thanks!

PR #1490, opened 2017-07-15
HSEARCH-2818 Optimise encoding of GSON elements into HTTP client buffers

Sanne:
Similar optimisations as proposed before but using a different approach, now more carefully taking into account the AWS digest requests.

Also taking into account flow control of the underlying buffers, this was a bug in the previous proposal.

https://hibernate.atlassian.net/browse/HSEARCH-2818

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
@@ -29,9 +29,9 @@ import org.apache.http.nio.ContentEncoder;
 import org.apache.http.nio.IOControl;
 import org.apache.http.nio.entity.HttpAsyncContentProducer;
 import org.apache.http.protocol.HTTP;
+import org.hibernate.search.exception.SearchException;
 
 import org.hibernate.search.elasticsearch.spi.DigestSelfSigningCapable;
-import org.hibernate.search.exception.SearchException;
 
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
@@ -54,7 +54,13 @@ import com.google.gson.JsonObject;
  * be written when the flow control is re-enabled.
  *
  * A side effect of this strategy is that the total content length which
- * is being produced is not known in advance.
+ * is being produced is not known in advance. Not reporting the length
+ * in advance to the Apache Http client causes it to use chunked-encoding,
+ * which is great for large blocks but not optimal for small messages.
+ * For this reason we attempt to start encoding into a small buffer
+ * upfront: if all data we need to produce fits into that then we can
+ * report the content length; if not the encoding completion will be deferred
+ * but not resetting so to avoid repeating encoding work.
  *
  * @author Sanne Grinovero (C) 2017 Red Hat Inc.
  */
@@ -82,11 +88,13 @@ public final class GsonHttpEntity implements HttpEntity, HttpAsyncContentProduce
 	private final List<JsonObject> bodyParts;
 
 	/**
-	 * We don't want to compute the length in advance as it would defeat all optimisations.
+	 * We don't want to compute the length in advance as it would defeat the optimisations
+	 * for large bulks.
 	 * Still it's possible that we happen to find out, for example if a Digest from all
-	 * content needs to be computed.
+	 * content needs to be computed, or if the content is small enough as we attempt
+	 * to serialise at least one page.
 	 */
-	private long contentLength = -1l;
+	private long contentLength;
 
 	/**
 	 * We can lazily compute the contentLenght, but we need to avoid changing the value
@@ -127,6 +135,7 @@ public final class GsonHttpEntity implements HttpEntity, HttpAsyncContentProduce
 		Objects.requireNonNull( bodyParts );
 		this.gson = gson;
 		this.bodyParts = bodyParts;
+		this.contentLength = attemptOnePassEncoding();
 	}
 
 	@Override
@@ -185,6 +194,55 @@ public final class GsonHttpEntity implements HttpEntity, HttpAsyncContentProduce
 		//Nothing to close
 	}
 
+	/**
+	 * Let's see if we can fully encode the content without ever needing to rotate
+	 * output buffers. This will allow us to keep the memory consumption reasonable
+	 * while also being able to hint the client about the {@link #getContentLength()}.
+	 * Incidentally, having this information would avoid chunked output encoding
+	 * which is ideal precisely for small messages which can fit into a single buffer.
+	 * @return the size of the content, if all was written, or -1.
+	 */
+	private long attemptOnePassEncoding() {
+		// Essentially attempt to use the writer without going NPE on the output sink
+		// as it's not set yet.
+		try {
+			triggerFullWrite();
+		}
+		catch (IOException e) {
+			// Unlikely: there's no output buffer yet!
+			throw new SearchException( e );
+		}
+		if ( writer.flowControlPushingBack == false ) {
+			//Can't flip the buffer yet but the position is final,
+			//as we know the entire content has been rendered already.
+			return writer.availableBuffer.position();
+		}
+		else {
+			return -1l;
+		}
+	}
+
+	/**
+	 * Higher level write loop. It will start writing the JSON objects
+	 * from either the  beginning or the next object which wasn't written yet
+	 * but simply stop and return as soon as the sink can't accept more data.
+	 * Checking state of writer.flowControlPushingBack will reveal if everything
+	 * was written.
+	 * @throws IOException
+	 */
+	private void triggerFullWrite() throws IOException {
+		while ( nextBodyToEncodeIndex < bodyParts.size() ) {
+			JsonObject bodyPart = bodyParts.get( nextBodyToEncodeIndex++ );
+			gson.toJson( bodyPart, bufferedWriter );
+			bufferedWriter.append( '\n' );
+			bufferedWriter.flush();
+			if ( writer.flowControlPushingBack ) {
+				//Just quit: return control to the caller and trust we'll be called again.
+				return;
+			}
+		}
+	}
+
 	@Override
 	public void produceContent(ContentEncoder encoder, IOControl ioctrl) throws IOException {
 		Objects.requireNonNull( encoder );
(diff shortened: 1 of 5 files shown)


elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/GsonStreamedEncodingTest.java
```diff
@@ -0,0 +1,326 @@
...
+			final byte[] productionTwo = produceContentWithCustomEncoder( entity );
+			assertArrayEquals( productionOne, productionTwo );
+		}
+		catch (IOException e) {
+			throw new RuntimeException( "We're mocking IO operations, this should not happen?", e );
+		}
+	}
+
+	@Test
+	public void testDigestToTriggerLenghtComputation() {
```
yrodiere: typo here, should be `testDigestToTriggerLengthComputation`

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/DigestSelfSigningCapable.java
```diff
@@ -0,0 +1,22 @@
...
+import java.io.IOException;
+import java.security.MessageDigest;
+
+/**
+ * In some cases we can have more efficient code by delegating
+ * the computation of digests to the container object.
+ *
+ * @author Sanne Grinovero (C) 2017 Red Hat Inc.
+ */
+public interface DigestSelfSigningCapable {
```
yrodiere: See also my comment on `GsonHttpEntity#writeTo`: we may not need this interface at all.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
```diff
@@ -0,0 +1,335 @@
...
+		}
+
+		public void insertNewline() throws IOException {
+			writeOrBufferWrite( ByteBuffer.wrap( NEWLINE ) );
+		}
+
+		private void writeOrBufferWrite(ByteBuffer toWriteBytes) throws IOException {
+			totalWrittenBytes += toWriteBytes.remaining();
+			flushPendingBuffers();
+			if ( flowControlPushingBack == false ) {
```
Sanne: It's not as readable as an explicit `== false` :)

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	@Override
+	public void consumeContent() throws IOException {
+		//not used (and deprecated)
+	}
+
+	@Override
+	public void close() throws IOException {
+		//Nothing to close
```
Sanne: that's not possible: the async http client expects to be able to call us multiple times and resume the data stream from the last time, so I can't just reset for the sake of it.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
```diff
@@ -0,0 +1,335 @@
...
+		//This could be implemented but would be sub-optimal compared to using produceContent().
+		//We therefore prefer throwing the exception so that we can easily spot unintended usage via tests.
+		throw new UnsupportedOperationException( "Not implemented! Expected to produce content only over produceContent()" );
+	}
+
+	@Override
+	public void writeTo(OutputStream outstream) throws IOException {
+		//This could be implemented but would be sub-optimal compared to using produceContent().
+		//We therefore prefer throwing the exception so that we can easily spot unintended usage via tests.
+		throw new UnsupportedOperationException( "Not implemented! Expected to produce content only over produceContent()" );
```
Sanne: I'll not do it as the code is already very complex, and this method isn't even ever called so not worth the effort..

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
```diff
@@ -0,0 +1,335 @@
...
+
+	private static final BasicHeader CONTENT_TYPE = new BasicHeader( HTTP.CONTENT_TYPE, ContentType.APPLICATION_JSON.toString() );
+
+	private final Gson gson;
+	private final List<JsonObject> bodyParts;
+
+	/**
+	 * We don't want to compute the length in advance as it would defeat all optimisations.
+	 * Still it's possible that we happen to find out, for example if a Digest from all
+	 * content needs to be computed.
```
Sanne: This was actually buggy and I figured it out later: once the content length is provided we need to stick to it or the http client driver gets seriously confused. The new design is better and guarantees a "stable" response w/o drawbacks.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/DigestSelfSigningCapable.java
Sanne: having the explicit hint that this class is being invoked for the purpose of the digest computation helps a lot in making it efficient..

Could make it SPI but IMO it's borderline as something we'd not support others to use / rely on.
Is any invocation "across _our_ modules" needing to be an SPI ?

Sanne:
I've updated it, fixed the typos and applied several more optimisations.

The previous version would avoid all allocations "out of TLAB" - which are the most expensive.
After the second round, we don't see any allocations within TLAB either - the only remaining significant allocations are performed during response parsing.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
```diff
@@ -192,81 +235,133 @@ public void fillDigest(MessageDigest digest) throws IOException {
...
 
-		public void insertNewline() throws IOException {
-			writeOrBufferWrite( ByteBuffer.wrap( NEWLINE ) );
+		private void storeForLater(CharBuffer toWriteLater) {
+			if ( unwrittenLazyBuffers == null ) {
+				unwrittenLazyBuffers = new ArrayList<>( 16 );
+			}
+			//The original CharBuffer is not safe as it's backed by a char array
+			//which happens to be the GSON write buffer: it will get mutated.
+			CharBuffer defensiveCopy = CharBuffer.allocate( toWriteLater.remaining() );
```
yrodiere: It may be silly, but... since you have to perform an allocation anyway, why not directly allocate ByteBuffers, and directly encode the characters in there? You could then remove this list of char buffers, and use a list of ByteBuffers instead. This would mean turning "needsWritingBuffer" into a list, if you will: it would initially only contain one buffer, and would grow as necessary.

I tested it a bit with the performance tests, and it seems having a lot of "unwritten lazy buffers" is not rare; I was above the 16 buffer initial capacity about 25% of the time, I think, going all the way up to more than 50 buffers. So it could be useful.
But more importantly, I wonder if it couldn't make the code a bit simpler. Do you think it could?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
```diff
@@ -192,81 +235,133 @@ public void fillDigest(MessageDigest digest) throws IOException {
 			digestWriter.insertNewline();
 		}
 		//Now we finally know the content size in bytes:
-		this.contentLength = digestWriter.getContentLenght();
+		hintContentLenght( digestWriter.getContentLenght() );
+	}
+
+	private void hintContentLenght(long contentLenght) {
```
yrodiere: Typo: "hintContentLenght" => "hintContentLength", "contentLenght" => "contentLength"

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
```diff
@@ -262,6 +316,15 @@ private void hintContentLenght(long contentLenght) {
...
 		boolean flowControlPushingBack = false;
 
+		//Accumulate the content length while it's being produced so we can hint
+		//other components.
+		//This is only used for the case in which we can encode the full message
+		//without ever rotating the active buffer (in one pass), so that we
+		//can calculate the accurate size upfront.
+		//Computing this size after multiple buffers is possible but not helpful,
+		//as we'll already be using the chunked encoding mode.
+		long totalWrittenBytes = 0;
```
Sanne: thanks, I'll remove it. It's just a leftover from a previous approach which didn't work out.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/DigestSelfSigningCapable.java
Sanne: I've tried such things before. It didn't work out for various reasons.

1. the `writeTo(Outpustream)` is invoked in a different context by the http client and you can't do flow control in that case - not without buffering which is what I'm working to avoid.

> We could use a wrapper around the OutputStream to detect the content length if necessary.

That's too late. It's going to need the content lenght before the content is being produced, as the content lenght needs to be written in the output stream first thing, and also it might affect how the rest of the content is actually encoded (chunked, buffered, etc..).

Also you can't report a different lenght after you've produced the content as that gets you in trouble (bug of the client?).

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
Sanne: BTW I don't want to buffer w/o boundarier. I'm using a buffer of a very specific size which I know is a good balance for our use case.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
Sanne: I'm not sure I understand what you'd hope to achieve. I'd love to simplify the code a bit but not at cost of performance, and I'm not sure if you've considered all aspect .. for example that you actually don't know how large the bytebuffers need to be as that depends on the encoding details, and guessing gets you into re-sizing trouble.

I need to wrap this up, let's apply this as-is but then if you want to play with it and show me I'm happy to review some follow-up polishing PR.

Sanne:
I've rebased it and fixed the problems with missing `static final` and what you highlighted about close() not guaranteeing to make it repeatable.

Also promoted that disgest interface to SPI as suggested.

I didn't change the essential strategy around buffers as I'm not seeing a clearly better idea. Happy to be proven wrong but let's split that out as a different PR if you think it's worth trying.

yrodiere:
Merged, thanks! Good to see this one in :)
I'll try to submit a follow-up PR to show what I was talking about.

PR #1492, opened 2017-07-21
HSEARCH-2830 Extend Elasticsearch performance tests for better orchestration testing

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2830


yrodiere:
Actually, it seems the auxiliary counters are artificially multiplied by 3, probably due to a bug in JMH: http://mail.openjdk.java.net/pipermail/jmh-dev/2017-July/002598.html
I'll try to work it around...

Sanne:
Nice, merged!

PR #1494, opened 2017-07-25
HSEARCH-2801 QueryDescriptor requires a List<CustomTypeParameter> where a Set or Map would make more sense

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2801

### DIFF
--- a/engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
+++ b/engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
@@ -9,10 +9,7 @@ package org.hibernate.search.query.engine.impl;
 import static org.hibernate.search.util.impl.CollectionHelper.newHashMap;
 
 import java.io.Serializable;
-import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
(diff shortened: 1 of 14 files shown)


engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
```diff
@@ -130,7 +122,15 @@ private AbstractHSQuery(ExtendedSearchIntegrator extendedIntegrator, IndexedType
 				throw LOG.targetedEntityTypesNotIndexed( StringHelper.join( targetedEntities, "," ) );
 			}
 		}
-		this.customTypeMetadata = customTypeMetadata == null ? Collections.emptyList() : new ArrayList<>( customTypeMetadata );
+		if ( customTypeMetadata == null ) {
+			this.customTypeMetadata = IndexedTypeMaps.empty();
+		}
+		else {
+			this.customTypeMetadata = IndexedTypeMaps.hashMap();
+			for ( Map.Entry<IndexedTypeIdentifier, CustomTypeMetadata> entry : customTypeMetadata.entrySet() ) {
```
Sanne: looks like you're making a defensive copy here. Why is that?

The reason the modifiers on the `IndexedTypeMap` are deprecated is to eventually make the contract explicitly immutable - granted this is not done yet so we still need to be careful but I think we should work into that direction.

Sanne: I've tested and removed it. No point in waiting for this ;)

Sanne:
merged. Thanks! Nice to see the cleanup making progress..

engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
yrodiere: It was mainly because `customTypeMetadata` is basically "user" input, so I can't trust it not to change. As you said, we want to make the map immutable _eventually_, but we're not quite there yet :)
Anyway, ok to remove the copy, it's not that critical.

PR #1495, opened 2017-07-25
HSEARCH-2823 The elasticsearch-aws JAR and its dependencies are missing from the .tar.gz/zip distributions

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2823


Sanne:
great catch! Merged

PR #1498, opened 2017-07-26
HSEARCH-2640 Improve resilience of cleanup methods (close, etc.)

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2640

This isn't strictly necessary in 5.8, but I think it's an improvement, and I'll need the `Closer` class to fix an issue in our JSR-352 integration (https://hibernate.atlassian.net/browse/HSEARCH-2673).


Sanne:
extreme measures :)

yrodiere:
Well... I can't think of another way of making sure we properly close all resources... There's the fake try-with-resources pattern we applied in some Elasticsearch components, but it doesn't work anymore once you have to close a collection of resources.
So I'm not sure this is an extreme solution, it's just the only one I could think of. Though I agree it may be a bit extreme to consider the problem worth solving to begin with.

Sanne:
merged

PR #1499, opened 2017-07-26
HSEARCH-2833 Avoid confusing variable uses (bad naming, no declaration) in the documentation

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2833


Sanne:
nice! merged

PR #1491, opened 2017-07-18
HSEARCH-2827 + HSEARCH-2828 : WorkerTestCase improvements

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2827
https://hibernate.atlassian.net/browse/HSEARCH-2828

PR #1500, opened 2017-07-28
HSEARCH-2764 Improve orchestration of Elasticsearch works

yrodiere:
~~This PR is based on #1491 (HSEARCH-2827 and HSEARCH-2828), which should be merged first.
We should also probably merge #1492 (HSEARCH-2830) before this one.~~ => Done.

Relevant JIRA ticket: https://hibernate.atlassian.net//browse/HSEARCH-2764

So, that's a big one... Sorry about that, but once again most of the changes are linked. I tried to organize the commits in such a way that they are easier to understand.

# Purpose

## Performance improvements

First, if you're dreaming about dramatic performance improvements, forget about it. Currently, Hibernate Search spends most of its time waiting for Elasticsearch to process works. And I mean this literally: threads are usually parked or blocked on a monitor, waiting for an Elasticsearch response, more than 90% of the time.
So improving raw throughput in Hibernate Search, while possible, likely won't have much of an effect on overall throughput (when we take Elasticsearch into account).
In order to improve performance, I tried to be a bit smarter about what we ask Elasticsearch to do, and how we ask Elasticsearch to do it:

 * Avoid performing unnecessary works (like unnecessary refreshes)
 * Minimize the impact of network latency (minimize the number of round-trips between Hibernate Search and the Elasticsearch cluster)
 * Leverage the distributed nature of an Elasticsearch cluster (send more unrelated works in parallel)

The last one, in particular, is made much easier by embracing the reactive nature of the underlying HTTP client: by submitting a response handler along with each request, instead of submitting a request and waiting for it to finish, we can achieve a high level of parallelism with a small number of threads. That's what the commits introducing CompletableFuture everywhere are about: making Elasticsearch work processing reactive.

## Technical improvements

Along the way, I had to tackle some technical issues, mainly because if I didn't, the performance tests would be skewed, or would simply crash. Both of the changes are about `AsyncRequestProcessor` (now `BatchingSharedElasticsearchWorkOrchestrator`).

1. I discovered a race condition in which a thread waiting for async works to be finished would be released too early. See commit "HSEARCH-2764 Avoid a very rare race condition in async work processing ...".
2. I switched the `MultiWriteDrainableLinkedList` for a safer, bounded `ArrayBlockingQueue`. `MultiWriteDrainableLinkedList` being unbounded means we get very quickly an `OutOfMemoryException` when works are submitted faster than they are consumed. And this is, in fact, an issue that's been reported before: https://forum.hibernate.org/viewtopic.php?f=9&t=1044534 . Note we don't have this issue in the Lucene backend because in that case we only use it for synchronous execution, while with Elasticsearch we use it for asynchronous execution, meaning the submitting threads are released after having submitted a work and can submit more work before the first even finished executing. See "HSEARCH-2764 Limit the size of Elasticsearch work queues"

## Functional improvements

A side effect of one of this PR is that synchronously-executed works from different threads, but relating to the same documents, are no longer executed out of order. That's because we now submit synchronous works for each index to a single queue, which is then processed serially.

Obviously it's not something that will improve performance per se, since we previously executed these works in parallel.
But on the other hand, this change allows us to skip the `refresh=true` parameter in bulk requests when `refreshAfterWrite` is disabled.
So, all in all... I figured we may as well give this change a try.

See commit "HSEARCH-2764 Make Elasticsearch non-stream orchestrators index-specific".

# Results

## Complexity

The overall architecture of Elasticsearch work processing is arguably more complex after this patch. I think that's counterbalanced by more thorough testing and more clearly separated responsibilities, but I'll let you be the judge of that.

## Performance tests results

By running the performance tests we merged previously (with a few improvements, see #1492) before and after this patch, I got some numbers: https://docs.google.com/spreadsheets/d/14pLNWIewrCNr_pj_rzMyq6UTMf7lf9VBHMjY8b_gapU/edit#gid=2063776958

The interesting sheet is "nonbuggy-2017-07-28-2". Here is how to read it:

* column A is the client mode:
  * `blackhole-5.4.0` means a stub Elasticsearch 5.4 server processing requests in almost-constant time
  * `default` means an actual Elasticsearch 5.3 server hosted on AWS (a c4-large instance)
* column B is the benchmark: 
  * `Stream.write` is about `IndexManager.performStreamOperation`
  * `NonStream.concurrentReadWriteTest:readWriteTestWriter` is about `IndexManager.performOperation` with concurrent search queries
  * `NonStream.write` is about `IndexManager.performOperation` without concurrent search queries
  * `NonStream.queryBooksByBestRating` is about querying only
* column C is the worker execution mode (only relevant for non-stream writes): sync or async
* column D is the value of `refreshAfterWrite`
* columns E and F show the scores and errors for the original code (without this patch)
* columns G and H show the scores and errors for the "reactive" code (with this patch)
* columns I to J show the relative improvements with this patch. Color indicates improvements/decreases from the original build to the "reactive" build:
  * values in green are statistically significant improvements
  * values in red are statistically significant decreases
  * other values do not reveal any statistically significant difference
* columns K to N show the scores, errors, and improvements when applying this patch *and*, on top of that, using parallel processing everywhere (even for non-stream work). This means we run the risk of conflicts between add/updates and deletes on the same document, so it's not an option, but it shows us the cost of serial processing.
* columns O to R show the scores, errors, and improvements when applying this patch *and*, on top of that, using serial processing everywhere (even for stream work). It shows the benefit of parallel processing in that case.

Note that due to what's probably a bug in JHM, [the auxiliary counters are skewed](http://mail.openjdk.java.net/pipermail/jmh-dev/2017-July/002598.html). So please don't use the `:changeset` and `:add` counters when analyzing the results.

#### "query" method in concurrent tests

For concurrent benchmarks, the "query" threads and the "write" threads are competing for one resource: the Elasticsearch server in "default" mode, and the CPU in "black hole" mode.

Therefore, looking at the performance of the "query" benchmark is not very conclusive, because this performance will automatically degrade when write performance improves, and will automatically improve when write performance decreases.

#### Blackhole mode

In blackhole mode, we either have seemingly huge improvements or inconclusive results. I'd argue that the huge improvements are not conclusive either, because they mostly result from the last commit in this PR, where we remove the artificial 100ms delay between two async processing batches.

I think the only conclusion to draw from blackhole tests is that, be it before or after this PR, the internal processing in Hibernate Search is faster than the actual indexing by Elasticsearch by several orders of magnitude. Thus we shouldn't bother ourselves too much about raw code performance for now, at least not if our goal is to increase throughput.

### Actual Elasticsearch server

In real-world conditions (with an actual Elasticsearch server), the numbers are more conclusive, and quite encouraging.

My comments below only look at the "worst case" improvements, i.e. the difference between (score before + error) and (score after - error); thus the improvements are probably better than what I mention here.

* query throughput seems to have decreased. I cannot explain why exactly, since I barely even touched query code. Note that the numbers are quite different depending on the worker execution mode and the value of `refreshAfterWrite`, which is a bit strange since we don't perform any write during these tests. I would be inclined to attribute the changes to error margins, but we should probably work a bit on query performance in another ticket.
* non-stream throughput with `refreshAfterWrite=false`, be it sync or async, improved massively (+20% to +240%), probably because we no longer do unnecessary refreshes in bulk API calls.
* non-stream async throughput with `refreshAfterWrite=true` decreased, though only in a non-concurrent setup. This is likely a result of more frequent refreshes: while we used to do a refresh for each batch of works, we now do a refresh for every bulk API call (in this setup, 1 batch = at most 2500 changesets = at most 30000 works = approximately 250 times less bulk API calls, i.e. at most 120). We could optimize it, I would say "we don't care", since this setup is particularly pointless.
* non-stream sync throughput with `refreshAfterWrite=true` increased in the non-concurrent setup, but decreased in the concurrent setup. I don't think it's very important, because `refreshAfterWrite=true` will ruin performance anyway, but if I had to explain why, I would say:
  * the increase is caused by the fact we now bulk together requests from different threads, leading to lower average latency when the Elasticsearch server is the bottleneck
  * the decrease is probably caused by the fact we now (correctly) execute sync works serially, which must reduce the throughput when write threads compete with read threads, somehow.
* streamed write throughput improved about 20%, probably because of parallel work processing. This is confirmed by the fact the Elasticsearch server did show significantly higher CPU usage during those tests, as shown [on this CPU graph taken during three run](http://imgur.com/a/EB4kc) (the "original" run is on the left, the "patched" run on the right and slightly shorter).

### Conclusion

Performance improvements seem to be here, at least in the important cases (stream, and non-stream with `refreshAfterWrite=false`).
Let's merge it, and work on the sore points in the scope of another ticket if you think it's necessary?


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/processor/impl/ElasticsearchWorkProcessor.java
```diff
@@ -87,313 +121,157 @@ public void close() {
...
+		/*
+		 * The non-stream orchestrator requires a strict execution ordering,
+		 * because subsequent changesets may be inter-dependent (an addition in one changeset
+		 * followed by a deletion in the next, for instance).
+		 * To make sure we apply works in order, we use serial orchestrators.
+		 */
+		FlushableElasticsearchWorkOrchestrator delegate =
+				createSerialOrchestrator( contextSupplier, NON_STREAM_MIN_BULK_SIZE, refreshInBulkApiCall );
+
+		if ( sync ) {
```
Sanne: is the only difference here the message string?

Sanne:
> First, if you're dreaming about dramatic performance improvements, forget about it. Currently, Hibernate Search spends most of its time waiting for Elasticsearch to process works.

I see you might be a bit disappointed because of the figures you got, but this work might actually provide some dramatic improvements for people who have an higher-end configured ES server, or just a better tuned local network link.

> In order to improve performance, I tried to be a bit smarter about what we ask Elasticsearch to do

See, that's great and exactly what people hope to get as "free lunch" when using a library like this.

The figures we get here are just not representative of what someone else will get from it - unless they use the same cheap AWS configuration. Still that doesn't make our figures irrelevant, they are an essential factor of this whole exercise which helps us to understand and acquire the knowledge to figure out what "smarter about what we ask Elasticsearch to do" requires... today and in the future.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/processor/impl/ElasticsearchWorkProcessor.java
yrodiere: Turns out it is. There was other differences before, but I removed them along the commits.
The thing is, the orchestrators themselves work the same way whether we're in synchronous mode or not. It's just the index manager that will wait for submitted works to finish in synchronous mode, and won't wait in asynchronous mode.
I'll add a fixup commit.

yrodiere: I added two commits. We'll just have to remember to execute `git rebase -i --autosquash` before merging!

Sanne: thanks!

Sanne:
merged!

PR #1501, opened 2017-08-04
HSEARCH-2837 Clarify errors when interrupted during submission of work to the ES client

yrodiere:
This PR is based on #1500 , which should be merged first.

Relevant ticket: https://hibernate.atlassian.net//browse/HSEARCH-2837

~~@Sanne As mentioned on HipChat, the interruptions you experienced were probably generated by JMH, not by Hibernate Search: the threads that got interrupted were not part of a pool owned by Hibernate Search, so Hibernate Search had no reason to even try interrupting them.~~

~~Thus I think that in your case, throwing an error is the right choice: we need to inform the client that his attempt at submitting a new changeset failed, since it failed for a reason that is outside of our control.~~
 
=> As discussed together, I removed the cause in the SearchException, which should make the stack trace a bit smaller.

We do have issues with overly verbose error messages in threads owned by Hibernate Search, but that's when processing changesets, not when submitting them (https://hibernate.atlassian.net/browse/HSEARCH-2832). However, errors during processing are usually passed to the ErrorHandler, so I'm not sure bypassing the ErrorHandler and logging a warning would be the best solution. Please comment on HSEARCH-2832 if you have an opinion about that issue: I personally don't know what to do about it...


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/logging/impl/Log.java
```diff
@@ -489,4 +489,9 @@ SearchException invalidAnalysisFactoryParameter(@FormatWith(ClassFormatter.class
 	@Message(id = ES_BACKEND_MESSAGES_START_ID + 90,
 			value = "Elasticsearch response indicates a failure." )
 	SearchException elasticsearchResponseIndicatesFailure();
+
+	@Message(id = ES_BACKEND_MESSAGES_START_ID + 91,
+			value = "The thread was interrupted while a changeset was being submitted to '%1$s'."
+					+ " The changeset has been ignored." )
```
Sanne: I'd like to replace "ignored" with "discarded". would you agree?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/processor/impl/BatchingSharedElasticsearchWorkOrchestrator.java
```diff
@@ -49,6 +51,8 @@
 	private final BlockingQueue<Changeset> changesetQueue;
 	private final List<Changeset> changesetBuffer;
 	private final AtomicBoolean processingScheduled;
+	private boolean open;
```
Sanne: I'll add a coment for the _open_ field to point out it's guarded by the shutdownLock

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/logging/impl/Log.java
yrodiere: Sure!

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/processor/impl/BatchingSharedElasticsearchWorkOrchestrator.java
yrodiere: Sure!

Sanne:
merged. Thanks!

PR #1502, opened 2017-08-07
HSEARCH-2838 Add a maxConnection parameter to Elasticsearch performance tests

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2838


Sanne:
merged

PR #1506, opened 2017-08-10
HSEARCH-2843 Changing the limit/offset of a query does not clear the cached results

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2843


Sanne:
merged

PR #1508, opened 2017-08-11
HSEARCH-2845 Add dependency management entries for the dependencypath plugin and jsonassert

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2845


Sanne:
merged, thanks!

PR #1509, opened 2017-08-11
HSEARCH-2842 Verify and document interaction of full-text queries with an ORM filter

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2842

### DIFF
--- a/documentation/src/main/asciidoc/query.asciidoc
+++ b/documentation/src/main/asciidoc/query.asciidoc
@@ -1486,8 +1486,26 @@ for (Object[] result : results) {
 Be careful, building the explanation object is quite expensive, it is roughly as expensive as
 running the Lucene query again. Don't do it if you don't need the object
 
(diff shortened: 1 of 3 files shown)


documentation/src/main/asciidoc/query.asciidoc
```diff
@@ -1486,8 +1486,26 @@ for (Object[] result : results) {
 Be careful, building the explanation object is quite expensive, it is roughly as expensive as
 running the Lucene query again. Don't do it if you don't need the object
 
-[[query-filter]]
-=== Filters
+=== [[query-filter]] Filters
+
+==== [[query-filter-orm]] Hibernate ORM filters
```
Sanne: I'll move this section as the last one of the `Filters` as it's the least interesting one. Also nice to not break current chapter numbering.

Sanne:
merged (some days ago!)

PR #1510, opened 2017-08-11
HSEARCH-2818 Remove the need to allocate char buffers in GsonHttpEntity

yrodiere:
Follow-up on #1490 .

After this patch, we only use byte buffers to store data that couldn't be flushed (as many buffers as needed), whereas we previously used a mix of byte buffers and char buffers (2 byte buffers per request + as many char buffers as needed).
Functionally, this is identical to the previous implementation.
Performance-wise, there might be a small gain, but I doubt it can be significant.

All in all, the only reason for this patch is to make the code simpler. I think it succeeds, but that's my opinion.

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/AutoGrowingPagedBuffer.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/AutoGrowingPagedBuffer.java
@@ -1,0 +1,188 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.elasticsearch.util.impl;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.CharacterCodingException;
+import java.nio.charset.CharsetEncoder;
+import java.nio.charset.CoderResult;
+import java.util.Deque;
+import java.util.Iterator;
+import java.util.LinkedList;
+
+import org.apache.http.nio.ContentEncoder;
+
+/**
+ * An automatically growing, paged buffer.
+ * <p>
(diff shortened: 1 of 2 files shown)


yrodiere:
Note that the last commit is a change of mind. I tried to replicate the previous strategy of swapping between two buffers (one "available" and one "needing write"), but it just felt like I was taking up memory for no good reason... Another solution would be to limit the size of "reuseablePages" to some small number (like 2), but I still feel like this is overkill.

Sanne:
applied the first commit as it's straight forward.
I don't agree on the second commit, it's not making it clearer to initialize fields by side-effect..

I'll need to spend some more time on the remaining commits to understand the benefit; I'll admit I'm slightly biased towards not liking it because of the splitting up complex logic between two different yet tightly coupled classes. I guess you might argue that they are independent types, but all reasons for those detailed methods are driven by that one single client class?

yrodiere:
>  I guess you might argue that they are independent types, but all reasons for those detailed methods are driven by that one single client class?

I think the sore point is the `put` method in `AutoGrowingPagedBuffer`, which effectively introduce writing responsibilities in a class that should only be about buffering. I'll rework that.

yrodiere:
> I think the sore point is the put method in AutoGrowingPagedBuffer, which effectively introduces writing responsibilities in a class that should only be about buffering. I'll rework that.

Ok, I had a look and it's not a good idea, basically it would require breaking encapsulation and would introduce other problems, such as the possibility that pages are tampered with *after* being put to the `needWriting` queue. Of course I could introduce new interfaces for callbacks accepted by the buffer to add/extract content to/from it, making it clearer that buffers are only to be used within a very limited scope, but at this point it would just be overkill.

> I guess you might argue that they are independent types, but all reasons for those detailed methods are driven by that one single client class?

Yes, buffering in this exact fashion is only needed in a single client class, but the details of how this buffering is implemented are of no importance to this client class: it just *needs* buffering. Its main concern is to orchestrate writing (redirect writes to the buffer, and consume the buffer as soon as possible). So why add the complexity of managing the buffer pages to `ProgressiveCharBufferWriter`?

yrodiere:
Closing, I'll submit these commits again soon in a PR for https://hibernate.atlassian.net/browse/HSEARCH-2886

PR #1511, opened 2017-08-11
HSEARCH-2818 Use only standard interfaces to implement digest generation for Elasticsearch requests

yrodiere:
Follow-up on #1490 .

After this patch, additional features such as AWS request signing do not need to use a specific interface to access the content of the request and create a digest from it: they can simply use the `org.apache.http.HttpEntity.writeTo(OutputStream)` method.
One advantage is that other implementors could use the same method to do anything with the request's content, beyond just populating a MessageDigest.

Proof it works on AWS: http://ci.hibernate.org/view/Personal%20runs/job/hibernate-search-master-elasticsearchAWS-yoann/4/


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
```diff
@@ -169,14 +169,27 @@ public Header getContentEncoding() {
...
+				+ " or writeTo(OutputStream) if blocking calls are acceptable for your use case." );
 	}
 
 	@Override
-	public void writeTo(OutputStream outstream) throws IOException {
-		//This could be implemented but would be sub-optimal compared to using produceContent().
-		//We therefore prefer throwing the exception so that we can easily spot unintended usage via tests.
-		throw new UnsupportedOperationException( "Not implemented! Expected to produce content only over produceContent()" );
+	public void writeTo(OutputStream out) throws IOException {
+		/*
```
Sanne: A nasty side-effect of implementing this method is that we won't notice if we actually start using this as primary serialization mechanism, so doing introducing a significant performance regression.
I'd prefer to throw an exception.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
```diff
@@ -169,14 +169,27 @@ public Header getContentEncoding() {
...
+		 */
+		CountingOutputStream countingStream = new CountingOutputStream( out );
+		Writer writer = new OutputStreamWriter( countingStream, CHARSET );
+		for ( JsonObject bodyPart : bodyParts ) {
+			gson.toJson( bodyPart, writer );
+			writer.append( '\n' );
+		}
+		writer.flush();
+		//Now we finally know the content size in bytes:
+		hintContentLenght( countingStream.getBytesWritten() );
```
Sanne: aren't you confusing the size in chars from the size in bytes? I'm not seeing the right encoding being applied.

yrodiere: The code uses the number of bytes as a content length even before this patch.
As for the encoding, I'll have a look.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
yrodiere: Hum, yes, there's a risk of the entity being misused. Not sure it's a big one, but it exist nonetheless.
However, implementing this method was the point of this whole PR. If we decide we want to throw an exception here, it means this PR is rejected.
Since you marked the PR with the "Changes required" label, I guess your message was only meant as an observation?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
yrodiere: So after having dig up a bit, I discovered this content length "hinting" when computing the hash is unfortunately useless. The `getContentLength()` method is always called by `org.apache.http.protocol.RequestContent` before the AWS interceptor is called, so we always freeze the content length before we compute the hash.
The only reason you can sometime witness chunk encoding being used is that we try to compute the size in the GsonEntity constructor if it's not too high.
I'll open a ticket about that, but it's unrelated to this PR.

yrodiere:
Rebased and answered your comments.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/util/impl/GsonHttpEntity.java
yrodiere: The ticket: https://hibernate.atlassian.net/browse/HSEARCH-2849

Sanne:
@yrodiere Is this PR still relevant after your other PRs ?

yrodiere:
@Sanne It needs some rebasing, but yes it is. The purpose was rather orthogonal to other PRs'.

yrodiere:
Rebased on master

Sanne:
Commit re-worded as https://hibernate.atlassian.net/browse/HSEARCH-2913 then merged!

Thanks

yrodiere:
@Sanne Thanks!

Sanne:
thanks to you for your patience ;)

PR #1512, opened 2017-08-16
HSEARCH-2850 options to run WildFly TS against prepared distribution …

jmartisk:
…and built-in modules

https://hibernate.atlassian.net/browse/HSEARCH-2850

So far this is for 5.5 only (where it is needed for EAP testing), if accepted, I will do the same for master and other branches as needed.

### DIFF
--- a/integrationtest/wildfly/README.md
+++ b/integrationtest/wildfly/README.md
@@ -1,0 +1,8 @@
+## Explanation of testing profiles
+### Obtaining WildFly distribution
+- The default is that the TS obtains WildFly from Maven and prepares it automatically.
+- To test against WildFly distributions which you prepared yourself, just pass the `-DjbossHome.node1` and `-DjbossHome.node2` properties containing paths to the distributions
+
+### Testing included modules vs. preparing modules 
+- The default mode is that the TS unzips Hibernate Search and required libraries as WildFly modules into the distributions
+- If you want to test against the modules included in the official WildFly distributions, pass the property `-DuseBuiltinModules`. In this case, you should also specify the expected Hibernate Search version (the one included in the WildFly distribution), using the `-DbuiltinVersion` property. If not specified, the test suite will expect that the included version is the same as the one you are running the test suite from. 
--- a/integrationtest/wildfly/pom.xml
+++ b/integrationtest/wildfly/pom.xml
@@ -30,6 +30,7 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <byteman.agent.path>${org.jboss.byteman:byteman:jar}</byteman.agent.path>
         <byteman.script.path>${basedir}${file.separator}src${file.separator}test${file.separator}resources${file.separator}disablejpadapters.btm</byteman.script.path>
+        <test.hibernate.search.module.slot>main</test.hibernate.search.module.slot>
     </properties>
 
     <dependencyManagement>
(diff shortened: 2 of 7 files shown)


Sanne:
@Hibernate-CI add to whitelist

@jmartisk thanks!

integrationtest/wildfly/README.md
```diff
@@ -0,0 +1,8 @@
+## Explanation of testing profiles
+### Obtaining WildFly distribution
+- The default is that the TS obtains WildFly from Maven and prepares it automatically.
+- To test against WildFly distributions which you prepared yourself, just pass the `-DjbossHome.node1` and `-DjbossHome.node2` properties containing paths to the distributions
+
+### Testing included modules vs. preparing modules 
+- The default mode is that the TS unzips Hibernate Search and required libraries as WildFly modules into the distributions
+- If you want to test against the modules included in the official WildFly distributions, pass the property `-DuseBuiltinModules`
```
yrodiere: I know we don't do this consistently, but maybe using qualified names for properties would be interesting here? Be it only to make it clear, from the property name, what it will affect?
I'd use something like `test.wildfly.node1.home`, `test.wildfly.node2.home` and `test.wildfly.useBuiltinModules`. Or maybe `it.wildfly.node1.home`, `it.wildfly.node2.home` and `it.wildfly.useBuiltinModules`, if you want to clearly separate integration tests from the rest.

integrationtest/wildfly/pom.xml
```diff
@@ -30,6 +30,7 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <byteman.agent.path>${org.jboss.byteman:byteman:jar}</byteman.agent.path>
         <byteman.script.path>${basedir}${file.separator}src${file.separator}test${file.separator}resources${file.separator}disablejpadapters.btm</byteman.script.path>
+        <hibernate.search.module.slot>main</hibernate.search.module.slot>
```
yrodiere: Careful with that property name: we use the same name in another module, not for testing but to generate module aliases (I think).
Maybe that's all the more reason to use qualified property names? `test.hibernate.search.module.slot`, or `it.hibernate.search.module.slot` here?

Sanne:
looks like it's failing integration tests?
http://ci.hibernate.org/job/hibernate-search-PR/2458/org.hibernate$hibernate-search-integrationtest-wildfly/console

integrationtest/wildfly/README.md
jmartisk: I see your point, I'm personally just not sure if it's worth the trouble  because it makes the properties quite long, so it's more work for the user to write that when running the TS from command line. But I don't really care, if that's how you guys prefer it, I can change it. What does Sanne think?

integrationtest/wildfly/pom.xml
jmartisk: You're right, thanks for noticing. Ok I'll change this, my point about the user having to write long names does not apply here anyway.

jmartisk:
Oops, yeah, I fixed that now.

jmartisk:
Btw, when the TS is run against built-in modules, it typically throws these two failures which stem from assertions about Search version, which is usually different in the distribution than what the Maven project sees:

``
ModuleMemberRegistrationIT.HibernateSearchVersion:95 expected:<5.5.[8-SNAPSHOT]> but was:<5.5.[2.Final-redhat-1]>
ModuleMemberRegistrationEarArchiveWithJbossDeploymentIT.HibernateSearchVersion
``

Surely you can fix this by running the tests from the same tag from which Search was built for that WildFly/EAP distribution, but typically you want to be able to run the latest TS against an older WF/EAP build and not get unexpected failures due to that.
What should we do with this? Should I make a change so that they get skipped when the TS is running against built-in modules?

yrodiere:
> What should we do with this? Should I make a change so that they get skipped when the TS is running against built-in modules?

I would say we override the values in this files using the new properties that you added, and add more properties to the pom so that we can override them if we need to `integrationtest/wildfly/src/test/resources/module-versions.properties`

I say "add more properties to the pom" because I'm not sure overriding `luceneVersion` (for instance) directly just for tests is a good idea, as it would impact other parts of the build. Maybe use a property such as `test.luceneVersion`, with its default value set to `luceneVersion`.

For the record, this property file is used in `org.hibernate.search.test.integration.VersionTestHelper`. Maybe it's used elsewhere, but I don't think so.

jmartisk:
You mean we create a new property containing the "expected" Search version which then the `VersionTestHelper` will then read instead of reading the value of `dependency.version.HibernateSearch`? But how would you generally determine its value? That should be done somehow automatically, but the value would have to be somehow extracted from the particular WildFly distribution you are testing against, which isn't easy. And requiring the user to pass it manually seems quite useless to me.

yrodiere:
> And requiring the user to pass it manually seems quite useless to me.

Might be, when testing against a Wildfly distribution with a single version of Hibernate Search. If it's really too much hassle, ok to skip the version checks in your particular case.

But even if we skip the version check, some tests still seem to require the dependency versions in order to execute properly: see in particular `org.hibernate.search.test.integration.jms.SearchNewEntityJmsMasterSlaveIT.LibrariesLoader` and `org.hibernate.search.test.integration.jms.transaction.TransactionalJmsMasterSlaveIT.LibrariesLoader`. Those tests are about allowing users to bundle Hibernate Search as libraries in their deployment, if I understood correctly. So in this case I don't think you would need to provide the version of Hibernate Search in your distribution, just one that is available in the maven repositories you configured.

Sanne:
I would prefer to not skip the version check: the reason they are in place is to verify that you are indeed testing the version you want.

So to make sure that we are indeed testing the version that WildFly (EAP) is including I'd suggest we check that this version is indeed the one being run.

jmartisk:
A version check makes sense, I'm just not sure how to detect the expected version. 

Options which come to my mind:
- make the TS somehow look into the `org.hibernate.search.orm:main` module to guess the included version
- make the test assert that Hibernate Search classes are loaded from a jar located in the `org.hibernate.search.orm:main` module
- require the user to specify the expected version, which, as I already said, is quite a lot of hassle

Sanne:
> require the user to specify the expected version, which, as I already said, is quite a lot of hassle

I think this would be my preference as the goal is to verify things. Why do you think it's an hassle?

Essentially I expect you to run it either giving no parameters at all - which implies running the default tests with the currently being built version - or to explicitly override it all and provide a) a custom application server b) which Search version is expected to be found in such server

I don't think automating the detection of the Search version out of a custom server build is useful. It's not future proof either as such an autodetection strategy would likely break with future versions, or possibly need to take into consideration multiple versions.

Since we run such a test [with a custom application server build] only occasionally - and possibly automated - I see no much benefit in trying to avoid a second parameter.

jmartisk:
OK, I added the `builtinVersion` property to let the user specify the expected Search version. If not specified, it will default to the version obtained from Maven project. Also updated the README to make this clear.

> Since we run such a test [with a custom application server build] only occasionally - and possibly automated

Yes, I am going to create a job for running the 5.5 branch test suite regularly against EAP 7.1 builds. There it will be relatively easy to automatically detect the version from EAP distribution, so I guess you're right it's not that such big deal.

yrodiere:
> Yes, I am going to create a job for running the 5.5 branch test suite regularly against EAP 7.1 builds. There it will be relatively easy to automatically detect the version from EAP distribution, so I guess you're right it's not that such big deal.

Ok! Just tell us when you tried this job, so that we're sure you can detect the version from EAP distributions, and I'll merge the PR.

jmartisk:
> Ok! Just tell us when you tried this job, so that we're sure you can detect the version from EAP distributions, and I'll merge the PR.

I tested this on my laptop against EAP 7.1.0.ER3 and it worked. I can't fully finish the job automation without this PR being merged, but I'm pretty confident it will work - actually my comment was a little imprecise, the job will not detect the Search version from the EAP distribution, but from metadata which we receive from the productization team when they deliver an EAP build to us.
From my point of view, this can be merged now.

yrodiere:
@jmartisk Merged, thanks! Before I mark the ticket as resolved, could you please "forward-port" this to the other branches? It's definitely needed in 5.8, but I'm not sure about 5.6 and 5.7. If you feel like those two branches will never require any testing, I guess we can forget about them.

jmartisk:
@yrodiere thanks for reviews and merging! I will now create PRs for the other branches, I'll try to do that for 5.6 and 5.7 as well

PR #1516, opened 2017-08-17
HSEARCH-2854 IllegalStateException when sending an Elasticsearch request with a content-length

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2854


Sanne:
thanks! merged

PR #1520, opened 2017-08-22
HSEARCH-2858 Using @ContainedIn in a @Spatial-annotated entity triggers a NullPointerException on startup

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2858

To be backported to 5.6 and 5.7. Maybe even 5.5, but I didn't try that one.


Sanne:
checkstyle is unhappy ;)

yrodiere:
> checkstyle is unhappy ;)

Fixed, sorry about that.

Sanne:
The fix looks reasonable but I'm not seeing the relation with the test?

I tried to run the test after reverting the fix and indeed it's still passing.

yrodiere:
@Sanne I tried to simplify the test too hard... Will correct that.

yrodiere:
@Sanne Should be okay now... The test fails without the fix, succeeds with the fix, checkstyle passes, and I added a few comments in the test to explain what's going on. Let's wait for the CI just in case...

Sanne:
thanks, understand it now and it works! merged

PR #1521, opened 2017-08-23
HSEARCH-2853 customize configuration for tests programmatically rathe…

jmartisk:
…r than using a static config file

https://hibernate.atlassian.net/browse/HSEARCH-2853

To discuss:
- Anything else needed to be configured right now apart from the org.hibernate.search logging category?
- Do we need to switch to creating JMS queues using management API instead of `activemq-jms.xml`? I don't think it's that useful to change this, but still I'm asking in case someone has a different opinion.


integrationtest/wildfly/src/test/java/org/hibernate/search/test/integration/arquillian/LoggingConfigurator.java
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Configures WildFly so that the org.hibernate.search logger exists for the DEBUG level.
+ */
+public class LoggingConfigurator {
+
+	@Inject
+	private Instance<ManagementClient> managementClient;
+
+	private Logger logger = Logger.getLogger(LoggingConfigurator.class.getName());
```
yrodiere: It's stange that checkstyle didn't notice it, but you should use spaces after opening parentheses and before closing ones.
It's rather dull to fix that manually, so you can use the codestyles for IDEA and Eclipse from there: https://github.com/hibernate/hibernate-ide-codestyles

Sanne:
oh, right we need the timeout changed. Sorry I had forgotten about it!

integrationtest/wildfly/src/test/java/org/hibernate/search/test/integration/arquillian/LoggingConfigurator.java
jmartisk: Fixed that, thanks

jmartisk:
Ok, I added a server setup task to configure the transaction timeout. IIUC a comment implied that this is only needed for `MassIndexingTimeoutIT` so I only applied it to that one test. If it's needed for more tests, I can change that.

yrodiere:
@jmartisk Merged, thanks! Do you need to backport this to 5.5, or even other branches (5.6 and 5.7)? If so, can you do it? I'll mark the ticket as resolved afterward.

jmartisk:
@yrodiere Yeah, I will backport it today or tomorrow.

PR #1522, opened 2017-08-24
HSEARCH-2850 options to run WildFly TS against prepared distribution …

jmartisk:
…and built-in modules


yrodiere:
Merged, thanks!

PR #1525, opened 2017-08-25
HSEARCH-2853 customize configuration for tests programmatically rathe…

jmartisk:
…r than using a static config file


yrodiere:
Merged, thanks!

PR #1528, opened 2017-08-29
HSEARCH-2863 CalendarBridgeProvider creates StringEncodingDateBridge

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2863

**CAUTION**: to be backported to 5.7 and 5.6 before we mark the ticket as resolved.

### DIFF
--- a/orm/src/test/java/org/hibernate/search/test/bridge/BridgeTest.java
+++ b/orm/src/test/java/org/hibernate/search/test/bridge/BridgeTest.java
@@ -39,8 +39,10 @@ import org.hibernate.search.bridge.builtin.StringEncodingCalendarBridge;
 import org.hibernate.search.query.dsl.QueryBuilder;
 import org.hibernate.search.test.SearchTestBase;
 import org.hibernate.search.testsupport.TestConstants;
+import org.hibernate.search.testsupport.junit.SkipOnElasticsearch;
 import org.hibernate.testing.SkipForDialect;
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


orm/src/test/java/org/hibernate/search/test/bridge/BridgeTest.java
```diff
@@ -359,6 +360,74 @@ public void testCalendarBridge() throws Exception {
...
+		tx.commit();
+
+		tx = s.beginTransaction();
+		FullTextSession session = Search.getFullTextSession( s );
+
+		BooleanQuery.Builder booleanQueryBuilder = new BooleanQuery.Builder();
+
+		TermQuery termQuery = new TermQuery( new Term( "dateDayStringEncoding",
+				DateTools.dateToString( date, DateTools.Resolution.DAY ) ) );
+		booleanQueryBuilder.add( termQuery, BooleanClause.Occur.MUST );
```
Sanne: you shouldn't need to wrap it in a boolean query "MUST when it's the only clause?

I'll simplify both methods.

yrodiere: Right, I'll rebase and fix that

yrodiere: Ah, didn't see your comment. Thanks :)

Sanne:
merged and tested on the 3 branches. Thanks!

PR #1529, opened 2017-08-30
HSEARCH-2864 Use anchors correctly in the documentation

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2864


Sanne:
thanks. Merged!

PR #1537, opened 2017-09-12
HSEARCH-2867 Duplicate package names in orm/engine, causes problem with jdk9 modules

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2867

Turns out this change is not really useful, because we're far from being the only ones using the same (exported) package name in multiple JARs. It's also the case in Apache Lucene, and to a lesser extent in Hibernate ORM and Hibernate Annotations.

But since it's not a big change... let's merge it, at least it's a step in the right direction.

The errors before and after this patch: https://gist.github.com/yrodiere/473f2db4902d90b0a261ac134737f404
(note there are as many errors after than before, but I suspect it's because the JVM can now inspect Hibernate Search modules that previously failed loading, and thus discovers errors that it previously couldn't detect).

Note that removing the `module-info.java` file will make the test pass, since obviously we won't be using modules anymore.

### DIFF
--- a/integrationtest/jdk9-modules/pom.xml
+++ b/integrationtest/jdk9-modules/pom.xml
@@ -1,0 +1,113 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ ~ Hibernate Search, full-text search for your domain model
+ ~
+ ~ License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ ~ See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <artifactId>hibernate-search-parent</artifactId>
+        <groupId>org.hibernate</groupId>
+        <version>5.8.0-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>hibernate-search-integrationtest-jdk9-modules</artifactId>
+    <packaging>jar</packaging>
+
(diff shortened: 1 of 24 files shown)


yrodiere:
Jenkins, retest this please

yrodiere:
Having a look, the last commit seems to have unexpected side-effects...

Sanne:
That looks like some of the explicit exports we have in the manifest for OSGi needs updating as well. Let me know if you need help with that?

https://github.com/hibernate/hibernate-search/blob/master/engine/pom.xml#L170-L245

yrodiere:
OSGi tests should be okay now.

@gunnarmorling So, if you want to run this test:

 * revert the last commit (since it disables the test)
 * Build and install the test's dependencies with JDK8: `mvn clean install -pl orm -am -Dtest.elasticsearch.host.provided=true -DskipTests`
 * Build and run the test with JDK9: `JAVA_HOME=<your JDK9 HOME> mvn clean install -pl integrationtest/jdk9-modules`
 * Enjoy the compilation failure

Tested with JDK 9+181.

The purpose of this PR was to remove the failures that were specifically caused by Hibernate Search, so you can revert the other commits (except the first one which introduces the test, obviously) to have a look at the original failures.

integrationtest/jdk9-modules/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+    <description>Hibernate Search integration tests for JDK9 modules</description>
+
+    <properties>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>hibernate-search-orm</artifactId>
+            <version>${project.version}</version>
```
Sanne: FYI there's no need to set versions, we manage them all in the root pom.

yrodiere: True... But I think I copy/pasted this, so there must be the same mistake somewhere else.

Sanne:
Very nice! I've been playing with some test improvements so that we don't have to disable it; this way we'll be able to catch regressions on the package duplicates.

One thing isn't clear to me: while we had forgotten to import some more modules, this doesn't seem to matter. But that's something we can investigate another time; I suspect it needs real modules for the stricter rules to kick in, maybe @gunnarmorling 's Moditect will come in handy.

Following up with #1539

PR #1538, opened 2017-09-12
Infinispan needs

Sanne:
A couple of improvements noticed while synching up Infinispan to our latest version.

 - https://hibernate.atlassian.net/browse/HSEARCH-2872
 - https://hibernate.atlassian.net/browse/HSEARCH-2873

### DIFF
--- a/engine/src/main/java/org/hibernate/search/spi/impl/DelegatingIndexedTypeMap.java
+++ b/engine/src/main/java/org/hibernate/search/spi/impl/DelegatingIndexedTypeMap.java
@@ -19,7 +19,7 @@ import org.hibernate.search.util.logging.impl.Log;
 import org.hibernate.search.util.logging.impl.LoggerFactory;
 
 // Must be serializable to support query serialization
-public class DelegatingIndexedTypeMap<V> implements IndexedTypeMap<V>, Serializable {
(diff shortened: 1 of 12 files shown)


engine/src/main/java/org/hibernate/search/spi/impl/DelegatingIndexedTypeMap.java
```diff
@@ -29,6 +29,7 @@
 
 	private final Map<IndexedTypeIdentifier,V> map;
 	private final Map<String,IndexedTypeIdentifier> name2keyMapping;
+	private volatile IndexedTypeSet keyset;
```
yrodiere: I don't understand, how is this related to removing the "put" method? Seems it isn't used, either.

Sanne: Indeed, it's part of a WIP I separated. Forgot to remove this piece.

FYI the (unrelated) work is that we use `keySet` quite regularly and it's not efficient to rebuild the returned `IndexedTypeSet` each time. I don't like the volatile path though, so more changes will be needed to fix that.

thanks for catching that!

Sanne:
merged

PR #1539, opened 2017-09-13
HSEARCH-2867 Duplicate package names in orm/engine, causes problem with jdk9 modules

Sanne:
Folllow up on pr #1537 
https://hibernate.atlassian.net/browse/HSEARCH-2867


yrodiere:
Merged, thanks!

PR #1541, opened 2017-09-14
HSEARCH-2881 Re-enable Wildfly integration tests with JDK9

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2881

Tested locally, tests passed.
Build in progress on Jenkins: http://ci.hibernate.org/view/Personal%20runs/job/hibernate-search-yoann-jdk9/1/

### DIFF
--- a/integrationtest/wildfly/pom.xml
+++ b/integrationtest/wildfly/pom.xml
@@ -173,6 +173,11 @@
             <artifactId>jboss-transaction-api_1.2_spec</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jboss.spec.javax.annotation</groupId>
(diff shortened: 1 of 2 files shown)


integrationtest/wildfly/pom.xml
```diff
@@ -173,6 +173,11 @@
             <artifactId>jboss-transaction-api_1.2_spec</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
```
Sanne: Is this because of `@Generated` et al?

Sanne:
first 5.9 patch merged!

integrationtest/wildfly/pom.xml
yrodiere: Because of `@PostConstruct`. Seems it was moved out of the core libraries in JDK9, so we need a dependency.

PR #1544, opened 2017-09-20
HSEARCH-2832 Hibernate Search produces huge stack traces when Elasticsearch request processing threads are interrupted

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2832

This makes exceptions much shorter by removing the body of each request from the exception message.
Rationale:
 * when something was wrong in the request body, Elasticsearch error messages include a snippet of what was wrong, so the request body is not strictly necessary
 * in case of recurring issues, users can always enable request logging
 * when a lot of exception occur, for instance because the ES server is down, or security is mis-configured, including the full body of each request leads to huge exception messages, to the point of being unreadable and consuming a significant amount of resources.

While this PR does not solve the issue completely, at least it reduces the magnitude of the problem by several factors: most requests are bulk requests, containing up to 500 JSON objects. We used to log those 500 objects, plus the response, which is generally short where there was an error; now we only log the response and a one-line summary of the request (method, path, parameters).

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/logging/impl/ElasticsearchRequestFormatter.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/logging/impl/ElasticsearchRequestFormatter.java
@@ -7,7 +7,6 @@
 package org.hibernate.search.elasticsearch.logging.impl;
 
 import org.hibernate.search.elasticsearch.client.impl.ElasticsearchRequest;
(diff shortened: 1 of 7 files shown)


yrodiere:
Closing, wrong branch...

yrodiere:
Damn, I must be out of it. Re-opening: it was actually the right branch.

yrodiere:
Damn, I must be out of it. Re-opening: it was actually the right branch.

Sanne:
merged! thanks

PR #1546, opened 2017-09-25
HSEARCH-2887

Sanne:
This is the result of looking into the metadata cost, and exploring what our code is costing in terms of native memory utilization - the areas of JVM costs in which we normally don't look.

 - https://hibernate.atlassian.net/browse/HSEARCH-2887

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/logging/impl/Log.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/logging/impl/Log.java
@@ -28,17 +28,22 @@ import org.jboss.logging.annotations.FormatWith;
 import org.jboss.logging.annotations.LogMessage;
 import org.jboss.logging.annotations.Message;
 import org.jboss.logging.annotations.MessageLogger;
+import org.jboss.logging.annotations.ValidIdRange;
 
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 
 /**
  * Hibernate Search log abstraction for the Elasticsearch integration.
+ * Valid ranges for the message ids are between 400000 and 500000.
  *
  * @author Gunnar Morling
  */
 @MessageLogger(projectCode = "HSEARCH")
+@ValidIdRange(min = 400001, max = 499999)
 public interface Log extends BaseHibernateSearchLogger {
+
+	int ES_BACKEND_MESSAGES_START_ID = 400000;
 
 	@Message(id = ES_BACKEND_MESSAGES_START_ID + 1,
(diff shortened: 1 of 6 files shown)


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/logging/impl/Log.java
```diff
@@ -37,7 +38,7 @@
  * @author Gunnar Morling
  */
 @MessageLogger(projectCode = "HSEARCH")
-public interface Log extends org.hibernate.search.util.logging.impl.Log {
+public interface Log extends BaseHibernateSearchLogger {
```
yrodiere: I know it's not _exactly_ the point of this PR, but since it's related... shouldn't we also make sure the "engine" interface isn't used in other modules? And that they rather use the `BaseHibernateSearchLogger` methods on their own child interface?

It seems we currently do use the "engine" logger in other modules, for instance there:

 * `org.hibernate.search.backend.impl.jms.AbstractJMSHibernateSearchController`
 * `org.hibernate.search.elasticsearch.analyzer.impl.NamedElasticsearchAnalyzerReference`
 * ...

engine/src/main/java/org/hibernate/search/util/logging/impl/Log.java
```diff
@@ -200,10 +195,12 @@
 	@Message(id = 48, value = "Unable to retrieve named analyzer: %1$s")
 	void unableToRetrieveNamedAnalyzer(String value);
 
+	/* Moved to the parent logger interface to be shared across compilation units
```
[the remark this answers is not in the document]
Sanne: I'll remove them. This was primarily useful to keep the changesets understandable.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/logging/impl/Log.java
Sanne: yes that's the ideal perfect world scenario. 
I'm worried that we're adding a bit too much "red tape" which really doesn't add much value but makes development slower.

For now I'm inclined to let us use a bit more flexibility and choose to share an error, we can refine this after seeing how it works out.

Sanne: updated this comment as it was worded confusing.

Sanne:
rebased & updated!

Sanne:
N.B. it's also taking advantage of `@ValidIdRange` now, which beyond checking the range in the specific Logger it's annotating, I noticed it also verifies that different Loggers (among the ones it sees - not clear how far this goes) that they stay clear from each other's ranges.

yrodiere:
Merged, thanks!

PR #1547, opened 2017-09-26
HSEARCH-2883 Follow-up: perfomance tests fix

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2883

To be backported to 5.8, I guess (it's not useful to users, but since we backported the previous commits for the same ticket...)


Sanne:
no problem in merging, but out of curiosity: did this cause any failure? I didn't see them, wondering what I'm doing wrong.

Sanne:
merged and backported. Thanks!

yrodiere:
> no problem in merging, but out of curiosity: did this cause any failure? I didn't see them, wondering what I'm doing wrong.

Yes, the elasticsearch performance tests failed when using the "blackhole-5.4" client for instance. But those tests are not ran as part of the build.

Sanne:
got it, thanks for clarifying.

PR #1549, opened 2017-09-26
Cleaning up integration tests from unnecessary legacy

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2889
 - https://hibernate.atlassian.net/browse/HSEARCH-2890

Neither change affects end users.

I'd suggest to backport HSEARCH-2889 just to keep the 5.8 branch aligned to 5.9: could make it easier to backport further stuff, but we could do this if (and ever) there's actual usefulness.


integrationtest/performance/orm/src/test/java/org/hibernate/search/test/performance/TestRunnerArquillian.java
```diff
@@ -58,7 +57,6 @@
 				.addPackages( true, TestRunnerArquillian.class.getPackage() )
 				.addClass( TestConstants.class )
 				.addAsResource( createPersistenceXml(), "META-INF/persistence.xml" )
-				.addAsLibraries( PackagerHelper.hibernateSearchTestingLibraries() )
```
yrodiere: Not to be adressed in this ticket, but we also seem to be using two Hibernate Search JARs in other tests:

* `MemberRegistrationEarArchiveIT`: we add one as library, and also define `wildfly.jpa.hibernate.search.module`
* `CDIInjectionIT`: we using the `hibernateSearchTestingLibraries`, which depend on Hibernate Search. It's convenient, but we could probably avoid any JAR conflict by adding the `hibernate-search-engine` and `hibernate-search-orm` JARs (at least) to the transitive dependencies exclusions for the testing libraries.

Should I create a ticket for that?

Sanne: Nice catch, I'm having a look.

Sanne: Addressed by PR #1550

PR #1552, opened 2017-09-28
HSEARCH-2903 Incorrect word @FieldBrige in documentation

frekele:
https://hibernate.atlassian.net/browse/HSEARCH-2903


Sanne:
@frekele Thanks! I was about to solve the issue but then saw your PR. Merging!

I'll only incorporate the documentation fix; the changelog refers to archived issues so I prefer to keep it as-is.

PR #1557, opened 2017-10-06
HSEARCH-2912 Reenable all integration tests on all JDK profiles

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2912


yrodiere:
Merged, thanks!

PR #1558, opened 2017-10-06
HSEARCH-2905 Follow-up: Fix JDK9 modules test following database profile changes

yrodiere:
I forgot to add the dependency to the database driver... Before the
HSEARCH-2905 fix, the dependency was implicitly added for every module.

Ticket (already resolved): https://hibernate.atlassian.net/browse/HSEARCH-2905


Sanne:
merged! thanks, I was just about to check this issue ..

PR #1559, opened 2017-10-09
hibernate-search-orm performance test fixes

yrodiere:
Relevant tickets:

* [HSEARCH-2893](https://hibernate.atlassian.net/browse/HSEARCH-2893): Fix ORM performance tests on Wildfly (they fail silently)
* [HSEARCH-2894](https://hibernate.atlassian.net/browse/HSEARCH-2894): Fix off-by-one error in initial index size of ORM performance tests
* [HSEARCH-2895](https://hibernate.atlassian.net/browse/HSEARCH-2895): Prepare ORM performance tests for the introduction of completely different scenarios


Sanne:
Looks like you also removed the configuration for the `HikariCPConnectionProvider` ? We need a reasonable connection provider to run these tests.

integrationtest/performance/orm/src/test/resources/arquillian.xml
```diff
@@ -22,7 +22,11 @@
     <container qualifier="wildfly" default="true">
         <configuration>
             <property name="jbossHome">${project.build.directory}/wildfly-${wildflyVersion}</property>
-            <property name="javaVmArguments">-Xmx512m -Djava.net.preferIPv4Stack=true -Djgroups.bind_addr=127.0.0.1</property>
+            <property name="javaVmArguments">
+                -Xmx512m -Djava.net.preferIPv4Stack=true -Djgroups.bind_addr=127.0.0.1
+                -Dscenario=${scenario}
+                -Dorg.hibernate.search.enable_performance_tests=${org.hibernate.search.enable_performance_tests}
+            </property>
```
Sanne: Are you sue that's necessary? Adding them won't hurt but I'm wondering, as while WildFly is run in a separate JVM the test execution driver is not: the test methods are invoked over a remote protocol.

integrationtest/performance/orm/src/test/java/org/hibernate/search/test/performance/scenario/TestReporter.java
```diff
@@ -106,9 +106,14 @@ private static void printTaskInfo(TestScenarioContext ctx, PrintWriter out) {
 			String taskAverageTime = "n/a";
 			if ( ctx.testContext.measureTaskTime ) {
 				long taskTotalMilis = TimeUnit.MILLISECONDS.convert( task.getTimerValue(), TimeUnit.NANOSECONDS );
-				long taskAverageMilis = taskTotalMilis / task.getCounterValue();
+				long taskCount = task.getCounterValue();
 				taskTotalTime = DurationFormatUtils.formatDuration( taskTotalMilis, "mm:ss.SSS" );
-				taskAverageTime = DurationFormatUtils.formatDuration( taskAverageMilis, "mm:ss.SSS" );
+				if ( taskCount == 0L ) {
```
Sanne: good point :)

integrationtest/performance/orm/src/test/java/org/hibernate/search/test/performance/TestRunnerArquillian.java
```diff
@@ -100,7 +102,7 @@ private static Asset createPersistenceXml() {
 
 	public static Asset manifest() {
 		String manifest = Descriptors.create( ManifestDescriptor.class )
-				.attribute( "Dependencies", "org.apache.commons.lang" )
+				.attribute( "Dependencies", "org.apache.commons.lang,com.google.guava" )
```
Sanne: Where is the Guava requirement coming from? Can we avoid it?

integrationtest/performance/orm/src/test/java/org/hibernate/search/test/performance/scenario/TestReporter.java
```diff
@@ -125,15 +124,15 @@ private static void printTaskInfo(TestContext ctx, PrintWriter out) {
...
 		out.println( "    measure performance : " + TestConstants.arePerformanceTestsEnabled() );
 		out.println( "    threads             : " + THREADS_COUNT );
-		out.println( "    measured cycles     : " + ctx.scenario.measuredCyclesCount );
-		out.println( "    warmup cycles       : " + ctx.scenario.warmupCyclesCount );
-		out.println( "    initial book count  : " + ctx.scenario.initialBookCount );
-		out.println( "    initial author count : " + ctx.scenario.initialAuthorCount );
+		out.println( "    measured cycles     : " + testCtx.measuredCyclesCount );
+		out.println( "    warmup cycles       : " + testCtx.warmupCyclesCount );
+		out.println( "    initial book count  : " + testCtx.initialBookCount );
+		out.println( "    initial author count : " + testCtx.initialAuthorCount );
```
Sanne: Could you s/initial author count :/initial author count:/ ? Just to keep all semicolons nicely aligned ;)

integrationtest/performance/orm/src/test/java/org/hibernate/search/test/performance/scenario/TestContext.java
```diff
@@ -41,20 +41,20 @@
...
 		else {
 			this.initialOffset = 100;
 			this.initialAuthorCount = 10;
 			this.initialBookCount = 100;
-			this.warmupCyclesCount = 1;
-			this.measuredCyclesCount = 1;
 		}
+
+		this.threadCount = 2;
+		this.warmupCyclesCount = 1;
```
Sanne: This looks like suspicious. Just one cycle for warming up the JVM even when performance tests are enabled?

integrationtest/performance/orm/src/test/java/org/hibernate/search/test/performance/TestRunnerArquillian.java
yrodiere: The Stopwatch class. The one from Apache only has millisecond resolution. If that's okay with you, then we can switch to the Apache one; otherwise we'll have to code our own, and I don't see the point as this is only a test.

integrationtest/performance/orm/src/test/resources/arquillian.xml
yrodiere: Yes I'm sure. The properties are read from our own test classes, executed inside WildFly, namely in `TestContext` (when initializing `org.hibernate.search.test.performance.scenario.TestContext#performanceEnabled`) and `org.hibernate.search.test.performance.scenario.TestScenarioFactory#create`.

integrationtest/performance/orm/src/test/java/org/hibernate/search/test/performance/TestRunnerArquillian.java
Sanne: ok, let's use the Guava one.

integrationtest/performance/orm/src/test/java/org/hibernate/search/test/performance/scenario/TestContext.java
yrodiere: I moved this configuration to the tests themselves, because it makes no sense to say "hey, let's do a thousand warmup cycles" when we don't know how expensive one cycle is. I can guarantee that one thousand mass indexing cycles are way too much.

See `org.hibernate.search.test.performance.scenario.ReadWriteTestScenario#initContext`

integrationtest/performance/orm/src/test/java/org/hibernate/search/test/performance/TestRunnerArquillian.java
Sanne: BTW I don't see how it can be more precise that milliseconds :) but ok that's not too relevant.

integrationtest/performance/orm/src/test/java/org/hibernate/search/test/performance/scenario/TestContext.java
Sanne: oh, ok. Add a comment here for future understanding?

yrodiere:
Right, I'll see about this HikariCP thing. I expect it's not needed in Wildfly, since there must be some built-in, production-grade connection pool?

integrationtest/performance/orm/src/test/java/org/hibernate/search/test/performance/TestRunnerArquillian.java
yrodiere: It uses `System.nanoTime()`. But yeah, maybe it's about time drift rather than resolution, I keep confusing those things.

Sanne:
That's right, we only need HikariCP (or similar) when running standalone. WildFly has its own pool implementation.

yrodiere:
@Sanne Thanks, I addressed your comments.

Sanne:
merged!

PR #1560, opened 2017-10-09
HSEARCH-2915 Update to Narayana 5.5.30.Final

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2915


yrodiere:
Merged, thanks!

PR #1561, opened 2017-10-09
HSEARCH-2874 Update to Checkstyle 8.3

Sanne:


Sanne:
Added a second commit: it doesn't work on JDK9 yet but there's an easy workaround.

yrodiere:
Merged, thanks!

PR #1565, opened 2017-10-13
HSEARCH-2916 Avoid random failures in TestRunnerStandalone.runPerformanceTest

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2916

Turns out there was a rather simple solution. I'm not sure it works, since I've been unable to reproduce the issue, but let's merge and hope it does?


Sanne:
merged, thanks!

PR #1570, opened 2017-10-26
Update the previous.stable property in the POM

yrodiere:


Sanne:
Ok I'll take it but does it matter?

yrodiere:
@Sanne In this case I don't know, but it can matter when we backport a new API in a micro, say 5.8.2. In that case we must not change that API in 5.9.0 for instance, and if we don't update previous.stable some backward-incompatible changes might go unnoticed.

PR #1571, opened 2017-10-26
HSEARCH-2930 + HSEARCH-2931 + HSEARCH-2932 JSR-352 test fixes

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2930
https://hibernate.atlassian.net//browse/HSEARCH-2931
https://hibernate.atlassian.net//browse/HSEARCH-2932

Note: this does not fix tests on MariaDB completely, only part of it. We'll have to wait for https://hibernate.atlassian.net/browse/HSEARCH-2933[HSEARCH-2933] to be fixed before JSR-352 can work with the MariaDB JDBC driver.

### DIFF
--- a/integrationtest/jsr352/pom.xml
+++ b/integrationtest/jsr352/pom.xml
@@ -37,6 +37,11 @@
         </dependency>
         <dependency>
             <groupId>org.hibernate</groupId>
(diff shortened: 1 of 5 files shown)


Sanne:
merged! thanks

PR #1572, opened 2017-10-30
Logger optimisation

DavideD:
@Sanne THese are the changes to the log to improve performance, I don't know if there is a JIRA for it so you will probably need to update the commit history.

I also left your original commit.

Cheers


engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
```diff
@@ -42,9 +43,9 @@
 import org.hibernate.search.spatial.DistanceSortField;
 import org.hibernate.search.spi.CustomTypeMetadata;
 import org.hibernate.search.spi.IndexedTypeIdentifier;
+import org.hibernate.search.spi.IndexedTypeMap;
```
Sanne: just curious: seems something special happened on this case?

gunnarmorling:
Probably too late now, but you could have applied my automated refactoring for this: https://github.com/hibernate/hibernate-validator/pull/874. Maybe try and use it for OGM?

engine/src/main/java/org/hibernate/search/query/engine/impl/AbstractHSQuery.java
DavideD: I probably organized the imports on this and didn't notice

DavideD:
@gunnarmorling  I did the Hibernate OGM one and this one on Friday with regular expressions. Before I've seen your commit

Sanne:
merged as HSEARCH-2935! thanks

PR #1573, opened 2017-10-31
Build cleanup

Sanne:
Cleaning up some technical debt in the build:
 - https://hibernate.atlassian.net/browse/HSEARCH-2249
 - https://hibernate.atlassian.net/browse/HSEARCH-2936
 - https://hibernate.atlassian.net/browse/HSEARCH-2937

Incidentally, this gives a big speedup of the build when skipping tests. Useful during various dev cycles.


yrodiere:
Very nice! I couldn't think of a way to implement the Elasticsearch bit, but here it is :) Merged!

PR #1576, opened 2017-11-02
HSEARCH-2926 Upgrade surefire/failsafe to 2.20.1

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2926


Sanne:
this was merged. thanks!

PR #1577, opened 2017-11-02
HSEARCH-1885 BridgeFactory unable to create instances of custom bridges in a modular environment

Sanne:
RFC
 - https://hibernate.atlassian.net/browse/HSEARCH-1885

Preview: requires at least documentation changes still.

This is mostly to make sure we can test it properly. The JIRA was suggesting we also find a way to actually load things from the other modules but I no longer think that's a good idea: that implies to load classes regularly as if the current context classloader was different?
Not sure I like the idea to override the normal module workings.

BTW this looks way simpler now than what I had in POC branches. The key is in the exception handling, especially catching the `LinkageError`.

Comments welcome


orm/src/main/java/org/hibernate/search/cfg/impl/DelegatingClassLoaderService.java
```diff
@@ -44,7 +44,7 @@ public DelegatingClassLoaderService(org.hibernate.boot.registry.classloading.spi
 		try {
 			return hibernateClassLoaderService.classForName( className );
 		}
-		catch (ClassLoadingException e) {
+		catch (ClassLoadingException | LinkageError e) {
```
yrodiere: If I understand correctly, this commit is about much, much more than just translating a failure to load the TikaBridge into a different exception. It is also about implementing a different behavior to handle that exception. Maybe some comments would be in order?

The new behavior seems to be the following: if we fail to link a Hibernate Search class to the classes it depends on (in our case, `TikaBridge` to Tika classes), we try to load this Hibernate Search class again from the user's class loader, which, on top of allowing to load Hibernate Search classes, may also allow to load classes from additional dependencies (in our case, the Apache Tika module).

This looks like a good idea. It's also great that it doesn't require moving our Tika-related classes to a separate module, though we may want to do it anyway in the next major version.

But... I can't begin to understand why the modular class loader would allow a class from one dependency to access classes from another dependency, when those dependencies are unrelated to each other... I would expect our `TikaBridge` class to be loaded from the Hibernate Search module classloader, and all linkage to be performed in the context of this classloader, not in the context of the top-level (user application) classloader...
Otherwise, in the case of user-facing API classes, with multiple user modules depending on the Hibernate Search module, wouldn't we run the risk of class loading behaving differently depending on which user module accesses the API classes first? Which seems really bad...
Aren't we exploiting a flaw somehow? Do you know for sure it works like that on purpose?

Sanne: No the module system is not going to allow linking across two different modules which have no direct visibility.
Catching the `LinkageException` is just to handle the *invalid modules* as the not-right way of loading the extension point, so to either throw a more helpful exception or to allow re-trying over a different classpath which might be valid.

The conceptual problem here is that I don't want to expose *private modules* to the user's classpath, but we seem to expect that the user's classpath is exposed to our private modules.

That first step is conceptually doable but to initialize the (hypotetical) private extension points (e.g. tika-core) with additional extension points (e.g. a specific format Tika extension) provided by the user would require such extension points to explicitly handle classloaders right :(

I've been trying some approaches around that and this resulted only in opening more JIRAs.

I plan to have a second look this weekend about another POC I made a while back - having a new perspective - but I suppose we need to move the entire extension into the user's deployment, like I'm doing in the integration tests (although I sneaked in an additional module to avoid splitting our current API - as you suggested we'd need to separate that out to avoid needing the module).

yrodiere:
Jenkins, retest this please

orm/src/main/java/org/hibernate/search/cfg/impl/DelegatingClassLoaderService.java
yrodiere: > No the module system is not going to allow linking across two different modules which have no direct visibility.
> Catching the LinkageException is just to handle the invalid modules as the not-right way of loading the extension point, so to either throw a more helpful exception or to allow re-trying over a different classpath which might be valid.

In that case I don't understand what this "different classpath which might be valid" is. There is one, right? And that's how you solved the issue?

How I see it:

 * Hibernate Search's classloader has access to `TikaBridge`, but not to Tika classes
 * The user's classloader has access to Tika classes and to `TikaBridge` through its module dependencies, but as you said "the module system is not going to allow linking across two different modules which have no direct visibility"

So... do you know what is going on?

Sanne:
pushed the rename of the modules to `hibernate-search-wildfly-testingmodules` . Still need to address the other points.

orm/src/main/java/org/hibernate/search/cfg/impl/DelegatingClassLoaderService.java
Sanne: Sorry for the delay, finally getting back at this.

Where I said "different classpath which might be valid" I meant a different _classloader_ ; the problem being that we try 1) the user's classloader 2) our own 3) others as listed by the integrator. 
I don't like this approach of looping-until-we-find-it but that's the best we can get from the current ORM's SPI `AggregatedClassLoader`.

The problem was that one of these could have the correct Bridge (and dependencies) but it's possible that another classloader would fail (critically) before we'd try the right one. In this case the `LinkageException` is safe to be ignored - so that we don't abort but try the next classloader.

Sanne: > Otherwise, in the case of user-facing API classes, with multiple user modules depending on the Hibernate Search module, wouldn't we run the risk of class loading behaving differently depending on which user module accesses the API classes first? Which seems really bad...
Aren't we exploiting a flaw somehow? Do you know for sure it works like that on purpose?

I see no flaw; let me try to clarify one point - not sure if that's the confusing one though: feel free to correct me.
We'll be loading an extension _instance_ for the current SearchIntegrator _instance_; there are no threadlocals nor static constants which would risk leaking the extension nor the class definition being obtained to a different deployment.

Sanne:
Ultimately I think we'll have to go with this solution; I'll add a documentation patch to explain to module users how to package Tika extensions.

orm/src/main/java/org/hibernate/search/cfg/impl/DelegatingClassLoaderService.java
yrodiere: > We'll be loading an extension instance for the current SearchIntegrator instance; there are no threadlocals nor static constants which would risk leaking the extension nor the class definition being obtained to a different deployment.

My concern was more that the `TikaBridge` class will be "linked" to classes from the user's deployment (the Tika classes). Isn't it possible that the `TikaBridge` class itself (which includes links to the user's deployment classes) will later be re-used in another deployment? Thereby providing (indirect) access from one deployment to the classes of another deployment?

I probably just don't understand how class linking works... I expected linking to require a change of some sort to the class definition (so as to point to the linked classes).

Or maybe the `TikaBridge` class will be loaded successfully by the user's deployment classloader only, and thus will never be re-used in another deployment?

Sanne: The `TikaBridge` which might get linked to the user deployment is the one defined in the (currrent) TCCL - so it's the one from the user's deployment classloader (And that's why the user code must explicitly import the Tika dependency). Overriding the TCCL (Thread Context ClassLoader) during the scope of the Tika bridge initialization is the alternative I've explored and ultimately discarded, as such concerns would indeed apply.

Sanne:
> You didn't include my PR at Sanne#6, is it on purpose?

Not on purpose. Sorry forgot about it, will integrate it now with.

documentation/src/main/asciidoc/mapping.asciidoc
```diff
@@ -1710,6 +1711,8 @@ metadata.
 <<example-tika-mapping>>) need to explicitly enable the option ignoreFieldBridge().
 ====
 
+To use the `@TikaBridge` in combination with WildFly modules, make sure to package your Tika extensions as described in <<modules-tika>>.
```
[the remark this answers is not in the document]
Sanne: ok I'll change it. I was torn myself on that, but FYI decided that "when in doubt I'll avoid adding more boxes".. breaks too much readability. The threahold of "when in doubt" is of course affected if others also think it should :)

documentation/src/main/asciidoc/configuration.asciidoc
```diff
@@ -1416,7 +1418,8 @@ is compatible with the Hibernate ORM version you choose.
 [WARNING]
 ====
 This version of Hibernate Search `{hibernateSearchVersion}` requires an Hibernate ORM version `5.2.3.Final` or a later version
-of the 5.2 branch, such as `5.2.8.Final`.
+of the 5.2 branch, such as `{hibernateVersion}`.
```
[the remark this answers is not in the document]
Sanne: I'd rather not in this case, as the whole paragraph is specific for 5.2. Makes it easier to remove it all when I search for such markers later..

documentation/src/main/asciidoc/configuration.asciidoc
```diff
@@ -1406,8 +1406,10 @@ configuration.
 Hibernate Search is included in the WildFly application server, and since WildFly 10 the module
 is automatically activated (added to the classpath of your deployment) if you are using Hibernate ORM and have any indexed entities.
 
-Alternatively you can opt to use a different version of the module by downloading and unzipping a different
-moduleset and setting the `wildfly.jpa.hibernate.search.module` property in your `persistence.xml`.
+WildFly includes an older version of Hibernate Search; if you want to use that then you won't need to download anything else and you can stop reading this chapter; however you might miss out on several new features and improvements.
```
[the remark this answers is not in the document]
Sanne: Which version should I add? All versions of WildFly include some Hibernate Search version, and it will always be older than what the latest version of this documentation mentiones.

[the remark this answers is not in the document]
yrodiere: I'd just go with the one we target in our tests. It only seems logical that older versions of WildFly ship with even older versions of Hibernate Search.

documentation/src/main/asciidoc/configuration.asciidoc
```diff
@@ -1497,6 +1497,33 @@ Hibernate Search and come with a different edition of this guide specifically ta
...
+
+Include Apache Tika {tikaModuleVersion} and any extensions and dependencies from Tika that you might need in the module.
+
+Next, you need to have your application depend on this same module as well:
+
+.Manifest header to add in your application
+====
+[source]
+[subs="verbatim,attributes"]
+Dependencies org.apache.tika:{tikaModuleVersion}
```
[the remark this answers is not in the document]
Sanne: good point. thanks!

Sanne:
Suggestions applied. Couldn't run the integration tests locally though, let's wait for CI.

yrodiere:
Merged, thanks!

PR #1578, opened 2017-11-03
HSEARCH-2934 Pending Elasticsearch streamed works may fail when stopping Hibernate Search and using the drop-and-create-and-drop strategy

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2934

The solution I chose was to provide each index manager with its own view of the stream orchestrator, so that it can close its view when closing the index manager, which will automatically wait for pending stream works to be executed.

I could have simply added an explicit wait on the "shared" stream orchestrator, but my solution has the additional benefit of preventing new stream works to be accepted by the index manager while waiting for pending stream works to be executed.

### DIFF
--- a/integrationtest/elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/ElasticsearchIndexManagerClosingIT.java
+++ b/integrationtest/elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/ElasticsearchIndexManagerClosingIT.java
@@ -1,0 +1,221 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.elasticsearch.test;
+
+import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ForkJoinPool;
+import java.util.regex.Pattern;
+
+import org.hibernate.search.annotations.DocumentId;
+import org.hibernate.search.annotations.Indexed;
+import org.hibernate.search.backend.AddLuceneWork;
+import org.hibernate.search.bridge.util.impl.ContextualExceptionBridgeHelper;
+import org.hibernate.search.elasticsearch.cfg.ElasticsearchEnvironment;
+import org.hibernate.search.elasticsearch.cfg.IndexSchemaManagementStrategy;
+import org.hibernate.search.elasticsearch.client.impl.ElasticsearchClientFactory;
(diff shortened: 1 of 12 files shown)


Sanne:
Great fix!

I have only a minor consideration about how it's tested, I'll comment on the specific change.

integrationtest/elasticsearch/src/test/java/org/hibernate/search/elasticsearch/test/ElasticsearchIndexManagerClosingIT.java
```diff
@@ -0,0 +1,204 @@
...
+				);
+		// Important: we actually want to perform a stream, async operation here.
+		searchIntegrator.getIndexManager( Entity.INDEX_NAME )
+				.performStreamOperation( work, null, true );
+
+		// Schedule the work to be actually executed a long time from now,
+		// so that we can be sure the closing process waited for it to execute
+		ForkJoinPool.commonPool().submit( () -> {
+			try {
+				Thread.sleep( 1_000 );
```
Sanne: Could you avoid waiting for an arbitrary amount of time? It's almost never a good idea in tests: either we end up with very slow tests as they get littered with "wait" operations, or often we're not even sure if the delay is going to be enough for the machines actually running the tests.

In this case you could just have a barrier, which is broken right after the searchIntegrator.close() ? Does it even have to be a different thread to mark the "complete" phase?

yrodiere: The change needs to happen sometime *during* the call to `searchIntegrator.close()`, because if all goes well the thread closing the search integrator is actually waiting for the change to happen. If the change happens before the call to `close()`, we won't be testing that the wait occurs, and if it happens after the call to `close()`, the call to `close()` will never terminate (it will wait forever).

So I definitely need another thread, and I'm afraid I can't use a barrier, because the barrier would actually need to be broken sometime *during* the `searchIntegrator.close()`... I would need some extension point inside the index manager allowing me to break the barrier just after the index manager started waiting on pending threads...

Please, don't answer "use byteman" xD

Sanne: use ASM then! xD

Sanne: Can you possibly invert the logic? do an "asynch close" which is blocking forever - unless the event is done.
Something along these lines should allow you to avoid picking an arbitrary sleep time.

Sanne: my bad I misunderstood the intent. We agreed on adding a reminder comment and merge.

yrodiere:
Merged!

PR #1579, opened 2017-11-03
HSEARCH-2944 Clarify that all methods of MassIndexerProgressMonitor are incremental

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2944


Sanne:
Looks good! Thanks.

I'll just make a micro change where you say "his method can be invoked from several threads and should thus be thread-safe." I'll point out it's the implementor which needs to be.

Sanne:
merged

PR #1587, opened 2017-11-28
Lazy Loading Exception after indexing an entity

sagarghai:
**Update** 
AbstractLazyInitializer.java to fix below issue for setting index from search in version hibernate-core 5.2.8. Below is the Exception.

`HHH000260: Exception calling user Synchronization` `[org.hibernate.search.backend.impl.EventSourceTransactionContext$BeforeCommitSynchronizationDelegator@b4884752] : org.hibernate.LazyInitializationException: failed to lazily initialize a collection of role: com.myapp.loan.LoanApplication.subloans, could not initialize proxy - the owning Session was closed`

`Caused by: org.hibernate.LazyInitializationException: failed to lazily initialize a collection of role: com.myapp.entities.LoanApplication.subloans, could not initialize proxy - the owning Session was closed`

Below is the code line causing this issue. We are using hibernate search 5.7.

`org.hibernate.search.jpa.FullTextEntityManager fullTextEntityManager = org.hibernate.search.jpa.Search
.getFullTextEntityManager(entityManager);
fullTextEntityManager.index(loanApplication);`

Session is marked as closed which doesn't allow the entity to update the indexes. Please see the attached testcase.

> Old Code
```
if (!(this.session.isOpen()))
{
throw new LazyInitializationException("could not initialize proxy - the owning Session was closed");
}
```

> Updated Code

```
if (!(this.session.isOpenOrWaitingForAutoClose()))
{
throw new LazyInitializationException("could not initialize proxy - the owning Session was closed");
}
```

Unable to reproduce the exact issue using only hibernate core. Hence reporting it here! 

P.S. Extending from the previous PR sent to hibernate core (HHH-11902 - HHH000260: Exception calling user Synchronization - hibernate update indexes #1969). Also here is the JIRA Id for the same : Hibernate ORM : Jira Id : HHH-11902 .

> Attaching a sample test case to reproduce the issue.,

[sampleTestCase.zip](https://github.com/hibernate/hibernate-search/files/1509738/sampleTestCase.zip)


yrodiere:
This is not the right place at all. You created a pull request, which is used to submit fixes to us so that we add the fixes to Hibernate Search.

If you want to discuss something, please continue on the JIRA ticket: https://hibernate.atlassian.net/browse/HHH-11902 . We will move the ticket to the Hibernate Search project if necessary.

Closing this pull request, since it's not really a pull request...

sagarghai:
Hello,
Creating an issue instead since the changes have been accepted into a future upgrade of hibernate-core.

sagarghai:
Hello,
Although the changes have been accepted into hibernate core, it still fails for previous versions and the issue seems to persist when using hibernate core 5.2.8 with hibernate search. The issue is only reproduced with hibernate search indexing. Would it be possible to look into the issue for backward compatibility?

yrodiere:
Replying on the ticket, since you asked the same question there.
Please keep the discussion there, it will be easier for everyone ;)

Sanne:
@sagarghai as I mentioned on HHH-11902 the issue is fixed already. You'll have to update Hibernate ORM (core) to _5.2.12.Final_ . We don't fix older versions because we can't do time travel.

PR #1590, opened 2017-11-29
HSEARCH-2961 Downgrade the version of the MariaDB plugin

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2961


Sanne:
merged

PR #1591, opened 2017-11-29
 HSEARCH-2958 Rework Mockito tests

mincong-h:
https://hibernate.atlassian.net/browse/HSEARCH-2958

The first 2 commits transferred Mockito tests into EasyMock tests. But later, I realized that there's even no need to use mock! Using Java polymorphisme did the work, so I pushed my 3rd commit. I didn't squash them into one because I'm not sure if you want to keep the EasyMock version or the no-mock one.

### DIFF
--- a/jsr352/core/pom.xml
+++ b/jsr352/core/pom.xml
@@ -50,8 +50,8 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.mockito</groupId>
-            <artifactId>mockito-all</artifactId>
+            <groupId>org.easymock</groupId>
+            <artifactId>easymock</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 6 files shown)


Sanne:
Hi @mincong-h ! Thanks, indeed avoiding the mocking libraries altogether is my favourite solution too.

Looks like some of the integration tests in WildFly need to be updated as well though:
 - http://ci.hibernate.org/job/hibernate-search-PR/org.hibernate$hibernate-search-integrationtest-wildfly/lastCompletedBuild/testReport/

yrodiere:
@Sanne I don't see why writing our own mocks by hand, with an implementation for each and every method we don't need, would be an improvement... ?
If you really want to merge that last commit, ok, but otherwise I'd rather stick to not reinventing the wheel...

mincong-h:
Mon own implementation is indeed too complex and makes the test logic
difficult to understand. Another option is to use concrete class directly,
like  `com.ibm.jbatch.container.context.impl.StepContextImpl` . Otherwise
I'm fine with the easymock version. (That's why kept both commits)

yrodiere:
> That's why kept both commits

And I thank you for that :+1: 
Let's see what Sanne thinks, if he agrees I'll merge the Easymock version.

Sanne:
> Let's see what Sanne thinks, if he agrees I'll merge the Easymock version.

As you all prefer. Ease of maintenance gets precedence.

I _personally_ tend to avoid mocks when the overhead (in amounts of code) is not too high, but what is "too much" is up to you.

I do suspect however that in a couple of years neither of you will rememer details of this test and we'll suddenly need to remove easymock for whatever reason.. you might regret what you decide today  :P

yrodiere:
Tested on Travis and merged (with EasyMock). Thanks!

PR #1592, opened 2017-11-30
 HSEARCH-2963 Remove duplicate dependency declarations

mincong-h:
https://hibernate.atlassian.net/browse/HSEARCH-2963

I also added an EnforcerRule [_"Ban Duplicate Pom Dependency Versions"_][1] to avoid similar ambiguity in the future.

[1]: https://maven.apache.org/enforcer/enforcer-rules/banDuplicatePomDependencyVersions.html

### DIFF
--- a/integrationtest/wildfly/pom.xml
+++ b/integrationtest/wildfly/pom.xml
@@ -88,25 +88,9 @@
             <type>test-jar</type>
             <scope>test</scope>
         </dependency>
-        
-        <dependency>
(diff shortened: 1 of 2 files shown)


yrodiere:
Merged, thank you!
Note the enforcer rule won't have any effect on the wildfly integration tests module, because we skip the enforcer due to problems in transitive dependencies (out of our control). But at least other modules will benefit from it :)

PR #1593, opened 2017-11-30
HSEARCH-2959 + HSEARCH-2960 build improvements

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2959
https://hibernate.atlassian.net//browse/HSEARCH-2960

Example of a working build on Travis: https://travis-ci.org/yrodiere/hibernate-search/builds/309389247
And the logs on ES2: https://travis-ci.org/yrodiere/hibernate-search/jobs/309389248


yrodiere:
Wait, I just noticed the checkstyle execution doesn't stop the build when it fails. Working on it now...

yrodiere:
Fixed the problem with checkstyle failures not aborting the build. Ready to review!

Sanne:
Nice. Feel free to merge; just some things I'm wondering about (hypotehtically):
- you could skip ChecksTyle on the last phase? ( the "clean install").
- looks like you're not skipping PDF builds in that same last phase. Intentional?
- about the initial build which is meant to satisfy checkstyle: we could limit it to a short list of GAVs we actually need? (probably not worth to maintain such a list?)

yrodiere:
> you could skip ChecksTyle on the last phase? ( the "clean install").

Right, I will do that then merge. Just wanted to make this last phase as close to the release as possible, but I guess this is not a dangerous change.

> looks like you're not skipping PDF builds in that same last phase. Intentional?

Yeah, I didn't want to change what we do in the build, just avoid redundancy. If you think we should avoid building the PDF during the default build, I can address that in another ticket, but I think I will change the documentation POM rather than `.travis.yaml` (so that the change also applies to local builds).

> about the initial build which is meant to satisfy checkstyle: we could limit it to a short list of GAVs we actually need? (probably not worth to maintain such a list?)

It tried to, but it proved a bit complicated. For instance some performance testing modules have plugins that depend on the WildFly integration test module... So the plugin dependencies are not very straightforward, and I suspect in the end we'll need to build almost everything.

gsmet:
> about the initial build which is meant to satisfy checkstyle: we could limit it to a short list of GAVs we actually need? (probably not worth to maintain such a list?)

Don't think it's worth it.

yrodiere:
Merged, thanks :)

gsmet:
@yrodiere +1 to be consistent and remove the PDF build by default.

The `-Pdocumentation-pdf` option is already included in the master job so the master job will fail if the PDF cannot be generated.

PR #1594, opened 2017-12-01
HSEARCH-2964 Add the JSR 352 modules to the distribution

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-2964
 * https://hibernate.atlassian.net/browse/HSEARCH-2778
 * https://hibernate.atlassian.net/browse/HSEARCH-2969
 * https://hibernate.atlassian.net/browse/HSEARCH-2968


distribution/pom.xml
```diff
@@ -119,6 +141,9 @@
                         <link>http://lucene.apache.org/core/5_5_4/analyzers-common/</link>
                         <link>http://lucene.apache.org/core/5_5_4/queryparser/</link>
                         <link>http://jgroups.org/javadoc</link>
+                        <!-- for javax.batch in the JSR 352 modules -->
+                        <link>https://docs.oracle.com/javaee/7/api/</link>
+                        <link>http://docs.jboss.org/jberet/${jberetVersion}/javadoc/jberet-core/</link>
```
gsmet: It will probably be useless if/when you only include the API/SPI in the javadoc but if one day we want to go back to the current setup, at least it will be comprehensive.

jsr352/pom.xml
```diff
@@ -21,90 +21,9 @@
...
-            <comments>See also: http://hibernate.org/license</comments>
-        </license>
-    </licenses>
-
-    <developers>
-        <developer>
-            <id>mincong-h</id>
-            <name>Mincong Huang</name>
-            <email>mincong.h@gmail.com</email>
-            <url>http://mincong-h.github.io</url>
```
gsmet: Added Mincong to the parent pom.

distribution/pom.xml
```diff
@@ -70,6 +70,26 @@
...
             <artifactId>hibernate-search-backend-jgroups</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-search-jsr352-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>javax.inject</groupId>
+            <artifactId>javax.inject</artifactId>
+        </dependency>
```
yrodiere: Why did you add those dependencies to javax.inject and javax.batch? Is this necessary to generate the javadoc?

gsmet: Yup. There are references to them in the analyzed code we need them. That's why we already have most of the other dependencies.

yrodiere:
Adding the Changes required label: Guillaume is working on making this faster (replacing the javadoc:jar goal with javadoc:javadoc in particular).

distribution/pom.xml
```diff
@@ -117,36 +117,27 @@
...
-                </configuration>
                 <executions>
                     <execution>
-                        <id>aggregate-javadoc</id>
-                        <phase>install</phase>
-                        <goals>
-                            <goal>javadoc</goal>
-                        </goals>
+                        <!-- Aggregate the javadocs of the selected modules  -->
+                        <id>generate-javadoc</id>
```
yrodiere: You might want to document the hack here?

gsmet: Done in a fixup commit.

yrodiere:
Merged, thanks a lot :)

PR #1595, opened 2017-12-01
HSEARCH-2965 Do not generate the PDF documentation by default

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2965


yrodiere:
Merged, thanks!

PR #1596, opened 2017-12-04
HSEARCH-2967 Remove internal classes from the Javadoc

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2967

Downsides:

1. Links to internal classes from API/SPI will be dead links. There are some, unfortunately, for example from `AbstractDocumentBuilder` to `TypeMetadata`.
2. We will have some errors in the logs (example below) because some artifacts do not have any SPI or API, so their javadoc is not generated. When times come for the aggregation, their javadoc cannot be found and this error appears. Everything works fine though, since the error is ignored.

Example of errors:
```
[ERROR] Error fetching link: /home/travis/build/yrodiere/hibernate-search/serialization/avro/target/site/apidocs/package-list. Ignored it.
```

### DIFF
--- a/documentation/pom.xml
+++ b/documentation/pom.xml
@@ -36,6 +36,8 @@
 
         <!-- Skip artifact deployment -->
         <maven.deploy.skip>true</maven.deploy.skip>
+        <!-- Skip javadoc generation -->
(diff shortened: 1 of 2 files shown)


Sanne:
looks good! Merging..

PR #1598, opened 2017-12-20
HSEARCH-2978 + HSEARCH-2979 Target CDI 2.0 + upgrade to ORM 5.3.0.Beta1/JPA 2.2 + use ORM ManagedBeanRegistry for FieldBridge injection

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2978
https://hibernate.atlassian.net/browse/HSEARCH-2979


integrationtest/wildfly/pom.xml
```diff
@@ -418,6 +420,75 @@
                             </dependency>
                         </dependencies>
                     </plugin>
+                    <plugin>
```
Sanne: I guess this works, but you could have asked the provisioning plugin to just replace the version during initial resolution?

yrodiere: I wrote this (actually, you did and I copy/pasted for CDI) before we were even using feature packs :) I can replace it though.

Sanne: I just checked, that might not be easy. The <version-overrides> schema in [1] seems able to re-define the version, but not remap the other GAV elements to a different artifact altogether :(

1 - https://github.com/wildfly/wildfly-build-tools/blob/master/provisioning/src/main/resources/wildfly-server-provisioning-1_2.xsd

yrodiere:
@Sanne So to sum up, feature-pack overrides seem to be a no-go for the JPA upgrade. Regarding the Weld upgrade, I'm not sure a simply version override would work, since the Weld team may have added new modules or dependencies. I can try if you really want, but I feel like there's little value in it...

Maybe we can merge this one?

Sanne:
agreed, I'll re-test this and merge.

Sanne:
merged!

PR #1599, opened 2017-12-22
HSEARCH-2942 Move Wildfly module packaging of dependencies to external repositories

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2942

In order to build this branch, you will need to build and install in your local Maven repository the following projects:

* https://github.com/hibernate/lucene-jbossmodules
* https://github.com/yrodiere/elasticsearch-client-jbossmodules
* https://github.com/yrodiere/gson-jbossmodules
* https://github.com/yrodiere/aws-v4-signer-java-jbossmodules

Remaining work:

* ~Release the projects listed above and change the dependencies in Hibernate Search to target non-snapshot versions~
* ~Update the documentation to mention that dependencies need to be added to the WildFly server too~
* ~Check whether it's safe to use the EPL license in the `lucene-modules` project (Lucene itself is ASL2, so I don't see the point). Also check wether it's safe to use ASL2 in the other JBoss modules projects listed above.~
* ~Check wether the last commit in this PR is safe: I removed a dependency to `org.apache.lucene.internal` in a JBoss module. Tests pass, but I wonder why we had this dependency to begin with. If the module is named "org.apache.lucene.**internal**", is it really a good idea to depend on it? Also, if we really must depend on it, we will have to release an alias for this module as part of the `lucene-modules` project (currently there is only an alias for the main module, `org.apache.lucene`).~


Sanne:
@yrodiere I'm releasing lucene-jbossmodules as `5.5.5.Final`.

I haven't looked into the other 3 repositories yet, but let's start by transferring them to Hibernate?

Sanne:
lucene-jbossmodules released as 5.5.5.hibernate01

yrodiere:
@Sanne I added a commit to target non-snapshot versions of the module dependencies. Obviously this will only work if these versions are available in a Maven repository. I confirmed the build works locally, but maybe it will be easier to work on this ticket if we release the jboss modules?

Sanne:
Thanks, I'll build them in background here. I asked about the license, hopefully that will be fast: then we release them all.

yrodiere:
@Sanne I added a commit to load the Elasticsearch SPI as we discussed. For now we need an explicit dependency in the user deployment though, which really is a shame considering:

1. We didn't use to require this dependency (backward compatibility was broken).
2. And more importantly we still don't require this dependency in other cases, such as the JGroups/JMS backend or the Infinispan directory provider (inconsistency).

So we'll definitely need some "automatic injection rules", i.e. ways to fetch the dependency dynamically even though the user didn't declare them, based on the properties passed to Hibernate Search. I don't know how to do that though, so it's a task for another day :)

Sanne:
> For now we need an explicit dependency in the user deployment though

Why is that, it looks like you didn't remove the optional dependency to `org.hibernate.search.elasticsearch` ?


>  So we'll definitely need some "automatic injection rules" [...] I don't know how to do that though, so it's a task for another day :)

Definitely. The main problem is that that's code within WildFly so we can't update it whenever we need (nor right now). Also we will need to agree specifics with Scott Marlow.

yrodiere:
> Why is that, it looks like you didn't remove the optional dependency to org.hibernate.search.elasticsearch ?

I made a mistake while rebasing. I amended the commit, the dependency is gone now.

> Definitely. The main problem is that that's code within WildFly so we can't update it whenever we need (nor right now). Also we will need to agree specifics with Scott Marlow.

I slept on it, and I am now wondering whether this is a good idea... There are big drawbacks to having such code in WildFly:

1. We will no longer be able to just "drop in" any newer version of Hibernate Search in an older WildFly instance if there are new/removed modules, because the auto-injection mechanism will not be compatible. This is bad for users willing to experiment/use cutting-edge technology, but also for us when working on the next version of HSearch.
2. The process of upgrading Hibernate Search within WildFly will become much more fragile: the auto-injection mechanism will be one more potential point of failure.

Combined to the fact that WildFly is currently lagging way behind with its Hibernate Search version, I feel like this auto-injection mechanism will be a huge source of problems...

Now, we obviously have alternatives:

1. we could remove the optional dependencies, and just require users to explicitly add the dependencies to their deployment. The main drawback is that we would be breaking compatibility quite thoroughly.
2. or we could keep the optional dependencies, but split our JBoss modules ZIPs so that users will only have the modules they need in their WildFly instance.

I must admit this last solution feels way better to me:

* We stay consistent: all of our optional modules are handled implicitly (elasticsearch, jgroups, ...).
* We break compatibility, but only for users choosing to upgrade explicitly to a newer version of Hibernate Search. Users upgrading WildFly, in particular, will not be affected.
* Users upgrading explicitly and failing to add all the modules to WildFly will get a clear error: something along the lines of "module hibernate-search-elasticsearch:<the newer version> was not found".
* It's arguably a nice move towards WildFly Swarm users, because they won't have to bundle modules they don't need.

The main drawback, of course, is that it will get harder for users to upgrade to a newer version. But this is an advanced use case, and we're already making it more painful by moving the dependencies (GSON, es-client, ...) to separate projects.
Maybe we could address this by publishing a "full" WildFly patch, including all the Hibernate Search modules (including hsearch-elasticsearch and hsearch-jgroups and such) and all the dependencies (including GSON, es-client and such)? It arguably makes more sense in the case of a WildFly patch, since the point is to upgrade what is already in WildFly, and all the Hibernate Search modules plus their dependencies are supposed to be in WildFly.

To sum up the idea would be to have two "ways" of upgrading WildFly:

* Using fine-grained module ZIPs. This would mostly be useful with automated tools such as WildFly Swarm, or for the WildFly team itself when working on the next version of WildFly.
* Using a coarse-grained WildFly patch, including all the Hibernate Search modules and their non-standard dependencies. This would mostly be useful when upgrading Hibernate Search to a newer version in a full (non-Swarm) WildFly instance.

Sanne:
Excellent summary of the situation. The only thing I might be able to add is that for Hibernate ORM we've broken up JipiJapa (the WildFly component dealing with JPA deployment postprocessing - aka the code adding the depedencies) into multiple pieces, allowing ORM to "bring his own" adaptor.
Eventually we could explore such a thing - but I'm only mentioning it to clarify the options, not suggesting we pursue this.

At this time I believe we should go for fine-grained packages, making things easier for WildFly Swarm (which happens to be a roadmap goal).

No doubt the "WildFly full" users will have do fetch & unpack some more modules, but as you said that's "advanced users" anyway. Also I believe that while it's annoying to have to figure out one needs to download and unpack more modules, once the mechanism is put in place it's no big deal to change it to download two (or more) of the same.

Sanne:
The last commit is not safe, please remove it:

Lucene will attempt to load a compatible codec using the "current" classloader (by name) within the context of IndexReader initialization; so it needs to be able to find a compatible codec on the classpath of Hibernate Search. In most cases this will work fine even if you remove the line, but it will fail if you're opening an index which was written by a previous version of Lucene.... The Lucene project moves the out of date and "previous version" codec into the `lucene-backwards-codec` jar.

An alternative solution would be to add these jars to the main lucene module, but that would allow applications to use these classes directly as well; I'd rather keep them private.

Sanne:
Seems to work nicely, great work!

Just two final steps needed:
- [ ] Explain in the docs which other dependencies "module packs" people need to get (Looks like just section "Update and activate latest Hibernate Search version in WildFly" ?)
- [ ] Remove the last commit

yrodiere:
> The last commit is not safe, please remove it:

@Sanne Ok, then we need to re-release the Lucene modules. Currently there is no alias for the "internal" module, which is thus only available under the fully-qualified slot, not the abbreviated "5.5" slot. Which is why I had to change *something*.
I'll send a PR there.

>  Explain in the docs which other dependencies "module packs" people need to get (Looks like just section "Update and activate latest Hibernate Search version in WildFly" ?)

Right, will do.

Sanne:
> Currently there is no alias for the "internal" module, which is thus only available under the fully-qualified slot, not the abbreviated "5.5" slot. 

There is intentionally no alias to make sure it's understood that one shouldn't depend on it.. Back when we create them I thought it better to depend on the module slot directly (explicitly) as an implementation detail.

We can revisit this decision, but I'm mentioning this so you see it's not a black/white decision..

yrodiere:
@Sanne Done! I had to release verion 5.5.5.hibernate03 of the Lucene JBoss modules, but now it works. Also added the documentation changes.

yrodiere:
Addressed your comments from HipChat and merged. Thanks!

PR #1600, opened 2018-01-02
HSEARCH-2983 Take boosts into account for all queries built with the DSL + HSEARCH-2981 

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-2983
 * https://hibernate.atlassian.net/browse/HSEARCH-2981

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/analyzer/impl/NamedElasticsearchAnalyzerReference.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/analyzer/impl/NamedElasticsearchAnalyzerReference.java
@@ -35,6 +35,8 @@ public class NamedElasticsearchAnalyzerReference extends ElasticsearchAnalyzerRe
 
 	private static final Log LOG = LoggerFactory.make( MethodHandles.lookup() );
 
+	private static final ElasticsearchAnalysisDefinitionRegistryPopulator NOOP_DEFINITION_REGISTRY_POPULATOR = (r) -> { };
+
 	protected final String name;
 
 	private ElasticsearchAnalysisDefinitionRegistryPopulator definitionRegistryPopulator;
@@ -77,7 +79,7 @@ public class NamedElasticsearchAnalyzerReference extends ElasticsearchAnalyzerRe
 	protected ElasticsearchAnalysisDefinitionRegistryPopulator createRegistryPopulator(ElasticsearchAnalysisDefinitionRegistry definitionRegistry) {
 		AnalyzerDefinition analyzerDefinition = definitionRegistry.getAnalyzerDefinition( name );
 		if ( analyzerDefinition == null ) {
-			return (r) -> { }; // No-op
+			return NOOP_DEFINITION_REGISTRY_POPULATOR; // No-op
 		}
 
 		String tokenizerName = analyzerDefinition.getTokenizer();
@@ -114,8 +116,10 @@ public class NamedElasticsearchAnalyzerReference extends ElasticsearchAnalyzerRe
 		sb.append( getClass().getSimpleName() );
 		sb.append( "<" );
 		sb.append( name );
-		sb.append( "," );
-		sb.append( definitionRegistryPopulator );
+		if ( definitionRegistryPopulator != NOOP_DEFINITION_REGISTRY_POPULATOR ) {
+			sb.append( "," );
+			sb.append( definitionRegistryPopulator );
+		}
 		sb.append( ">" );
 		return sb.toString();
 	}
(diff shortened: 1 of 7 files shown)


Sanne:
Looks great, but I'll need to have a better look tomorrow. In particular I'm puzzled at why we didn't already treat the boost the way you're suggesting to do.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/analyzer/impl/NamedElasticsearchAnalyzerReference.java
```diff
@@ -35,6 +35,12 @@
 
 	private static final Log LOG = LoggerFactory.make( MethodHandles.lookup() );
 
+	private static final ElasticsearchAnalysisDefinitionRegistryPopulator NOOP_DEFINITION_REGISTRY_POPULATOR = new ElasticsearchAnalysisDefinitionRegistryPopulator() {
+		@Override
+		public void populate(ElasticsearchAnalysisDefinitionRegistry registry) {
+		}
+	};
```
yrodiere: You really don't like lambdas, do you? :) You could have used one here.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/analyzer/impl/ScopedElasticsearchAnalyzerReference.java
```diff
@@ -63,6 +63,20 @@ public CopyBuilder startCopy() {
 		return new Builder( globalAnalyzerReference, scopedAnalyzerReferences );
 	}
 
+	@Override
+	public String toString() {
```
yrodiere: This is a good idea, but don't you think we should also add the names of the targeted fields in `RemoteMatchQuery#toString()`? From what I see [here](https://forum.hibernate.org/viewtopic.php?f=9&t=1045140) the field name isn't printed out, and it seems like an even greater problem to me... We should probably have a look at `org.apache.lucene.search.TermQuery#toString(String)` and use the same principles in `RemoteMatchQuery#toString()`.

engine/src/test/java/org/hibernate/search/test/dsl/Coffee.java
```diff
@@ -48,6 +49,8 @@ public void setId(String id) {
 	public void setDescription(String description) { this.description = description; }
 	private String description;
 
+	@Field
+	@NumericField
```
yrodiere: `@NumericField` is not necessary here, the field should be detected as numeric automatically.

engine/src/main/java/org/hibernate/search/query/dsl/impl/ConnectedMultiFieldsTermQueryBuilder.java
```diff
@@ -107,19 +113,17 @@ private Query createQuery(FieldContext fieldContext, ConversionContext conversio
 		final String searchTerm = buildSearchTerm( fieldContext, documentBuilder, conversionContext );
 
 		if ( !applyTokenization || BridgeAdaptorUtils.unwrapAdaptorOnly( fieldBridge, IgnoreAnalyzerBridge.class ) != null ) {
-			perFieldQuery = createTermQuery( fieldContext, searchTerm );
+			return createTermQuery( fieldContext, searchTerm );
+		}
+
+		// we need to build differentiated queries depending of if the search terms should be analyzed
```
yrodiere: Since it's you, I'm sure you'll understand :D : could you please replace "of if" with "on if" while you're at it?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/analyzer/impl/NamedElasticsearchAnalyzerReference.java
gsmet: You know, I'm an old man now :). Fixed.

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/analyzer/impl/ScopedElasticsearchAnalyzerReference.java
gsmet: Mmmh, yes, somehow missed it. Will fix too.

engine/src/test/java/org/hibernate/search/test/dsl/Coffee.java
gsmet: Yes, that was my recollection but I wanted to be on the safe side. Removed.

Sanne:
Very nice. Merging!

PR #1601, opened 2018-01-04
HSEARCH-2972 + HSEARCH-2984 Fix dependencies in the distribution ZIP

yrodiere:
* https://hibernate.atlassian.net//browse/HSEARCH-2972 Missing provided dependencies in the distribution ZIP
* https://hibernate.atlassian.net/browse/HSEARCH-2984 Remove an unnecessary jberet dependency in the distribution module

The added dependencies are licensed under ASL2, so this should be fine.


Sanne:
merged, thanks!

PR #1602, opened 2018-01-10
HSEARCH-2955 Faceting query fails when the facet field is not populated for any document

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2955

### DIFF
--- a/engine/src/main/java/org/hibernate/search/query/engine/impl/QueryHits.java
+++ b/engine/src/main/java/org/hibernate/search/query/engine/impl/QueryHits.java
@@ -79,6 +79,10 @@ public class QueryHits {
 	private final LazyQueryState searcher;
 	private final QueryFilters filters;
 	private final Sort sort;
+	/**
+	 * True to collect hits, false to only collect hit count.
+	 */
+	private final boolean collectHits;
 	private final Map<FacetingRequest, FacetMetadata> facetingRequestsAndMetadata;
 	private final TimeoutManagerImpl timeoutManager;
 
@@ -94,6 +98,7 @@ public class QueryHits {
 	public QueryHits(LazyQueryState searcher,
 			QueryFilters filters,
 			Sort sort,
+			boolean collectHits,
 			TimeoutManagerImpl timeoutManager,
 			Map<FacetingRequest, FacetMetadata> facetingRequestsAndMetadata,
 			Coordinates spatialSearchCenter,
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/hibernate/search/query/engine/impl/QueryHits.java
```diff
@@ -399,13 +399,20 @@ private FacetResult getFacetResultForLongRange(RangeFacetRequest<?> facetRequest
...
 		SortedSetDocValuesReaderState docValuesReaderState;
 		try {
 			docValuesReaderState = new DefaultSortedSetDocValuesReaderState( searcher.getIndexReader() );
 		}
 		catch (IllegalArgumentException e) {
-			// happens in case there are no facets at all configured for the matching documents
-			throw log.unknownFieldNameForFaceting( facetRequest.getFacetingName(), facetRequest.getFieldName() );
+			/*
+			 * Happens in case there are no facets at all stored in the matching documents.
+			 * But we know some the target field is correctly configured to generate facets,
```
gsmet: s/some// ?

engine/src/main/java/org/hibernate/search/query/engine/impl/QueryHits.java
```diff
@@ -399,13 +399,20 @@ private FacetResult getFacetResultForLongRange(RangeFacetRequest<?> facetRequest
...
+			 * because we managed to retrieve the FacetMetadata.
+			 * So we can safely return an empty list: the matching documents simply do not have
+			 * any value for this field.
+			 */
+			return facets;
```
gsmet: Can't we use `Collections.emptyList()` and keep the array list initialization where it was? I think the list is readonly after that? Let's not initialize an array list for the sake of it.

yrodiere: Well the facet list we return to the user has always been mutable, and I'm dead certain some people must rely on it.
Plus the calling code mutates the list in some cases. Not when it's empty, granted, but still I would feel bad returning an immutable list.

So... Not in a minor?

engine/src/main/java/org/hibernate/search/query/engine/impl/QueryHits.java
yrodiere: Right, done.

engine/src/main/java/org/hibernate/search/query/engine/impl/QueryHits.java
gsmet: Right, I thought we would have made them immutable at some point.

Merged!

PR #1603, opened 2018-01-11
HSEARCH-2986 Put the distribution files to upload into distribution/target/dir so that release scripts find them

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2986

### DIFF
--- a/distribution/src/main/assembly/dist.xml
+++ b/distribution/src/main/assembly/dist.xml
@@ -18,7 +18,10 @@
     <formats>
         <format>tar.gz</format>
         <format>zip</format>
+        <!-- Uncomment the following line to simplify content inspection: -->
+        <!--
         <format>dir</format>
+        -->
     </formats>
 
     <!-- Configure the module artifacts which make it into the distribution bundle -->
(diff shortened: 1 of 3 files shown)


distribution/src/main/assembly/dist.xml
```diff
@@ -18,7 +18,10 @@
     <formats>
         <format>tar.gz</format>
         <format>zip</format>
+        <!-- Uncomment the following line to simplify content inspection: -->
+        <!--
```
Sanne: Thanks, I wanted to do this as well :)

Sanne:
Looks good. The filter with "zip", that's on top of additional artifacts like the README, etc. right?

yrodiere:
Yes, basically we put whatever files we want to upload into `distribution/target/dist`, and the release scripts will upload that plus the README etc.
See https://github.com/hibernate/hibernate-noorm-release-scripts/pull/4 for more information.
Maybe do not merge this PR right now, there seems to be some unexpected controversy on the changes I planned for the release scripts...

yrodiere:
The release scripts have been updated, we're good to go! Merging...

PR #820, opened 2015-04-15
Create hibernate-search feature pack

ctomc:
@Sanne lets talk about this.

There is also second part on wildfly build side that i am still working on.
WIP branch for inclusion in wildfly https://github.com/ctomc/wildfly/tree/hsearch-feature-pack

PR #1607, opened 2018-01-19
HSEARCH-2989 + HSEARCH-2990 + HSEARCH-2191 Make Hibernate Search Modules a valid Wildfly Feature pack file

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2989
https://hibernate.atlassian.net//browse/HSEARCH-2990

Noteworthy:

* JBoss modules are now packaged as "feature packs", i.e. ZIPs that only include the metadata allowing to retrieve the dependencies (other feature packs + JARs) from Maven
* JBoss modules will **not** published as ZIPs containing the full modules (metadata + JARs) anymore. In particular, JBoss modules will not be published to Sourceforge anymore.
* Moving to feature packs was going to require actions from users anyway, so I renamed the main WildFly-related Maven artifact from `hibernate-search-modules` to `hibernate-search-jbossmodules-engine` and removed the ORM parts from it, . Removing the ORM integration will allow easier consumption by Infinispan in particular. The Hibernate-Search / ORM integration is now packaged as a separate feature pack in `hibernate-search-jbossmodules-orm`.
* Since Hibernate ORM does not provide a feature pack yet, I packaged ORM as a feature pack in module `hibernate-search-jbossmodules-hibernate-orm`. Users should not have to know about this, because the dependencies between feature packs are transitive.
* I bumped the versions of our dependencies to external JBoss modules (Lucene, Elasticsearch clients, ...) and released those, since the previous releases did not include feature packs and we need them. Please allow a few minutes until the JBoss Nexus repository syncs up.

Also, at the moment the feature packs use versioned slots, and not the "main" slot as in the [original pull request](https://github.com/hibernate/hibernate-search/pull/820/). I am not sure if using the main slots would be a good idea, since the `wildfly-feature-pack` we depend on already includes those modules...
Maybe once WildFly is properly split up in separate feature packs we will be able to just use the main slot and expect users to only have one version of Hibernate Search in their application. For the time being, it doesn't look easy.


Sanne:
Great work! I've merged it and marked the issues resolved, but will send an additional PR with some additional touches. They are optional, so while I'll reuse the HSEARCH-2191 I'm considering it solved already.

PR #1608, opened 2018-01-21
Follow up on  HSEARCH-2191

Sanne:


jbossmodules/elasticsearch-aws/pom.xml
```diff
@@ -27,16 +27,34 @@
...
+                    <artifactId>*</artifactId>
+                    <groupId>*</groupId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>hibernate-search-jbossmodules-elasticsearch</artifactId>
             <type>zip</type>
+            <exclusions>
```
yrodiere: Ok for the non-JBoss modules dependencies, but JBoss modules already have their transitive dependencies excluded in the parent POM's dependency management section.
I'm ok to move the exclusions here if you think it's clearer, but then maybe we should remove the exclusions from the parent POM?
The only other place where we have dependencies to the JBoss modules is in the dependencies section of the `wildfly-server-provisioning-maven-plugin` configuration in the integration tests, and I suspect dependency manangement doesn't apply there. And anyway, we don't need it there since the plugin will ultimately download the transitive dependencies.

documentation/src/main/asciidoc/configuration.asciidoc
```diff
@@ -1516,9 +1516,11 @@ In that case, you should https://developer.jboss.org/wiki/MavenGettingStarted-Us
 [[provisioning-wildfly-nonmaven]]
 ===== Server provisioning via other build tools
 
-Plugins for other build tools have not been implemented yet, but this should be quite straight forward to do:
-the above Maven plugin is just a thin wrapper invoking other libraries;
-these other libraries are build agnostic and are responsible for performing most of the work.
+The provisioning tools have been developed mainly for Maven users, but there are options for users of other build tools too.
+
+Gradle users can use the link:https://plugins.gradle.org/plugin/org.wildfly.build.provision[org.wildfly.build.provision] plugin, available on the official Gradle plugin portal.
```
yrodiere: Nice work :) Seems like your weekend was productive!

Sanne:
> If you agree, I will remove the transitive dependencies exclusions for JBoss modules in the parent POM though. They are no longer necessary. See my comment.

As you prefer. I think I'd rather keep it, as many things might happen: stuff will chance, people will copy paste this as an example to make new modules, etc.. Hopefully we'll get better control on what versions are really being included in the feature pack definition so I wouldn't look at this as the definitive structure.

yrodiere:
Merged, thanks!

PR #1609, opened 2018-01-22
 HSEARCH-2191 follow-up: Include ORM specifics in the WildFly configuration doc

yrodiere:

Simply referring to the ORM documentation for WildFly isn't enough
anymore, since with Hibernate Search we install ORM modules differently.
So we'd better explain ORM specifics as well.


Sanne:
Thanks, good idea. WDYT of changing "activate" to "select" ?  I'm thinking activation as a concept might confuse people, as the module isn't _activated_ globally but rather chosen on a per-deployment base.

yrodiere:
> WDYT of changing "activate" to "select" ? I'm thinking activation as a concept might confuse people, as the module isn't activated globally but rather chosen on a per-deployment base.

Good idea, I'll change that and push

PR #1610, opened 2018-01-22
HSEARCH-2191 follow-up: Remove the now-unused property "hibernate-search.module.slot"

yrodiere:
Everything is now defined in the "jbossmodules" POM, and is derived from
properties set in the parent POM.


Sanne:
merged!

PR #1612, opened 2018-01-23
HSEARCH-2880 Improve documentation regarding field bridges on IDs

yrodiere:
... and the documentation of two-way bridges in general while I'm at it.

https://hibernate.atlassian.net//browse/HSEARCH-2880

### DIFF
--- a/documentation/src/main/asciidoc/mapping.asciidoc
+++ b/documentation/src/main/asciidoc/mapping.asciidoc
@@ -1859,64 +1859,7 @@ An example is a bridge that deals with enums in a custom fashion but needs to ac
 type. Any bridge implementing `AppliedOnTypeAwareBridge` will get the type the bridge is applied on
 injected. Like parameters, the type injected needs no particular care with regard to thread-safety.
 
-[[section-two-way-bridge]]
-====== Two-way bridge
-
-If you expect to use your bridge implementation on an id property (ie annotated with `@DocumentId`
-), you need to use a slightly extended version of `StringBridge` named `TwoWayStringBridge`. Hibernate
-Search needs to read the string representation of the identifier and generate the object out of it.
-There is no difference in the way the `@FieldBridge` annotation is used.
-
-
-.Implementing a `TwoWayStringBridge` usable for id properties
-====
-[source, JAVA]
-----
-public class PaddedIntegerBridge implements TwoWayStringBridge, ParameterizedBridge {
-
-    public static String PADDING_PROPERTY = "padding";
-    private int padding = 5; //default
-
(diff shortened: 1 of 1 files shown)


documentation/src/main/asciidoc/mapping.asciidoc
```diff
@@ -437,16 +437,28 @@ For instance you may leave this to the default (`false`) during testing to ident
...
+===== Document identifier: @DocumentId or @Id
 
-Finally, the id property of an entity is a special property used by Hibernate Search to ensure index
-unicity of a given entity. By design, an id has to be stored and must not be tokenized. It is also
-always string encoded, even if the id is a number. To mark a property as index id, use the
-`@DocumentId` annotation. If you are using JPA and you are using `@Id` you can omit `@DocumentId`.
-The chosen entity id will also be used as document id.
+Finally, the document identifier is a special field used by Hibernate Search to ensure
+there will only ever be one document per entity in the index,
+and to retrieve documents when they should be deleted.
```
Sanne: s/retrieve/identify/ ?

I wouldn't suggest we actually "retrieve" a document to delete it

documentation/src/main/asciidoc/mapping.asciidoc
```diff
@@ -456,32 +468,43 @@ public class Essay {
     ...
 
     @Id
-    @DocumentId
+    @GeneratedValue
```
Sanne: Feel free to keep this, just a though to share:
I typically assume it's ok in our examples to assume "application assigned ids", so to not have to add annotations which might be distracting / non essential to the example.

documentation/src/main/asciidoc/mapping.asciidoc
```diff
@@ -456,32 +468,43 @@ public class Essay {
...
+    ...
 }
 ----
 ====
 
-<<example-annotated-entity>> defines an index with four fields: `id`, `Abstract`, `text` and
-`grade`. Note that by default the field name is de-capitalized, following the JavaBean
-specification. The `grade` field is annotated as numeric with a slightly larger `precisionStep` than
-the default.
+If you don't use JPA,
```
Sanne: We had set some style "house rules":
 - any period should be on a single line, with exceptions of "within reason" to break very long ones.
 - we don't like the style of "it is recommended". Better to use "we suggest"

I see it's gradually being forgotten as I'm getting non-picky so today I'll be extra demanding as you might not even ever have heard of these :)

Sanne:
thanks! Ok to merge

Sanne:
merged

PR #1613, opened 2018-01-23
HSEARCH-2991 Avoid needing jsr250-api during project build

Sanne:
https://hibernate.atlassian.net/browse/HSEARCH-2991

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1208,7 +1208,9 @@
                                 <processors>
                                     <processor>org.jboss.logging.processor.apt.LoggingToolsProcessor</processor>
                                 </processors>
-                                <compilerArguments>-source 1.8 -target 1.8</compilerArguments>
+                                <!-- Disable usage of the @Generated annotation: too much trouble as requirements are different across JDK8 vs JDK9 -->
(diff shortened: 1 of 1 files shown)


yrodiere:
@Sanne I just tried it, and the `@Generated` annotation still gets generated:

```java
/**
 * Warning this class consists of generated code.
 */
@Generated(value = "org.jboss.logging.processor.generator.model.MessageLoggerImplementor", date = "2018-01-23T14:20:13+0100")
public class BaseHibernateSearchLogger_$logger extends DelegatingBasicLogger implements BaseHibernateSearchLogger, BasicLogger, Serializable {
    private static final long serialVersionUID = 1L;
    private static final String FQCN = BaseHibernateSearchLogger_$logger.class.getName();
```

Or was it about another `@Generated` annotation, from another package than `javax.annotation`?

gsmet:
Nope, it's this very one.

Sanne:
Thanks for checking :) It worked in JDK9 but that's for other reasons: it automatically apply "the right" kind of Generated annotation now.

Fixed the PR, a I still want to get rid of this for the sake of tooling.

yrodiere:
Merged, thanks!

PR #1614, opened 2018-01-23
HSEARCH-2988 Make sure Hibernate Search can be used without relying on JBoss's Nexus repo

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2988

I created a CI job running the build with a default repository configuration (only Maven Central): http://ci.hibernate.org/view/Search/job/hibernate-search-master-mavencentral/

The HEAD of this PR builds fine: http://ci.hibernate.org/view/Search/job/hibernate-search-master-mavencentral/2/
And adding a JBoss-only dependency makes the build fail as expected: http://ci.hibernate.org/view/Search/job/hibernate-search-master-mavencentral/4/

Since the job requires downloading each and every dependency from Maven central, and does not re-use the local Maven repo from previous executions (to avoid false positives), I made sure to throttle it to only run once a day, and only if something changed on master. We can adjust that if needed.

### DIFF
--- a/integrationtest/performance/orm/pom.xml
+++ b/integrationtest/performance/orm/pom.xml
@@ -310,4 +310,61 @@
         </profile>
     </profiles>
 
+    <!--
+        Enable a hard-coded JBoss repository configuration in WildFly-related tests.
+        This allows us to run a CI job with default Maven settings,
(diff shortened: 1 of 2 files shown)


Sanne:
Great, thanks for doing all those cross-checks. Merged!

PR #1620, opened 2018-01-31
HSEARCH-2999 Upgrade to Elasticsearch 5.6.7

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2999


yrodiere:
We should also update the Elasticsearch test version; actually that's probably the most important part of the upgrade: from what I saw, the Java REST client rarely changes.
I'll update it and push a commit.

Sanne:
thanks, apparently I missed that at the previous round as well.

Merged.

PR #1622, opened 2018-02-08
Follow-up to HSEARCH-2978 - Use the new automatic module name for Hibernate ORM 5.3 in JDK9 tests

yrodiere:
Follow-up to the previous PR #1598, because the JDK9 build started to fail.

@Sanne On a related note, the Hibernate ORM team defined module names explicitly in their manifest. Maybe we should do it for Search, too? Would org.hibernate.search.* work for you?


yrodiere:
Succeeding build on JDK9: http://ci.hibernate.org/view/Personal%20runs/job/hibernate-search-yoann-jdk9/3/

Sanne:
Yes let's do that as well. However there have been some strong opinions in the community about naming conventions - I don't remember the details now but we'd need to verify thoser before picking names.

>> https://hibernate.atlassian.net/browse/HSEARCH-3000

Merged, thanks!

PR #1627, opened 2018-02-14
HSEARCH-3006 Make the version of the Surefire Maven plugin a property

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-3006
 - https://hibernate.atlassian.net/browse/HSEARCH-3007

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -196,7 +196,7 @@
         <jandexVersion>2.0.3.Final</jandexVersion>
         <classMateVersion>1.3.0</classMateVersion>
         <javassistVersion>3.22.0-GA</javassistVersion>
-        <bytebuddyVersion>1.6.14</bytebuddyVersion>
+        <bytebuddyVersion>1.7.9</bytebuddyVersion>
 
         <!-- OSGi ranges for some primary dependencies -->
         <!-- Warning:
--- a/pom.xml
+++ b/pom.xml
@@ -181,6 +181,11 @@
         <!-- Can be overridden by subprojects if dependency convergence cannot be achieved -->
         <enforcer.dependencyconvergence.skip>false</enforcer.dependencyconvergence.skip>
 
+        <!-- Maven plugin versions -->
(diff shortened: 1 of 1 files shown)


Sanne:
Wait a second, I'm going to send an improved version to actually use the right defaults with profiles.

Sanne:
Updated.

pom.xml
```diff
@@ -1878,20 +1878,44 @@ org.hibernate.search.spi.SearchIntegratorBuilder#buildSearchIntegrator()
...
+                <maven.plugin.failsafe.version>2.20.1</maven.plugin.failsafe.version>
+            </properties>
             <modules>
                 <module>documentation</module>
             </modules>
         </profile>
 
         <profile>
-            <id>jigsaw</id>
+            <id>jdk9</id>
```
yrodiere: Ok if I rename the "non-jigsaw" profile to jdk8 then?

pom.xml
```diff
@@ -1878,20 +1878,44 @@ org.hibernate.search.spi.SearchIntegratorBuilder#buildSearchIntegrator()
                 <!-- Asciidoctor and Karaf won't work on JDK9 yet  -->
                 <jdk>1.8</jdk>
             </activation>
+            <properties>
+                <!-- Surefire versions are a minefield of bugs, so depending on context we need to choose a different one -->
+                <maven.plugin.surefire.version>2.20.1</maven.plugin.surefire.version>
+                <maven.plugin.failsafe.version>2.20.1</maven.plugin.failsafe.version>
```
yrodiere: Would you mind if I simply made this the default, and removed the overrides for jdk8 and jdk9? I'm afraid we'll forget to update both the next time we upgrade...

pom.xml
Sanne: ok

pom.xml
Sanne: ok. The reason I didn't set a default is to catch mistakes, I never trust its "inheritance rules" ;)

yrodiere:
Merged, thanks!

PR #1631, opened 2018-02-15
HSEARCH-3011 Re-enable building the documentation on all JDKs

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-3011
 - https://hibernate.atlassian.net/browse/HSEARCH-3012

### DIFF
--- a/documentation/pom.xml
+++ b/documentation/pom.xml
@@ -210,7 +210,6 @@
                                         <pdf-style>hibernate</pdf-style>
                                         <pdf-fontsdir>${asciidoctor.aggregated-resources-dir}/theme/fonts</pdf-fontsdir>
                                         <pagenums/>
-                                        <idprefix/>
                                         <idseparator>-</idseparator>
                                     </attributes>
                                 </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -292,7 +292,7 @@
         <puppycrawl.checkstyle.version>8.8</puppycrawl.checkstyle.version>
 
         <!-- Asciidoctor -->
-        <hibernate-asciidoctor-theme.version>1.0.1.Final</hibernate-asciidoctor-theme.version>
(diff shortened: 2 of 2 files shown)


documentation/pom.xml
```diff
@@ -210,7 +210,6 @@
                                         <pdf-style>hibernate</pdf-style>
                                         <pdf-fontsdir>${asciidoctor.aggregated-resources-dir}/theme/fonts</pdf-fontsdir>
                                         <pagenums/>
-                                        <idprefix/>
```
gsmet: I think this change will break all the unqualified anchors which doesn't sound as a good idea.

Is there a reason why you removed it? If there is an issue with Asciidoctor, we should open a bug there.

pom.xml
```diff
@@ -292,10 +293,10 @@
 
         <!-- Asciidoctor -->
         <hibernate-asciidoctor-theme.version>1.0.1.Final</hibernate-asciidoctor-theme.version>
```
gsmet: While we're at it, let's update the theme to 1.0.2.Final. It should improve the code formatting in the PDF doc.

documentation/pom.xml
Sanne: See the discussions I linked to.. also, this is only PDF so it's not going to break any links.

pom.xml
Sanne: ok will do, but that's a separate issue.

documentation/pom.xml
gsmet: Ah, yes sorry, I didn't have the full context :).

pom.xml
Sanne: Added a commit for that ;)

gsmet:
I checked the PDF and HTML outputs. Everything looks good, will merge.

PR #1632, opened 2018-02-16
HSEARCH-3017 Upgrade to Hibernate ORM 5.3.0.CR1

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3017

The dependency versions in Hibernate-core did not change.


gsmet:
Merged, thanks!

PR #1634, opened 2018-02-20
Provide ability to customize parser in TikaBridge

nikowitt:
Hi there,

while updating to a more recent version of Tika, we cannot rely on the default AutoDetectParser anymore, as we need to define a custom Tika instance with a special tika-config.xml to drop a specific parser (GrobIdParser). To reflect this customization in the TikaBridge as well, I need to provide a custom TikaBridge that should just extend from the built-in bridge, just use the previously customized parser.

Thank you for any feedback,

Regards,
Niko

### DIFF
--- a/engine/src/main/java/org/hibernate/search/bridge/builtin/TikaBridge.java
+++ b/engine/src/main/java/org/hibernate/search/bridge/builtin/TikaBridge.java
@@ -32,6 +32,7 @@ import org.hibernate.search.util.impl.ClassLoaderHelper;
 import org.hibernate.search.util.logging.impl.Log;
 import org.hibernate.search.util.logging.impl.LoggerFactory;
 import java.lang.invoke.MethodHandles;
+import java.util.Objects;
 
 import static org.apache.tika.io.IOUtils.closeQuietly;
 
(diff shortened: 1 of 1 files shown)


yrodiere:
Ok to test

yrodiere:
@nikowitt This is interesting, but unfortunately your solution would prevent you (or anyone else interested in this feature) from using the `@TikaBridge` annotation or the equivalent feature in the programmatic mapping. You would have to configure and apply the `TikaBridge` manually, which may be fine for some use cases, but probably not all of them.

Do you really need to pass a parser instance? Would the ability to provide the path of the `tika-config,xml` on the classpath work for you? We could simply add a parameter to the annotation, e.g. `@TikaBridge(config = "/my-dir/tika-config.xml")`.
Alternatively, if you really cannot use `AutoDetectParser`, we could introduce a `TikaParserProvider` alongside the `TikaMetadataProcessor` and `TikaParseContextProvider` (see `org.hibernate.search.annotations.TikaBridge`). This class would expose a factory method returning a parser instance that would then be used by the `TikaParser`.

In any case, we will need a ticket before we merge anything :) Could you please create one on [our JIRA instance](https://hibernate.atlassian.net/projects/HSEARCH/issues), and continue the conversation there? We will be notified automatically and will answer ASAP.

Sanne:
+1 for a `TikaParserProvider`, so that users can inject a custom one at configuration time.

Niko: thank you so much for letting us know about the problem.

nikowitt:
Thank you for the quick feedback! Sure, I will create a JIRA tasks for that one.

yrodiere:
Thank you very much @nikowitt , let's continue the discussion on the ticket you created. I will submit another PR when necessary, so I'm going to close this one.

PR #1635, opened 2018-02-20
HSEARCH-3019 Provide ability to customize parser in TikaBridge

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3019

Ready for review if the build succeeds.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/bridge/builtin/TikaBridge.java
+++ b/engine/src/main/java/org/hibernate/search/bridge/builtin/TikaBridge.java
@@ -22,11 +22,14 @@ import org.apache.tika.parser.AutoDetectParser;
 import org.apache.tika.parser.ParseContext;
 import org.apache.tika.parser.Parser;
 import org.apache.tika.sax.WriteOutContentHandler;
+
+import org.hibernate.search.bridge.AppliedOnTypeAwareBridge;
 import org.hibernate.search.bridge.LuceneOptions;
 import org.hibernate.search.bridge.MetadataProvidingFieldBridge;
 import org.hibernate.search.bridge.MetadataProvidingTikaMetadataProcessor;
 import org.hibernate.search.bridge.TikaMetadataProcessor;
 import org.hibernate.search.bridge.TikaParseContextProvider;
+import org.hibernate.search.bridge.TikaParserProvider;
 import org.hibernate.search.bridge.spi.FieldMetadataBuilder;
 import org.hibernate.search.util.impl.ClassLoaderHelper;
 import org.hibernate.search.util.logging.impl.Log;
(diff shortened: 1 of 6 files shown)


engine/src/main/java/org/hibernate/search/bridge/builtin/TikaBridge.java
```diff
@@ -182,6 +209,13 @@ private FileInputStream openInputStream(File file) {
 		}
 	}
 
+	private static class AutoDetectParserProvider implements TikaParserProvider {
```
gsmet: Couldn't we have only one instance of this one? TikaBridge could be used several times.

Wondering also if we could have only one instance of the parser for all the TikaBridges (IIRC the metadata are read sequentially so we shouldn't have concurrent accesses?).

engine/src/main/java/org/hibernate/search/bridge/builtin/TikaBridge.java
```diff
@@ -40,28 +43,52 @@
...
 	private TikaParseContextProvider parseContextProvider;
 
 	public TikaBridge() {
+		setParserProviderClass( null );
 		setMetadataProcessorClass( null );
 		setParseContextProviderClass( null );
 	}
 
+	@Override
+	public void setAppliedOnType(Class<?> returnType) {
```
gsmet: Don't know if it's already planned for 6 but having a proper hook would help :).

gsmet:
Added a few minor comments inline.

To be honest, I'm wondering if we should backport it to 5.9. I expect people won't be able to upgrade to 5.10 and ORM 5.3 just after the release and this patch seems to be useful and very low risk.

engine/src/main/java/org/hibernate/search/bridge/builtin/TikaBridge.java
gsmet: > Not sure it's obvious, as the AutoDetectParser relies on system properties and environment variables for its configuration, and those can change. We can create a ticket if you want, but I wouldn't try that without serious regression testing.

Yeah, OK, considering I would like to consider backporting it, let's be conservative :).

engine/src/main/java/org/hibernate/search/bridge/builtin/TikaBridge.java
yrodiere: It sure would. I didn't feel like adding such a think in 5... Having one interface per hook is already such a mess :)
In the Search 6 we don't really need a specific hook, because:

1. Binding aside, bridges are immutable. Parameters, if any, are processed by the bridge builder, and the builder can execute such initialization code in the build method (or you can execute it in the bridge constructor).
2. If push comes to shove, you already have a bind() method where you can execute such initialization code.

gsmet: Ok, cool, just wanted to check we were covered for 6!

yrodiere:
@gsmet I addressed your comments, thanks. I'll let @Sanne give us his opinion on the last commit, since I know he's not a fan of static variables :) But in this case I think it's okay.
Ok for me to backport to 5.9.

engine/src/main/java/org/hibernate/search/bridge/builtin/TikaBridge.java
```diff
@@ -210,13 +210,16 @@ private FileInputStream openInputStream(File file) {
 	}
 
 	private static class AutoDetectParserProvider implements TikaParserProvider {
+		private static AutoDetectParserProvider INSTANCE = new AutoDetectParserProvider();
+
```
gsmet: You invoking Sanne made me think maybe it's not such a good idea considering you are throwing it just after you used it and it's not kept at runtime.

It might make more sense for the other 2.

Sanne:
> I'll let @Sanne give us his opinion on the last commit, since I know he's not a fan of static variables :) But in this case I think it's okay.

They would look better if you could make them `final` as well.

Other than that I have no problem with it, but yea it looks a bit overkill. I tend to like the `INSTANCE` pattern for stuff which is otherwise heavily instantiated, but I don't think it's worth their cost in bytecode (and code metadata) when they are likely only created a couple of these at bootstrap. It's a matter of tradeoffs ;) As Guillaume mentioned, it's likely thrown out after bootstrap.. but it won't as you made it static. No problem at all though, just explaining why I don't like them much in such case but I understand it's far fetched and maybe unreasonable :+1:

Sanne:
P.S. I wouldn't backport to 5.9 unless there is specific strong demand, I understand it's tempting but so is any new feature. As a rule, we don't backport new features in micros.

We definitely don't backport a new API right away. Let it hit master first and gauge some feedback; whis specific issue would be marked resolved on JIRA, we can always re-evaluate later with a separate JIRA which is specific for backports.

yrodiere:
Ok, well, let's forget about these static instances, that probably won't bring us much. And yeah I forgot the `final`, sorry about that.
I removed the last commit.

Sanne:
merged. Thanks!

PR #1638, opened 2018-02-22
Amend the changelog to mention HSEARCH-3019 in 5.10

yrodiere:
The ticket didn't have the right fix version when I released...


Sanne:
I mean: my bad. Should not do unusual things.

yrodiere:
Not problem, I actually remember setting the initial fix version myself, so I should have thought about it ;)

PR #1640, opened 2018-03-05
HSEARCH-3013 HSEARCH-3015 Update JGroups and Classmate dependencies

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-3013
 * https://hibernate.atlassian.net/browse/HSEARCH-3015


Sanne:
updated, thanks! FYI also exploring update to JGroups 4.0.10, but it's a complex one so not sure if we'll make it.

gsmet:
Yes, I saw the issue but it was my feeling that it wouldn't be that easy. As the update was easy, I made it anyway.

PR #1643, opened 2018-03-15
Add profile for the maven-dependency-check plugin

gustavonalle:
This is a plugin we're now using on Infinispan to spot CVEs in case you're interested. 

The report is saved on ```target/dependency-check-report.html``` and can be generated with 
```mvn -Powasp clean install -DskipTests=true```


Sanne:
thanks, interesting. Coincidence: I was just at a security related meetup yesterday night and they persuaded me into trying their tool which does something similar. Let's start with this though :+1:

PR #1644, opened 2018-03-22
HSEARCH-3032 Add test coverage analysis in an optional build profile

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3032

To generate a report, run `mvn clean install -Pcoverage`. You can then find the report in `reports/target/site/jacoco-aggregate/index.html`.

I added Coveralls support in the `.travis.yml`, but since some of us are not fans of continuous code coverage analysis, I will not configure it in Jenkins. If you want to enable coveralls on your fork, log into https://coveralls.io/ and add your repo.
Alternatively, we could go for https://codecov.io/, but I'm not sure we want to execute a bash script downloaded from the web in our CI :]
See https://coveralls.io/github/yrodiere/hibernate-search for an idea of what reports look like in Coveralls. There seems to be a problem with the tree view, but that has to do with Coveralls (the JaCoCo report is fine).

Regarding the problem I was originally interested in... Indeed, we do not need, nor test, executing delete works in a "stream" fashion: https://coveralls.io/builds/16121711/source?filename=engine%2Fsrc%2Fmain%2Fjava%2Forg%2Fhibernate%2Fsearch%2Fbackend%2Fimpl%2FStreamingOperationExecutorSelector.java#L62 . But as Sanne mentioned, this may be used in Infinispan.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ env:
 before_install:
   # Build options must be set before install, so that
   # we properly retrieve *all* the necessary dependencies
-  - BUILD_OPTIONS='-Pdist'
+  - BUILD_OPTIONS='-Pdist -Pcoverage'
   - case $DB in
     "h2") ;;
     "pgsql") BUILD_OPTIONS+=' -Pci-postgresql' ;;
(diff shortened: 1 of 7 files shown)


Sanne:
nice! merged, thanks

PR #1645, opened 2018-03-22
HSEARCH-3033 Internal links to "section-custom-similarity" in the documentation are dead

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3033


Sanne:
thanks!

PR #1646, opened 2018-03-26
HSEARCH-3034 Replace fest-assert with AssertJ

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-3034

In `FutureAssert`, we lose the cause but I haven't found anything in AssertJ allowing to pass the cause.

We could either throw the exception and have the message as a comment or write custom failure handling. Not sure it's worth it.

Another option could be to contribute to AssertJ and allow to pass the cause to `failWithMessage()` (or more probably `failWithMessageAndCause()` as the methods would conflict).

### DIFF
--- a/elasticsearch/pom.xml
+++ b/elasticsearch/pom.xml
@@ -81,8 +81,8 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.easytesting</groupId>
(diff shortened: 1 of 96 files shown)


yrodiere:
I added a commit to fix the stacktrace issue as discussed on HipChat. Reviewing now.

yrodiere:
Merged, thanks!

PR #1649, opened 2018-03-26
HSEARCH-3035 Use source and target properties for JBoss logging processor

gsmet:
https://hibernate.atlassian.net/browse/HSEARCH-3035

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1365,7 +1365,7 @@
                                 </processors>
                                 <!-- Disable usage of the @Generated annotation: too much trouble as requirements are different across JDK8 vs JDK9 -->
                                 <!-- See also: https://github.com/jboss-logging/jboss-logging-tools/pull/60#issuecomment-293088198 -->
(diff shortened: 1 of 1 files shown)


gsmet:
The build failed because the new version of hibernate common annotations is not yet synced to Central.

Sanne:
Why do you need this?

gsmet:
Why do I need what? This commit? I thought it would be better to use the properties.

Sanne:
I'm wondering what's the reason to force the annotation processor to use the same as the main compiler version; while I understand that typically they are the same this isn't a given, so I'm curious why you feel this is a change which needs to be done.

Sanne:
(BTW just being curious - I see no problem with it either, so go ahead if that's what you prefer)

yrodiere:
Merged, thanks!

gsmet:
@Sanne Well, it's mostly to avoid updating them when we will update the source/target versions at some point.

We can still use specific versions if need be but in the default case they will be consistent, which I think is an improvement.

Sanne:
Ok. Just remember not all annotation processors work equally well on all JDKs ;)

PR #1650, opened 2018-03-30
HSEARCH-3039 Register a pass through analyzer for the id field

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-3039

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
+++ b/engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
@@ -385,6 +385,11 @@ public class AnnotationMetadataProvider implements MetadataProvider {
 
 		Field.TermVector termVector = AnnotationProcessingHelper.getTermVector( TermVector.NO );
 
+		if ( !parseContext.skipAnalyzers() ) {
+			// Register a pass through analyzer for the document id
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/hibernate/search/engine/metadata/impl/AnnotationMetadataProvider.java
```diff
@@ -385,6 +385,11 @@ private void createIdPropertyMetadata(XProperty member,
 
 		Field.TermVector termVector = AnnotationProcessingHelper.getTermVector( TermVector.NO );
 
+		if ( !parseContext.skipAnalyzers() ) {
+			// Register a pass through analyzer for the document id
+			typeMetadataBuilder.addToScopedAnalyzerReference( fieldPath, null, Field.Index.NOT_ANALYZED );
+		}
```
yrodiere: I think you also need to fix `createPropertyMetadataForEmbeddedId`. It's the one called when including the ID in an `@IndexedEmbedded`, either though `@IndexedEmbedded.includePaths` or `@IndexedEmbedded.includeEmbeddedObjectId`.
And we need a test for it, too.

gsmet: Added a commit to take this case into account and rebased.

PR #1652, opened 2018-04-04
Updates: JGroups 4 & WildFly 12

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-2598
 - https://hibernate.atlassian.net/browse/HSEARCH-3022

I'm unhappy about the design of the JGroups backend - I could hardly grasp the logic :)

But while I had started a rewrite, I wasn't able to cap that work nor finish it.. so I backed off to the minimal needed to get the version upgrade and trying to not change too much else. Let's focus on getting the WildFly upgrade, the clustering improvements will have to wait a bit more.


backends/jgroups/src/main/java/org/hibernate/search/backend/jgroups/impl/DispatchMessageSender.java
```diff
@@ -71,68 +72,58 @@ public View getView() {
 	}
 
 	@Override
-	public void send(final Message message, final boolean synchronous, final long timeout) throws Exception {
+	public void send(final Buffer data, final boolean synchronous, final long timeout) throws Exception {
 		final RequestOptions options = synchronous ? RequestOptions.SYNC() : RequestOptions.ASYNC();
-		options.setExclusionList( dispatcher.getChannel().getAddress() );
+		options.exclusionList( dispatcher.getChannel().getAddress() );
+		options.setTransientFlags( TransientFlag.DONT_LOOPBACK );
```
yrodiere: Are **both** of those options necessary? In my mind excluding the local address and preventing loopback are the same, but I may be missing something...

backends/jgroups/src/test/resources/log4j.properties
```diff
@@ -24,10 +18,10 @@ log4j.logger.org.jboss=info
 
 log4j.logger.org.hibernate=warn
 
-log4j.logger.org.hibernate.search=debug
+log4j.logger.org.hibernate.search=info
+log4j.logger.org.hibernate.search.backend=info
```
yrodiere: Is there a specific reason to do that? I'd rather have extensive logs in tests by default... It's particularly useful when encountering transient failures on CI. We already use the `debug`/`trace` level in Elasticsearch integration tests, for instance.

backends/jgroups/src/test/resources/testing-flush-loopback.xml
```diff
@@ -7,14 +7,14 @@
   -->
 <config xmlns="urn:org:jgroups"
 		xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-		xsi:schemaLocation="urn:org:jgroups http://www.jgroups.org/schema/JGroups-3.6.xsd">
+		xsi:schemaLocation="urn:org:jgroups http://www.jgroups.org/schema/jgroups-4.0.xsd">
 
 	<SHARED_LOOPBACK
-		bind_addr="127.0.0.1"
-		diagnostics_addr="127.0.0.1"
+		stats="false"
```
yrodiere: Did you run into issues with stats... ? Or is it just best practice?

backends/jgroups/src/test/resources/testing-flush-loopback.xml
```diff
@@ -7,14 +7,14 @@
   -->
 <config xmlns="urn:org:jgroups"
 		xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-		xsi:schemaLocation="urn:org:jgroups http://www.jgroups.org/schema/JGroups-3.6.xsd">
+		xsi:schemaLocation="urn:org:jgroups http://www.jgroups.org/schema/jgroups-4.0.xsd">
 
 	<SHARED_LOOPBACK
-		bind_addr="127.0.0.1"
-		diagnostics_addr="127.0.0.1"
```
yrodiere: Default seems to be `224.0.75.75`, so not the same... Was this unnecessary?

backends/jgroups/src/test/resources/testing-flush-loopback.xml
```diff
@@ -7,14 +7,14 @@
   -->
 <config xmlns="urn:org:jgroups"
 		xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-		xsi:schemaLocation="urn:org:jgroups http://www.jgroups.org/schema/JGroups-3.6.xsd">
+		xsi:schemaLocation="urn:org:jgroups http://www.jgroups.org/schema/jgroups-4.0.xsd">
 
 	<SHARED_LOOPBACK
-		bind_addr="127.0.0.1"
```
yrodiere: I couldn't find the default for this, but I'll assume it's more or less `127.0.0.1`.

backends/jgroups/src/test/resources/testing-flush-loopback.xml
```diff
@@ -27,14 +27,13 @@
 			xmit_table_num_rows="50"
 			xmit_table_msgs_per_row="1024"
 			xmit_table_max_compaction_time="30000"
-			max_msg_batch_size="100" />
+			discard_delivered_msgs="true" />
```
yrodiere: I can't even begin to understand how this is related to our tests, so... I'll have to trust you :)

pom.xml
```diff
@@ -200,7 +200,7 @@
 
         <version.org.jboss.spec.javax.jms.jboss-jms-api_2.0_spec>1.0.1.Final</version.org.jboss.spec.javax.jms.jboss-jms-api_2.0_spec>
         <version.org.apache.activemq.broker>5.13.3</version.org.apache.activemq.broker>
-        <version.org.jgroups>3.6.13.Final</version.org.jgroups>
+        <version.org.jgroups>4.0.11.Final</version.org.jgroups>
```
yrodiere: For the record, [WildFly 12.0.0.Final uses `4.0.10.Final`](https://github.com/wildfly/wildfly/blob/12.0.0.Final/pom.xml#L236). I guess you have reasons not to be aligned exactly this time.

backends/jgroups/src/main/java/org/hibernate/search/backend/jgroups/impl/MasterNodeSelector.java
```diff
@@ -24,26 +22,21 @@
...
+public final class MasterNodeSelector implements NodeSelectorStrategy {
 
 	@Override
 	public boolean isIndexOwnerLocal() {
 		return true;
 	}
 
 	@Override
-	public void setLocalAddress(Address address) {
+	public void setLocalAddress(final Address address) {
```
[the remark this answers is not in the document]
Sanne: yea sorry I didn't revert all changes, these should be harmless. Just a matter of style.

backends/jgroups/src/test/java/org/hibernate/search/backend/jgroups/impl/SyncJGroupsBackendTest.java
```diff
@@ -74,52 +76,20 @@ public void testSynchAsConfigured() {
 		JGroupsReceivingMockBackendQueueProcessor dvdBackendMock = extractMockBackend( "dvds" );
 
 		dvdBackendMock.resetThreadTrap();
-		boolean timeoutTriggered = false;
 		try {
 			//DVDs are sync operations so they will timeout:
 			final long presendTimestamp = System.nanoTime();
-			log.trace( "[PRESEND] Timestamp: " + presendTimestamp );
+			System.out.println( "[PRESEND] Timestamp: " + presendTimestamp );
```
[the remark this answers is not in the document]
Sanne: it's a Test. A bit annoying to have to open the logs when debugging / working on it, no?

backends/jgroups/src/main/java/org/hibernate/search/backend/jgroups/logging/impl/Log.java
```diff
@@ -67,7 +64,7 @@
 	@Message(id = 200009, value = "MuxId '%1$d' configured on the JGroups was already taken. Can't register handler!")
 	SearchException jGroupsMuxIdAlreadyTaken(short n);
 
-	@LogMessage(level = WARN)
+	@LogMessage(level = DEBUG)
```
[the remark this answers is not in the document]
Sanne: Apparently FLUSH is deprecated / not really essential anymore.

backends/jgroups/src/test/java/org/hibernate/search/backend/jgroups/impl/SyncJGroupsBackendTest.java
```diff
@@ -74,52 +76,20 @@ public void testSynchAsConfigured() {
...
 			final long postsendTimestamp = System.nanoTime();
 			final long differenceInMilliseconds = TimeUnit.MILLISECONDS.convert( (postsendTimestamp - presendTimestamp), TimeUnit.NANOSECONDS );
-			log.trace( "[POSTSEND] Timestamp: " + postsendTimestamp + " Diff: " + differenceInMilliseconds + " ms." );
-		}
-		catch (SearchException se) {
-			//Expected: we're inducing the RPC into timeout by blocking receiver processing
-			Throwable cause = se.getCause();
-			Assert.assertTrue( "Cause was not a TimeoutException but a " + cause, cause instanceof TimeoutException );
-			timeoutTriggered = true;
+			System.out.println( "[POSTSEND] Timestamp: " + postsendTimestamp + " Diff: " + differenceInMilliseconds + " ms." );
```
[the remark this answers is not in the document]
Sanne: Yes you understand correctly. This is the very problem that led me to take additional days to investigate :(
Apparently now when doing a multi-cast, if some (any) nodes don't answer it just moves on. One should build the backend differently. It will still attempt to deliver the messages in the right order automatically, but nodes in the cluster group are rather allowed to "lag behind" than to block it all.

Therefore the previous assertions no longer make sense. I was tempted to either:
 - remove the test
 - rewrite the whole backend differently

Yet while working further on it, this test actually prevented me from making some silly mistakes - so while I agree it's all puzzling I'd rather keep it now, move on, and see if we can make some time to rewrite the whole thing.

backends/jgroups/src/main/java/org/hibernate/search/backend/jgroups/impl/DispatchMessageSender.java
Sanne: In most cases you'll only need one but they are subtly different depending on the protocols being used.

the exclusion list is applied on the packet headers, like destination delivery with UDP anycast.
The "loopback" option affects the loop delivery, so if a message is broadcasted - it's still received by the local network interface but we have JGroups "not-deliver" it back up again.

backends/jgroups/src/test/resources/log4j.properties
Sanne: ok I can revert these.
I typically prefer to "opt-in" on the details as there was too much noise but you have a good point about CI failures.

yrodiere:
Merged, thanks!

PR #1655, opened 2018-04-11
HSEARCH-3030 Remove the JGroups integration from the WildFly feature packs

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3030


Sanne:
I see the previous module definition (the xml file) for the JGroups backend was not deleted. Was that intentional?

yrodiere:
Good catch. Fixed, sorry about that.

Sanne:
merged. Thanks!

PR #1656, opened 2018-04-11
HSEARCH-3021 Document the integration to DI frameworks

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3021

### DIFF
--- a/documentation/src/main/asciidoc/mapping.asciidoc
+++ b/documentation/src/main/asciidoc/mapping.asciidoc
@@ -2226,6 +2226,34 @@ CustomBridge cb = classLoaderService.classForName( "com.package.CustomBridge" );
 ----
 ====
 
+[[section-bridge-dependency-injection]]
+==== Dependency injection in bridges
+
+Hibernate Search offers SPIs that allow dependency injection frameworks
+to be used when instantiating custom bridges, enabling dependency injection in bridges.
+
+When using the Hibernate Search ORM integration,
+Hibernate Search is wired to Hibernate ORM's dependency injection support,
+meaning that if your framework already provides dependency injection to Hibernate ORM
+(for `EntityListener`s and `AttributeConverter`s in particular),
+then dependency injection should automatically work for Hibernate Search bridges, too.
+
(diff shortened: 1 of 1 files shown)


documentation/src/main/asciidoc/mapping.asciidoc
```diff
@@ -2226,6 +2226,34 @@ CustomBridge cb = classLoaderService.classForName( "com.package.CustomBridge" );
...
+and if not found it will default to reflection-based instantiation (using the public, no-argument constructor).
+
+[WARNING]
+====
+Be careful to define the scope of your bridge as appropriate.
+For instance if the bridge implements `AppliedOnTypeAwareBridge`, a singleton bridge will not work,
+since the bridge may be used in multiple places on different properties with different types.
+In that case, a "prototype" (Spring terminology) or "dependent" (CDI terminology) scope is what you want,
+and it's generally the safest choice.
+====
```
Sanne: good point! I'm not very familiar with CDI  myself (only used its precursors). Do you think we can actually verify this to be sane at bootstrap?

Sanne:
great,  merged! 

The single comment I had should be treated as a different JIRA - if at all.

documentation/src/main/asciidoc/mapping.asciidoc
yrodiere: Not cleanly with the current SPIs at my disposal: I don't have any way to know what the scope of a component is. 

For CDI I would have to bypass Hibernate ORM's SPIs to get access to the bean manager, and even then I could not handle Spring.

We may be able to add something to Hibernate ORM's SPIs, but... TBH it's be hard enough to get where we are now :)

Other than that, I guess I could keep a map "bridge type" => "set of created instances so far" and check that a newly created instance is not part of the set of created instances so far, but I think there would be a lot of false negatives, especially with that habit of DI frameworks' to add proxies everywhere.

Sanne: ok let's not explore this further, at least until there's demand.

PR #1658, opened 2018-04-19
Add *.autosave to .gitignore

yrodiere:
Because of how DIA handles automatic, periodic saves.

See https://github.com/hibernate/hibernate-search-6-poc/pull/40#pullrequestreview-113508290


gsmet:
Rebased and merged, thanks!

PR #1659, opened 2018-04-19
HSEARCH-3132 Normalizers are not closed on shutdown

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3132


gsmet:
Merged, thanks!

PR #1661, opened 2018-04-23
HSEARCH-3125 Allow native access to the Elasticsearch RestClient

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3125
Also fixes https://hibernate.atlassian.net/browse/HSEARCH-3133

I tried to do something conceptually similar to [what will be available in Search 6](https://github.com/hibernate/hibernate-search-6-poc/pull/41). See the [test in Search 6](https://github.com/hibernate/hibernate-search-6-poc/pull/41/files#diff-9524dadfd8aa7e1f285ef679979ca04eR284) vs. the test in this PR (`ElasticsearchLowLevelClientAccessIT#indexFamily_getClient()`).

Unfortunately, I couldn't use the same wording as in Search 6, since the term "backend" has a very different meaning in Search 5. I chose to borrow Sanne's idea of "index families".

### About IndexManagerType

For now I used a trick to expose `IndexFamilyType` on the API side while still relying on `IndexManagerType` internally, because the concept of `IndexManagerType` did not seem to make sense from an API point of view:

1. Index managers are an SPI concept, rarely even seen on the API side
2. I find the term `IndexManagerType` way too generic and confusing to use this wording in APIs

`IndexManagerType` is really more of a `IndexFamilyTypeImplementor`:

1. the very javadoc of `IndexManagerType` states it "avoid susing the class type of an IndexManager to allow creating multiple custom implementations for the same type of technology", so there could actually be multiple `IndexManager` classes (types) for a given `IndexManagerType`.
2. if you have a look at how `IndexManagerType` is used, it's generally to refer to an index family (either Lucene or Elasticsearch).
3. after this PR, `IndexManagerType` extends `IndexFamilyType` by adding some SPI-specific methods. Thus it's really the implementor side of `IndexFamilyType`.

However, I didn't rename the (SPI only) concept of `IndexManagerType`, because that would lead to even more changes in our code, and I don't think it would bring much, since I don't expect we will be using this concept a lot in 5.


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/indexes/spi/ElasticsearchIndexFamily.java
```diff
@@ -10,4 +10,15 @@
...
 public interface ElasticsearchIndexFamily extends IndexFamily {
 
+	/**
+	 * Retrieve the underlying, low-level client used to communicate with the Elasticsearch cluster.
+	 * <p>
+	 * <strong>WARNING - Unsupported API:</strong> the underlying client class may change without notice.
+	 *
+	 * @param clientClass The {@link Class} representing the expected client type
+	 * @return The client.
+	 * @throws SecurityException if the underlying client does not implement the given class.
```
gsmet: `SecurityException`?

elasticsearch/src/main/java/org/hibernate/search/elasticsearch/logging/impl/Log.java
```diff
@@ -516,4 +516,9 @@ SearchException invalidAnalysisFactoryParameter(@FormatWith(ClassFormatter.class
 	void executedRequest(String method, String path, Map<String, String> getParameters, long timeInMs,
 			int responseStatusCode, String responseStatusMessage,
 			String requestBodyParts, String responseBody);
+
+	@Message(id = ES_BACKEND_MESSAGES_START_ID + 94,
+			value = "Attempt to unwrap the Elasticsearch low-level client to %1$s,"
+					+ " but the client can only be unwrapped to %2$s." )
+	SearchException clientUnwrappingWithUnkownType(Class<?> requestedClass, Class<?> actualClass);
```
gsmet: s/Unkown/Unknown/

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -1116,6 +1116,32 @@ The optimization features documented in <<search-optimize>> are only partially i
  * The automatic optimization is not implemented and most probably never will be.
  * The manual optimization (`searchFactory.optimize()`) is implemented.
 
+[[elasticsearch-client-access]]
+=== Direct access to the Elasticsearch client
+
+When writing complex applications with advanced needs,
+it may be necessary from time to time to send requests to the Elasticsearch cluster directly,
+in particular if Hibernate Search does not send this kind of requests out of the box.
```
gsmet: s/send/support/?

documentation/src/main/asciidoc/elasticsearch-integration.asciidoc
```diff
@@ -1116,6 +1116,32 @@ The optimization features documented in <<search-optimize>> are only partially i
  * The automatic optimization is not implemented and most probably never will be.
  * The manual optimization (`searchFactory.optimize()`) is implemented.
 
+[[elasticsearch-client-access]]
+=== Direct access to the Elasticsearch client
+
+When writing complex applications with advanced needs,
```
gsmet: s/needs/requirements/?

yrodiere:
@gsmet Thanks. I applied your suggestions and rebased.

yrodiere:
> I like this approach, but since the unwrap takes only a type parameter it will be an ambiguous operation when having multiple families registered. I think that's ok but the docs/API should already hint that this will result in an exception in the future when:

Hmm, no it will not. When you call `unwrap`, you already retrieved an index family, thus the ambiguity, if any, has already been resolved.

With our current API, if we allowed multiple index families to be created, the ambiguity would lay where we retrieve the index family, i.e. the second line of this code:

```
SearchFactory searchFactory = fullTextSession.getSearchFactory();
IndexFamily indexFamily = searchFactory.getIndexFamily( ElasticsearchIndexFamilyType.get() );
ElasticsearchIndexFamily elasticsearchIndexFamily = indexFamily.unwrap( ElasticsearchIndexFamily.class );
RestClient restClient = elasticsearchIndexFamily.getClient( RestClient.class );
```

Solving this is easy: you simply give a name to each index family, and use that name instead of the `ElasticsearchIndexFamilyType` to retrieve the family. And fortunately, that's what we did in Search 6 (it's [already been merged](https://github.com/hibernate/hibernate-search-6-poc/blob/133417c995ae6fb1657f280686cfc47c66a5cab6/integrationtest/backend-elasticsearch/src/test/java/org/hibernate/search/v6poc/integrationtest/backend/elasticsearch/ExtensionIT.java#L284)). The code thus becomes:

```
Backend backend = mappingRepository.getBackend( "myBackend" );
ElasticsearchBackend elasticsearchBackend = backend.unwrap( ElasticsearchBackend.class );
RestClient restClient = elasticsearchBackend.getClient( RestClient.class );
```

`mappingRepository` being what replaces the `SearchFactory` in Search 6. Yes, the name is awful and will need to change :)

Of course the string cannot be just anything; it must match something you have in your configuration, for instance:

```
hibernate.search.backend.myBackend.type = elasticsearch
hibernate.search.backend.myBackend.host = https://elasticsearch.acme.com:9200/
...
```

> I'm bit confused currently about the differences between
> org.hibernate.search.indexes.spi.IndexManagerType
> org.hibernate.search.indexes.IndexFamilyType
> org.hibernate.search.indexes.IndexFamily

* `IndexFamilyType` represents a technology that a set of indexes could rely on: Lucene or Elasticseach. It is accessible through static methods, making it handy as a key in our APIs.
* `IndexFamily` represents a set of indexes using the same technology and the same "global" configuration (e.g. the host for the Elasticsearch). It actually contains information derived from the configuration and mapping, thus it cannot be accessed directly through static methods. It can be retrieved throught the `SearchFactory` by passing an `IndexFamilyType` as a key, though.
As it happens, in Search 5 you will only have one of those per technology, because we do not allow to have multiple sets of "global" configuration per technology; so you essentially have a 1-1 mapping between `IndexFamilyType` and `IndexFamily`. But in 6, you will be able to have multiple `IndexFamily` instances per `IndexFamilyType`.
* `IndexManagerType` is the SPI-level equivalent of `IndexFamilyType`. There's a 1-1 mapping between these two classes, but we simply don't want to expose the methods of `IndexManagerType` to users.

In order to clarify that, I could add what I wrote above to each type's javadoc, and maybe rename `IndexFamilyType` to `IndexFamilyKey`. Would that help?

> Perhaphs it would help to - at least internally - identify each new IndexFamily via the couple { IndexFamilyType, some name} rather than just conflate the IndexFamily concept with the IndexFamilyType concept?

Well, to be honest this PR contains mainly throwaway code. It's more of a backport from 6 to 5 than anything else. The same concept is already implemented in the Search 6 POC, and since we didn't have the same constraints as we have in 5 (in particular the whole lot of index-centric configuration that should be "index-family"-centric), I've been able to implement it in a much more straightforward way.
With that in mind, I'm not sure trying to make sure this code can evolve towards what we want in 6 is really useful.

What I can do, if the "IndexFamilyType" thing annoys you, is use strings in the API allowing to retrieve the index families: "lucene" or "elasticsearch", maybe exposed as constants, instead of `LuceneIndexFamilyType.get()` and `ElasticsearchIndexFamilyType.get()`. Then I will implement some magic to convert that to an `IndexManagerType` internally.

Sanne:
> Hmm, no it will not. [...]

Yes I understand, that's what I meant.

> IndexManagerType is the SPI-level equivalent of IndexFamilyType. There's a 1-1 mapping between these two classes, but we simply don't want to expose the methods of IndexManagerType to users.

Great, thanks: I had missed that detail and yes it might be worth adding that to the javadoc.

> Well, to be honest this PR contains mainly throwaway code. It's more of a backport from 6 to 5 than anything else.  [..]

That's ok and not bothering me particularly. I was thinking of the SPI could be made more similar to what 6 will have - but since it's SPI I agree it's not a priority.

> What I can do, if the "IndexFamilyType" thing annoys you, is use strings in the API allowing to retrieve the index families: "lucene" or "elasticsearch", maybe exposed as constants, instead of LuceneIndexFamilyType.get() and ElasticsearchIndexFamilyType.get(). Then I will implement some magic to convert that to an IndexManagerType internally.

Mmm no I don't think that's worth it, not sure it clarifies much either: it's ok to have a notion of `IndexFamilyType`. What I suggested in my previous comment is not to remove `IndexFamilyType` but to consistently use a *couple* { `IndexFamilyType` + _name_ } as "key" for a backend group.

e.g.

    IndexManagerType#createIndexFamily(ServiceManager serviceManager, SearchConfiguration cfg, String familyName);

In the current PR you have #createIndexFamily(ServiceManager serviceManager, SearchConfiguration cfg); without the name, which I found puzzling.. but then yes ok as long as we now this SPI will have to change I can live with that.

yrodiere:
I pushed an additional commit to update the javadoc and rebased.
Will merge as soon as CI is green.

yrodiere:
Merged, thanks to you both!

PR #1662, opened 2018-04-25
HSEARCH-3134 Error message in MassIndexerImpl#toRootEntities mentions "a subclass of an indexed entity" but should mention "supertype" instead

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3134

### DIFF
--- a/orm/src/main/java/org/hibernate/search/batchindexing/impl/MassIndexerImpl.java
+++ b/orm/src/main/java/org/hibernate/search/batchindexing/impl/MassIndexerImpl.java
@@ -17,6 +17,7 @@ import org.hibernate.search.MassIndexer;
 import org.hibernate.search.batchindexing.MassIndexerProgressMonitor;
 import org.hibernate.search.batchindexing.spi.MassIndexerWithTenant;
 import org.hibernate.search.engine.integration.impl.ExtendedSearchIntegrator;
+import org.hibernate.search.util.StringHelper;
 import org.hibernate.search.util.jmx.impl.JMXRegistrar;
 import org.hibernate.search.spi.IndexedTypeSet;
 import org.hibernate.search.spi.SearchIntegrator;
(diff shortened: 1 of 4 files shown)


orm/src/main/java/org/hibernate/search/batchindexing/impl/MassIndexerImpl.java
```diff
@@ -78,16 +79,19 @@ protected MassIndexerImpl(SearchIntegrator searchIntegrator, SessionFactoryImple
...
 	private static IndexedTypeSet toRootEntities(ExtendedSearchIntegrator extendedIntegrator, Class<?>... selection) {
-		Set<Class<?>> entities = new HashSet<Class<?>>();
+		IndexedTypeSet selectionAsSet = IndexedTypeSets.fromClasses( selection );
 		//first build the "entities" set containing all indexed subtypes of "selection".
-		for ( Class<?> entityType : selection ) {
-			IndexedTypeSet targetedClasses = extendedIntegrator.getIndexedTypesPolymorphic( IndexedTypeSets.fromClass( entityType ) );
-			if ( targetedClasses.isEmpty() ) {
-				String msg = entityType.getName() + " is not an indexed entity or a subclass of an indexed entity";
-				throw new IllegalArgumentException( msg );
+		Set<Class<?>> entities = extendedIntegrator.getIndexedTypesPolymorphic( selectionAsSet ).toPojosSet();
```
Sanne: This refactoring isn't semantically equivalent. Was that intentional?

yrodiere: Done, it should be better now.

orm/src/main/java/org/hibernate/search/batchindexing/impl/MassIndexerImpl.java
```diff
@@ -78,16 +79,32 @@ protected MassIndexerImpl(SearchIntegrator searchIntegrator, SessionFactoryImple
...
+			else {
+				entities.addAll( targetedClasses.toPojosSet() );
+			}
+		}
+
+		if ( !nonIndexedEntities.isEmpty() ) {
+			IndexedTypeSet nonIndexedEntitiesAsSet = IndexedTypeSets.fromClasses( nonIndexedEntities );
+			IndexedTypeSet configuredTargetEntities =
+					extendedIntegrator.getConfiguredTypesPolymorphic( nonIndexedEntitiesAsSet );
+			if ( configuredTargetEntities.isEmpty() ) {
```
Sanne: Looks good, but looking forward remember the plan about IndexedTypeSets#fromClasses :

those static helpers should disappear so to use a stateful registry of types instead. That means that the equivalent of "fromClasses" is likely going to throw an error when attempting to convert a totally unknown (neither configured nor indexed) set of classes into an `IndexedTypeSet`. The idea is that `IndexedTypeSet` will be a _set of indexed types_ so you won't be able to use it to represent non-indexed-types or even non-configured-types.

Sanne:
merged in master

orm/src/main/java/org/hibernate/search/batchindexing/impl/MassIndexerImpl.java
yrodiere: Right. But I didn't need to use the concept of `IndexedTypeIdentifier` anywhere else than at bootstrap in the Search 6 POC, at least for now. So I'm not sure we will need it in the mass indexer in 6.

I guess we'll see if we need to reintroduce it at runtime when we actually implement protobuf/JSON mapping. If we need to re-use code from the POJO mapper, then we may need to use `IndexedTypeIdentifier` or a similar concept at runtime.

PR #1667, opened 2018-05-02
HSEARCH-1085 MappingModelMetadataProvider should pass the chosen Clas…

Sanne:
…sLoader rather than overriding the context

A 6 years old workaround finally being removed :)

 - https://hibernate.atlassian.net/browse/HSEARCH-1085


yrodiere:
Nice! Merged, thanks.

PR #1668, opened 2018-05-02
ForbiddenAPI update and improved compatibility with future JDKs

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-3153
 - https://hibernate.atlassian.net/browse/HSEARCH-3154

### DIFF
--- a/build-config/src/main/resources/forbidden-runtime.txt
+++ b/build-config/src/main/resources/forbidden-runtime.txt
@@ -29,3 +29,8 @@ java.lang.StringBuffer
 # Probably meant the other Objects:
 
 org.jboss.logging.processor.util.Objects
+
+################################################################################################################
+# Methods from Hibernate ORM :
+# This one is only safe to use with specific configurations:
+org.hibernate.SharedSessionContract#getTransaction()
--- a/build-config/src/main/resources/forbidden-runtime.txt
+++ b/build-config/src/main/resources/forbidden-runtime.txt
@@ -1,0 +1,31 @@
+# Hibernate Search, full-text search for your domain model
+#
+# License: GNU Lesser General Public License (LGPL), version 2.1 or later
+# See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+
+# This file contains forbidden API signatures common to all modules
+
+@ignoreUnresolvable
+@defaultMessage Should not be used
+
+################################################################################################################
+# Some packages that should not be used
+java.awt.**
+sun.**
+org.slf4j.**
+junit.framework.**
+
+################################################################################################################
+# org.testng.Assert#fail() should be used instead of:
+org.assertj.core.api.Fail#fail()
+org.testng.FileAssert#fail()
+org.assertj.core.api.Assertions#fail()
+
+################################################################################################################
+# Nobody should be using StringBuffer anymore
+java.lang.StringBuffer
+
+################################################################################################################
+# Probably meant the other Objects:
+
+org.jboss.logging.processor.util.Objects
--- a/build-config/src/main/resources/forbidden-tests.txt
+++ b/build-config/src/main/resources/forbidden-tests.txt
@@ -23,5 +23,9 @@ org.assertj.core.api.Assertions#fail()
 
 ################################################################################################################
 # Probably meant the other Objects:
-
 org.jboss.logging.processor.util.Objects
+
+################################################################################################################
+# Various, from our own code.
+# This one needs to be handled with special care to avoid tests breaking out of their isolation (see exclusions):
+org.hibernate.search.spi.SearchIntegratorBuilder#buildSearchIntegrator()
--- a/build-config/src/main/resources/forbidden-tests.txt
+++ b/build-config/src/main/resources/forbidden-tests.txt
@@ -1,0 +1,27 @@
+# Hibernate Search, full-text search for your domain model
+#
+# License: GNU Lesser General Public License (LGPL), version 2.1 or later
+# See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+
+# This file contains forbidden API signatures common to all modules
+
+@ignoreUnresolvable
+@defaultMessage Should not be used
+
+################################################################################################################
+# Some packages that should not be used
+java.awt.**
+sun.**
+org.slf4j.**
+junit.framework.**
+
+################################################################################################################
+# org.testng.Assert#fail() should be used instead of:
+org.assertj.core.api.Fail#fail()
+org.testng.FileAssert#fail()
+org.assertj.core.api.Assertions#fail()
+
+################################################################################################################
+# Probably meant the other Objects:
+
+org.jboss.logging.processor.util.Objects
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/settings/impl/translation/JsonPassThroughAnalysisDefinitionFactory.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/settings/impl/translation/JsonPassThroughAnalysisDefinitionFactory.java
@@ -12,7 +12,7 @@ import java.util.Map.Entry;
 
 import org.hibernate.search.elasticsearch.logging.impl.Log;
 import org.hibernate.search.elasticsearch.settings.impl.model.AnalysisDefinition;
-import org.hibernate.search.exception.AssertionFailure;
+import org.hibernate.search.util.impl.ClassLoaderHelper;
 import org.hibernate.search.util.logging.impl.LoggerFactory;
 import java.lang.invoke.MethodHandles;
 
@@ -43,13 +43,7 @@ class JsonPassThroughAnalysisDefinitionFactory<D extends AnalysisDefinition> imp
 
 	@Override
 	public D create(Map<String,String> parameters) {
-		D result;
-		try {
-			result = targetClass.newInstance();
-		}
-		catch (InstantiationException | IllegalAccessException e) {
-			throw new AssertionFailure( "Unexpected failure while instanciating a definition", e );
-		}
+		D result = ClassLoaderHelper.untypedInstanceFromClass( targetClass, "analysis definition" );
 
 		Map<String, JsonElement> parameterMap = new LinkedHashMap<>();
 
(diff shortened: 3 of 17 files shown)


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/settings/impl/translation/JsonPassThroughAnalysisDefinitionFactory.java
```diff
@@ -43,13 +43,7 @@ public String getType() {
...
 	@Override
 	public D create(Map<String,String> parameters) {
-		D result;
-		try {
-			result = targetClass.newInstance();
-		}
-		catch (InstantiationException | IllegalAccessException e) {
-			throw new AssertionFailure( "Unexpected failure while instanciating a definition", e );
-		}
+		D result = ClassLoaderHelper.untypedInstanceFromClass( targetClass, "Analyzer definition" );
```
yrodiere: "Analyzer definition" => "analysis definition".
This encompasses tokenizer definitions, token filter definitions, char filter definitions as well as analyzer definitions.

engine/src/main/java/org/hibernate/search/engine/impl/ConfigContext.java
```diff
@@ -157,12 +158,9 @@ private FilterDef interpretFullTextFilterDef(FullTextFilterDef defAnn) {
...
 		}
 		try {
-			filterDef.getImpl().newInstance();
+			ClassLoaderHelper.untypedInstanceFromClass( filterDef.getImpl(), "filter definition" );
 		}
-		catch (IllegalAccessException e) {
-			throw new SearchException( "Unable to create Filter class: " + filterDef.getImpl().getName(), e );
-		}
-		catch (InstantiationException e) {
+		catch (Exception e) {
```
yrodiere: This really should catch `RuntimeException`: there's no reason to catch `Exception` if no checked exception is thrown, and this way we will have a compile-time error if we ever add a checked exception to `ClassLoaderHelper.untypedInstanceFromClass`.

engine/src/main/java/org/hibernate/search/engine/service/beanresolver/spi/ReflectionBeanResolver.java
```diff
@@ -32,16 +33,7 @@
 
 	@Override
 	public <T> T resolve(Class<?> classOrFactoryClass, Class<T> expectedClass) {
-		Object instance;
-		try {
-			instance = classOrFactoryClass.newInstance();
-		}
-		catch (InstantiationException e) {
-			throw log.noPublicNoArgConstructor( classOrFactoryClass.getName() );
```
yrodiere: I guess it's on purpose, but `log.unableToAccessClass` is not used anymore. Maybe remove the method from the logger?

orm/src/test/java/org/hibernate/search/test/directoryProvider/MultipleSFTestCase.java
```diff
@@ -74,13 +75,13 @@ protected void buildSessionFactories(Class[] classes, String[] packages, String[
 		}
 	}
 
-	@Override
+	@Before
 	protected void setUp() throws Exception {
```
yrodiere: You should probably make the method `public` just for consistency... Same for `tearDown`.

pom.xml
```diff
@@ -1588,34 +1588,43 @@
...
+                                     <bundledSignature>jdk-internal</bundledSignature>
+                                 </bundledSignatures>
+                                 <failOnMissingClasses>false</failOnMissingClasses>
+                                 <failOnUnresolvableSignatures>false</failOnUnresolvableSignatures>
+                                 <!--
+                                 The method SharedSessionContract#getTransaction() should never be used
+                                 as it is only reliable in certain configurations.
+                                  -->
+                                 <signatures>
 org.hibernate.SharedSessionContract#getTransaction()
```
yrodiere: Is there a specific reason not to include this in `forbidden-runtime.txt` and `forbidden-tests.txt`?

engine/src/main/java/org/hibernate/search/util/impl/ClassLoaderHelper.java
```diff
@@ -69,28 +75,42 @@ private ClassLoaderHelper() {
...
 	 *
 	 * @throws SearchException wrapping other error types with a proper error message for all kind of problems, like
-	 * missing proper constructor, wrong type, security errors.
+	 * missing proper constructor, wrong type, securitymanager errors.
 	 */
 	public static <T> T instanceFromClass(Class<T> targetSuperType, Class<?> classToLoad, String componentDescription) {
 		checkClassType( classToLoad, componentDescription );
-		checkHasNoArgConstructor( classToLoad, componentDescription );
-		Object instance;
+		final Constructor<?> constructor = getNoArgConstructor( classToLoad, componentDescription );
```
yrodiere: This variable is never used. Maybe this line should be removed?

engine/src/main/java/org/hibernate/search/util/impl/ClassLoaderHelper.java
```diff
@@ -69,28 +75,42 @@ private ClassLoaderHelper() {
...
+	 * @param <T> the type of targetSuperType: defines the return type
+	 * @param classToLoad the class to be instantiated
+	 * @param componentDescription a role name/description to contextualize error messages. Ideally should be provided, but it can handle null.
+	 *
+	 * @return a new instance of classToLoad
+	 *
+	 * @throws SearchException wrapping other error types with a proper error message for all kind of problems, like
+	 * missing proper constructor, securitymanager errors.
+	 */
+	public static <T> T untypedInstanceFromClass(final Class<T> classToLoad, final String componentDescription) {
```
yrodiere: Ideally, shouldn't you call `checkClassType( classToLoad, componentDescription );` here?

Sanne: +1 nice improvement

pom.xml
Sanne: I could move it. The only reason I didn't was that I prefer small, fine-grained patches :)

engine/src/main/java/org/hibernate/search/engine/impl/ConfigContext.java
Sanne: ok.

Sanne:
Thanks @yrodiere , suggestions applied.

build-config/src/main/resources/forbidden-runtime.txt
```diff
@@ -0,0 +1,36 @@
...
+
+################################################################################################################
+# Some packages that should not be used
+java.awt.**
+sun.**
+org.slf4j.**
+junit.framework.**
+
+################################################################################################################
+# org.testng.Assert#fail() should be used instead of:
```
gsmet: This one looks weird, considering you're using JUnit on Search :).

build-config/src/main/resources/forbidden-tests.txt
```diff
@@ -0,0 +1,31 @@
...
+
+################################################################################################################
+# Some packages that should not be used
+java.awt.**
+sun.**
+org.slf4j.**
+junit.framework.**
+
+################################################################################################################
+# org.testng.Assert#fail() should be used instead of:
```
gsmet: Same here about TestNG.

engine/src/main/java/org/hibernate/search/engine/spi/SearchMappingHelper.java
```diff
@@ -81,7 +82,7 @@ private static SearchMapping getNewInstanceOfSearchMapping(Class<?> clazz, Metho
 		SearchMapping mapping;
 		try {
 			LOG.debugf( "invoking factory method [ %s.%s ] to get search mapping instance", clazz.getName(), method.getName() );
-			Object instance = clazz.newInstance();
+			final Object instance = ClassLoaderHelper.untypedInstanceFromClass( clazz, "SearchMapping provider" );
```
gsmet: Real question: what does this `final` bring us?

engine/src/main/java/org/hibernate/search/util/impl/ClassLoaderHelper.java
```diff
@@ -69,28 +75,42 @@ private ClassLoaderHelper() {
 	 * @return a new instance of classToLoad
 	 *
 	 * @throws SearchException wrapping other error types with a proper error message for all kind of problems, like
-	 * missing proper constructor, wrong type, security errors.
+	 * missing proper constructor, wrong type, securitymanager errors.
```
gsmet: `security manager` with a space would be better, wouldn't it?

pom.xml
```diff
@@ -1588,34 +1588,40 @@
...
-                                     <!--
-                                     The method SharedSessionContract#getTransaction() should never be used
-                                     as it is only reliable in certain configurations.
-                                      -->
-                                     <signatures>
-org.hibernate.SharedSessionContract#getTransaction()
-                                     </signatures>
-                                 </configuration>
+                            <configuration>
+                                 <targetVersion>10</targetVersion>
```
gsmet: Isn't there a risk that we introduce calls to APIs introduced in Java 10?

build-config/src/main/resources/forbidden-runtime.txt
Sanne: it's a comment :)

pom.xml
Sanne: It's the other way around: it's about banning calls, not introducing calls.

Essentially the sets of deprecated/removed methods included in ForbiddenAPI for JDK10 are more up to date than those in JDK9.

Sadly the JDK11 list isn't included yet, but at least if we avoid anything which is deprecated in JDK10 changes are good that it will be fine in JDK11 as well.

engine/src/main/java/org/hibernate/search/util/impl/ClassLoaderHelper.java
Sanne: I'm referring specifically to the securitymanager system of the JVM, not a generic "security manager". I changed it as the previous one was a bit alarming but it's not very important.

engine/src/main/java/org/hibernate/search/engine/spi/SearchMappingHelper.java
Sanne: readability

yrodiere: Yeah, it's mostly readability. If Java used constants instead of variables by default, there would be reasons to use some `mutable` or `var` keyword from time to time, but since it's the other way around, it's just easier to leave everything mutable.

I don't think the compiler is able to optimize things more when the `final` keyword is used on a local variable, since even without it the compiler is able to detect "effectively final" variables. The "effectively final" concept is used for lambdas: it's what allows you to refer to a variable out of the scope of the lambda, on the condition the variable is never re-affected. 

Previously (Java 7 and below) you had to mark a variable as `final` to use it in an anonymous class, but with Java 8 and this "effectively final" concept, I'm not sure it's necessary any longer. I should try it out some time.

PR #1669, opened 2018-05-03
HSEARCH-3000 Pick Jigsaw Automatic Module names for all published mod…

Sanne:
…ules

 - https://hibernate.atlassian.net/browse/HSEARCH-3000

### DIFF
--- a/elasticsearch/pom.xml
+++ b/elasticsearch/pom.xml
@@ -121,9 +121,6 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-source-plugin</artifactId>
             </plugin>
-            <!-- Elasticsearch integration currently not supported in OSGi;
-                 This section commented out to not make it appear it might work.
-                 If you need this to work in OSGi, please see and vote for: https://hibernate.atlassian.net/browse/HSEARCH-2524
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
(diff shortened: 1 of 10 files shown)


elasticsearch/pom.xml
```diff
@@ -136,12 +133,17 @@
                         </goals>
                         <configuration>
                             <archive>
-                                <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+                                <manifestEntries>
+                                    <Automatic-Module-Name>org.hibernate.search.indexing.elasticsearch</Automatic-Module-Name>
```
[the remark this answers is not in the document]
Sanne: No I wasn't planning to change this later; would be nicer to commit on a choice and stick to it.. even if you don't like it later :)
Feelings about names are personal, we all need to compromise yet the only thing we know for sure is that changing things confuses everyone.

I do agree that "backend" is a suited term for the architectural picture. I'm not convinced about module names / in the context of dependencies: people are going to use "Hibernate" and "Elasticsearch", most of them won't care less of how it's bolted together, many won't even care that we call this development group "Hibernate Search".

@gsmet proposed the winning name: http://lists.jboss.org/pipermail/hibernate-dev/2018-February/017252.html

If you no longer like the "indexing" part, happy to talk about it... on the mailing list I had initially suggested "org.hibernate.search.elasticsearch".

I'd be inclined to get back to that one as it clearly expresses what people are looking for, no need to confusing/debatable and superfluous prefixes.

You think we should revisit this name? Please challenge it on the mailing list, as this PR is merely implementing the decision documented on JIRA.

Sanne:
updated

jsr352/core/pom.xml
```diff
@@ -80,8 +80,23 @@
...
+                    <execution>
+                        <id>default-jar</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>jar</goal>
+                        </goals>
+                        <configuration>
+                            <archive>
+                                <manifestEntries>
+                                    <Automatic-Module-Name>org.hibernate.search.jsr352</Automatic-Module-Name>
```
yrodiere: Maybe rather `org.hibernate.search.jsr352.core`, since we need an additional module for JBeret.

By the way, maybe you should also add a module name to `hibernate-search-jsr352-jberet`? `org.hibernate.search.jsr352.jberet`, maybe?

elasticsearch/pom.xml
```diff
@@ -136,12 +133,17 @@
                         </goals>
                         <configuration>
                             <archive>
-                                <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+                                <manifestEntries>
+                                    <Automatic-Module-Name>org.hibernate.search.backend.elasticsearch</Automatic-Module-Name>
```
yrodiere: You should also add a module name to the `hibernate-search-elasticsearch-aws` module. I would go with `org.hibernate.search.backend.elasticsearch.aws`.

Sanne:
updated again

yrodiere:
Merged, thanks!

PR #1672, opened 2018-05-10
HSEARCH-3158 Define default index name and possibility to have index type

tbounsiar:
- Add possibility to define Indexed Type Name for hibernate-search-elasticsearch by defining a new attribute type in @Indexed
 - Force hibernate-search-elasticsearch to use type as name if it defined or simple class name (without package)

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ElasticsearchIndexWorkVisitor.java
@@ -32,6 +32,7 @@ import org.hibernate.search.elasticsearch.gson.impl.UnexpectedJsonElementTypeExc
 import org.hibernate.search.elasticsearch.gson.impl.UnknownTypeJsonAccessor;
 import org.hibernate.search.elasticsearch.impl.NestingMarker.NestingPathComponent;
 import org.hibernate.search.elasticsearch.logging.impl.Log;
+import org.hibernate.search.elasticsearch.util.impl.ElasticsearchEntityHelper;
 import org.hibernate.search.elasticsearch.util.impl.FieldHelper;
 import org.hibernate.search.elasticsearch.util.impl.FieldHelper.ExtendedFieldType;
 import org.hibernate.search.elasticsearch.util.impl.ParentPathMismatchException;
@@ -118,7 +119,7 @@ class ElasticsearchIndexWorkVisitor implements IndexWorkVisitor<IndexingMonitor,
 		 * Deleting only the given type.
 		 * Inheritance trees are handled at a higher level by creating multiple purge works.
 		 */
-		builder.type( URLEncodedString.fromString( work.getEntityType().getName() ) );
+		builder.type( URLEncodedString.fromString( ElasticsearchEntityHelper.getIndexedTypeName( work.getEntityType() ) ) );
 		return builder.build();
 	}
 
@@ -145,7 +146,8 @@ class ElasticsearchIndexWorkVisitor implements IndexWorkVisitor<IndexingMonitor,
 				searchIntegrator.getIndexBinding( work.getEntityType() ).getDocumentBuilder(),
 				work.getDeletionQuery()
 		);
-		URLEncodedString typeName = URLEncodedString.fromString( work.getEntityType().getName() );
+		URLEncodedString typeName = URLEncodedString.fromString(
+				ElasticsearchEntityHelper.getIndexedTypeName( work.getEntityType() ) );
 
 		JsonObject payload = createDeleteByQueryPayload( convertedQuery, work.getTenantId() );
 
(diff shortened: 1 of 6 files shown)


Sanne:
add to whitelist

Sanne:
@tbounsiar that's brilliant, thanks! 

In this version though I'd prefer to not add an Elasticsearch specific attribute to the @Indexed annotation. We will improve the API in version 6 but so far we need to be very conservative.

May I suggest you to define this same attribute in a configuration property scoped by index name?

It could be for example

    hibernate.search.default.elasticsearch.index_type x

which makes it overrideable for each index:

     hibernate.search.[indexname].elasticsearch.index_type y

Incidentally, the build failed because @apiNote is not a valid javadoc.

tbounsiar:
@Sanne  thank you for comment and your suggestion, but i think if we add hibernate.search.default.elasticsearch.index_type  to configuration it will be applied to all Entities, and we will have conflicts.

yrodiere:
Thanks for the contribution, but as Sanne mentioned, this feature is inherently Elasticsearch-specific and we'd rather not introduce ES-specific bits in the common, abstract API.

Also, there are other problems related to the fact that metadata can also be provided through a programmatic mapping API, so we can't just use Java reflection to get it, we have to go through an abstraction layer. It's a bit complex, but if you want to have a look at how it's done, see `org.hibernate.search.indexes.impl.IndexManagerHolder#getIndexName`. Basically you need to use the `ReflectionManager` to turn the class to an `XClass`.

Sanne's configuration properties idea would indeed be a problem when you have multiple type mappings in a single index, so for now it's not an option. That being said, Elasticsearch 6.0 and above only allow one type per index, and our plans so far are to follow the same path in Hibernate Search 6.0 (i.e. allow only one type per index, even with Elasticsearch 5 and below or in embedded Lucene mode).

So, theoretically, we could make this a configuration property in 6. For now, in the Search 6 prototype, the type name [is hardcoded](https://github.com/hibernate/hibernate-search-6-poc/blob/50432d6a564bcd31e6950bfb9103e049d0bfcecc/backend-elasticsearch/src/main/java/org/hibernate/search/v6poc/backend/elasticsearch/index/impl/ElasticsearchIndexManagerBuilder.java#L56), because I didn't see the point of naming the type if there's only one per index.

I think there's a more general problem we should think about in Search 6, though: passing backend-specific mapping options to the backend through a mapping API that is inherently backend-independent. I [added a section about this](https://docs.google.com/document/d/16PAa__LsxyLZcbW3q1MvgyIznh4ZnCYLupbNMXKMRds/edit?pli=1#heading=h.tmalk5pijr3) to the "TODO" document for Search 6.

As to Search 5... I can only see two solutions:

1. Go with the configuration properties idea, but format the property key differently so that it's type-scoped. Something like `hibernate.search.[fully-qualified class name].elasticsearch.index_type y` instead of `hibernate.search.[indexname].elasticsearch.index_type y`
2. Use a separate, Elasticsearch-specific annotation, and add an SPI for the Elasticsearch integration to access this annotation.

yrodiere:
EDIT: Submitted my comment too fast... Conclusion below.

Personally, I'd rather go with the configuration property idea (solution 1), since that's how most other Elasticsearch-specific features work. And the Elasticsearch-specific annotations (solution 2) would introduce a lot of complexity if we want to preserve them in 6, where we may support mapping data that doesn't come from POJOs...

@tbounsiar If you want to give it a try, know that index options are provided through `org.hibernate.search.elasticsearch.impl.ElasticsearchIndexManager#initialize`. Don't hesitate to ask if you need some guidance.

tbounsiar:
Hello @yrodiere, thanks for comments, i will take a look.

Sanne:
@tbounsiar regarding my previous suggestion: you're right, it shouldn't be coupled 1:1 to the index of course. Sorry I was in a hurry that day and didn't think clearly :+1: 

I agree with @yrodiere 's summary and the preference for option 1.

There's a third option to consider:

3. don't have this feature in 5.x

I'm happy to have this in 5.x if the solution is doable and there are practical benefits. However I haven't understood if there are strong benefits?
The fact that 5.x doesn't have a perfect Elasticsearch mapping yet is no secret, that's why we are changing many APIs 6 in and why its integration is currently documented as experimental.

In short, happy to solve what's possible in 5.x - even with a "dirty" workaround if it's important for people - but let's not waste time on an itch to make the mapping perfect as it will never be in 5.x.

On the other hand, I expect it will take about a year to release 6 so if it's important for you then we're happy to help.

I hope that makes sense..

yrodiere:
We thought about this use case and created this ticket to handle backend-specific, index-scoped configuration in Hibernate Search 6: https://hibernate.atlassian.net/browse/HSEARCH-3306

The implementation would be very different from this PR, though, so I will close this PR.

If you're interested in contributing this feature to Search 6, feel free to [reach out to us](http://hibernate.org/community/#contribute) and we'll help you getting started.

PR #1673, opened 2018-05-14
HSEARCH-3156 @ContainedIn fails at runtime when targeting a class for which only some subclasses are configured in Search

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3156

To be backported to 5.9 at the very least; maybe it would make sense to backport to 5.6 too.


engine/src/main/java/org/hibernate/search/engine/impl/WorkPlan.java
```diff
@@ -111,12 +111,36 @@ public int size() {
...
+	}
+
+	/**
+	 * @param tenantId the tenant identifier
+	 * @param typeIdentifier The entity class for which to retrieve the work
+	 *
+	 * @return the {@code PerClassWork} from the {@link #byClass} map, creating it if necessary and if possible
+	 * (if the type is configured). Otherwise, return null.
+	 */
+	private PerClassWork getClassWorkIfConfigured(String tenantId, IndexedTypeIdentifier typeIdentifier) {
```
gsmet: I suppose there's a good reason you didn't use an `Optional` here? It would make a lot of sense.

engine/src/main/java/org/hibernate/search/engine/impl/WorkPlan.java
```diff
@@ -111,12 +111,36 @@ public int size() {
 	 * @param typeIdentifier The entity class for which to retrieve the work
 	 *
 	 * @return returns (and creates if needed) the {@code PerClassWork} from the {@link #byClass} map.
+	 * @throws SearchException if the specified type is not @Indexed nor hosts any @ContainedIn
 	 */
 	private PerClassWork getClassWork(String tenantId, IndexedTypeIdentifier typeIdentifier) {
```
gsmet: Ah and one additional question: I suppose you kept this one because there are cases where it makes sense to use it?

It's not very obvious when we should use one or the other, maybe a comment would help understanding when it's necessary to use the one throwing an exception?

engine/src/main/java/org/hibernate/search/engine/impl/WorkPlan.java
yrodiere: Mainly because we don't use such features a lot in Search 5. But now that I think of it, it also makes the code more painful to backport to 5.6, which uses Java 7.

engine/src/main/java/org/hibernate/search/engine/impl/WorkPlan.java
yrodiere: Yeah, there are other places that need to access the class work, but have nothing to do with `@ContainedIn`, and thus do not care about/do not expect non-configured types.

I don't expect us to add other uses of this method (this code will be removed in 6), so I'm not really sure what to add to the javadoc... Essentially you should use the one that throws an exception when you want it to throw an exception if the entity type is not configured in Search, and the other when you want to handle the problem of non-configured entity types yourself...

engine/src/main/java/org/hibernate/search/engine/impl/WorkPlan.java
gsmet: Right, makes sense.

engine/src/main/java/org/hibernate/search/engine/impl/WorkPlan.java
gsmet: All good then.

gsmet:
Merged and also merged all the backports.

PR #1676, opened 2018-05-15
HSEARCH-3159 Upgrade to Hibernate ORM 5.3.0.Final

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3159

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,10 +34,8 @@ over its REST API.
 This version of Hibernate Search requires:
 
 * Java SE 8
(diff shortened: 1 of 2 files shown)


Sanne:
looks great! 
Just the readme needs a fix?

yrodiere:
Right, done. At some point, we should probably replace this information with a link to the right page at hibernate.org, now that we have a compatibility matrix and all.

PR #1677, opened 2018-05-15
HSEARCH-3161 Clarify which LGPL version exactly in the POM file

Sanne:
- https://hibernate.atlassian.net/browse/HSEARCH-3161


yrodiere:
Merged, thanks!

PR #1679, opened 2018-05-23
HSEARCH-3168 JBoss Modules should not use the deprecated module of Hi…

Sanne:
…bernate ORM

 - https://hibernate.atlassian.net/browse/HSEARCH-3168


yrodiere:
Merged, thanks!

PR #1680, opened 2018-05-23
HSEARCH-3163 On bootstrap failure the SessionFactory is closed without propagating the root cause

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3163


Sanne:
Awesome, works nicely and the test is spot on.

Merged.

I added a trivial change at the end to remove some unnecessary resources from the deployment.

Minor inconvenience, but unrelated to this issue: apparently I no longer can run integration tests from the IDE because of t hose jacoco properties:

    Error: Could not find or load main class ${container.jacoco.args}

PR #1681, opened 2018-05-24
Backports to 5.10:  HSEARCH-3163,  HSEARCH-3168,  HSEARCH-2933

yrodiere:
Because master is now 5.11, not 5.10.

https://hibernate.atlassian.net/browse/HSEARCH-2933
https://hibernate.atlassian.net/browse/HSEARCH-3168
https://hibernate.atlassian.net/browse/HSEARCH-3163


Sanne:
ah, ok. Good catch! merged.

We normally wait a little bit so to not need backports as there always is a .1 :)

PR #1682, opened 2018-05-24
HSEARCH-3172 + HSEARCH-3184 + HSEARCH-3188 Fix build problems affecting IDEA

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3172
https://hibernate.atlassian.net/browse/HSEARCH-3184
https://hibernate.atlassian.net/browse/HSEARCH-3188

The first and last commits are only cleanup.

The second commit actually fixes the JaCoCo problem.

The third commit is not absolutely necessary, but it seems to make more sense that way... Also, it allows to make IDEA generate the WildFly instance by setting "Maven > Importing > Phase to be used for folders update" to "process-test-resources" in the IDEA settings, then clicking the "Generate Sources and Update Folders for All Projects" button in the Maven tool panel.

There doesn't seem to be any way to make IDEA generate the WildFly instance automatically, at least not without heavy customization of your IDEA project configuration.

### DIFF
--- a/integrationtest/performance/orm/pom.xml
+++ b/integrationtest/performance/orm/pom.xml
@@ -222,7 +222,7 @@
                         <goals>
                             <goal>build</goal>
                         </goals>
-                        <phase>process-test-resources</phase>
+                        <phase>generate-test-resources</phase>
                         <configuration>
                             <skip>${skipWildFlyPreparation}</skip>
                             <config-file>server-provisioning.xml</config-file>
@@ -256,6 +256,7 @@
                 <executions>
                     <execution>
                         <id>copy-drivers</id>
+                        <!-- Must execute after "generate-test-resources", during which we create the WildFly directory -->
                         <phase>process-test-resources</phase>
                         <goals>
                             <goal>copy</goal>
@@ -275,6 +276,7 @@
                     </execution>
                     <execution>
                         <id>fetch-jpa-patch</id>
+                        <!-- Must execute after "generate-test-resources", during which we create the WildFly directory -->
                         <phase>process-test-resources</phase>
                         <goals>
                             <goal>copy</goal>
--- a/integrationtest/performance/orm/pom.xml
+++ b/integrationtest/performance/orm/pom.xml
@@ -30,7 +30,7 @@
         <jbosshome>${project.build.directory}/${serverName}/</jbosshome>
 
         <!-- Relax the JVM restrictions for integration tests, when using Jigsaw -->
-        <additionalRuntimeArgLine>${additionalRuntimeArgLineForWildFlyTests}</additionalRuntimeArgLine>
+        <test.jvm.args>${test.wildfly.jvm.args}</test.jvm.args>
     </properties>
 
     <dependencyManagement>
--- a/integrationtest/performance/orm/pom.xml
+++ b/integrationtest/performance/orm/pom.xml
@@ -377,7 +377,7 @@
         <repository>
             <id>jboss-public-repository-group</id>
             <name>JBoss Public Maven Repository Group</name>
-            <url>${repository.jboss-public.url}</url>
+            <url>${jboss.public.repo.url}</url>
             <layout>default</layout>
             <releases>
                 <enabled>true</enabled>
@@ -403,7 +403,7 @@
         <pluginRepository>
             <id>jboss-public-repository-group</id>
             <name>JBoss Public Maven Repository Group</name>
-            <url>${repository.jboss-public.url}</url>
+            <url>${jboss.public.repo.url}</url>
             <layout>default</layout>
             <releases>
                 <enabled>true</enabled>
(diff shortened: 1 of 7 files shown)


gsmet:
+1 to merge

yrodiere:
Thanks @gsmet! Before I merge... @Sanne can you confirm the first commit is fine? Or is the configuration property `additionalRuntimeArgLine` overridden by the product team, and we can't touch it under any circumstance?

integrationtest/performance/orm/pom.xml
```diff
@@ -377,7 +377,7 @@
         <repository>
             <id>jboss-public-repository-group</id>
             <name>JBoss Public Maven Repository Group</name>
-            <url>${repository.jboss-public.url}</url>
+            <url>${jboss.public.repo.url}</url>
```
Sanne: This is the perfect example regarding my previous comment :)

Don't you think it would have been better if the property had never been changed in f25b4ed to begin with?

A lot of times a "cleanup" just creates more work for other teams / other components to have to a)figure out why stuff no longer works b) adjust things

integrationtest/performance/orm/pom.xml
```diff
@@ -222,7 +222,7 @@
                         <goals>
                             <goal>build</goal>
                         </goals>
-                        <phase>process-test-resources</phase>
+                        <phase>generate-test-resources</phase>
```
Sanne: Very interesting find! Thanks

integrationtest/performance/orm/pom.xml
```diff
@@ -30,7 +30,7 @@
         <jbosshome>${project.build.directory}/${serverName}/</jbosshome>
 
         <!-- Relax the JVM restrictions for integration tests, when using Jigsaw -->
-        <additionalRuntimeArgLine>${additionalRuntimeArgLineForWildFlyTests}</additionalRuntimeArgLine>
+        <test.jvm.args>${test.wildfly.jvm.args}</test.jvm.args>
```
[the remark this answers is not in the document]
yrodiere: > also doesn't really help to make it clear that this is not a version?

Well let's be reasonable, version properties start with `version.`, I think that's a pretty clear indication those are versions and other properties are not :) I don't see why people would start assuming a property is about a version just because it doesn't use camel case.

> why you consider this a cleanup

Simply because it's more consistent with other properties. For instance, other properties that include runtime args all end with `.args`: `surefire.system.args`, `failsafe.jacoco.args`, ... When properties with similar purposes have similar names, it's arguably easier to understand they are related.

I wouldn't have done it just for the sake of consistency, but since I was working on this part of the configuration, and had to understand how it worked before I made other changes, I thought I might as well update the configuration to reflect what I understood.

Now if you think those names are not clear enough, I can of course change them.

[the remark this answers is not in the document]
Sanne: don't worry, I don't care :)

I'm just suggesting you shouldn't care either, such changes just cause more work, risk of trouble, and _maybe_ they add some value but I doubt it.

Sanne:
It's still not working for me in IDEA, as this property is not resolved:
 
    <byteman.agent.path>${org.jboss.byteman:byteman:jar}</byteman.agent.path>

I wonder if it's an IDEA specific limitation as this doesn't seem changed recently and AFAIK it worked fine in Eclipse.
Trying now some older branch in IDEA..

integrationtest/performance/orm/pom.xml
gsmet: @Sanne 
Hmmm, you were the one initially suggesting that we should try to improve the consistency with the JBoss world, that's the reason of f25b4ed. I admit it sure would have been better if I hadn't missed a few things when doing it.

Sanne: that's different; having the same version property format has indeed been requested by the productization team, and is useful to compare e.g. with WildFly poms.

gsmet: Yeah, taking a look at it, this thing has been added while I was doing the clean up, that's why I missed them.

gsmet: You might have missed that I used the same variable names than in the JBoss parent pom: https://github.com/jboss/jboss-parent-pom/blob/master/pom.xml#L118

That's why I changed them :).

Sanne:
Not working on branch 5.5 with latest IDEA 2018.1.4 either. Does it work for you @yrodiere ?

If not, I'd still like to merge these as it does go in the right direction - but it doesn't solve the problem so maybe we could repurpose/rename the JIRA description to solve the jacoco properties problem only.

integrationtest/performance/orm/pom.xml
Sanne: this whole discussion is missing the point :)
I am not questioning the previous patches, I'm just suggesting to be careful with further property renames. Clearly, this commit is a necessary fix.

yrodiere:
@Sanne As discussed on HipChat, IDEA is working against us in this case, so I'll work on something more useful. I repurposed the ticket and will merge.

PR #1687, opened 2018-05-30
HSEARCH-3183 Remove obsolete and invalid Maven configuration in hibernate-search-jbossmodules-parent

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3183

### DIFF
--- a/jbossmodules/pom.xml
+++ b/jbossmodules/pom.xml
@@ -46,31 +46,4 @@
         -->
         <enforcer.dependencyconvergence.skip>true</enforcer.dependencyconvergence.skip>
     </properties>
-
-    <build>
-        <pluginManagement>
-            <plugins>
-                <plugin>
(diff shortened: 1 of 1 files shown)


gsmet:
Looks good, you can merge it.

gsmet:
@Sanne before this PR, we did override the parent pom enforcer configuration, now we simply use it.

yrodiere:
@Sanne Sorry I didn't see your comment. The answer is [yes we do in the root pom](https://github.com/yrodiere/hibernate-search-6-poc/blob/c9b2906372e1f0c07bc552b4b0fb436fda0960f0/pom.xml#L436).

This was actually just a copy of the root pom configuration to remove the dependency convergence rule, but still keep the other rules. This is no longer necessary since, a few months ago, I set up a way to disable the dependency convergence rule on a per-module basis through properties.

yrodiere:
Merged, thanks!

PR #1691, opened 2018-06-01
HSEARCH-3194 Upgrade feature packs and integration tests to WildFly 13

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3194


Sanne:
Nice cleanup! No problems caused by the fact that the ORM modules depend on WF12? Was the dependency override good enough to handle that?

Sanne:
cool, thanks for checking. Sounds like the main dependency already applies the override - might make sense.

Sanne:
merged (first commit only).

PR #1694, opened 2018-06-14
Backport to 5.11: HSEARCH-3197 Allow to set the minimum number of should clauses required to match for boolean predicates in the DSL

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3197

This is a backport of work done in the Search 6 proof of concept: https://github.com/hibernate/hibernate-search-6-poc/pull/68

I think the syntax works fine even with Search 5. The only doubt I have is regarding the `.end()` syntax, which we don't have anywhere else in Search 5 and might be renamed in 6 at some point (to `.up()` for instance).

### DIFF
--- a/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ToElasticsearch.java
+++ b/elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ToElasticsearch.java
@@ -344,6 +344,11 @@ public class ToElasticsearch {
 
 		clauses.append( boostAppender( booleanQuery ) );
 
+		int minimumShouldMatchNumber = booleanQuery.getMinimumNumberShouldMatch();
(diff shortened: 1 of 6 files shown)


elasticsearch/src/main/java/org/hibernate/search/elasticsearch/impl/ToElasticsearch.java
```diff
@@ -344,6 +344,11 @@ else if ( filters.size() == 1 ) {
 
 		clauses.append( boostAppender( booleanQuery ) );
 
+		int minimumShouldMatch = booleanQuery.getMinimumNumberShouldMatch();
```
gsmet: You call it `minimumShouldMatchNumber` with `number` at the end elsewhere.

yrodiere:
Fixed the problem you mentioned and merged. Thanks!

PR #1695, opened 2018-06-14
Backport HSEARCH-3197 to branch 5.10 - Allow to set the minimum number of should clauses required to match for boolean predicates in the DSL

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3197

Backport of #1694 , which should be merged first.


yrodiere:
Merged, as #1694 was approved.

PR #1696, opened 2018-06-22
HSEARCH-3215 Reference guide typos

sant0s:


Sanne:
good catches! thank you so much

PR #1697, opened 2018-06-22
HSEARCH-3216 Deprecated API usage in example Implementing a class bridge

sant0s:


documentation/src/main/asciidoc/mapping.asciidoc
```diff
@@ -2002,9 +2002,7 @@ public class CatFieldsClassBridge implements FieldBridge, ParameterizedBridge {
             fieldValue2 = "";
         }
         String fieldValue = fieldValue1 + sepChar + fieldValue2;
-        Field field = new Field( name, fieldValue, luceneOptions.getStore(),
-            luceneOptions.getIndex(), luceneOptions.getTermVector() );
-        field.setBoost( luceneOptions.getBoost() );
+        Field field = new StringField( name, fieldValue, luceneOptions.getStore() );
```
yrodiere: Actually, you should rather replace that with `luceneOptions.addFieldToDocument( name, fieldValue, document )`, which will add all the correct options under the hood.

yrodiere: And remove the line below, of course.

Thanks for spotting this, by the way :)

yrodiere: For your own information (since I guess you are using this kind of code): according to its javadoc `StringField` is never tokenized. `TextField` is, on the other hand. So there are a few subtleties to keep in mind with these non-deprecated types...

sant0s: Thanks.

yrodiere:
Rebased, squashed and merged. Thanks!

PR #1698, opened 2018-06-27
Formatted code to avoid validation problems with future checkstyle releases

ps-sp:
Checkstyle is updating one of its checks (`RightCurlyCheck`) in accordance with issue [#4022](https://github.com/checkstyle/checkstyle/issues/4022). The corresponding PR [#4247](https://github.com/checkstyle/checkstyle/pull/4247) has wercker CI failure due to an unexpected violation with Hibernate Search. 

Please have a look at [PR #4247 comment](https://github.com/checkstyle/checkstyle/pull/4247#issuecomment-400294580) and the ones that follow for further reference.

This PR formats the code so as to sync the style with the latest updates in checkstyle.

PR #1703, opened 2018-07-15
Formatted Code

ps-sp:
Similar to #1698. Wercker CI bails out on the first error it sees. Initial wercker failure resulted in the creation of #1698. Another count of similar inappropriate formatting was realized soon after but by then #1698 had been merged. Right now Wercker passes on the master of my fork of hibernate-search.

We sincerely apologize for the inconvenience, as we should have rather fixed everything using my fork of hibernate-search first and only then created #1698 with all the necessary changes in a single PR.

Anyhow, please merge this PR to avoid problems with future checkstyle releases.

Apologies again. Thank you.


Sanne:
Thanks! Added reference to the existing HSEARCH-3218 and merged.

PR #1708, opened 2018-07-16
HSEARCH-3226 Add test dependencies to javax.xml.bind when building with JDK9 and above

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3226

This does not fix everything, far from it, but at least the tests now mostly pass with JDK11.

Disabling a feature of JAXB is necessary in order for the tests to run with JDK11, because this feature makes use of `sun.misc.Unsafe`. So I disabled it in tests. The feature is apparently going to be removed in the next release of JAXB, see the commit message.

In the end I had to add a dependency to javax.xml.bind in the Elasticsearch module. For now the dependency only shows up when building with JDK9 and above, which means our released JARs simply will not work with JDK11. I will try to remove the dependency as part of https://hibernate.atlassian.net/browse/HSEARCH-3237, and if I can't, we'll see what we want to do. On the bright side, we only rely on the API, not on the runtime.

### DIFF
--- a/elasticsearch/pom.xml
+++ b/elasticsearch/pom.xml
@@ -175,4 +175,43 @@
             </plugin>
         </plugins>
     </build>
+
+    <profiles>
+        <profile>
+            <id>jdk9</id>
+            <activation>
+                <jdk>9</jdk>
+            </activation>
+            <dependencies>
+                <dependency>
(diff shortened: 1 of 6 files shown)


Sanne:
merged!

PR #1709, opened 2018-07-19
HSEARCH-3240 Use the pre-installed JDK in the Travis build instead of trying to update it in each build

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3240


yrodiere:
Merged following the discussion on #1710

PR #1710, opened 2018-07-19
Backport HSEARCH-3240 to branch 5.10 - Use the pre-installed JDK in the Travis build instead of trying to update it in each build

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3240

#1709 should be merged first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@
 # therefore deactivated. Activating Travis for your own fork is as easy as
 # activating it in the travis site of your fork.
 
-sudo: required
+sudo: false
 dist: trusty
 language: java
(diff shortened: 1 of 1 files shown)


gsmet:
I'm a bit worried that the version could change without us seeing it but I suppose that's OK as long as we keep our CI?

yrodiere:
Yes, and I plan to keep our CI. I plan to use it even more, actually, since we're introducing Jenkins pipelines in Search 6.
Also, the version could change without us seing before this patch, too, since we were effectively upgrading to the latest version in each build.

yrodiere:
Merged, thanks!

PR #1711, opened 2018-07-19
HSEARCH-3241 Reorganize and clarify Maven properties used to pass JVM args

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3241

PR #1712, opened 2018-07-19
JDK11 related 5.10 backports

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3219
https://hibernate.atlassian.net//browse/HSEARCH-3224
https://hibernate.atlassian.net//browse/HSEARCH-3232
https://hibernate.atlassian.net//browse/HSEARCH-3236
https://hibernate.atlassian.net//browse/HSEARCH-3235
https://hibernate.atlassian.net//browse/HSEARCH-3226
https://hibernate.atlassian.net//browse/HSEARCH-3241

Must wait for #1711 to be merged first.

Not sure it's worth it, let's discuss this.

We want to make Search work with JDK11 ultimately, and a version of HSearch compatible with JDK11 should ship with Wildfly 14, ideally.

We have to options:

* Release a micro on the 5.10 branch and update WF14 to use that micro
* Release a 5.11.0 and update WF14 to use 5.11.0

So far, the amount of changes is not enough to justify a 5.11, so I'd say let's target JDK11 support in 5.10. But on the ORM side we may need to add new dependencies (to JaxB in particular), and I'd argue adding new dependencies in a micro is not very nice.

### DIFF
--- a/elasticsearch/pom.xml
+++ b/elasticsearch/pom.xml
@@ -175,4 +175,43 @@
             </plugin>
         </plugins>
     </build>
+
+    <profiles>
+        <profile>
+            <id>jdk9</id>
+            <activation>
+                <jdk>9</jdk>
+            </activation>
+            <dependencies>
+                <dependency>
+                    <groupId>javax.xml.bind</groupId>
+                    <artifactId>jaxb-api</artifactId>
+                </dependency>
+            </dependencies>
+        </profile>
+        <profile>
+            <id>jdk10</id>
+            <activation>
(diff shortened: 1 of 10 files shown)


gsmet:
+1 to merge this in 5.10. I did the same for HV.

As for ORM, I don't plan to add the JAXB dependency for now: we did that in WF.

When JDK 11 will be out, we will need to add one and we'll see if it's worth releasing a new 5.x.

yrodiere:
Sounds good. Merged, thanks!

PR #1714, opened 2018-07-23
HSEARCH-3233 Overriding an analyzer in the queryBuilder with an unknown analyzer should lead to a proper SearchException

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3233


yrodiere:
Fixed the typo plus a failing test on Elasticsearch and merged. Thanks!

PR #1717, opened 2018-07-24
HSEARCH-3248 Restore lenient illegal access checks on the client side of Arquillian during integration tests

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3248

Tested locally against JDK 9.0.4: it works.

Note the problem probably also affected JDK11, but I can't test it since other failures happen earlier in the bootstrap process.


yrodiere:
Let's hope so. It should get better once we switch to JDK11 as the minimal requirement, but it'll probably be a long time from now...
Merged, thanks!

PR #1724, opened 2018-09-06
HSEARCH-3246 Generate Coveralls and Sonar reports from the Jenkinsfile instead of Travis

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3246

~Based on #1723 , which should be merged first.~ => Done

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,25 +25,12 @@ import org.jenkinsci.plugins.credentialsbinding.impl.CredentialNotFoundException
  */
 
 /*
- * Expected configuration (optional): multibranch-configuration.yaml, set up using the config file provider plugin
- * (https://plugins.jenkins.io/config-file-provider).
- * Expected structure of this file:
- *
- * notification:
- *   email:
- *     recipients: ... # string containing a space-separated list of email addresses to notify in case of failing non-PR builds>
- *
- * The following credentials are necessary for some features:
- *
- * - 'aws-elasticsearch' AWS credentials, to test Elasticsearch as a service on AWS
- * - 'coveralls-repository-token' secret text credentials containing the repository token,
- * to send coverage reports to coveralls.io. Note these credentials should be registered at the job level, not system-wide.
- * - 'sonarcloud-hibernate-token' secret text credentials containing a Sonar access token for the Hibernate organization,
- * to send Sonar analysis input to sonarcloud.io.
- *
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -320,6 +327,24 @@ stage('Default build') {
...
+									-DpullRequest=${env.CHANGE_ID} \\
+							""" : """ \\
+									-Dbranch=${env.BRANCH_NAME} \\
+							"""} \\
+					"""
+				}
+			}
+			catch (CredentialNotFoundException e) {
+				echo "No Coveralls token configured - skipping Coveralls report. Error was: ${e}"
+			}
```
gsmet: You prefer skipping it (and probably it going unnoticed) rather than throwing the exception?

Jenkinsfile
```diff
@@ -61,6 +49,43 @@ import org.jenkinsci.plugins.credentialsbinding.impl.CredentialNotFoundException
...
+ *       email:
+ *         recipients: ... # string containing a space-separated list of email addresses to notify in case of failing non-PR builds>
+ *
+ * #### Credentials
+ *
+ * The following credentials are necessary for some features:
+ *
+ * - 'aws-elasticsearch' AWS credentials, to test Elasticsearch as a service on AWS
+ * - 'coveralls-repository-token' secret text credentials containing the repository token,
+ * to send coverage reports to coveralls.io. Note these credentials should be registered at the job level, not system-wide.
```
gsmet: Does this remark apply to all three credentials or only this one?

Jenkinsfile
yrodiere: It's for the benefit of anyone configuring a job for his fork and not wanting to also set up coveralls, sonar, and so on. For example if you (@gsmet) set up a job for your own fork, I suppose you'll want the job to work out of the box, and only later, maybe, set up additional integrations.

Note that we only ignore the `CredentialNotFoundException`. As long as you provide credentials, errors should be reported correctly and fail the build. Which means in particular that, for the job currently configured on master, this `catch` block has no effect.

Jenkinsfile
yrodiere: Only this one. The token is used both for authentication of the user and of the repository, so we can't use a global one. Other credentials are only used to authenticate the user.

yrodiere:
Merging, thanks

PR #1725, opened 2018-09-07
HSEARCH-3337 Generate external javadoc URLs automatically from version properties

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3337

~This is based on #1723 , which should be merged first.~ => Done

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -176,8 +176,6 @@
         <javadoc.org.apache.lucene.core.url>http://lucene.apache.org/core/${javadoc.org.apache.lucene.tag}/core/</javadoc.org.apache.lucene.core.url>
         <javadoc.org.apache.lucene.analyzers-common.url>http://lucene.apache.org/core/${javadoc.org.apache.lucene.tag}/analyzers-common/</javadoc.org.apache.lucene.analyzers-common.url>
         <javadoc.org.apache.lucene.queryparser.url>http://lucene.apache.org/core/${javadoc.org.apache.lucene.tag}/queryparser/</javadoc.org.apache.lucene.queryparser.url>
-        <!-- TODO define feature pack version for Lucene in this module, not in the legacy module (requires a release of Lucene feature packs) -->
-        <version.feature-pack.org.apache.lucene>OVERRIDDEN_IN_LEGACY</version.feature-pack.org.apache.lucene>
 
         <!-- >>> Elasticsearch -->
         <!-- When updating the following versions you will likely need to update the
--- a/pom.xml
+++ b/pom.xml
@@ -1083,10 +1083,8 @@
                             <link>http://lucene.apache.org/core/${version.org.apache.lucene.javadoc}/core/</link>
                             <link>http://lucene.apache.org/core/${version.org.apache.lucene.javadoc}/analyzers-common/</link>
                             <link>http://lucene.apache.org/core/${version.org.apache.lucene.javadoc}/queryparser/</link>
-                            <link>http://jgroups.org/javadoc/</link>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -1083,10 +1083,8 @@
                             <link>http://lucene.apache.org/core/${version.org.apache.lucene.javadoc}/core/</link>
                             <link>http://lucene.apache.org/core/${version.org.apache.lucene.javadoc}/analyzers-common/</link>
                             <link>http://lucene.apache.org/core/${version.org.apache.lucene.javadoc}/queryparser/</link>
-                            <link>http://jgroups.org/javadoc/</link>
-                            <!-- for javax.batch in the JSR 352 modules -->
+                            <!-- For Java EE APIs, in particular javax.batch in the JSR 352 modules -->
                             <link>https://docs.oracle.com/javaee/${version.javaee.javadoc}/api/</link>
-                            <link>http://docs.jboss.org/jberet/${version.org.jberet.javadoc}/javadoc/jberet-core/</link>
```
gsmet: I think I would have kept it if we have a clear plan to keep the JSR 352 support. Or do you expect this particular jberet part to be dropped (as you keep the EE part).

pom.xml
```diff
@@ -1570,6 +1570,29 @@
...
+                                    <goal>javadoc-no-fork</goal>
+                                </goals>
+                                <phase>prepare-package</phase>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+
```
gsmet: I don't understand why you're doing that? I mean we don't plan to build a distribution of the legacy code?

pom.xml
yrodiere: We will restore JSR 352 support at some point.

However, I'd like to point out that what you're suggesting is the exact opposite of what you wanted for everything else in the POM, namely to only copy the necessary configuration to the new POM and migrate the rest later as we need it (if we need it).

Should I understand that, for the modules that we're sure to restore eventually, you'd like us to move, right now, the configuration from the legacy pom to the parent one? And not for the ones we're not sure about?

I can do what you're asking, but I'll have to move (or copy) some properties from the legacy pom to this one, mainly because empty `<link>` elements trigger an NPE during the build (and if you have a look at the next commits, defining the property only in the legacy pom will lead to empty `<link>` elements).

pom.xml
yrodiere: I don't understand the link between this change and building a distribution of the legacy code.

I'm just making sure we build the javadoc for the *new* modules when the `-Pdist` profile is enabled, which unfortunately we didn't use to.

gsmet: Ah my bad, I didn't see that it was the parent pom.

yrodiere:
Jenkins, retest this please. (just testing the new access tokens)

yrodiere:
Jenkins, retest this please

yrodiere:
Jenkins, retest this please

pom.xml
gsmet: Yeah, I can see how it can look as contradictory but at least in my head, it's a little more subtle. I think it's typically an element we could forget to restore. And if we are sure we will need it at some point, I wouldn't have removed it.

Anyway, it's not a big deal, I can live with it :).

yrodiere:
Jenkins, retest this please

pom.xml
yrodiere: I added a commit to move the JSR 352 properties up to the parent POM. I don't thing there is much else to move, that we could forget about, so I'll leave it to that.

yrodiere:
Merged, thanks!

PR #1726, opened 2018-09-07
Follow-up to HSEARCH-3179: Fix invalid OSGi version range in legacy POM

yrodiere:
I made a mistake in ee805bb

Full build passed: http://ci.hibernate.org/blue/organizations/jenkins/hibernate-search-yoann/detail/HSEARCH-3179/21/pipeline


yrodiere:
The full build passed and the fix is simple enough; merging.

PR #1727, opened 2018-09-07
HSEARCH-840 Split IndexManager interface in API and SPI

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-840

Since there weren't any APIs for index managers, and in Search 6 there isn't any method that could qualify as an API (yet), this is more about creating index manager API interfaces as well as a mechanism to allow users to access these index managers.

I used the same principles for index-scoped APIs (`IndexManager` interface and sub-interfaces) as for backend-scoped APIs (`Backend` interface and sub-interfaces) meaning: the user gets a (generic) `IndexManager` from the `SearchMappingRepository` (yes, the name has to change, there's a ticket about that :wink:), and then is able to unwrap it to a backend-specific interface.

I took this opportunity to also cleanup the backend-scoped APIs (and I mean API, not SPI): there were a few problems with the current implementations.


yrodiere:
Jenkins, retest this please

engine/src/main/java/org/hibernate/search/engine/search/ProjectionConstants.java
```diff
@@ -20,7 +21,7 @@ private ProjectionConstants() {
 	 * Project to an object representing the match.
 	 * <p>
 	 * The actual type of the object depends on the entry point
-	 * for your query query: an {@link org.hibernate.search.engine.backend.index.spi.IndexManager}
+	 * for your query query: an {@link MappedIndexManager}
```
gsmet: s/an/a/

engine/src/main/java/org/hibernate/search/engine/search/ProjectionConstants.java
```diff
@@ -36,7 +37,7 @@ private ProjectionConstants() {
 	 * Project to a reference to the match.
 	 * <p>
 	 * The actual type of the reference depends on the entry point
-	 * for your query query: an {@link org.hibernate.search.engine.backend.index.spi.IndexManager}
+	 * for your query query: an {@link MappedIndexManager}
```
gsmet: Same here.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/index/impl/ElasticsearchIndexSearchTargetBuilder.java
```diff
@@ -49,7 +49,7 @@ void add(SearchBackendContext searchBackendContext, ElasticsearchIndexManager in
 	@Override
 	public IndexSearchTarget build() {
 		// Use LinkedHashSet to ensure stable order when generating requests
-		Set<ElasticsearchIndexModel> indexModels = indexManagers.stream().map( ElasticsearchIndexManager::getModel )
+		Set<ElasticsearchIndexModel> indexModels = indexManagers.stream().map( ElasticsearchIndexManagerImpl::getModel )
```
gsmet: Do we really want to use the impl in this class, given we have an interface?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -94,13 +94,13 @@ void executedRequest(String method, String path, Map<String, String> getParamete
...
 	SearchException cannotMixElasticsearchSearchTargetWithOtherType(IndexSearchTargetBuilder firstTarget,
-			ElasticsearchIndexManager otherTarget, @Param EventContext context);
+			ElasticsearchIndexManagerImpl otherTarget, @Param EventContext context);
 
 	@Message(id = ID_OFFSET_3 + 3,
 			value = "A search query cannot target multiple Elasticsearch backends."
 					+ " First target was: '%1$s', other target was: '%2$s'")
 	SearchException cannotMixElasticsearchSearchTargetWithOtherBackend(IndexSearchTargetBuilder firstTarget,
-			ElasticsearchIndexManager otherTarget, @Param EventContext context);
+			ElasticsearchIndexManagerImpl otherTarget, @Param EventContext context);
```
gsmet: Same question here: I would have used the interface.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/index/impl/LuceneIndexSearchTargetBuilder.java
```diff
@@ -31,14 +31,14 @@
 	private final SearchBackendContext searchBackendContext;
 
 	// Use LinkedHashSet to ensure stable order when generating requests
-	private final Set<LuceneIndexManager> indexManagers = new LinkedHashSet<>();
+	private final Set<LuceneIndexManagerImplementor> indexManagers = new LinkedHashSet<>();
```
gsmet: We are not consistent between Es and Lucene. See here: https://github.com/hibernate/hibernate-search/pull/1727/commits/deaa1adba34aea50211e69c5ea80c9df22243c7c#diff-e34a887279fa744ebd6fbd73ab2e553aR33 .

Is it something expected e.g. you use the `Impl` in Es and an `Implementor` interface in Lucene?

yrodiere: This inconsistency is pre-existing: there's an (internal) interface for index managers on the Lucene side, but not on the Elasticsearch side.

On the Elasticsearch side the interface is clearly useless, because there's no reason to have multiple index manager implementations; worst case some *internal* components could have multiple implementations.

On the Lucene side, we talked about it, and I wanted to just remove the interface, but you were worried that we might need multiple index manager types in the future and thought it was better this way. Maybe there was more to it, but then I don't remember.

Currently the interface is only used in a single package, which is the package where its only implementation is defined, so the interface is not really useful.

Maybe I should just replace references to the interface with references to the implementation class, and we can reintroduce the interface later if necessary, once we know what exactly is generic to every implementation and what is specific to the "directory-based" index managers?

Note that in any case, since the "index manager" already implements SPIs defined in the engine, in my mind it already has one responsibility and should not assume another one. If we ever need index-scoped, backend-specific, internal operations to be exposed to other packages, I'd rather create separate objects, much like what we did with `org.hibernate.search.backend.lucene.index.impl.IndexingBackendContext` and `org.hibernate.search.backend.lucene.search.query.impl.SearchBackendContext`. It helps keep the inter-package dependencies straightforward and manageable.

yrodiere: Ok, I added two commits just for that.

gsmet: Looks good.

yrodiere:
Merging, thanks!

PR #1729, opened 2018-09-10
HSEARCH-3339 @SpatialBridge in "contained" entity triggers exception at bootstrap

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3339


gsmet:
Looks as good as it can be :)

yrodiere:
Merged, thanks

PR #1736, opened 2018-09-10
HSEARCH-3346 Only enable Sonar analysis on select Jenkins jobs

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3346

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -69,7 +69,7 @@ import org.jenkinsci.plugins.credentialsbinding.impl.CredentialNotFoundException
  *
  * The configuration file is optional. Its purpose is to host job-specific configuration, such as notification recipients.
  *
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -388,9 +391,6 @@ stage('Default build') {
 					"""
 				}
 			}
-			catch (CredentialNotFoundException e) {
-				echo "No SonarCloud token configured - skipping SonarCloud report. Error was: ${e}"
-			}
```
gsmet: See, you end up removing this catch ;).

yrodiere: Yeah, if someone explicitly configured a sonar organization and credentials are not found, then it's definitely an error :)

yrodiere:
Merged, thanks

PR #1737, opened 2018-09-11
HSEARCH-3341 Do not send SUCCESS notifications when a  Jenkins pipeline build fails or is cancelled

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3341

This one is a bit embarrassing... I tested it, it should work now: I got a FAILURE notification for http://ci.hibernate.org/blue/organizations/jenkins/hibernate-search-yoann/detail/ci-test-this-please/8/pipeline .

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -522,13 +522,23 @@ stage('Release') {
 	}
 }
 
-currentBuild.result = 'SUCCESS'
 } // End of the code triggering notifications
 catch (any) {
 	mainScriptException = any
 	throw any
 }
 finally {
+	/*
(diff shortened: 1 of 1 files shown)


yrodiere:
... And it doesn't work for cancelled builds. Having a look.

yrodiere:
Ok, so apparently it's not possible to distinguish between cancelled builds and failing builds, because of [this bug](https://issues.jenkins-ci.org/browse/JENKINS-28822). So any cancelled build will be notified as a failure.

I'd say it's close enough.

gsmet:
Looks good to me.

yrodiere:
Merged, thanks.

PR #1738, opened 2018-09-11
HSEARCH-3348 Consider backend TCK and test utils as test code during the Sonar analysis

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3348

It's a rather ugly hack, but I couldn't find any other way. At least it should not cause much maintenance hassle.

Results of a full analysis after this patch: https://sonarcloud.io/project/activity?branch=sonartest-yrodiere&id=org.hibernate.search%3Ahibernate-search-parent&selected_date=2018-09-11T11%3A54%3A22%2B0000

To be compared with: https://sonarcloud.io/project/activity?id=org.hibernate.search%3Ahibernate-search-parent&selected_date=2018-09-11T12%3A13%3A24%2B0000

So yeah, that's quite a lot of noise we got rid of.

### DIFF
--- a/.empty/README.md
+++ b/.empty/README.md
@@ -1,0 +1,1 @@
+Do not move, rename or remove this directory, it *is* useful. See property 'rootProject.emptySubdirectory' in the parent POM.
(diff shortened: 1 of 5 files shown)


gsmet:
Looks good.

yrodiere:
Merged, thanks

PR #1739, opened 2018-09-11
HSEARCH-383 Hibernate Search does not respect the @AccessType annotation in respect to @Id fields.

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-383

~Based on #1740, which should be merged first.~ => Done

I doubt the problem was still relevant even in Search 5, but the test makes sense, so let's just add it.

### DIFF
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/model/ProxyIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/model/ProxyIT.java
@@ -1,0 +1,136 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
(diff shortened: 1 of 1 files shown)


integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/model/ProxyIT.java
```diff
@@ -0,0 +1,136 @@
...
+		public String getText() {
+			return text;
+		}
+
+		public void setText(String text) {
+			this.text = text;
+		}
+	}
+
+	@Entity(name = "propertyAccess")
```
gsmet: `PropertyAccess` with a capital maybe?

yrodiere: Done

yrodiere:
Merged, thanks

PR #1740, opened 2018-09-12
HSEARCH-3359 Search 6 groundwork - Cleanup and organize mapper integration tests

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3359


util/internal/integrationtest/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/orm/OrmSetupHelper.java
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	@Override
+	protected SetupContext createSetupContext() {
+		return new SetupContext();
+	}
+
+	@Override
+	protected void close(SessionFactory toClose) {
+		// TODO
```
gsmet: Hmmm?

yrodiere: Good catch, I'll fix that.

yrodiere:
Merged, thanks!

PR #1741, opened 2018-09-12
HSEARCH-2759 + HSEARCH-2847 : ITs regarding how GeoPointBridge accesses entity data 

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2759
https://hibernate.atlassian.net/browse/HSEARCH-2847

~This PR is based on #1740 , which should be merged first.~ => Done

### DIFF
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/model/BridgeUsingPropertyMarkerAccessIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/model/BridgeUsingPropertyMarkerAccessIT.java
@@ -38,7 +38,7 @@ import org.junit.runners.Parameterized;
  * @param <TIndexed> The entity class under test.
  */
 @RunWith(Parameterized.class)
-@TestForIssue(jiraKey = "HSEARCH-2759")
+@TestForIssue(jiraKey = { "HSEARCH-2759", "HSEARCH-2847" })
 public class BridgeUsingPropertyMarkerAccessIT<TIndexed> {
 
 	private static final String INDEX_NAME = "IndexedEntity";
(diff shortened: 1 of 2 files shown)


integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/model/BridgeUsingPropertyMarkerAccessIT.java
```diff
@@ -0,0 +1,196 @@
...
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+/**
+ * Test that a bridge using markers to retrieve and access properties
+ * will behave as expected regarding access mode.
+ *
```
gsmet: A `<p>` would be in order (even if not an API, I think it's better to have a proper Javadoc).

util/internal/test/src/main/java/org/hibernate/search/util/impl/test/annotation/TestForIssue.java
```diff
@@ -18,11 +19,18 @@
...
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ ElementType.METHOD, ElementType.TYPE })
+@Repeatable(TestForIssue.List.class)
 public @interface TestForIssue {
 	/**
 	 * The key of a JIRA issue tested.
 	 *
 	 * @return The jira issue key
 	 */
 	String jiraKey();
```
gsmet: In another project (see https://github.com/hibernate/hibernate-ogm/blob/master/core/src/test/java/org/hibernate/ogm/utils/TestForIssue.java#L28), we made `jiraKey` an array.

yrodiere: Right. I amended the commit to do the same.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/model/BridgeUsingPropertyMarkerAccessIT.java
yrodiere: Right, done.

yrodiere:
Merged, thanks

PR #1745, opened 2018-09-17
HSEARCH-3340 + HSEARCH-3334 Add support for snapshot releases to the Jenkinsfile

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3340
https://hibernate.atlassian.net//browse/HSEARCH-3334

This PR is twofold:

* First, as part of HSEARCH-3340, I tried to restore something similar to nightly releases for master and maintenance branches. As it turns out, it's easier to just deploy snapshot artifacts as part of the main build, so that's what I did. Apparently the JBoss Nexus removes the older artifacts for a given snapshot version each time we deploy, so this shouldn't cause storage space issues.
* Second, since the snapshot artifact deployment is now done as part of the main build, a failing integration test could lead to a partial deployment: some artifacts are on Nexus, others are not. I had a look at HSEARCH-3334, and it turns out that just using the `nexus-staging-maven-plugin`, which replaces the (apparently useless) `nexus-maven-plugin`, does the trick: it allows us to stage artifacts locally, and only push them to the Nexus repository once every module built successfully.

This brings us one step closer to continuous delivery and proper integration testing, where we would actually test the JARs that we pushed to the staging repository, instead of what we do currently (build a first time, test, then build a second time and release). While we're still far from it, and I don't intend to actively work toward it for now, I still think it's nice to at least clear the way.

Regarding the Maven configuration, I tested both the snapshot deployment (on a local Nexus) and the non-snapshot deployment (with remote staging, on the JBoss Nexus). It worked as expected.
Regarding the pipeline, I tested snapshot deployment and it worked well too: http://ci.hibernate.org/blue/organizations/jenkins/hibernate-search-yoann/detail/ci-test-this-please/12/pipeline/23 (don't mind the subsequent IT failures, they are unrelated).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -366,7 +366,7 @@ stage('Default build') {
 		return
 	}
 	node(NODE_PATTERN_BASE) {
-		checkout scm
+		initWorkspace()
 		withDefaultedMaven {
 			sh """ \\
 					mvn clean install -Pdist -Pcoverage -Pjqassistant \\
@@ -431,8 +431,8 @@ stage('Non-default environment ITs') {
 	jdkEnvs.each { itEnv ->
 		executions.put(itEnv.tag, {
 			node(NODE_PATTERN_BASE) {
+				initWorkspace()
 				withDefaultedMaven(jdk: itEnv.tool) {
-					checkout scm
 					mavenNonDefaultIT itEnv,
 							"clean install --fail-at-end"
 				}
@@ -444,6 +444,7 @@ stage('Non-default environment ITs') {
 	databaseEnvs.each { itEnv ->
 		executions.put(itEnv.tag, {
 			node(NODE_PATTERN_BASE) {
+				initWorkspace()
 				withDefaultedMaven {
 					resumeFromDefaultBuild()
 					mavenNonDefaultIT itEnv, """ \\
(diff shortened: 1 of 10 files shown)


Jenkinsfile
```diff
@@ -374,6 +374,7 @@ stage('Default build') {
 		return
 	}
 	node(NODE_PATTERN_BASE) {
+		cleanWs()
```
gsmet: Hum, does that mean that we will need to download again all the dependencies before building? Wondering if we could maybe only clean the artifacts we built?

yrodiere: I implemented a more selective cleanup and it seems to work, as this build confirms: http://ci.hibernate.org/blue/organizations/jenkins/hibernate-search-yoann/detail/ci-test-this-please/31/pipeline/68 (see how the apache artifacts are there all along, but the Hibernate Search artifacts are deleted before the build).
I had to also avoid cleaning up the `.git` directory, because deleting it would cause the next `checkout scm` command to clean up the whole workspace.

It turns out the workspace is only shared among all the builds **for the same branch/PR**. So t's really unlikely for the in-workspace cache to be of any use across builds. But it'll be there...

yrodiere:
Merged, thanks!

PR #1748, opened 2018-09-19
HSEARCH-2765 Rename the SearchFactory API

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2765

Context: as part of the Search 6 proof-of-concept, the `SearchFactory` has already been replaced by another component, the `SearchMappingRepository`. But as we added more features, this component began to provide access to more than just the mappings (the backends, the index managers), and thus the name started to be less and less fitting. We have to find a better name for it.

I ended up using the name `SearchIntegration` for the main component of Hibernate Search, the one that can provide access to all backends, indexes and mappings.

I'm not very happy with this name, but I couldn't find anything better, so it'll have to do for now.

I also thought of the following:

* `SearchAssembly`, because that's what it is: an assembly of all the top-level Search components. But to me at least, it rings like the output of some manufacturing plant... Not really the impression I'm looking for.
* `SearchManager`, because of the similarity with the `CacheManager` in Infinispan. But then we will have to rename the `PojoSearchManager` et. al., so named because of their similarity with the `EntityManager` in JPA. `SearchSession` maybe? But even then it will be weird to have a `FullTextEntityManager` and a `SearchManager` with radically different roles.
* `SearchIntegrator`, as a variation of `SearchIntegration`. But this makes it look like the `SearchIntegrator` allows further integration, while in fact it's just the result of it.
* `SearchInstance`, but that's awfully generic, to the point of being barely understandable (this could refer to a search query, for all we know).

Suggestions are welcome...


gsmet:
The new name is not that bad IMHO.

yrodiere:
Merged, thanks

PR #1749, opened 2018-09-19
HSEARCH-3263 Remove support for Elasticsearch < 5.6

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3263

I also removed implementation code for ES2 and ES50 from the legacy code, because we will have to migrate most of this code to Search 6 at some point, and I'd rather untangle this mess now than later :)


README.md
```diff
@@ -111,10 +109,10 @@ and run Elasticsearch-related tests against your own server using the
 
     > mvn clean install -Dtest.elasticsearch.host.provided=true -Dtest.elasticsearch.host.url=http://localhost:9200
 
-If you want to run tests against an older Elasticsearch version  (2.x for instance),
+If you want to run tests against a different Elasticsearch version  (6.x for instance),
 you will still have to select a profile among those listed above, and disable the default profile:
 
-    > mvn clean install -P!elasticsearch-5.2,elasticsearch-2.2 -Dtest.elasticsearch.host.provided=true -Dtest.elasticsearch.host.url=http://localhost:9200
+    > mvn clean install -P!elasticsearch-5.6,elasticsearch-6.0 -Dtest.elasticsearch.host.provided=true -Dtest.elasticsearch.host.url=http://localhost:9200
```
gsmet: Here you negate the `elasticsearch-5.6` profile, but you don't do it there: https://github.com/hibernate/hibernate-search/pull/1749/commits/6e9a18363128d6eefd754277a705033f4a5fe677#diff-600376dffeb79835ede4a0b285078036R2003 .

pom.xml
```diff
@@ -2000,234 +2000,12 @@
              To run tests against a different version of Elasticsearch, use the appropriate profile
              and provide the Elasticsearch version:
 
-                 mvn clean test -Pelasticsearch-2.0 -Dtest.elasticsearch.version=2.1.0
+                 mvn clean test -Pelasticsearch-6.0 -Dtest.elasticsearch.version=6.0.0
           -->
 
-        <!-- Elasticsearch [2.0,2.2) test environment -->
+        <!-- Elasticsearch 5.6+ test environment (default) -->
```
gsmet: Shouldn't the latest stable be the default?

Jenkinsfile
```diff
@@ -133,20 +133,12 @@ import org.jenkinsci.plugins.credentialsbinding.impl.CredentialNotFoundException
...
 		new EsLocalITEnvironment(versionRange: '[6.0,6.x)', mavenProfile: 'elasticsearch-6.0', status: ITEnvironmentStatus.SUPPORTED)
 ]
 @Field EsLocalITEnvironment defaultEsLocalEnv
 @Field final List<EsAwsITEnvironment> esAwsEnvs = [
-		// TODO add support for AWS (needs the plugin currently only available in Search 5)
-		new EsAwsITEnvironment(version: '2.3', mavenProfile: 'elasticsearch-2.0', status: ITEnvironmentStatus.EXPERIMENTAL),
-		new EsAwsITEnvironment(version: '5.1', mavenProfile: 'elasticsearch-5.0', status: ITEnvironmentStatus.EXPERIMENTAL),
-		new EsAwsITEnvironment(version: '5.3', mavenProfile: 'elasticsearch-5.2', status: ITEnvironmentStatus.EXPERIMENTAL),
-		new EsAwsITEnvironment(version: '5.5', mavenProfile: 'elasticsearch-5.2', status: ITEnvironmentStatus.EXPERIMENTAL),
+		new EsAwsITEnvironment(version: '5.6', mavenProfile: 'elasticsearch-5.2', status: ITEnvironmentStatus.EXPERIMENTAL),
```
gsmet: The `mavenProfile` property here should be updated in the subsequent commit (https://github.com/hibernate/hibernate-search/pull/1749/commits/6e9a18363128d6eefd754277a705033f4a5fe677) but AFAICS it is not.

README.md
yrodiere: Right, that's because it's not necessary when the version is provided. I'll add it everywhere for consistency.

yrodiere: On second thought I won't, it just adds one more string that we may forget to update...

pom.xml
yrodiere: Ultimately it should, but right now this would break the integration tests in the legacy code, so... Let's do that when we get rid of the legacy code?

yrodiere: Note we execute tests against 6.0 regardless, that's just a matter of cleanliness.

Jenkinsfile
yrodiere: Good catch, fixed.

yrodiere:
Merged, thanks!

PR #1755, opened 2018-09-21
[DO NOT MERGE] Tracking patch for ORM 5.4 on master branch

yrodiere:
This PR is just a way to regularly try to build the master branch against the latest Hibernate ORM 5.4 snapshot, in order to check that no backward incompatibility is introduced.


yrodiere:
Turns out this doesn't work well. Will try another approach.

PR #1761, opened 2018-10-03
HSEARCH-3384 Search 6 groundwork - Restore support for ClassLoaderService

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3384


engine/src/main/java/org/hibernate/search/engine/environment/classpath/spi/ResourceResolver.java
```diff
@@ -0,0 +1,40 @@
...
+	 *
+	 * @param name The resource name.
+	 *
+	 * @return The stream of the located resource; may return {@code null} to indicate the resource was not found
+	 */
+	InputStream locateResourceStream(String name);
+}
+
+
+
```
gsmet: Lots of new lines here.

engine/src/main/java/org/hibernate/search/engine/environment/classpath/spi/DefaultClassAndResourceResolver.java
```diff
@@ -19,18 +19,18 @@
  *
  * @author Hardy Ferentschik
  */
-public final class DefaultClassResolver implements ClassResolver {
+public final class DefaultClassAndResourceResolver implements ClassResolver, ResourceResolver {
```
gsmet: I'm not sure I would have use the same class for the 2 interfaces. It seems a bit confusing to me.

Not a veto, just a remark.

yrodiere:

PR #1762, opened 2018-10-04
HSEARCH-3115 Search 6 groundwork - Restore the analyzer infrastructure

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3115

Goals:

* Reach the same level of feature coverage as Search 5 when it comes to defining named analyzers/normalizers.
* Do not bloat the mapper/engine with complex analyzer reference concepts: only manipulate names (strings) there, and expect analyzer/normalizer definition and resolution to be backend-specific.
* Provide an alternative to the `@Analyzer(impl = ...)` annotation from Search 5 (analyzers referenced by their type). This is done in the commit with message "HSEARCH-3115 Allow to assign name to Lucene analyzer/normalizer instances".

Non-goals:

* Provide default analyzer definitions in the Lucene backend. This will be implemented as part of [HSEARCH-2584](https://hibernate.atlassian.net/browse/HSEARCH-2584)
* Restore `AnalyzerDef`/`NormalizerDef` annotations. This won't happen in the main code, and will only be reintroduced int he compatibility module.
* Restore `Analyzer`/`Normalizer` and  their (quite complex) scoping mechanism. This won't happen in the main code, and will only be reintroduced int he compatibility module.
* Implement the complex mechanism we had in Search 5 to only push the analyzers that are actually referenced in each index to that index's settings. We'll implement [HSEARCH-3309](https://hibernate.atlassian.net/browse/HSEARCH-3309) instead
* Implement the complex Lucene => Elasticsearch analyzer definition translator we had in Search 5. This translator will be removed in Search 6.


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/analysis/impl/LuceneAnalysisComponentFactory.java
```diff
@@ -56,6 +56,10 @@ public Analyzer createAnalyzer(String name, TokenizerFactory tokenizerFactory,
 	public Analyzer createNormalizer(String name,
 			CharFilterFactory[] charFilterFactories, TokenFilterFactory[] filterFactories) {
 		Analyzer normalizer = new TokenizerChain( charFilterFactories, KEYWORD_TOKENIZER_FACTORY, filterFactories );
+		return wrapNormalizer( name, normalizer );
+	}
+
+	public Analyzer wrapNormalizer(String name, Analyzer normalizer) {
 		return new HibernateSearchNormalizerWrapper( normalizer, name );
```
gsmet: Minor nitpicking but it's weird to have the order of parameters reversed between these 2 lines.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/logging/impl/Log.java
```diff
@@ -68,10 +76,21 @@
 			value = "Unable to close the index reader. %1$s")
 	void unableToCloseIndexReader(@FormatWith(EventContextFormatter.class) EventContext context, @Cause Exception e);
 
+	@LogMessage(level = WARN)
+	@Message(id = ID_OFFSET_1 + 75,
+			value = "Configuration setting " + SearchBackendLuceneSettings.LUCENE_VERSION
+			+ " was not specified: using LATEST. %1$s")
+	void recommendConfiguringLuceneVersion(@FormatWith(EventContextFormatter.class) EventContext context);
```
gsmet: I wonder if we should provide a sensible value depending on the current Lucene version. That might help people initializing an application.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -285,4 +285,9 @@ SearchException indexManagerUnwrappingWithUnknownType(@FormatWith(ClassFormatter
 			value = "Cannot apply both an analyzer and a normalizer." )
 	SearchException cannotApplyAnalyzerAndNormalizer(@Param EventContext context);
 
+	@Message(id = ID_OFFSET_3 + 36,
+			value = "Cannot apply an analyzer on a sortable field. Use a normalizer instead."
+					+ " If an actual analyzer (with tokenization) is necessary, define two separate fields:"
+					+ " one that with an analyzer is not sortable, and one with a normalizer that is sortable.")
```
gsmet: `one that with`
`.....^`

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/logging/impl/Log.java
```diff
@@ -133,6 +134,12 @@
 			+ " but you should fix your normalizer definition." )
 	void normalizerProducedMultipleTokens(String normalizerName, int token);
 
+	@Message(id = ID_OFFSET_1 + 345,
+			value = "Cannot apply an analyzer on a sortable field. Use a normalizer instead."
+			+ " If an actual analyzer (with tokenization) is necessary, define two separate fields:"
+			+ " one that with an analyzer is not sortable, and one with a normalizer that is sortable.")
```
gsmet: `one that with`
`.....^`

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/dsl/impl/StringIndexSchemaFieldContext.java
```diff
@@ -73,10 +73,9 @@ public StringIndexSchemaFieldContext sortable(Sortable sortable) {
 	@Override
 	protected void contribute(IndexSchemaFieldDefinitionHelper<String> helper, LuceneIndexSchemaNodeCollector collector,
 			LuceneIndexSchemaObjectNode parentNode) {
-		// TODO For now, we allow the sortable - analyzed combination, it will be disallowed later
-//		if ( Sortable.YES.equals( getSortable() ) && analyzer != null ) {
-//			throw log.cannotUseAnalyzerOnSortableField( getRelativeFieldName() );
-//		}
+		if ( Sortable.YES.equals( sortable ) && analyzer != null ) {
+			throw log.cannotUseAnalyzerOnSortableField( getSchemaContext().getEventContext() );
+		}
```
gsmet: In the Es backend, you have this `switch` construct. Either it brings something to the plate (having a warning when a new value is added) and we should use it in both backends or it doesn't and I think we should simplify the Es backend.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/predicate/impl/StringMatchPredicateBuilder.java
```diff
@@ -30,7 +31,14 @@
 	@Override
 	protected Query doBuild(LuceneSearchPredicateContext context) {
 		if ( queryBuilder != null ) {
-			return queryBuilder.createBooleanQuery( absoluteFieldPath, value );
+			Query analyzed = queryBuilder.createBooleanQuery( absoluteFieldPath, value );
+			if ( analyzed == null ) {
+				// Either the value was an empty string
+				// or the analysis removed all tokens (that can happen if the value contained only stopwords, for example)
+				// In any case, use the same behavior as Elasticsearch: don't match anything
+				analyzed = new MatchNoDocsQuery( "No tokens after analysis of the value to match" );
```
gsmet: Good catch.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/logging/impl/Log.java
yrodiere: Not sure that' what you meant, but I added the exact version to the log message (7.y.z).

yrodiere:
Merged, thanks!

PR #1763, opened 2018-10-05
HSEARCH-3310 Split the @Field annotation into multiple, type-specific annotations

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3310

~Based on #1762 , which should be merged first.~ => Done


mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/FullTextField.java
```diff
@@ -0,0 +1,80 @@
...
+ * <p>
+ * Note that this annotation only creates tokenized (multi-word) text fields.
+ * As a result:
+ * <ul>
+ *     <li>You must assign an analyzer when using this annotation</li>
+ *     <li>This annotation does not allow to make the field sortable (analyzed fields cannot be sorted on)</li>
+ * </ul>
+ * <p>
+ * If you don't want the field to be analyzed, or want it to be sortable,
+ * use the {@link GenericField} annotation instead.
```
gsmet: Wouldn't we recommend `KeywordField`?

yrodiere: Right. I clarified the documentation to also mention we expect the field to be of type String, and that KeywordField should be used in this case if the field is a String.

yrodiere:
Still waiting for CI. Will merge on Monday.

yrodiere:
Merged, thanks!

PR #1764, opened 2018-10-05
HSEARCH-3389 Fix bugs in the Lucene backend when maxResults is set to 0

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-3389


integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/SearchQueryIT.java
```diff
@@ -156,6 +156,22 @@ public void paging_reuse_query() {
 				.hasReferencesHitsExactOrder( INDEX_NAME, DOCUMENT_1, DOCUMENT_2, DOCUMENT_3 );
 	}
 
+	@Test
```
yrodiere: Add a `@TestForIssue` here, maybe? Otherwise it looks good to merge.

PR #1767, opened 2018-10-09
HSEARCH-3189 + HSEARCH-3225: small API improvements

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3189
https://hibernate.atlassian.net//browse/HSEARCH-3225

* HSEARCH-3189: Instantiate immutable implementations of data-carrier interfaces through static factory methods only
* HSEARCH-3225: Remove search() methods that implicitly create a search target encompassing all mapped types


yrodiere:
Merged, thanks

PR #1768, opened 2018-10-11
HSEARCH-3380 Find a way to add JAXB modules to the OSGi feature in OSGi integration tests

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3380

This is just a hack, but hopefully we'll be able to remove it, along with all the explicit dependencies to JAXB, when Hibernate ORM is fixed and we fix [HSEARCH-3376](https://hibernate.atlassian.net//browse/HSEARCH-3376)


gsmet:
I suppose you tested this dark magic so let's merge it :).

yrodiere:
> I suppose you tested this dark magic so let's merge it :).

I did, quite a few times...

Rebased and merged. Thanks!

PR #1769, opened 2018-10-11
HSEARCH-3396 Allow to use snapshot artifacts in OSGi integration tests

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3396

PR #1770, opened 2018-10-11
Backport HSEARCH-3396 to branch 5.10 - Allow to use snapshot artifacts in OSGi integration tests

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3396

This is based on #1769 , which should be merged first.


yrodiere:
Merged

PR #1771, opened 2018-10-11
HSEARCH-1710 Child collection being unnecessarily initialized through @ContainedIn

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-1710

Note I added a separate package dedicated to non-regression tests, so that we can differentiate between those and actual, more complete functional tests. Hopefully that will make it easier to find the most relevant tests when we need to add a functional test: in most cases we don't want to change a non-regression tests, but we sometimes want to add tests to a functional test class.

I'm open to revisit this if you think it's unnecessary.


gsmet:
Apart from the discussion we had on HipChat about the leading `_` for the table name, it looks good.

yrodiere:
Fixed the table name and merged. Thanks.

PR #1772, opened 2018-10-15
HSEARCH-3126 Search 6 groundwork - Rework the syntax when using extensions

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3126

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/logging/impl/Log.java
+++ b/engine/src/main/java/org/hibernate/search/engine/logging/impl/Log.java
@@ -9,11 +9,11 @@ package org.hibernate.search.engine.logging.impl;
 
 import java.util.List;
 
-import org.hibernate.search.engine.mapper.model.spi.MappableTypeModel;
-import org.hibernate.search.util.EventContext;
 import org.hibernate.search.engine.logging.spi.MappableTypeModelFormatter;
-import org.hibernate.search.util.SearchException;
+import org.hibernate.search.engine.mapper.model.spi.MappableTypeModel;
 import org.hibernate.search.engine.spatial.GeoPoint;
+import org.hibernate.search.util.EventContext;
+import org.hibernate.search.util.SearchException;
 import org.hibernate.search.util.impl.common.MessageConstants;
 import org.hibernate.search.util.impl.common.logging.ClassFormatter;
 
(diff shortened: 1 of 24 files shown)


engine/src/main/java/org/hibernate/search/engine/logging/impl/Log.java
```diff
@@ -166,4 +166,17 @@ SearchException relativeFieldNameCannotContainDot(String relativeFieldName,
 			value = "Multiple calls to createAccessor() for the same field definition."
 					+ " You must call createAccessor() exactly once.")
 	SearchException cannotCreateAccessorMultipleTimes(@Param EventContext context);
+
+	@Message(id = ID_OFFSET_2 + 25,
+			value = "Cannot call ifSupported(...) after otherwise(...)."
```
gsmet: I think you mean `orElse()` here instead of `otherwise()`. At least from a user facing POV, that's what he will see.

Maybe we should be consistent to avoid this issue?

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/ExtensionIT.java
```diff
@@ -99,19 +99,20 @@ public void predicate_fromJsonString() {
 		SearchQuery<DocumentReference> query = searchTarget.query( sessionContext )
 				.asReferences()
 				.predicate().bool( b -> {
-					b.should().withExtensionOptional(
+					b.should().extension().ifSupported(
 							LuceneExtension.get(),
 							// FIXME find some way to forbid using the context twice... ?
 							c -> c.fromLuceneQuery( new TermQuery( new Term( "string", "text 1" ) ) )
 					);
```
gsmet: Is it valid to call `ifSupported()` without having a `orElse()` or `orElseFail()`?

engine/src/main/java/org/hibernate/search/engine/logging/impl/Log.java
yrodiere: Right, I forgot these when I renamed `otherwise()` to `orElse()` while working on it. Which I did to be consistent with `Optional#orElse`.

Fixed.

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/ExtensionIT.java
yrodiere: Right... so, two problems:

1. I forgot to change this test to use the simpler `extension( MyExtension.get() ).fromLuceneQuery( ... )` syntax like I did in the Elasticsearch test. Fixed.
2. I forgot to test the very case you're mentioning, which should ideally trigger an error at some point. I'll see what I can do on Monday.

yrodiere: > which should ideally trigger an error at some point

I mean, it probably should if the extension was not supported.

yrodiere: It was (much) easier to allow the syntax `ifSupported()` without a following `orElse`, so I added some tests and allowed it.

yrodiere:
Merged. Thanks!

PR #1777, opened 2018-10-23
HSEARCH-3079 Throw an error when mixing different ORM SessionFactories

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3079

### DIFF
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/spi/DifferentSessionFactoriesIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/spi/DifferentSessionFactoriesIT.java
@@ -1,0 +1,96 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.integrationtest.mapper.orm.spi;
+
+import javax.persistence.Entity;
+import javax.persistence.EntityManager;
+import javax.persistence.Id;
+
+import org.hibernate.SessionFactory;
+import org.hibernate.engine.spi.SessionFactoryImplementor;
+import org.hibernate.search.mapper.orm.impl.HibernateSearchContextService;
+import org.hibernate.search.mapper.orm.mapping.HibernateOrmMapping;
+import org.hibernate.search.mapper.orm.mapping.HibernateOrmSearchManager;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.Indexed;
+import org.hibernate.search.util.SearchException;
+import org.hibernate.search.util.impl.integrationtest.common.rule.BackendMock;
+import org.hibernate.search.util.impl.integrationtest.orm.OrmSetupHelper;
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+/**
+ * Tests the creation of a {@link HibernateOrmSearchManager}, using a {@link HibernateOrmMapping} with an {@link EntityManager} owned by a different {@link SessionFactory}.
+ *
+ * @author Fabio Massimo Ercoli
+ */
+public class DifferentSessionFactoriesIT {
+
+	@Rule
+	public BackendMock backendMock = new BackendMock( "stubBackend" );
+
+	@Rule
+	public OrmSetupHelper ormSetupHelper = new OrmSetupHelper();
+
+	@Rule
+	public ExpectedException thrown = ExpectedException.none();
+
+	private SessionFactory sessionFactory;
+	private SessionFactory sessionFactoryAlt;
+	private EntityManager entityManager;
+
+	@Before
+	public void setup() {
+		sessionFactory = initABasicSessionFactory();
+		sessionFactoryAlt = initABasicSessionFactory();
+		entityManager = sessionFactory.createEntityManager();
+	}
+
+	@Test
+	public void tryToUseDifferentSessionFactories() {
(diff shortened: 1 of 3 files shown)


mapper/orm/src/main/java/org/hibernate/search/mapper/orm/mapping/impl/HibernateOrmMappingImpl.java
```diff
@@ -49,7 +54,12 @@ public boolean isWorkable(Object entity) {
 
 	private HibernateOrmSearchManagerBuilder createSearchManagerBuilder(EntityManager entityManager) {
 		SessionImplementor sessionImplementor = entityManager.unwrap( SessionImplementor.class );
-		// TODO check that the session refers to the same session factory used when building the mapping
+		SessionFactoryImplementor givenSessionFactory = sessionImplementor.getSessionFactory();
+
+		if ( !givenSessionFactory.getUuid().equals( sessionFactoryImplementor.getUuid() ) ) {
```
yrodiere: I think you really should use `.equals` instead. If two different instances of `SessionFactoryImplementor` have the same UUID, something is terribly wrong, and I'd rather have this code fail...

fax4ever: No problem to change that. Just for curiosity, is it really possible? :)

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/spi/DifferentSessionFactoriesIT.java
```diff
@@ -0,0 +1,95 @@
...
+	private SessionFactory initABasicSessionFactory() {
+		backendMock.expectSchema( AutomaticIndexingBasicIT.IndexedEntity.INDEX, b -> b
+				.field( "indexedField", String.class )
+				.field( "noReindexOnUpdateField", String.class )
+				.field( "indexedElementCollectionField", String.class )
+				.field( "noReindexOnUpdateElementCollectionField", String.class )
+		);
+
+		SessionFactory factory = ormSetupHelper.withBackendMock( backendMock )
+				.setup( AutomaticIndexingBasicIT.IndexedEntity.class );
```
yrodiere: Do not reuse an entity from another test, create your own inside the test class. It can be as simple as this:

```
	@Entity(name = "indexed")
	@Indexed(name = IndexedEntity.INDEX)
	public static class IndexedEntity {
		public static String INDEX = "IndexedEntity";

		@Id
		Integer id;

		@Id
		public Integer getId() {
			return id;
		}

		public void setId(Integer id) {
			this.id = id;
		}
	}
```

... but please, don't reuse entity classes across tests. I don't want changing one class in one test to affect other tests...

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/spi/DifferentSessionFactoriesIT.java
```diff
@@ -0,0 +1,95 @@
...
+		// service mapping is taken from the alternative session factory
+		HibernateOrmMapping serviceMapping = sessionFactoryAlt.unwrap( SessionFactoryImplementor.class )
+			.getServiceRegistry().getService( HibernateSearchContextService.class ).getMapping();
+
+		// try to use an entityManager owned by the original session factory instead
+		serviceMapping.createSearchManager( entityManager );
+	}
+
+	@After
+	public void teardown() {
```
yrodiere: This is not necessary. The `ormSetupHelper` rule will take care of this. You can remove this method.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/spi/DifferentSessionFactoriesIT.java
```diff
@@ -0,0 +1,95 @@
...
+		if ( sessionFactory != null ) {
+			sessionFactory.close();
+		}
+		if ( sessionFactoryAlt != null ) {
+			sessionFactoryAlt.close();
+		}
+	}
+
+	private SessionFactory initABasicSessionFactory() {
+		backendMock.expectSchema( AutomaticIndexingBasicIT.IndexedEntity.INDEX, b -> b
```
yrodiere: You can use `backendMock.expectAnySchema( IndexedEntity.INDEX )` here, since you don't really care about the content of the schema.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/spi/DifferentSessionFactoriesIT.java
fax4ever: Nice! We should introduce something similar on Hibernate OGM

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/spi/DifferentSessionFactoriesIT.java
fax4ever: Nice to weak the expectation here. Thanks

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/spi/DifferentSessionFactoriesIT.java
fax4ever: I agree with you, I also prefer decoupling over reuse.

fax4ever:
Wow tests are executed very fast here. I made some changes, if they are OK, I could squash the commits.

yrodiere:
Looks good, squashed the two test commits and merged. Thanks!

fax4ever:
Thanks!

PR #1778, opened 2018-10-24
 HSEARCH-3399 Use "builder factories" consistently for DSL builders in backends

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-3399


gsmet:
@yrodiere I think this is good to go.

I'm wondering if we should create a test for all the `isDslCompatible()` methods (e.g. for every supported type)?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/sort/impl/GeoPointFieldSortBuilder.java
```diff
@@ -4,19 +4,22 @@
...
+import org.hibernate.search.backend.lucene.search.sort.impl.LuceneSearchSortBuilder;
+import org.hibernate.search.backend.lucene.search.sort.impl.LuceneSearchSortCollector;
 import org.hibernate.search.engine.logging.spi.EventContexts;
 import org.hibernate.search.engine.search.dsl.sort.SortOrder;
 import org.hibernate.search.engine.search.sort.spi.DistanceSortBuilder;
 import org.hibernate.search.engine.spatial.GeoPoint;
 import org.hibernate.search.util.impl.common.LoggerFactory;
 
-class DistanceSortBuilderImpl extends AbstractSearchSortBuilder
+public class GeoPointFieldSortBuilder extends AbstractSearchSortBuilder
```
yrodiere: Not sure this new name is a great one; first because this class implements `DistanceSortBuilder`, not `FieldSortBuilder`, second because it gives the false impression that this will sort by the GeoPoint's value somehow, and third because it's inconsistent with how the distance projection builders are named.

Perhaps what you're after is making the various `GeoPoint`-related classes sort nicely in the package, in which case I would suggest `GeoPointDistanceSortBuilder`?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/sort/impl/LocalDateFieldSortBuilderFactory.java
```diff
@@ -0,0 +1,57 @@
...
+		this.converter = converter;
+	}
+
+	@Override
+	public FieldSortBuilder<LuceneSearchSortBuilder> createFieldSortBuilder(String absoluteFieldPath) {
+		return new LocalDateFieldSortBuilder( absoluteFieldPath, converter );
+	}
+
+	@Override
+	public DistanceSortBuilder<LuceneSearchSortBuilder> createDistanceSortBuilder(String absoluteFieldPath,
```
yrodiere: You could have an `AbstractStandardLuceneFieldSortBuilderFactory`, similarly to the `AbstractStandardLuceneFieldPredicateBuilderFactory`, that would implement the the `converter` field, and the `createDistanceSortBuilder` and `isDslCompatibleWith` methods. Not a big improvement, but that would remove a little bit of duplicated code.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/projection/impl/GeoPointFieldProjectionBuilderFactory.java
```diff
@@ -37,7 +37,7 @@ public boolean isDslCompatibleWith(ElasticsearchFieldProjectionBuilderFactory ob
 		if ( this == obj ) {
 			return true;
 		}
-		if ( !( obj instanceof GeoPointFieldProjectionBuilderFactory ) ) {
```
yrodiere: You're right, it's better. While you're at it, maybe you could fix `GeoPointFieldPredicateBuilderFactory#isDslCompatibleWith` (the Lucene one): it just compares the classes, but I find your pattern of comparing `INSTANCE == obj` more foolproof (if we remove the `INSTANCE` constant one day, we'll have to have a look at the `isDslCompatibleWith` method).

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/projection/impl/FieldSearchProjectionBuilderImpl.java
```diff
@@ -11,7 +11,7 @@
 import org.hibernate.search.engine.search.SearchProjection;
 import org.hibernate.search.engine.search.projection.spi.FieldSearchProjectionBuilder;
 
-public class FieldSearchProjectionBuilderImpl<T> implements FieldSearchProjectionBuilder<T> {
+public class FieldSearchProjectionBuilderImpl<T, U> implements FieldSearchProjectionBuilder<U> {
```
yrodiere: Shouldn't `T` be renamed to `F`? From what I understand it's the actual field type, which is named `F` everywhere else, including in the codec.

If I understand correctly, `U` is not the same as the type `T` in the converter, though (it the type requested by the user, not the type used internally when querying).

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/projection/impl/FieldSearchProjectionImpl.java
```diff
@@ -14,7 +14,7 @@
 import org.hibernate.search.backend.lucene.types.converter.impl.LuceneFieldConverter;
 import org.hibernate.search.engine.search.query.spi.ProjectionHitCollector;
 
-class FieldSearchProjectionImpl<T> implements LuceneSearchProjection<T> {
+class FieldSearchProjectionImpl<T, U> implements LuceneSearchProjection<U> {
```
yrodiere: Same here, I think `T` should be named `F` instead.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/impl/IndexSchemaFieldNodeComponentRetrievalStrategy.java
```diff
@@ -0,0 +1,22 @@
...
+import org.hibernate.search.util.SearchException;
+
+public interface IndexSchemaFieldNodeComponentRetrievalStrategy<T> {
+
+	T extractComponent(ElasticsearchIndexSchemaFieldNode<?> schemaNode);
+
+	boolean areCompatible(T component1, T component2);
+
+	SearchException createCompatibilityException(String absoluteFieldPath,
+			ElasticsearchIndexSchemaFieldNode<?> schemaNode1, ElasticsearchIndexSchemaFieldNode<?> schemaNode2,
```
yrodiere: Given that you're already passing the absolute field path to this method, wouldn't it make more sense to only pass the "components" (of type `T`) instead of the schema nodes here? So that we display the actual components that are not compatible in the error message? That would remove a bit of noise.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/projection/impl/GeoPointFieldProjectionBuilderFactory.java
gsmet: > (if we remove the INSTANCE constant one day, we'll have to have a look at the isDslCompatibleWith method)

That's exactly why I did it that way :).

Fixed `GeoPointFieldPredicateBuilderFactory`.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/projection/impl/FieldSearchProjectionBuilderImpl.java
gsmet: Right, done.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/projection/impl/FieldSearchProjectionImpl.java
gsmet: Fixed.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/impl/IndexSchemaFieldNodeComponentRetrievalStrategy.java
gsmet: Right, done.

PR #1779, opened 2018-10-25
HSEARCH-3296 Search 6 groundwork - Pass context objects to bridges

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3296

This is not an easy topic, so I expect we will have to discuss it and make some adjustments before we merge the PR.

The first thing to notice is that not all bridge operations are executed in the same "scope":

* the "write" operations in type bridges/property bridges, for example, are executed in a "session" scope: there's always an open ORM session when we execute these methods.
* other operations, such as `org.hibernate.search.mapper.pojo.bridge.ValueBridge#toIndexedValue`, could theoretically be executed outside the "session" scope, in what I would call a "mapping" scope. Even if in practice we turned out to always have an open session when we execute these operations, they are used to generate `SearchPredicate` and `SearchSort` instances, which are supposedly re-useable; thus the operations should return the same result regardless of the session, thus the session should not matter for these operations.

Keeping that in mind, we could decide to have one context type per "scope": essentially one type for the session scope and one for the mapping scope. Each operation where we want to pass a context would accept one or the other: for example `RoutingKeyBridge.toRoutingKey(String, Object, PojoElement)` would become `RoutingKeyBridge.toRoutingKey(String, Object, PojoElement, SessionContext)`, `PropertyBridge.write(DocumentElement, PojoElement)` would become `PropertyBridge.write(DocumentElement, PojoElement, SessionContext)`, and `ValueBridge.toIndexedValue(Object)` would become `ValueBridge.toIndexedValue(Object, MappingContext)`.

That's all well and good, but let's remember why we wanted to pass context objects to bridges in the first place: we wanted to be able to evolve the APIs without breaking them, which the context objects would allow us to do: they would let us introduce new getters or operations that new bridge could use, but that wouldn't break compatibility with older bridges.

With those two types, we're fine as long as the new getters or operations we want to introduce are related to the scope (some session-related or mapping-related operation), but as soon as we want to be more specific (some operation that only makes sense for a `RoutingKeyBridge`, for example), we're kind of stuck.

To address this problem, the second approach that I tried to explore is to introduce one "context" interface per bridge operation (commit with message "HSEARCH-3296 Use one specific context interface per bridge/converter" and following commits). For example, with this approach,  `RoutingKeyBridge.toRoutingKey(String, Object, PojoElement)` would become `RoutingKeyBridge.toRoutingKey(String, Object, PojoElement, RoutingKeyBridgeToRoutingKeyContext)`, `PropertyBridge.write(DocumentElement, PojoElement)` would become `PropertyBridge.write(DocumentElement, PojoElement, PropertyBridgeWriteContext)`, and `ValueBridge.toIndexedValue(Object)` would become `ValueBridge.toIndexedValue(Object, ValueBridgeToIndexedValueContext)`.

The second approach definitely makes the APIs more future-proof, but it comes at the cost of clarity: we end up with a rather "exploded" API, where you have trouble finding common denominators between all the bridge types. On the other hand, using tricks such as the "extension" pattern we're already using in several places, we can make it easy for users to get back on their feet: we can totally declare a single `OrmExtension` for example, and make sure that calling `RoutingKeyBridgeToRoutingKeyContext.extension(OrmExtension.get())` returns the same type as calling `PropertyBridgeWriteContext.extension(OrmExtension.get())`. Then, for the most simple needs, the fact that each operation takes a different context as a parameter would not matter, since you could use the context similarly (with the same extension, and the same result). I tried to implement such an extension in the ORM case, that allows to access the session/sessionFactory from within a bridge. See `org.hibernate.search.mapper.orm.OrmExtension`, which returns a `HibernateOrmSessionContext` for operations in the "session" scope and a `HibernateOrmMappingContext` for operations in the "mapping" scope.

My choice would personally be to go with the second approach, but YMMV. Have a look at the code and make your own opinion :)


engine/src/main/java/org/hibernate/search/engine/mapper/session/context/SessionContext.java
```diff
@@ -0,0 +1,28 @@
...
+ * Depending on the mapper being used, this may give access, through {@link #unwrap(Class)},
+ * to different objects, such as a Hibernate ORM Session.
+ */
+public interface SessionContext {
+
+	/**
+	 * Unwrap the session context to some implementation-specific type.
+	 *
+	 * @param clazz The {@link Class} representing the expected type
+	 * @param <T> The expected type
```
gsmet: Very minor but dots are missing for the 2 above.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/session/context/impl/HibernateOrmSessionContextImpl.java
```diff
@@ -6,12 +6,40 @@
...
+import org.hibernate.search.mapper.pojo.model.spi.PojoRuntimeIntrospector;
+import org.hibernate.search.util.impl.common.CollectionHelper;
+import org.hibernate.search.util.impl.common.LoggerFactory;
+
+public class HibernateOrmSessionContextImpl implements PojoSessionContextImplementor, HibernateOrmSessionContext {
+
+	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
+
+	/**
+	 * SessionImplementor.unwrap is a bit dodgy: if you define your own interface that extends Session,
```
gsmet: `unwrap()` maybe as it's a method?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/session/context/impl/HibernateOrmSessionContextImpl.java
```diff
@@ -21,6 +49,27 @@ public HibernateOrmSessionContextImpl(SessionImplementor sessionImplementor) {
...
+		if ( clazz.isAssignableFrom( HibernateOrmSessionContext.class ) ) {
+			return (T) this;
+		}
+		/*
+		 * As a shortcut, we allow to unwrap directly to the Hibernate ORM session instead of
+		 * having to write .unwrap( HibernateOrmSessionContext.class ).getSession().
+		 */
+		else if ( SUPPORTED_SESSION_IMPLEMENTOR_UNWRAP_CLASSES.contains( clazz ) ) {
+			return (T) sessionImplementor;
+		}
```
gsmet: Not a big fan of it, to be honest.

It saves a method call but I don't think the additional code/complexity/potential lack of clarity is worth it.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/bridge/PropertyBridge.java
```diff
@@ -52,11 +53,12 @@
 	 * Reading from the {@link PojoElement} should be done using
 	 * {@link org.hibernate.search.mapper.pojo.model.PojoModelElementAccessor}s retrieved when the
 	 * {@link #bind(PropertyBridgeBindingContext)} method was called.
-	 *
-	 * @param target The {@link DocumentElement} to write to.
+	 *  @param target The {@link DocumentElement} to write to.
```
gsmet: There's an additional space before `@param`.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/bridge/PropertyBridge.java
```diff
@@ -52,11 +53,12 @@
...
 	 * @param source The {@link PojoElement} to read from.
+	 * @param sessionContext A {@link SessionContext} that can be {@link SessionContext#unwrap(Class) unwrapped}
+	 * to a more useful type, such as a Hibernate ORM Session (if using the Hibernate ORM mapper).
```
gsmet: If we take into account my comment about session unwrapping, we will need to update these comments.

backend/elasticsearch/src/test/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/DistanceToFieldSearchProjectionTest.java
```diff
@@ -30,10 +32,14 @@ public void projection_script() {
...
-		SearchProjectionExecutionContext searchProjectionExecutionContext =
-				elementCollector.toSearchProjectionExecutionContext();
+		SessionContextImplementor sessionContext = createMock( SessionContextImplementor.class );
 
+		resetAll();
+		replayAll();
+		SearchProjectionExecutionContext searchProjectionExecutionContext =
+				elementCollector.toSearchProjectionExecutionContext( sessionContext );
 		projection.contributeRequest( requestBody, searchProjectionExecutionContext );
+		verifyAll();
```
gsmet: Why do we need all this noise? Doesn't look like an improvement for the readability of the test?

Wouldn't a stub `SessionContext` be better?

mapper/orm/src/test/java/org/hibernate/search/mapper/orm/OrmExtensionTest.java
```diff
@@ -0,0 +1,73 @@
...
+public class OrmExtensionTest extends EasyMockSupport {
+	private final SessionImplementor sessionImplementor = createMock( SessionImplementor.class );
+	private final HibernateOrmSessionContextImpl sessionContext = new HibernateOrmSessionContextImpl( sessionImplementor );
+
+	@Test
+	public void identifierBridge() {
+		IdentifierBridgeFromDocumentIdentifierContext fromDocumentContext = new BridgeSessionContextImpl( sessionContext );
+		resetAll();
+		replayAll();
+		assertThat( fromDocumentContext.extension( OrmExtension.get() ) ).isSameAs( sessionContext );
```
gsmet: It means the user will have to do that to get the ORM session?

Same remark about the EasyMock machinery being harmful here.

gsmet:
I added a couple of comments inline.

A couple are outdated if we go the multiple contexts path as you have removed the code in these commits.

While I agree separate contexts are probably better, I can't say I'm very excited about the extension stuff being required after these changes.

It's probably more future-proof though so I suppose it's for the better.

backend/elasticsearch/src/test/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/DistanceToFieldSearchProjectionTest.java
yrodiere: This "noise" is necessary for EasyMock mocks to be in the correct state when they are used. Without it, the mock would be in "record" mode during the test, which is not the way mocks are supposed to be used. As an additional benefit, putting the mock in the correct state allows us to check that no method is called on the session context, in other words that `toSearchProjectionExecutionContext` does not rely on the session context. Not really necessary, but as I said it's not the main reason.

As to whether coding empty interfaces yourself is better than using a mocking library, well... I personally don't want to have to change this test whenever we add a method to `SessionContextImplementor`, but YMMV.

mapper/orm/src/test/java/org/hibernate/search/mapper/orm/OrmExtensionTest.java
yrodiere: The user will have to call `fromDocumentContext.extension( OrmExtension.get() )`, yes.

About EasyMock, if you don't want to use it and want to implement `SessionImplementor` instead, be me guest, but I doubt you do ;)

yrodiere:
Fixed the typos and merged. Thanks!

PR #1780, opened 2018-10-26
HSEARCH-3082 Search 6 groundwork - Create all SearchExceptions using a JBoss Logger

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3082

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/dsl/impl/ElasticsearchIndexSchemaFieldContextImpl.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/dsl/impl/ElasticsearchIndexSchemaFieldContextImpl.java
@@ -6,8 +6,10 @@
  */
 package org.hibernate.search.backend.elasticsearch.document.model.dsl.impl;
 
+import java.lang.invoke.MethodHandles;
 import java.time.LocalDate;
 
+import org.hibernate.search.backend.elasticsearch.logging.impl.Log;
 import org.hibernate.search.engine.backend.document.model.dsl.IndexSchemaFieldTypedContext;
 import org.hibernate.search.engine.backend.document.model.dsl.StandardIndexSchemaFieldTypedContext;
 import org.hibernate.search.engine.backend.document.model.dsl.StringIndexSchemaFieldTypedContext;
@@ -26,7 +28,7 @@ import org.hibernate.search.backend.elasticsearch.util.impl.ElasticsearchFields;
 import org.hibernate.search.engine.logging.spi.EventContexts;
 import org.hibernate.search.engine.spatial.GeoPoint;
 import org.hibernate.search.util.EventContext;
-import org.hibernate.search.util.SearchException;
+import org.hibernate.search.util.impl.common.LoggerFactory;
 
 
 /**
@@ -35,6 +37,8 @@ import org.hibernate.search.util.SearchException;
 class ElasticsearchIndexSchemaFieldContextImpl
 		implements ElasticsearchIndexSchemaFieldContext, ElasticsearchIndexSchemaNodeContributor<PropertyMapping>,
 				IndexSchemaContext {
+
+	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
 
 	private final AbstractElasticsearchIndexSchemaObjectNodeBuilder parent;
 	private final String relativeFieldName;
@@ -65,7 +69,7 @@ class ElasticsearchIndexSchemaFieldContextImpl
 		}
 		else {
 			// TODO implement other types
-			throw new SearchException( "Cannot guess field type for input type " + inputType );
+			throw log.cannotGuessFieldType( inputType, getEventContext() );
 		}
 	}
 
@@ -109,7 +113,7 @@ class ElasticsearchIndexSchemaFieldContextImpl
 
 	private <T extends ElasticsearchIndexSchemaNodeContributor<PropertyMapping>> T setDelegate(T context) {
 		if ( delegate != null ) {
-			throw new SearchException( "You cannot set the type of a field more than once" );
+			throw log.tryToSetFieldTypeMoreThanOnce( getEventContext() );
 		}
 		delegate = context;
 		return context;
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
@@ -312,4 +312,13 @@ public interface Log extends BasicLogger {
 	SearchException conflictingFieldTypesForProjection(String absoluteFieldPath,
 			ElasticsearchFieldProjectionBuilderFactory component1, ElasticsearchFieldProjectionBuilderFactory component2,
 			@Param EventContext context);
+
+	@Message(id = ID_OFFSET_3 + 44, value = "Failed to shut down the Elasticsearch backend.")
+	SearchException failedToShutdownBackend(@Cause Exception cause, @Param EventContext context);
+
+	@Message(id = ID_OFFSET_3 + 45, value = "Cannot guess field type for input type %1$s.")
+	SearchException cannotGuessFieldType(@FormatWith(ClassFormatter.class) Class<?> inputType, @Param EventContext context);
+
+	@Message(id = ID_OFFSET_3 + 46, value = "You cannot set the type of a field more than once.")
+	SearchException tryToSetFieldTypeMoreThanOnce(@Param EventContext context);
 }
--- a/mapper/javabean/src/main/java/org/hibernate/search/mapper/javabean/log/impl/Log.java
+++ b/mapper/javabean/src/main/java/org/hibernate/search/mapper/javabean/log/impl/Log.java
@@ -11,8 +11,10 @@ import org.hibernate.search.mapper.pojo.logging.spi.PojoTypeModelFormatter;
 import org.hibernate.search.mapper.pojo.model.spi.PojoRawTypeModel;
 import org.hibernate.search.util.SearchException;
 import org.hibernate.search.util.impl.common.MessageConstants;
+import org.hibernate.search.util.impl.common.logging.ClassFormatter;
 
 import org.jboss.logging.BasicLogger;
+import org.jboss.logging.annotations.Cause;
 import org.jboss.logging.annotations.FormatWith;
 import org.jboss.logging.annotations.Message;
 import org.jboss.logging.annotations.MessageLogger;
@@ -34,4 +36,9 @@ public interface Log extends BasicLogger {
 	SearchException cannotReadProperty(@FormatWith(PojoTypeModelFormatter.class) PojoRawTypeModel<?> typeModel,
 			String propertyName);
 
+	@Message(id = ID_OFFSET_1 + 3, value = "Exception while retrieving the type model for '%1$s'.")
+	SearchException errorRetrievingTypeModel(@FormatWith(ClassFormatter.class) Class<?> clazz, @Cause Exception cause);
+
+	@Message(id = ID_OFFSET_1 + 4, value = "Exception while retrieving property type model for '%1$s' on '%2$s'")
+	SearchException errorRetrievingPropertyTypeModel(String propertyModelName, @FormatWith(PojoTypeModelFormatter.class) PojoRawTypeModel<?> parentTypeModel, @Cause Exception cause);
 }
--- a/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/logging/impl/Log.java
+++ b/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/logging/impl/Log.java
@@ -105,4 +105,7 @@ public interface Log extends BasicLogger {
 
 	@Message(id = ID_OFFSET_2 + 11, value = "Mapping service cannot create a search manager using a different session factory. Expected: '%1$s'. In use: '%2$s'.")
 	SearchException usingDifferentSessionFactories(SessionFactory expectedSessionFactory, SessionFactory usedSessionFactory);
+
+	@Message(id = ID_OFFSET_2 + 12, value = "Exception while retrieving property type model for '%1$s' on '%2$s'.")
+	SearchException errorRetrievingPropertyTypeModel(String propertyModelName, @FormatWith(PojoTypeModelFormatter.class) PojoRawTypeModel<?> parentTypeModel, @Cause Exception cause);
 }
--- a/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/model/impl/HibernateOrmPropertyModel.java
+++ b/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/model/impl/HibernateOrmPropertyModel.java
@@ -7,6 +7,7 @@
 package org.hibernate.search.mapper.orm.model.impl;
 
 import java.lang.annotation.Annotation;
+import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Field;
 import java.lang.reflect.Member;
 import java.lang.reflect.Method;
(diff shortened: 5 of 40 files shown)


fax4ever:
Just a preview, only creation using constructor SearchException(m, c) is currently covered. Other changes will be on subsequent commits.

mapper/javabean/src/main/java/org/hibernate/search/mapper/javabean/log/impl/Log.java
```diff
@@ -34,4 +36,11 @@ SearchException cannotFindProperty(@FormatWith(PojoTypeModelFormatter.class) Poj
 	SearchException cannotReadProperty(@FormatWith(PojoTypeModelFormatter.class) PojoRawTypeModel<?> typeModel,
 			String propertyName);
 
+	@Message(id = ID_OFFSET_1 + 3,
+			value = "Exception while retrieving the type model for '%1$s'.")
+	SearchException errorRetrivingTypeModel(@FormatWith(ClassFormatter.class) Class<?> clazz, @Cause Exception cause);
+
+	@Message(id = ID_OFFSET_1 + 4,
+			value = "Exception while retrieving property type model for '%1$s' on '%2$s'")
+	SearchException errorRetrivingPropertyTypeModel(String propertyModelName, String parentTypeModel, @Cause Exception cause);
```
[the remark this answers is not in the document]
yrodiere: Also, you could pass the type model directly instead of using toString(). Just use a formatter, like on `org.hibernate.search.mapper.pojo.logging.impl.Log#unableToResolveDefaultIdentifierBridgeFromSourceType` for example.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/logging/impl/Log.java
```diff
@@ -105,4 +105,7 @@ SearchException cannotFindReadableProperty(@FormatWith(PojoTypeModelFormatter.cl
 
 	@Message(id = ID_OFFSET_2 + 11, value = "Mapping service cannot create a search manager using a different session factory. Expected: '%1$s'. In use: '%2$s'.")
 	SearchException usingDifferentSessionFactories(SessionFactory expectedSessionFactory, SessionFactory usedSessionFactory);
+
+	@Message(id = ID_OFFSET_2 + 12, value = "Exception while retrieving property type model for '%1$s' on '%2$s'")
+	SearchException errorRetrivingPropertyTypeModel(String propertyModelName, String parentTypeModel, @Cause Exception cause);
```
yrodiere: `errorRetrivingPropertyTypeModel` => `errorRetrivingPropertyTypeModel`

yrodiere: Also, you could pass the type model directly instead of using toString(). Just use a formatter, like on `org.hibernate.search.mapper.pojo.logging.impl.Log#unableToResolveDefaultIdentifierBridgeFromSourceType` for example.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/logging/impl/Log.java
```diff
@@ -280,4 +280,9 @@ SearchException invalidFieldEncodingForFullTextFieldMapping(IndexSchemaFieldType
 	)
 	SearchException invalidFieldEncodingForKeywordFieldMapping(IndexSchemaFieldTypedContext<?, ?> context,
 			@FormatWith(ClassFormatter.class) Class<?> expectedContextType);
+
+	@Message(id = ID_OFFSET_2 + 33,
+			value = "Exception while invoking '%1$s' on '%2$s'."
+	)
+	SearchException errorInvokingMember(String member, String component);
```
yrodiere: You forgot to pass the `@Cause`.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -312,4 +312,8 @@ SearchException conflictingFieldTypesForSort(String absoluteFieldPath,
 	SearchException conflictingFieldTypesForProjection(String absoluteFieldPath,
 			ElasticsearchFieldProjectionBuilderFactory component1, ElasticsearchFieldProjectionBuilderFactory component2,
 			@Param EventContext context);
+
+	@Message(id = ID_OFFSET_3 + 44,
+			value = "Failed to shut down the Elasticsearch backend.")
+	SearchException failedToShutdownBackend(@Cause Exception cause);
```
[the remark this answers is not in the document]
fax4ever: nice!

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/model/impl/AbstractPojoModelCompositeElement.java
```diff
@@ -44,7 +48,7 @@
 	@SuppressWarnings("unchecked") // The cast is checked using reflection
 	public final <T> PojoModelElementAccessor<T> createAccessor(Class<T> requestedType) {
 		if ( !isAssignableTo( requestedType ) ) {
-			throw new SearchException( "Requested incompatible type for '" + createAccessor() + "': '" + requestedType + "'" );
+			log.incompatibleRequestedType( createAccessor(), requestedType );
```
yrodiere: Could you please improve this while you're at it; I would expect to pass a path here, not the accessor itself.

Maybe do something like this:

```
log.incompatibleRequestedType( getModelPathTypeNode().toUnboundPath(), requestedType );
```

... and use `PojoModelPathFormatter` to format the path.

fax4ever:
I think I've finished the refactoring here. Let me know whether I have to change something else. Thanks

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/dsl/impl/ElasticsearchIndexSchemaFieldContextImpl.java
```diff
@@ -109,7 +112,7 @@ public EventContext getEventContext() {
 
 	private <T extends ElasticsearchIndexSchemaNodeContributor<PropertyMapping>> T setDelegate(T context) {
 		if ( delegate != null ) {
-			throw new SearchException( "You cannot set the type of a field more than once" );
+			throw log.tryToSetFieldTypeMoreThanOnce( getEventContext() );
```
yrodiere: Good idea, some context here and in the exception above could help.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/dsl/impl/ElasticsearchIndexSchemaFieldContextImpl.java
```diff
@@ -64,8 +68,7 @@ else if ( GeoPoint.class.equals( inputType ) ) {
 			return (StandardIndexSchemaFieldTypedContext<?, F>) asGeoPoint();
 		}
 		else {
-			// TODO implement other types
-			throw new SearchException( "Cannot guess field type for input type " + inputType );
+			throw log.cannotGuessFieldType( inputType, getEventContext() );
```
yrodiere: You removed a TODO by mistake, I think.

mapper/javabean/src/main/java/org/hibernate/search/mapper/javabean/model/impl/JavaBeanTypeModel.java
```diff
@@ -28,7 +28,7 @@
 import org.hibernate.search.mapper.pojo.util.spi.JavaClassOrdering;
 import org.hibernate.search.util.impl.common.LoggerFactory;
 
-class JavaBeanTypeModel<T> implements PojoRawTypeModel<T> {
+public class JavaBeanTypeModel<T> implements PojoRawTypeModel<T> {
```
yrodiere: Making this class public looks unnecessary?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/model/impl/HibernateOrmPropertyModel.java
```diff
@@ -16,12 +17,15 @@
...
 import org.hibernate.annotations.common.reflection.XProperty;
+import org.hibernate.search.mapper.orm.logging.impl.Log;
 import org.hibernate.search.mapper.pojo.model.spi.PojoGenericTypeModel;
 import org.hibernate.search.mapper.pojo.model.spi.PojoPropertyModel;
 import org.hibernate.search.mapper.pojo.model.spi.PropertyHandle;
-import org.hibernate.search.util.SearchException;
+import org.hibernate.search.util.impl.common.LoggerFactory;
 
-class HibernateOrmPropertyModel<T> implements PojoPropertyModel<T> {
+public class HibernateOrmPropertyModel<T> implements PojoPropertyModel<T> {
```
yrodiere: Making this class public looks unnecessary?

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/impl/PropertyIdentifierMapping.java
```diff
@@ -46,8 +50,7 @@ else if ( property != null ) {
 			return (I) caster.cast( id );
 		}
 		else {
-			throw new SearchException( "No identifier was provided, and this mapping does not define"
```
yrodiere: Actually this code looks useless: `property` is always non-null. Could you remove this `else` branch and replace the `else if` above with an `else`? Then you can remove the `cannotExtractIdentifier`  method from the `Log` interface.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/model/spi/MemberPropertyHandle.java
```diff
@@ -48,7 +52,7 @@ public Object get(Object thiz) {
 			if ( e instanceof InterruptedException ) {
 				Thread.currentThread().interrupt();
 			}
-			throw new SearchException( "Exception while invoking '" + member + "' on '" + thiz + "'" , e );
+			throw log.errorInvokingMember( member.getName(), thiz.toString(), e );
```
yrodiere: Just pass `thiz` with type `Object` instead of `thiz.toString()` here, the JBoss logger should call `toString()` automatically, and if for some reason `thiz` is null, it will not throw an NPE like this code.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/dsl/impl/ElasticsearchIndexSchemaFieldContextImpl.java
fax4ever: thanks!

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/dsl/impl/ElasticsearchIndexSchemaFieldContextImpl.java
fax4ever: It is. I'll put back it

mapper/javabean/src/main/java/org/hibernate/search/mapper/javabean/model/impl/JavaBeanTypeModel.java
fax4ever: You're right. I'll remove them.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/model/impl/HibernateOrmPropertyModel.java
fax4ever: You're right. I'll remove them.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/model/spi/MemberPropertyHandle.java
fax4ever: Good idea. Shall I keep ` member.getName()` or I have to change this too?

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/impl/PropertyIdentifierMapping.java
fax4ever: May I remove the else branch at all here?

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/model/spi/MemberPropertyHandle.java
yrodiere: Right, I didn't see you switched from using `toString()` to using `getName`; we actually want `toString()` here, e.g. for a method we don't want just the name, but the return type too. Just pass "member" and expect JBoss Logging to use `toString()`.

fax4ever:
Thanks @yrodiere. I think that it would be a very good idea, but actually I don't know whether is really applicable. Even if we've avoided using the constructors, there are other many usages such as:

```
... catch (SearchException | IOException e) {
  throw log.couldNotNormalizeField( fieldName, e );
}

```
Maybe I wrote the wrong rule in `forbidden-runtime.txt`:

> org.hibernate.search.util.SearchException @ Use an instance of org.jboss.logging.BasicLogger instead

Did I? Thanks in advance

yrodiere:
> org.hibernate.search.util.SearchException @ Use an instance of org.jboss.logging.BasicLogger instead

You referenced the class, you need to reference the constructors instead. For example `org.hibernate.search.util.SearchException#<init>(java.lang.String)`, etc. See [the documentation](https://github.com/policeman-tools/forbidden-apis/wiki/SignaturesSyntax#syntax-of-forbidden-signatures).

fax4ever:
Much better scoping the rule on constructor. We have very few issue now, such as:

> log.unexpectedEmptyEventContext( new SearchException( "Exception for stack trace" ) );

used in `RootFailureCollector#115'. Shall I refactor this case too?

yrodiere:
> used in RootFailureCollector#115. Shall I refactor that case too?

We don't really care about the type of this exception, so let's sidestep the issue: just use an `IllegalStateException`,

fax4ever:
The problem now seems to be bounded to test area. 
Do you know whether there's a way to tell the Forbidden API plugin to skip the test folders?
I tried to use the goal `testCheck` and to exclude the folder...

yrodiere:
> The problem now seems to be bounded to test area.

The test code should have its own set of rules (`forbidden-tests.txt` or something like that).

Maybe it's the main code in the integration test modules that is affected? Like in the TCK module? If so, I think you should be able to solve this by adding something like this to the pluginManagement section of the `hibernate-search-integrationtest-parent` module POM:

```
                <plugin>
                    <groupId>de.thetaphi</groupId>
                    <artifactId>forbiddenapis</artifactId>
                    <executions>
                        <execution>
                            <id>verify-forbidden-apis</id>
                            <phase>none</phase> <!-- Do not use the main rules at all in integration tests, see below -->
                        </execution>
                        <execution>
                            <id>verify-forbidden-test-apis</id>
                            <goals>
                                <goal>check</goal>  <!-- Apply the test rules to all code in integration tests, even code in src/main -->
                                <goal>testCheck</goal>
                            </goals>
                        </execution>
                    </executions>
                </plugin>
```

Does it solve the problem?

fax4ever:
> > The problem now seems to be bounded to test area.
> 
> The test code should have its own set of rules (`forbidden-tests.txt` or something like that).
> 
> Maybe it's the main code in the integration test modules that is affected? Like in the TCK module? If so, I think you should be able to solve this by adding something like this to the pluginManagement section of the `hibernate-search-integrationtest-parent` module POM:
> 
> ```
>                 <plugin>
>                     <groupId>de.thetaphi</groupId>
>                     <artifactId>forbiddenapis</artifactId>
>                     <executions>
>                         <execution>
>                             <id>verify-forbidden-apis</id>
>                             <phase>none</phase> 
>                         </execution>
>                         <execution>
>                             <id>verify-forbidden-test-apis</id>
>                             <goals>
>                                 <goal>check</goal>  
>                                 <goal>testCheck</goal>
>                             </goals>
>                         </execution>
>                     </executions>
>                 </plugin>
> ```
> Does it solve the problem?

Trying to apply that configuration, it seem that tests pass, but no rules have been fired. I've tried exactly this:

> <executions>
                        <execution>
                            <id>verify-forbidden-apis</id>
                            <configuration>
                                ...
                            </configuration>
                            <phase>none</phase>
                            <goals>
                                <goal>check</goal>
                            </goals>
                        </execution>
                        <execution>
                            <id>verify-forbidden-test-apis</id>
                            <configuration>
                                ...
                            </configuration>
                            <phase>test-compile</phase>
                            <goals>
                                <goal>check</goal>  
                                <goal>testCheck</goal>
                            </goals>
                        </execution>
                    </executions>

yrodiere:
> Trying to apply that configuration, it seem that tests pass, but no rules have been fired

I don't know what you did exactly, so I had to try; it does work. I pushed two commits that do the work.

Also, it seems everything else was solved, so I merged it. Thanks!

fax4ever:
Thanks @yrodiere!

PR #1781, opened 2018-10-29
HSEARCH-3064 - HSEARCH-3253 Add more checks for unsortable/non-projectable fields

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-3064
 * https://hibernate.atlassian.net/browse/HSEARCH-3253

Maybe we could refactor things a bit more, creating abstract classes or utils. Not sure it's really worth it but open to discussion.

Better reviewed commit per commit because the store -> projectable commit contains a lot of noise.

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/sort/impl/StandardFieldSortBuilderFactory.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/sort/impl/StandardFieldSortBuilderFactory.java
@@ -12,7 +12,6 @@ import org.hibernate.search.backend.elasticsearch.logging.impl.Log;
 import org.hibernate.search.backend.elasticsearch.search.sort.impl.ElasticsearchSearchSortBuilder;
 import org.hibernate.search.backend.elasticsearch.search.sort.impl.FieldSortBuilderImpl;
 import org.hibernate.search.backend.elasticsearch.types.converter.impl.ElasticsearchFieldConverter;
-import org.hibernate.search.engine.backend.document.model.dsl.Sortable;
 import org.hibernate.search.engine.logging.spi.EventContexts;
 import org.hibernate.search.engine.search.sort.spi.DistanceSortBuilder;
 import org.hibernate.search.engine.search.sort.spi.FieldSortBuilder;
@@ -23,11 +22,11 @@ public class StandardFieldSortBuilderFactory implements ElasticsearchFieldSortBu
 
 	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
 
-	private final Sortable sortable;
+	private final boolean sortable;
 
 	private final ElasticsearchFieldConverter converter;
 
-	public StandardFieldSortBuilderFactory(Sortable sortable, ElasticsearchFieldConverter converter) {
+	public StandardFieldSortBuilderFactory(boolean sortable, ElasticsearchFieldConverter converter) {
 		this.sortable = sortable;
 		this.converter = converter;
 	}
@@ -61,14 +60,10 @@ public class StandardFieldSortBuilderFactory implements ElasticsearchFieldSortBu
 		return converter.isDslCompatibleWith( other.converter );
 	}
 
-	private static void checkSortable(String absoluteFieldPath, Sortable sortable) {
-		switch ( sortable ) {
-			case YES:
-				break;
-			case DEFAULT:
-			case NO:
-				throw log.unsortableField( absoluteFieldPath,
-						EventContexts.fromIndexFieldAbsolutePath( absoluteFieldPath ) );
+	private static void checkSortable(String absoluteFieldPath, boolean sortable) {
+		if ( !sortable ) {
+			throw log.unsortableField( absoluteFieldPath,
+					EventContexts.fromIndexFieldAbsolutePath( absoluteFieldPath ) );
 		}
 	}
 }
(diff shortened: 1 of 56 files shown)


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/projection/impl/GeoPointFieldProjectionBuilderFactory.java
```diff
@@ -63,4 +72,15 @@ public boolean isDslCompatibleWith(LuceneFieldProjectionBuilderFactory obj) {
 		return codec.isCompatibleWith( other.codec ) &&
```
yrodiere: I think the "isCompatible" method should take the "projectable" field into acccount. Otherwise you run the risk of the field being projectable in one index and not the other, and allowing projection anyway. Maybe we should also adapt the exception message to mention that it can happen when the field is not projectable in every index.

Same goes for all the other factories, obviously.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/spatial/DistanceSearchProjectionIT.java
```diff
@@ -189,6 +190,25 @@ public void distanceProjection_nullUnit() {
 				.toProjection();
 	}
 
+	@Test
+	public void distanceProjection_nonProjectable() {
+		IndexSearchTarget searchTarget = indexManager.createSearchTarget().build();
+
+		SubTest.expectException( () -> {
+			searchTarget.projection().field( "nonProjectableGeoPoint", GeoPoint.class ).toProjection();
```
yrodiere: As far as I can see, we already test for this in `SearchProjectionIT`. But I guess it doesn't hurt to do it here, too...

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/sort/impl/StandardFieldSortBuilderFactory.java
```diff
@@ -54,4 +60,15 @@ public boolean isDslCompatibleWith(ElasticsearchFieldSortBuilderFactory obj) {
 
 		return converter.isDslCompatibleWith( other.converter );
```
yrodiere: Same here for the sort  builder factories, we should probably check for the "sortable" parameter to be compatible.

By the way, "comparing" here is a bit tricky, as `DEFAULT` is, theroretically, compatible with `NO`... So before we compare the values with each other, I think we should convert "default" to whatever the default is... Maybe we should use a boolean at this point? If so, I think it should be done in the `*TypeContext`, where we already have a switch on the enum: just add a `boolean sortableDefaulted` variable and populate it in the switch.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/spatial/AbstractSpatialWithinSearchPredicateIT.java
```diff
@@ -112,14 +112,17 @@ protected void initData() {
...
 		final IndexFieldAccessor<GeoPoint> nonProjectableGeoPoint;
+		final IndexFieldAccessor<GeoPoint> unsortableGeoPoint;
 		final IndexFieldAccessor<String> string;
 
 		IndexAccessors(IndexSchemaElement root) {
-			geoPoint = root.field( "geoPoint" ).asGeoPoint().projectable( Projectable.YES ).createAccessor();
-			geoPoint_1 = root.field( "geoPoint_1" ).asGeoPoint().projectable( Projectable.YES ).createAccessor();
-			geoPoint_2 = root.field( "geoPoint_2" ).asGeoPoint().projectable( Projectable.YES ).createAccessor();
-			geoPoint_with_longName = root.field( "geoPoint_with_a_veeeeeeeeeeeeeeeeeeeeerrrrrrrrrrrrrrrrrryyyyyyyyyyyyyyyy_long_name" ).asGeoPoint().projectable( Projectable.YES ).createAccessor();
+			geoPoint = root.field( "geoPoint" ).asGeoPoint().sortable( Sortable.YES ).projectable( Projectable.YES ).createAccessor();
```
yrodiere: Do we really need the fields to be projectable or sortable for the *predicate* tests? That looks odd.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/projection/impl/GeoPointFieldProjectionBuilderFactory.java
gsmet: Right, fixed.

At the moment, the message is very vague. Not a big fan of making it too specific while a lot of things move every day. Better vague than wrong. Maybe let's wait for some feedback from the field.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/sort/impl/StandardFieldSortBuilderFactory.java
gsmet: Right, fixed.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/spatial/AbstractSpatialWithinSearchPredicateIT.java
gsmet: It's also the base class for DistanceSearchProjectionIT. And we project and sort on some of those fields.

gsmet:
@yrodiere pushed a few additional commits to address your comments.

yrodiere:
I just rebased following the merging of #1780 . Had to fix a few conflicts, let's wait for CI...

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/spatial/AbstractSpatialWithinSearchPredicateIT.java
yrodiere: Mkay. I guess that spares us some copy-pasting.

PR #1782, opened 2018-10-31
HSEARCH-3290 Search 6 groundwork - Finalize the Search DSL syntaxes

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3290

Best reviewed commit by commit, as some of these commits are just renamings and introduce a lot of noise. And best reviewed in the topological order, i.e. not the order GitHub displays commits in...

Here is a summary of what this PR brings:

1. It removes the fluent ("fluid") syntax from the predicate/sort DSLs, as we agreed to during the face-to-face meeting, in order to reduce the confusion. Only the object-based and lambda-based syntaxes remain.
3. It renames the `.end()` methods in the predicate/sort DSLs to `toPredicate()`/`toSort()`, as we agreed to during the face-to-face meeting.
4. Additionally, it changes the lambda syntax substantially in order to avoid confusing contexts when nesting predicates with the lambda syntax. See commit `
HSEARCH-3290 Remove any notion of "position in the predicate tree" fr… ` and its commit message for details.
5. And finally, it allows to pass `SearchPredicateTerminalContext` to DSL methods such as `must(...)` or `.except(...)`, in order to reduce the verbosity of the DSL. It's a bit dangerous in my opinion, as it may lead people into using `SearchPredicateTerminalContext` to model cached predicates instead of `SearchPredicate`, but it's far less dangerous now than it used to be: before item 4., it could lead to unpredictable results, especially when using the cached predicate at different locations of the tree; now it's just sub-optimal (as we would keep references to DSL context objects that are no longer needed).


gsmet:
Merged, thanks!

PR #1783, opened 2018-10-31
HSEARCH-3259 Search 6 groundwork - Restore support for projections on fields within flattened object fields in Elasticsearch

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3259

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/gson/impl/JsonObjectAccessorImpl.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/gson/impl/JsonObjectAccessorImpl.java
@@ -6,6 +6,8 @@
  */
 package org.hibernate.search.backend.elasticsearch.gson.impl;
 
+import java.util.regex.Pattern;
+
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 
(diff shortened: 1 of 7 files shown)


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/gson/impl/JsonObjectAccessorImpl.java
```diff
@@ -37,4 +41,17 @@ public UnknownTypeJsonAccessor property(String propertyName) {
 		return new ObjectPropertyJsonAccessor( this, propertyName );
 	}
 
+	@Override
+	public UnknownTypeJsonAccessor path(String dotSeparatedPath) {
+		String[] components = DOT_REGEX.split( dotSeparatedPath );
```
gsmet: Wondering if we should add a `Contracts#assertNotNullNorEmpty()` call for this method and the method above.

yrodiere: I don't think it can happen in this specific case, but it cannot hurt to check. Added a commit, will merge as soon as CI is green.

yrodiere:
Merged, thanks!

PR #1784, opened 2018-10-31
HSEARCH-3388 Search 6 groundwork - Output the full property key (as provided by the user) in case of parsing error

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3388

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/impl/LuceneBackendFactory.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/impl/LuceneBackendFactory.java
@@ -104,7 +104,11 @@ public class LuceneBackendFactory implements BackendFactory {
 		}
 		else {
 			Version latestVersion = SearchBackendLuceneSettings.Defaults.LUCENE_VERSION;
-			log.recommendConfiguringLuceneVersion( latestVersion, backendContext );
+			log.recommendConfiguringLuceneVersion(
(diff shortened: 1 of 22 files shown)


gsmet:
Can't say I'm excited by this one. Wouldn't it be better to keep the original property handy so that we don't have to resolve it back?

yrodiere:
> Wouldn't it be better to keep the original property handy so that we don't have to resolve it back?

Not sure what you're suggesting, but one problem is that we don't know the full property key where we use it, we only know the radical. So we can't really display the full property key without some additional code to resolve it.

For example in the ORM integration we use the `hibernate.search.` prefix, but in Infinispan I know for a fact that they currently (with Search 5) prepend the prefix in their integration, and users don't use it when configuring (or they use another prefix, I'm not sure). So if you consider the property defining the backend to use for "myIndex", the full key will be "hibernate.search.index.myIndex.backend" in the ORM integration, but it will be something like "index.myIndex.backend" in Infinispan. Or maybe it will be even more different, because Infinispan is configured per-cache and has only one index per cache; but in any case, this PR will allow their configuration property source to resolve the key to a string that makes sense for end users.

Out of curiosity, what is your main concern? Maybe I can do something about it.

gsmet:
> Out of curiosity, what is your main concern? Maybe I can do something about it.

Well, it's just that it feels weird to have the property to begin with, then lose some information about it, and then resolve it back to what it was.

But I suppose it's how it is.

Merging.

gsmet:
Merged!

PR #1785, opened 2018-11-02
HSEARCH-3217 Search 6 groundwork - Make sure the index schema definition DSL allows for non-standard or extended accessor types

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3217

Essentially I only had to make sure that `IndexSchemaFieldTypedContext` does not extend `IndexSchemaFieldTerminalContext`, but only its sub-interfaces do (where relevant). This will allow us, in the future, to introduce new "typed contexts" whose terminal context is not exactly the same, i.e. it returns a different type of accessor that offers a different API, like for example an accessor that expects two values instead of one (the center of a circle and its radius). Not sure we will need it, but at least we'll be able to do it if need be.

You can find an example of what this would allow to change in our APIs in the HEAD commit of this branch: https://github.com/yrodiere/hibernate-search/tree/HSEARCH-3217_proofitworks

Looking back, it's not a very interesting change, nor a very useful one since we'll probably rework these APIs in [HSEARCH-3291](https://hibernate.atlassian.net/browse/HSEARCH-3291) . But it's done and doesn't hurt, so let's merge it?


gsmet:
Merged, thanks.

PR #1787, opened 2018-11-05
HSEARCH-3398 Use dedicated APIs for the ORM mapper instead of extending the POJO mapper APIs, where possible

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3398


mapper/orm/src/main/java/org/hibernate/search/mapper/orm/mapping/HibernateOrmMapping.java
```diff
@@ -8,15 +8,38 @@
...
 
-public interface HibernateOrmMapping extends PojoMapping {
+public interface HibernateOrmMapping {
 
 	HibernateOrmSearchManager createSearchManager(EntityManager entityManager);
 
 	HibernateOrmSearchManagerBuilder createSearchManagerWithOptions(EntityManager entityManager);
 
+	/**
+	 * @param type a Java type
```
gsmet: Don't you usually have a capital and a dot?

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/spi/PojoMappingDelegate.java
```diff
@@ -22,4 +21,26 @@
 	<T> PojoSearchTargetDelegate<T> createPojoSearchTarget(Collection<? extends Class<? extends T>> targetedTypes,
 			PojoSessionContextImplementor sessionContext);
 
+	/**
+	 * @param type a Java type
```
gsmet: Same question here about the javadoc format?

mapper/javabean/src/main/java/org/hibernate/search/mapper/javabean/session/impl/JavaBeanSearchManagerImpl.java
```diff
@@ -44,9 +45,9 @@ public void close() {
 	}
 
 	@Override
-	public PojoWorkPlan getMainWorkPlan() {
+	public JavaBeanWorkPlan getMainWorkPlan() {
 		if ( workPlan == null ) {
-			workPlan = getDelegate().createWorkPlan();
+			workPlan = new JavaBeanWorkPlanImpl( getDelegate().createWorkPlan() );
```
gsmet: Is this change related to this commit? It seems like a totally different thing?

BTW, what is the reason for introducing this class? Will we do the same for the ORM mapper?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/mapping/HibernateOrmMapping.java
yrodiere: Done

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/spi/PojoMappingDelegate.java
yrodiere: Done

mapper/javabean/src/main/java/org/hibernate/search/mapper/javabean/session/impl/JavaBeanSearchManagerImpl.java
yrodiere: It was not related, apparently I merged two commits together by mistake. I split them back in two, the new commit is ["HSEARCH-3398 Move PojoWorkPlan to SPI"](https://github.com/hibernate/hibernate-search/pull/1787/commits/4fdac2961ca73c4e6cd052526ca55dc326ae5a0b).

In short, I had to create this class because I wanted to move `PojoWorkPlan` to SPI and hide it from users of the JavaBean mapper (which is the main goal of this PR).

On the ORM side, I don't really need to do that, because the work plan is only used internally and is not exposed through APIs (yet): indexing is done automatically. If we ever expose manual indexing interfaces to users, then yes we may need a dedicated interface. Though I'm not sure we will: we may decide to hide the concept of work plan and just add methods to the `FullTextSession` interface (for example an `FullTextSession#index(Object)` method like we had in Search 5).

If you need a tangible proof that it's not needed, well, JQAssistant checks didn't fail, even though `PojoWorkPlan` was moved to an SPI package, so we are sure it's not referenced from any API class or interface.

gsmet: OK, thanks for the explanation!

yrodiere:
Merged, thanks!

PR #1788, opened 2018-11-05
HSEARCH-3395 Make the .reference() and .object() projections type-safe in the Projection DSL

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3395

Based on #1787 , which should be merged first.


gsmet:
The additional commits look good.

yrodiere:
Merged, thanks!

PR #1789, opened 2018-11-05
HSEARCH-3383 Upgrade to a version of jqassistant that works correctly when we skip modules

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3383


gsmet:
Merged, thanks!

PR #1790, opened 2018-11-05
HSEARCH-3336 Rename artifact hibernate-search-orm to hibernate-search-mapper-orm (and change jigsaw module name)

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3336


gsmet:
Merged, thanks!

PR #1792, opened 2018-11-06
HSEARCH-3238 Upgrade to Easymock 4.0.1

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-3238


yrodiere:
Just launched a build on all environments (all DBs, all JDKs, ...): http://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-1792/2/

It will fail, but let's see what works exactly, and perhaps adjust our expectations (move JDK11 from "experimental" to "supported"?)

yrodiere:
As discussed, let's add `--add-opens java.base/java.lang=ALL-UNNAMED` to the JVM args with JDK11.

I created https://github.com/easymock/easymock/issues/235 , maybe it will get fixed soon.

yrodiere:
Closing, will open another PR with more changes than just Easymock.

PR #1795, opened 2018-11-06
HSEARCH-2496 Smarter dirty checking when using @ContainedIn with custom field bridges

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2496

Based on #1794 , which should be merged first.

I only had to add a test, as this is already supported in Search 6.


gsmet:
Merged, thanks!

PR #1796, opened 2018-11-07
HSEARCH-3407 Boolean fields in embedded collections are not indexed properly when using Elasticsearch

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3407


gsmet:
Merged, thanks!

PR #1798, opened 2018-11-08
HSEARCH-1788 Projection of a single element through FullTextQuery should not be wrapped in a single-element array

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-1788

Best reviewed commit by commit, as there are a few test refactorings.


engine/src/main/java/org/hibernate/search/engine/common/impl/MappedIndexSearchTargetImpl.java
```diff
@@ -83,6 +84,26 @@ public String toString() {
 		);
 	}
 
+	@Override
+	public <P, T, Q> SearchQueryResultContext<Q> queryAsProjections(SessionContextImplementor sessionContext,
```
gsmet: Shouldn't it be `queryAsProjection()` without an s?

engine/src/main/java/org/hibernate/search/engine/common/impl/MappedIndexSearchTargetImpl.java
```diff
@@ -91,7 +112,7 @@ public String toString() {
 			Function<SearchQuery<T>, Q> searchQueryWrapperFactory,
 			SearchProjection<?>... projections) {
 		HitAggregator<ProjectionHitCollector, List<T>> hitAggregator =
-				new ProjectionHitAggregator<>( documentReferenceTransformer, objectLoader, hitTransformer,
+				new MultipleProjectionHitAggregator<>( documentReferenceTransformer, objectLoader, hitTransformer,
```
gsmet: In this case, shouldn't it be `MultipleProjectionsHitAggregator`? It's weird to have Multiple and a singular word.

engine/src/main/java/org/hibernate/search/engine/search/query/impl/SingleProjectionHitAggregator.java
```diff
@@ -0,0 +1,72 @@
...
+		}
+		return hitTransformer.apply( loadedHit );
+	}
+
+	@Override
+	protected AbstractProjectionHitCollector getHitCollector() {
+		return hitCollector;
+	}
+
+	private class MultipleProjectionHitCollectorImpl extends AbstractProjectionHitCollector {
```
gsmet: It should be `Single` shouldn't it?

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/projection/SearchProjectionIT.java
```diff
@@ -454,19 +459,20 @@ public void field_inNestedObject() {
...
-				b.list( fieldModel.document1Value );
-				b.list( fieldModel.document2Value );
-				b.list( fieldModel.document3Value );
-				b.list( null ); // Empty document
+			SubTest.expectSuccess( fieldModel, model -> {
+				String fieldPath = indexMapping.nestedObject.relativeFieldName + "." + model.relativeFieldName;
+
+				assertThat(
+						searchTarget.query()
+								.asProjections( searchTarget.projection().field( fieldPath, model.type ).toProjection() )
```
gsmet: I think I would name it `asProjection()`. Otherwise, I would kinda expect the return type to be consistent when adding another projection to be returned.

The methods have totally different behavior.

yrodiere: So, two things.

Regarding the ability to use this API with "generic" code: generic code will call the `asProjections()` method that accepts an array of projections as a parameter. It will read something like `.asProjections( myProjections.toArray( new SearchProjection[0] ) )`. So the fact that an overload with a single parameter exists does not matter: the method will return a `SearchQuery<List<?>>`, even if their array only contains one element.

About consistency, the other methods are named asObjects/asEntities/asReferences, depending on where you look. But they only return one element per hit. So the "s" here is about the fact there can be multiple hits, so multiple objects/entities/references. With that reasoning, we should keep `asProjections()`, with an "s", even when there is one projection: there will still be multiple hits, so multiple "projection results".

So if you want to remove the "s", I say okay, but let's remove it from the other methods too (except the one that take a projection array in parameter).

engine/src/main/java/org/hibernate/search/engine/common/impl/MappedIndexSearchTargetImpl.java
yrodiere: I thought of it like "a three-child family" (which is grammatically correct). But whatever, I'll change it.

engine/src/main/java/org/hibernate/search/engine/search/query/impl/SingleProjectionHitAggregator.java
yrodiere: Right; fixed.

engine/src/main/java/org/hibernate/search/engine/common/impl/MappedIndexSearchTargetImpl.java
gsmet: Looks like this one got forgotten or was it intentional?

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/spi/MappedIndexSearchTarget.java
```diff
@@ -40,6 +40,21 @@
 			Function<R, T> hitTransformer,
 			Function<SearchQuery<T>, Q> searchQueryWrapperFactory);
 
+	/*
+	 * IMPLEMENTATION NOTE: we *must* only accept an object loader with the same R/O type parameters as this class,
+	 * otherwise some casts in ObjectProjectionContextImpl and ReferenceProjectionContextImpl
+	 * will be wrong.
+	 * In particular, we cannot accept an ObjectLoader<R, T> like we do in queryAsLoadedObjects(...).
+	 */
+	<P, T, Q> SearchQueryResultContext<Q> queryAsProjections(
```
gsmet: Same here.

engine/src/main/java/org/hibernate/search/engine/search/query/impl/MultipleProjectionsHitAggregator.java
```diff
@@ -0,0 +1,77 @@
...
+		}
+		return hitTransformer.apply( hit );
+	}
+
+	@Override
+	protected AbstractProjectionHitCollector getHitCollector() {
+		return hitCollector;
+	}
+
+	private class MultipleProjectionHitCollectorImpl extends AbstractProjectionHitCollector {
```
gsmet: It should also be called `MultipleProjections` then.

engine/src/main/java/org/hibernate/search/engine/common/impl/MappedIndexSearchTargetImpl.java
yrodiere: I renamed the method with the same name in PojoSearchTargetDelegate and forgot there was also this one. And of course that's the only one I forgot to check with a global search afterwards...

Well spotted. Fixed.

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/spi/MappedIndexSearchTarget.java
yrodiere: Fixed

engine/src/main/java/org/hibernate/search/engine/search/query/impl/MultipleProjectionsHitAggregator.java
yrodiere: Fixed

gsmet:
Merged, thanks!

PR #1801, opened 2018-11-12
HSEARCH-3361 Test behavior of default value bridges when projecting

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3361

Unfortunately I had to add a bit more logic to the stub backend in order to test the projection converters, but I think it's a reasonable trade-off.


gsmet:
Merged, thanks!

PR #1802, opened 2018-11-12
HSEARCH-2641 Allow for Elasticsearch analyzer definitions to be applied in class bridges/custom field bridges

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2641


integrationtest/mapper/pojo/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/BridgeIT.java
```diff
@@ -31,22 +34,153 @@
...
+				return id;
+			}
+			public String getStringProperty() {
+				return stringProperty;
+			}
+		}
+
+		backendMock.expectSchema( INDEX_NAME, b ->
+				b.field( "someField", String.class, b2 -> {
+					b2.analyzerName( "myAnalyzer" ); // For HSEARCH-2641
```
yrodiere: @gsmet Here, I check that we forward the information to the backend "you should use this analyzer for this field". Whether the backend actually complies with this instruction is another problem, and we already test that in the backend TCK (`org.hibernate.search.integrationtest.backend.tck.analysis.AnalysisCustomIT` in particular).

PR #1803, opened 2018-11-13
HSEARCH-3413 Search 6 groundwork - Restore support for implicit index name

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3413

### DIFF
--- a/mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/impl/AnnotationMappingDefinitionContextImpl.java
+++ b/mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/impl/AnnotationMappingDefinitionContextImpl.java
@@ -83,8 +83,12 @@ public class AnnotationMappingDefinitionContextImpl implements AnnotationMapping
 				.forEach( typeModel -> {
 					Optional<Indexed> indexedAnnotation = typeModel.getAnnotationByType( Indexed.class );
 					if ( indexedAnnotation.isPresent() ) {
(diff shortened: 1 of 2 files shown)


mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/impl/AnnotationMappingDefinitionContextImpl.java
```diff
@@ -83,8 +83,12 @@ public void configure(MappingBuildContext buildContext, ConfigurationPropertySou
 				.forEach( typeModel -> {
 					Optional<Indexed> indexedAnnotation = typeModel.getAnnotationByType( Indexed.class );
 					if ( indexedAnnotation.isPresent() ) {
+						String defaultedIndexName = indexedAnnotation.get().index();
```
gsmet: Wondering if we should apply a `trim()`.

yrodiere: I wouldn't want to start supporting `@Indexed(index = "    ")` as a way to say "use the default mapper". So I wouldn't trim the index name here.

We may want to clean up index names automatically at some point, but if we do, I would do it either in the backend or in the implementation of `collector.mapToIndex`. Not sure it's a good thing either, though... Let's wait for users to provide valid reasons to pass index names that require some cleanup.

yrodiere:
Merged, thanks!

PR #1805, opened 2018-11-16
HSEARCH-3088 Query id in DSL

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-3088

I hope it makes sense.
At the moment it only supports strings.

It looks like:
```
		SearchQuery<DocumentReference> query = searchTarget.query()
				.asReference()
				.predicate( f -> f.id().matching( DOCUMENT_1 ).toPredicate() )
				.build();
```

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/search/dsl/predicate/MatchIdPredicateContext.java
+++ b/engine/src/main/java/org/hibernate/search/engine/search/dsl/predicate/MatchIdPredicateContext.java
@@ -6,6 +6,8 @@
  */
 package org.hibernate.search.engine.search.dsl.predicate;
 
+import java.util.Collection;
+
 /**
  * The context used when defining a match on an identifier.
  */
@@ -16,8 +18,23 @@ public interface MatchIdPredicateContext extends SearchPredicateTerminalContext 
 	 * <p>
 	 * If used multiple times, it will target any of the specified values.
 	 * <p>
+	 * @see #matchingAny(Collection)
 	 * @param value the value of the id we want to match.
 	 * @return {@code this} for method chaining.
 	 */
 	MatchIdPredicateContext matching(Object value);
+
+	/**
+	 * Target the identifiers matching any of the values in a collection.
+	 * <p>
+	 * @param values the collection of identifiers to match.
+	 * @return {@code this} for method chaining.
+	 */
+	default MatchIdPredicateContext matchingAny(Collection<Object> values) {
+		MatchIdPredicateContext context = null;
+		for ( Object value : values ) {
+			context = matching( value );
+		}
+		return context;
+	}
 }
--- a/engine/src/main/java/org/hibernate/search/engine/search/dsl/predicate/MatchIdPredicateContext.java
+++ b/engine/src/main/java/org/hibernate/search/engine/search/dsl/predicate/MatchIdPredicateContext.java
@@ -1,0 +1,23 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.engine.search.dsl.predicate;
+
+/**
+ * The context used when defining a match on an identifier.
+ */
+public interface MatchIdPredicateContext extends SearchPredicateTerminalContext {
+
+	/**
+	 * Target the identifier with the given id.
+	 * <p>
+	 * If used multiple times, it will target any of the specified values.
+	 * <p>
+	 * @param value the value of the id we want to match.
+	 * @return {@code this} for method chaining.
+	 */
+	MatchIdPredicateContext matching(Object value);
+}
--- a/engine/src/main/java/org/hibernate/search/engine/search/dsl/predicate/impl/MatchIdPredicateContextImpl.java
+++ b/engine/src/main/java/org/hibernate/search/engine/search/dsl/predicate/impl/MatchIdPredicateContextImpl.java
@@ -1,0 +1,37 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.engine.search.dsl.predicate.impl;
+
+import org.hibernate.search.engine.search.dsl.predicate.MatchIdPredicateContext;
+import org.hibernate.search.engine.search.dsl.predicate.SearchPredicateFactoryContext;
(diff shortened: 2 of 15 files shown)


engine/src/main/java/org/hibernate/search/engine/search/dsl/predicate/impl/MatchIdPredicateContextImpl.java
```diff
@@ -0,0 +1,37 @@
...
+	private final MatchIdPredicateBuilder<B> matchIdBuilder;
+
+	MatchIdPredicateContextImpl(SearchPredicateBuilderFactory<?, B> factory, SearchPredicateFactoryContext factoryContext) {
+		super( factory );
+		this.matchIdBuilder = factory.id();
+	}
+
+	@Override
+	public MatchIdPredicateContext matching(Object value) {
+		matchIdBuilder.value( value );
```
yrodiere: I see you allow multiple values, which is nice; wouldn't be useful to also provide a `matchingAny(Collection<String>)` method? From what I can see it can be implemented easily here with a simple loop, without changing the backend.

engine/src/main/java/org/hibernate/search/engine/search/predicate/spi/MatchIdPredicateBuilder.java
```diff
@@ -0,0 +1,13 @@
...
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.engine.search.predicate.spi;
+
+public interface MatchIdPredicateBuilder<B> extends SearchPredicateBuilder<B> {
+
+	void value(Object values);
```
yrodiere: `values` should be singular here (`value`), I think? Or did I misunderstand how it works?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/predicate/impl/ElasticsearchFieldPredicateBuilderFactory.java
```diff
@@ -29,6 +30,9 @@
 	 */
 	boolean isDslCompatibleWith(ElasticsearchFieldPredicateBuilderFactory other);
 
+	MatchIdPredicateBuilder<ElasticsearchSearchPredicateBuilder> createMatchIdPredicateBuilder(
```
yrodiere: This is unnecessary; you can remove this method and its implementations.

To be clear:

* `SearchPredicateBuilderFactory` is the entry point for the engine to retrieve predicate builders. It's implemented by the backend.
* In some cases, for field-scoped predicates (match, range), `SearchPredicateBuilderFactory` implementations delegate to a field-scoped, backend-specific `ElasticsearchFieldPredicateBuilderFactory` or `LuceneFieldPredicateBuilderFactory`. But that's only when the predicate can have a different behavior depending on the field being targeted: the `match` predicate will accept different input based on whether you target an integer or `LocalDate` field, for example, but this concept doesn't make sense for the `bool` predicate, which doesn't really "target" a field.

So in the case of your ID predicate, you can see that you don't need to change the `ElasticsearchFieldPredicateBuilderFactory`, only the `SearchPredicateBuilderFactory` (which you already did).

I'll try to add some javadoc in a separate PR to make that clearer.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/predicate/impl/LuceneFieldPredicateBuilderFactory.java
```diff
@@ -32,6 +33,9 @@
 	 */
 	boolean isDslCompatibleWith(LuceneFieldPredicateBuilderFactory other);
 
+	MatchIdPredicateBuilder<LuceneSearchPredicateBuilder> createMatchIdPredicateBuilder(
```
yrodiere: As for the Elasticsearch backend, this is unnecessary; you can remove this method and its implementations.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/MatchSearchPredicateIT.java
```diff
@@ -69,6 +69,35 @@ public void setup() {
 		initData();
 	}
 
+	@Test
+	public void match_id() {
```
yrodiere: We tried to create one test class per type of predicate: `MatchSearchPredicateIT`, `RangeSearchPredicateIT`, etc.

Could you please move this to a different test class?

The class can be quite simple, much simpler than this one, because you don't really have to test multiple types of fields. Maybe you could take inspiration from what @fax4ever did here: https://github.com/hibernate/hibernate-search/commit/d0d1c3e1c8e94704451e976f95553cb5071bf694

yrodiere:
Follow-up ticket: https://hibernate.atlassian.net/browse/HSEARCH-3416

engine/src/main/java/org/hibernate/search/engine/search/dsl/predicate/impl/MatchIdPredicateContextImpl.java
DavideD: Sure, sounds good

engine/src/main/java/org/hibernate/search/engine/search/predicate/spi/MatchIdPredicateBuilder.java
DavideD: no, it's a typo

DavideD:
I've applied the remarks and added the method `matchingAny`.

engine/src/main/java/org/hibernate/search/engine/search/dsl/predicate/MatchIdPredicateContext.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.engine.search.dsl.predicate;
+
+import java.util.Collection;
+
+/**
+ * The context used when starting to define a match predicate.
```
yrodiere: Just noticed this: this is apparently a copy/paste from the match predicate, it should be rephrased or removed.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/predicate/impl/LuceneFieldPredicateBuilderFactory.java
yrodiere: It seems you didn't take my comment into account? This method and its implementations should be removed.

DavideD:
Should be fixed now.

I've updated the javadoc and removed the method I forgot to remove before (I did it for elasticsearch but I forgot to do it for Lucene).

yrodiere:
Merged, thanks!

PR #1808, opened 2018-11-16
HSEARCH-3409 Add support for the Instant type in backends and use it to provide support for java.util.Date

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3409


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/codec/impl/InstantFieldCodec.java
```diff
@@ -15,31 +15,31 @@
 import com.google.gson.JsonPrimitive;
 
 /**
- * Encode and decode a {@link Date} to a json form.
+ * Encode and decode a {@link Instant} to a json form.
```
gsmet: s/a/an/

yrodiere:
Merged, thanks

PR #1812, opened 2018-11-19
HSEARCH-3390 Implement composite projections

gsmet:
https://hibernate.atlassian.net/browse/HSEARCH-3390


util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/rule/SearchWorkCall.java
```diff
@@ -44,13 +60,35 @@
...
 		long totalHitCount = behavior.getTotalHitCount();
-		@SuppressWarnings({ "unchecked", "rawtypes" })
-		List<U> hits = (List<U>) ((StubHitExtractor) actualCall.hitExtractor).extract( behavior.getRawHits() );
-		return new SearchResultImpl<>( totalHitCount, hits );
+
+		return new SearchResultImpl<>( totalHitCount, getResults( actualCall.convertContext,
+				actualCall.projectionHitMapper, actualCall.rootProjection, behavior.getRawHits() ) );
+	}
+
+	private static <U> List<U> getResults(FromIndexFieldValueConvertContext actualConvertContext,
```
yrodiere: Keeping the `StubHitExtractor` class and encapsulating the `convertContext`, `projectionHitMapper` and `rootProjection` in it would have spared you this code, you would just have delegated to the extractor. But I suppose it's just as well.

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/rule/SearchWorkCall.java
```diff
@@ -44,13 +60,35 @@
 				.as( "Search work on indexes " + indexNames + " did not match: " )
 				.matches( work );
 		/*
-		 * Checking the result type in the work assertion above guarantees
-		 * that the hit collector has the correct type. We can cast safely.
+		 * Checking the result type in the work assertion above guarantees that the hit collector has the correct type.
```
yrodiere: This comment is obsolete, you can just remove it.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/query/projection/LoadingProjectionIT.java
```diff
@@ -0,0 +1,234 @@
...
+
+		sessionFactory = ormSetupHelper.withBackendMock( backendMock )
+				.setup( Book.class );
+
+		backendMock.verifyExpectationsMet();
+	}
+
+	@Test
+	public void index() {
+		OrmUtils.withinTransaction( sessionFactory, session -> {
```
yrodiere: The database is cleared between each test method, so if you want to use the persisted data, you have to persist it in the same test method as you use it. In this case, you need to persist it in `search`.

You will tell me "but the test pass!" or "you did the same in the smoke tests" and I will agree: the tests are buggy, and it's concerning.

I think it's because the calls to `session.get( Book.class, 1 )` return null, and similarly the projection returns null books because the database is empty.

An easy way to fix all this would be to move this indexing code to a private `initData` method and call that method in `setup`.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/query/projection/LoadingProjectionIT.java
```diff
@@ -0,0 +1,234 @@
...
+					b -> { },
+					StubSearchWorkBehavior.of(
+							3L,
+							Arrays.asList(
+									Arrays.asList( StubBackendUtils.reference( Book.INDEX, "1" ), AUTHOR_4_3_2_1 ),
+									1.0F
+							),
+							Arrays.asList(
+									Arrays.asList( StubBackendUtils.reference( Book.INDEX, "2" ), AUTHOR_CIDER_HOUSE ),
+									1.0F
```
yrodiere: Maybe use different scores?

engine/src/main/java/org/hibernate/search/engine/search/dsl/projection/SearchProjectionFactoryContext.java
```diff
@@ -78,4 +84,38 @@
 	 * @return A context allowing to define the projection more precisely.
 	 */
 	DistanceToFieldProjectionContext distance(String absoluteFieldPath, GeoPoint center);
+
+	default CompositeProjectionContext<List<?>> composite(SearchProjection<?>... projections) {
+		return composite( Function.identity(), projections );
+	}
+
+	/**
+	 * Creates a composite projection e.g. a projection composed of other projections.
```
yrodiere: You mean `i.e.`, not `e.g.`?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ReferenceSearchProjectionBuilderImpl.java
```diff
@@ -6,19 +6,21 @@
...
 package org.hibernate.search.backend.elasticsearch.search.projection.impl;
 
-import org.hibernate.search.backend.elasticsearch.search.extraction.impl.DocumentReferenceExtractorHelper;
 import org.hibernate.search.engine.search.SearchProjection;
 import org.hibernate.search.engine.search.projection.spi.ReferenceSearchProjectionBuilder;
 
 
 class ReferenceSearchProjectionBuilderImpl implements ReferenceSearchProjectionBuilder {
 
+	@SuppressWarnings("rawtypes")
```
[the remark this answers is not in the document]
gsmet: Yes, done.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/SearchQueryBuilderFactoryImpl.java
```diff
@@ -41,49 +40,54 @@
 
 	@Override
 	public <O> SearchQueryBuilder<O, ElasticsearchSearchQueryElementCollector> asObject(
```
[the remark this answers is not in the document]
gsmet: I think I prefer to keep them in the `SearchQueryBuilderFactory` and keep the `MappedIndexSearchTargetImpl` not knowing about how we indeed build the queries.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/SearchQueryBuilderFactoryImpl.java
```diff
@@ -41,49 +40,54 @@
...
+	public <T> SearchQueryBuilder<T, ElasticsearchSearchQueryElementCollector> asProjection(
+			SessionContextImplementor sessionContext, ProjectionHitMapper<?, ?> projectionHitMapper,
+			SearchProjection<T> projection) {
+		return createSearchQueryBuilder( sessionContext, projectionHitMapper,
+				searchProjectionFactory.toImplementation( projection ) );
+	}
 
-		return createSearchQueryBuilder( sessionContext, hitExtractor, hitAggregator );
+	@Override
+	public SearchQueryBuilder<List<?>, ElasticsearchSearchQueryElementCollector> asProjections(
```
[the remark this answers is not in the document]
gsmet: Same answer.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/SearchQueryBuilderFactoryImpl.java
```diff
@@ -44,43 +41,46 @@
 
 	@Override
 	public <O> SearchQueryBuilder<O, LuceneSearchQueryElementCollector> asObject(
```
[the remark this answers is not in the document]
gsmet: Same answer.

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/rule/SearchWorkCall.java
gsmet: So the idea is that I would like to have the Stub part as consistent with the rest as possible. It's already quite different tbh and it's a bit hard to make the changes here after doing the changes in the backends.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/query/projection/LoadingProjectionIT.java
gsmet: Ah, I saw it at one point but forgot to change it. Done now.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchSearchProjection.java
```diff
@@ -6,10 +6,42 @@
...
-import org.hibernate.search.backend.elasticsearch.search.extraction.impl.HitExtractor;
 import org.hibernate.search.engine.search.SearchProjection;
-import org.hibernate.search.engine.search.query.spi.ProjectionHitCollector;
+import org.hibernate.search.engine.search.query.spi.LoadingResult;
+import org.hibernate.search.engine.search.query.spi.ProjectionHitMapper;
 
-public interface ElasticsearchSearchProjection<T> extends SearchProjection<T>, HitExtractor<ProjectionHitCollector> {
+import com.google.gson.JsonObject;
 
+public interface ElasticsearchSearchProjection<T> extends SearchProjection<T> {
```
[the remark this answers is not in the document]
gsmet: Yup, I was hesitant because I thought it would be a mess but in the end, I think it's better than what we had.

Pushed an additional commit with it.

gsmet:
@yrodiere fixed all the elements marked as resolved.

Happy to discuss the remaining ones if you think we need to discuss them.

yrodiere:
Looks good to me, I'll merge it. Thanks a lot!

Could you please answer to this before I start creating follow-up tickets?

> I think the following matters should be addressed in a follow-up ticket; please tell me if I forget anything:
> 
> *    Lambda syntax for projections, i.e. asProjection( f -> f.field( "title", String.class ).toProjection() ) or .composite( Book::new, f -> f.field( "title", String.class ).toProjection() ). That would allow to simplify a few calls in our tests.
>  *   I think you also wanted to offer easy-to-use asProjection methods in the mapping APIs, that would duplicate the composite( Function, SearchProjection) and similar methods. Not sure if it's still worth it, especially if we allow the lambda syntax above?
>   *  I think that we could build on your changes to move parts of the implementation of projections to the engine. That would require introducing a few SPIs, but it may be doable without too much hassle. I'd like to have a look, at least.

gsmet:
> Lambda syntax for projections, i.e. asProjection( f -> f.field( "title", String.class ).toProjection() ) or .composite( Book::new, f -> f.field( "title", String.class ).toProjection() ). That would allow to simplify a few calls in our tests.

Yes, we discuss that, I think we need it to be consistent with the rest and avoid some boiler plate code.

> I think you also wanted to offer easy-to-use asProjection methods in the mapping APIs, that would duplicate the composite( Function, SearchProjection) and similar methods. Not sure if it's still worth it, especially if we allow the lambda syntax above?

I now think it's better if we don't do it. Considering we have nested composites, Let's be consistent and enforce composites everywhere.

> I think that we could build on your changes to move parts of the implementation of projections to the engine. That would require introducing a few SPIs, but it may be doable without too much hassle. I'd like to have a look, at least.

Yes, that might well be the case. Not sure it should be high priority though.

yrodiere:
Created https://hibernate.atlassian.net/browse/HSEARCH-3430 and https://hibernate.atlassian.net/browse/HSEARCH-3429.

Merged, thanks!

PR #1813, opened 2018-11-20
HSEARCH-3422 Use _doc to implement index order sorting on Elasticsearch

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-3422


yrodiere:
Merged, thanks!

PR #1814, opened 2018-11-20
HSEARCH-3420 Make the build fully compatible with JDK11

yrodiere:
Multiple tickets:
* https://hibernate.atlassian.net//browse/HSEARCH-3238
* https://hibernate.atlassian.net//browse/HSEARCH-3420

Supersedes #1792 .

I have backports to 5.11 ready in one of my branches (https://github.com/yrodiere/hibernate-search/commits/tracking-5.11), so please don't bother with backporting this.


yrodiere:
I tested a full build yesterday on my branch, but just to be sure, here is a full build of this PR (including JDK11): http://ci.hibernate.org/job/hibernate-search/job/PR-1814/

gsmet:
Some tests are failing: http://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-1814/2/pipeline#step-30-log-950 .

Jenkinsfile
```diff
@@ -172,8 +172,7 @@ stage('Configure') {
 					// See http://www.oracle.com/technetwork/java/javase/eol-135779.html
 					new JdkITEnvironment(version: '8', tool: 'Oracle JDK 8', status: ITEnvironmentStatus.USED_IN_DEFAULT_BUILD),
 					new JdkITEnvironment(version: '10', tool: 'Oracle JDK 10.0.1', status: ITEnvironmentStatus.SUPPORTED),
```
gsmet: I think we can remove JDK 10 then.

Jenkinsfile
```diff
@@ -172,8 +172,7 @@ stage('Configure') {
...
-					// TODO HSEARCH-3238 (and maybe more) to support JDK11 builds
-					new JdkITEnvironment(version: '11', tool: 'OpenJDK 11 Latest', status: ITEnvironmentStatus.EXPERIMENTAL)
+					new JdkITEnvironment(version: '11', tool: 'OpenJDK 11 Latest', status: ITEnvironmentStatus.SUPPORTED)
```
gsmet: Does it run in parallel when doing a full build or is there a way to only trigger a JDK 11 build if the JDK 8 build passes?

yrodiere: It's only triggered if the default build passes, and the default build uses JDK8.

Jenkinsfile
yrodiere: Right, will do.

yrodiere:
Fixed the issues and launched a new build: http://ci.hibernate.org/job/hibernate-search/job/PR-1814/4/

PR #1819, opened 2018-11-22
HSEARCH-3268 Search 6 groundwork - Restore support for the massindexer

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3268


mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/impl/PojoWorkPlanImpl.java
```diff
@@ -120,6 +122,21 @@ public void prepare() {
 		}
 	}
 
+	@Override
+	public <E> Set<Class<? extends E>> getIndexedTypesPolymorphic(Class<E> entityType) {
```
yrodiere: This has nothing to do with the work plan. Rule of thumb: work plans should only provide operations related to... working on the index.

The `getIndexedTypesPolymorphic` method should be exposed in `PojoMappingDelegate` and `HibernateOrmMapping`. You can already see metadata-querying methods in there: `isWorkable`, `isIndexable`, ...

`PojoMappingDelegateImpl` would have the actual implementation, and `HibernateOrmMappingImpl` would just delegate.

In `MassIndexerImpl`, this should not be a problem: you have a `mapping` attribute, so you can just use it instead of the work plan.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/batchindexing/impl/BatchCoordinator.java
```diff
@@ -0,0 +1,174 @@
...
+		this.objectLoadingBatchSize = objectLoadingBatchSize;
+		this.optimizeAtEnd = optimizeAtEnd;
+		this.purgeAtStart = purgeAtStart;
+		this.optimizeAfterPurge = optimizeAfterPurge;
+		this.objectsLimit = objectsLimit;
+		this.endAllSignal = new CountDownLatch( rootEntities.size() );
+	}
+
+	@Override
+	public void runWithErrorHandler() {
```
yrodiere: Good call removing the `BatchBackend` here. It doesn't make much sense in Search 6.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/batchindexing/impl/BatchCoordinator.java
```diff
@@ -0,0 +1,174 @@
...
+			//TODO: Implement a monitor, then close it here
+		}
+	}
+
+	/**
+	 * Will spawn a thread for each type in rootEntities, they will all re-join
+	 * on endAllSignal when finished.
+	 *
+	 * @throws InterruptedException if interrupted while waiting for endAllSignal.
+	 * @throws ExecutionException
```
yrodiere: `ExecutionException` is not thrown anymore.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/batchindexing/impl/BatchIndexingWorkspace.java
```diff
@@ -0,0 +1,174 @@
...
+								long objectsLimit,
+								int idFetchSize,
+								Integer transactionTimeout,
+								String tenantId) {
+		this.indexedType = type;
+		this.idFetchSize = idFetchSize;
+		this.transactionTimeout = transactionTimeout;
+		this.tenantId = tenantId;
+
+		HibernateOrmSearchManager searchManager = mapping.createSearchManager( sessionFactory.createEntityManager() );
```
yrodiere: This is awful, you're creating an entity manager but never closing it. You're effectively leaking JDBC connections here. Don't do that.

I don't see why you create a `WorkPlan` here. You should remove it. Here are the uses of this work plan I found:

* You pass it to `BatchTransactionalContext`, but it doesn't use it
* You pass it to `IdentifierConsumerDocumentProducer`, but that class is supposed to be thread safe, and the work plan is not. As we already discussed, the work plan in `IdentifierConsumerDocumentProducer` should be created in `org.hibernate.search.mapper.orm.batchindexing.impl.IdentifierConsumerDocumentProducer#run` (using the session you create in the same method) and passed as a parameter to other methods.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/batchindexing/impl/DefaultMassIndexerFactory.java
```diff
@@ -0,0 +1,33 @@
...
+import org.hibernate.engine.spi.SessionFactoryImplementor;
+import org.hibernate.search.mapper.orm.MassIndexer;
+import org.hibernate.search.mapper.orm.batchindexing.spi.MassIndexerFactory;
+
+/**
+ * The {@link MassIndexer} implementation used when none is specified in the configuration.
+ *
+ * @author Davide D'Alto
+ */
+public class DefaultMassIndexerFactory implements MassIndexerFactory {
```
yrodiere: Hmm, I didn't know about that.

Mass indexer factories don't make much sense until we provide a way for user to access internal metadata: what could they do with a session factory and a list of entity types?

I think we should remove the whole concept of "mass indexer factory" for now and just use `new MassIndexerImpl()` wherever this was used. We'll reintroduce factories later if necessary; I suspect it's only needed for OGM, so we'll have a look again when we start integrating with OGM.

Could do that in a separate commit, please?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/batchindexing/impl/IdentifierConsumerDocumentProducer.java
```diff
@@ -0,0 +1,232 @@
...
+			String tenantId, PojoWorkPlan workPlan) {
+		this.source = fromIdentifierListToEntities;
+		this.sessionFactory = sessionFactory;
+		this.cacheMode = cacheMode;
+		this.type = indexedType;
+		this.idName = idName;
+		this.producerEndSignal = producerEndSignal;
+		this.transactionTimeout = transactionTimeout;
+		this.tenantId = tenantId;
+		this.workPlan = workPlan;
```
yrodiere: As I said somewhere else, there shoudn't be a work plan here: instances of this class is supposed to be used concurrently, and work plan is not thread safe. The work plan should be created in `run()` and passed to other methods.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/batchindexing/impl/IdentifierConsumerDocumentProducer.java
```diff
@@ -0,0 +1,232 @@
...
+					// temporary re-throw the exception
+					throw new RuntimeException( errorMsg, e );
+				}
+			}
+		}
+
+		workPlan.execute().get();
+	}
+
+	@SuppressWarnings("unchecked")
```
yrodiere: This `@SuppressWarnings("unchecked")` is no longer necessary.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/batchindexing/impl/MassIndexerImpl.java
```diff
@@ -0,0 +1,235 @@
...
+	//TODO: restore the MassIndexerProgressMonitor
+	private int idFetchSize = 100; //reasonable default as we only load IDs
+	private String tenantIdentifier;
+	private Integer idLoadingTransactionTimeout;
+
+	protected MassIndexerImpl(SessionFactoryImplementor sessionFactory, Class<?>... entities) {
+		this.sessionFactory = sessionFactory;
+		this.mapping = sessionFactory.getServiceRegistry().getService( HibernateSearchContextService.class ).getMapping();
+		this.workPlan = mapping.createSearchManager( sessionFactory.createEntityManager() ).createWorkPlan();
+		this.rootEntities = toRootEntities( entities );
```
yrodiere: Please don't remove any code without at least leaving a TODO for later. You removed this:

```
		if ( extendedIntegrator.isJMXEnabled() ) {
			monitor = new JMXRegistrar.IndexingProgressMonitor();
		}
		else {
			monitor = new SimpleIndexingProgressMonitor();
		}
```

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/batchindexing/impl/MassIndexerImpl.java
```diff
@@ -0,0 +1,235 @@
...
+
+	/**
+	 * From the set of classes a new set is built containing all indexed
+	 * subclasses, but removing then all subtypes of indexed entities.
+	 *
+	 * @param selection
+	 *
+	 * @return a new set of entities
+	 */
+	private Set<Class<?>> toRootEntities(Class<?>... selection) {
```
yrodiere: You made the method non-static, I don't really know why.

When you call a method in a constructor, and it doesn't need to be overridden, I find it's better to make the method static and pass all the necessary parameters. That way, you're sure it won't use any field that is initialized later in the constructor.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/batchindexing/spi/MassIndexerWithTenant.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+public interface MassIndexerWithTenant extends MassIndexer {
+
+	/**
+	 * Set the tenant that is associated to this {@link MassIndexer}.
+	 *
+	 * @param tenantIdentifier the identifier of the tenant associated this {@link MassIndexer}
+	 * @return <tt>this</tt> for method chaining
+	 */
+	MassIndexerWithTenant tenantIdentifier(String tenantIdentifier);
```
yrodiere: I think you can (in a separate commit) remove this interface and pass the tenant identifier directly to the `MassIndexerImpl` constructor. We don't need two separate interfaces.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/MassIndexer.java
```diff
@@ -0,0 +1,129 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.mapper.orm;
```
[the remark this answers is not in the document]
fax4ever: About the other classes, for instance `MassIndexerImpl`. Could be OK to use `org.hibernate.search.massindexing.impl` as its package?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/batchindexing/impl/MassIndexerImpl.java
```diff
@@ -0,0 +1,235 @@
...
+	private boolean optimizeAfterPurge = true;
+	//TODO: restore the MassIndexerProgressMonitor
+	private int idFetchSize = 100; //reasonable default as we only load IDs
+	private String tenantIdentifier;
+	private Integer idLoadingTransactionTimeout;
+
+	protected MassIndexerImpl(SessionFactoryImplementor sessionFactory, Class<?>... entities) {
+		this.sessionFactory = sessionFactory;
+		this.mapping = sessionFactory.getServiceRegistry().getService( HibernateSearchContextService.class ).getMapping();
+		this.workPlan = mapping.createSearchManager( sessionFactory.createEntityManager() ).createWorkPlan();
```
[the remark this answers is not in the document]
fax4ever: Sure, it is not used neither closed. I'll remove it.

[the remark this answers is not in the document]
fax4ever: The idea  was leaving it there as a placeholder. But, it would have been better using a comment.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/MassIndexer.java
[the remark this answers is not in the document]
yrodiere: Yes, as a general rule, try to put everything you can in `.impl` packages. The CI build will fail and tell you if you leak an implementation type or SPI type from an API type, so don't be afraid to try :)

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/batchindexing/impl/DefaultMassIndexerFactory.java
fax4ever: Sure

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/batchindexing/spi/MassIndexerWithTenant.java
fax4ever: Good idea

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/spi/MassIndexer.java
```diff
@@ -4,7 +4,7 @@
  * License: GNU Lesser General Public License (LGPL), version 2.1 or later
  * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
  */
-package org.hibernate.search.mapper.orm;
+package org.hibernate.search.mapper.orm.massindexing.spi;
```
yrodiere: Why SPI? Users have to use the mass indexer, this particular interface should be API.

The build should have failed because this class is referenced from API. I will investigate. In the meantime, please move this interface to API.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/IdentifierConsumerDocumentProducer.java
```diff
@@ -190,37 +191,38 @@ private void rollbackTransaction(SessionImplementor session) throws Exception {
...
-					);
-
-					// TODO: implements exception handler
-					// errorHandler.handleException( errorMsg, e );
-					// temporary re-throw the exception
-					throw new RuntimeException( errorMsg, e );
-				}
+
+		//TODO: switch to specific batch-optimized workPlan
+		PojoWorkPlan workPlan = mapping.createSearchManager( session ).createWorkPlan();
```
yrodiere: No need to change it, since you will have to switch to a different kind of work plan later, but just for your information: work plans are reusable. You could have created it in the run() method and reused it for each `List` of entities. Though you may have wanted a `clear()` method to call in case something goes wrong.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/IdentifierConsumerDocumentProducer.java
```diff
@@ -190,37 +191,38 @@ private void rollbackTransaction(SessionImplementor session) throws Exception {
...
+
+		log.tracef( "received a list of objects to index: %s", entities );
+		for ( Object object : entities ) {
+			try {
+				index( workPlan, object );
+				// TODO: implements monitor
+				// monitor.documentsBuilt( 1 );
+			}
+			catch (RuntimeException e) {
```
yrodiere: Note for later, when you implement the other "work plan": we really need to process the entity in this `index()` method. Currently that task is delayed to the execute() call below, so we're unlikely to ever get an exception here.

fax4ever: Sure

integrationtest/showcase/library/src/test/java/org/hibernate/search/integrationtest/showcase/ManualIndexingIT.java
```diff
@@ -0,0 +1,191 @@
+/*
+ * Hibernate Search, full-text search for your domain model
```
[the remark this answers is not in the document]
fax4ever: Sorry for this oversight!

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/session/spi/HibernateOrmSearchManager.java
```diff
@@ -27,4 +28,6 @@
 	 */
 	PojoWorkPlan createWorkPlan();
 
+	SessionWorkExecutor createSessionWorkExecutor();
```
yrodiere: Seeing this makes me think the naming is a bit inconsistent. Do you think you could rename `MappingWorkExecutor` to `PojoMappingWorkExecutor` and `SessionWorkExecutor` to `PojoSessionWorkExecutor`? No need to rename the methods (`createSessionWorkExecutor` etc.) or the variables, just renaming the type will be enough.

fax4ever:
Thank @yrodiere. Pushed the fixed-up commits.

engine/src/main/java/org/hibernate/search/engine/backend/index/spi/IndexingMonitor.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.engine.backend.index.spi;
+
+public interface IndexingMonitor {
```
yrodiere: If possible, I'd prefer this interface to be in the ORM module, next to the mass indexer. Maybe also rename it `MassIndexingMonitor`?
As far as I can tell, it should be possible, since all these methods will be called directly from the mass indexer.

Also, this should be API, not SPI: users are expected to implement it for their own purposes (displaying a progress bar in their web application, for example).

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/work/impl/PojoSessionWorkExecutorImpl.java
```diff
@@ -28,7 +31,7 @@
 	private final PojoSessionContextImplementor sessionContext;
 	private final PojoRuntimeIntrospector introspector;
 
-	private Map<Class<?>, IndexWorkExecutor> indexWorkExecutor;
+	private Map<Class<?>, PojoTypeWorkExecutor> indexWorkExecutor = new HashMap<>();
```
yrodiere: Please don't use raw types. It should be `PojoTypeWorkExecutor<?, ?, ?>`.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/impl/PojoTypeWorkExecutor.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Supplier;
+
+import org.hibernate.search.engine.backend.document.DocumentElement;
+import org.hibernate.search.engine.backend.index.spi.DocumentReferenceProvider;
+import org.hibernate.search.engine.backend.index.spi.IndexWorkExecutor;
+import org.hibernate.search.mapper.pojo.session.context.spi.PojoSessionContextImplementor;
+
+public class PojoTypeWorkExecutor<I, E, D extends DocumentElement> {
```
yrodiere: Good call. I don't think we discussed it, but adding a intermediary "type executor" here can definitely help.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/work/impl/PojoSessionWorkExecutorImpl.java
```diff
@@ -39,13 +42,34 @@ public PojoSessionWorkExecutorImpl(PojoIndexedTypeManagerContainer indexedTypeMa
...
+	public CompletableFuture<?> add(Object entity) {
+		return add( null, entity );
 	}
 
 	@Override
-	public CompletableFuture<?> add(Object entity) {
-		return add( null, entity );
+	public CompletableFuture<?> add(Object id, Object entity) {
+		Class<?> clazz = introspector.getClass( entity );
+		PojoTypeWorkExecutor indexWork = this.indexWorkExecutor.get( clazz );
```
yrodiere: Maybe "typeExecutor" rather than "indexWork"?

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/work/impl/PojoSessionWorkExecutorImpl.java
```diff
@@ -39,13 +42,34 @@ public PojoSessionWorkExecutorImpl(PojoIndexedTypeManagerContainer indexedTypeMa
...
+		if ( indexWork == null ) {
+			indexWork = createIndexWork( clazz );
+			indexWorkExecutor.put( clazz, indexWork );
+		}
+
+		return indexWork.add( id, entity );
+	}
+
+	private PojoTypeWorkExecutor createIndexWork(Class<?> clazz) {
```
yrodiere: Same here, maybe "createTypeExecutor"?

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/work/impl/PojoSessionWorkExecutorImpl.java
```diff
@@ -39,13 +42,34 @@ public PojoSessionWorkExecutorImpl(PojoIndexedTypeManagerContainer indexedTypeMa
...
+		Optional<? extends PojoIndexedTypeManager<?, ?, ?>> exactClass = indexedTypeManagers.getByExactClass( clazz );
+		if ( exactClass.isPresent() ) {
+			return exactClass.get().createWorkExecutor( sessionContext );
+		}
+
+		// I don't know if it makes sense to seek on the containedTypeManagers too
```
yrodiere: It does not, you can remove this. Maybe use a different error message, too.

fax4ever:
Thanks @yrodiere! All very clear, I'll do the changes here in a subsequent fixup commit.
I was thinking about the fact that a `CompletableFuture` seems not to be 'interruptible', according to the Java APIs. So maybe that could be a problem in order to handle correctly a thread interruption signal on the coordinator. What do you think about?

yrodiere:
>  I was thinking about the fact that a CompletableFuture seems not to be 'interruptible', according to the Java APIs. So maybe that could be a problem in order to handle correctly a thread interruption signal on the coordinator. What do you think about?

I don't see what you are talking about. Let's talk about it on gitter.

engine/src/main/java/org/hibernate/search/engine/backend/index/spi/IndexingMonitor.java
fax4ever: OK Even if I think that the monitor feature could be used for a general purpose tracing, very common these days. No problem to confine the monitor to the ORM module at the moment, we can refactor it later on.

fax4ever:
Hi @yrodiere! 
Much better change the backend mock test APIs, offering the method:
> backendMock.expectWorksAnyOrder(...)

So that the `out of order` property is defined in the test itself. 
I made some tests and all works as expected.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/IdentifierConsumerDocumentProducer.java
```diff
@@ -196,39 +197,85 @@ private void indexAllQueue(SessionImplementor session, List<?> entities) throws
...
-				// temporary re-throw the exception
-				throw new RuntimeException( errorMsg, e );
+		for ( int i = 0; i < entities.size(); i++ ) {
+			tasks[i] = index( workExecutor, entities.get( i ) );
+		}
+
+		try {
+			CompletableFuture.allOf( tasks ).join();
+		}
+		catch (RuntimeException e) {
```
yrodiere: This is not how you're supposed to use completable futures, especially not when you intend to (mostly) ignore and log exceptions.

A better implementation would be to add something like this, just after the `tasks[i] = index( workExecutor, entities.get( i ) );` (inside the loop):

```
tasks[i] = index( workExecutor, entities.get( i ) );
tasks[i].exceptionally( exception -> handleException( entities.get( i ), exception ) );
```


Then you can safely ignore the exception when you join, since it's handled on a per-work basis:

```
CompletableFuture.allOf( tasks ).exceptionally( exception -> null ).join();
```

Ideally we should use an instances of nested classes instead of lambdas, but you get the idea.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/IdentifierConsumerDocumentProducer.java
```diff
@@ -196,39 +197,85 @@ private void indexAllQueue(SessionImplementor session, List<?> entities) throws
...
+			if ( taskWithExceptionIndex == null ) {
+				throw new RuntimeException( "This row should be never reached, according to CompletableFuture#allOf API", e );
 			}
+
+			// handling error on executing task phase
+			handleException( entities.get( taskWithExceptionIndex ), e );
 		}
 
-		workPlan.execute().get();
+		Integer completedTasks = this.countCompletedTasks( tasks );
```
yrodiere: Good point. In Search 5 we used to increment the monitor even if a work failed, but we probably should not.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/IdentifierConsumerDocumentProducer.java
```diff
@@ -196,39 +197,85 @@ private void indexAllQueue(SessionImplementor session, List<?> entities) throws
...
-	private void index(PojoWorkPlan workPlan, Object entity) throws InterruptedException {
+	private CompletableFuture<?> index(PojoSessionWorkExecutor workExecutor, Object entity) throws InterruptedException {
 		// abort if the thread has been interrupted while not in wait(), I/O or similar which themselves would have
 		// raised the InterruptedException
 		if ( Thread.currentThread().isInterrupted() ) {
 			throw new InterruptedException();
 		}
 
-		workPlan.add( entity );
+		// after that a CompletableFuture may not be interrupted anymore, according to the current Java APIs
```
yrodiere: I fail to see how this is relevant.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/IdentifierConsumerDocumentProducer.java
```diff
@@ -196,39 +197,85 @@ private void indexAllQueue(SessionImplementor session, List<?> entities) throws
 			return;
 		}
 
-		// TODO use 'SessionWorkExecutor workExecutor = mapping.createSearchManager( session ).createSessionWorkExecutor();' instead
-		PojoWorkPlan workPlan = mapping.createSearchManager( session ).createWorkPlan();
+		// TODO <<monitor>> indexingMonitor.entitiesLoaded( entities.size() );
+		PojoSessionWorkExecutor workExecutor = mapping.createSearchManager( session ).createSessionWorkExecutor();
+		CompletableFuture<?>[] tasks = new CompletableFuture<?>[entities.size()];
```
yrodiere: Please call these what they are: futures. A "task" generally provides the code to execute, here these are just futures that give us information about the *result* of a task.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/IdentifierConsumerDocumentProducer.java
```diff
@@ -196,39 +197,85 @@ private void indexAllQueue(SessionImplementor session, List<?> entities) throws
...
-		workPlan.add( entity );
+		// after that a CompletableFuture may not be interrupted anymore, according to the current Java APIs
+		CompletableFuture<?> task = null;
+		try {
+			task = workExecutor.add( entity );
+		}
+		catch (Throwable e) {
+			// handling error on creating task phase
+			handleException( entity, e );
+		}
```
yrodiere: Alternatively you can use the pattern we use elsewhere:

```
task = Futures.create( () -> workExecutor.add( entity ) );
```

Then any exception occurring while creating the future will simply result in Futures.create() returning a future that completed exceptionally. Thus you can get rid of the try/catch.

Yes it's a lamba, but in this case, I think it's worth it.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/IdentifierConsumerDocumentProducer.java
fax4ever: Much better as you pointed out, thanks!

fax4ever:
@yrodiere thanks the many reviews. I've just pushed two commits. I think that many others have to come before we finish this issue.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/massindexing/IndexWorkExecutorIT.java
```diff
@@ -0,0 +1,155 @@
...
+	public void before() {
+		setupHelper.withDefaultConfiguration()
+				.withIndex(
+						"MappedType", INDEX_NAME,
+						ctx -> this.indexAccessors = new IndexAccessors( ctx.getSchemaElement() ),
+						indexManager -> this.indexManager = indexManager
+				)
+				.setup();
+
+		initData();
```
yrodiere: What you are testing is the data initialization itself. This should be in the test method.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/massindexing/IndexWorkExecutorIT.java
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	@Test
+	public void findBulk() {
+		SearchQuery<DocumentReference> query = queryWithPredicate( f -> f.matchAll().toPredicate() );
```
yrodiere: Not sure we need anything more than this test, really. Checking that the index contains everything that you wanted to index should be enough. The various field types, predicates, etc. are already tested elsewhere.

More importantly, the point of the work executor is to index *lots* of documents. Here, by complicating your schema, you forced yourself to write the value for each document by hand, and thus you couldn't index more than a few documents.

I think you should use a dramatically simpler schema (just one String field) and index a few hundred documents instead of just five. Not that I expect anything to go wrong: it's just what this interface is for, so let's test it like it is supposed to be used.

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/rule/IndexWorkCall.java
```diff
@@ -16,16 +16,16 @@
...
 
-	enum Operation {
+	enum WorkPhase {
 		PREPARE,
 		EXECUTE
 	}
 
 	private final String indexName;
-	private final Operation operation;
+	private final WorkPhase operation;
```
yrodiere: Well that's nice, but maybe you could also remove the fields and variables accordingly :) And similarly change the error messages (see below).

fax4ever: +1

fax4ever:
@yrodiere rebasing on top of the master, squashing fixups

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/impl/ElasticsearchStubWorkFactory.java
```diff
@@ -103,7 +103,9 @@ public ElasticsearchStubWorkFactory(GsonProvider gsonProvider) {
 	}
 
 	@Override
-	public ElasticsearchWork<?> delete(URLEncodedString indexName, URLEncodedString typeName) {
+	public ElasticsearchWork<?> delete(URLEncodedString indexName, URLEncodedString typeName, String tenantId) {
```
yrodiere: Adding a comment so that we remember: this cannot be merged as is.

1. The implementation should change: a simple "delete" just won't work, in particular because the index, its schema and its settings must still be there after the purge (they may have been modified by the user, so we can't just restore them ourselves). A delete-by-query would be more appropriate.
2. As a consequence, The method should be named "purge" or "delete by query".
3. We should have tests for a purge in a multi-tenant environment, to be sure no data is lost.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/index/impl/IndexingBackendContext.java
```diff
@@ -66,4 +67,8 @@ Directory createDirectory(String indexName) throws IOException {
 		return new LuceneIndexWorkExecutor( workFactory, multiTenancyStrategy, orchestrator,
 				indexName, sessionContext );
 	}
+
+	public CompletableFuture<?> optimize(LuceneIndexWorkOrchestrator streamOrchestrator, String indexName) {
```
yrodiere: Ok, the fact you have to add delegate methods all over the place is concerning. There may be something wrong with what we planned.

Don't you think it would be better if renamed "IndexWorkExecutor" to "IndexDocumentWorkExecutor", and moved the optimize/purge/flush operations to a new "IndexWorkExecutor" interface? Then you could remove the various operations you added in `MappedIndexManager` (just leave a createIndexWorkExecutor method) and here (same, just leave a createIndexWorkExecutor method).

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/work/impl/LuceneStubWorkFactory.java
```diff
@@ -38,6 +38,11 @@ public LuceneStubWorkFactory(MultiTenancyStrategy multiTenancyStrategy) {
 		return multiTenancyStrategy.createDeleteEntryLuceneWork( indexName, tenantId, id );
 	}
 
+	@Override
+	public LuceneIndexWork<?> delete(String indexName, String tenantId) {
```
yrodiere: Maybe rename this "deleteAll"? It's a bit dangerous to name it as if it deleted only one entry.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/work/impl/LuceneTermBasedDeleteAllEntriesWork.java
```diff
@@ -0,0 +1,23 @@
...
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.lucene.work.impl;
+
+import java.io.IOException;
+
+import org.apache.lucene.index.IndexWriter;
+
+public class LuceneTermBasedDeleteAllEntriesWork extends AbstractLuceneDeleteAllEntriesWork {
```
yrodiere: I don't see how this class is "term-based". Please find a better name?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/work/impl/LuceneStubWorkFactory.java
fax4ever: This was the other option I had thought. I will change it.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/index/impl/IndexingBackendContext.java
fax4ever: Good idea, I'll do it in a subsequent commit, since it requires a little bit of redesign.

yrodiere: Sure this can definitely wait until you're done with everything else.

fax4ever: What do you think of either `IndexBatchWorkExecutor` or `IndexMassiveWorkExecutor` or `IndexBulkWorkExecutor`  as a name for the new `IndexWorkExecutor' interface?

yrodiere: What's wrong with `IndexWorkExecutor`/`IndexDocumentWorkExecutor`? That would make it clear that the first creates works scoped to the index, the second creates works scoped to the document.

If that doesn't work for you, then let's see the other names...

`Bulk` won't do because it means something different in Elasticsearch.
`Batch` won't do either, since it's not about batching: on the contrary, it's about executing one big work atomically.
`Massive` would work, I suppose, but it focuses on the fact these works are resource-intensive, while what's important is their scope.

`IndexGeneralWorkExecutor`/`IndexDocumentWorkExecutor`, maybe? `General` feels clumsy too, but at least it focuses on the scope.

fax4ever: OK

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/MassIndexerImpl.java
```diff
@@ -0,0 +1,250 @@
...
+	public MassIndexerImpl(SessionFactoryImplementor sessionFactory, String tenantIdentifier, Class<?>... entities) {
+		this.sessionFactory = sessionFactory;
+		this.tenantIdentifier = tenantIdentifier;
+
+		HibernateSearchContextService searchContextService = sessionFactory.getServiceRegistry().getService( HibernateSearchContextService.class );
+		this.mapping = searchContextService.getMapping();
+		this.rootEntities = toRootEntities( mapping, entities );
+
+		// TODO Maybe it is a little bit complex
+		ConfigurationService configurationService = sessionFactory.getServiceRegistry().getService( ConfigurationService.class );
```
fax4ever: Actually I've not found a better way to read the property. I don't like the one I implemented here.

fax4ever:
So far I think that we've implemented all the expected feature for the relative issue. May I open a followup issue for enabling  chunk processing?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/IdentifierConsumerDocumentProducer.java
```diff
@@ -0,0 +1,250 @@
...
+			}
+
+			transactionManager.begin();
+		}
+		else {
+			session.beginTransaction();
+		}
+	}
+
+	private void rollbackTransaction(SessionImplementor session) throws Exception {
```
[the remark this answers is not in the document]
fax4ever: it was so in v5, I'll change it.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/MassIndexerImpl.java
```diff
@@ -0,0 +1,250 @@
...
+			}
+		}
+
+		// check for not indexed entities
+		if ( !nonIndexedEntities.isEmpty() ) {
+			throw log.someTargetedEntityTypesNotIndexed( StringHelper.join( nonIndexedEntities, "," ) );
+		}
+
+		Set<Class<?>> cleaned = new HashSet<Class<?>>();
+		Set<Class<?>> toRemove = new HashSet<Class<?>>();
```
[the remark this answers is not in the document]
fax4ever: As above, it was so in v5, I'll change it.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/ElasticsearchJsonHelper.java
```diff
@@ -0,0 +1,28 @@
...
+package org.hibernate.search.backend.elasticsearch.client.impl;
+
+import com.google.gson.JsonObject;
+
+public class ElasticsearchJsonHelper {
+
+	private ElasticsearchJsonHelper() {
+	}
+
+	public static JsonObject muchAllHavingTenant(String tenantId) {
```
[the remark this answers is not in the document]
fax4ever: There is a TODO below: 

> // TODO filter by tenantId

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/BatchCoordinator.java
```diff
@@ -0,0 +1,166 @@
...
+	private final boolean optimizeAfterPurge;
+	private final CountDownLatch endAllSignal;
+	private final MassIndexerProgressMonitor monitor;
+	private final long objectsLimit;
+	private final int idFetchSize;
+	private final Integer transactionTimeout;
+	private final String tenantId;
+	private final List<Future<?>> indexingTasks = new ArrayList<>();
+
+	public BatchCoordinator(Set<Class<?>> rootEntities, SessionFactoryImplementor sessionFactory, HibernateOrmMapping mapping,
```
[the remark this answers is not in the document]
fax4ever: I agree, but it was taken from v5 code. @yrodiere shall I use a DTO for it?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/IdentifierProducer.java
```diff
@@ -0,0 +1,193 @@
...
+	 * @param sessionFactory the Hibernate SessionFactory to use to load entities
+	 * @param objectLoadingBatchSize affects mostly the next consumer: IdentifierConsumerEntityProducer
+	 * @param indexedType the entity type to be loaded
+	 * @param monitor
+	 * @param objectsLimit if not zero
+	 * @param errorHandler
+	 * @param idFetchSize the fetch size
+	 * @param tenantId the tenant identifier
+	 */
+	public IdentifierProducer(
```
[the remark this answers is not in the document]
fax4ever: as above

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/massindexing/IndexWorkExecutorIT.java
```diff
@@ -0,0 +1,103 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.integrationtest.backend.tck.massindexing;
```
yrodiere: From the point of view of the backend, "mass indexing" does not exist. It's purely a mapper thing.
Maybe move this to package `org.hibernate.search.integrationtest.backend.tck.work`?

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/massindexing/IndexWorkExecutorIT.java
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	@Test
+	public void runOptimizePurgeAndFlushInSequence() {
+		IndexWorkExecutor workExecutor = indexManager.createWorkExecutor();
+
+		workExecutor.optimize().join();
+		assertBookNumberIsEqualsTo( NUMBER_OF_BOOKS );
+
+		workExecutor.purge( null ).join();
```
yrodiere: You should probably also test this operation with a tenant ID, see if it properly only deletes documents for that tenant ID, and leave the other documents untouched.

You can add that test in `org.hibernate.search.integrationtest.backend.tck.MultiTenancyIT`. If you think it makes more sense to test it here, in a separate method with a separate setup, let's talk about it first.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/monitor/IndexingMonitor.java
```diff
@@ -0,0 +1,31 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.mapper.orm.massindexing.monitor;
```
yrodiere: It does not make sense to separate the interfaces if you put them both in the "mass indexing" package. Please merge them back into a single `MassIndexingMonitor` interface, like we had agreed to. We can always extract a superinterface later, if it turns out to make sense.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/BatchCoordinator.java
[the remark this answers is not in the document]
yrodiere: I suspect we'll end up completely changing these classes in the near future: like we discussed, some names are confusing, and at some point we would like to use a publisher/subscriber pattern if possible. When we do, it's likely we will have to change constructors.

So, I would say don't bother with this right now.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/BatchIndexingWorkspace.java
```diff
@@ -44,6 +45,9 @@
 	private final CountDownLatch producerEndSignal; //released when we stop adding Documents to Index
 	private final CountDownLatch endAllSignal; //released when we release all locks and IndexWriter
 
+	// progress monitor
```
yrodiere: This comment is not necessary.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/IdentifierConsumerDocumentProducer.java
```diff
@@ -214,7 +214,7 @@ private void indexAllQueue(SessionImplementor session, List<?> entities) throws
 		CompletableFuture.allOf( futures ).exceptionally( exception -> null ).join();
 
 		Integer completedFutures = this.countCompletedTasks( futures );
-		// TODO <<monitor>> indexingMonitor.documentAdded( completedFutures );
+		monitor.documentsAdded( completedFutures.longValue() );
```
yrodiere: Just thought about this:

1. The "countCompletedTasks" method is useless at the moment. It counts all the tasks that are completed, either successfully or in error. Since we called `CompletableFuture.allOf( futures ).exceptionally( exception -> null ).join();` just two lines before, we know that all the futures completed in some way. So you could just do `futures.length` or `entities.size()`, and remove the `countCompletedTasks` method.
2. I originally though the "countCompletedTasks" method was counting the futures that succeeded. That could be a nice improvement. But if we did that, then we should add a `documentsFailed` method to the monitor to also inform it that some documents could not be indexed properly.

So... Let's just keep it simple for now, use `entities.size()`. And maybe create a ticket if you want to improve the monitor to differenciate successfully added documents from failed ones?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/IdentifierProducer.java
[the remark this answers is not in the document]
yrodiere: Same answer.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/work/impl/PojoMappingWorkExecutorImpl.java
```diff
@@ -0,0 +1,75 @@
...
+
+		for ( Class<?> type : types ) {
+			// TODO Should we support concurrent accesses here?
+			if ( !workExecutors.containsKey( type ) ) {
+				workExecutors.put( type, createWorkExecutor( type ) );
+			}
+			futures[typeCounter++] = operation.apply( workExecutors.get( type ) );
+		}
+
+		// TODO use an << errorHandler >> here
```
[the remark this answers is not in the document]
fax4ever: Maybe it is better not to use an error handler for `MappingWorkExecutor` operations: optimize, purge and flush.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/monitor/impl/JMXRegistrar.java
```diff
@@ -0,0 +1,174 @@
...
+import org.hibernate.search.mapper.orm.massindexing.monitor.IndexingProgressMonitorMBean;
+import org.hibernate.search.mapper.orm.massindexing.monitor.MassIndexerProgressMonitor;
+import org.hibernate.search.util.impl.common.LoggerFactory;
+
+/**
+ * Helper class to register JMX beans.
+ *
+ * @author Hardy Ferentschik
+ */
+public final class JMXRegistrar {
```
yrodiere: This is completely out of scope, and this PR is taking long enough as it is. Adding back JMX support was never something we discussed.

Please remove this commit from the PR. You can keep it around somewhere if you want, but let's try to focus a bit?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/MassIndexerImpl.java
yrodiere: As mentioned, JMX support was never part of the scope. Please remove this commit.

As to configuration properties, they should be read at startup, and stored somewhere. This kind of runtime configuration parsing should not be necessary at all.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/IdentifierConsumerDocumentProducer.java
```diff
@@ -0,0 +1,250 @@
...
+			indexAllQueue( session, list );
+			session.clear();
+		}
+		finally {
+			// it's read-only, so no need to commit
+			rollbackTransaction( session );
+		}
+	}
+
+	private void beginTransaction(Session session) throws Exception {
```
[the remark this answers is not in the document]
fax4ever: In this case we should throw 2 different exception. Is it not even worse to do this:

> private void beginTransaction(Session session) throws SystemException, NotSupportedException {

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/error/ErrorHandler.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+package org.hibernate.search.mapper.orm.massindexing.error;
+
+/**
+ * Implementations handle errors and exceptions occurring during indexing.
+ *
+ * @author Amin Mohammed-Coleman
+ * @since 3.2
+ */
+public interface ErrorHandler {
```
yrodiere: Again, this is out of scope. We even explicitly discussed it, and I told you we have a separate ticket to add back support for error handling. Please remove this commit from this PR. You can keep it somewhere for when we address error handlers, but right now it's not necessary.

Also, you took the liberty to move the error handler to the mass-indexing package. It's not specific to mass indexing at all; in fact, in Search 5 it's used for about every background work.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/IdentifierConsumerDocumentProducer.java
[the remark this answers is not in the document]
yrodiere: Sure.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/ElasticsearchJsonHelper.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.elasticsearch.client.impl;
```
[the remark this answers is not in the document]
fax4ever: You're right, it makes definitely no sense to extract an helper here.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/massindexing/IndexWorkExecutorIT.java
fax4ever: Do you mean having a test with 2 different tenants and verify that the purge will affect only one of them?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/monitor/IndexingMonitor.java
fax4ever: I kept both in order to preserve the v5 authoring. I'll merge them in a subsequent (not fixup) commit.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/IdentifierConsumerDocumentProducer.java
fax4ever: Yes, I think we should create a follow-up ticket for that.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/monitor/impl/JMXRegistrar.java
fax4ever: OK I will keep them in some patch files.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/error/ErrorHandler.java
fax4ever: OK. I ll preserve this commit too. I think we can refactor the name package when we work on the relative ticket.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/massindexing/IndexWorkExecutorIT.java
yrodiere: Yes.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/BatchCoordinator.java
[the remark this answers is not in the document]
fax4ever: OK thank you

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/monitor/IndexingMonitor.java
yrodiere: Fair enough.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/index/impl/ElasticsearchIndexWorkExecutor.java
```diff
@@ -40,11 +43,27 @@ public ElasticsearchIndexWorkExecutor(ElasticsearchWorkFactory workFactory, Mult
 	@Override
 	public CompletableFuture<?> purge(String tenantId) {
 		multiTenancyStrategy.checkTenantId( tenantId, eventContext );
-		return orchestrator.submit( workFactory.deleteByQuery( indexName, tenantId ) );
+		JsonObject document = ( tenantId == null ) ? matchAll() : matchAllByTenant( tenantId );
```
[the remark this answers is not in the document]
fax4ever: I think the collaboration pattern you described makes sense. I will apply it.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/work/IndexWorkExecutorIT.java
```diff
@@ -60,24 +65,31 @@ public void before() {
 				.withMultiTenancy()
 				.setup();
 
-		createBookIndexes();
-		assertBookNumberIsEqualsTo( NUMBER_OF_BOOKS );
+		createBookIndexes( tenant1SessionContext );
```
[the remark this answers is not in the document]
fax4ever: I think I got it. I will do that

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/index/impl/ElasticsearchIndexWorkExecutor.java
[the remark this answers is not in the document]
fax4ever: What do you think instead to delegate only the payload/document construction to the `multiTenancyStrategy` instance here and not the whole work?
We could reuse the same request structure, with the exception of the payload.

fax4ever:
I've just pushed some fixup commits, relative to the last review. I'm starting now to update the manual and the result will be in a subsequent commit.
At the end I choose to delegate to multiTenancyStrategy only the query decoration, using: 

> multiTenancyStrategy.decorateJsonQuery( matchAll, tenantId ) )

The reason is not just the reuse, but most of all the fact that the resulting code is cleaner in this way.

documentation/src/test/java/org/hibernate/search/documentation/gettingstarted/withhsearch/withoutanalysis/GettingStartedWithoutAnalysisIT.java
```diff
@@ -79,6 +84,24 @@ public void test() {
 			bookIdHolder.set( book.getId() );
 		} );
 
+		OrmUtils.withinJPATransaction( entityManagerFactory, entityManager -> {
+			// tag::manual-index[]
+			// Not shown: get the entity manager and open a transaction
+			FullTextEntityManager fullTextEntityManager = Search.getFullTextEntityManager( entityManager ); // <1>
+
+			MassIndexer indexer = fullTextEntityManager.createIndexer( Book.class ); // <2>
```
yrodiere: I think we should at least hint at the fact that there are options that users can fine-tune. Maybe add a call to `threadsToLoadObjects` and a note saying what it does, and mentioning there are other options? You can also redirect to the dedicated section for more details: `<<manual-index-changes>>`

documentation/src/main/asciidoc/getting-started.asciidoc
```diff
@@ -245,10 +245,19 @@ include::{sourcedir}/org/hibernate/search/documentation/gettingstarted/withhsear
 
 However, keep in mind that data already present in your database when you add the Hibernate Search integration
 is unknown to Hibernate Search, and thus has to be indexed through a batch process.
+Use the mass indexer api to do that, as shown in the following code:
```
yrodiere: Maybe "To that end, you can use the mass indexer API, as shown [...]"? There are (or at least will be) other options.

yrodiere:
I rebased and squashed some commits together where it made sense. Some I would have preferred to keep separated, but due to the various fix-up commits it was just easier to squash them (otherwise conflict resolution would have been hell).

fax4ever:
Thanks

yrodiere:
Solved a few bugs reported by Sonar. Waiting for CI.

yrodiere:
Merged. Thanks a lot!

PR #1820, opened 2018-11-22
HSEARCH-3410 + HSEARCH-3237 Upgrade to ORM 5.4.0.CR1 + HCANN 5.1.0.Final and remove direct JAXB dependencies

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3410
https://hibernate.atlassian.net//browse/HSEARCH-3376
https://hibernate.atlassian.net//browse/HSEARCH-3237

I grouped the ORM upgrade and the removal of direct JAXB dependencies because the ORM upgrade was the last thing we were waiting for to remove the JAXB dependencies.


yrodiere:
Please hold. This breaks the JDK10 build.

pom.xml
```diff
@@ -238,12 +238,9 @@
...
 
-        <!--
-            Using JAXB 2.3.0 as 2.2.11 seems to be broken in more than a few ways.
-            In particular version 2.2.11 of the impl depends on version 2.2.12-b140109.1041 of the API.
-         -->
-        <version.javax.xml.bind.jaxb-api>2.3.0</version.javax.xml.bind.jaxb-api>
-        <version.org.glassfish.jaxb>2.3.0.1</version.org.glassfish.jaxb>
+        <!-- This should be aligned with the version used in ORM -->
+        <!-- HSEARCH-3237: this should be removed when we remove the dependency from the Elasticsearch module to JAXB -->
+        <version.javax.xml.bind.jaxb-api>2.2.12-b140109.1041</version.javax.xml.bind.jaxb-api>
```
gsmet: Didn't you upgrade to 2.3 in ORM? Or you plan to upgrade it when moving to CR2?

elasticsearch/pom.xml
```diff
@@ -183,6 +183,7 @@
                 <jdk>9</jdk>
             </activation>
             <dependencies>
+                <!-- HSEARCH-3237: this should be removed when we remove the dependency from the Elasticsearch module to JAXB -->
```
gsmet: Why do we use JAXB in the Elasticsearch module? IIRC at some point we used it to format a date? But maybe we should get rid of it in 5.11 and be done with it?

pom.xml
yrodiere: I'll try to remember to upgrade when moving to CR2, but I remove this in one of the following commits anyway.
To be clear, as mentioned above, this PR is on hold precisely because I know CR1 has bugs that make our build (JDK10, and also JDK11) fail. So I won't merge it until CR2 is out and I updated the PR to target CR2 instead of CR1.

elasticsearch/pom.xml
yrodiere: > But maybe we should get rid of it in 5.11 and be done with it?

Yep, exactly what I do in the following commits.

yrodiere:
Rebased and modified to upgrade to 5.4.0.CR2. Just launched a full CI build. Let's see how it goes...

gsmet:
@yrodiere looks good to me

PR #1832, opened 2018-12-11
HSEARCH-3442 Search 6 groundwork - Properly handle missing backend / backend type in configuration properties

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3442

~Based on #1828 which should be merged first.~ => Done

@fax4ever As discussed, this could be a nice start if you want to review PRs on Search. Just mind the fact that the first commits are from another PRs and should not be considered.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/common/impl/IndexManagerBuildingStateHolder.java
+++ b/engine/src/main/java/org/hibernate/search/engine/common/impl/IndexManagerBuildingStateHolder.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.search.engine.common.impl;
 
+import java.lang.invoke.MethodHandles;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -14,6 +15,7 @@ import org.hibernate.search.engine.backend.document.model.dsl.spi.IndexSchemaRoo
 import org.hibernate.search.engine.cfg.spi.OptionalConfigurationProperty;
 import org.hibernate.search.engine.environment.bean.BeanReference;
 import org.hibernate.search.engine.environment.bean.BeanHolder;
+import org.hibernate.search.engine.logging.impl.Log;
 import org.hibernate.search.engine.mapper.mapping.spi.MappedIndexManager;
 import org.hibernate.search.engine.backend.index.spi.IndexManagerBuilder;
 import org.hibernate.search.engine.backend.index.spi.IndexManagerImplementor;
@@ -27,6 +29,7 @@ import org.hibernate.search.engine.mapper.mapping.building.impl.RootIndexModelBi
 import org.hibernate.search.engine.mapper.mapping.building.spi.IndexManagerBuildingState;
 import org.hibernate.search.engine.mapper.mapping.building.spi.IndexModelBindingContext;
 import org.hibernate.search.util.AssertionFailure;
+import org.hibernate.search.util.impl.common.LoggerFactory;
 import org.hibernate.search.util.impl.common.SuppressingCloser;
 
 
@@ -34,6 +37,8 @@ import org.hibernate.search.util.impl.common.SuppressingCloser;
  * @author Yoann Rodiere
  */
 class IndexManagerBuildingStateHolder {
+
+	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
 
 	private static final OptionalConfigurationProperty<String> INDEX_BACKEND_NAME =
 			ConfigurationProperty.forKey( "backend" ).asString().build();
@@ -58,8 +63,14 @@ class IndexManagerBuildingStateHolder {
 	public IndexManagerBuildingState<?> startBuilding(String indexName, boolean multiTenancyEnabled) {
 		ConfigurationPropertySource indexPropertySource = propertySource.withMask( "indexes." + indexName )
 				.withFallback( defaultIndexPropertySource );
-		// TODO HSEARCH-3442 more checks on the backend name (is non-null, non-empty) => use INDEX_BACKEND_NAME.getOrThrow?
-		String backendName = INDEX_BACKEND_NAME.get( indexPropertySource ).get();
+		String backendName = INDEX_BACKEND_NAME.getOrThrow(
+				indexPropertySource,
+				key -> log.indexBackendCannotBeNullOrEmpty(
+						indexName, key,
+						// Retrieve the resolved *default* key (*.indexes.default.backend) from the default source
+						INDEX_BACKEND_NAME.resolveOrRaw( defaultIndexPropertySource )
+				)
+		);
 		BackendBuildingState<?> backendBuildingstate =
 				backendBuildingStateByName.computeIfAbsent( backendName, this::createBackend );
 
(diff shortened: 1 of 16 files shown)


engine/src/main/java/org/hibernate/search/engine/common/impl/IndexManagerBuildingStateHolder.java
```diff
@@ -35,31 +38,39 @@
...
 
 	public IndexManagerBuildingState<?> startBuilding(String indexName, boolean multiTenancyEnabled) {
 		ConfigurationPropertySource indexPropertySource = propertySource.withMask( "indexes." + indexName )
 				.withFallback( defaultIndexPropertySource );
-		// TODO HSEARCH-3442 more checks on the backend name (is non-null, non-empty) => use INDEX_BACKEND_NAME.getOrThrow?
-		String backendName = INDEX_BACKEND_NAME.get( indexPropertySource ).get();
+		String backendName = INDEX_BACKEND_NAME.getOrThrow(
+				indexPropertySource,
+				key -> log.indexBackendCannotBeNullOrEmpty(
+						indexName, key, INDEX_BACKEND_NAME.resolveOrRaw( defaultIndexPropertySource )
```
fax4ever: Instead of use again `INDEX_BACKEND_NAME' inside the function as a closure variable, you could think use a bi-function such as: 
>  ( key, resolved ) ->  log.indexBackendCannotBeNullOrEmpty( indexName. key, resolved )

to gather the resolved value.

yrodiere: That will not work here.
To be clear, the key being passed to the lambda has already been resolved. Here, I add an explicit call to resolve because I want to resolve the key against a different configuration source: the default one.
Basically `key` here is resolved to something like `hibernate.search.indexes.myIndex.backend`, whereas `INDEX_BACKEND_NAME.resolveOrRaw( defaultIndexPropertySource )` returns `hibernate.search.indexes.default.backend`.
It's definitely not clear, though. I will add a comment.

fax4ever: Makes sense. Sorry. I didn't notice the use of two different property sources.

engine/src/main/java/org/hibernate/search/engine/logging/impl/Log.java
```diff
@@ -238,4 +238,16 @@ SearchException invalidBeanReferencePropertyValue(@FormatWith(ClassFormatter.cla
...
+
+	@Message(id = ID_OFFSET_2 + 49,
+			value = "Missing backend type for backend '%1$s'."
+					+ " Set the property '%2$s' to a supported value."
+	)
+	SearchException backendTypeCannotBeNullOrEmpty(String backendName, String key);
+
+	@Message(id = ID_OFFSET_2 + 50,
+			value = "Missing backend reference for index '%1$s'."
+					+ " Set the property '%2$s' to a supported value or set '%3$s' to set a default for all indexes."
```
gsmet: I think I would prefer `to set a **default** value for all indexes`

engine/src/test/java/org/hibernate/search/engine/common/impl/IndexManagerBuildingStateHolderTest.java
```diff
@@ -0,0 +1,121 @@
...
+		EasyMock.expect( configurationSourceMock.get( "indexes.indexName.backend" ) )
+				.andReturn( (Optional) Optional.of( "backendName" ) );
+		EasyMock.expect( configurationSourceMock.get( "backends.backendName.type" ) )
+				.andReturn( (Optional) Optional.of( "" ) );
+		EasyMock.expect( configurationSourceMock.resolve( "backends.backendName.type" ) )
+				.andReturn( Optional.of( keyPrefix + "backends.backendName.type" ) );
+		replayAll();
+		thrown.expect( SearchException.class );
+		thrown.expectMessage( "Missing backend type for backend 'backendName'" );
+		thrown.expectMessage( "Set the property 'somePrefix.backends.backendName.type' to a supported value" );
```
gsmet: If you change the message, it should also be changed here.

engine/src/test/java/org/hibernate/search/engine/common/impl/IndexManagerBuildingStateHolderTest.java
```diff
@@ -0,0 +1,121 @@
...
+				.andReturn( Optional.of( keyPrefix + "indexes.indexName.backend" ) );
+		EasyMock.expect( configurationSourceMock.get( "indexes.default.backend" ) )
+				.andReturn( (Optional) Optional.of( "" ) );
+		EasyMock.expect( configurationSourceMock.resolve( "indexes.default.backend" ) )
+				.andReturn( Optional.of( keyPrefix + "indexes.default.backend" ) );
+		replayAll();
+		thrown.expect( SearchException.class );
+		thrown.expectMessage( "Missing backend reference for index 'indexName'" );
+		thrown.expectMessage( "Set the property 'somePrefix.indexes.indexName.backend' to a supported value" );
+		thrown.expectMessage( "or set 'somePrefix.indexes.default.backend' to set a default for all indexes" );
```
gsmet: If you change the message, it should also be changed here.

engine/src/test/java/org/hibernate/search/engine/common/impl/IndexManagerBuildingStateHolderTest.java
```diff
@@ -0,0 +1,121 @@
...
+				.andReturn( Optional.of( keyPrefix + "indexes.indexName.backend" ) );
+		EasyMock.expect( configurationSourceMock.get( "indexes.default.backend" ) )
+				.andReturn( Optional.empty() );
+		EasyMock.expect( configurationSourceMock.resolve( "indexes.default.backend" ) )
+				.andReturn( Optional.of( keyPrefix + "indexes.default.backend" ) );
+		replayAll();
+		thrown.expect( SearchException.class );
+		thrown.expectMessage( "Missing backend reference for index 'indexName'" );
+		thrown.expectMessage( "Set the property 'somePrefix.indexes.indexName.backend' to a supported value" );
+		thrown.expectMessage( "or set 'somePrefix.indexes.default.backend' to set a default for all indexes" );
```
gsmet: If you change the message, it should also be changed here.

engine/src/test/java/org/hibernate/search/engine/cfg/spi/ConfigurationPropertyBeanReferenceTest.java
```diff
@@ -29,13 +30,7 @@
 public class ConfigurationPropertyBeanReferenceTest extends EasyMockSupport {
```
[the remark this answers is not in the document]
yrodiere: Right, done

engine/src/test/java/org/hibernate/search/engine/common/impl/IndexManagerBuildingStateHolderTest.java
```diff
@@ -0,0 +1,121 @@
...
+				.andReturn( Optional.of( keyPrefix + "backends.backendName.type" ) );
+		replayAll();
+		thrown.expect( SearchException.class );
+		thrown.expectMessage( "Missing backend type for backend 'backendName'" );
+		thrown.expectMessage( "Set the property 'somePrefix.backends.backendName.type' to a supported value" );
+		holder.startBuilding( "indexName", false );
+		verifyAll();
+	}
+
+	private abstract static class AbstractConfigurationPropertySource implements ConfigurationPropertySource {
```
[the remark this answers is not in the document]
yrodiere: Right. Removed.

engine/src/main/java/org/hibernate/search/engine/logging/impl/Log.java
yrodiere: Done.

PR #1830, opened 2018-12-10
HSEARCH 3416 id bridges

DavideD:
https://hibernate.atlassian.net/browse/HSEARCH-3416

I'm not sure I can dedicate some more time to this issue.
It should work but, if something is wrong, feel free to update it as you see fit.

Cheers

PR #1833, opened 2018-12-11
HSEARCH-3416 Use ID bridges from the mapper when building ID predicates

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3416

Supersedes #1830

No need to review: most of this is not my code and I reviewed it already. Just creating this to have the CI check it.


yrodiere:
Merged! Thanks @DavideD!

PR #1834, opened 2018-12-12
HSEARCH-3351 follow-up: use executeCount() in mass indexing tests + fix a potential bug

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3351

@fax4ever I noticed that just after merging... See anything wrong with it?

Waiting for CI, I didn't run all the tests locally.


fax4ever:
Good idea to add a test to probe different `firstResultIndex` values. I merged it, thanks!

PR #1837, opened 2018-12-13
HSEARCH-3449 Fix random test failures caused by improper mocking of ObjectLoader

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3449


util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/stub/mapper/StubMapperUtils.java
```diff
@@ -20,4 +31,28 @@ public static DocumentReferenceProvider referenceProvider(String identifier) {
 	public static DocumentReferenceProvider referenceProvider(String identifier, String routingKey) {
 		return new StubDocumentReferenceProvider( identifier, routingKey );
 	}
+
+	public static <R, O> void expectLoad(ObjectLoader<R, O> objectLoaderMock,
+			Consumer<LoadingDefinitionContext<R, O>> loadingDefinition) {
```
fax4ever: Maybe you can use a `Function<Void, Map<R, O>>` instead

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/stub/mapper/StubMapperUtils.java
```diff
@@ -20,4 +31,28 @@ public static DocumentReferenceProvider referenceProvider(String identifier) {
...
+		LoadingDefinitionContext<R, O> context = new LoadingDefinitionContext<>();
+		loadingDefinition.accept( context );
+
+		EasyMock.expect( objectLoaderMock.load(
+				EasyMockUtils.collectionAnyOrderMatcher( new ArrayList<>( context.loadingMap.keySet() ) )
```
fax4ever: I think there is no reason here to match the exact parameter, maybe you could use just `anyObject(List.class)`. 
So that we could remove the inner helper method.

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/stub/mapper/StubMapperUtils.java
```diff
@@ -20,4 +31,28 @@ public static DocumentReferenceProvider referenceProvider(String identifier) {
...
+		) )
+				.andAnswer(
+						() -> ( (List<R>) EasyMock.getCurrentArguments()[0] ).stream()
+						.map( context.loadingMap::get )
+						.collect( Collectors.toList() )
+				);
+	}
+
+	public static class LoadingDefinitionContext<R, O> {
```
fax4ever: Maybe we can define `expectLoad` method and `LoadingDefinitionContext` class in the same project where they are used: `hibernate-search-integrationtest-backend-tck`. So that we could remove the generics R and O, improving the readability.

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/stub/mapper/StubMapperUtils.java
yrodiere: I don't understand why? `Supplier<Map<R, O>>` could be an idea, but even then I would lose the fluent syntax (`.load(...).load(...)`)

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/stub/mapper/StubMapperUtils.java
yrodiere: The idea was to check the references are correctly transformed by the backend. If they are not, the load call will fail and tell us the parameters were not the expected ones.

I suppose something would ultimately fail too (because of null loaded objects) if the wrong parameters were passed to these methods, but the error message would probably be less clear?

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/stub/mapper/StubMapperUtils.java
yrodiere: I'll add a comment explaining what R and O are.

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/stub/mapper/StubMapperUtils.java
fax4ever: Yes a `Supplier<Map<R, O>>`, as wish it not critical at all. Code would look like:

```
StubMapperUtils.expectLoad(
   objectLoaderMock,
   () -> {
     HashMap<StubTransformedReference, StubLoadedObject> result = new HashMap<>();
     result.put( mainTransformedReference, mainLoadedObject );
     result.put( emptyTransformedReference, emptyLoadedObject );
     return result;
   }
 );
```
```
public static <R, O> void expectLoad(ObjectLoader<R, O> objectLoaderMock, Supplier<Map<R, O>> loadingDefinition) {
  EasyMock.expect( objectLoaderMock.load(
    EasyMockUtils.collectionAnyOrderMatcher( new ArrayList<>( loadingDefinition.get().keySet() ) )
  ) )
  .andAnswer(
    () -> ( (List<R>) EasyMock.getCurrentArguments()[0] ).stream()
      .map( loadingDefinition.get()::get )
    );
}
``

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/stub/mapper/StubMapperUtils.java
fax4ever: You're right, in this case it makes sense to check that the backend will produce the expected references. I guess the same.

yrodiere:
Merged. Thanks!

PR #1838, opened 2018-12-13
HSEARCH-3446 Search 6 groundwork - Clean up and factorize code around converters and type-specific features in backends

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3446

The first commit renames a few APIs that had become confusing.

The next commits change the internals, improving the situation regarding type-specific features in three ways:

* POJO value bridges are more throroughly tested, in particular their DSL/projection converter features.
* When adding a new supported type in the main (non-test) code, we should have less code to write, in particular for Lucene, and in particular for numerically-encoded types.
* When adding tests for that newly supported type, the classes to change should be more straightforward: basically we only need to implement a new `FieldTypeDescriptor` in the backend TCK and a new `PropertyTypeDescriptor` in the POJO mapper integration tests.


yrodiere:
Force-pushed an update to address issues reported by Sonar.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneStandardFieldCodec.java
```diff
@@ -0,0 +1,24 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.lucene.types.codec.impl;
+
+/**
+ * @param <F> The field type as declared on
```
gsmet: Unfinished sentence.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneStandardFieldCodec.java
```diff
@@ -0,0 +1,24 @@
...
+ * @param <E> The encoded type. For example, for a {@code LocalDate} field this will be {@code Long}
```
gsmet: Missing dot at the end.

yrodiere:
Addressed your comments and merged. Thanks!

PR #1841, opened 2018-12-14
Search 6 groundwork - Restore the full Elasticsearch work factory/orchestrators from Search 5

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3053

~~This is a very `preview` of what will be the final PR.~~ I prefer to submit these very first commits early in order to have a fast feedback. Since this issue seems to be quite extensive.
The phase 1 will consist in replace all the stub work items, maintaining the not very smart old orchestrator.


engine/src/main/java/org/hibernate/search/engine/search/SearchResult.java
```diff
@@ -17,6 +17,12 @@
 
 	List<T> getHits();
 
-	// TODO aggregations, facets, ...
+	// TODO implement => T getAggregations();
```
yrodiere: Please remove these changes. None of them makes sense in the `SearchResult` class from the engine. You can get rid of the code responsible for computing these values, too; it will be re-implemented differently.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchResultExtractorImpl.java
```diff
@@ -20,21 +20,25 @@
...
 
-	private static final JsonAccessor<Long> HITS_TOTAL_ACCESSOR =
-			HITS_ACCESSOR.property( "total" ).asLong();
+	private static final JsonAccessor<Long> HITS_TOTAL_ACCESSOR = HITS_ACCESSOR.property( "total" ).asLong();
+
+	private static final JsonAccessor<Integer> TOOK_ACCESSOR = JsonAccessor.root().property( "took" ).asInteger();
+
+	private static final JsonAccessor<Boolean> TIMED_OUT_ACCESSOR = JsonAccessor.root().property( "timed_out" ).asBoolean();
+
+	private static final JsonAccessor<String> SCROLL_ID_ACCESSOR = JsonAccessor.root().property( "_scroll_id" ).asString();
```
yrodiere: You can remove the added accessors, they are not useful. See also my comment on `SearchResult.java`

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/real/accessor/impl/ElasticsearchRequestSuccessAssessor.java
```diff
@@ -0,0 +1,33 @@
...
+
+	/**
+	 * Check the given response, throwing an exception if the reponse indicates a failure.
+	 * @param response The response, containing information about the outcome of the request.
+	 * @throws SearchException If the result is a failure.
+	 */
+	void checkSuccess(ElasticsearchResponse response) throws SearchException;
+
+	/**
+	 * Check the given bulk response item, return {@code true} if it is successful, {@code false} otherwise.
```
yrodiere: I think this comment is obsolete, could you please replace it with this?

```
Check the given bulk response item, throwing an exception if it indicates a failure.
```

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchResultExtractorImpl.java
```diff
@@ -58,6 +58,24 @@ public long getHitCount() {
...
+			@Override
+			public boolean getTimedOut() {
+				// TODO implement
+				return false;
+			}
+
+			@Override
+			public String getScrollId() {
+				// TODO implement
+				return null;
```
yrodiere: See, this does not make sense.

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/rule/SearchWorkCall.java
```diff
@@ -119,6 +119,21 @@ public long getHitCount() {
...
+		}
+
+		@Override
+		public boolean getTimedOut() {
+			return false;
+		}
+
+		@Override
+		public String getScrollId() {
+			return null;
```
yrodiere: Same here, this does not make any sense.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchBackendFactory.java
```diff
@@ -92,12 +94,13 @@
...
 			MultiTenancyStrategy multiTenancyStrategy = getMultiTenancyStrategy( name, propertySource );
 			ElasticsearchWorkFactory workFactory = new ElasticsearchStubWorkFactory( dialectSpecificGsonProvider, multiTenancyStrategy );
+			ElasticsearchWorkBuilderFactory workBuilderFactory = new ElasticsearchWorkBuilderFactoryImpl( dialectSpecificGsonProvider );
 
 			ElasticsearchAnalysisDefinitionRegistry analysisDefinitionRegistry =
 					getAnalysisDefinitionRegistry( backendContext, buildContext, propertySource );
 
 			return new ElasticsearchBackendImpl(
-					client, name, workFactory,
+					client, name, workFactory, workBuilderFactory,
```
yrodiere: I suppose this is temporary, because you want to create one commit per work and you want everything to still compile in the meantime?

Ultimately (at the end of this PR) the work factory (the old one) will disappear, right?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/builder/factory/impl/ElasticsearchWorkBuilderFactory.java
```diff
@@ -0,0 +1,78 @@
...
+import org.hibernate.search.backend.elasticsearch.work.builder.impl.OptimizeWorkBuilder;
+import org.hibernate.search.backend.elasticsearch.work.builder.impl.PutIndexSettingsWorkBuilder;
+import org.hibernate.search.backend.elasticsearch.work.builder.impl.RefreshWorkBuilder;
+import org.hibernate.search.backend.elasticsearch.work.builder.impl.ScrollWorkBuilder;
+import org.hibernate.search.backend.elasticsearch.work.builder.impl.SearchWorkBuilder;
+
+import com.google.gson.JsonObject;
+
+/**
+ * @author Yoann Rodiere
```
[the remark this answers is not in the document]
fax4ever: I prefer to keep them, so that it is more clear the original authoring.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/ElasticsearchStubWorkExecutionContext.java
```diff
@@ -24,4 +26,14 @@ public ElasticsearchStubWorkExecutionContext(ElasticsearchClient client) {
...
 	}
+
+	@Override
+	public void setIndexDirty(URLEncodedString indexName) {
+		// TODO scope execution context per request
+	}
+
+	@Override
+	public GsonProvider getGsonProvider() {
+		return client.getGsonProvider();
```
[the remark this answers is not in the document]
fax4ever: Having more lines of code for something that could change in the future, from my experience could be a premature optimization. On the other hand the impact here is not very critical, following the collaboration I would have +10 -5 LOC change. So I will change ;)

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchResultExtractorImpl.java
fax4ever: You're right

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/ElasticsearchStubWorkExecutionContext.java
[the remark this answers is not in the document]
yrodiere: Thanks. In general it may be true, but here we're talking about controlling the flow of information, the responsibilities of each class, and encapsulation in general. So I wouldn't call that "optimization", it's just design. Adding a getter here and there may be easier in the short term, but over time it introduces dependencies between components and makes changes harder (because they have a larger impact on the codebase).

Anyway, in this particular case I'm working on a PR where, indeed, I need to introduce another implementation of the Elasticsearch client for tests. So, rest assured it's not for nothing.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchBackendFactory.java
fax4ever: Exactly, eventually we will have only the latter. This is a transitory state that won't exist at the end of phase 1 of this issue.

fax4ever:
I took some time to rebase it, merging all fix-up commits. So that the fix-up change won't be visible on the commit history.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/index/impl/ElasticsearchIndexWorkPlan.java
```diff
@@ -40,12 +38,11 @@
 
 	private final List<ElasticsearchWork<?>> works = new ArrayList<>();
 
-	ElasticsearchIndexWorkPlan(ElasticsearchWorkFactory factory, MultiTenancyStrategy multiTenancyStrategy,
```
yrodiere: Could you please preserve the original order of parameters? The factory was before the multi-tenancy strategy, because it's more important.

fax4ever: sure

fax4ever: Do we need to keep the same order, factory < multi-tenancy strategy, in `ElasticsearchIndexWorkExecutor` and `ElasticsearchIndexDocumentWorkExecutor` too?

fax4ever:
The merge with v5 work items is completed. I'm going to add an extra commit to rearrange the package inside the subtree `org.hibernate.search.backend.elasticsearch.work`. My idea is to remove nested `real` package and merge it with `org.hibernate.search.backend.elasticsearch.work.impl`. Any other ideas are well accepted.

fax4ever:
Just rebased. It was quite extensive, so I hope that I didn't do something wrong.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -386,4 +386,7 @@ SearchException conflictingFieldTypesForProjection(String absoluteFieldPath,
 			value = "Multiple conflicting types for identifier: '%1$s' vs. '%2$s'.")
 	SearchException conflictingIdentifierTypesForPredicate(ToDocumentIdentifierValueConverter<?> component1,
 			ToDocumentIdentifierValueConverter<?> component2, @Param EventContext context);
+
+	@Message(id = ID_OFFSET_3 + 50, value = "Error on a previous work exception")
+	SearchException previousWorkException(@Cause Throwable throwable);
```
yrodiere: I suppose you added this to satisfy some checkstyle rule or similar?

I'm not against it, but this method should return `PreviousWorkException`. We actually use `instanceof` to discriminate between these exceptions and other exceptions, so this is important. See where this type is referenced.

I'm a bit surprised this didn't fail the tests, by the way... maybe a test is missing?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/AbstractElasticsearchBarrierWorkOrchestrator.java
```diff
@@ -63,13 +63,29 @@ protected final String getName() {
 		}
 	}
 
+	// TODO avoid code duplication here
```
yrodiere: Indeed. You could simply move the concept of "Changeset" to this class and have four methods:

* `public <T> CompletableFuture<T> submit(ElasticsearchWork<T>)` creates a `SingleWorkChangeset`, calls `submit(Changeset<T>)` and returns the future passed to the constructor of the changeset.
* `public CompletableFuture<?> submit(List<ElasticsearchWork<?>>)` creates a `MultiWorkChangeset`, calls `submit(Changeset<T>)` and returns the future passed to the constructor of the changeset.
* `private <T> void submit(Changeset<T>)` implements the code below and delegates to `doSubmit(Changeset<T>)`
* `protected abstract <T> void doSubmit(Changeset<T>)` is abstract and implemented in subclasses

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/ElasticsearchWorkExecutor.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.util.concurrent.CompletableFuture;
+
+import org.hibernate.search.backend.elasticsearch.work.impl.ElasticsearchWork;
+import org.hibernate.search.backend.elasticsearch.work.impl.ElasticsearchWorkExecutionContext;
+
+/**
+ * @author Yoann Rodiere
+ */
+public interface ElasticsearchWorkExecutor {
```
yrodiere: I thought you wanted to remove this?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
fax4ever: You're right, indeed it does not make sense using a log for an exception that is supposed to be catch and not throw to the outside. This code is not covered by test yet. This is the commit to handle the single work.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/AbstractElasticsearchBarrierWorkOrchestrator.java
fax4ever: Good idea. Thank you. I will apply it.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/ElasticsearchWorkExecutor.java
fax4ever: Yes I did, but I would keep at the this state, to preserve commit authoring from v5.

fax4ever:
Brought back v5 orchestrators, applying a minimal adapting change. There are still two thing I would like to do before closing the task:

1. Remove the `refresh` on `IndexWork` and `DeleteWork`
2. Set default of `REFRESH_AFTER_WRITE` back to false

If I try now, some tests fails. I have to figured out why.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/impl/BulkWork.java
```diff
@@ -0,0 +1,195 @@
...
+			return builder.build();
+		}
+
+		@Override
+		public BulkWork build() {
+			return new BulkWork( this );
+		}
+	}
+
+	private static class BulkResultDeafualt implements BulkResult {
```
yrodiere: There are typos: `BulkResultDeafualt` => `BulkResultDefault`

Also, I'm not sure why it's not simply named `BulkResultImpl`, as in Search 5? Some JQAssistant check were failing maybe?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/AbstractElasticsearchBarrierWorkOrchestrator.java
```diff
@@ -63,28 +95,49 @@ protected final String getName() {
...
 	}
 
-	@Override
-	public <T> CompletableFuture<T> submit(ElasticsearchWork<T> work) {
-		throw new UnsupportedOperationException( "Single work execution is not supported yet!" );
-	}
+	interface Changeset {
+		void applyToDelegate(ElasticsearchFlushableWorkOrchestrator delegate);
+		CompletableFuture<?> getFuture();
+	};
```
yrodiere: There's an extra semicolon after the interface.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/index/impl/ElasticsearchIndexWorkPlan.java
```diff
@@ -98,8 +98,11 @@ public void prepare() {
 	@Override
 	public CompletableFuture<?> execute() {
 		try {
-			CompletableFuture<?> future = orchestrator.submit( works );
-			return future;
+			// TODO verify with the team if this strategy makes sense:
+			// forcing flush to make subsequent operations consistent with the future
```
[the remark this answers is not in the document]
fax4ever: Ok. I'm going to apply the other change here.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/ElasticsearchWorkProcessor.java
```diff
@@ -0,0 +1,265 @@
...
+	 * @param indexName The name of the index that all submitted works will target.
+	 * @return the orchestrator for stream background works.
+	 */
+	public ElasticsearchBarrierWorkOrchestrator createStreamOrchestrator(String indexName) {
+		return streamOrchestrator.createChild(
+				"Elasticsearch stream work orchestrator for index " + indexName
+		);
+	}
+
+	private <T> CompletableFuture<T> start(ElasticsearchWork<T> work, ElasticsearchWorkExecutionContext context) {
```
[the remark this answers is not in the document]
fax4ever: 100% agreed

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchBackendImpl.java
```diff
@@ -65,23 +67,22 @@
...
 			MultiTenancyStrategy multiTenancyStrategy) {
 		this.client = client;
 		this.name = name;
+		this.workProcessor = new ElasticsearchWorkProcessor( new LogErrorHandler(), client, gsonProvider, workFactory );
 		this.analysisDefinitionRegistry = analysisDefinitionRegistry;
 		this.multiTenancyStrategy = multiTenancyStrategy;
-		this.streamOrchestrator = new ElasticsearchStubWorkOrchestrator( client, gsonProvider );
-		this.queryOrchestrator = new ElasticsearchStubWorkOrchestrator( client, gsonProvider );
+		this.streamOrchestrator = workProcessor.createStreamOrchestrator( name );
+		this.queryOrchestrator = workProcessor.createStreamOrchestrator( name );
```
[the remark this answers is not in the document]
fax4ever: Good idea, Moreover, I would like to remove `streamOrchestrator` instance from `IndexingBackendContext`, so that it will be used just by initializeIndex method. Do you agree with that idea?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/index/impl/ElasticsearchIndexManagerImpl.java
```diff
@@ -59,12 +60,14 @@
...
 		this.refreshAfterWrite = refreshAfterWrite;
 		this.workPlanOrchestrator = indexingBackendContext.createWorkPlanOrchestrator( elasticsearchIndexName.encoded, refreshAfterWrite );
+		this.streamOrchestrator = streamOrchestrator;
 	}
 
 	@Override
 	public void close() {
 		// Index managers own the work plan context, but not the stream context (which is shared)
 		workPlanOrchestrator.close();
+		streamOrchestrator.close();
```
yrodiere: If the first orchestrator throws an exception when closing, the second one will never get closed.
Please use a closer instead; see `org.hibernate.search.backend.elasticsearch.impl.ElasticsearchBackendImpl#close` for an example.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchBackendImpl.java
```diff
@@ -86,11 +85,10 @@
 		this.orchestratorFactory = new ElasticsearchWorkOrchestratorFactory( new LogErrorHandler(), client, gsonProvider, workFactory );
 		this.analysisDefinitionRegistry = analysisDefinitionRegistry;
 		this.multiTenancyStrategy = multiTenancyStrategy;
-		this.streamOrchestrator = orchestratorFactory.createStreamOrchestrator( name );
 		this.queryOrchestrator = orchestratorFactory.createStreamOrchestrator( name );
```
yrodiere: You should pass a more meaningful name for the orchestrator. Something like `"Elasticsearch query orchestrator for backend " + name`.

fax4ever:
rebased

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/ElasticsearchDefaultWorkSequenceBuilder.java
```diff
@@ -0,0 +1,194 @@
...
+ * <p>
+ * Execution of works will not be performed by this class;
+ * instead, it delegates to an "executor" passed to the constructor.
+ * <p>
+ * Works will be executed inside a sequence-scoped context (a {@link ElasticsearchFlushableWorkExecutionContext}),
+ * ultimately leading to a {@link ElasticsearchFlushableWorkExecutionContext#flush()}.
+ */
+class ElasticsearchDefaultWorkSequenceBuilder implements ElasticsearchWorkSequenceBuilder {
+
+	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
```
yrodiere: This logger is not used, please remove it.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/builder/impl/DeleteByQueryWorkBuilder.java
```diff
@@ -6,7 +6,7 @@
  */
 package org.hibernate.search.backend.elasticsearch.work.builder.impl;
 
-import org.hibernate.search.backend.elasticsearch.util.impl.URLEncodedString;
```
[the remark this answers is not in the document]
fax4ever: Right, I merge the wrong commit on rebasing,

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/builder/factory/impl/ElasticsearchWorkBuilderFactory.java
```diff
@@ -0,0 +1,78 @@
...
+
+	CloseIndexWorkBuilder closeIndex(URLEncodedString indexName);
+
+	IndexExistsWorkBuilder indexExists(URLEncodedString indexName);
+
+	GetIndexSettingsWorkBuilder getIndexSettings(URLEncodedString indexName);
+
+	PutIndexSettingsWorkBuilder putIndexSettings(URLEncodedString indexName, IndexSettings settings);
+
+	// TODO restore method => GetIndexTypeMappingsWorkBuilder getIndexTypeMappings(URLEncodedString indexName);
```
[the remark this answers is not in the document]
fax4ever: Is there something similar to `org.hibernate.search.elasticsearch.schema.impl.model.TypeMapping` in v6?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/real/impl/IndexWork.java
```diff
@@ -0,0 +1,89 @@
...
+
+	@Override
+	protected Void generateResult(ElasticsearchWorkExecutionContext context, JsonObject bulkResponseItem) {
+		return null;
+	}
+
+	@Override
+	protected CompletableFuture<?> afterSuccess(ElasticsearchWorkExecutionContext context) {
+		return super.afterSuccess( context );
+	}
```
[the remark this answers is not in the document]
fax4ever: removed from history

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/builder/factory/impl/ElasticsearchWorkBuilderFactory.java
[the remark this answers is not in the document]
yrodiere: Right, @gsmet renamed that because we both thought it was cleaner for various reasons. 

`TypeMapping` has been split in two: `AbstractTypeMapping`, the base class, and `RootTypeMapping`, the concrete class.

In this case I think you need `RootTypeMapping`. That's the one you used in `org.hibernate.search.backend.elasticsearch.work.impl.CreateIndexWork.Builder#mapping`.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/impl/ElasticsearchStubWorkFactory.java
```diff
@@ -182,7 +167,7 @@ public ElasticsearchStubWorkFactory(GsonProvider gsonProvider, MultiTenancyStrat
 				.body( payload );
 
 		if ( !routingKeys.isEmpty() ) {
-			builder.param( "_routing", routingKeys.stream().collect( Collectors.joining( "," ) ) );
+			builder.param( "routing", routingKeys.stream().collect( Collectors.joining( "," ) ) );
```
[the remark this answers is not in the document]
fax4ever: Seen :)

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/real/impl/SearchWork.java
```diff
@@ -114,11 +124,15 @@ protected ElasticsearchRequest buildRequest() {
 				builder.param( "scroll", scrollTimeout );
 			}
 
+			if ( !routingKeys.isEmpty() ) {
+				builder.param( "routing", routingKeys.stream().collect( Collectors.joining( "," ) ) );
```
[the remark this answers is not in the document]
fax4ever: good idea

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/work/IndexDocumentWorkExecutorIT.java
```diff
@@ -67,6 +69,11 @@ public void checkAllDocumentsAreSearchable() throws Exception {
 		}
 		CompletableFuture.allOf( tasks ).join();
 
+		// TODO verify with the team if this strategy makes sense:
```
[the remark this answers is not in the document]
fax4ever: I wanted to be sure about it, Thanks

yrodiere:
Looks good, apart from a few forgotten `<T>` that I added. I just force-pushed, let's see what CI has to say...

fax4ever:
Merged, thanks. :)

yrodiere:
Great! Thanks a lot for all the work :)

PR #1854, opened 2018-12-17
HSEARCH-3455 Add a way to test the JSON sent to the Elasticsearch cluster

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3455


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/gson/spi/GsonProvider.java
```diff
@@ -4,7 +4,7 @@
  * License: GNU Lesser General Public License (LGPL), version 2.1 or later
  * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
  */
-package org.hibernate.search.backend.elasticsearch.gson.impl;
+package org.hibernate.search.backend.elasticsearch.gson.spi;
```
gsmet: Do we really want to expose Gson in our SPI?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/gson/spi/JsonLogHelper.java
```diff
@@ -19,7 +19,7 @@
  *
  * @author Yoann Rodiere
  */
-public class JsonLogHelper {
+public final class JsonLogHelper {
```
gsmet: Same question here, what's the purpose of exposing a helper class you cannot extend and that does not implement an interface (so you can't provide an alternative implementation)?

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/util/ElasticsearchClientSubmitCall.java
```diff
@@ -0,0 +1,91 @@
...
+				break;
+		}
+	}
+
+	private String toComparableJson(List<JsonObject> bodyParts) {
+		JsonArray array = new JsonArray( bodyParts.size() );
+		for ( JsonObject bodyPart : bodyParts ) {
+			array.add( bodyPart );
+		}
+		return array.toString();
```
gsmet: We are sure this will format the JSON in a reliable way?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/impl/ElasticsearchStubWorkFactory.java
```diff
@@ -72,7 +72,7 @@ public ElasticsearchStubWorkFactory(GsonProvider gsonProvider, MultiTenancyStrat
 				.body( document );
 		builder.param( "refresh", true );
 		if ( routingKey != null ) {
-			builder.param( "_routing", routingKey );
+			builder.param( "routing", routingKey );
```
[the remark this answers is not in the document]
gsmet: Have to say I tend to agree with Sonar, let's make it a constant.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/util/ElasticsearchClientSubmitCall.java
yrodiere: At least the current implementation produces simple, clean JSON. But I'll change this to use a Gson object just in case...

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/gson/spi/GsonProvider.java
yrodiere: We have to, if we want to allow to replace the Elasticsearch client: it takes JSON as input. We could format it to a String, but that would prevent some optimizations in the client itself (related to non-blocking requests).

I would not expose Gson in APIs, but in SPIs, especially this single SPI related to the Elasticsearch client, I think it's safe. If we were to replace Gson with something else, we wouldn't do it in a micro anyway, would we? And we allow ourselves to break SPIs in minors.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/gson/spi/JsonLogHelper.java
yrodiere: The helper is provided for use by the Elasticsearch client (since it's... a helper). It's not supposed to be implemented by anyone else, and if people don't want to use it, they can ignore it.

The helper is only here to help logging the requests/responses, though. I could improve the situation: I could only pass a Gson object to the client (not a GsonProvider), and have Hibernate Search wrap all clients with a wrapper that implements the logging. Something like this:

```
public class LoggingElasticsearchClientWrapper implements ElasticsearchImplementor {
    CompletableFuture<ElasticsearchResponse> submit(ElasticsearchRequest request) {
        int startTime = ...;
        CompletableFuture<ElasticsearchResponse> result = delegate.submit( request );
        if ( requestLog.isDebugEnabled() ) {
            result.thenAccept( response -> log( request, startTime, response ) );
        }
        return result;
    }

    // ... delegate the other methods...

	private void log(ElasticsearchRequest request, long start, ElasticsearchResponse response) {
            // Log according to backend settings (passed to the constructor)
	}
}
```

Then the SPI would only be tainted with `Gson`, not with our own classes (`GsonProvider` and `JsonHelper`).

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/impl/ElasticsearchStubWorkFactory.java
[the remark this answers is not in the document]
yrodiere: This code is being replaced by Fabio, who is working on porting the real (non-stub) work factory from Search 5. We'll see what Sonar thinks of that code.

yrodiere:
As discussed on Gitter, I'll leave the SPI as they are and we'll have a closer look when we switch from Gson to something else: https://hibernate.atlassian.net/browse/HSEARCH-2817

Merged. Thanks!

PR #1857, opened 2018-12-21
HSEARCH-3291 Search 6 groundwork - Split the field definition API in two: encoding definition and field definition

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3291

So... admittedly, this makes the syntax of the index schema definition DSL worse. It does clean up the implementation, but it's not worth it if the API is unusable.

On the other hand, this change also allows us to clean up the `ValueBridge` API, requiring users to define a type, and not a field, which is more in line with the semantics of the `ValueBridge` (the field name and the accessor are none of the `ValueBridge`'s business).

I wouldn't say the new API is "unusable", just a bit awkward, and fortunately in a similar way to the predicate/sort/projection DSLs. Thus I intend to try to solve the API issues in the same ticket as the other DSLs: https://hibernate.atlassian.net/browse/HSEARCH-3444. Let's merge it in the meantime? Worst case I should be able to revert only the commit that splits the APIs, and just keep the commit that splits the internals.


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/document/model/dsl/impl/LuceneIndexSchemaBuildContext.java
```diff
@@ -6,10 +6,10 @@
...
 package org.hibernate.search.backend.lucene.document.model.dsl.impl;
 
-import org.hibernate.search.engine.backend.document.model.dsl.spi.IndexSchemaContext;
+import org.hibernate.search.engine.backend.document.model.dsl.spi.IndexSchemaBuildContext;
 
-public interface LuceneIndexSchemaContext extends IndexSchemaContext {
+public interface LuceneIndexSchemaBuildContext extends IndexSchemaBuildContext {
 
-	LuceneIndexSchemaRootContext getRoot();
+	LuceneIndexBuildContext getRoot();
```
gsmet: This change looks weird with the new interface name vs the method name.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/dsl/impl/AbstractElasticsearchIndexSchemaObjectNodeBuilder.java
```diff
@@ -66,19 +65,17 @@ public IndexSchemaObjectFieldNodeBuilder createExcludedObjectField(String relati
...
-			PropertyMapping propertyMapping = propertyContributor.contribute( collector, node );
-			mapping.addProperty( propertyName, propertyMapping );
+		for ( Map.Entry<String, ElasticsearchIndexSchemaNodeContributor> entry : content.entrySet() ) {
+			ElasticsearchIndexSchemaNodeContributor propertyContributor = entry.getValue();
+			propertyContributor.contribute( collector, node, mapping );
 		}
 	}
 
-	abstract ElasticsearchIndexSchemaRootNodeBuilder getRootNodeBuilder();
+	abstract ElasticsearchIndexSchemaRootNodeBuilder getRoot();
```
gsmet: I fail to see how this is an improvement?

engine/src/main/java/org/hibernate/search/engine/backend/document/model/dsl/IndexSchemaElement.java
```diff
@@ -7,19 +7,73 @@
...
  */
 public interface IndexSchemaElement {
 
-	IndexFieldTypeFactoryContext field(String relativeFieldName);
+	/**
+	 * Add a field to this index schema element with the given type.
+	 *
+	 * @param relativeFieldName The relative name of the new field.
+	 * @param type The type of the new field.
+	 * @param <F> The type values help by the field.
```
gsmet: `The type values help by the field.` I don't understand this javadoc? `help` seems wrong but maybe I'm missing something? Same for all the other methods below.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/spatial/AbstractSpatialWithinSearchPredicateIT.java
```diff
@@ -114,14 +114,38 @@ protected void initData() {
...
+			)
+					.createAccessor();
+			geoPoint_with_longName = root.field(
+					"geoPoint_with_a_veeeeeeeeeeeeeeeeeeeeerrrrrrrrrrrrrrrrrryyyyyyyyyyyyyyyy_long_name",
+					f -> f.asGeoPoint().sortable( Sortable.YES ).projectable( Projectable.YES ).toIndexFieldType()
+			)
+					.createAccessor();
+			nonProjectableGeoPoint = root.field(
+					"nonProjectableGeoPoint",
+					f -> f.asGeoPoint().projectable( Projectable.NO ).toIndexFieldType()
```
gsmet: Can we make `.toIndexFieldType()` optional for the lambda version or is it not possible?

yrodiere:
Merged, thanks!

PR #1859, opened 2019-01-10
HSEARCH-3421 Support for ID predicates when multi-tenancy is enabled

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3421


yrodiere:
Merged, thanks!

PR #1864, opened 2019-01-17
HSEARCH-3424 Clean up the showcases to turn it into a full-stack ORM integration test suite

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3424


integrationtest/fullstack/library/src/test/java/org/hibernate/search/integrationtest/fullstack/OrmEsManualIndexingIT.java
```diff
@@ -0,0 +1,85 @@
...
+					"The Divine Comedy is composed of 14,233 lines that are divided into three cantiche (singular cantica) – Inferno (Hell), Purgatorio (Purgatory), and Paradiso (Paradise)",
+					"literature,poem,afterlife" );
+		}
+		documentRepo.createBooks( Arrays.asList( books ) );
+	}
+
+	@Test
+	public void testMassIndexing() {
+		checkNothingIsIndexed();
+		FullTextEntityManager ftEntityManager = Search.getFullTextEntityManager( documentRepo.entityManager() );
```
yrodiere: If you cannot inject the entity manager directly in tests (???), then please move the reindexing operation to a dedicated service in the main code, and call it from here. It makes much more sense than breaking encapsulation like this. It's a showcase, let's not showcase bad practices ;)

integrationtest/fullstack/library/src/main/java/org/hibernate/search/integrationtest/fullstack/library/cdi/EntityManagerFactoryProducer.java
```diff
@@ -0,0 +1,47 @@
...
+	public EntityManagerFactory autoIndexingEMF() {
+		Map<String, Object> props = new HashMap<>();
+		props.put( "javax.persistence.bean.manager", beanManager );
+		return Persistence.createEntityManagerFactory( "elasticsearchAutoIndexing", props );
+	}
+
+	@Produces
+	@ApplicationScoped
+	@ManualIndexingEM
+	public EntityManagerFactory manualIndexingEMF() {
```
yrodiere: We should probably add a TODO here: once HSEARCH-1383 is fixed, we should be able to use the same EMF.

integrationtest/fullstack/library/src/test/java/org/hibernate/search/integrationtest/fullstack/support/JtaEnvironment.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.integrationtest.fullstack.support;
+
+import org.junit.rules.ExternalResource;
+
+import com.arjuna.ats.jta.utils.JNDIManager;
+import org.jnp.server.NamingBeanImpl;
+
+public class JtaEnvironment extends ExternalResource {
```
yrodiere: Do we really need JTA in our case? I would have thought it was only required if we needed XA transactions.

integrationtest/fullstack/library/src/test/java/org/hibernate/search/integrationtest/fullstack/OrmEsManualIndexingIT.java
fax4ever: Good idea

integrationtest/fullstack/library/src/main/java/org/hibernate/search/integrationtest/fullstack/library/cdi/EntityManagerFactoryProducer.java
fax4ever: Fine

integrationtest/fullstack/library/src/test/java/org/hibernate/search/integrationtest/fullstack/support/JtaEnvironment.java
fax4ever: If I remember well JTS is the JTA support for XA, JTA would be used even for local transaction. I'll check what happens with out it.

fax4ever:
@yrodiere about the complexity I agree with you, but actually I copied that our team did here: https://github.com/fax4ever/hibernate-demos/tree/master/other/cdi-jpa-testing. I'm going to check if we could simplify something.

fax4ever:
I'm closing this preview. I'll open another one based on a different branch: https://github.com/fax4ever/hibernate-search/tree/3424-enanche-showcase-module.

PR #1873, opened 2019-01-31
HSEARCH-3424 enhance showcase module

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3424

~~Based on top of https://github.com/hibernate/hibernate-search/pull/1877 for some Lucene tests.~~ done!

~~Preview of alternative solution based on Spring Boot runtime.
It's a very `preview`, so that please treat the changes here as very transient.~~ not anymore!


integrationtest/showcase/library/src/test/resources/application-elasticsearch.yaml
```diff
@@ -1,11 +1,16 @@
 spring.jpa:
-  hibernate:
-    ddl-auto: create-drop
   properties:
     hibernate.search:
       indexing_strategy: ${indexing.strategy}
+      default_backend: elasticsearch1
       backends:
         elasticsearch1:
+          type: elasticsearch
```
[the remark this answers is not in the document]
fax4ever: I think I wrote this: 
> Temporary commit. Many things are transient here, such as change LuceneTextFieldPredicateBuilderFactory.java is supposed to be done in other issue, OrmLuceneLibraryShowcaseIT.java is a temporary copy and paste, I have an issue on sorting lowercase/uppercase string, and so and on so forth. So do not review this commit, please.

Didn't I?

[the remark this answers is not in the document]
yrodiere: My bad, I didn't see that. So I suppose it's no longer a preview PR... for the time being :)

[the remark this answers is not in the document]
fax4ever: Not a problem ;)

README.md
```diff
@@ -166,6 +166,12 @@ and do the work of converting between user entities and documents to be indexed.
 * `reports`: Module built last, producing reports related to test coverage in particular.
 * `util`: Various modules containing util classes, both for runtime and for tests.
 
+## Showcase Project
+
+The project [Hibernate Search with Hibernate ORM Showcase](hibernate-search/tree/master/integrationtest/showcase/library) shows possible Hibernate Search application cases.
```
yrodiere: I think this link is wrong, it does not work.
Maybe rather point to `integrationtest/showcase/library`?

Also, when I wrote the ticket I was merely thinking about adding a description of the showcase in the list above. Something like:

```
* `integrationtest`: Integration tests for backends (Elasticsearch, Lucene) and mappers (Hibernate ORM),
as well as any other technology Hibernate Search integrates with.
Here are some notable sub-directories:
  *  `performance`: performance tests.
  *  `showcase/library`: a sample application using Hibernate Search in a Spring Boot environment.
```

For users wanting to know more, the paragraph you added to the reference documentation should do the trick.

documentation/src/main/asciidoc/getting-started.asciidoc
```diff
@@ -441,6 +441,13 @@ include::{sourcedir}/org/hibernate/search/documentation/gettingstarted/withhsear
 ----
 ====
 
+=== Showcase Project
+
+The project [Hibernate Search with Hibernate ORM Showcase](https://github.com/hibernate/hibernate-search/tree/master/integrationtest/showcase/library)
```
yrodiere: This feels a bit... out of place? Maybe this would be better at the end of the "What's next" section below.

Also, rephrased like this, it would feel more natural:

```
If you want to see an example project using Hibernate Search,
you can also have a look at the ["Library" showcase](https://github.com/hibernate/hibernate-search/tree/master/integrationtest/showcase/library),
a sample application using Hibernate Search in a Spring Boot environment.
```

integrationtest/showcase/library/src/main/java/org/hibernate/search/integrationtest/showcase/library/config/SessionFactoryConfig.java
```diff
@@ -0,0 +1,84 @@
...
+import org.hibernate.search.integrationtest.showcase.library.model.Library;
+import org.hibernate.search.integrationtest.showcase.library.model.Person;
+import org.hibernate.search.integrationtest.showcase.library.model.Video;
+import org.hibernate.search.integrationtest.showcase.library.model.VideoCopy;
+import org.hibernate.search.mapper.orm.cfg.HibernateOrmIndexingStrategyName;
+import org.hibernate.search.mapper.orm.cfg.HibernateOrmMapperSettings;
+import org.hibernate.service.ServiceRegistry;
+import org.hibernate.tool.schema.Action;
+
+public class SessionFactoryConfig {
```
yrodiere: While it's a good idea, I really don't see how this is related to this commit in any way.

integrationtest/showcase/library/src/test/java/org/hibernate/search/integrationtest/showcase/library/OrmElasticsearchLibraryShowcaseIT.java
```diff
@@ -0,0 +1,594 @@
...
+import org.springframework.test.context.TestPropertySource;
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+@ActiveProfiles("test")
+@TestPropertySource(properties = { "indexing.strategy=event" })
+public class OrmElasticsearchLibraryShowcaseIT {
+
+	private static final String PREFIX = HibernateOrmMapperSettings.PREFIX;
```
yrodiere: This constant is no longer necessary and should be removed.

integrationtest/showcase/library/src/main/java/org/hibernate/search/integrationtest/showcase/library/analysis/LibraryAnalysisConfigurer.java
```diff
@@ -6,25 +6,9 @@
  */
 package org.hibernate.search.integrationtest.showcase.library.analysis;
 
-import org.hibernate.search.backend.elasticsearch.analysis.ElasticsearchAnalysisConfigurer;
-import org.hibernate.search.backend.elasticsearch.analysis.model.dsl.ElasticsearchAnalysisDefinitionContainerContext;
+public abstract class LibraryAnalysisConfigurer {
```
yrodiere: I was a bit surprised by the fact you define an common superclass just to define constants, but ok. 
That being said, sonar complains because this class looks like a "constants" class, which should be final and define a private constructor.

Maybe you could just rename this class to `LibraryAnalyzers`, make it final, add a private constructor, and not extend it from the configurers?

integrationtest/showcase/library/src/main/java/org/hibernate/search/integrationtest/showcase/library/repository/impl/DocumentRepositoryImpl.java
```diff
@@ -144,4 +147,25 @@ public long count() {
...
+				.asProjections(
+						target.projection().field( "author", String.class ).toProjection()
+				)
+				.predicate( f -> f.bool( b -> {
+					b.must( f.match()
+							.onField( "title" ).boostedTo( 2.0f )
+							.orField( "summary" )
+							.matching( terms )
+					);
+				} ) )
```
yrodiere: You don't need a boolean query here, since you're not combining anything. This would work just fine:

```
				.predicate( f -> f.match()
					.onField( "title" ).boostedTo( 2.0f )
					.orField( "summary" )
					.matching( terms )
				) 
```

integrationtest/showcase/library/src/main/java/org/hibernate/search/integrationtest/showcase/library/repository/impl/DocumentRepositoryImpl.java
```diff
@@ -144,4 +147,25 @@ public long count() {
 
 		return query.getResultList();
 	}
+
+	@Override
+	public List<String> getAuthorsOfBooksHavingTerms(String terms, SortOrder order) {
+		FullTextSearchTarget<Document> target = entityManager.search( Document.class );
+		FullTextQuery<List<?>> query = target.query()
+				.asProjections(
```
yrodiere: Use `asProjection` (without the trailing "s") and you'll get a `FullTextQuery<String>` and will not need the awkward stream/map/collect below.

Also, you could use lambdas here, too:

```
.asProjections( f -> f.field( "author", String.class ) )
```

fax4ever:
Rebased, now I'm addressing the changes.

integrationtest/showcase/library/src/main/java/org/hibernate/search/integrationtest/showcase/library/model/Document.java
```diff
@@ -64,6 +64,29 @@
...
+
+	public Document(int id, String title, String author, String summary, String tags) {
+		this.id = id;
+		this.title = title;
+		this.author = author;
+		this.summary = summary;
+		this.tags = tags;
+	}
+
+	public C getCopy(Library library, int copyIndex) {
```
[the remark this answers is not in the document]
fax4ever: In general, I prefer to push the logic inside the entity in order to avoid an [Anemic Domain Model](https://www.martinfowler.com/bliki/AnemicDomainModel.html). Furthermore, some people say that with Hibernate it is impossible to build a model avoiding an Anemic Domain and this would be a good counterexample example of that. Anyway let's do as you wish ;)

integrationtest/showcase/library/src/main/java/org/hibernate/search/integrationtest/showcase/library/service/BorrowalService.java
```diff
@@ -0,0 +1,123 @@
...
+
+	public List<Person> listTopShortTermBorrowers(int offset, int limit) {
+		return listTopBorrowers( "account.borrowals.shortTermCount", offset, limit );
+	}
+
+	public List<Person> listTopLongTermBorrowers(int offset, int limit) {
+		return listTopBorrowers( "account.borrowals.longTermCount", offset, limit );
+	}
+
+	public List<Person> searchPerson(String terms, int offset, int limit) {
```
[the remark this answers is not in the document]
fax4ever: You're right, good catch-up

integrationtest/showcase/library/src/main/java/org/hibernate/search/integrationtest/showcase/library/service/DocumentService.java
```diff
@@ -4,66 +4,100 @@
...
+	public Book createBook(int id, String isbn, String title, String author, String summary, String tags) {
+		return documentRepo.save( new Book( id, isbn, title, author, summary, tags ) );
+	}
+
+	public Video createVideo(int id, String title, String author, String summary, String tags) {
+		return documentRepo.save( new Video( id, title, author, summary, tags ) );
+	}
+
+	public Iterable<Book> createBooks(List<Book> books) {
+		return documentRepo.saveAll( books );
```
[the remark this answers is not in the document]
fax4ever: You're right. In fact I created this method when we didn't have `TestDataService`. I'll remove it.

integrationtest/showcase/library/src/test/java/org/hibernate/search/integrationtest/showcase/library/OrmElasticsearchLibraryShowcaseIT.java
```diff
@@ -0,0 +1,594 @@
...
+	public void cleanUpAllDataAndIndexes() {
+		adminService.deleteAllBorrowalsLibrariesAndAccounts();
+		documentService.deleteAll();
+	}
+
+	@Test
+	public void search_library() {
+		List<Library> libraries = libraryService.search( "library", 0, 10 );
+		assertThat( libraries ).containsExactly(
+				libraryService.load( CITY_CENTER_ID ),
```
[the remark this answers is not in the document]
fax4ever: Good idea

integrationtest/showcase/library/src/main/java/org/hibernate/search/integrationtest/showcase/library/config/SessionFactoryConfig.java
fax4ever: You're right, I will push this change in a different commit.

integrationtest/showcase/library/src/main/java/org/hibernate/search/integrationtest/showcase/library/repository/impl/DocumentRepositoryImpl.java
fax4ever: Good api ;)

fax4ever:
@yrodiere thank you for the review. I'll add one commit more, the one to push some logic from entities to services. Even if I don't very agree with that change :P

yrodiere:
Looks good! I rebased, squashed a few commits, and added small fixup commits.
I'm testing a full build on CI, and I will merge once I'm certain the full build works correctly.

fax4ever:
@yrodiere thanks

yrodiere:
Merged. Thanks!

PR #1877, opened 2019-02-05
HSEARCH-3477 Incompatible types error querying a full text field across different indexes

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3477


yrodiere:
@fax4ever I added a commit to move your test to `MatchSearchPredicateIT`, so as to be consistent with what we do in `FieldSearchProjectionIT#multiIndex()`, and to clear the way for https://hibernate.atlassian.net/browse/HSEARCH-3307.

If that change looks good to you, we can merge!

fax4ever:
Merged #1877. Thanks

PR #1878, opened 2019-02-05
HSEARCH-3054 +  HSEARCH-3080 Clear the way for Elasticsearch 7 support (type name + dialects)

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3054 : restore support for Elasticsearch dialects
https://hibernate.atlassian.net//browse/HSEARCH-3080 : pick an appropriate type name

See the commits messages for details.


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/dialect/impl/es56/Elasticsearch56Dialect.java
```diff
@@ -0,0 +1,19 @@
...
+ */
+package org.hibernate.search.backend.elasticsearch.dialect.impl.es56;
+
+import org.hibernate.search.backend.elasticsearch.dialect.impl.ElasticsearchDialect;
+import org.hibernate.search.backend.elasticsearch.dialect.impl.es6.Elasticsearch6Dialect;
+
+/**
+ * The dialect for Elasticsearch 5.6.
+ */
+public class Elasticsearch56Dialect extends Elasticsearch6Dialect implements ElasticsearchDialect {
```
gsmet: Oh, you chose this way. ORM does the opposite. Is there a particular reason, it doesn't seem very natural?

Especially, you could introduce issues in an older dialect by changing things in the newer one. Doesn't look like the best behavior.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/builder/factory/impl/Elasticsearch6WorkBuilderFactory.java
```diff
@@ -63,14 +63,11 @@
 
 import com.google.gson.JsonObject;
 
-/**
- * @author Yoann Rodiere
- */
-public class ElasticsearchWorkBuilderFactoryImpl implements ElasticsearchWorkBuilderFactory {
+public class Elasticsearch6WorkBuilderFactory implements ElasticsearchWorkBuilderFactory {
```
gsmet: So what if you use the same for 7 too? It will be named 6 but will also be used by 5 and 7?

fax4ever:
@yrodiere nice Job!

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/dialect/impl/es56/Elasticsearch56Dialect.java
gsmet: OK. Can't say I'm excited about this idea because you're right: the last dialect will be the most tested so it's where we don't risk any issue.

Whereas the old ones won't be and we put them at risk.

In any case, I understand where you're coming from so it's not a blocker for me. Just be prepared for my "I told you so" at the first regression of this kind :).

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/builder/factory/impl/Elasticsearch6WorkBuilderFactory.java
gsmet: OK, makes sense.

yrodiere:
Merged, thanks!

PR #1879, opened 2019-02-07
HSEARCH-3478 + HSEARCH-3425 Code cleanup

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3425
https://hibernate.atlassian.net//browse/HSEARCH-3478

Note I don't expect a full review, it would be quite boring.

@gsmet, could you simply double-check that the warnings are gone when you use Eclipse? They are gone in my instance of Eclipse, but who knows.
Of course you need to ignore the legacy modules; personally I moved the non-legacy modules to a separate working set, I configured the "Problems" view to only show problems in the selected projects, and I selected the "non-legacy" working set.


yrodiere:
Note for future reference.

IDEA doesn't offer a global view of all warnings in the codebase, making it difficult to keep the number of warnings to 0. The feature was requested a long time ago and never implemented, so it's unlikely it will ever be: https://youtrack.jetbrains.com/issue/IDEA-68854

Thus I wanted to find *some* solution to be aware of added warnings and fix them as I code.

I tried to fail the build on warnings by passing `-Werror` to the compiler, but it seems to stop after the very first warning. Clearly not good.

I also tried to make all warnings appear in IDEA in the "Messages" view, shown when we trigger a full build. That's better than nothing, I thought. Well, no: by passing `-Xlint:all` to the compiler for non-legacy modules, and `-Xlint:none -nowarn` for legacy modules, and IDEA correctly detected the options (they appear in the build settings). But for some reason, a lot of warning still appear even for legacy modules. Also, the output seems to include irrelevant warnings, such as warnings we suppressed using `@SuppressWarnings`, or heap pollution warnings even though we added `@SafeVarargs`. So we can't control which warnings appear, and there are way too many warnings. Useless.

I'm not going to go back to Eclipse, because even without a global view of warnings, IDEA is still better. And more importantly I won't force people working on Hibernate Search to use one IDE or another.

So... I'm out of idea. I guess we'll have to periodically clean up the codebase...

yrodiere:
Approved over the chat. Merged, thanks!

PR #1881, opened 2019-02-08
HSEARCH-3480 Rename fromExternalRepresentation() to of() in enums

gsmet:
* https://hibernate.atlassian.net/browse/HSEARCH-3480

Maybe don't merge it just yet but incorporate it in our private branch so that we can test that everything is working fine before merging.

Note that I also changed things to consistently normalize the input values and log the normalized values as it's the one compared with the accepted values.

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
@@ -9,6 +9,7 @@ package org.hibernate.search.backend.elasticsearch.logging.impl;
 
 import static org.jboss.logging.Logger.Level.WARN;
 
+import java.util.List;
 import java.util.Map;
 
 import org.hibernate.search.backend.elasticsearch.client.spi.ElasticsearchRequest;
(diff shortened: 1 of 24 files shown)


yrodiere:
Just pushed a commit to use the same implementation everywhere, as discussed over the chat.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -297,9 +299,9 @@ SearchException cannotMixElasticsearchSearchTargetWithOtherBackend(IndexSearchTa
 			value = "Index '%1$s' requires multi-tenancy but the backend does not support it in its current configuration.")
 	SearchException multiTenancyRequiredButNotSupportedByBackend(String indexName, @Param EventContext context);
 
-	@Message(id = ID_OFFSET_3 + 15,
-			value = "Unknown multi-tenancy strategy '%1$s'.")
-	SearchException unknownMultiTenancyStrategyConfiguration(String multiTenancyStrategy);
+	@Message(id = ID_OFFSET_3 + 15, value = "Invalid multi-tenancy strategy name: '%1$s'."
+			+ " Valid names are: %2$s.")
```
gsmet: maybe valid strategies?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -455,4 +457,8 @@ SearchException conflictingIdentifierTypesForPredicate(ToDocumentIdentifierValue
 
 	@Message(id = ID_OFFSET_3 + 51, value = "The operation was skipped due to the failure of a previous work in the same changeset.")
 	SearchException elasticsearchSkippedBecauseOfPreviousWork(@Cause Throwable skippingCause);
+
+	@Message(id = ID_OFFSET_3 + 52, value = "Invalid index lifecycle strategy name: '%1$s'."
+			+ " Valid names are: %2$s.")
```
gsmet: Same here.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
yrodiere: It looks better to me this way, but feel free to change it if you feel strongly against it.

yrodiere:
I rebased and squashed this PR with changes you made in another branch.
The build passes, so I'll merge. Thanks!

PR #1883, opened 2019-02-11
HSEARCH-3468 Restore the ability for the Elasticsearch schema validation to report errors across multiple indexes

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3468


engine/src/main/java/org/hibernate/search/engine/logging/impl/RootFailureCollector.java
```diff
@@ -236,15 +241,19 @@ void appendFailuresTo(ToStringTreeBuilder builder) {
 		}
 
 		private void doAdd(Throwable failure, String failureMessage) {
+			StringJoiner contextJoiner = new StringJoiner( COMMON_MESSAGES.contextSeparator() );
+			appendContextTo( contextJoiner );
+			log.newBootstrapCollectedFailure( contextJoiner.toString(), failure );
```
gsmet: So in the case of an exception, you log it directly but not in the case of a simple string message?

yrodiere: Yes. The main purpose of logging the exception directly is to show the full stack trace.

I suppose I could also log simple String messages, but then in the case of schema validation it will probably be very verbose. And since we currently only use simple String messages for schema validation, it would be rather useless, I think...

yrodiere:
Merged, thanks!

PR #1885, opened 2019-02-12
HSEARCH-3485 + HSEARCH-3484 JDK12 compatibility

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3485
https://hibernate.atlassian.net/browse/HSEARCH-3484

Full build including a build with JDK12: http://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-1885/2/pipeline/66

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -397,7 +397,7 @@ stage('Default build') {
 					"} \
 					-Pdist -Pcoverage -Pjqassistant \
 					${enableDefaultEnvIT ? '' : '-DskipITs'} \
-					${enableDefaultEnvLegacyIT ? '-Dsurefire.legacy.skip=false -Dfailsafe.legacy.skip=false' : ''} \
+					${enableDefaultEnvLegacyIT ? '-Dlegacy.skip=false' : ''} \
 			"""
 
 			// Don't try to report to Coveralls.io or SonarCloud if coverage data is missing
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -1771,6 +1771,38 @@
...
+                </surefire.jvm.args.java-version.strict>
+                <arquillian.wildfly.jvm.args.java-version>
+                    -Djdk.attach.allowAttachSelf=true
+                </arquillian.wildfly.jvm.args.java-version>
+            </properties>
+            <modules>
+                <!-- Nothing at the moment -->
+            </modules>
+        </profile>
+
```
gsmet: I think I would just have made the version of the above profile an open bound. Not sure it's worth having the exact same profile for N+ JDK.

Jenkinsfile
```diff
@@ -394,7 +394,7 @@ stage('Default build') {
 					"} \
 					-Pdist -Pcoverage -Pjqassistant \
 					${enableDefaultEnvIT ? '' : '-DskipITs'} \
-					${enableDefaultEnvLegacyIT ? '-Dsurefire.legacy.skip=false -Dfailsafe.legacy.skip=false' : ''} \
+					${enableDefaultEnvLegacyIT ? '-Dlegacy.skip=false' : ''} \
```
gsmet: Ah nice, I suggested that a while ago, it's definitely better.

pom.xml
yrodiere: I didn't know we could do that. You're right, it would be better; will do.

yrodiere:
Updated according to your comment and launched a full build. Waiting for CI.

yrodiere:
Merged, thanks!

PR #1888, opened 2019-02-13
HSEARCH-3315 Search 6 groundwork - Move publicly exposed util types to a dedicated module

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3315

Obviously there are too many changes to review, so please just have a look at the commit messages and see if you agree with that. Maybe you can also have a look at the resulting package structure of the hibernate-search-util-common module?


yrodiere:
Merged, thanks

PR #1890, opened 2019-02-15
HSEARCH-3473 Search 6 groundwork - Document the configuration options in Search 6

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3473


documentation/src/main/asciidoc/configuration.asciidoc
```diff
@@ -1,19 +1,159 @@
...
+* `hibernate.search.default_backend`: defines the name of the backend used by default on all indexes.
+
++
+Other global properties are explained in the relevant parts of this documentation:
+
+* <<mapper-orm-configuration,Hibernate ORM integration>>
+
+Backend properties::
+These properties affect a single backend.
+They a grouped under a common root that includes the backend name: `hibernate.search.backends.<backend name>`.
```
gsmet: s/They a /They are /

documentation/src/main/asciidoc/configuration.asciidoc
```diff
@@ -1,19 +1,159 @@
...
+* <<backend-elasticsearch-configuration,Elasticsearch backend>>
+
+Index properties::
+These properties affect either one or multiple indexes, depending on the root.
++
+With the root `hibernate.search.backends.<backend name>.index_defaults`,
+they set defaults for all indexes of the referenced backend.
++
+With the root `hibernate.search.indexes.<index name>`,
+they set defaults for all indexes of the referenced backend.
```
gsmet: Hmmm, not really? :)

documentation/src/main/asciidoc/mapper-orm-configuration.asciidoc
```diff
@@ -0,0 +1,25 @@
...
+[[search-configuration-event]]
+
+== Enabling the integration
+
+The Hibernate ORM integration is enabled by default as soon as it is present in the classpath.
+
+If for some reason you need to disable it,
+set the `hibernate.search.autoregister_listeners` <<configuration-property-types,boolean property>> to `false`.
+
+== Configuration property tracking
```
gsmet: How is this paragraph related to ORM?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/cfg/ElasticsearchIndexSettings.java
```diff
@@ -8,16 +8,41 @@
...
 	public static final String LIFECYCLE_STRATEGY = "lifecycle.strategy";
 
+	/**
+	 * The minimal required status of the index on startup, before Hibernate Search can start using it.
+	 * <p>
+	 * Expects an {@link ElasticsearchIndexStatus} value, or a String representation of such value.
+	 * <p>
+	 * Defaults to {@link Defaults#LIFECYCLE_REQUIRED_STATUS}.
+	 */
 	public static final String LIFECYCLE_REQUIRED_STATUS = "lifecycle.required_status";
```
gsmet: In all the doc, you're talking about minimal required status. And it resonates with the discussion we had the other day.

I would change this one to `LIFECYCLE_MINIMAL_REQUIRED_STATUS` and the one below to `LIFECYCLE_MINIMAL_REQUIRED_STATUS_WAIT_TIMEOUT`

documentation/src/main/asciidoc/backend-lucene.asciidoc
```diff
@@ -1,4 +1,118 @@
...
+While Hibernate Search strives to offer a backwards compatible API making it easy to port your
+application to newer versions, it still delegates to Apache Lucene to handle the index writing and
+searching. This creates a dependency to the Lucene index format. The Lucene developers of course
+attempt to keep a stable index format, but sometimes a change in the format can not be avoided. In
+those cases you either have to re-index all your data or use an index upgrade tool. Sometimes Lucene
+is also able to read the old format so you don't need to take specific actions (besides making
+backup of your index).
+
+While an index format incompatibility is a rare event, it can happen more often that Lucene's
+Analyzer implementations might slightly change its behavior. This can lead to a poor recall score,
```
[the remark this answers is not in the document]
yrodiere: I copy-pasted this from Search 5, so I didn't write this. Judging from the style, I'd say it was Sanne's.
And the explanation of "poor recall score" is in the very next sentence: "possibly missing many hits from the results". In short, documents that should have matched would not match.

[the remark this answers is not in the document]
gsmet: Yeah, that's what I understood too. Just think that we could be clearer.

documentation/src/main/asciidoc/mapper-orm-configuration.asciidoc
yrodiere: The feature is implemented in the ORM integration, because in order to detect unused properties, we need a list of all properties, which engine SPIs do not require (we only need it for this particular feature). Other integrations may not be able to provide this feature, for example if properties are retrieved from a datasource that only provides find-by-key retrieval.

Anyway... Ok, I'll admit I don't have a very good point. I'll move this to the top-level "Configuration" section.

yrodiere:
Thanks! I addressed your comments and force-pushed. Waiting for CI.

PR #1894, opened 2019-02-18
HSEARCH-3484 + HSEARCH-3485 + HSEARCH-3491: JDK12 compatibility on branch 5.11

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3484
https://hibernate.atlassian.net//browse/HSEARCH-3485
https://hibernate.atlassian.net//browse/HSEARCH-3491
https://hibernate.atlassian.net//browse/HSEARCH-3494

Full build (including JDK12): http://ci.hibernate.org/job/hibernate-search/job/PR-1894/4/
Note that OSGi and WildFly integration tests are disabled on JDK12, because Karaf and WildFly don't work on JDK12 yet.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -323,7 +323,7 @@
         <version.surefire.plugin>2.21.0</version.surefire.plugin>
         <version.failsafe.plugin>${version.surefire.plugin}</version.failsafe.plugin>
         <version.org.wildfly.build-tools>1.2.9.Final</version.org.wildfly.build-tools>
-        <version.jacoco.plugin>0.8.0</version.jacoco.plugin>
+        <version.jacoco.plugin>0.8.3</version.jacoco.plugin>
         <version.coveralls.plugin>4.3.0</version.coveralls.plugin>
         <version.org.jboss.bridger.plugin>1.5.Final</version.org.jboss.bridger.plugin>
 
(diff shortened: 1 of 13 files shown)


pom.xml
```diff
@@ -387,6 +387,9 @@
         <surefire.jvm.args>${surefire.jvm.args.memory} ${surefire.jvm.args.java-version} ${surefire.jvm.args.jacoco}</surefire.jvm.args>
         <failsafe.jvm.args>${surefire.jvm.args.memory} ${surefire.jvm.args.java-version} ${failsafe.jvm.args.jacoco}</failsafe.jvm.args>
 
+        <failsafe.wildfly.skip>false</failsafe.wildfly.skip>
+        <failsafe.osgi.skip>false</failsafe.osgi.skip>
```
gsmet: It seems you forgot to set them to true for the JDK 12+ profile? Or did I miss something?

yrodiere: I did that in the Jenkinsfile. This way, if someone tries to build by hand, they will see that it doesn't work (yet) on WF and OSGi, but our CI tests will still check everything else.

gsmet: Hum. Not sure it's the desired behavior. That means people won't be able to build our code with the latest JDK.

yrodiere: That was not my goal at all... but ok, I'll do it your way.

yrodiere:
Merged. Thanks!

PR #1895, opened 2019-02-19
HSEARCH-3494 Allow to build the javadoc with JDK12

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3494


yrodiere:
Merged, thanks

PR #1897, opened 2019-02-20
HSEARCH-3493 Test compatibility with JDK13

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3493

I only had to submit a PR to the elasticsearch-maven-plugin so that we can run Elasticsearch with a different JDK. Fortunately it was accepted and released right away, so here is a fully compatible build: everything else already worked.

~I'll start a full build soon to demonstrate that it works.~ => Done, see here: http://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-1897/2/pipeline

For now, I'd rather keep the JDK13 as experimental (= not run on master) and run a separate weekly job to check compatibility, because I expect most of the breaking changes in JDK13 are still to come.


yrodiere:
Note the build only failed because of AWS support, which isn't implemented yet. So everything is fine.

fax4ever:
Rebased and merged thanks

yrodiere:
Thanks!

PR #1898, opened 2019-02-20
Backport HSEARCH-3493 to branch 5.11 - Test compatibility with JDK13

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3493

Backport of #1897.

Note that tests don't pass, so I kept the build as experimental (= not run on CI builds except when explicitly requested).
But all the errors seem related to Byteman not supporting JDK13 yet (see https://developer.jboss.org/message/988091).


yrodiere:
#1897 was approved; merging.

PR #1900, opened 2019-02-25
HSEARCH-3500 Manage bytebuddy version following HHH-13283

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3500

Currently running full ITs, including in JDK10. Let's wait for the result.


yrodiere:
It worked; merged.

PR #1906, opened 2019-03-04
HSEARCH-3509 Document the limitations of the date/time conversion in the Pojo mapper for java.util.Calendar and java.util.Date

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3509


fax4ever:
@yrodiere good work! I have just a suggestion.
Shall we add to the documentation that using Java legacy type is not recommended?

yrodiere:
Ok, I'll add that and merge. Thanks

PR #1907, opened 2019-03-05
HSEARCH-3257 Allow to bypass bridges in projections

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3257

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchSearchProjectionBuilderFactory.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchSearchProjectionBuilderFactory.java
@@ -19,6 +19,7 @@ import org.hibernate.search.backend.elasticsearch.search.impl.IndexSchemaFieldNo
 import org.hibernate.search.backend.elasticsearch.types.projection.impl.ElasticsearchFieldProjectionBuilderFactory;
 import org.hibernate.search.engine.search.SearchProjection;
 import org.hibernate.search.engine.search.predicate.spi.DslConverter;
+import org.hibernate.search.engine.search.projection.spi.ProjectionConverter;
 import org.hibernate.search.engine.search.projection.spi.CompositeProjectionBuilder;
 import org.hibernate.search.engine.search.projection.spi.DistanceToFieldProjectionBuilder;
 import org.hibernate.search.engine.search.projection.spi.DocumentReferenceProjectionBuilder;
@@ -57,10 +58,10 @@ public class ElasticsearchSearchProjectionBuilderFactory implements SearchProjec
 	}
 
 	@Override
-	public <T> FieldProjectionBuilder<T> field(String absoluteFieldPath, Class<T> expectedType) {
+	public <T> FieldProjectionBuilder<T> field(String absoluteFieldPath, Class<T> expectedType, ProjectionConverter projectionConverter) {
 		return searchTargetModel
 				.getSchemaNodeComponent( absoluteFieldPath, PROJECTION_BUILDER_FACTORY_RETRIEVAL_STRATEGY )
-				.createFieldValueProjectionBuilder( absoluteFieldPath, expectedType );
+				.createFieldValueProjectionBuilder( absoluteFieldPath, expectedType, projectionConverter );
 	}
 
 	@Override
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/projection/impl/ElasticsearchGeoPointFieldProjectionBuilderFactory.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/projection/impl/ElasticsearchGeoPointFieldProjectionBuilderFactory.java
@@ -14,6 +14,7 @@ import org.hibernate.search.backend.elasticsearch.search.projection.impl.Elastic
 import org.hibernate.search.backend.elasticsearch.types.codec.impl.ElasticsearchFieldCodec;
 import org.hibernate.search.engine.backend.types.converter.FromDocumentFieldValueConverter;
 import org.hibernate.search.engine.reporting.spi.EventContexts;
+import org.hibernate.search.engine.search.projection.spi.ProjectionConverter;
 import org.hibernate.search.engine.search.projection.spi.DistanceToFieldProjectionBuilder;
 import org.hibernate.search.engine.search.projection.spi.FieldProjectionBuilder;
 import org.hibernate.search.engine.spatial.GeoPoint;
@@ -26,10 +27,7 @@ public class ElasticsearchGeoPointFieldProjectionBuilderFactory implements Elast
 	private final boolean projectable;
 
 	private final FromDocumentFieldValueConverter<? super GeoPoint, ?> converter;
-
-	// TODO passing rawConverter to the projection builder
 	private final FromDocumentFieldValueConverter<? super GeoPoint, GeoPoint> rawConverter;
-
 	private final ElasticsearchFieldCodec<GeoPoint> codec;
 
 	public ElasticsearchGeoPointFieldProjectionBuilderFactory(boolean projectable,
@@ -44,15 +42,16 @@ public class ElasticsearchGeoPointFieldProjectionBuilderFactory implements Elast
 	@Override
 	@SuppressWarnings("unchecked") // We check the cast is legal by asking the converter
 	public <T> FieldProjectionBuilder<T> createFieldValueProjectionBuilder(String absoluteFieldPath,
-			Class<T> expectedType) {
+			Class<T> expectedType, ProjectionConverter projectionConverter) {
 		checkProjectable( absoluteFieldPath, projectable );
 
-		if ( !converter.isConvertedTypeAssignableTo( expectedType ) ) {
+		FromDocumentFieldValueConverter<? super GeoPoint, ?> requestConverter = getConverter( projectionConverter );
+		if ( !requestConverter.isConvertedTypeAssignableTo( expectedType ) ) {
 			throw log.invalidProjectionInvalidType( absoluteFieldPath, expectedType,
 					EventContexts.fromIndexFieldAbsolutePath( absoluteFieldPath ) );
 		}
 
-		return (FieldProjectionBuilder<T>) new ElasticsearchFieldProjectionBuilder<>( absoluteFieldPath, converter, codec );
+		return (FieldProjectionBuilder<T>) new ElasticsearchFieldProjectionBuilder<>( absoluteFieldPath, requestConverter, codec );
 	}
 
 	@Override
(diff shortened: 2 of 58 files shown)


engine/src/main/java/org/hibernate/search/engine/search/dsl/projection/impl/DefaultSearchProjectionFactoryContext.java
```diff
@@ -45,14 +46,26 @@ public DocumentReferenceProjectionContext documentReference() {
...
+	@Override
+	public <T> FieldProjectionContext<T> rawField(String absoluteFieldPath, Class<T> clazz) {
+		Contracts.assertNotNull( clazz, "clazz" );
+
+		return new FieldProjectionContextImpl<>( factory, absoluteFieldPath, clazz, ProjectionConverter.DISABLED );
+	}
+
+	@Override
+	public FieldProjectionContext<Object> rawField(String absoluteFieldPath) {
+		return field( absoluteFieldPath, Object.class );
```
yrodiere: This seems wrong, it should be `rawField( absoluteFieldPath, Object.class )`. And tests do not fail, so you're missing a test here.

engine/src/main/java/org/hibernate/search/engine/search/dsl/projection/SearchProjectionFactoryContext.java
```diff
@@ -72,6 +72,30 @@
 	 */
 	FieldProjectionContext<Object> field(String absoluteFieldPath);
 
+	/**
+	 * Project to a field of the indexed document.
+	 *
+	 * Unlike {@link #field(String, Class)} this method bypass any {@code DslConverter} defined on the field,
+	 * so that the values will be provided exactly as they are stored in the backend.
```
yrodiere: Please don't mention DSL converters here. I suppose it's an oversight and you already know this, because I already told you so multiple times on the chat, but DSL converters are unrelated to projection converters:

* DSL converters convert parameters passed to the DSL to index field values, typically from the type of the entity properties, so `JavaUtilDateBridge` would convert a `java.util.Date` to an `Instant`.
* Projection converters convert projected field values to a type expected by the user, typically to the type of the entity properties, so `JavaUtilDateBridge` would convert an `Instant` to a `java.util.date`.

Here is a perhaps better javadoc for this method (please adapt it for the other method below). Ideally we should explain converters more in detail, but that'll have to wait:

```
	/**
	 * Project to a field of the indexed document
	 * <p>
	 * Using this method instead of {@link #field(String, Class)} will disable some of the conversion applied to the field value,
	 * allowing to retrieve values directly from the backend.
	 *
	 * @param absoluteFieldPath The absolute path of the field.
	 * @return A context allowing to define the projection more precisely.
	 */
```

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchSearchProjectionBuilderFactory.java
```diff
@@ -19,6 +19,7 @@
 import org.hibernate.search.backend.elasticsearch.types.projection.impl.ElasticsearchFieldProjectionBuilderFactory;
 import org.hibernate.search.engine.search.SearchProjection;
 import org.hibernate.search.engine.search.predicate.spi.DslConverter;
+import org.hibernate.search.engine.search.predicate.spi.ProjectionConverter;
```
yrodiere: `ProjectionConverter` should be in the `projection` package... `org.hibernate.search.engine.search.projection.spi.ProjectionConverter`

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/projection/impl/ElasticsearchGeoPointFieldProjectionBuilderFactory.java
```diff
@@ -44,15 +42,15 @@ public ElasticsearchGeoPointFieldProjectionBuilderFactory(boolean projectable,
 	@Override
 	@SuppressWarnings("unchecked") // We check the cast is legal by asking the converter
 	public <T> FieldProjectionBuilder<T> createFieldValueProjectionBuilder(String absoluteFieldPath,
-			Class<T> expectedType) {
+			Class<T> expectedType, ProjectionConverter projectionConverter) {
 		checkProjectable( absoluteFieldPath, projectable );
 
-		if ( !converter.isConvertedTypeAssignableTo( expectedType ) ) {
+		if ( projectionConverter.isEnabled() && !converter.isConvertedTypeAssignableTo( expectedType ) ) {
```
yrodiere: You should check whether the type is assignable in all cases. Just, when the projection converter is disabled, you will call `isConvertedTypeAssignableTo` on the *raw* converter.

Note that your tests didn't fail, so you're missing a test where you try to pass the wrong type to `rawField(String, Class)`.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/projection/impl/ElasticsearchStandardFieldProjectionBuilderFactory.java
```diff
@@ -43,15 +42,15 @@ public ElasticsearchStandardFieldProjectionBuilderFactory(boolean projectable,
 	@Override
 	@SuppressWarnings("unchecked") // We check the cast is legal by asking the converter
 	public <T> FieldProjectionBuilder<T> createFieldValueProjectionBuilder(String absoluteFieldPath,
-			Class<T> expectedType) {
+			Class<T> expectedType, ProjectionConverter projectionConverter) {
 		checkProjectable( absoluteFieldPath, projectable );
 
-		if ( !converter.isConvertedTypeAssignableTo( expectedType ) ) {
+		if ( projectionConverter.isEnabled() && !converter.isConvertedTypeAssignableTo( expectedType ) ) {
```
yrodiere: You should check whether the type is assignable in all cases. Just, when the projection converter is disabled, you will call `isConvertedTypeAssignableTo` on the *raw* converter.

Note that your tests didn't fail, so you're missing a test where you try to pass the wrong type to `rawField(String, Class)`.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/projection/impl/LuceneGeoPointFieldProjectionBuilderFactory.java
```diff
@@ -44,15 +42,15 @@ public LuceneGeoPointFieldProjectionBuilderFactory(boolean projectable,
 	@Override
 	@SuppressWarnings("unchecked") // We check the cast is legal by asking the converter
 	public <T> FieldProjectionBuilder<T> createFieldValueProjectionBuilder(String absoluteFieldPath,
-			Class<T> expectedType) {
+			Class<T> expectedType, ProjectionConverter projectionConverter) {
 		checkProjectable( absoluteFieldPath, projectable );
 
-		if ( !converter.isConvertedTypeAssignableTo( expectedType ) ) {
+		if ( projectionConverter.isEnabled() && !converter.isConvertedTypeAssignableTo( expectedType ) ) {
```
yrodiere: You should check whether the type is assignable in all cases. Just, when the projection converter is disabled, you will call `isConvertedTypeAssignableTo` on the *raw* converter.

Note that your tests didn't fail, so you're missing a test where you try to pass the wrong type to `rawField(String, Class)`.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/projection/impl/LuceneStandardFieldProjectionBuilderFactory.java
```diff
@@ -47,15 +46,15 @@ public LuceneStandardFieldProjectionBuilderFactory(boolean projectable,
 	@Override
 	@SuppressWarnings("unchecked") // We check the cast is legal by asking the converter
 	public <T> FieldProjectionBuilder<T> createFieldValueProjectionBuilder(String absoluteFieldPath,
-			Class<T> expectedType) {
+			Class<T> expectedType, ProjectionConverter projectionConverter) {
 		checkProjectable( absoluteFieldPath, projectable );
 
-		if ( !converter.isConvertedTypeAssignableTo( expectedType ) ) {
+		if ( projectionConverter.isEnabled() && !converter.isConvertedTypeAssignableTo( expectedType ) ) {
```
yrodiere: You should check whether the type is assignable in all cases. Just, when the projection converter is disabled, you will call `isConvertedTypeAssignableTo` on the *raw* converter.

Note that your tests didn't fail, so you're missing a test where you try to pass the wrong type to `rawField(String, Class)`.

engine/src/main/java/org/hibernate/search/engine/search/dsl/projection/impl/DefaultSearchProjectionFactoryContext.java
fax4ever: You're right! There's a very dangerous typo here!

I've fixed it and added a test for using `rawField` without expressing a return type. 
Pre-exist test used `rawField`, always expressing the return type.

engine/src/main/java/org/hibernate/search/engine/search/dsl/projection/SearchProjectionFactoryContext.java
fax4ever: Yep quoting `DSL converters` here was just an oversight.

fax4ever: I was intending bypass the `DSL converter`'s **dual** => `Projection converter`

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchSearchProjectionBuilderFactory.java
fax4ever: you're right! makes sense

engine/src/main/java/org/hibernate/search/engine/search/dsl/projection/impl/DefaultSearchProjectionFactoryContext.java
yrodiere: Nice idea to have changed the existing test for `.field` too.

engine/src/main/java/org/hibernate/search/engine/search/dsl/projection/SearchProjectionFactoryContext.java
yrodiere: You didn't update the other `rawField` method... Will do.

yrodiere:
Hu, I just noticed that your PR has `HSEARCH-3257` in the title, `HSEARCH-3512` in the body, and `HSEARCH-3256` in the commit messages...

I'll fix that.

engine/src/main/java/org/hibernate/search/engine/search/dsl/projection/SearchProjectionFactoryContext.java
fax4ever: Thanks!

fax4ever:
Thanks

yrodiere:
Merged, thanks!

PR #1908, opened 2019-03-05
HSEARCH-3512 throw an exception when the field is non projectable due to the bridge

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3512
Based on top of https://github.com/hibernate/hibernate-search/pull/1907.
Issue is scoped on commits having comments having prefix `HSEARCH-3512`.

This is the way I understood the HSEARCH-3512.
@yrodiere, I would like to present this solution, even if it is probably wrong :P

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
@@ -460,4 +460,8 @@ public interface Log extends BasicLogger {
 	@Message(id = ID_OFFSET_3 + 53,
 			value = "Text predicates (phrase, fuzzy, wildcard, simple query string) are not supported by this field's type.")
 	SearchException textPredicatesNotSupportedByFieldType(@Param EventContext context);
+
+	@Message(id = ID_OFFSET_3 + 54, value = "Projection on field `%1$s` is not allowed, since bridge did not specify a bind back for this field. " +
+			"Please define a projection converter or use .rawField(`%1$s`) to extract the raw values bypassing the bridge.")
+	SearchException projectionConverterRequired(String absoluteFieldPath, @Param EventContext context);
 }
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/AbstractElasticsearchIndexFieldTypeConverterContext.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/AbstractElasticsearchIndexFieldTypeConverterContext.java
@@ -57,9 +57,8 @@ abstract class AbstractElasticsearchIndexFieldTypeConverterContext<S extends Abs
 
 	final FromDocumentFieldValueConverter<? super F, ?> createIndexToProjectionConverter() {
 		/*
-		 * TODO HSEARCH-3257 when no projection converter is configured, create a projection converter that will throw an exception
-		 * with an explicit message.
-		 * Currently we create a pass-through converter because users have no way to bypass the converter.
+		 * TODO HSEARCH-3512 Throw an exception when the field is non projectable due to the bridge.
+		 * E.g. if the field is bridged and the bridge did not specify a "FromIndexFieldValueConverter".
 		 */
 		return indexToProjectionConverter == null ? createFromDocumentRawConverter() : indexToProjectionConverter;
 	}
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/AbstractElasticsearchIndexFieldTypeConverterContext.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/AbstractElasticsearchIndexFieldTypeConverterContext.java
@@ -56,11 +56,18 @@ abstract class AbstractElasticsearchIndexFieldTypeConverterContext<S extends Abs
 	}
 
 	final FromDocumentFieldValueConverter<? super F, ?> createIndexToProjectionConverter() {
-		/*
-		 * TODO HSEARCH-3512 Throw an exception when the field is non projectable due to the bridge.
-		 * E.g. if the field is bridged and the bridge did not specify a "FromIndexFieldValueConverter".
-		 */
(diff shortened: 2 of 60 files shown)


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -460,4 +460,8 @@ SearchException conflictingIdentifierTypesForPredicate(ToDocumentIdentifierValue
 	@Message(id = ID_OFFSET_3 + 53,
 			value = "Text predicates (phrase, fuzzy, wildcard, simple query string) are not supported by this field's type.")
 	SearchException textPredicatesNotSupportedByFieldType(@Param EventContext context);
+
+	@Message(id = ID_OFFSET_3 + 54, value = "Projection on field `%1$s` is not allowed, since bridge did not specify a bind back for this field. " +
+			"Please define a projection converter or use .rawField(`%1$s`) to extract the raw values bypassing the bridge.")
```
yrodiere: Bridges are a mapper-specific concept, I'd rather avoid mentioning them in the backend:

```
"Projection on field '%1$s' is not allowed, since no projection converter was defined for this field. "+
"Please define a projection converter or use .rawField(\"%1$s\") to extract the raw values, bypassing the projection converter."
```

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/AbstractElasticsearchIndexFieldTypeConverterContext.java
```diff
@@ -56,11 +56,18 @@ final ElasticsearchIndexFieldTypeBuildContext getBuildContext() {
...
-		 * TODO HSEARCH-3512 Throw an exception when the field is non projectable due to the bridge.
-		 * E.g. if the field is bridged and the bridge did not specify a "FromIndexFieldValueConverter".
-		 */
-		return indexToProjectionConverter == null ? createFromDocumentRawConverter() : indexToProjectionConverter;
+		if ( indexToProjectionConverter != null ) {
+			// this case corresponds to a two-way bridge
+			return indexToProjectionConverter;
+		}
+
+		if ( dslToIndexConverter != null ) {
```
yrodiere: Again... the DSL converter is completely unrelated and should not be mentioned here. This entire method can be replaced with:

```
return indexToProjectionConverter;
```

That's all.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/projection/impl/ElasticsearchGeoPointFieldProjectionBuilderFactory.java
```diff
@@ -45,6 +45,11 @@ public ElasticsearchGeoPointFieldProjectionBuilderFactory(boolean projectable,
 			Class<T> expectedType, ProjectionConverter projectionConverter) {
 		checkProjectable( absoluteFieldPath, projectable );
 
+		if ( projectionConverter.isEnabled() && converter == null ) {
+			throw log.projectionConverterRequired( absoluteFieldPath,
```
yrodiere: :+1:

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/logging/impl/Log.java
```diff
@@ -418,4 +418,8 @@ SearchException conflictingIdentifierTypesForPredicate(ToDocumentIdentifierValue
 	@Message(id = ID_OFFSET_2 + 70,
 			value = "Text predicates (phrase, fuzzy, wildcard, simple query string) are not supported by this field's type.")
 	SearchException textPredicatesNotSupportedByFieldType(@Param EventContext context);
+
+	@Message(id = ID_OFFSET_2 + 71, value = "Projection on field `%1$s` is not allowed, since bridge did not specify a bind back for this field. " +
+			"Please define a projection converter or use .rawField(`%1$s`) to extract the raw values bypassing the bridge.")
```
yrodiere: Same as for Elasticsearch, this would be more fitting:

```
"Projection on field '%1$s' is not allowed, since no projection converter was defined for this field. "+
"Please define a projection converter or use .rawField(\"%1$s\") to extract the raw values, bypassing the projection converter."
```

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/dsl/impl/AbstractLuceneStandardIndexFieldTypeContext.java
```diff
@@ -76,11 +76,18 @@ protected final LuceneIndexFieldTypeBuildContext getBuildContext() {
...
+			return indexToProjectionConverter;
+		}
+
+		if ( dslToIndexConverter != null ) {
+			// this case corresponds to a one-way bridge
+			return null;
+		}
+
+		// this case corresponds to no bridge
+		return createFromDocumentRawConverter();
```
yrodiere: Same as for Elasticsearch, you can replace the whole method with:


```
return indexToProjectionConverter;
```

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/projection/FieldSearchProjectionIT.java
```diff
@@ -288,6 +288,44 @@ public void withProjectionConverters_rawValues() {
 		}
 	}
 
+	@Test
+	public void withDslConverters_rawValues() {
```
yrodiere: The tests shouldn't be about DSL converters. They should be about projecting with `field()` when no projection converter is defined (this should fail) or projecting with `rawField()` when no projection converter is defined (this should work).

fax4ever:
@yrodiere you know.

My goal here was just to avoid grabbing the raw values on projection without using rawField.

For instance when I have DSLConvertion and not a ProjecitonConverter (one way bridge), data is different (raw vs original), using byField would grab raw data.
In my opinion, but this is very personal, it would be better throw an exception and invite the use to use rawField instead of field.

Anyway, I think that's not critical at all.
So it is fine to mark the ticket as rejected ;)

PR #1909, opened 2019-03-06
HSEARCH-3403 + HSEARCH-3511 Polish the entry points to the Search DSL and make them more consistent

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3403: Remove the inheritance from our Session/Query APIs to Hibernate ORM/JPA APIs.
https://hibernate.atlassian.net//browse/HSEARCH-3511: Make the entry points to the Search API more intuitive

Regarding HSEARCH-3403, note I ended up removing the inheritance in `FullTextSession` too, and exposing two methods: `.toHibernateOrmSession()`/`toJpaEntityManager()`, so that the use case mentioned in the ticket is still covered.

Regarding HSEARCH-3511, I'd really like @gsmet to have a look, please, since you were the one to mention to me that the APIs were not that intuitive when trying to use them without having a look at the documentation. I think I addressed your main concern, but I went a bit beyond that, so it would be great if you could have a quick look at the resulting syntax.

In a nutshell:

* We consistently use the `Search` prefix in API types.
* We use `toQuery` to build the query (consistent with predicates, sorts, etc.).
* The concept of "target" has been renamed to "scope": the scope of the search.
* There are now two separate entry points to the query DSL: the simpler one, `.search()`, will work just fine with the lambda-based syntax. The more "advanced" one, `.scope()`, will be necessary for the object-based syntax, and we may one day add more operations than just search there: mass indexer, delete queries, ...
* The methods on the `SearchQuery` interface no longer follow the same naming scheme as JPA/ORM: they are named `fetchXXX()` and offsets/limits are passed through parameters instead of setters. The reasons are:
  1. JPA naming does not play well with how versatile a "result" can be in Elasticsearch/Lucene. A result is not just hits, it can also include the total hit count, aggregations, facets, ... so having a method named `getResultList` can be confusing. Even if we decided to deal with it, finding appropriate names would be quite challenging...
  2. The methods actually retrieve data from disk or the network. Naming them `getXXX` gives the impression that you're just calling a getter.
  3. Some methods are not affected with the limit/offset (e.g. `fetchTotalHitCount`)... but passing their value through setters gives the impression that every method in `SearchQuery` is affected.

You can find a few examples in the tests, `org.hibernate.search.integrationtest.mapper.orm.search.SearchQueryIT` in particular, but below are some simple ones.

I changed this:

```
FullTextSession ftSession = Search.getFullTextSession( session );
FullTextSearchTarget<Book> target = ftSession.search( Book.class );
SearchQuery<String> query = target.query()
		.asProjection( target.projection().field( "title", String.class ).toProjection() )
		.predicate( target.predicate().matchAll() )
		.build();
List<String> results = query.fetchHits();
```

to this:

```
SearchSession searchSession = Search.getSearchSession( session );
SearchScope<Book> scope = searchSession.scope( Book.class );
SearchQuery<String> query = scope.search()
		.asProjection( scope.projection().field( "title", String.class ).toProjection() )
		.predicate( scope.predicate().matchAll() )
		.toQuery();
List<String> results = query.fetchHits();
```

And I introduced a shortcut so that the lambda syntax (the one we will recommend for most use cases) feels more natural, so this:

```
FullTextQuery<Book> query = ftSession.search( Book.class ).query()
		.asEntity()
		.predicate( f -> f.matchAll() )
		.build();
```

became this:

```
SearchQuery<Book> query = searchSession.search( Book.class ) // No trailing .query() call
		.asEntity()
		.predicate( f -> f.matchAll() )
		.toQuery();
```


yrodiere:
@gsmet , @fax4ever : there's a lot of refactoring here, so please don't let the large number of changes scare you. The large commits (the ones with "Rename X to Y" in the commit message) are generally not worth reviewing in details as long as you agree with their purpose.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/hibernate/FullTextQuery.java
```diff
@@ -8,12 +8,8 @@
...
 
-public interface FullTextQuery<T> extends Query<T>, org.hibernate.search.mapper.orm.jpa.FullTextQuery<T> {
+public interface FullTextQuery<T> extends org.hibernate.search.mapper.orm.jpa.FullTextQuery<T> {
 
-	// TODO facets
-
-	// TODO timeouts
-
-	// TODO explain
+	Query<T> toHibernateOrmQuery();
```
yrodiere: Suggested by @gsmet : `toOrmQuery` instead of this.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/hibernate/FullTextSession.java
```diff
@@ -11,12 +11,14 @@
...
-public interface FullTextSession extends Session, FullTextEntityManager {
+public interface FullTextSession extends FullTextEntityManager {
 
 	@Override
 	<T> FullTextSearchTarget<T> search(Class<T> type);
 
 	@Override
 	<T> FullTextSearchTarget<T> search(Collection<? extends Class<? extends T>> types);
 
+	Session toHibernateOrmSession();
```
yrodiere: Suggested by @gsmet : `toOrmSession` instead of this.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/jpa/FullTextEntityManager.java
```diff
@@ -11,11 +11,14 @@
...
-public interface FullTextEntityManager extends EntityManager {
+public interface FullTextEntityManager {
 
 	<T> FullTextSearchTarget<T> search(Class<T> type);
 
 	<T> FullTextSearchTarget<T> search(Collection<? extends Class<? extends T>> types);
 
 	MassIndexer createIndexer(Class<?>... types);
+
+	EntityManager toJpaEntityManager();
```
yrodiere: Suggested by @gsmet : `toEntityManager` instead of this.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/query/SearchQuery.java
```diff
@@ -72,19 +72,35 @@
...
+	 * @return {@code this} for method chaining.
+	 */
+	SearchQuery<T> setFirstResult(Long firstResult);
+
+	/**
+	 * @deprecated Pass a {@code long} value and use {@link #setFirstResult(Long)} instead.
 	 * @param firstResult The offset of the first result. Must be positive or zero.
 	 * @return {@code this} for method chaining.
 	 */
+	@Deprecated
```
gsmet: Should we remove them instead? My rationale is that people will have a tendency to use them considering you don't need the `L` and they would have a warning.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/impl/FullTextQueryImpl.java
```diff
@@ -69,11 +72,26 @@ else if ( type.equals( SearchQuery.class ) ) {
 		}
 	}
 
+	@Override
+	public TypedQuery<R> toJpaQuery() {
```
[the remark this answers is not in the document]
yrodiere: It mainly provides more resilience to API changes in ORM, which can be useful since the `Query` contract is not meant to be implemented by third parties, thus they could add methods unexpectedly.

Also, it allows to more easily implement methods such as `getSingleResult()` in a way that throws the correct exceptions.

Since we're splitting the interfaces, I could create a wrapper class, separate from this one, and instantiate it when `toJpaQuery` is called? That would clean up the code of `SearcheQueryImpl` quite a lot.
The main drawback is that we wouldn't notice when our own interfaces deviate from JPA/ORM interfaces in an incompatible way (same signature with different return type). But I suppose it's not such a big deal, since we're unlikely to go back to making our interfaces extend JPA/ORM interfaces.

mapper/javabean/src/main/java/org/hibernate/search/mapper/javabean/session/SearchSession.java
```diff
@@ -9,27 +9,66 @@
...
+	 * Initiate the building of a search query.
+	 * <p>
+	 * The query will target the indexes mapped to the given type, or to any of its sub-types.
+	 *
+	 * @param type An indexed type, or a supertype of all indexed types that will be targeted by the search query.
+	 * @return A context allowing to define the search query,
+	 * and ultimately {@link SearchQueryContext#toQuery() get the resulting query}.
+	 * @see SearchQueryResultDefinitionContext
+	 */
 	default <T> SearchQueryResultDefinitionContext search(Class<T> type) {
```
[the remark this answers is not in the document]
yrodiere: Yes: generics in varargs lead to call-site warnings ("heap pollution yada yada"). Usually the solution to avoid these warnings is to add `@SafeVarargs`, but this can only be used on static or final methods, so not here in an interface.

For the JavaBean mapper, we could get rid of the generics, but not for the ORM mapper, where they are actually useful.

Our only other option would be to declare one method per number of parameters: one type, two types, etc. I'm really not thrilled by the idea.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/query/SearchQuery.java
```diff
@@ -14,22 +14,102 @@
...
+	 */
 	SearchQuery<T> setFirstResult(int firstResult);
 
+	/**
+	 * Set the JDBC fetch size for this query.
+	 *
+	 * @param fetchSize The fetch size. Must be positive or zero.
+	 * @return {@code this} for method chaining.
+	 * @see Query#setFetchSize(int)
+	 */
```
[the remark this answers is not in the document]
yrodiere: > is it used for loading the elements from the database

Yes.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/query/SearchQuery.java
```diff
@@ -14,22 +14,102 @@
...
+	/**
+	 * Convert this query to a {@link TypedQuery JPA query}.
+	 * <p>
+	 * Note that the resulting query <strong>does not support all operations</strong>
+	 * and may behave slightly differently in some cases
+	 * (including, but not limited to, the type of thrown exceptions).
+	 * For these reasons, it is recommended to only use this method
+	 * when integrating to an external library that expects JPA queries.
+	 *
+	 * @return A representation of this query as a JPA query.
```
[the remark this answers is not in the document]
yrodiere: I'm not sure I'm following, we did agree it was useful to be able to use Hibernate Search queries through JPA/ORM interfaces, didn't we? I just wanted to move this away from our main interfaces because we know that it will always be just a hack and we cannot implement all the methods of `TypedQuery`.

It's not a JPA query, it's a representation of a Hibernate Search query as a JPA query. The only purpose of this method is to allow using JPA interfaces to manipulate Hibernate Search queries, thereby integrating more easily into existing frameworks that are used to Hibernate ORM.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/query/SearchQuery.java
```diff
@@ -15,9 +15,19 @@
...
-	 * Execute the query and retrieve the results as a {@link List}
+	 * Execute the query and retrieve the results as a {@link SearchResult}.
 	 *
-	 * @return The results as a {@link List} containing one element for each matched entity.
+	 * @return The results.
+	 * @throws org.hibernate.search.util.common.SearchException If something goes wrong while executing the query.
+	 * @throws org.hibernate.HibernateException If something goes wrong while fetching results from the database.
+	 * @throws javax.persistence.PersistenceException If something goes wrong while fetching results from the database.
+	 */
+	SearchResult<T> getResult();
```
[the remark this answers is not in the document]
yrodiere: So it seems we're on the same page, but just a reminder for readers: what we call "result" in Hibernate Search is the whole thing: hits, total hit count, aggregations, facets, ... While a "hit" is only the part of the result that represents a document matching the query.

I agree the names you proposed are more correct. What I was trying to do here is to still more or less fit the JPA naming, so as not to be too confusing.

If we are to give up on that, then yes, I would pick different names. But I would go even further, because I want to make it extra clear that calling one of these methods does not just "get" the hits from a Java object that is already there, fully populated: it actually executes the query.
I would also get rid of the `getSingleHit` variant, since returning an Optional provides the same feature, and more.

So ideally, if we forget about using the same naming as JPA, I would use:

* `SearchResult<T> fetch()`
* `List<T> fetchHits()` (=> `fetch().getHits()`)
* `Optional<T> fetchSingleHit()`  (=> `fetch().getHits()`, but returns an Optional for 0 or 1 hit, throws an exception if more than 1 hit)
* `long fetchCount()` (=> `fetch().getHitCount()`, but optimized to not fetch any hit)
* variants of `fetch()` and `fetchHits` accepting two paramters to handle paging: `fetch(long limit)`, `fetch(long limit, long offset)`... **OR** at least rename `setFirstResult` to `setHitOffset` and `setMaxResults` to `setHitLimit`.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/query/SearchQuery.java
yrodiere: If I remove them, people trying to pass an `int` (either a literal without a trailing `L`, or a variable) will get a compilation error. There is no automatic boxing from `int` to `Long`, just from `long` to `Long`.

If I removed the `int` variants and changed the parameter type from `Long` to `long`, then users would benefit from automatic casting from `int` to `long`. But then they couldn't reset the value by passing `null` to `setFirstResult` or `setMaxResults` anymore. Which I suppose is fine, since they could pass `0` to `setFirstResult` and `Long.MAX_VALUE` to setMaxResults`?

mapper/javabean/src/main/java/org/hibernate/search/mapper/javabean/session/SearchSession.java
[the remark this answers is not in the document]
gsmet: Hum. This is annoying :/.

What makes them useful for the ORM case?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/query/SearchQuery.java
gsmet: Well, what I mean is: either you support it and you don't have them deprecated or you don't.

But what you're saying in your first paragraph means they will end up with a deprecation warnings.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/query/SearchQuery.java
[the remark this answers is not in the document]
gsmet: TBH, I don't know.

I suppose people would expect an API similar to what they are used to with JPA but it's really 2 completely different things.

Another thing is that a lot of people are using layers on top of JPA so maybe they are not that used to the JPA API.

I like your `fetch(limit, offset)` proposal.

mapper/javabean/src/main/java/org/hibernate/search/mapper/javabean/session/SearchSession.java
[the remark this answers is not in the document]
yrodiere: The generics, you mean? In short it allows for typed queries when requesting an entity result type. Without it, we wouldn't know which type to put in `SearchQuery<T>`.
It's not useful in the JavaBean mapper because it cannot load entities from the database, since there isn't any database.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/query/SearchQuery.java
yrodiere: Ok, so I suppose "them" meant the `@Deprecated` annotations, not the methods. Anyway... It's probably not relevant anymore if we switch to a `fetch(limit, offset)` method.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/query/SearchQuery.java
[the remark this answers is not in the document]
yrodiere: I added a few commits to use `fetch(...)` as I proposed.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/query/SearchQuery.java
yrodiere: I removed the deprecation, we now have two variants of each `fetch` method: one accepting longs and one accepting integers.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/impl/FullTextQueryImpl.java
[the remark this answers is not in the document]
yrodiere: I split the implementations and moved the adaptor to a separate class.

integrationtest/showcase/library/src/main/java/org/hibernate/search/integrationtest/showcase/library/repository/indexsearch/IndexSearchDocumentRepositoryImpl.java
```diff
@@ -52,22 +49,18 @@
...
+						.toQuery();
 
-		return Optional.ofNullable( query.uniqueResult() );
+		return query.fetchSingleHit();
 	}
 
 	@Override
-	public List<Book> searchByMedium(String terms, BookMedium medium, int offset, int limit) {
-		FullTextQuery<Book> query = Search.getFullTextEntityManager( entityManager ).search( Book.class ).query()
+	public List<Book> searchByMedium(String terms, BookMedium medium, long offset, long limit) {
```
fax4ever: Maybe we should change the order of the parameter `offset` and ` limit` into `limit` and `offset`, in order to be more consist with the new Search API.

integrationtest/showcase/library/src/main/java/org/hibernate/search/integrationtest/showcase/library/repository/indexsearch/IndexSearchPersonRepositoryImpl.java
```diff
@@ -24,56 +22,46 @@
 	private EntityManager entityManager;
 
 	@Override
-	public List<Person> listTopBorrowers(int offset, int limit) {
+	public List<Person> listTopBorrowers(long offset, long limit) {
```
fax4ever: Same as above, maybe it would be better move limit param before offset.

integrationtest/showcase/library/src/main/java/org/hibernate/search/integrationtest/showcase/library/repository/indexsearch/IndexSearchPersonRepositoryImpl.java
```diff
@@ -24,56 +22,46 @@
...
 		return listTopBorrowers( "account.borrowals.totalCount", offset, limit );
 	}
 
 	@Override
-	public List<Person> listTopShortTermBorrowers(int offset, int limit) {
+	public List<Person> listTopShortTermBorrowers(long offset, long limit) {
```
fax4ever: Same as above, maybe it would be better move limit param before offset.

integrationtest/showcase/library/src/main/java/org/hibernate/search/integrationtest/showcase/library/repository/indexsearch/IndexSearchPersonRepositoryImpl.java
```diff
@@ -24,56 +22,46 @@
...
 		return listTopBorrowers( "account.borrowals.shortTermCount", offset, limit );
 	}
 
 	@Override
-	public List<Person> listTopLongTermBorrowers(int offset, int limit) {
+	public List<Person> listTopLongTermBorrowers(long offset, long limit) {
```
fax4ever: Same as above, maybe it would be better move limit param before offset.

integrationtest/showcase/library/src/main/java/org/hibernate/search/integrationtest/showcase/library/repository/indexsearch/IndexSearchPersonRepositoryImpl.java
```diff
@@ -24,56 +22,46 @@
...
 		return listTopBorrowers( "account.borrowals.longTermCount", offset, limit );
 	}
 
 	@Override
-	public List<Person> searchPerson(String terms, int offset, int limit) {
+	public List<Person> searchPerson(String terms, long offset, long limit) {
```
fax4ever: Same as above, maybe it would be better move limit param before offset.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/query/impl/HibernateOrmSearchQuery.java
```diff
@@ -0,0 +1,111 @@
...
+
+public class HibernateOrmSearchQuery<R> implements SearchQuery<R> {
+
+	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
+
+	private final IndexSearchQuery<R> delegate;
+	private final SessionImplementor sessionImplementor;
+	private final MutableObjectLoadingOptions loadingOptions;
+
+	private HibernateOrmSearchQueryAdapter<R> wrapper;
```
fax4ever: Maybe you could call `adapter` the field name or `HibernateOrmSearchQueryWrapper` the class. I would prefer the former option.

yrodiere:
@fax4ever Thanks for the review. I applied all of your suggestions.

Waiting for CI, then I'll merge.

yrodiere:
Merged. Thanks a lot for your reviews, @gsmet and @fax4ever !

PR #1910, opened 2019-03-07
HSEARCH-3503 Add dedicated default value bridges for java.sql.Timestamp, java.sql.Date and java.sql.Time

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3503

It's incredible that `java.sql.Date` and even `java.sql.Time` behaves exact as `java.sql.Timestamp`.

The methods `java.sql.Date#toLocalDate`, ` java.sqlTimestamp#toLocalDateTime` and so on have a lot of issue relative to the default JVM timezones applied on the conversion.
Thus I avoided to use them or their reverse.


fax4ever:
Added them to the table types and quoted them under `Support for legacy java.util date/time APIs' chapter.
Since I think that we have similar issue with them.

yrodiere:
Crap, I saw you used the wrong ticket number in your last commit too late... Please mind next time?
Anyway, it's merged, thanks.

fax4ever:
Sorry, I did not notice it.

PR #1917, opened 2019-03-18
HSEARCH-3498 Move optional predicate DSL parameters to the terminal contexts

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3498

@gsmet Any opinion on this?

The motivations and some examples are in the ticket descriptions.

@fax4ever @gsmet I'd be interested in your opinion on HSEARCH-3522, too?


engine/src/main/java/org/hibernate/search/engine/search/dsl/predicate/impl/RangePredicateFieldSetContextImpl.java
```diff
@@ -92,8 +92,8 @@ public void contributePredicateBuilders(Consumer<B> collector) {
 		}
 	}
 
-	static class CommonState<B> extends AbstractBooleanMultiFieldPredicateCommonState<B, RangePredicateFieldSetContextImpl<B>>
-			implements RangePredicateTerminalContext {
+	static class CommonState<B> extends AbstractBooleanMultiFieldPredicateCommonState<CommonState<B>, B, RangePredicateFieldSetContextImpl<B>>
+			implements RangePredicateTerminalContext, RangePredicateLimitTerminalContext {
```
fax4ever: Maybe we could remove the first interface after implements here.

yrodiere: Done

yrodiere:
I addressed some comments. Here are my answers:

> The issue mentions (see Pros#3) that we would be able to do more validations, such as validate fuzzy is not applied to an Integer field. Do we have some tests for these cases?

It's not about making more validations, since we already had this validation. It's about throwing the exception at the right time, i.e. when `fuzzy()` is called instead of when `matching` is called. Yes we have a test in `org.hibernate.search.integrationtest.backend.tck.search.predicate.MatchSearchPredicateIT#error_fuzzy_unsupportedFieldType`.

> Moreover, I was thinking that maybe it would be nice change MultiFieldPredicateFieldSetContext#boostedTo method name in something different, like boostField, in order to avoid confusion between predicate and field level boosts.

Not sure it's a great name, but feel free to create a ticket so that we think about it.

yrodiere:
Merged, thanks!

PR #1920, opened 2019-03-20
HSEARCH-3136  Throw an exception when an element of @IndexedEmbedded.includePaths did not match anything

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3136

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/ConfiguredIndexSchemaNestingContext.java
+++ b/engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/ConfiguredIndexSchemaNestingContext.java
@@ -6,6 +6,8 @@
  */
 package org.hibernate.search.engine.mapper.mapping.building.impl;
 
+import java.util.LinkedHashSet;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.BiFunction;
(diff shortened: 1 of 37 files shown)


engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/ConfiguredIndexSchemaNestingContext.java
```diff
@@ -91,6 +93,11 @@ public String toString() {
 			while ( nextDotIndex >= 0 ) {
 				String objectName = prefixToParse.substring( afterPreviousDotIndex, nextDotIndex );
 				contextBuilder.appendObject( objectName );
+
+				// Make sure to mark the paths as encountered in the filter
+				String objectNameRelativeToFilter = prefixToParse.substring( 0, nextDotIndex );
+				filter.isPathIncluded( objectNameRelativeToFilter );
```
gsmet: I would use the name `markPathIncluded` or `markPathAsIncluded`.

gsmet:
Oh and it needs a rebase.

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/ConfiguredIndexSchemaNestingContext.java
yrodiere: The thing is, this method returns a boolean telling you whether the path is included or not. Here's I'm just twisting it: I know it has a side effect, which is marking the path as encountered, so I use it just for the side effect.

Every where else we really mainly use the method to know whether the path is included or not.

I'll add a comment to clarify.

yrodiere:
Merged, thanks!

PR #1927, opened 2019-03-26
HSEARCH-3521 Move index-specific configuration properties below the hibernate.search.backends.<backend name>.indexes property

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3521

### DIFF
--- a/documentation/src/main/asciidoc/configuration.asciidoc
+++ b/documentation/src/main/asciidoc/configuration.asciidoc
@@ -72,14 +72,6 @@ For ORM entities, the default index name is the name of the indexed class, witho
 Index names can be customized in the mapping.
 
 +
-Notable properties:
-
-* `hibernate.search.indexes.<index name>.backend`: the backend to use for this index.
-Set this to the name of your backend.
-Note that you do not need to set this explicitly for each index,
-you can simply set the global property `hibernate.search.default_backend` (see above).
-
-+
 Examples:
 
 * `hibernate.search.backends.myBackend.index_defaults.lifecycle.strategy = validate`
--- a/documentation/src/main/asciidoc/configuration.asciidoc
+++ b/documentation/src/main/asciidoc/configuration.asciidoc
@@ -63,10 +63,11 @@ These properties affect either one or multiple indexes, depending on the root.
 +
 With the root `hibernate.search.backends.<backend name>.index_defaults`,
 they set defaults for all indexes of the referenced backend.
+The backend name must match the name defined in the mapping.
 +
-With the root `hibernate.search.indexes.<index name>`,
+With the root `hibernate.search.backends.<backend name>.indexes.<index name>`,
 they set the value for a specific index, overriding the defaults (if any).
-The index name must match the name defined in the mapping.
+The backend and index names must match the names defined in the mapping.
 For ORM entities, the default index name is the name of the indexed class, without the package:
 `org.mycompany.Book` will have `Book` as its default index name.
 Index names can be customized in the mapping.
--- a/engine/src/main/java/org/hibernate/search/engine/common/impl/IndexManagerBuildingStateHolder.java
+++ b/engine/src/main/java/org/hibernate/search/engine/common/impl/IndexManagerBuildingStateHolder.java
@@ -131,6 +131,7 @@ class IndexManagerBuildingStateHolder {
 	class BackendInitialBuildState<D extends DocumentElement> {
 		private final String backendName;
 		private final BackendBuildContext backendBuildContext;
+		private final ConfigurationPropertySource backendPropertySource;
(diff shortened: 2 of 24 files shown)


documentation/src/main/asciidoc/configuration.asciidoc
```diff
@@ -63,22 +63,15 @@ These properties affect either one or multiple indexes, depending on the root.
 +
 With the root `hibernate.search.backends.<backend name>.index_defaults`,
 they set defaults for all indexes of the referenced backend.
+The backend name must match the name defined in the mapping.
 +
-With the root `hibernate.search.indexes.<index name>`,
+With the root `hibernate.search.backends.<backend name>.indexes.<index name>`,
 they set the value for a specific index, overriding the defaults (if any).
-The index name must match the name defined in the mapping.
+The backend and index name must match the names defined in the mapping.
```
gsmet: s/backend and index name/backend and index names/ maybe?

engine/src/main/java/org/hibernate/search/engine/common/impl/IndexManagerBuildingStateHolder.java
```diff
@@ -163,22 +142,32 @@ private BackendInitialBuildState(
...
-			IndexedEntityBindingContext bindingContext = new IndexedEntityBindingContextImpl( schemaRootNodeBuilder );
-			return new IndexManagerInitialBuildState<>( backendName, indexName, builder, bindingContext );
+		public IndexManagerInitialBuildState<?> getIndexManagerBuildingState(
+				String indexName, boolean multiTenancyEnabled) {
+			IndexManagerInitialBuildState<?> state = indexManagerBuildStateByName.get( indexName );
+			if ( state == null ) {
+				ConfigurationPropertySource indexPropertySource =
+						EngineConfigurationUtils.getIndex( backendPropertySource, defaultIndexPropertySource, indexName );
+				ConfigurationPropertySource defaultedIndexPropertySource =
+						EngineConfigurationUtils.addIndexDefaults( indexPropertySource, defaultIndexPropertySource );
```
yrodiere: I just noticed this: it's not necessary and should be removed. The line above already adds the default configuration as a fallback.

engine/src/main/java/org/hibernate/search/engine/common/impl/IndexManagerBuildingStateHolder.java
```diff
@@ -71,42 +68,23 @@
...
-		BackendInitialBuildState<?> backendBuildingstate =
-				backendBuildStateByName.computeIfAbsent( backendName, this::createBackend );
-
-		IndexManagerInitialBuildState<?> state = indexManagerBuildStateByName.get( indexName );
-		if ( state == null ) {
-			state = backendBuildingstate.createIndexManagerBuildingState(
-					indexName, multiTenancyEnabled, indexPropertySource
-			);
-			indexManagerBuildStateByName.put( indexName, state );
+	public BackendInitialBuildState<?> getBackend(String backendName) {
```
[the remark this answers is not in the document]
yrodiere: This class is package-private, so it doesn't matter much. I'll reduce the visibility of this method to be clearer.

yrodiere:
Parsing the property keys to deduce the backend to use for a given index would not be easy given how configuration currently works: it's a sort of key-value datastore, with no possibility to list all the keys or to query the keys. Changing that in a consistent way (taking into account the masks, prefixes, etc.) would require a significant amount of effort.

On top of that, there are several corner cases that would require workarounds:

* What if the user didn't set any configuration property for this index? Then there's no configuration property beginning with `hibernate.search.backends.es.indexes.Book`.
* What if the user wants his application to work with two backends depending on a switch of his, like we did in the showcase? He will probably configure both backends in the configuration properties and rely on the default backend property to select the backend to use. Which means there will potentially be a `hibernate.search.backends.es.indexes.Book.something` property key, and a `hibernate.search.backends.lucene.indexes.Book.something` property key. Which one do we choose?

That all sounds very complicated. Let's have a look later if you really think it's a good idea, but that sounds like a lot of risk and I'm not sure there will be a lot of benefits.

yrodiere:
Thanks for the reviews; I addressed the comments.
Waiting for CI.

yrodiere:
Merged, thanks!

fax4ever:
Hi @yrodiere! In my mind:
* If the user didn't set any configuration property for this index, it would be needed just to add an empty property such as `hibernate.search.backends.es.indexes.Book`.
* For the case of application working with two backends, I think that we would have the same problem with the annotation. Wouldn't we?

Anyway I don't think that it is be very critical, since 99% of the time the user defines a default backend for all the indexes. :P

PR #1930, opened 2019-03-27
HSEARCH-3538 Disabling projection converters in projections is ignored when checking compatibility

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3538


fax4ever:
I think it's perfect. You complete very well the work on dsl / projection converter. Now it is very symmetrical and consistent. I'll merge it in a moment.

PR #1931, opened 2019-03-27
HSEARCH-3540 The NONE Elasticsearch index lifecycle strategy throws an AssertionFailure

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3540


yrodiere:
Jenkins, retest this please

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/management/ElasticsearchSchemaNoneStrategyIT.java
```diff
@@ -0,0 +1,89 @@
...
+	public TestElasticsearchClient elasticsearchClient = new TestElasticsearchClient();
+
+	/**
+	 * Just check that the NONE strategy works correctly and does not involve any call to the Elasticsearch cluster
+	 * (except the one to get the Elasticsearch version, but that one is not caused by the lifecycle strategy).
+	 */
+	@Test
+	@TestForIssue(jiraKey = "HSEARCH-3540")
+	public void noCall() {
+		elasticsearchClient.index( INDEX_NAME )
```
fax4ever: Do we need this call too? Since it seems that test works fine even without it. This block could generate some confusion, because the log shows index creation call. I would keep it only if we needed to work on this index.

yrodiere: You're right.
I intended to also send some requests to check that everything works right even if we don't touch the index, but I ended up not doing it.
I'll remove this.

fax4ever:
Merged, thanks.

PR #1933, opened 2019-03-28
HSEARCH-3070 Remove the concept of ServiceManager

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3070


yrodiere:
Merged, thanks

PR #1934, opened 2019-03-28
HSEARCH-3051 Restore the Elasticsearch AWS integration

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3051

~Based on #1933 (HSEARCH-3070), which should be merged first.~ => Done

Full build (including AWS ES 5.6 and AWS ES 6.4): http://ci.hibernate.org/view/Search/job/hibernate-search/view/change-requests/job/PR-1934/9/

Note the coverage drop is artificial: the number doesn't take into account the executions of our integration tests against AWS...


backend/elasticsearch-aws/src/main/java/org/hibernate/search/backend/elasticsearch/aws/cfg/ElasticsearchAwsBackendSettings.java
```diff
@@ -0,0 +1,67 @@
...
+	 * No default: must be provided when signing is enabled.
+	 */
+	public static final String SECRET_KEY = "aws.secret_key";
+
+	/**
+	 * The AWS region.
+	 * <p>
+	 * Expects a String value such as {@code us-east-1}.
+	 * <p>
+	 * No default: must be provided when signing is enabled.
```
gsmet: It's only needed when signing is enabled?

backend/elasticsearch-aws/src/main/java/org/hibernate/search/backend/elasticsearch/aws/cfg/ElasticsearchAwsBackendSettings.java
```diff
@@ -0,0 +1,67 @@
...
+
+	/**
+	 * Whether requests should be signed using the AWS credentials.
+	 * <p>
+	 * Expects a Boolean value such as {@code true} or {@code false},
+	 * or a string that can be parsed to such Boolean value.
+	 * <p>
+	 * Defaults to {@link Defaults#SIGNING_ENABLED}.
+	 */
+	public static final String SIGNING_ENABLED = "aws.signing.enabled";
```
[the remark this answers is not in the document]
yrodiere: Ok, I'll add `.signing.` everywhere.

backend/elasticsearch-aws/src/main/java/org/hibernate/search/backend/elasticsearch/aws/cfg/ElasticsearchAwsBackendSettings.java
yrodiere: Yes. AWS simply rejects all requests if you don't sign them anyway, so the region is not useful if you don't sign your requests.

PR #1935, opened 2019-03-29
HSEARCH-3546 SearchException is missing from the javadoc

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3546


yrodiere:
I need this fixed to work on https://hibernate.atlassian.net/browse/HSEARCH-3375, and it's not like there's a lot to review: I'll go ahead and merge it now.

PR #1936, opened 2019-04-01
HSEARCH-2464 Add support for "missing"/"exists" predicates

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2464

Note that:

* "missing" predicates are not supported explicitly: you just need to use a "must not" clause containing an "exists" query in a boolean junction. Just like any other predicate.
* "exists" predicates on object fields are out of scope and will be addressed in [HSEARCH-2389](https://hibernate.atlassian.net/browse/HSEARCH-2389)


integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/LuceneExtensionIT.java
```diff
@@ -285,6 +285,19 @@ public void predicate_nativeField_nativeQuery() {
 				.hasDocRefHitsAnyOrder( INDEX_NAME, FIRST_ID );
 	}
 
+	@Test
+	public void predicate_nativeField_exists() {
+		StubMappingSearchScope scope = indexManager.createSearchScope();
+
+		IndexSearchQuery<DocumentReference> query = scope.query()
+				.asReference()
+				.predicate( f -> f.exists().onField( "nativeField" ) )
```
gsmet: `onField` does not feel natural for `exists`? But maybe consistency is better?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/AbstractLuceneNumericFieldCodec.java
```diff
@@ -0,0 +1,82 @@
...
+			doEncodeForProjection( documentBuilder, absoluteFieldPath, value, encodedValue );
+		}
+
+		LuceneNumericDomain<E> domain = getDomain();
+
+		if ( sortable ) {
+			documentBuilder.addField( domain.createDocValuesField( absoluteFieldPath, encodedValue ) );
+		}
+		else {
+			// For createExistsQuery()
```
fax4ever: I agree with this optimization idea. Is there a way to apply it even if the field isn't sortable?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/predicate/impl/AbstractLuceneFieldPredicateBuilderFactory.java
```diff
@@ -69,4 +92,13 @@ public LuceneSimpleQueryStringPredicateBuilderFieldContext createSimpleQueryStri
 				EventContexts.fromIndexFieldAbsolutePath( absoluteFieldPath )
 		);
 	}
+
+	@Override
+	public ExistsPredicateBuilder<LuceneSearchPredicateBuilder> createExistsPredicateBuilder(String absoluteFieldPath) {
+		return new LuceneExistsPredicateBuilder( absoluteFieldPath, getCodec() );
+	}
+
+	protected abstract LuceneFieldCodec<?> getCodec();
```
fax4ever: I think that's a good idea. Why do not move `codec` and `converter` field in there too?

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/LuceneExtensionIT.java
yrodiere: Yes I'd keep it for consistency for now. If we end up implementing [HSEARCH-3522](https://hibernate.atlassian.net/browse/HSEARCH-3522), we might change the syntax to `.field( "name" ).exists()`, which in my opinion would be much better, but there should be some debate before we go there.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/document/impl/LuceneRootDocumentBuilder.java
```diff
@@ -42,6 +50,10 @@ public LuceneIndexEntry build(String indexName, MultiTenancyStrategy multiTenanc
 		rootDocument.add( new StringField( LuceneFields.indexFieldName(), indexName, Store.YES ) );
 		rootDocument.add( new StringField( LuceneFields.idFieldName(), id, Store.YES ) );
 
+		for ( String fieldName : fieldNames ) {
+			rootDocument.add( new StringField( LuceneFields.fieldNamesFieldName(), fieldName, Store.NO ) );
```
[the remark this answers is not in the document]
yrodiere: It's not overridden, each new `StringField` instance just adds a new value. That's how Lucene works, and that's what we want here :)

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/AbstractLuceneNumericFieldCodec.java
yrodiere: No, we need docvalues in order to avoid indexing the field name. And currently we only have docvalues when the field is sortable.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/predicate/impl/AbstractLuceneFieldPredicateBuilderFactory.java
yrodiere: Because subtypes require the codec and converter to have a narrower type, so it's easier to just let them declare a field it with the right type.

fax4ever:
@yrodiere +1 to merge it, as soon we've solved the issue with the master.

yrodiere:
@fax4ever Let's not merge just yet, as Guillaume mentioned some tests are missing. I'm on it.

yrodiere:
@gsmet

> It seems to rely on some internal behavior of Lucene.

As far as I can tell it does not, the query classes are in public packages and are not marked as experimental.
The features they provide don't seem that obscure, either: they just check if norms or docvalues exist. Both concepts are exposed to users, not internal ones.

> Or do we have tests for all the corner cases introduced in the optimization?

Now we do.

yrodiere:
Merged, thanks!

PR #1940, opened 2019-04-02
HSEARCH-3532 + HSEARCH-3550 Java module fixes (not enough, but that's a start)

yrodiere:
* [HSEARCH-3532](https://hibernate.atlassian.net//browse/HSEARCH-3532): Fix the java module name for hibernate-search-util-common
* [HSEARCH-3550](https://hibernate.atlassian.net//browse/HSEARCH-3550): Fix the many problems of java module integration tests

It turns out we were not running the java module tests properly. Even when we were indeed compiling the code as a module, the tests used to put this module in the classpath... so all module configuration was ignored.

Unfortunately, it turns out Hibernate Search does not currently work when a module depends on it. We get the following error:

```
org.hibernate.service.UnknownServiceException: Unknown service requested [org.hibernate.search.mapper.orm.impl.HibernateSearchContextService]
at org.hibernate.search.integrationtest.java.module/org.hibernate.search.integrationtest.java.module.service.AuthorServiceIT.test(AuthorServiceIT.java:29)
```

See [here](http://ci.hibernate.org/blue/organizations/jenkins/hibernate-search-yoann/detail/HSEARCH-3550/2/pipeline#step-85-log-637) for a failing build with this error.
See [HSEARCH-3551](https://hibernate.atlassian.net/browse/HSEARCH-3551) for a follow-up ticket that describes the problem.

For now I had to disable the test execution. The modules are still compiled, though, so we test just as much as we used to.


pom.xml
```diff
@@ -262,7 +262,7 @@
         <version.org.jboss.bridger.plugin>1.5.Final</version.org.jboss.bridger.plugin>
         <version.com.buschmais.jqassistant.plugin>1.5.0</version.com.buschmais.jqassistant.plugin>
         <version.com.github.alexcojocaru.elasticsearch.plugin>6.12</version.com.github.alexcojocaru.elasticsearch.plugin>
-        <version.moditech.plugin>1.0.0.Beta2</version.moditech.plugin>
```
gsmet: Hmmm, it wasn't used anywhere?

yrodiere: No...

PR #1941, opened 2019-04-02
HSEARCH-3507 Improve consistency of naming for methods related to native Elasticsearch/Lucene features (fromJsonString, etc.)

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3507

I ended up going for minimal changes, so I did not fix *everything* that was mentioned in the ticket. But I think it's the best compromise between consistency and user-friendliness (because `fromJson` is definitely clearer than `byNative`, for example).

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/ElasticsearchIndexFieldTypeFactoryContext.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/ElasticsearchIndexFieldTypeFactoryContext.java
@@ -14,6 +14,6 @@ import org.hibernate.search.engine.backend.types.dsl.IndexFieldTypeFactoryContex
  */
 public interface ElasticsearchIndexFieldTypeFactoryContext extends IndexFieldTypeFactoryContext {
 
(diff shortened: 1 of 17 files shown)


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/ElasticsearchIndexFieldTypeFactoryContext.java
```diff
@@ -14,6 +14,33 @@
...
+	 * @param jsonString A string representing an Elasticsearch field mapping as a JSON object.
+	 * The JSON object must be a syntactically correct Elasticsearch field mapping.
+	 * See <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping.html">the Elasticsearch documentation</a>.
+	 * @return A context allowing to get the resulting predicate.
+	 */
+	/**
+	 *
+	 * @param mappingJsonString
+	 * @return
+	 */
```
gsmet: There is an issue with the javadoc. You have 2 javadoc blocks and in the first one `jsonString` should be `mappingJsonString`.

PR #1943, opened 2019-04-03
HSEARCH-3548 + HSEARCH-3549 + HSEARCH-3555 + HSEARCH-3557 Tests and fixes for indexing problems in Elasticsearch and Lucene

yrodiere:
* [HSEARCH-3548](https://hibernate.atlassian.net//browse/HSEARCH-3548): ZonedDateTime with zone = ZoneOffset.UTC fails when indexing in Elasticsearch 5.6
* [HSEARCH-3549](https://hibernate.atlassian.net//browse/HSEARCH-3549): Indexing MonthDay.of( 2, 29 ) fails on Elasticsearch 5.6
* [HSEARCH-3555](https://hibernate.atlassian.net//browse/HSEARCH-3555): Indexing Year.of( Year.MAX_VALUE ) fails with Elasticsearch 5
* [HSEARCH-3557](https://hibernate.atlassian.net//browse/HSEARCH-3557): ZonedDateTime at later offset during DST change gets parsed incorrectly

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/format/impl/Elasticsearch6DefaultFieldFormatProvider.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/format/impl/Elasticsearch6DefaultFieldFormatProvider.java
@@ -58,8 +58,17 @@ public class Elasticsearch6DefaultFieldFormatProvider implements ElasticsearchDe
 		map.put( LocalDateTime.class, asImmutableList( "yyyy-MM-dd'T'HH:mm:ss.SSS", "yyyyyyyyy-MM-dd'T'HH:mm:ss.SSSSSSSSS" ) );
 		map.put( OffsetTime.class, asImmutableList( "HH:mm:ss.SSSZZ", "HH:mm:ss.SSSSSSSSSZZ" ) );
 		map.put( OffsetDateTime.class, asImmutableList( "yyyy-MM-dd'T'HH:mm:ss.SSSZZ", "yyyyyyyyy-MM-dd'T'HH:mm:ss.SSSSSSSSSZZ" ) );
-		// ZoneRegionId is optional for ZonedDateTime, but we need the offset to handle ambiguous date/times at DST overlap
-		map.put( ZonedDateTime.class, asImmutableList( "yyyy-MM-dd'T'HH:mm:ss.SSSZZ'['ZZZ']'", "yyyyyyyyy-MM-dd'T'HH:mm:ss.SSSSSSSSSZZ'['ZZZ']'" ) );
+		/*
+		 * ZoneRegionId is optional for ZonedDateTime, but we need the offset to handle ambiguous date/times at DST overlap.
+		 *
+		 * Also, in some cases the ZoneId might actually be a ZoneOffset (since ZoneOffset extends ZoneId),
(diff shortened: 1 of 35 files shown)


yrodiere:
I almost did it, but on second thought I'll keep 0 as the leap year used for MonthDay. I know I'd be pissed to see some random year pop up in my data as a user, so let's run the tests and if it works on ES5, 6 and 7 I think we're safe.

Tests running.

yrodiere:
Tests pass (well there's a problem with AWS, but it's a problem on master). I'll merge.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/format/impl/Elasticsearch6DefaultFieldFormatProvider.java
```diff
@@ -58,15 +58,22 @@
...
 		 */
-		map.put( MonthDay.class, asImmutableList( "--MM-dd" ) );
+		map.put( ZonedDateTime.class, asImmutableList(
+				"yyyy-MM-dd'T'HH:mm:ss.SSSZZ'['ZZZ']'",
+				"yyyyyyyyy-MM-dd'T'HH:mm:ss.SSSSSSSSSZZ'['ZZZ']'",
+				"yyyyyyyyy-MM-dd'T'HH:mm:ss.SSSSSSSSSZZ'['ZZ']'"
+		) );
+		map.put( Year.class, asImmutableList( "yyyy", "yyyyyyyyy" ) );
+		map.put( YearMonth.class, asImmutableList( "yyyy-MM", "yyyyyyyyy-MM" ) );
+		map.put( MonthDay.class, asImmutableList( "yyyy-MM-dd" ) );
```
fax4ever: Do we have 2 put with the same key `MonthDay.class`?

documentation/src/main/asciidoc/backend-elasticsearch.asciidoc
```diff
@@ -368,10 +368,26 @@ or `yyyy\|\|yyyyyyyyy` (ES6 and below)
...
 `uuuu-MM` (ES7 and above)
 or `yyyy-MM\|\|yyyyyyyyy-MM` (ES6 and below)
-|java.time.MonthDay|`date` with format `--MM-dd`
+|java.time.MonthDay|`date` with format `uuuu-MM-dd` (ES7 and above)
+or `yyyy-MM-dd` (ES6 and below).
+**The year is always set to 0**.
 |org.hibernate.search.engine.spatial.GeoPoint|`geo_point`
 |====
 
+jshell> java.time.Instant.ofEpochMilli( Long.MIN_VALUE )
```
fax4ever: I would format this as Monospaced font

yrodiere: Wow, this shouldn't be in there. Good catch, I'll remove it.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/format/impl/Elasticsearch6DefaultFieldFormatProvider.java
yrodiere: Good catch, I'll remove one.

yrodiere:
Merged, thanks!

PR #1944, opened 2019-04-04
HSEARCH-3534 HSEARCH-3535  Default value  and out of bound message for "minimumShouldMatch"

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3534
https://hibernate.atlassian.net/browse/HSEARCH-3535

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchBooleanJunctionPredicateBuilder.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchBooleanJunctionPredicateBuilder.java
@@ -112,6 +112,11 @@ class ElasticsearchBooleanJunctionPredicateBuilder extends AbstractElasticsearch
 		contributeClauses( context, innerObject, SHOULD_ACCESSOR, shouldClauseBuilders );
 		contributeClauses( context, innerObject, FILTER_ACCESSOR, filterClauseBuilders );
 
+		// Forcing to the Lucene's defaults. See HSEARCH-3534
+		if ( minimumShouldMatchConstraints == null && hasAtLeastOneMustOrFilterPredicate() ) {
+			minimumShouldMatchNumber( 0, 0 );
+		}
+
 		if ( minimumShouldMatchConstraints != null ) {
 			MINIMUM_SHOULD_MATCH_ACCESSOR.set(
 					innerObject,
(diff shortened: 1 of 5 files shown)


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchBooleanJunctionPredicateBuilder.java
```diff
@@ -124,6 +129,10 @@ protected JsonObject doBuild(ElasticsearchSearchPredicateContext context,
 		return outerObject;
 	}
 
+	private boolean hasAtLeastOneMustPredicate() {
+		return mustClauseBuilders != null || mustNotClauseBuilders != null;
```
yrodiere: From the Elasticsearch 6.7 documentation:

> If the bool query is in a filter context or has neither must or filter then at least one of the should queries must match a document for it to match the bool query. This behavior may be explicitly controlled by setting the minimum_should_match parameter.

https://www.elastic.co/guide/en/elasticsearch/reference/6.7/query-dsl-bool-query.html

The documentation mentions `must` and `filter`, but your condition mentions `must` and `mustNot`. I wonder if it's accurate? Maybe you should add tests (in the same test method) with "one should, one filter" and "one should, one mustNot" in a filter context, just to be sure?

fax4ever: Good idea. Let's add the test so that to verify whether  `|| mustNotClauseBuilders != null` is really need.

fax4ever: You were right, according to the new test. A must-not is treated differently from a must.

yrodiere: And what about "filter"? Or maybe you're still working on it?

fax4ever: Sorry, I forgot it.
I think I've addressed it now.

yrodiere:
Merged, thanks

PR #1945, opened 2019-04-05
HSEARCH-3058 Search 6 groundwork - Add generic type parameters to PropertyHandle

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3058

Actually I'm not sure whether it is a good idea to make the method `PojoPropertyModel#getHandle` return a `PropertyHandle<? super T>` instead of just `PropertyHandle<T>`.

This was my first choice, since the field `GenericContextAwarePojoPropertyMode#rawPropertyModel` has a type `PojoPropertyModel<? super T>`. On the other hand, I'm not sure about the solution I applied.

### DIFF
--- a/mapper/javabean/src/main/java/org/hibernate/search/mapper/javabean/model/impl/JavaBeanBootstrapIntrospector.java
+++ b/mapper/javabean/src/main/java/org/hibernate/search/mapper/javabean/model/impl/JavaBeanBootstrapIntrospector.java
@@ -87,7 +87,7 @@ public class JavaBeanBootstrapIntrospector implements PojoBootstrapIntrospector 
 		return annotationHelper.getAnnotationsByMetaAnnotationType( annotatedElement, metaAnnotationType );
 	}
 
-	PropertyHandle createPropertyHandle(String name, Method method) throws IllegalAccessException {
+	PropertyHandle<?> createPropertyHandle(String name, Method method) throws IllegalAccessException {
 		return propertyHandleFactory.createForMethod( name, method );
 	}
 
--- a/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/model/impl/HibernateOrmPropertyModel.java
+++ b/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/model/impl/HibernateOrmPropertyModel.java
@@ -39,7 +39,7 @@ class HibernateOrmPropertyModel<T> implements PojoPropertyModel<T> {
  	 */
 	private final List<XProperty> declaredXProperties;
 
-	private PropertyHandle handle;
+	private PropertyHandle<T> handle;
 	private PojoGenericTypeModel<T> typeModel;
 
 	HibernateOrmPropertyModel(HibernateOrmBootstrapIntrospector introspector, HibernateOrmRawTypeModel<?> holderTypeModel,
@@ -90,10 +90,11 @@ class HibernateOrmPropertyModel<T> implements PojoPropertyModel<T> {
 	}
 
 	@Override
-	public PropertyHandle getHandle() {
+	@SuppressWarnings("unchecked") // By construction, we know the member returns values of type T
+	public PropertyHandle<T> getHandle() {
 		if ( handle == null ) {
 			try {
-				handle = introspector.createPropertyHandle( name, member );
+				handle = (PropertyHandle<T>) introspector.createPropertyHandle( name, member );
 			}
 			catch (IllegalAccessException | RuntimeException e) {
 				throw log.errorRetrievingPropertyTypeModel( getName(), holderTypeModel, e );
--- a/mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/impl/PropertyIdentifierMapping.java
+++ b/mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/impl/PropertyIdentifierMapping.java
@@ -22,10 +22,10 @@ import org.hibernate.search.util.common.impl.Closer;
 public class PropertyIdentifierMapping<I, E> implements IdentifierMapping<I, E> {
 
 	private final PojoCaster<? super I> caster;
-	private final PropertyHandle<? super I> property;
+	private final PropertyHandle<I> property;
 	private final BeanHolder<? extends IdentifierBridge<I>> bridgeHolder;
 
-	public PropertyIdentifierMapping(PojoCaster<? super I> caster, PropertyHandle<? super I> property,
+	public PropertyIdentifierMapping(PojoCaster<? super I> caster, PropertyHandle<I> property,
 			BeanHolder<? extends IdentifierBridge<I>> bridgeHolder) {
 		this.caster = caster;
 		this.property = property;
@@ -46,10 +46,7 @@ public class PropertyIdentifierMapping<I, E> implements IdentifierMapping<I, E> 
 		if ( providedId != null ) {
 			return (I) caster.cast( providedId );
 		}
-
-		Object id = property.get( entitySupplier.get() );
-		// TODO avoid this cast? By construction, the property handle should always return type I
-		return (I) caster.cast( id );
+		return property.get( entitySupplier.get() );
 	}
 
 	@Override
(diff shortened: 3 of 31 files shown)


mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/model/spi/PojoTypeModel.java
```diff
@@ -18,5 +18,5 @@
 	 */
 	PojoRawTypeModel<? super T> getRawType();
 
-	PojoPropertyModel<?> getProperty(String propertyName);
```
yrodiere: That will have to be a no from me.

What you're saying here is "this method can provide you with a `PojoPropertyModel<P>`, where P is whatever you want". Which is wrong; actually this method can provide callers with a `PojoPropertyModel<P>`, where `P` is some pre-defined type. Not "whatever they want".

So this method should actually return `PojoPropertyModel<?>`, and if callers need something else, they will have to handle it on the call site.

From what I can see, there's only one place where my suggestion would raise an issue: `org.hibernate.search.mapper.pojo.model.path.impl.BoundPojoModelPathTypeNode#property`. But even there, that's only a problem because you changed the signature from:

```
public BoundPojoModelPathPropertyNode<T, ?> property(PropertyHandle propertyHandle)
```

to:

```
public <P> BoundPojoModelPathPropertyNode<T, P> property(PropertyHandle<P> propertyHandle)
```

If you just use a wildcard instead of a generic, you should be fine:

```
public BoundPojoModelPathPropertyNode<T, ?> property(PropertyHandle<?> propertyHandle)
```

Callers do not need more anyway.

mapper/pojo/src/test/java/org/hibernate/search/mapper/pojo/dirtiness/building/impl/PojoAssociationPathInverterTest.java
```diff
@@ -189,6 +189,7 @@ public void detectInfiniteRecursion() {
 
 	private PropertyHandle setupPropertyStub(PojoTypeModel<?> holdingTypeMock, String propertyName,
 			PojoGenericTypeModel<?> propertyTypeMock) {
+		// It seems that we cannot add a generic type such a <? super P> here
```
yrodiere: Just use `PropertyHandle<?>` here, and cast it to `(PropertyHandle)` where you get the compilation error.
That's how we have to deal with generics when using mocks, unfortunately. You can find similar examples all over the place.

Also, please add the missing `<?>` at line 72 (there's a raw `PropertyHandle`).

mapper/javabean/src/main/java/org/hibernate/search/mapper/javabean/model/impl/JavaBeanBootstrapIntrospector.java
```diff
@@ -87,7 +87,7 @@ public JavaBeanBootstrapIntrospector(MethodHandles.Lookup lookup) {
 		return annotationHelper.getAnnotationsByMetaAnnotationType( annotatedElement, metaAnnotationType );
 	}
 
-	PropertyHandle<?> createPropertyHandle(String name, Method method) throws IllegalAccessException {
+	<T> PropertyHandle<T> createPropertyHandle(String name, Method method) throws IllegalAccessException {
```
yrodiere: As I mentioned in a previous comment, this is not a good idea. Here you're saying "the factory can provide the caller with a `PropertyHandle<T>`, and you can choose `T`, regardless of the `Method` passed". Which is wrong: `T` is predefined, depends on the `Method` passed, and cannot be chosen by the caller.

So this should really be:

```
PropertyHandle<?> createPropertyHandle(String name, Method method) throws IllegalAccessException {
```

Which is what you had before.

I know what you wanted to do was to solve another problem; I'll add a comment there.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/model/impl/HibernateOrmPropertyModel.java
```diff
@@ -93,8 +93,7 @@ public String getName() {
 	public PropertyHandle<T> getHandle() {
 		if ( handle == null ) {
 			try {
-				// TODO HSEARCH-3058
-				handle = (PropertyHandle<T>) introspector.createPropertyHandle( name, member );
+				handle = introspector.createPropertyHandle( name, member );
```
yrodiere: This effectively hides the problem under the carpet. The problem is that the introspector (and, below that, the factory) cannot guarantee that the property handle will be of type `PropertyHandle<T>` for any member. With this solution you could pass a method representing `toString` and request a `PropertyHandle<Integer>` and that would compile just fine. At runtime you would get an exception eventually, but it would compile.

There's no nice solution here, but I think there's a more readable one: remove this entire commit, and just replace the beginning of this method with this:

```
@SuppressWarnings("unchecked") // By construction, we know the member returns values of type T
public PropertyHandle<T> getHandle() {
		if ( handle == null ) {
			try {
				handle = (PropertyHandle<T>) introspector.createPropertyHandle( name, member );
```

And do something similar in `org.hibernate.search.mapper.javabean.model.impl.JavaBeanPropertyModel#getHandle`.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/model/spi/PojoPropertyModel.java
```diff
@@ -23,6 +23,6 @@
 	 */
 	PojoGenericTypeModel<T> getTypeModel();
 
-	PropertyHandle getHandle();
+	PropertyHandle<? super T> getHandle();
```
yrodiere: I think this should be `PropertyHandle<T>`.

I know that will cause problems in `GenericContextAwarePojoPropertyModel`; I'll add another comment there.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/model/spi/GenericContextAwarePojoPropertyMode.java
```diff
@@ -43,7 +43,7 @@ public String getName() {
 	}
 
 	@Override
-	public PropertyHandle getHandle() {
+	public PropertyHandle<? super T> getHandle() {
 		return rawPropertyModel.getHandle();
```
yrodiere: This should really return `PropertyHandle<T>`.

The thing is, the "raw property model" represents the property with its raw type, and this class represents the property with generics.

Let's say the property has type `List<U>` and is declared in a class `MyEntity<U>`. The raw property model will represent a property
of type `List`. A generic context aware propery model in the context of `MyEntity<Integer>` will represent a property of type `List<Integer>`.
So you will have a `GenericContextAwarePojoPropertyModel<List<Integer>>` which contains a rawPropertyModel of type `PojoPropertyModel<? super List<Integer>>` which is really a `PojoPropertyModel<List>`.
And we know that, in the context of `MyEntity<List<Integer>>`, the values contained in the property *can* be casted from `List` to `List<Integer>`. We have this knowledge right here, in this very class.

So this is the right place to perform the cast:

```
	@Override
	@SuppressWarnings("unchecked") // We know that, in the current generic context, this cast is legal
	public PropertyHandle<T> getHandle() {
		return (PropertyHandle<T>) rawPropertyModel.getHandle();
	}
```

fax4ever:
@yrodiere thanks for the nice explanation. 

If I got it well, the 'rawTypeModel' instance cannot be any `PojoRawTypeModel<? super T>`, it must have the same raw type of the generic type model.

The only difference can reside in the generic of the generic 'T' type.

I think I've addressed your corrections here.
Thank you for the review too.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/impl/PropertyIdentifierMapping.java
```diff
@@ -22,10 +22,10 @@
 public class PropertyIdentifierMapping<I, E> implements IdentifierMapping<I, E> {
 
 	private final PojoCaster<? super I> caster;
-	private final PropertyHandle property;
+	private final PropertyHandle<? super I> property;
```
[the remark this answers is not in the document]
yrodiere: You forgot this one§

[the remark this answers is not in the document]
fax4ever: Yes I did. Sorry. Addressed now.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/model/spi/PojoTypeModel.java
yrodiere: In `BoundPojoModelPathTypeNode` you just reverted your changes. I gave you the correct code, which includes a `<?>` instead of leaving a raw type.

I'll do it.

yrodiere:
I fixed an oversight and renamed a class that had a typo in its name. Waiting for CI then I'll merge.

fax4ever:
Thanks

yrodiere:
Merged, thanks!

PR #1947, opened 2019-04-08
HSEARCH-3099 HSEARCH-3553 Restore support for "indexNullAs" in @Field

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3099
https://hibernate.atlassian.net/browse/HSEARCH-3553


mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/programmatic/impl/PropertyGenericFieldMappingContextImpl.java
```diff
@@ -33,4 +41,18 @@ PropertyGenericFieldMappingContext thisAsS() {
...
 
+	@Override
+	protected <F> void contributeIndexNullAs(StandardIndexFieldTypeContext<?, F> contextConverter, String indexNullAs, Class<?> type) {
+		try {
+			Optional<Object> parsed = ParserUtils.parse( indexNullAs, type );
+			if ( !parsed.isPresent() ) {
+				throw log.unsupportedTypeForIndexNullAs( type );
+			}
+
+			contextConverter.indexNullAs( (F) parsed.get() );
```
yrodiere: There are several problems with this solution:

1. It's obviously not extensible. If the user wants to add support for a custom type by writing a custom bridge, he won't be able to use `@GenericField.indexNullAs`.
2. It lacks support for a lot of otherwise supported types. `Instant`, `java.util.Date`, basically all of `java.time`... All those types were supported in Search 5, why wouldn't they be in 6?
3. You ask the user to tell us the expected type, then you cast it to `F` and hope for the best. But we don't have *any* guarantee that `F` is the same type... If the user provides the wrong type, this will fail, and because the cast to `F` is ineffective (generic casts are just cosmetic, they don't do anything), this will fail in a very bizarre way.

So in my opinion, this is not the right approach. 

I think the support for converting from String to `F` should be built in `ValueBridge`. `ValueBridge` is where we know about both the property type and the indexed field type, it's where all the conversion happens and it's what we have to implement to add support for another property type. It feels natural that support for converting from a string to the correct type would be implemented there.

I'll suggest a solution, but some choices were a matter of taste, so feel free to challenge it:

* Expose a `F parse(String value)` method in `ValueBridge`.
* Add the appropriate javadoc to this method.
* Add a default implementation to this method that simply does `throw new UnsupportedOperationException( "Bridge " + toString() + " does not support parsing a value from a String" )`
* Implement this method in all built-in bridges. For the pass-through bridge, I would expect a `Function<String, F>` to be passed to the constructor, and it would be called when `parse` is called.
  Note you could reuse the `ConvertUtils` to generate method references that do what you want : `ConvertUtils::convertBoolean`, `ConvertUtils::convertLong`, etc., and add more methods as necessary. The additional methods may be useful one day for configuration properties.

Now you have extensible support for all property types, but it's not used anywhere. You still need to call this `parse` method:

* Move `org.hibernate.search.engine.mapper.mapping.building.spi.FieldModelContributor` to `org.hibernate.search.mapper.pojo.mapping.building.spi`. I don't know why it was in the engine, it should not be there.
* Change `org.hibernate.search.engine.mapper.mapping.building.spi.FieldModelContributor#contribute` so that it now takes two parameters: `fieldTypeContext` (the current parameter, renamed) and `bridgeContext` (with a new type, `FieldModelContributorBridgeContext`)
* Define `FieldModelContributorBridgeContext` with just one `indexNullAs(String)` method
* Change `org.hibernate.search.mapper.pojo.mapping.definition.programmatic.impl.PojoCompositeFieldModelContributor#contribute` to pass the bridge context to the delegates (which will now be `BiConsumers` instead of just `Consumers`).
* Define an implementation of `FieldModelContributorBridgeContext` that holds a reference to both the `ValueBridge<?, F>` and the `StandardIndexFieldTypeContext<?, F>`. It will be able to implement `indexNullAs(String value)` simply by calling `fieldTypeContext.indexNullAs( bridge.parse( value ) )`.
* Instantiate this implementation in `org.hibernate.search.mapper.pojo.mapping.building.impl.PojoIndexModelBinderImpl#bindValueBridge` add pass it to the field contributor (the line that reads `contributor.contribute( fieldTypeContext );`).

integrationtest/mapper/pojo/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/IndexNullAsValueIT.java
```diff
@@ -0,0 +1,227 @@
...
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.GenericField;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.Indexed;
+import org.hibernate.search.util.common.SearchException;
+import org.hibernate.search.util.impl.integrationtest.common.rule.BackendMock;
+import org.hibernate.search.util.impl.test.SubTest;
+
+import org.junit.Rule;
+import org.junit.Test;
+
+public class IndexNullAsValueIT {
```
yrodiere: There are many types that should be supported but are not, mainly the `java.time` types.

I know it's more verbose, but since this is a type-sensitive feature, I think it would be safer to rely on the `PropertyTypeDescriptors` to test it.

I think you should be able to just extend `org.hibernate.search.integrationtest.mapper.pojo.mapping.definition.FieldDefaultBridgeIT`. Simply add a new property in the mapped types (there is one for each `DefaultValueBridgeExpectations` implementation). Name the property `myPropertyWithIndexNullAs` for example, and add a `@GenericField` annotation with the appropriate "indexNullAs". You'll then be able to add a method in `FieldDefaultBridgeIT` that tests the nominal case.

To test for error cases, I think you'll be better off with a separate test. Let's talk about it once you get there?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/dsl/impl/AbstractLuceneStandardIndexFieldTypeContext.java
```diff
@@ -60,7 +63,7 @@ public S projectable(Projectable projectable) {
 
 	@Override
 	public S indexNullAs(F indexNullAs) {
-		// TODO implement indexNullAs in the backends
+		indexNullAsValue = Optional.of( indexNullAs );
```
yrodiere: `null` is not an acceptable value for `indexNullAs`. You can use `null` to represent "no value was selected", no need for an Optional.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/document/model/impl/LuceneIndexSchemaFieldNode.java
```diff
@@ -94,11 +96,20 @@ public String toString() {
...
 	}
 
 	private EventContext getEventContext() {
 		return EventContexts.fromIndexFieldAbsolutePath( absoluteFieldPath );
 	}
+
+	public void encode(AbstractLuceneDocumentBuilder documentBuilder, String absoluteFieldPath, F value) {
+		if ( value == null && indexNullAsValue.isPresent() ) {
+			value = indexNullAsValue.get();
+		}
```
yrodiere: To me it would make more sense to implement this in the codec itself... There's already a `if ( value == null )` block there, that would be the ideal place.

If you really want to put this in the schema node, which was initially supposed to simply hold metadata, I won't force your hand, but I suspect we'll have to revert it at some point.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/mapping/definition/IndexNullAsValueIT.java
```diff
@@ -0,0 +1,183 @@
...
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.DocumentId;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.GenericField;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.Indexed;
+import org.hibernate.search.util.impl.integrationtest.common.rule.BackendMock;
+import org.hibernate.search.util.impl.integrationtest.orm.OrmSetupHelper;
+
+import org.junit.Rule;
+import org.junit.Test;
+
+public class IndexNullAsValueIT {
```
yrodiere: This test, and this commit, are not really useful in my opinion. The classes involved are the same in both cases, so we shouldn't bother with testing them twice.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/AbstractElasticsearchTemporalIndexFieldTypeContext.java
```diff
@@ -36,6 +38,14 @@
 		return toIndexFieldType( mapping, formatter );
 	}
 
+	@Override
+	protected JsonPrimitive indexNullAsJsonPrimitive() {
+		ElasticsearchDefaultFieldFormatProvider defaultFieldFormatProvider = getBuildContext().getDefaultFieldFormatProvider();
+		DateTimeFormatter formatter = defaultFieldFormatProvider.getDefaultDateTimeFormatter( getFieldType() );
+
+		return new JsonPrimitive( formatter.format( indexNullAs ) );
```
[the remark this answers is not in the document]
fax4ever: You're definitely right. Does not make sense redo what codec does.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/mapping/IndexNullAsMappingIT.java
```diff
@@ -0,0 +1,143 @@
...
+import org.hibernate.search.integrationtest.backend.tck.testsupport.util.StandardFieldMapper;
+import org.hibernate.search.integrationtest.backend.tck.testsupport.util.rule.SearchSetupHelper;
+import org.hibernate.search.util.impl.integrationtest.common.stub.mapper.StubMappingIndexManager;
+import org.hibernate.search.util.impl.integrationtest.common.stub.mapper.StubMappingSearchScope;
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class IndexNullAsMappingIT {
```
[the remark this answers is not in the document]
fax4ever: In the very first commit you find it there, since the commit allows support for ES backend only. In a subsequent commit it has been moved to the TCK.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/mapping/definition/IndexNullAsValueIT.java
yrodiere: By "both cases" I meant the POJO mapper and the ORM mapper.

integrationtest/mapper/pojo/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/IndexNullAsValueIT.java
fax4ever: Good idea. Let's do that.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/types/AnalyzedStringFieldTypeDescriptor.java
```diff
@@ -76,4 +77,9 @@
 				"Aaron", "george", "Zach"
 		) );
 	}
+
+	@Override
+	public Optional<IndexNullAsExpectactions<String>> getIndexNullAsExpectations() {
+		return Optional.empty();
```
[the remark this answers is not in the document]
fax4ever: It is tested in a different method

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/types/GeoPointFieldTypeDescriptor.java
```diff
@@ -105,4 +106,9 @@ public boolean isFieldSortSupported() {
 				GeoPoint.of( 40, 80 )
 		) );
 	}
+
+	@Override
+	public Optional<IndexNullAsExpectactions<GeoPoint>> getIndexNullAsExpectations() {
+		return Optional.empty();
```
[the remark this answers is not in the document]
fax4ever: It is tested in a different method

fax4ever:
@yrodiere, thanks for the in depth review.  I think I've addressed most of the suggestions.

engine/src/main/java/org/hibernate/search/engine/cfg/spi/ParseUtils.java
```diff
@@ -0,0 +1,175 @@
...
+
+	private ParseUtils() {
+		// Private constructor, do not use
+	}
+
+	public static String parseString(String value) {
+		return value;
+	}
+
+	public static Instant parseInstant(String value) {
```
yrodiere: I'm not sure why you're using explicit formats everywhere while just calling `TheType.parse(String)` achieves exactly the same result, especially since the format is not likely to ever change.

Using `TheType.parse(String)` makes it easier to document we will only mention "use a string that can be parsed by `TheType.parse(String)`" and let users have a look at the javadoc of this method.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/building/impl/FieldModelContributorBridgeContextImpl.java
```diff
@@ -0,0 +1,27 @@
...
+import org.hibernate.search.engine.backend.types.dsl.StandardIndexFieldTypeContext;
+import org.hibernate.search.mapper.pojo.bridge.ValueBridge;
+import org.hibernate.search.mapper.pojo.mapping.building.spi.FieldModelContributorBridgeContext;
+
+public class FieldModelContributorBridgeContextImpl<F> implements FieldModelContributorBridgeContext {
+
+	private final ValueBridge<?, F> bridge;
+	private final StandardIndexFieldTypeContext<?, F> fieldTypeContext;
+
+	public FieldModelContributorBridgeContextImpl(ValueBridge<?, F> bridge, StandardIndexFieldTypeContext<?, F> fieldTypeContext) {
```
yrodiere: Should be `StandardIndexFieldTypeContext<?, ? super F> fieldTypeContext` to allow to call this constructor with a diamond operator in `PojoIndexModelBinderImpl`.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/building/impl/PojoIndexModelBinderImpl.java
```diff
@@ -311,7 +311,7 @@
 		);
 
 		// Then give the mapping a chance to override some of the model (add storage, ...)
-		contributor.contribute( fieldTypeContext );
+		contributor.contribute( fieldTypeContext, new FieldModelContributorBridgeContextImpl( bridge, fieldTypeContext ) );
```
yrodiere: This should use the diamond operator. You're instantiating a raw type, here.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/impl/PropertySortableFieldAnnotationProcessor.java
```diff
@@ -31,11 +31,21 @@
 			fieldContext.sortable( sortable );
 		}
 
+		String indexNullAs = getIndexNullAs( annotation );
+		// The empty String is the annotation default value, so it will interpreted as a null.
+		// If the user needs to set the empty String as indexNullAs value,
+		// he will have to use the programmatic APIs.
+		if ( indexNullAs != null && !indexNullAs.isEmpty() ) {
```
yrodiere: By default you could use a constant with a value that is unlikely to ever be picked by the user, for example `"__HibernateSearch_indexNullAs_default"`. If you do that, users will be able to use the empty string.

Note I'm not inventing anything: that's what we used to do in Search 5 with `org.hibernate.search.annotations.Field#DO_NOT_INDEX_NULL`.

engine/src/main/java/org/hibernate/search/engine/logging/impl/Log.java
```diff
@@ -275,4 +274,10 @@ SearchException unableToParseTemporal(@FormatWith(SimpleNameClassFormatter.class
 
 	@Message(id = ID_OFFSET_2 + 58, value = "Invalid %1$s value: expected either a Number or a String that can be parsed into a %1$s. %2$s")
 	SearchException invalidNumberPropertyValue(@FormatWith(SimpleNameClassFormatter.class) Class<? extends Number> type, String nestedErrorMessage, @Cause Exception cause);
+
+	@Message(id = ID_OFFSET_2 + 59, value = "Invalid '%1$s' value for type '%2$s'.")
+	SearchException invalidStringForType(String value, @FormatWith(SimpleNameClassFormatter.class) Class<?> type);
```
yrodiere: Maybe pass the cause in an additional `@Cause Exception cause` parameter?
IIRC you can pass null and it will just work (for your Character validation).

yrodiere: Also you could use the same method for both the enum and other types, it's not a big deal if we don't use the simple name :)

fax4ever: OK to add the cause. On the other hand I would like to keep a different message for Enum parsing error.

fax4ever:
Thank you @yrodiere. Is there other we can improve?

engine/src/main/java/org/hibernate/search/engine/cfg/spi/ParseUtils.java
yrodiere: I just noticed that you were displaying the formatter in the error message, so you probably wanted to actually use that formatter when parsing. That makes sense. Anyway, you already changed it, it will work just as well :)

yrodiere:
It looks good; I pushed a few commits with minor fixes and will merge as soon as CI goes green.

fax4ever:
Hi @yrodiere! Seen the fixed, they are good for me. Thank you ;)

yrodiere:
Merged. Thanks a lot!

fax4ever:
Thank you

PR #1949, opened 2019-04-11
HSEARCH-2941 Use long type only in MassIndexingMonitor

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-2941


yrodiere:
Looks good. Merged, thanks!

fax4ever:
Thanks!

PR #1952, opened 2019-04-12
Refactoring to support JDK 5 to 8 features

carstenartur:
Use diamond operator, use switch instead of if-then-else - even for strings, use lambda expressions and so on.
Signed-off-by: Carsten Hammer <carsten.hammer@t-online.de>


yrodiere:
Thanks for your PR, but I am going to close it for a few reasons:

1. It targets a maintenance branch of Hibernate Search, which we would like to avoid changing without good reasons (don't want to introduce bugs there).
2. It cannot be applied to the master branch, because a lot of the code is in the process of being rewritten.
3. The changes are quite massive and I'd have to take some time to review them. Since the changes will be lost in later versions (6.x), I'm not sure that's a wise use of my time.

I hope this didn't take too long for your. I assume you used some sort of automated tooling to perform the cleanup?

In any case, should you want to contribute again, we'd definitely like it; but let's open a ticket before, so we can discuss it :)

PR #1953, opened 2019-04-15
HSEARCH-3467 Add tests for all the attributes validated by the Elasticsearch schema validator

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3467

~~Still in preview, since several attributes have not been tested yet:~~ done

1. Property type attributes: ~~boots~~ removed!, ~~store, null_value, fielddata, doc_values, normalizer~~ done!.
2. ~~Normalizer definition attributes: char_filter, token_filter.~~ already exists
2. ~~Analyzer definition attributes: char_filter, token_filter, tokenizer.~~ already exists

### DIFF
--- a/integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/management/ElasticsearchSchemaAttributeValidationIT.java
+++ b/integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/management/ElasticsearchSchemaAttributeValidationIT.java
@@ -1,0 +1,783 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.integrationtest.backend.elasticsearch.management;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.hibernate.search.backend.elasticsearch.analysis.ElasticsearchAnalysisConfigurer;
+import org.hibernate.search.backend.elasticsearch.analysis.model.dsl.ElasticsearchAnalysisDefinitionContainerContext;
+import org.hibernate.search.backend.elasticsearch.cfg.ElasticsearchBackendSettings;
+import org.hibernate.search.backend.elasticsearch.cfg.ElasticsearchIndexLifecycleStrategyName;
+import org.hibernate.search.backend.elasticsearch.cfg.ElasticsearchIndexSettings;
+import org.hibernate.search.engine.backend.document.model.dsl.IndexSchemaElement;
+import org.hibernate.search.engine.backend.types.Projectable;
+import org.hibernate.search.integrationtest.backend.elasticsearch.testsupport.util.TestElasticsearchClient;
+import org.hibernate.search.integrationtest.backend.tck.testsupport.util.rule.SearchSetupHelper;
+import org.hibernate.search.util.common.SearchException;
+import org.hibernate.search.util.impl.integrationtest.common.FailureReportUtils;
+import org.hibernate.search.util.impl.test.SubTest;
+
+import org.junit.Assume;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class ElasticsearchSchemaAttributeValidationIT {
+
+	private static final String BACKEND_NAME = "myElasticsearchBackend";
+	private static final String INDEX_NAME = "IndexName";
+	private static final String SCHEMA_VALIDATION_CONTEXT = "schema validation";
+
+	@Rule
+	public SearchSetupHelper setupHelper = new SearchSetupHelper();
+
+	@Rule
+	public TestElasticsearchClient elasticSearchClient = new TestElasticsearchClient();
+
+	@Test
+	public void attribute_dynamic_missing() {
+		elasticSearchClient.index( INDEX_NAME ).deleteAndCreate();
+		elasticSearchClient.index( INDEX_NAME ).type().putMapping(
+				"{"
+					+ "'properties': {"
+							+ "'myField': {"
+									+ "'type': 'integer',"
+									+ "'index': true"
+							+ "}"
+					+ "}"
+				+ "}"
+		);
+
+		SubTest.expectException( () ->
+				validateSchemaConfig()
+						.withIndex( INDEX_NAME, ctx -> {
+									IndexSchemaElement root = ctx.getSchemaElement();
+									root.field( "myField", f -> f.asInteger() ).toReference();
+								}
+						)
+						.setup() )
+				.assertThrown()
+				.isInstanceOf( SearchException.class )
+				.hasMessageMatching( FailureReportUtils.buildFailureReportPattern()
+						.indexContext( INDEX_NAME )
+						.contextLiteral( SCHEMA_VALIDATION_CONTEXT )
+						.failure( "Invalid value for attribute 'dynamic'. Expected 'STRICT', actual is 'null'" )
+						.build() );
+	}
+
+	@Test
+	public void attribute_dynamic_invalid() {
+		elasticSearchClient.index( INDEX_NAME ).deleteAndCreate();
+		elasticSearchClient.index( INDEX_NAME ).type().putMapping(
+				"{"
+					+ "'dynamic': false,"
+					+ "'properties': {"
+							+ "'myField': {"
+									+ "'type': 'integer',"
+									+ "'index': true"
+							+ "}"
+					+ "}"
+				+ "}"
+		);
+
+		SubTest.expectException( () ->
+				validateSchemaConfig()
+						.withIndex( INDEX_NAME, ctx -> {
+									IndexSchemaElement root = ctx.getSchemaElement();
+									root.field( "myField", f -> f.asInteger() ).toReference();
+								}
+						)
+						.setup() )
+				.assertThrown()
+				.isInstanceOf( SearchException.class )
+				.hasMessageMatching( FailureReportUtils.buildFailureReportPattern()
+						.indexContext( INDEX_NAME )
+						.contextLiteral( SCHEMA_VALIDATION_CONTEXT )
+						.failure( "Invalid value for attribute 'dynamic'. Expected 'STRICT', actual is 'FALSE'" )
+						.build() );
+
+	}
+
+	@Test
+	public void attribute_properties_missing() {
+		elasticSearchClient.index( INDEX_NAME ).deleteAndCreate();
+		elasticSearchClient.index( INDEX_NAME ).type().putMapping(
+				"{"
+					+ "'dynamic': 'strict'"
+				+ "}"
+		);
+
+
+		SubTest.expectException( () ->
+				validateSchemaConfig()
+						.withIndex( INDEX_NAME, ctx -> {
+									IndexSchemaElement root = ctx.getSchemaElement();
+									root.field( "myField", f -> f.asInteger() ).toReference();
+								}
+						)
+						.setup() )
+				.assertThrown()
+				.isInstanceOf( SearchException.class )
+				.hasMessageMatching( FailureReportUtils.buildFailureReportPattern()
+						.indexContext( INDEX_NAME )
+						.contextLiteral( SCHEMA_VALIDATION_CONTEXT )
+						.indexFieldContext( "myField" )
+						.failure( "Missing property mapping" )
+						.build() );
+	}
+
+	@Test
+	public void attribute_properties_empty() {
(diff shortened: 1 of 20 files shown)


integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/management/ElasticsearchSchemaValidationIT.java
```diff
@@ -859,23 +853,6 @@ private void setupSimpleIndexWithLocalDateField(String indexName) {
...
-						// Don't contribute any analysis definitions, migration of those is tested in another test class
-						ElasticsearchBackendSettings.ANALYSIS_CONFIGURER,
-						new ElasticsearchAnalysisConfigurer() {
-							@Override
-							public void configure(ElasticsearchAnalysisDefinitionContainerContext context) {
-								// No-op
-							}
-						}
-				);
+		return setupHelper.withConfiguration( CONFIGURATION_ID );
```
yrodiere: I really don't see why this commit was necessary?
It looks like you're duplicating a properties file for no reason, whereas just using the default one and overriding what's relevant to the test used to work just fine.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/management/ElasticsearchSchemaAttributeValidationIT.java
```diff
@@ -0,0 +1,582 @@
...
+import org.hibernate.search.integrationtest.backend.tck.testsupport.util.rule.SearchSetupHelper;
+import org.hibernate.search.util.common.SearchException;
+import org.hibernate.search.util.impl.integrationtest.common.FailureReportUtils;
+import org.hibernate.search.util.impl.test.SubTest;
+
+import org.junit.Assume;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class ElasticsearchSchemaAttributeValidationIT {
```
yrodiere: Not sure why you're splitting it from the other test, since most of the validation is about attributes anyway... ?

fax4ever: I would like to keep apart tests on general validation aspects, such as general attribute leniency or aggregate errors and so on, from attribute-specific validation test.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/management/ElasticsearchSchemaValidationIT.java
fax4ever: I think that provide the configuration a `validate` schema strategy property type makes more readable the tests. But OK I'll revert this commit.

fax4ever:
@yrodiere thank you for the review. 
I ran this build: http://ci.hibernate.org/view/Search/job/hibernate-search/job/PR-1953/5/parameters/
to check the behavior for all supported ES versions.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/management/ElasticsearchSchemaAttributeValidationIT.java
```diff
@@ -0,0 +1,890 @@
...
+				.hasMessageMatching( FailureReportUtils.buildFailureReportPattern()
+						.indexContext( INDEX_NAME )
+						.contextLiteral( SCHEMA_VALIDATION_CONTEXT )
+						.failure( "Invalid value for attribute 'dynamic'. Expected 'STRICT', actual is 'FALSE'" )
+						.build() );
+
+	}
+
+	@Test
+	public void attribute_properties_missing() {
```
yrodiere: If we follow your line of reasoning regarding the moving of test method to a separate class, shouldn't this have remained in `ElasticsearchSchemaValidationIT`, next to `mapping_missing` and `property_fields_missing`?

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/management/ElasticsearchSchemaAttributeValidationIT.java
```diff
@@ -0,0 +1,890 @@
...
+				.hasMessageMatching( FailureReportUtils.buildFailureReportPattern()
+						.indexContext( INDEX_NAME )
+						.contextLiteral( SCHEMA_VALIDATION_CONTEXT )
+						.indexFieldContext( "myField" )
+						.failure( "Missing property mapping" )
+						.build() );
+	}
+
+	@Test
+	public void attribute_properties_empty() {
```
yrodiere: Same question regarding the moving of this method out of `ElasticsearchSchemaValidationIT`.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/management/ElasticsearchSchemaAttributeValidationIT.java
```diff
@@ -0,0 +1,890 @@
...
+				.hasMessageMatching( FailureReportUtils.buildFailureReportPattern()
+						.indexContext( INDEX_NAME )
+						.contextLiteral( SCHEMA_VALIDATION_CONTEXT )
+						.indexFieldContext( "myField" )
+						.failure( "Missing property mapping" )
+						.build() );
+	}
+
+	@Test
+	public void attribute_field_notPresent() {
```
yrodiere: Same question regarding the moving of this method out of `ElasticsearchSchemaValidationIT`.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/management/ElasticsearchSchemaAttributeValidationIT.java
```diff
@@ -0,0 +1,890 @@
...
+	private static final String SCHEMA_VALIDATION_CONTEXT = "schema validation";
+
+	@Rule
+	public SearchSetupHelper setupHelper = new SearchSetupHelper();
+
+	@Rule
+	public TestElasticsearchClient elasticSearchClient = new TestElasticsearchClient();
+
+	@Test
+	public void valid() {
```
yrodiere: I think this is already tested in `org.hibernate.search.integrationtest.backend.elasticsearch.management.ElasticsearchSchemaValidationIT#success_simple`.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/management/ElasticsearchSchemaAttributeValidationIT.java
```diff
@@ -0,0 +1,890 @@
...
+				.withIndex( INDEX_NAME, ctx -> {
+							IndexSchemaElement root = ctx.getSchemaElement();
+							root.field( "myField", f -> f.asString().analyzer( "default" ) ).toReference();
+						}
+				)
+				.setup();
+	}
+
+	@Test
+	public void attribute_docValues_ignored() {
```
yrodiere: Why "ignored"? It should not be ignored, there are cases where doc values are required and them being false should trigger an error.
For examples this is the case if you create a sortable field.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/management/ElasticsearchSchemaAttributeValidationIT.java
```diff
@@ -0,0 +1,890 @@
...
+				.hasMessageMatching( FailureReportUtils.buildFailureReportPattern()
+						.indexContext( INDEX_NAME )
+						.contextLiteral( SCHEMA_VALIDATION_CONTEXT )
+						.indexFieldContext( "myField" )
+						.failure( "Invalid value for attribute 'null_value'. Expected '739', actual is '777'" )
+						.build() );
+	}
+
+	@Test
+	public void attribute_fielddata_ignored() {
```
yrodiere: If we don't use fielddata, maybe it should just be removed from the model entirely, like you did for boost.
It only made sense when we allowed sorts on analyzed string fields, but we don't allow that anymore.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/management/ElasticsearchSchemaAttributeValidationIT.java
```diff
@@ -0,0 +1,890 @@
...
+				.withIndex( INDEX_NAME, ctx -> {
+							IndexSchemaElement root = ctx.getSchemaElement();
+							root.field( "myField", f -> f.asInteger().indexNullAs( 739 ) ).toReference();
+						}
+				)
+				.setup();
+	}
+
+	@Test
+	public void attribute_nullValue_missed() {
```
yrodiere: `missed` => `missing`?

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/management/ElasticsearchSchemaAttributeValidationIT.java
fax4ever: You're right. That's the reasoning.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/management/ElasticsearchSchemaAttributeValidationIT.java
fax4ever: Right. I moved it.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/management/ElasticsearchSchemaAttributeValidationIT.java
fax4ever: Same as above.

fax4ever:
Thanks @yrodiere. I think I've addressed the points here.
I started an new full built: http://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-1953/7.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/management/ElasticsearchSchemaAttributeValidationIT.java
yrodiere: So `properties`, which appears at the top level unlike other attributes, is an attribute nonetheless, but `fields`, which appears just below a property like other attributes, is not (`org.hibernate.search.integrationtest.backend.elasticsearch.management.ElasticsearchSchemaValidationIT#property_fields_missing`). I don't understand, but ok, I'll merge it anyway.

fax4ever: The name of the field itself is not fixed, it depends case by case. In this sense it would be not a fixed-name attribute.

yrodiere: They are structure in the exact same way, except fields can only be found at the same level as other attributes, while properties can also appear at the top level:

```
"properties": {
  "propertyName": {
    "type": "object",
    "properties": {
      "propertyName1": {
        "type": "integer"
      },
      "propertyName2": {
        "type": "keyword",
        "normalizer": "foo"
      },
  }
}
```

```
"properties": {
  "propertyName": {
    "type": "keyword",
    "fields": {
      "fieldName1": {
        "type": "integer"
      },
      "fieldName2": {
        "type": "keyword",
        "normalizer": "foo"
      }
  }
}
```

yrodiere:
Merged, thanks!

fax4ever:
Thanks @yrodiere. I'm reviewing yours.

PR #1958, opened 2019-04-16
HSEARCH-3563 Simpler dialect configuration

yrodiere:
[HSEARCH-3563](https://hibernate.atlassian.net//browse/HSEARCH-3563)

Based on #1956 , which should be merged first.


yrodiere:
Merged, thanks!

PR #1963, opened 2019-04-26
HSEARCH-3298 Mark the accessor-based entity access API and explicit reindexing declaration API in bridges as experimental

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3298

~Based on #1962 , which should be merged first.~ => Done

### DIFF
--- a/mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/bridge/binding/IdentifierBridgeBindingContext.java
+++ b/mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/bridge/binding/IdentifierBridgeBindingContext.java
@@ -16,6 +16,8 @@ public interface IdentifierBridgeBindingContext<T> {
 
 	/**
 	 * @return An entry point allowing to inspect the type of values that will be passed to this bridge.
+	 * @hsearch.experimental This feature is under active development.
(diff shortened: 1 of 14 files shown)


mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/bridge/binding/IdentifierBridgeBindingContext.java
```diff
@@ -16,6 +16,8 @@
 
 	/**
 	 * @return An entry point allowing to inspect the type of values that will be passed to this bridge.
+	 * @hsearch.experimental This feature is under active development.
+	 *    You should be prepared for incompatible changes in future releases.
```
fax4ever: Maybe using a 'you' in a Javadoc could be too much informal.
What do you think of `The API is is susceptible to change` ?

yrodiere: That's a message dating back to Search 5, but ok.
I changed it to: `Usual compatibility policies do not apply: incompatible changes may be introduced in any future release.`

yrodiere:
Merged, thanks!

PR #1965, opened 2019-04-30
HSEARCH-3536 Use Analyzer.normalize instead of our own code for normalization

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3536

@yrodiere as you found out Analyzer.normalize method works only if we use the Lucene's 'CustomAnalyzer.builder'.
I think that our analyzers don't. So they have to override them to work as 'CustomAnalyzer' does.

In a second commit I renamed the utility classes, we can drop it.

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/analysis/impl/TokenizerChain.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/analysis/impl/TokenizerChain.java
@@ -55,6 +55,25 @@ final class TokenizerChain extends Analyzer {
 	}
 
 	@Override
+	protected Reader initReaderForNormalization(String fieldName, Reader reader) {
+		// same as Lucene8's CustomAnalyzer
+		for ( CharFilterFactory charFilter : charFilters ) {
+			reader = charFilter.normalize( reader );
+		}
+		return reader;
+	}
+
+	@Override
+	protected TokenStream normalize(String fieldName, TokenStream in) {
+		// same as Lucene8's CustomAnalyzer
+		TokenStream result = in;
+		for ( TokenFilterFactory filter : filters ) {
+			result = filter.normalize( result );
+		}
+		return result;
+	}
+
+	@Override
 	public String toString() {
 		final StringBuilder sb = new StringBuilder( "TokenizerChain(" );
 		for ( CharFilterFactory filter : charFilters ) {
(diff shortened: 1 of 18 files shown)


yrodiere:
Jenkins, retest this please

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/analysis/impl/TokenizerChain.java
```diff
@@ -54,6 +54,15 @@ protected TokenStreamComponents createComponents(String fieldName) {
 		return new TokenStreamComponents( tk, ts );
 	}
 
+	@Override
+	protected TokenStream normalize(String fieldName, TokenStream in) {
```
yrodiere: I was surprised to see only token filters are mentioned here. It seems `CustomAnalyzer` also overrides another method to make sure char filters are applied: `org.apache.lucene.analysis.custom.CustomAnalyzer#initReaderForNormalization`. I think you should also copy this method. Maybe just below `initReader`?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneStringFieldCodec.java
```diff
@@ -110,7 +109,7 @@ public String normalize(String absoluteFieldPath, String value) {
 		}
 
 		return analyzerOrNormalizer != null
-				? AnalyzerUtils.normalize( analyzerOrNormalizer, absoluteFieldPath, value )
+				? analyzerOrNormalizer.normalize( absoluteFieldPath, value ).utf8ToString()
```
yrodiere: I think most of the callers of `normalize()` convert the String you return to a `BytesRef`. So maybe it would be better to drop the extra `utf8ToString()` call and make this method return a `BytesRef` directly?

fax4ever:
@yrodiere thanks for the review. I've applied the change here.
Do you think we need a test covering the override of `initReaderForNormalization`?

yrodiere:
@fax4ever It's a good idea if it's not too much work, yes. You would need to add something to `org.hibernate.search.integrationtest.backend.tck.analysis.AnalysisCustomIT`: a new normalizer that makes use of a char filter. You'll need a new entry in `org.hibernate.search.integrationtest.backend.tck.analysis.AnalysisCustomIT.AnalysisDefinitions`, a new test method, and of course new definitions in the corresponding analysis configurers (`org.hibernate.search.integrationtest.backend.elasticsearch.testsupport.configuration.AnalysisCustomITAnalysisConfigurer` and a similar one for Lucene).

fax4ever:
@yrodiere I did in the last commit

yrodiere:
Merged, thanks!

fax4ever:
Thank you

PR #1967, opened 2019-05-02
HSEARCH-1857 Throw exceptions on usage of a FullTextSession whose underlying Session is closed

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-1857

### DIFF
--- a/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/logging/impl/Log.java
+++ b/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/logging/impl/Log.java
@@ -183,4 +183,7 @@ public interface Log extends BasicLogger {
 	@Message(id = ID_OFFSET_2 + 15,
 			value = "Invalid property handle factory name: '%1$s'. Valid names are: %2$s.")
 	SearchException invalidPropertyHandleFactoryName(String invalidRepresentation, List<String> validRepresentations);
+
+	@Message(id = ID_OFFSET_2 + 16, value = "Error trying to access Hibernate ORM session." )
+	SearchException hibernateSessionAccessError(@Cause IllegalStateException cause);
 }
--- a/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/logging/impl/Log.java
+++ b/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/logging/impl/Log.java
@@ -186,4 +186,7 @@ public interface Log extends BasicLogger {
 
 	@Message(id = ID_OFFSET_2 + 16, value = "Error trying to access Hibernate ORM session." )
 	SearchException hibernateSessionAccessError(@Cause IllegalStateException cause);
+
+	@Message(id = ID_OFFSET_2 + 17, value = "Underlying Hibernate ORM Session seems to be closed." )
+	SearchException hibernateSessionIsClosed(@Cause IllegalStateException cause);
 }
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/hibernateormapis/ToHibernateOrmIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/hibernateormapis/ToHibernateOrmIT.java
@@ -261,6 +261,22 @@ public class ToHibernateOrmIT {
 				.hasMessage( "HSEARCH800016: Error trying to access Hibernate ORM session." );
 	}
 
+	@Test
+	@TestForIssue( jiraKey = "HSEARCH-1857" )
+	public void reuseSearchQueryAfterOrmSessionIsClosed_noMatching() {
+		Session session = sessionFactory.openSession();
+		SearchSession searchSession = Search.getSearchSession( session );
+		SearchQuery<IndexedEntity> query = createSimpleQuery( searchSession );
+		session.close();
+
+		SubTest.expectException( () -> {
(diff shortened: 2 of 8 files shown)


yrodiere:
Ok, so I don't think that's  really what this ticket was about.

Quoting:

> The tricky part is that it would sometimes work fine; the discriminating factor turned out to be that some of the operations exposed on the FullTextEntityManager don't actually make use of the underlying EntityManager.
> For example, when running a FullTextQuery you wouldn't have an exception about the EntityManager being closed if there are no matches for the query.

So we know that, as shown in your tests, some operations will already trigger an exception when you try to use a `SearchSession` whose underlying `Session` is closed.

What got Sanne concerned (rightfully so, IMO) is that some operations *will work*.

For example, if you do that:

```
Session session = sessionFactory.openSession();
SearchSession searchSession = Search.getSearchSession( session );
session.close();

searchSession.search( Object.class ).asProjection( f -> f.reference() ).predicate( f -> f.matchAll() ).toQuery().fetchHits();
```

Well, it might just work. This code doesn't rely on the underlying Hibernate ORM session, so it might work. Yet, there's something wrong with it, and we would like this to fail.

So, to sum up, what is needed here is to add checks, in strategic places in our code, that the session is open. This could simply be calls to `org.hibernate.engine.spi.SharedSessionContractImplementor#checkOpen()`, but we need to make those calls.

Now, where to make those calls?
First, I think we should make them at the start of every relevant method in `HibernateOrmSearchSession` that is not just a getter or setter.
Then, I think you could do the same with `HibernateOrmSearchQuery`: all the `fetch` methods should probably check the session is still open before doing anything.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/logging/impl/Log.java
```diff
@@ -183,4 +183,10 @@ SearchException cannotFindReadableProperty(@FormatWith(PojoTypeModelFormatter.cl
 	@Message(id = ID_OFFSET_2 + 15,
 			value = "Invalid property handle factory name: '%1$s'. Valid names are: %2$s.")
 	SearchException invalidPropertyHandleFactoryName(String invalidRepresentation, List<String> validRepresentations);
+
+	@Message(id = ID_OFFSET_2 + 16, value = "Error trying accessing Hibernate ORM session." )
+	SearchException hibernateSessionAccessError(@Cause IllegalStateException cause);
+
+	@Message(id = ID_OFFSET_2 + 17, value = "Error trying accessing entity manager." )
+	SearchException entityManagerAccessError(@Cause IllegalStateException cause);
```
yrodiere: No need to have two separate errors for Session and EntityManager; if you do what I suggested in my comment, this will complicate things a lot for you, for no good reason. Just have a single message that mentions the session?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/logging/impl/Log.java
```diff
@@ -186,4 +186,7 @@ SearchException cannotFindReadableProperty(@FormatWith(PojoTypeModelFormatter.cl
 
 	@Message(id = ID_OFFSET_2 + 16, value = "Error trying accessing Hibernate ORM session." )
 	SearchException hibernateSessionAccessError(@Cause IllegalStateException cause);
+
+	@Message(id = ID_OFFSET_2 + 17, value = "Underlining Hibernate ORM Session is closed." )
```
yrodiere: "Underlining" => "Underlying"

fax4ever:
@yrodiere thanks for the insights here.
You're right because it seems that we allowed to have queries on a existing Search session, having a closed ORM session.

Actually to reproduce it had to use the Search session just after its creation, since it is created lazily.

In the last commit I handled the case for mass indexing api too.

The commits:

1. HSEARCH-1857 Raise SearchException on lazy Search session creation Fabio Massimo Ercoli 5/2/19 4:37 PM
2. HSEARCH-1857 Raise SearchException trying unwrap closed session Fabio Massimo Ercoli 5/2/19 3:11 PM

are used to wrap the already raised exception by ORM with a `SearchException`. Maybe they are out of scope with the original intent of the issue.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/logging/impl/Log.java
```diff
@@ -183,4 +183,7 @@ SearchException cannotFindReadableProperty(@FormatWith(PojoTypeModelFormatter.cl
 	@Message(id = ID_OFFSET_2 + 15,
 			value = "Invalid property handle factory name: '%1$s'. Valid names are: %2$s.")
 	SearchException invalidPropertyHandleFactoryName(String invalidRepresentation, List<String> validRepresentations);
+
+	@Message(id = ID_OFFSET_2 + 16, value = "Error trying accessing Hibernate ORM session." )
```
yrodiere: I really think it's "Error trying to access" rather than "Error trying accessing".

"trying accessing" would mean "accessing" is something we wanted to try for the experience it gives us, because it's fun, something we've never done, etc. Which doesn't make much sense here.
 
See https://learnenglish.britishcouncil.org/intermediate-grammar/verbs-followed-ing-or-infinitive-2

fax4ever:
Thanks for the review. Anything else I can improve?

yrodiere:
Merged. Thanks!

fax4ever:
Happy about it. Thanks

PR #1968, opened 2019-05-02
HSEARCH-3571 Extra type name parameters trigger warnings in ES6.7+

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3571

I checked that the warnings are gone in the logs of this full build: http://ci.hibernate.org/view/Search/job/hibernate-search-yoann/job/HSEARCH-3571/6/

The only remaining warnings in Elasticsearch logs are the one we can't get rid of on ES 6.7 (see last commit), and some unrelated warnings about "minimumShouldMatch", which I think are false positives from Elasticsearch.


fax4ever:
Look good. Avoid a wrong warning is a good thing. Merged thanks.

PR #1970, opened 2019-05-08
HSEARCH-3487 Support java.math.BigDecimal and java.math.BigInteger as field types

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3487

This is a preview. Subsequent commits will arrive,
So far I've just pushed the commits to handle the decimal precision attribute taken both from both Hibernate Search annotation and Hibernate ORM schema metadata.

I've addressed the merging of the different contribution (mapper ORM + mapper Pojo) in the best way I could. Any suggestion here is more than welcome.


mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/AnnotationDefaultValues.java
```diff
@@ -14,6 +14,11 @@
 	 */
 	public static final String DO_NOT_INDEX_NULL = "__HibernateSearch_indexNullAs_default";
 
+	/**
+	 * Same default as scale attribute of JPA annotation Column when the precision attribute is not defined
```
yrodiere: I think this should rather be an extreme value that realistically can't be used, so that we can detect when the user didn't specify anything.

For example, `Integer.MAX_VALUE`. Elasticsearch does not support scaling by a decimal scale higher than approximately ~300 (since it stores the scaling factor in a double), so this would be rather safe.

That way, you will be able to use the ORM metadata as a default, *and* users would still be able to force the decimal scale to 2 here even when they set it to 3 or 4 in the `@Column` annotation.

Other advantage: you would be able to detect when the scale wasn't configured at all, neither from the ORM metadata (because not using the ORM mapper) nor from the mapping (because `decimalScale` was not used). And you could throw an exception to warn the user of something dodgy. Because, let's be real, `2` might work for some use cases, but if you're using a BigDecimal you probably have specific needs and should configure the scale explicitly.

Note we do not need to be consistent with JPA, since ORM will always contribute something for every property and thus will never use defaults. So mentioning JPA in the javadoc is not necessary.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/dsl/impl/LuceneIndexFieldTypeFactoryContextImpl.java
```diff
@@ -209,6 +210,12 @@ else if ( GeoPoint.class.equals( inputType ) ) {
 		return new LuceneGeoPointIndexFieldTypeContext( this );
 	}
 
+	@Override
+	public ScaledNumberIndexFieldTypeContext<?> asBigDecimal() {
```
yrodiere: Don't forget to also add the relevant `if` block in `as(Class<?>)`.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/ElasticsearchIndexFieldTypeFactoryContextImpl.java
```diff
@@ -211,6 +212,12 @@ else if ( GeoPoint.class.equals( inputType ) ) {
 		return new ElasticsearchGeoPointIndexFieldTypeContext( this );
 	}
 
+	@Override
+	public ScaledNumberIndexFieldTypeContext<?> asBigDecimal() {
```
yrodiere: Don't forget to also add the relevant `if` block in `as(Class<?>)`.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/model/JpaColumnScaleMappingIT.java
```diff
@@ -0,0 +1,72 @@
...
+	@Indexed(index = INDEX_NAME)
+	public static final class IndexedEntity {
+
+		@Id
+		@DocumentId
+		private Integer id;
+
+		@ScaledNumberField
+		@Column(precision = 14, scale = 7)
+		private BigDecimal scaled;
```
yrodiere: Shouldn't it work with `@GenericField`, too? I would expect it to.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/building/spi/PojoMetadataContributor.java
```diff
@@ -15,4 +15,19 @@
...
+	 * provided with {@link #contributeMapping(PojoMappingCollector)}.
+	 * <p>
+	 * Used only by contributions that must be provided before other contributions on the same collector.
+	 * For instance when we need to provide <b>a mapping default value</b> that could be overridden by an another contributor.
+	 * <p>
+	 * Use {@link #contributeMapping(PojoMappingCollector)} whenever a contribution could be applied at any time.
+	 *
+	 * @param collector the target collector for the contribution.
+	 */
+	default void contributePreMapping(MAC collector) {
```
yrodiere: I don't understand why you added this method. Contributing metadata before the mapping is exactly the point of `contributeAdditionalMetadata`. Let's talk about it, I think there's a misunderstanding.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/building/spi/PojoMappingCollectorValueNode.java
```diff
@@ -19,4 +19,5 @@ void valueBridge(BridgeBuilder<? extends ValueBridge<?, ?>> builder,
 
 	void indexedEmbedded(String relativePrefix, ObjectFieldStorage storage, Integer maxDepth, Set<String> includePaths);
 
+	void defaultScale(int scale);
```
yrodiere: This doesn't have anything to do with the mapping, since it doesn't create anything in the index schema. This is purely Pojo metadata and should be moved to `PojoAdditionalMetadataCollectorValueNode`.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/building/impl/FieldModelContributorIndirectContextImpl.java
```diff
@@ -0,0 +1,59 @@
...
+	public void decimalScale(int decimalScale) {
+		if ( fieldTypeContext instanceof ScaledNumberIndexFieldTypeContext ) {
+			( (ScaledNumberIndexFieldTypeContext) fieldTypeContext ).decimalScale( decimalScale );
+			useDefaultDecimalScale = false;
+		}
+	}
+
+	@Override
+	public void defaultDecimalScale(int decimalScale) {
+		if ( useDefaultDecimalScale && fieldTypeContext instanceof ScaledNumberIndexFieldTypeContext ) {
```
yrodiere: It's a nice a solution, but I wonder if it's better than the obvious one of applying contributions in the right order so that the last one wins.

We have three sources of contributions:

* the field bridge itself, which may apply a "hard-coded" decimal scale if it's a custom bridge.
* the `@ScaledNumberField` annotation (or programmatic mapping equivalent)
* the `@Column` annotation

I think what we want is this precedence:

1. `@ScaledNumberField` (takes precedence over everything)
2. Field bridge
3. `@Column` (only applied as a last-resort default)

What you implemented is this:

1. `@ScaledNumberField` (takes precedence over everything)
2. `@Column`
3. Field bridge (only applied as a last-resort default)

The thing is, if we simply applied the default before applying the `@ScaledNumberField`, we would get exactly the same precedence, with a much simpler implementation:

* Apply bridge
* Call `.decimalScale( <value from @Column> )` if a value was provided
* Call `.decimalScale( <value from @ScaledNumberField> )` if a value was provided

Now, if we wanted the Field bridge to take precedence over the default, then we would need to communicate the defaults to the backend somehow. This is doable: we would basically only need to pass an object with a `getDefaultDecimalScale()` method to `org.hibernate.search.engine.mapper.mapping.building.spi.IndexBindingContext#getTypeFactory`. But it would add a bit of complexity. Do we want to go there? Let's discuss it, maybe?

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/model/JpaColumnScaleMappingIT.java
fax4ever: Yep. There is a test for it.

yrodiere: As discussed, there is not. To do.

fax4ever: Added and you were right. It already works with `@GenericField`, since the subtype of the context depends on field type and not on the annotation we use.

fax4ever:
@yrodiere thanks. I think I've addressed the comments here.
I will close this preview as soon as you confirm the changes.

engine/src/main/java/org/hibernate/search/engine/backend/types/dsl/ScaledNumberIndexFieldTypeContext.java
```diff
@@ -16,4 +16,6 @@
 
 	S decimalScale(int decimalScale);
 
+	S defaultDecimalScale(int decimalScale);
```
yrodiere: I think I mentioned in one of my previous commits that this default should be contributed by passing an object to `org.hibernate.search.engine.mapper.mapping.building.spi.IndexBindingContext#getTypeFactory`. There was a reason, actually: the type you're modifying here is API. If you add a `defaultDecimalScale` method here, users will be able to call it. I don't know about you, but I don't want to have to care about what happens if they call it twice, or if they call it and the ORM mapper also calls it, etc. So I'd rather not expose this feature to users.

If, instead, you change `org.hibernate.search.engine.mapper.mapping.building.spi.IndexBindingContext#getTypeFactory` to pass a `IndexFieldTypeDefaultsProvider` object, I think the result will be easier to test thoroughly.

fax4ever: it does make sense!

fax4ever:
@yrodiere thanks. Anything else I can improve on this part of the issue? Otherwise I'll close it.

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/stub/backend/document/model/impl/StubIndexSchemaRootNodeBuilder.java
```diff
@@ -34,7 +35,9 @@ private StubIndexSchemaRootNodeBuilder(StubBackendBehavior backendBehavior, Stri
 	}
 
 	@Override
-	public IndexFieldTypeFactoryContext getTypeFactory() {
+	public IndexFieldTypeFactoryContext getTypeFactory(IndexFieldTypeDefaultsProvider defaultsProvider) {
+		// set defaultsProvider instance for the current request
+		typeFactoryContext.setDefaultsProvider( defaultsProvider );
```
yrodiere: Probably not a big deal in this case, but for the real backends it would be great if you could avoid setters and create a new type factory contexts for each call to `getTypeFactory()` instead, passing the defaults provider to the constructor.

fax4ever: That's would be good and maybe we should change also the name of the method consistently from `getTypeFactory()` in `createTypeFactory`. Shouldn't we?

yrodiere: Sure.

PR #1974, opened 2019-05-10
HSEARCH-3324 Keep track of which field may be multi-valued and force bridges to declare fields as such

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3324

Note this does not bring any feature in itself, it just makes sure the index schema is properly defined by users so that we will later be able to take advantage of the information "this field is multi-valued".

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/document/impl/AbstractLuceneNonFlattenedDocumentBuilder.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/document/impl/AbstractLuceneNonFlattenedDocumentBuilder.java
@@ -6,13 +6,16 @@
  */
 package org.hibernate.search.backend.lucene.document.impl;
 
-import java.util.HashSet;
+import java.lang.invoke.MethodHandles;
+import java.util.HashMap;
 import java.util.List;
-import java.util.Set;
+import java.util.Map;
 
 import org.hibernate.search.backend.lucene.document.model.impl.LuceneIndexSchemaObjectNode;
+import org.hibernate.search.backend.lucene.logging.impl.Log;
 import org.hibernate.search.backend.lucene.multitenancy.impl.MultiTenancyStrategy;
 import org.hibernate.search.backend.lucene.util.impl.LuceneFields;
+import org.hibernate.search.util.common.logging.impl.LoggerFactory;
 
 import org.apache.lucene.document.Document;
 import org.apache.lucene.document.Field;
@@ -25,8 +28,10 @@ import org.apache.lucene.index.IndexableField;
 abstract class AbstractLuceneNonFlattenedDocumentBuilder extends AbstractLuceneDocumentBuilder
 		implements LuceneDocumentBuilder {
 
+	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
+
 	final Document document = new Document();
-	private final Set<String> fieldNames = new HashSet<>();
+	private final Map<String, EncounteredFieldStatus> fieldStatus = new HashMap<>();
 
 	AbstractLuceneNonFlattenedDocumentBuilder(LuceneIndexSchemaObjectNode schemaNode) {
 		super( schemaNode );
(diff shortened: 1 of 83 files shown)


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/document/impl/AbstractLuceneNonFlattenedDocumentBuilder.java
```diff
@@ -0,0 +1,79 @@
...
+
+	@Override
+	public void addFieldName(String absoluteFieldPath) {
+		// If the status was already ENCOUNTERED, just replace it.
+		fieldStatus.put( absoluteFieldPath, EncounteredFieldStatus.ENCOUNTERED_AND_NAME_INDEXED );
+	}
+
+	@Override
+	void checkNoValueYetForSingleValued(String absoluteFieldPath) {
+		EncounteredFieldStatus previousValue = fieldStatus.putIfAbsent( absoluteFieldPath, EncounteredFieldStatus.ENCOUNTERED );
```
fax4ever: Is it possible that we have multiple concurrent requests here?

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/document/DocumentElementMultiValuedIT.java
```diff
@@ -0,0 +1,351 @@
...
+import org.hibernate.search.util.impl.integrationtest.common.stub.mapper.StubMappingIndexManager;
+import org.hibernate.search.util.impl.test.SubTest;
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+/**
+ * Test the behavior of implementations of {@link DocumentElement}
+ * when it comes to .
```
fax4ever: when it comes to multi valued fields?

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/document/DocumentElementMultiValuedIT.java
```diff
@@ -0,0 +1,351 @@
...
+			DocumentElement level1 = document.addObject( indexMapping.multiValuedNestedObject.self );
+			level1.addValue( indexMapping.multiValuedNestedObject.multiValuedString, "1" );
+			level1.addValue( indexMapping.multiValuedNestedObject.multiValuedString, "2" );
+			level1.addValue( indexMapping.multiValuedNestedObject.multiValuedString, "3" );
+		} );
+	}
+
+	@Test
+	public void inheritedFromFlattenedParent() {
+		// -> multi-valued flattened => children are multi (whether explicit or not)
```
fax4ever: is it supposed to be implemented?

engine/src/main/java/org/hibernate/search/engine/backend/types/dsl/IndexFieldTypeCardinalityContext.java
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.engine.backend.types.dsl;
+
+/**
+ * @param <S> The concrete type of this context.
+ */
+public interface IndexFieldTypeCardinalityContext<S extends IndexFieldTypeCardinalityContext<? extends S>> {
```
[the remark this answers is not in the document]
yrodiere: Good catch; I'll remove it.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/util/StandardFieldMapper.java
```diff
@@ -55,21 +56,40 @@ public M map(IndexSchemaElement parent, String name) {
...
-				}
-		)
-				.toReference();
+		return map( parent, name, false, additionalConfigurations );
+	}
+
+	@SafeVarargs
+	public final M mapMultiValued(IndexSchemaElement parent, String name,
+			Consumer<? super StandardIndexFieldTypeContext<?, F>>... additionalConfigurations) {
+		return map( parent, name, true, additionalConfigurations );
```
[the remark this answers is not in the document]
yrodiere: It's just test utils. I'll leave it for now, it'll be there if we need it.

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/AbstractIndexBindingContext.java
```diff
@@ -66,38 +67,53 @@ public IndexSchemaElement getSchemaElement(IndexSchemaContributionListener liste
...
-	public Optional<IndexedEmbeddedBindingContext> addIndexedEmbeddedIfIncluded(MappableTypeModel parentTypeModel,
+	public Optional<IndexedEmbeddedBindingContext> addIndexedEmbeddedIfIncluded(
+			MappableTypeModel parentTypeModel, boolean multiValued,
 			String relativePrefix, ObjectFieldStorage storage, Integer maxDepth, Set<String> includePaths) {
 		return nestingContext.addIndexedEmbeddedIfIncluded(
 				parentTypeModel, relativePrefix, maxDepth, includePaths,
-				new NestedContextBuilderImpl( indexSchemaRootNodeBuilder, indexSchemaObjectNodeBuilder, storage )
+				new NestedContextBuilderImpl(
+						indexSchemaRootNodeBuilder, indexSchemaObjectNodeBuilder, storage,
+						isMultiValuedFromParentNotPropagated() || multiValued
```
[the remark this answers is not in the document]
yrodiere: Ok, so the code is right, but the name is very, very bad indeed. When`isMultiValuedFromParentNotPropagated` is `true`, it actually means the parent is multi-valued, but that information could not be propagated to the index and thus should be propagated in children.

Think of a `@IndexedEmbedded` property of type `List<MyEmbedded>` which doesn't add any object field: its fields will automatically become multi-valued, because the property has multiple `MyEmbedded` values and each `MyEmbedded` value has to be translated into field values. So even if `MyEmbedded` only declares single-valued fields, in the end these fields will become multi-valued in the `@IndexedEmbedded`.

I'll try to find a better name for `isMultiValuedFromParentNotPropagated`.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/document/impl/AbstractLuceneNonFlattenedDocumentBuilder.java
yrodiere: Nope, the document builder is single-threaded. I just need `putIfAbsent` because of its semantics, not because I need atomicity.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/document/DocumentElementMultiValuedIT.java
yrodiere: Right, thanks

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/document/DocumentElementMultiValuedIT.java
yrodiere: Sorry, I was really messy this time... This was just something I wrote initially to sum up what needs to be tested. It's obsolete, I removed it.

yrodiere:
Merged, thanks!

fax4ever:
Thank you!

PR #1975, opened 2019-05-13
HSEARCH-3353 Restore support for search hit explanation

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3353

~Based on #1971 , which should be merged first.~ => Done

See the tests added to `ElasticsearchExtensionIT` and `LuceneExtensionIT` for how the feature will be used. It's a bit different from what we used to provide through `org.hibernate.search.jpa.FullTextQuery#explain`, but I think it is more convenient.


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/ElasticsearchSearchQuery.java
```diff
@@ -9,4 +9,30 @@
...
+	 *
+	 * @param id The id of the document to test.
+	 * @return A string representing a JSON object describing the score computation for the hit.
+	 * @throws org.hibernate.search.util.common.SearchException If the query targets multiple indexes,
+	 * or if the explain request fails.
+	 */
+	String explain(String id);
+
+	/**
+	 * Explain score computation of this query the document with the given id in the given index.
```
fax4ever: maybe to the document?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/LuceneSearchQuery.java
```diff
@@ -8,5 +8,32 @@
...
+	 *
+	 * @param id The id of the document to test.
+	 * @return An {@link org.apache.lucene.search.Explanation} describing the score computation for the hit.
+	 * @throws org.hibernate.search.util.common.SearchException If the query targets multiple indexes,
+	 * or if the explain request fails.
+	 */
+	Explanation explain(String id);
+
+	/**
+	 * Explain score computation of this query the document with the given id in the given index.
```
fax4ever: I wouldn't know... query to the document maybe?

yrodiere:
Thanks; I fixed the javadoc and merged.

fax4ever:
Thank you!

PR #1977, opened 2019-05-13
HSEARCH-3562 Reporting errors about duplicate fields fails with an AssertionFailure when field is declared by a class bridge or class-level @Spatial

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3562

To be backported to several other branches, see the ticket.


yrodiere:
Merged and backported. Thanks!

PR #1978, opened 2019-05-13
HSEARCH-3570 Switch to XML format for JaCoCo reports

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3570


yrodiere:
> I suppose we will see the results in Sonar soon enough.

Sorry I didn't mention it; I tested it and it worked: https://sonarcloud.io/dashboard?branch=sonartest-yrodiere&id=org.hibernate.search%3Ahibernate-search-parent . The analysis didn't report any error and the coverage makes sense. There are reported issues, but that's unrelated (it's because false positives reported on the master branch are not taken into account on other branches).

yrodiere:
Merged, thanks!

PR #1979, opened 2019-05-13
HSEARCH-3213 @Spatial bridge set at type level are applied regardless of @IndexedEmbedded.includePaths

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3213

This problem was already fixed in Search 6; this PR simply changes a test to also check that the problem is also fixed for GeoPoint bridge (previously known as "spatial bridge").


fax4ever:
Me too. +1 to merge from me

yrodiere:
Merged, thanks

PR #1980, opened 2019-05-13
HSEARCH-2663 Null handling for elements in container bridges isn't consistent with non-container bridges

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2663

Based on #1974, which should be merged first.

Just a non-regression test checking we don't have this problem in Search 6 anymore.

### DIFF
--- a/integrationtest/mapper/pojo/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/nonregression/mapping/definition/IndexNullAsOnNumericContainer.java
+++ b/integrationtest/mapper/pojo/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/nonregression/mapping/definition/IndexNullAsOnNumericContainer.java
@@ -1,0 +1,92 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.integrationtest.mapper.pojo.nonregression.mapping.definition;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.hibernate.search.integrationtest.mapper.pojo.testsupport.util.rule.JavaBeanMappingSetupHelper;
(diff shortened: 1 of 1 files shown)


integrationtest/mapper/pojo/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/nonregression/mapping/definition/IndexNullAsOnNumericContainer.java
```diff
@@ -0,0 +1,92 @@
...
+		public Integer getId() {
+			return id;
+		}
+
+		public void setId(Integer id) {
+			this.id = id;
+		}
+
+		@GenericField(indexNullAs = "42")
+		public List<Integer> getIntegerList() {
```
fax4ever: Not related to this issue, but my concern is: shall we allow to use define an `indexNullAs` for a container (collection) field?

yrodiere: Not sure what you mean exactly.

Do you mean an `indexNullAs` value that would be used when the collection itself is null, as opposed to its elements? No, I don't think we should provide that feature. If someone wants that, they can just use a custom bridge.

yrodiere:
Merged, thanks!

integrationtest/mapper/pojo/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/nonregression/mapping/definition/IndexNullAsOnNumericContainer.java
fax4ever: Yes and maybe we should throw an exception if the user tried to do that.

yrodiere: They cannot do that currently, there is no API to do that. We only allow to set an `indexNullAs` value for when the *field* is `null`, which means more or less when the value passed to the bridge (an element of the collection) is `null`. When the collection is `null`, it's just ignored.

PR #1981, opened 2019-05-16
HSEARCH-3487 Support java.math.BigDecimal and java.math.BigInteger as field types

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3487


fax4ever:
@yrodiere skipping test for ES 5 now full CI works

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/ElasticsearchBigIntegerIndexFieldTypeContext.java
```diff
@@ -0,0 +1,49 @@
...
+import org.hibernate.search.backend.elasticsearch.types.projection.impl.ElasticsearchStandardFieldProjectionBuilderFactory;
+import org.hibernate.search.backend.elasticsearch.types.sort.impl.ElasticsearchStandardFieldSortBuilderFactory;
+import org.hibernate.search.engine.backend.types.converter.FromDocumentFieldValueConverter;
+import org.hibernate.search.engine.backend.types.converter.ToDocumentFieldValueConverter;
+
+class ElasticsearchBigIntegerIndexFieldTypeContext
+		extends AbstractElasticsearchScalarFieldTypeContext<ElasticsearchBigIntegerIndexFieldTypeContext, BigInteger> {
+
+	ElasticsearchBigIntegerIndexFieldTypeContext(ElasticsearchIndexFieldTypeBuildContext buildContext) {
+		super( buildContext, BigInteger.class, DataType.LONG );
```
yrodiere: I understand there is no support for arbitrary-precision integers in Elasticsearch, but if we want to make it work regardless, we should try to find a compromise.

Indexing as long is not a compromise, it's straight up surrender: users need to rely on `BigInteger` when they need values that *cannot* fit into a `long`. Which means indexing, here, will always fail.

For `BigDecimal`, the compromise was that we provided higher precision than `double` (64 bits instead of 53 bits), but still limited precision, and with a scale that is as limited as in double (in the case of the Elasticsearch backend) or to what an integer can hold (in the case of the Lucene backend).

For `BigInteger`, we have several possible approaches:

1. Index as a string or byte[], get arbitrary precision for projections at the cost of meaningless range queries or sorts.
2. Index as a `scaled_float`, get arbitrary precisions for projections (I think?) at the cost of lower precision for range queries or sorts.

Solution number 2 seems better than nothing. It wouldn't provide arbitrary precision, but at least it would address *some* use cases. Which is better than indexing as a long.

For Lucene, I think we could theoretically ask users to give us a precision upfront, and then index as an array of integers. I'm not sure how well this would work, but it would be an option. We should probably start with a long and an integer scale for now, though (just like `BigDecimal`). Later we will be able to add an explicit "precision" parameter and to use a different encoding when that precision cannot be achieved with a `long`.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/fieldtype/ElasticsearchDecimalScaleIT.java
```diff
@@ -0,0 +1,210 @@
...
+		// decimal scale has been set to 3, overriding the default
+		// decimal scale should not have any effects on stored valued for projections
+		// any exceeding decimal value zero will be ignored
+		SearchQuery<Object> query = indexManager.createSearchScope().query()
+				.asProjection( p -> p.field( "scaled" ) )
+				.predicate( p -> p.matchAll() )
+				.toQuery();
+		assertThat( query ).hasHitsExactOrder( BigDecimal.valueOf( 739.11111 ) );
+	}
+
```
yrodiere: One thing you forgot to test (and, I think, to implement): when targeting multiple indexes in a search query, there should be an error when the targeted field of a match or range query, or of a sort, is indexed with a different decimal scale in both indexes.

Maybe some of this will work in Elasticsearch; but I'm pretty sure it won't work in our Lucene backend.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/types/BigIntegerFieldTypeDescriptor.java
```diff
@@ -0,0 +1,81 @@
...
+
+import org.hibernate.search.integrationtest.backend.tck.testsupport.types.expectations.ExistsPredicateExpectations;
+import org.hibernate.search.integrationtest.backend.tck.testsupport.types.expectations.FieldProjectionExpectations;
+import org.hibernate.search.integrationtest.backend.tck.testsupport.types.expectations.FieldSortExpectations;
+import org.hibernate.search.integrationtest.backend.tck.testsupport.types.expectations.IndexNullAsMatchPredicateExpectactions;
+import org.hibernate.search.integrationtest.backend.tck.testsupport.types.expectations.IndexingExpectations;
+import org.hibernate.search.integrationtest.backend.tck.testsupport.types.expectations.MatchPredicateExpectations;
+import org.hibernate.search.integrationtest.backend.tck.testsupport.types.expectations.RangePredicateExpectations;
+
+public class BigIntegerFieldTypeDescriptor extends FieldTypeDescriptor<BigInteger> {
```
yrodiere: This is great, but following my comment about encoding big integers as something else than longs, I think there should also be a test for various scales, similar to `DecimalScaleIT`, but for `BigInteger`.

There, you could test various decimal scales, and also test indexing a value that doesn't fit into a long. Those are the most likely to be indexed when people use a `BigInteger`. For example `BigInteger.valueOf( String.valueOf( Long.MAX_VALUE ) + String.valueOf( Long.MAX_VALUE ) )`.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneBigIntegerFieldCodec.java
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	@Override
+	public Long encode(BigInteger value) {
+		return value.longValue();
+	}
+
+	@Override
+	public LuceneNumericDomain<Long> getDomain() {
+		return LuceneNumericDomain.LONG;
```
yrodiere: Same as Elasticsearch: storing big integers as longs won't do.

integrationtest/mapper/pojo/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/ScaledNumberFieldIT.java
```diff
@@ -0,0 +1,112 @@
...
+						.pathContext( ".notScalable" )
+						.failure(
+								"This property is mapped to a scaled number field, but with a value bridge that creates a non-BigDecimal or otherwise incompatible field",
+								"bind() method returned context '",
+								"expected '" + ScaledNumberIndexFieldTypeContext.class.getName() + "'"
+						)
+						.build()
+				);
+	}
+
```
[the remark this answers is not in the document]
fax4ever: I think that there are in the subsequent commit. Exactly the three you pointed out. I only changed the name of the third in `customBridge_invalidBinding`, since I think it would be more expressive.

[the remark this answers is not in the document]
fax4ever: The commit is 
> HSEARCH-3487 Test bridges on ScaledNumberField annotation Fabio Massimo Ercoli 5/7/19 9:00 AM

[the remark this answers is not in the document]
yrodiere: Right, my bad.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/model/JpaColumnScaleMappingIT.java
```diff
@@ -0,0 +1,86 @@
...
+		// decimal scale value provided by Hibernate ORM
+		// is not used if the field is not a decimal-number field
+		@Column(precision = 14, scale = 3)
+		private Integer notDecimal;
+
+		@GenericField
+		// decimal scale value provided by Hibernate ORM
+		// is used even if we use a @GenericField to annotate the field
+		@Column(precision = 14, scale = 3)
+		private BigDecimal general;
```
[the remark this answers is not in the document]
fax4ever: I think that we already do: https://github.com/hibernate/hibernate-search/pull/1981/files/c457fe0b3863b97256738e6cb0c80ebf4b3e22f4..cfb916b9eb247898663b1b697d250a840e7232c8#diff-6684ab46e636c099e3926fda4b289819R59

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/impl/esnative/PropertyMapping.java
```diff
@@ -172,4 +175,12 @@ public String getNormalizer() {
 	public void setNormalizer(String normalizer) {
 		this.normalizer = normalizer;
 	}
+
+	public Integer getScalingFactor() {
```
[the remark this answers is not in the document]
fax4ever: You did, but I've forgotten it, so my bad :)

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/model/JpaColumnScaleMappingIT.java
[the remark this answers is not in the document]
yrodiere: No, that one uses the default `decimalScale`. My example is different.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneBigDecimalFieldCodec.java
```diff
@@ -0,0 +1,99 @@
...
+
+	private Long unscale(BigDecimal value) {
+		BigDecimal rounded = value.setScale( decimalScale, RoundingMode.HALF_EVEN );
+		if ( value.scale() > decimalScale ) {
+			log.bigDecimalRounded( value, value.scale(), decimalScale );
+		}
+		return rounded.unscaledValue().longValue();
+	}
+
+	private static BigDecimal customStripTrailingZeros(BigDecimal bigDecimal) {
```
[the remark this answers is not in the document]
fax4ever: Yes scale to 0 only if all decimal digits are zeros. I added, the method because Elasticsearch does the same, see `DecimalScaleIT#decimalScale_doesNotAffectProjections_exceedingZeros`.
If you want we can remove it, accepting different behaviors among backends. But I would like to keep it.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/ElasticsearchBigIntegerIndexFieldTypeContext.java
fax4ever: @yrodiere would option 2 consist in using always a decimal scale less or equal to 0?

yrodiere: Yes. I guess we could allow strictly positive scales if it's easier that way, but that wouldn't be very useful.

fax4ever: Keeping scales negative or at most equals to zero, it would be more consistent with use of scale for BigDecimal. The only differences that I would make are: 

1 To forbid positive scales for BigInteger, since decimal digits are not supported with them.  
2. To make `0` the default value for scale, since I think most of the time they are used as long.

yrodiere: > To make `0` the default value for scale, since I think most of the time they are used as long.

I don't know where you're getting this from? The purpose of a `BigInteger` is to hold values that cannot be held by a `long`, be it for storage, or just for intermediary values in calculations. Therefore it's unlikely that a `BigInteger` used in an application will be convertible to a long without any scale: it will be greater than `Long.MAX_VALUE`.

Please don't use `0` as a default. If you're right we can add it later if we get new information, but if I'm right we won't be able to remove the default easily and many users will use a wrong mapping an only notice the problem at runtime when indexing.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/fieldtype/ElasticsearchDecimalScaleIT.java
fax4ever: Can I add those tests to `DecimalScaleIT`?
Actually I would avoid to make `MatchSearchPredicateIT`, `RangeSearchPredicateIT`, `FieldSearchSortIT` more complex.
But as you prefer.

fax4ever: I would like for these classes same as for `FieldSearchProjectionIT` make them using only the field created by the FieldTypeDescriptor(s). Moving the rest on different classes. What you think about it?

yrodiere: > Can I add those tests to `DecimalScaleIT`?
> Actually I would avoid to make `MatchSearchPredicateIT`, `RangeSearchPredicateIT`, `FieldSearchSortIT` more complex.
> But as you prefer.

I really dislike the idea of testing a single feature (match, range, ...) in many different classes; to me it just makes it harder to see what is expected from that feature. Because in the end, it's not the decimal scale feature you're testing, it really is the field compatibility checks when building predicates.

I won't stop you, but I think it hurts more than it helps. If it's about making test classes smaller, splitting `MatchSearchPredicateIT` into multiple classes in the same package would be a better solution in my opinion (note I'm not suggesting to do it in this PR).

> I would like for these classes same as for FieldSearchProjectionIT make them using only the field created by the FieldTypeDescriptor(s).

I do not understand this sentence.

fax4ever: Not a problem. 
About the last sentence. In those tests we have 2 kind of field:

1. Some generated by an instance of 'FieldTypeDescriptor', such as `supportedFieldModels`
2. Others declared explicitly. like 'whitespaceAnalyzedField'
I would like to move all fields of type 1 out of this test class.

My concern is about readability of the test itself. In my opinion is getting too complex.
But as you wish.

yrodiere: > I would like to move the latter out of this test class.

`FieldSearchProjectionIT` is about field projections, while `SearchProjectionIT` is about projections in general. You can't apply the same reasoning to match predicates, as the match predicates always apply to a field.

You could have `MatchSearchPredicateBaseIT` for your item 2, and `MatchSearchPredicateTypeIT` for item 1. But then the "base" tests would miss some very basic tests... it would just be weird. There might be a way to split these classes, but it's not as obvious as you're suggesting it is.

If you want to reorganize tests, please open a ticket and state the problem so we can discuss it. Trying to sneak this kind of organizational change in a PR about something else entirely is bound to simply lead to an inconsistent codebase ;)

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/dialect/Elasticsearch60TestDialect.java
```diff
@@ -50,4 +50,10 @@ public void setTemplatePattern(JsonObject object, String pattern) {
 		array.add( pattern );
 		object.add( "index_patterns", array );
 	}
+
+	@Override
+	public boolean supportLargeBigIntegerProjections() {
```
yrodiere: Does it work with ES 6.5? If so, I could upgrade our testing infrastructure to use ES 6.5 on AWS (I should do it anyway), and we could remove this commit entirely. We'll just avoid testing regularly on ES 6.4.

documentation/src/main/asciidoc/search-dsl.asciidoc
```diff
@@ -254,3 +254,12 @@ or skip analysis completely with `.skipAnalysis()`.
 
 See <<search-dsl-predicate>> for more information about how to create predicates
 and about the available options.
+
+=== Incompatible decimal scale
```
yrodiere: This really boils down to a different encoding, which has its own section above. I'll update this in another commit.

yrodiere:
I think the PR also needs rebasing :)

fax4ever:
@yrodiere I'll rebase it. Thanks for the further commits.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/dialect/Elasticsearch60TestDialect.java
fax4ever: With ES 6.5 it works. Let's drop this commit. Of course full CI would not work until we upgrade our testing infrastructure.

fax4ever:
@yrodiere I think I've done with the changes so far. Anything else to improve before we merge it? Thanks

yrodiere:
I added a small correction. If the build passes, it's good to merge! I disabled the tests on AWS Elasticsearch 6.4 for now, I'm working on upgrading the infrastructure.

fax4ever:
@yrodiere have you disabled AWS ES 6.4 from `ENVIRONMENT_SET=SUPPORTED` pipeline? I ran a new build for that:
http://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-1981/36/parameters

yrodiere:
@fax4ever Yes, I disabled that.

fax4ever:
@yrodiere by mistake I overrode the very last commit with the last correction on the branch.
I re-did it and squash with the last commit. Sorry for that.
Anyway I merged, thanks.

yrodiere:
No problem. Glad to see this merged. Thanks!

PR #1983, opened 2019-05-16
HSEARCH-3577 Remove support for longs when specifying the limit/offset of a Search query

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3577


gsmet:
Not sure I understand the rationale of this one? The code was already there and the worst that can happen is that Elasticsearch fixes the offset limitation.

yrodiere:
We exposed the ability to use longs, but behind the scenes it just wasn't implemented.

So no, the worst that can happen is not that it gets fixed, the worst that can happen is that it does *not* get fixed and we keep APIs that do not work forever.

I had a choice between implementing correct input checking and throwing appropriate exceptions, and adding the relevant tests, or just removing the APIs because, well, there's no feature behind them. I chose the latter.

yrodiere:
Merged, thanks

PR #1984, opened 2019-05-20
HSEARCH-3572 Add fetch methods to terminal steps of the query DSL

yrodiere:
[HSEARCH-3572](https://hibernate.atlassian.net//browse/HSEARCH-3572): Add fetch methods to terminal steps of the query DSL

~Based on #1983 ([HSEARCH-3577](https://hibernate.atlassian.net//browse/HSEARCH-3577)), which should be merged first.~ => Done

To see how this improves APIs, have a look at `IndexSearchDocumentRepositoryImpl.java` or `org.hibernate.search.documentation.gettingstarted.withhsearch.withoutanalysis.GettingStartedWithoutAnalysisIT#test`

PR #1985, opened 2019-05-21
HSEARCH-3578 Make the call to asXXX() (asEntity, asReference) optional in the Search DSL

yrodiere:
[HSEARCH-3578](https://hibernate.atlassian.net//browse/HSEARCH-3578): Make the call to asXXX() (asEntity, asReference) optional in the Search DSL

Based on #1984 , which should be merged first.

I'm not very happy with how complex the generics are becoming in DSL interfaces, but that's the only way to make `asEntity()` optional, unfortunately: we need to make the first DSL interface as complex as the one it allows to "take a shortcut to".


yrodiere:
Merged, thanks.

PR #1986, opened 2019-05-21
HSEARCH-3579 Failing to boot the ORM integration no longer throws an exception

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3579


yrodiere:
Merged, thanks

PR #1988, opened 2019-05-21
[PREVIEW] HSEARCH-3581 Indexing entities whose bytecode was enhanced by Hibernate ORM fails to load lazy properties

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3581

@gsmet The test seems to pass. event though I had to disable some parts due to what is probably bugs in Hibernate ORM. I'm looking for more things to test at this point. Any idea?

~I'll try to reproduce and report the ORM bugs once we're done with this.~ => Created https://hibernate.atlassian.net//browse/HSEARCH-3584

~Could you try this in Quarkus? I suspect it won't work as is, because I need reflection on the read methods added by the enhancer, and I do not think it is enabled by default in GraalVM. But maybe there's an easy way for you to enable it?~ Confirmed by @gsmet, it works.


yrodiere:
Jenkins, retest this please

yrodiere:
Confirmed by @gsmet, it works. Merging.

PR #1989, opened 2019-05-22
HSEARCH-3582 Make the build compatible with Eclipse

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3582

I took this opportunity to improve the Jenkinsfile a little bit, allowing to run the build in an arbitrary environment using a new "ENVIRONMENT_FILTER" parameter. I used it to run only the part of the build checking that the eclipse compiler works fine, see here: http://ci.hibernate.org/job/hibernate-search-yoann/job/HSEARCH-3582/25/


yrodiere:
Checked on Eclipse, it works. I'll merge and adjust if @gsmet runs into any problem while developing on Eclipse.

PR #1990, opened 2019-05-23
HSEARCH-3118 Follow-up: fix JDK11 build

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3118

Follows up on #1982 . The JDK11 build started to fail after merging #1982 because of powermock: http://ci.hibernate.org/job/hibernate-search/job/master/252/

This fixes the JDK11 build by getting rid of powermock, which required to introduce a delegator on top of `IndexWriter` in our code, so that we can mock that new interface instead of the `IndexWriter` directly. `IndexWriter` has a lot of final methods that would require powermock if we wanted to mock them.

I also cleaned up the resulting code a little bit.


yrodiere:
A full, successful build of this PR: http://ci.hibernate.org/blue/organizations/jenkins/hibernate-search-yoann/detail/HSEARCH-3118/18/pipeline

fax4ever:
Looks good to me. Happy about we've removed PowerMock.

fax4ever:
Merged thanks

PR #1991, opened 2019-05-27
HSEARCH-3076 + HSEARCH-3077 Check identifier bridge / marker builder / bridge builder generic type parameters on bootstrap

yrodiere:
* [HSEARCH-3076](https://hibernate.atlassian.net//browse/HSEARCH-3076): Search 6 groundwork - Check that IdentifierBridges are suitable for the property type before applying them
* [HSEARCH-3077](https://hibernate.atlassian.net//browse/HSEARCH-3077): Search 6 groundwork - Check that Marker/Bridge builders instances are suitable for the annotation before initializing them


integrationtest/mapper/pojo/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/ValueBridgeBaseIT.java
```diff
@@ -1,126 +0,0 @@
...
-					// Stub backend is not supposed to use 'indexNullAs' option
-					.add( "1", b -> b.field( "integer", null ) )
-					.preparedThenExecuted();
-		}
-		backendMock.verifyExpectationsMet();
-	}
-
-	@Test
-	public void indexNullAs_noParsing() {
-		@Indexed(index = INDEX_NAME)
```
fax4ever: Why do not move this test too?

yrodiere: I did. I just renamed it to `error_indexNullAs_noParsing` to match the naming of other methods in the same class.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/building/impl/PojoIndexModelBinderImpl.java
```diff
@@ -104,20 +105,16 @@
 		if ( builder == null ) {
 			defaultedBuilder = bridgeResolver.resolveIdentifierBridgeForType( typeModel );
```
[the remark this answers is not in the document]
yrodiere: It's considered bad practice to assign values to method parameters. A lot of static analysis tools, Sonar included, will raise warnings when you do that. I don't quite understand how it's a problem, but avoiding it doesn't hurt and makes things clearer, I suppose.

yrodiere:
Merged, thanks!

PR #1992, opened 2019-05-27
HSEARCH-3587 Rename MarkerMapping#builder to MarkerMapping#marker

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3587

Based on #1991 , which should be merged first.


yrodiere:
Merged, thanks!

PR #1994, opened 2019-05-28
HSEARCH-3048 Search 6 groundwork - Add the missing common field type options compared to Search 5

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3048

1. `norms` 
2. `searchable`
3. `doc_values` ( test only, it seems it had already been implemented well )
4. `termVector`
5. `precision_step` not exists anymore (from ES 5+)

~~options added => preview.~~ I think that all missing attributes have been handled now


engine/src/main/java/org/hibernate/search/engine/backend/types/Searchable.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+public enum Searchable {
+	/**
+	 * Use the backend-specific default.
+	 */
+	DEFAULT,
+	/**
+	 * Do not index the field value. This field can thus not be searched,
+	 * but one can still access its contents provided it is
+	 * {@link Projectable projectable}.
```
yrodiere: This doesn't really explain what "searchable" means. Maybe just state this:

```
	/**
	 * The field is not searchable, i.e. search predicates cannot be applied to the field.
         * <p>
         * This can save some disk space if you know the field is only used for projections or sorts.
         */
```

And below:

```
	/**
	 * The field is searchable, i.e. search predicates can be applied to the field.
         */
```

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneStringFieldCodec.java
```diff
@@ -25,15 +26,17 @@
...
+	private final boolean searchable;
 
 	private final FieldType fieldType;
 
 	private String indexNullAsValue;
 
 	private final Analyzer analyzerOrNormalizer;
 
-	public LuceneStringFieldCodec(boolean sortable, FieldType fieldType, String indexNullAsValue, Analyzer analyzerOrNormalizer) {
+	public LuceneStringFieldCodec(boolean sortable, boolean searchable, FieldType fieldType, String indexNullAsValue, Analyzer analyzerOrNormalizer) {
```
yrodiere: Here `searchable` is after `sortable`, while in every other codec it's the opposite.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneStringFieldCodec.java
```diff
@@ -49,7 +52,14 @@ public void encode(LuceneDocumentBuilder documentBuilder, String absoluteFieldPa
...
 
-		documentBuilder.addField( new Field( absoluteFieldPath, value, fieldType ) );
+		if ( searchable ) {
+			// searchable either projectable or not projectable
+			documentBuilder.addField( new Field( absoluteFieldPath, value, fieldType ) );
+		}
+		else if ( fieldType.stored() ) {
+			// not searchable and projectable
+			documentBuilder.addField( new StoredField( absoluteFieldPath, normalize( absoluteFieldPath, value ) ) );
+		}
```
yrodiere: Not sure this is necessary? The `fieldType` should already take that into account.

See `org.hibernate.search.backend.lucene.types.dsl.impl.LuceneStringIndexFieldTypeContext#getFieldType`.

In particular, if the field is not searchable, `indexOptions` should be forced to `IndexOptions.NONE`.
You could disable norms and term vectors too, at least if they are not explicitly set, but I'm not sure it matters if `indexOptions` is set to `NONE`.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/spatial/AbstractSpatialWithinSearchPredicateIT.java
```diff
@@ -12,6 +12,7 @@
 import org.hibernate.search.engine.backend.document.DocumentElement;
 import org.hibernate.search.engine.backend.document.IndexFieldReference;
 import org.hibernate.search.engine.backend.document.model.dsl.IndexSchemaElement;
+import org.hibernate.search.engine.backend.types.Searchable;
 import org.hibernate.search.engine.backend.types.Sortable;
 import org.hibernate.search.engine.backend.types.Projectable;
```
yrodiere: Commit message: "spacial" => "spatial"

documentation/src/main/asciidoc/mapper-orm-mapping.asciidoc
```diff
@@ -174,6 +174,11 @@ Whether the field can be <<mapper-orm-query-projection, projected on>>,
 i.e. whether the field value is stored in the index to allow later retrieval when querying.
 +
 Value: `Projectable.YES`, `Projectable.NO`, `Projectable.DEFAULT`.
+`searchable`::
+Whenever the field can be searched on.
```
yrodiere: "Whenever" => "Whether"

documentation/src/main/asciidoc/mapper-orm-mapping.asciidoc
```diff
@@ -174,6 +174,11 @@ Whether the field can be <<mapper-orm-query-projection, projected on>>,
...
+i.e. whether the field is indexed. So the document will be searchable using this field, using it in the query predicates.
```
yrodiere: 1. A sentence may start with a conjunction, such as "so", but frankly it feels weird in the (relatively formal) context of documentation.
2. You're basically saying "Searchable: whether the field can be searched on, i.e. whether the field is indexed, so the document will be searchable". A bit confused, don't you think?

Try something like this:

```
i.e. whether the field is indexed in order to allow applying predicates later when querying.
```

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/spatial/AbstractSpatialWithinSearchPredicateIT.java
```diff
@@ -172,7 +172,7 @@ protected void initData() {
 		UnsearchableFieldsIndexMapping(IndexSchemaElement root) {
 			geoPoint = root.field(
 					// make the field not searchable
-					"geoPoint", f -> f.asGeoPoint().sortable( Sortable.YES ).projectable( Projectable.YES ).searchable( Searchable.NO )
```
yrodiere: Yes, that's weird. We suspect a bug, we hit it somewhere else. I'll report it, it's on my to-do list.

engine/src/main/java/org/hibernate/search/engine/backend/types/Norms.java
```diff
@@ -0,0 +1,28 @@
...
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.engine.backend.types;
+
+/**
+ * Whether index time scoring information for the field should be stored or not.
+ * <p>
+ * Not storing the norms can save a considerable amount of memory.
+ * Storing them can make scoring queries faster.
```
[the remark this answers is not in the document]
fax4ever: With `Storing them` I meant `enabling` and not `disabling` the norms. :P 
Anyway, let's change it as you suggested. In your version it's more clear the dichotomy of the two sentences.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/ElasticsearchStringIndexFieldTypeContext.java
```diff
@@ -140,4 +142,14 @@ public ElasticsearchStringIndexFieldTypeContext indexNullAs(String indexNullAs)
 	protected ElasticsearchStringIndexFieldTypeContext thisAsS() {
 		return this;
 	}
+
+	private boolean resolveNorms() {
+		if ( Norms.YES.equals( norms ) ) {
```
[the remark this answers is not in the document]
fax4ever: Yes, it does.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/dsl/impl/LuceneStringIndexFieldTypeContext.java
```diff
@@ -126,10 +129,20 @@ private LuceneAnalysisDefinitionRegistry getAnalysisDefinitionRegistry() {
...
+		}
+		return ( analyzerName != null );
+	}
+
+	private static FieldType getFieldType(boolean projectable, boolean analyzed, boolean norms) {
 		FieldType fieldType = new FieldType();
 		if ( analyzed ) {
-			// TODO HSEARCH-3048 take into account the norms and term vectors options
+			// TODO HSEARCH-3048 take into account term vectors option
 			fieldType.setOmitNorms( false );
```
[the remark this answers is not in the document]
fax4ever: I would remove it when I handle the term vectors option. The issue number means that the TODO is supposed to be removed with the current issue.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/mapping/ElasticsearchFieldAttributesIT.java
```diff
@@ -0,0 +1,110 @@
...
+	private final ElasticsearchTestDialect dialect = ElasticsearchTestDialect.get();
+
+	@Rule
+	public SearchSetupHelper setupHelper = new SearchSetupHelper();
+
+	@Rule
+	public ElasticsearchClientSpy clientSpy = new ElasticsearchClientSpy();
+
+	@Test
+	public void addUpdateDelete_routing() {
```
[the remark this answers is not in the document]
fax4ever: You're right. I did.

engine/src/main/java/org/hibernate/search/engine/backend/types/Norms.java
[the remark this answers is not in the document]
yrodiere: Right, that was wrong.

Even then, enabling norms should not speed up queries, since norms won't be computed at query time when they are disabled. At least there's no *obvious* reason to gain performance from that, as there's no extra cost at query time when norms are disabled.

documentation/src/main/asciidoc/mapper-orm-mapping.asciidoc
```diff
@@ -260,6 +260,19 @@ NOTE: Using Hibernate ORM mapping,
 a default `decimalScale` is taken automatically from the underlying `scale` value of the relative SQL `@Column`,
 using the Hibernate ORM metadata. The value could be overridden explicitly using the `decimalScale` attribute.
 
+==== Norms creation
+
+For both `@FullTextField` and `@KeywordField` fields it is possible to control the creation of the norms.
+The norms are index time scoring information for the fields.
+
+Storing them can make scoring queries faster.
```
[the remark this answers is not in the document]
fax4ever: I took this sentence from the documentation of Hibernate Search 5. 
As before, let's replace it with:
> Enabling norms will improve the quality of scoring

documentation/src/main/asciidoc/mapper-orm-mapping.asciidoc
```diff
@@ -260,6 +260,19 @@ NOTE: Using Hibernate ORM mapping,
 a default `decimalScale` is taken automatically from the underlying `scale` value of the relative SQL `@Column`,
 using the Hibernate ORM metadata. The value could be overridden explicitly using the `decimalScale` attribute.
 
+==== Norms creation
```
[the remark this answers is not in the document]
fax4ever: Sure. I think that's a good idea to split annotation descriptions from attribute descriptions.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneStringFieldCodec.java
fax4ever: You're right. It's better to keep the order consistent.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneStringFieldCodec.java
fax4ever: Ok. It makes sense to handle both projectable and searchable using the type created by `LuceneStringIndexFieldTypeContext#getFieldType`. Let's do that.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/spatial/AbstractSpatialWithinSearchPredicateIT.java
fax4ever: I made the same mistake again :P

documentation/src/main/asciidoc/mapper-orm-mapping.asciidoc
fax4ever: it sounds better

fax4ever:
@yrodiere thanks for the review. I applied the changes here. 
I'm going to add further options  in the subsequent commits.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/spatial/AbstractSpatialWithinSearchPredicateIT.java
yrodiere: FYI it's getting fixed: https://bugs.eclipse.org/bugs/show_bug.cgi?id=547807

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/mapping/LuceneFieldAttributesIT.java
```diff
@@ -53,7 +56,44 @@ public void setup() {
...
+
+		field = document.getField( "termVector" );
+		Assertions.assertThat( field.fieldType().storeTermVectors() ).isTrue();
+		Assertions.assertThat( field.fieldType().storeTermVectorPositions() ).isFalse();
+		Assertions.assertThat( field.fieldType().storeTermVectorOffsets() ).isFalse();
+		Assertions.assertThat( field.fieldType().storeTermVectorPayloads() ).isFalse();
+
+		field = document.getField( "moreOptions" );
+		Assertions.assertThat( field.fieldType().storeTermVectors() ).isTrue();
+		// TODO these are not true:
```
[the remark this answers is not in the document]
fax4ever: Probably I did something wrong trying to do that in my very last commit.

I used an `NGramTokenizer`, that it is supposed to have `PositionIncrementAttribute`, `PositionLengthAttribute` and `OffsetAttribute`, combined with a `TokenOffsetPayloadTokenFilter', that is supposed to have `PayloadAttribute` and  `OffsetAttribute`.

Uncommenting the rows we still got an error.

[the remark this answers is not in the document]
yrodiere: Ok, the problem is here: `org.hibernate.search.backend.lucene.search.extraction.impl.ReusableDocumentStoredFieldVisitor#stringField`.
The information is not available from FieldInfo and thus is not propagated to the field type of projected fields.

Checking this correctly would apparently require a more sophisticated approach. Let's skip that for now, just leave the TODO and maybe we'll do it one day.

fax4ever:
@yrodiere thanks for the review, anything else to improve here?

yrodiere:
Merged, thanks

fax4ever:
thank you

PR #1996, opened 2019-05-31
HSEARCH-1467 Test IndexedEmbedded with different depths

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-1467

### DIFF
--- a/integrationtest/mapper/pojo/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/nonregression/mapping/definition/IndexedEmbeddedDepthIT.java
+++ b/integrationtest/mapper/pojo/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/nonregression/mapping/definition/IndexedEmbeddedDepthIT.java
@@ -1,0 +1,105 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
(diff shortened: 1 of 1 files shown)


integrationtest/mapper/pojo/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/IndexedEmbeddedBaseIT.java
```diff
@@ -484,6 +484,79 @@ public IndexedEmbeddedLevel1 getLevel1() {
 		);
 	}
 
+	@Test
+	public void maxDepth_differentDepths() {
```
yrodiere: Could you please add this test to a new class in the `nonregression` package? Also please annotate with `@TestForIssue`.

It helps to understand why a test was added in the first place, if it ever starts failing again.

fax4ever: There was also a not used variable...

yrodiere:
Merged, thanks!

PR #1998, opened 2019-05-31
HSEARCH-1467 Test IndexedEmbedded with different depths

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-1467
Same test of #1996 for Hibernate Search 5.


yrodiere:
Merged, thanks!

PR #1999, opened 2019-05-31
HSEARCH-1467 Make MassIndexer#start return a CompletableFuture

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-2539


yrodiere:
Looks good, merged. Thanks.

I makes me think we probably should give users a way to pick their own executor and call `java.util.concurrent.CompletableFuture#runAsync(java.lang.Runnable, java.util.concurrent.Executor)`, since I'm not sure using the global one is always a good idea, but that can wait until we know more about use cases.

PR #2001, opened 2019-06-02
HSEARCH-3387 Add normalizedTypeOnMissingValueUse discretionary tck

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3387

### DIFF
--- a/integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/dialect/ElasticsearchTestDialect.java
+++ b/integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/dialect/ElasticsearchTestDialect.java
@@ -63,4 +63,8 @@ public interface ElasticsearchTestDialect {
 	default boolean worksFineWithStrictGraterThanRangedQueriesOnScaledFloatField() {
 		return true;
 	}
+
+	default boolean normalizedTypeOnMissingValueUse() {
+		return false;
+	}
 }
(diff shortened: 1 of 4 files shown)


integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/dialect/ElasticsearchTestDialect.java
```diff
@@ -63,4 +63,8 @@ static String getClusterVersion() {
 	default boolean worksFineWithStrictGraterThanRangedQueriesOnScaledFloatField() {
 		return true;
 	}
+
+	default boolean normalizedTypeOnMissingValueUse() {
+		return false;
```
yrodiere: As far as I understand, you effectively disabled tests for Elasticsearch, since this is not overridden anywhere. If that was your goal, I think you shouldn't have added a method in the dialect, since dialect is irrelevant.

Anyway, I only mentioned disabling tests in the ticket **if the problem is solved in later Elasticsearch versions**. Only then would it be acceptable to close the ticket as fixed, since the feature would work with the latest Elasticsearch versions, at least. Right now it seems it wouldn't work at all, with any version.
If the problem is not solved in Elasticsearch, we should try to work around the proble, as you did in HSEARCH-3255.
Since that is not possible (we can't normalize on the client side), I think we should leave our ticket open, report the bug to Elasticsearch (unless it's already done) and link our ticket to their bug report. Then we'll have to wait, or contribute if we can.

fax4ever: I agree with you. There is no reason to a new TCK feature method if it doesn't work for any version of Elasticsearch. I think we could contribute to Elasticsearch project, time permitting.

PR #2006, opened 2019-06-04
HSEARCH-3254 Handle .onMissingValue().use() properly for string fields in Lucene

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3254

### DIFF
--- a/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/sort/FieldSearchSortIT.java
+++ b/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/sort/FieldSearchSortIT.java
@@ -37,7 +37,6 @@ import org.hibernate.search.engine.search.query.SearchQuery;
 import org.hibernate.search.engine.search.dsl.sort.SearchSortContainerContext;
 import org.hibernate.search.integrationtest.backend.tck.testsupport.util.InvalidType;
 import org.hibernate.search.integrationtest.backend.tck.testsupport.util.StandardFieldMapper;
-import org.hibernate.search.integrationtest.backend.tck.testsupport.util.TckConfiguration;
 import org.hibernate.search.integrationtest.backend.tck.testsupport.util.ValueWrapper;
 import org.hibernate.search.integrationtest.backend.tck.testsupport.util.rule.SearchSetupHelper;
 import org.hibernate.search.util.common.SearchException;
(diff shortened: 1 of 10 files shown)


fax4ever:
@yrodiere Hi. I'm adding more tests for the coverage.

fax4ever:
done

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/assertion/SearchHitAssert.java
```diff
@@ -0,0 +1,44 @@
...
+	SearchHitAssert(H actual) {
+		this.actual = actual;
+	}
+
+	public SearchHitAssert<H> as(String description) {
+		this.description = description;
+		return this;
+	}
+
+	public void hasDocRefHit(String indexName, String id, String... orIds) {
```
yrodiere: `isInDocRefHits` would probably be more in line with what this method does.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/types/NormalizedStringFieldTypeDescriptor.java
```diff
@@ -71,13 +71,10 @@
...
-		return Optional.of( new FieldSortExpectations<>(
-				"Cecilia", "george", "Stefany",
-				"aaron",
-				( normalize ) ? "Daniel" : "daniel",
-				( normalize ) ? "Roger" : "roger",
-				"zach"
-		) );
+		return Optional.of( ( normalize ) ?
+				new FieldSortExpectations<>( "Cecilia", "george", "Stefany", "aaron", "Daniel", "Roger", "zach" ) :
+				new FieldSortExpectations<>( "cecilia", "george", "stefany", "aaron", "daniel", "roger", "zach" )
```
yrodiere: This looks wrong. Originally the first three values were about testing sorts, whithout considering missing values.
Then you added something that uses them as replacement for missing values, and when it failed, you lowercased everything.

As a result, we no longer test sorts with varying case in Elasticsearch, even when missing values are not involved and it's supposed to work. That's not a great idea.

Maybe you could move the lowercasing operation to the test method that needs it?
I.e. do something like:

```java
docRefHits = simpleQuery( b -> b.byField( fieldPath ).asc().onMissingValue().use( normalizeIfNecessary( fieldModel.document2Value.indexedValue ) ) ).fetchHits(); 
```

and add this method:

```java
@SuppressWarning("unchecked")
public <T> normalizeMissingValueReplacementIfNecessary(T value) {
    if ( !TckConfiguration.get().getBackendFeatures().normalizeStringMissingValues() && String.class.isInstance ( value ) ) {
        // The backend doesn't normalize missing value replacements automatically, we have to do it ourselves
        // TODO HSEARCH-3387 Remove this once all backends correctly normalize missing value replacements
        return (T) missingValue.toLowercase( Locale.ROOT );
    }
    else {
        return value;
    }
}
```

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/sort/FieldSearchSortIT.java
```diff
@@ -171,6 +171,42 @@ public void byField() {
 		}
 	}
 
+	@Test
+	public void byField_use_alreadyExists_missingValue() {
```
yrodiere: Please add `@TestForIssue`: this test is very specific and a bit obscure, we may need to come back to the ticket and PR one day.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/sort/FieldSearchSortIT.java
```diff
@@ -172,38 +186,68 @@ public void byField() {
 	}
 
 	@Test
-	public void byField_use_alreadyExists_missingValue() {
-		for ( ByTypeFieldModel<?> fieldModel : indexMapping.supportedFieldModels ) {
+	public void byField_multipleEmpty_missingValue() {
```
yrodiere: Same here, `@TestForIssue` would be nice.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/sort/FieldSearchSortIT.java
```diff
@@ -172,38 +186,68 @@ public void byField() {
...
+			assertThat( docRefHits ).ordinals( 3, 4, 5, 6 ).hasDocRefHitsAnyOrder( MULTIPLE_EMPTY_INDEX_NAME, EMPTY_1, EMPTY_2, EMPTY_3, EMPTY_4 );
+		}
+	}
 
-			assertThat( docRefHits ).ordinal( 0 ).as( "Must be DOCUMENT_1 or the EMPTY one" ).hasDocRefHit( INDEX_NAME, EMPTY, DOCUMENT_1 );
-			assertThat( docRefHits ).ordinal( 1 ).as( "Must be DOCUMENT_1 or the EMPTY one" ).hasDocRefHit( INDEX_NAME, EMPTY, DOCUMENT_1 );
-			assertThat( docRefHits ).ordinal( 2 ).as( "Must be DOCUMENT_2" ).hasDocRefHit( INDEX_NAME, DOCUMENT_2 );
-			assertThat( docRefHits ).ordinal( 3 ).as( "Must be DOCUMENT_3" ).hasDocRefHit( INDEX_NAME, DOCUMENT_3 );
+	@Test
+	public void byField_multipleEmpty_alreadyExists_missingValue() {
```
yrodiere: Same here, `@TestForIssue` would be nice.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/sort/FieldSearchSortIT.java
```diff
@@ -172,38 +186,68 @@ public void byField() {
 	}
 
 	@Test
-	public void byField_use_alreadyExists_missingValue() {
```
yrodiere: Not sure why you are replacing a test that you just added? Shouldn't you squash the two commits, if the first one ends up being useless?

fax4ever:
@yrodiere thanks for the review. I think I've addressed the comments.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/types/NormalizedStringFieldTypeDescriptor.java
fax4ever: You're right! Predicate values are not normalized, but  indexed values are.

yrodiere:
Merged, thanks!

PR #2008, opened 2019-06-06
HSEARCH-3591 Expect a function as parameter for SearchQueryContext#sort

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3591

Please review commit by commit, as commit messages explain why I do what I do.


engine/src/main/java/org/hibernate/search/engine/search/dsl/query/impl/DefaultSearchQueryContext.java
```diff
@@ -19,11 +19,11 @@
 						DefaultSearchQueryContext<H, C>,
 						H,
 						SearchPredicateFactoryContext,
-						SearchSortContainerContext,
+		SearchSortFactoryContext,
```
fax4ever: It seems not aligned

yrodiere:
Thanks. I addressed your comment and rebased. Waiting for CI.

yrodiere:
Merged, thanks!

PR #2010, opened 2019-06-12
HSEARCH-3616 + HSEARCH-3617 DSL fixes

yrodiere:
Just some fixes for problems I noticed while working on the documentation.

* [HSEARCH-3616](https://hibernate.atlassian.net//browse/HSEARCH-3616): Accept Collection<?> instead of Collection<Object>  in the ID predicate
* [HSEARCH-3617](https://hibernate.atlassian.net//browse/HSEARCH-3617): Search 6 groundwork - Remove the by(SearchSort) method from the sort DSL

### DIFF
--- a/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/sort/SearchSortIT.java
+++ b/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/sort/SearchSortIT.java
@@ -166,13 +166,6 @@ public class SearchSortIT {
 		assertThat( query )
 				.hasDocRefHitsExactOrder( INDEX_NAME, FIRST_ID, SECOND_ID, THIRD_ID, EMPTY_ID );
 
-		query = scope.query()
-				.predicate( f -> f.matchAll() )
-				.sort( f -> f.by( sortAsc ) )
-				.toQuery();
-		assertThat( query )
-				.hasDocRefHitsExactOrder( INDEX_NAME, FIRST_ID, SECOND_ID, THIRD_ID, EMPTY_ID );
-
 		SearchSort sortDesc = scope.sort()
 				.byField( "string" ).desc().onMissingValue().sortLast()
 				.toSort();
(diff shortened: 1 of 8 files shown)


integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/sort/SearchSortIT.java
```diff
@@ -183,39 +176,37 @@ public void separateSort() {
...
 
 		Assertions.assertThat( cache ).hasValue( null );
 
+		StubMappingScope scope = indexManager.createScope();
 		SearchQuery<DocumentReference> query;
 
-		query = simpleQuery( cachingContributor );
+		query = scope.query().predicate( f -> f.matchAll() )
+				.sort( cachingContributor.apply( scope.sort() ) )
+				.toQuery();
```
fax4ever: Probably it's not relative do this pull request.

But I think that with this code we're not actually testing the caching of a `SearchSort` by Hibernate Search.
It looks like we're testing our `cachingContributor' function instead.

I believe that reuse the same `SearchSort` two times would be the same, but simpler.

fax4ever: Since caching is in an internal behavior. We could test the reuse of an instance, not the caching.

yrodiere: You're right. This used to make sense when we used "caching lambdas", but it doesn't make much sense anymore.

When you're done adding more relevant tests in your next PR, maybe you can remove that test?

fax4ever: Yep! Let's merge it. I'll handle it in [HSEARCH-3192](https://hibernate.atlassian.net/browse/HSEARCH-3192).

PR #2013, opened 2019-06-12
HSEARCH-3597 Document all available predicates/sorts/projections and query DSL basics

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3597

Based on #2012, which should be merged first.

The easiest way to review this PR is probably to build the documentation and then read the "Searching" section:

```
mvn clean install -DskipTests -pl documentation -am
xdg-open ./documentation/target/asciidoctor/en-US/html_single/index.html
```

PR #2015, opened 2019-06-14
HSEARCH-3192 Allow "caching" predicates/sorts/projections for re-use

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3192

I'm waiting for #2013 is merged to add a commit with the documentation.

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchDistanceToFieldProjection.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchDistanceToFieldProjection.java
@@ -7,6 +7,7 @@
 package org.hibernate.search.backend.elasticsearch.search.projection.impl;
 
 import java.util.Optional;
+import java.util.Set;
 import java.util.regex.Pattern;
 
 import org.hibernate.search.backend.elasticsearch.gson.impl.JsonAccessor;
@@ -28,6 +29,7 @@ class ElasticsearchDistanceToFieldProjection implements ElasticsearchSearchProje
 
 	private static final Pattern NON_DIGITS_PATTERN = Pattern.compile( "\\D" );
 
+	private final Set<String> indexNames;
 	private final String absoluteFieldPath;
 
 	private final GeoPoint center;
@@ -36,7 +38,8 @@ class ElasticsearchDistanceToFieldProjection implements ElasticsearchSearchProje
 
 	private final String scriptFieldName;
 
-	ElasticsearchDistanceToFieldProjection(String absoluteFieldPath, GeoPoint center, DistanceUnit unit) {
+	ElasticsearchDistanceToFieldProjection(Set<String> indexNames, String absoluteFieldPath, GeoPoint center, DistanceUnit unit) {
+		this.indexNames = indexNames;
 		this.absoluteFieldPath = absoluteFieldPath;
 		this.center = center;
 		this.unit = unit;
@@ -92,6 +95,11 @@ class ElasticsearchDistanceToFieldProjection implements ElasticsearchSearchProje
 	public Double transform(LoadingResult<?> loadingResult, Double extractedData,
 			SearchProjectionTransformContext context) {
 		return extractedData;
+	}
+
+	@Override
+	public Set<String> getIndexNames() {
+		return indexNames;
 	}
 
 	@Override
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/SearchProjectionBackendContext.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/SearchProjectionBackendContext.java
@@ -6,55 +6,45 @@
  */
 package org.hibernate.search.backend.elasticsearch.search.projection.impl;
 
+import java.util.Set;
+
 import org.hibernate.search.engine.search.projection.spi.SearchProjectionBuilder;
 
 import com.google.gson.Gson;
 
 public class SearchProjectionBackendContext {
 
-	private final ElasticsearchDocumentReferenceProjectionBuilder documentReferenceProjectionBuilder;
-	@SuppressWarnings("rawtypes")
-	private final ElasticsearchEntityProjectionBuilder objectProjectionBuilder;
-	@SuppressWarnings("rawtypes")
-	private final ElasticsearchReferenceProjectionBuilder referenceProjectionBuilder;
-	private final ElasticsearchScoreProjectionBuilder scoreProjectionBuilder;
-	private final ElasticsearchSourceProjectionBuilder sourceProjectionBuilder;
-	private final ElasticsearchExplanationProjectionBuilder explanationProjectionBuilder;
+	private final DocumentReferenceExtractorHelper documentReferenceExtractorHelper;
+	private final Gson userFacingGson;
 
 	@SuppressWarnings("rawtypes")
 	public SearchProjectionBackendContext(DocumentReferenceExtractorHelper documentReferenceExtractorHelper,
 			Gson userFacingGson) {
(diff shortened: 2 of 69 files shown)


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
```diff
@@ -29,15 +29,15 @@
 
 	private final Set<ElasticsearchIndexModel> indexModels;
 	private final Set<String> hibernateSearchIndexNames;
-	private final Set<URLEncodedString> elasticsearchIndexNames;
```
yrodiere: They were named that way for a reason. These names are not the ones used in Hibernate Search, they were normalized to fit Elasticsearch's requirements (no uppercase, in particular).

I don't think it's wise to just call them "index names", as they could be mistaken for the names used internally in Hibernate Search.

Also, when displaying error messages, I think you should use `hibernateSearchIndexNames`, since those are the names that the user will use in the mapping and configuration files.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchDistanceToFieldProjection.java
```diff
@@ -36,7 +39,8 @@
 
 	private final String scriptFieldName;
 
-	ElasticsearchDistanceToFieldProjection(String absoluteFieldPath, GeoPoint center, DistanceUnit unit) {
+	ElasticsearchDistanceToFieldProjection(Set<URLEncodedString> indexNames, String absoluteFieldPath, GeoPoint center, DistanceUnit unit) {
```
yrodiere: Wouldn't an abstract class make sense here? You could even turn `ElasticsearchSearchProjection` into an abstract class.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/SearchProjectionBackendContext.java
```diff
@@ -6,55 +6,48 @@
...
-		return documentReferenceProjectionBuilder;
+	ElasticsearchDocumentReferenceProjectionBuilder createDocumentReferenceProjectionBuilder(Set<URLEncodedString> indexNames) {
+		return new ElasticsearchDocumentReferenceProjectionBuilder( indexNames, documentReferenceExtractorHelper );
 	}
 
 	@SuppressWarnings("unchecked")
-	<E> ElasticsearchEntityProjectionBuilder<E> getEntityProjectionBuilder() {
-		return objectProjectionBuilder;
+	<E> ElasticsearchEntityProjectionBuilder<E> createEntityProjectionBuilder(Set<URLEncodedString> indexNames) {
+		return new ElasticsearchEntityProjectionBuilder( indexNames, documentReferenceExtractorHelper );
```
yrodiere: Missing `<>`. And the `@SuppressWarnings` is no longer necessary.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/SearchProjectionBackendContext.java
```diff
@@ -6,55 +6,48 @@
...
 	}
 
 	@SuppressWarnings("unchecked")
-	<R> ElasticsearchReferenceProjectionBuilder<R> getReferenceProjectionBuilder() {
-		return referenceProjectionBuilder;
+	<R> ElasticsearchReferenceProjectionBuilder<R> createReferenceProjectionBuilder(Set<URLEncodedString> indexNames) {
+		return new ElasticsearchReferenceProjectionBuilder( indexNames, documentReferenceExtractorHelper );
```
yrodiere: Missing `<>`. And the `@SuppressWarnings` is no longer necessary.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/SearchPredicateIT.java
```diff
@@ -37,8 +37,8 @@
 
 public class SearchPredicateIT {
 
-	private static final String INDEX_NAME = "IndexName";
-	private static final String ANOTHER_INDEX_NAME = "AnotherIndexName";
```
yrodiere: See. You had to do this because you used the Elasticsearch index names, not the Hibernate Search index names.

This change shouldn't be necessary. Please remove this change and use the correct index names in the Elasticsearch backend.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/projection/SearchProjectionIT.java
```diff
@@ -59,8 +59,8 @@
  */
 public class SearchProjectionIT extends EasyMockSupport {
 
-	private static final String INDEX_NAME = "IndexName";
-	private static final String ANOTHER_INDEX_NAME = "AnotherIndexName";
+	private static final String INDEX_NAME = "indexname";
+	private static final String ANOTHER_INDEX_NAME = "anotherindexname";
```
yrodiere: Same as above, this change should be removed.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/sort/SearchSortIT.java
```diff
@@ -53,8 +53,8 @@
  */
 public class SearchSortIT {
 
-	private static final String INDEX_NAME = "IndexName";
-	private static final String ANOTHER_INDEX_NAME = "AnotherIndexName";
+	private static final String INDEX_NAME = "indexname";
+	private static final String ANOTHER_INDEX_NAME = "anotherindexname";
```
yrodiere: Same as above, this change should be removed.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchDistanceToFieldProjection.java
fax4ever: When I wrote it I was thinking to do so. However then I asked to myself: what can we share among all `ElasticsearchSearchProjection`s?
I think that extract an abstract class just for the field `Set<String> indexNames;` is too much.

fax4ever:
@yrodiere thakns for the review. You're right, it is better to rely on Hibernate Search's index names.
I think I've applied the changes here. I'm not sure whether it is the case to extract an abstract class for Elasticsearch projection predicates.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchDistanceToFieldProjection.java
yrodiere: Just to be clear, I wasn't talking about just Elasticsearch, I think it would make sense for Lucene too.

I feel like turning `ElasticsearchSearchProjection` wouldn't be such a big change, especially considering all the other changes you had to make and considering that it would bring projections closer to `ElasticsearchSearchPredicate` and `ElasticsearchSearchSort`.

But I can understand your point of view too. It can stay that way, it's not a big deal.

yrodiere:
Looks good to me!

Sonar reported a few duplicated blocks, but that's related to the ElasticsearchSearchProjection/LuceneSearchProjection, and if we don't add an abstract class there's not much we can do about it.

Feel free to add the abstract class or to merge as is.

fax4ever:
@yrodiere merged thanks!

PR #2016, opened 2019-06-18
HSEARCH-3618 Distance projection may lead to missing hits in the Lucene and Elasticsearch backends

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3618

Turns out there were two very different problems, but leading to the same symptoms.

A full build is running right now to check other versions of Elasticsearch: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search-yoann/detail/HSEARCH-3618/6/


fax4ever:
Looks good!

yrodiere:
Merged, thanks

PR #2017, opened 2019-06-19
HSEARCH-3619 Clean up the code in org.hibernate.search.mapper.pojo.mapping.impl

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3619

@fax4ever The changeset is quite large, but it's mainly about restructuring code. And I think I was the only one to ever touch most of that code. So if that's okay with you, I'll just go ahead and merge it without review; I'll simply have a look at the Sonar report. I think a review would just waste your time in this specific case.


yrodiere:
I just addressed a few problems reported by sonar. Will merge as soon as CI goes green.

fax4ever:
@yrodiere  It's OK for me.

PR #2018, opened 2019-06-19
HSEARCH-3230 Clarify the behavior of Lucene GeoPoint field codecs with respect to searchable/sortable/projectable

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3230

Actually I think that the issue is out of date.
Anyway I take the change to add some tests. Maybe too many. I think I could merge them into the current ones.

About all the integration tests pass splitting: 
~~~ java
if ( searchable ) {
  documentBuilder.addField( new LatLonPoint( absoluteFieldPath, value.getLatitude(), value.getLongitude() ) );
}

if ( sortable ) {
  documentBuilder.addField( new LatLonDocValuesField( absoluteFieldPath, value.getLatitude(), value.getLongitude() ) );
}
~~~

I was wrong. All the integration tests pass, with the exception that ones on `exists` predicate.
Anyway I tried to optimize a bit field creations. It is possible that a more significant optimization it is possible changing a bit the implementation of `exists` predicate.

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneGeoPointFieldCodec.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneGeoPointFieldCodec.java
@@ -15,10 +15,13 @@ import org.apache.lucene.document.LatLonDocValuesField;
 import org.apache.lucene.document.LatLonPoint;
 import org.apache.lucene.document.StoredField;
 import org.apache.lucene.index.IndexableField;
+import org.apache.lucene.index.Term;
 import org.apache.lucene.search.DocValuesFieldExistsQuery;
 import org.apache.lucene.search.Query;
+import org.apache.lucene.search.TermQuery;
 
 import org.hibernate.search.backend.lucene.document.impl.LuceneDocumentBuilder;
+import org.hibernate.search.backend.lucene.util.impl.LuceneFields;
 import org.hibernate.search.engine.spatial.GeoPoint;
 
 public final class LuceneGeoPointFieldCodec implements LuceneFieldCodec<GeoPoint> {
@@ -54,10 +57,14 @@ public final class LuceneGeoPointFieldCodec implements LuceneFieldCodec<GeoPoint
 			documentBuilder.addField( new StoredField( getLongitudeAbsoluteFieldPath( absoluteFieldPath ), value.getLongitude() ) );
 		}
 
-		if ( sortable || searchable ) {
-			// Generally the LatLonDocValuesField field is required only for sorting,
-			// but we rely on this field to support exists predicate too.
+		if ( sortable || projectable ) {
+			// The projectable term here is present only to support distance projections.
+			// Since distances are derived from a DocValuesField, see DistanceCollector.
 			documentBuilder.addField( new LatLonDocValuesField( absoluteFieldPath, value.getLatitude(), value.getLongitude() ) );
+		}
+		else {
+			// For createExistsQuery()
+			documentBuilder.addFieldName( absoluteFieldPath );
 		}
 
 		if ( searchable ) {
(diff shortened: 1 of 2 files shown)


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneGeoPointFieldCodec.java
```diff
@@ -54,9 +54,13 @@ public void encode(LuceneDocumentBuilder documentBuilder, String absoluteFieldPa
 			documentBuilder.addField( new StoredField( getLongitudeAbsoluteFieldPath( absoluteFieldPath ), value.getLongitude() ) );
 		}
 
-		if ( searchable || sortable ) {
-			// doc values fields are required for predicates, distance projections and distance sorts
+		if ( sortable || searchable ) {
+			// Generally the LatLonDocValuesField field is required only for sorting,
+			// but we rely on this field to support exists predicate too.
```
yrodiere: To address the `exists` predicate, see how we do it for numeric fields.

In particular, when encoding:

```
		if ( sortable ) {
			documentBuilder.addField( domain.createDocValuesField( absoluteFieldPath, encodedValue ) );
		}
		else {
			// For createExistsQuery()
			documentBuilder.addFieldName( absoluteFieldPath );
		}
```

and when creating the `exists` query:

```
		if ( sortable ) {
			return new DocValuesFieldExistsQuery( absoluteFieldPath );
		}
		else {
			return new TermQuery( new Term( LuceneFields.fieldNamesFieldName(), absoluteFieldPath ) );
		}
```

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/spatial/DistanceSearchSearchableSortableIT.java
```diff
@@ -98,6 +98,31 @@ public void searchableNotSortable() {
 		assertThat( query ).hasDocRefHitsAnyOrder( INDEX_NAME, CHEZ_MARGOTTE_ID, IMOUTO_ID );
 	}
 
+	@Test
+	public void searchableDefSortable() {
```
yrodiere: "def"? Does it mean "default" or "definition"? Maybe avoid that kind of shortcut?

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/spatial/DistanceSearchSearchableSortableIT.java
```diff
@@ -124,22 +149,39 @@ public void notSearchableSortable() {
 		assertThat( query ).hasDocRefHitsAnyOrder( INDEX_NAME, CHEZ_MARGOTTE_ID, IMOUTO_ID, OURSON_QUI_BOIT_ID );
 	}
 
+	@Test
+	public void defSearchableSortable() {
```
yrodiere: Same comment about "def".

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/spatial/DistanceSearchSearchableSortableIT.java
```diff
@@ -154,12 +196,16 @@ private void initData() {
 	protected static class IndexMapping {
 		final IndexFieldReference<GeoPoint> searchableSortable;
 		final IndexFieldReference<GeoPoint> searchableNotSortable;
+		final IndexFieldReference<GeoPoint> searchableDefSortable;
```
yrodiere: Same comment about "def".

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/spatial/DistanceSearchSearchableSortableIT.java
```diff
@@ -154,12 +196,16 @@ private void initData() {
...
 		final IndexFieldReference<GeoPoint> notSearchableSortable;
+		final IndexFieldReference<GeoPoint> defSearchableSortable;
```
yrodiere: Same comment about "def".

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneGeoPointFieldCodec.java
fax4ever: I think that it's a nice solution. I applied it.
I had only to add an `| projectable` since it seems that a `DocValuesField` is required by the our `DistanceCollector` to project a distance form the very point.

fax4ever:
Thank @yrodiere for review.
There's anything else I could do to improve it?

yrodiere:
Merged, thanks!

fax4ever:
Thanks for merge it.

PR #2020, opened 2019-06-20
HSEARCH-3614 Allow integrators to opt out of using MethodHandles completely

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3614

@gsmet This should address a potential problem when putting two `@GenericField` annotations on the same field/method. I suspect that, when Quarkus' static initialization runs in GraalVM, it currently fails because we use method handles. If it does not, well... I guess this is a nice cleanup anyway.

Most of the changeset is restructuring and testing. The most important part is commit "HSEARCH-3614 Take advantage of the ValueReadHandleFactory in the AnnotationHelper", where we make sure to only use method handles when appropriate.

I tried to test this correctly in Hibernate Search, but unfortunately I cannot easily check that method handles are not used. I tried Byteman, but it is not able to manipulate the bytecode of `java.lang` classes (unless we use a complicated setup) and thus cannot help when it comes to method handles.
I'll try a bit more and may send another PR with more restrictive tests.


yrodiere:
I addressed the relevant Sonar issues. Others are mostly false positives or unrelated to this PR.

yrodiere:
Rebased and merged. Thanks!

PR #2021, opened 2019-06-20
HSEARCH-3605 Use a consistent strategy to decide on the package hosting each annotation

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3605

Annotations are now organized this way:

* all annotations that users will put directly on their entity types and methods are in the `pojo.mapping.definition.annotation` package, like they used to, next to the `pojo.mapping.definition.programmatic` package which hosts similar entry points for the programmatic mapping.
* extractor-related annotations are in the `pojo.extractor.mapping.annotation` package, next to the `pojo.extractor.mapping.programmatic` package which hosts interfaces for similar features used in the programmatic mapping.
* bridge-related annotations are in the `pojo.bridge.mapping.annotation` package, next to the `pojo.bridge.mapping.programmatic` package which hosts interfaces for similar features used in the programmatic mapping.
* builtin bridge annotations such as `@GeoPointBridge` are in the `pojo.bridge.builtin.annotation` package, next to the `pojo.bridge.builtin.programmatic` package which hosts interfaces for similar features used in the programmatic mapping.


fax4ever:
@yrodiere after we've rebased it on top of current master +1 to merge!

yrodiere:
Merged, thanks!

PR #2022, opened 2019-06-20
HSEARCH-3610 Rename Search.getSearchSession(Session) to Search.session(Session)

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3610

This:

```
SearchResult<Book> result = Search.getSearchSession(em).search(Book.class)
        .predicate(f -> f.matchAll())
        .fetch(20, 0);
```

now looks a bit shorter, like this:

```
SearchResult<Book> result = Search.session(em).search(Book.class)
        .predicate(f -> f.matchAll())
        .fetch(20, 0);
```


gsmet:
Not totally convinced. I think I would prefer `getSession()` better. But it might be just me so feel free to ignore :).

fax4ever:
@yrodiere I like the new API, I think is more smart and compact. So +1 to merge from me as it is.

yrodiere:
Merged, thanks!

PR #2023, opened 2019-06-20
HSEARCH-3074 Avoid depending on internal Hibernate ORM types

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3074

### DIFF
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingEmbeddedIdIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingEmbeddedIdIT.java
@@ -1,0 +1,240 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.integrationtest.mapper.orm.massindexing;
+
+import static org.assertj.core.api.Fail.fail;
+
+import java.io.Serializable;
+import java.util.UUID;
+import javax.persistence.Embeddable;
+import javax.persistence.EmbeddedId;
+import javax.persistence.Entity;
+import javax.persistence.Table;
+
+import org.hibernate.SessionFactory;
+import org.hibernate.search.engine.backend.work.execution.DocumentCommitStrategy;
+import org.hibernate.search.engine.backend.work.execution.DocumentRefreshStrategy;
+import org.hibernate.search.mapper.orm.Search;
+import org.hibernate.search.mapper.orm.cfg.HibernateOrmAutomaticIndexingStrategyName;
+import org.hibernate.search.mapper.orm.cfg.HibernateOrmMapperSettings;
+import org.hibernate.search.mapper.orm.massindexing.MassIndexer;
+import org.hibernate.search.mapper.orm.session.SearchSession;
+import org.hibernate.search.mapper.pojo.bridge.IdentifierBridge;
+import org.hibernate.search.mapper.pojo.bridge.runtime.IdentifierBridgeFromDocumentIdentifierContext;
+import org.hibernate.search.mapper.pojo.bridge.runtime.IdentifierBridgeToDocumentIdentifierContext;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.DocumentId;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.GenericField;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.IdentifierBridgeRef;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.Indexed;
+import org.hibernate.search.util.impl.integrationtest.common.rule.BackendMock;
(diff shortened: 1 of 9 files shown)


integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingEmbeddedIdIT.java
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	@Entity
+	@Table(name = "book")
+	@Indexed(index = Book.INDEX)
+	public static class Book {
+		public static final String INDEX = "Book";
+
+		@EmbeddedId
+		private BookId id;
```
yrodiere: Could you maybe give another name to the ID property, such as "foobar"? I'm not sure using "id" in a query string will have Hibernate ORM auto-magically unrestand we're refering to the ID.

In particular, what if another property is named "id", but is not the entity id?

util/internal/integrationtest/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/orm/SimpleSessionFactoryBuilder.java
```diff
@@ -28,8 +28,8 @@
 	private final List<Consumer<MetadataSources>> metadataSourcesContributors = new ArrayList<>();
 	private final List<Consumer<SessionFactoryBuilder>> sessionFactoryBuilderContributors = new ArrayList<>();
 
-	public SimpleSessionFactoryBuilder setTcclLookupPrecedence(TcclLookupPrecedence tcclLookupPrecedence) {
-		return onBootstraServiceRegistryBuilder( builder -> builder.applyTcclLookupPrecedence( tcclLookupPrecedence ) );
+	public SimpleSessionFactoryBuilder setTcclLookupPrecedenceBefore() {
+		return onBootstraServiceRegistryBuilder( builder -> builder.applyClassLoader( new TcclLookupPrecedenceClassLoader() ) );
```
yrodiere: I'm really not sure about this.

You're implementing your own classloader just to change the TCCL precedence in Hibernate ORM, which is a feature that is already exposed through SPIs through `org.hibernate.boot.registry.BootstrapServiceRegistryBuilder#applyTcclLookupPrecedence`. There's a fair amount of risk involved in deviating too much from the what our users' environment will be: our tests may end up not being representative.

Yes, `TcclLookupPrecedence` itself is marked as internal, but then it's exposed in `org.hibernate.boot.registry.BootstrapServiceRegistryBuilder` which is clearly an SPI. So in this case an implementation type is leaking through Hibernate ORM's SPI. The problem is in ORM, not in our code.

If you want to limit the usage of `TcclLookupPrecedence`, I'd advise to apply your changes up to this method, and give up here, calling `applyTcclLookupPrecedence( TcclLookupPrecedence.BEFORE )` directly.

If Hibernate ORM breaks this SPI, well, we can always fall back to your solution. It's not a big deal, as this is test code and a compatibility break will not affect our users.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingEmbeddedIdIT.java
fax4ever: Good idea!

fax4ever: You're right! If I change the name of the Hibernate ORM identifier, it works. But if I add a non-identifier field having `id` as a name, it doesn't!

util/internal/integrationtest/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/orm/SimpleSessionFactoryBuilder.java
fax4ever: Is a SPI method that has an internal parameter a little strange? Anyway let's limit as much as possible the internal exposition.

fax4ever: As the comment says... Actually it was my first intention.

fax4ever:
@yrodiere thanks for the review. Anything else? Thanks

util/internal/integrationtest/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/orm/SimpleSessionFactoryBuilder.java
yrodiere: > Is a SPI method that has an internal parameter a little strange

Yes I totally agree it's strange. But for now we'll have to deal with it.

yrodiere:
@fax4ever Looks good! Maybe you could just add forbiddenapis rules to forbid using classes in packages `org.hibernate.internal.**` and `org.hibernate.**.internal.**`?
Just launch the build after your added the rules, to check it fails, then add the necessary exclusions in `pom.xml` (around line 1411).

fax4ever:
@yrodiere thanks! I'll added the commit with the rule for the forbiddenapis plugin. I'll wait the CI, then I'll merge it.

fax4ever:
Merged thanks!

PR #2024, opened 2019-06-20
HSEARCH-3603 Swap the names of BeanResolver and BeanProvider

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3603
update: // -> /

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/environment/bean/BeanResolver.java
+++ b/engine/src/main/java/org/hibernate/search/engine/environment/bean/BeanResolver.java
@@ -11,7 +11,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Function;
 
-import org.hibernate.search.engine.environment.bean.spi.BeanResolver;
+import org.hibernate.search.engine.environment.bean.spi.BeanProvider;
 import org.hibernate.search.util.common.SearchException;
 import org.hibernate.search.util.common.impl.Contracts;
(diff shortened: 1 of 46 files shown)


engine/src/main/java/org/hibernate/search/engine/environment/bean/BeanResolver.java
```diff
@@ -11,74 +11,77 @@
...
+ * {@link #resolve(Class) by their type},
+ * or {@link #resolve(Class, String) by their type and name},
+ * or {@link #resolveRole(Class) by their role}.
  * <p>
- * This interface may be used by any Hibernate Search module,
- * but should only be implemented by the Hibernate Search engine itself;
- * if you are looking for implementing your own bean resolver,
- * you should implement {@link BeanResolver} instead.
+ * This interface is API,
+ * but should only be implemented by Hibernate Search itself;
```
fax4ever: Then maybe we should move it from a spi to an impl package.

yrodiere: I was trying to say that even though users can *call* the methods of this interface, nobody should not try to *implement* this interface. If people want to customize the bean retrieval behavior, they should implement `BeanProvider` instead (as explained below).

yrodiere:
Rebased, waiting for CI

yrodiere:
Merged, thanks!

PR #2025, opened 2019-06-21
HSEARCH-3067 Remove the notion of explicit initialization of lazy collection/map/array in PojoRuntimeIntrospector

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3067

It turns out that notion is obsolete.

There are remnants in the legacy code in
`org.hibernate.search.spi.InstanceInitializer#initializeCollection`
for example, but all implementations are no-op as of
8bc0e23.

Judging from that commit's message, this initialization was only useful
when dealing detached collections, and "we don't have to" do it anymore.
It was an SPI, but as far as I can tell it wasn't useful in any know
integration, so let's remove it for now and restore it if someone asks
for it (nicely).


fax4ever:
Merged thanks!

PR #2026, opened 2019-06-21
HSEARCH-3604 Replace PojoReference with mapper-specific types

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3604

Based on #2021, which should be merged first.

### DIFF
--- a/mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/bridge/mapping/impl/IdentifierMappingImplementor.java
+++ b/mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/bridge/mapping/impl/IdentifierMappingImplementor.java
@@ -8,14 +8,14 @@ package org.hibernate.search.mapper.pojo.bridge.mapping.impl;
 
 import java.util.function.Supplier;
 
+import org.hibernate.search.mapper.pojo.bridge.mapping.spi.IdentifierMapping;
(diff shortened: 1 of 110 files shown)


mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/bridge/mapping/impl/IdentifierMappingImplementor.java
```diff
@@ -4,18 +4,19 @@
...
+import org.hibernate.search.mapper.pojo.bridge.mapping.spi.IdentifierMapping;
 import org.hibernate.search.mapper.pojo.mapping.context.spi.AbstractPojoMappingContextImplementor;
 import org.hibernate.search.mapper.pojo.session.context.spi.AbstractPojoSessionContextImplementor;
 
 /**
  * @param <I> The identifier type for the mapped entity type.
  * @param <E> The entity type mapped to an index.
  */
-public interface IdentifierMapping<I, E> extends AutoCloseable {
+public interface IdentifierMappingImplementor<I, E> extends IdentifierMapping, AutoCloseable {
```
fax4ever: Extending `IdentifierMapping', I think we can remove `toDocumentIdentifier` method from here.

yrodiere:
Merged, thanks!

PR #2027, opened 2019-06-21
HSEARCH-2473 Get test naming practices straight

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-2473

I tested the new JQAssistant rules locally by adding tests with the wrong suffix: it worked.


fax4ever:
Merged thanks!

PR #2028, opened 2019-06-25
HSEARCH-3602 Use the "Step" suffix instead of the "Context" suffix in DSLs

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3602

That's a big, dumb refactoring, so I won't ask for a review. I'll see what Sonar has to say and merge.


yrodiere:
Sonar is happy. Merging.

PR #2034, opened 2019-07-04
HSEARCH-3609 Replace bridge builders with *binders*

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3609

~Based on #2038 , which should be merged first.~ => Done

This is an attempt at making the bridge 2.0 API simpler for simple use cases, and not-too-complicated-yet-powerful for advanced use cases.

Due to the refactorings involved, this will not be easy to review, so I'll describe the result here instead. If that doesn't seem too bad to you, I'll merge. Then I'll work on documenting this... soon-ish...

## Simple use case: `ValueBridge` only

To bind a single property to a single field, a custom `ValueBridge` can be used. It is similar to JPA's `AttributeConverter`.

Define the bridge class:

```java
// Assuming "SKU" is a user-defined type
public static class SKUBridge implements ValueBridge<SKU, String> {
	@Override
	public String toIndexedValue(SKU value, ValueBridgeToIndexedValueContext context) {
		return value == null ? null : value.toString();
	}

	// Implementing this is optional (the default impl throws an exception), but useful if you need projections
	@Override
	public SKU fromIndexedValue(String value, ValueBridgeFromIndexedValueContext context) {
		return value == null ? null : SKU.fromString( value );
	}
}
```

Use it in the mapping:

```java
@Entity
@Indexed
public class Product {

    // ...

    @Basic 
    @GenericField(valueBridge = @ValueBridgeRef(type = SKUBridge.class))
    private SKU sku;

    // ...
}
```

And that's it! Hibernate Search will automatically detect that the index field is of type String, and you'll be able to query the "sku" field as a text field.

## More advanced use case: `ValueBridge` with a `ValueBinder`

The `ValueBinder` allows more fine-grained control over the index field type.

For example, if we wanted the sku fields to be normalized, we could do it that way:

```java
@Entity
@Indexed
public class Product {

    // ...

    @Basic 
    @KeywordField(valueBridge = @ValueBridgeRef(type = SKUBridge.class), normalizer = "sku")
    private SKU sku;

    // ...
}
```

... but that's a bit redundant.

Another way of doing this would be to define a binder that applies the normalizer automatically.

Define the bridge and binder classes:

```java
// Assuming "SKU" is a user-defined type
public static class SKUBridge implements ValueBridge<SKU, String> {
	@Override
	public String toIndexedValue(SKU value, ValueBridgeToIndexedValueContext context) {
		return value == null ? null : value.toString();
	}

	// Implementing this is optional (the default impl throws an exception), but useful if you need projections
	@Override
	public SKU fromIndexedValue(String value, ValueBridgeFromIndexedValueContext context) {
		return value == null ? null : SKU.fromString( value );
	}

	public static class Binder implements ValueBinder {
		public void bind(ValueBindingContext context) {
			context.setBridge(
				SKU.class, new SKUBridge(),
				context.getTypeFactory().asString().normalizer( "sku" )
			);
		}
	}
}
```

Use the binder in the mapping:

```java
@Entity
@Indexed
public class Product {

    // ...

    @Basic 
    @GenericField(valueBinder = @ValueBinderRef(type = SKUBridge.Binder.class))
    private SKU sku;

    // ...
}
```

Note that ValueBinder is a bit more powerful than that: it also allows inspecting the input value type, which could allow the implementation of binders that create a different bridge depending on the input type. That's what we do for enums, in particular.

## The bazooka: `TypeBridge` and `PropertyBridge`

If you need to bind multiple properties and combine them into a single field value, or to bind one or more properties to multiple fields in the same bridge, `ValueBridge` cannot help: one of its trade-offs is that it only allows one-to-one binding.

`TypeBridge` and `PropertyBridge` are here just for that. They are a bit more complex to set up, but far more powerful.

Note they allow things that were not possible in Search 5:

* These bridges can define precisely the options of every index field they populate, including in particular analyzers and normalizers. In Search 5, you had to use a default analyzer for all fields except the "main" one (the one whose name was defined in the `@Field` annotation).
* These bridges can access properties of the bridged element, yet preserve consistent reindexing, thanks to dependency declaration

The first step to define such a bridge is to define an annotation:

```java
@PropertyBinding(binder = @PropertyBinderRef( type = LocalizedTextBridge.Binder.class ) )
@Retention(RetentionPolicy.RUNTIME)
@Target({ ElementType.FIELD, ElementType.METHOD })
@Documented
public @interface LocalizedTextBinding {
    String analyzerEn() default "text-en";
    String normalizerEn() default "text-en-sort";
    String analyzerFr() default "text-fr";
    String normalizerFr() default "text-fr-sort";
    String analyzerCrossLanguage() default "text-cross-language";
}
```

Then the bridge and binder classes:

```java
public class LocalizedTextBridge implements PropertyBridge {

	private final IndexFieldReference<String> enFieldReference;
	private final IndexFieldReference<String> enSortFieldReference;
	private final IndexFieldReference<String> frFieldReference;
	private final IndexFieldReference<String> frSortFieldReference;
	private final IndexFieldReference<String> crossLanguageFieldReference;

	private LocalizedTextBridge(IndexSchemaElement schema, String fieldName,
			LocalizedTextBinding annotation) {
		this.enFieldReference = schema.field(
				fieldName + "_en",
				f -> f.asString().analyzer( annotation.analyzerEn() )
		).toReference();
		this.enSortFieldReference = schema.field(
				fieldName + "_en_sort",
				f -> f.asString().normalizer( annotation.normalizerEn() )
		).toReference();
		this.frFieldReference = schema.field(
				fieldName + "_fr",
				f -> f.asString().analyzer( annotation.analyzerFr() )
		).toReference();
		this.frSortFieldReference = schema.field(
				fieldName + "_fr_sort",
				f -> f.asString().normalizer( annotation.normalizerFr() )
		).toReference();
		// An extra field for cross-language search.
		// We want a different analyzer in that case:
		// neither the French one nor the English one will do.
		this.crossLanguageFieldReference = schema.field(
				fieldName,
				f -> f.asString().analyzer( annotation.analyzerCrossLanguage() )
		).multiValued().toReference();
	}

	@Override
	public void write(DocumentElement target, Object bridgedElement, TypeBridgeWriteContext context) {
		if ( bridgedElement == null ) {
			return;
		}

		LocalizedText text = (LocalizedText) bridgedElement;

		target.addValue( enFieldReference, text.getEn() );
		target.addValue( enSortFieldReference, text.getEn() );
		target.addValue( crossLanguageFieldReference, text.getEn() );
		target.addValue( frFieldReference, text.getFr() );
		target.addValue( frSortFieldReference, text.getFr() );
		target.addValue( crossLanguageFieldReference, text.getFr() );
	}

	public static class Binder implements PropertyBinder<LocalizedTextBinding> {
		private LocalizedTextBinding annotation;

		@Override
		public void initialize(LocalizedTextBinding annotation) {
			this.annotation = annotation;
		}

		@Override
		public void bind(PropertyBindingContext context) {
			// By declaring the properties we rely on,
			// we make sure that changes to these properties will trigger reindexing
			context.getDependencies().use( "fr" ).use( "en" );

			context.setBridge(
					new LocalizedTextBridge(
							context.getBridgedElement().getName(),
							getIndexSchemaElement(),
							annotation
					)
			);
		}
	}
}
```

And finally use the annotation in the mapping:

```java
@Entity
@Indexed
public class Product {

    // ...

    @Embedded
    // Parameters can be passed to the binder through the annotation attributes
    @LocalizedTextBinding(analyzerCrossLanguage = "product-name-cross-language")
    private LocalizedText name;

    // ...
}
```


yrodiere:
@fax4ever I just pushed an additional commit to address a problem that @gsmet mentioned to me: the cast() method in Value and Identifier bridges was not necessary anymore, since after this PR the binders are expected to pass the expected type of values/identifiers to Hibernate Search, so we can perform the cast ourselves.
Anyway, it's fixed, and I updated the PR description to remove the weird-looking `cast` methods.

fax4ever:
I haven't reviewed the full code, I think that the new API is very powerful and it looks good to me. So +1 to merge from me.

yrodiere:
Thanks. I'll merge as soon as the Travis build succeeds.

PR #2035, opened 2019-07-08
HSEARCH-3622 - Backport HSEARCH-3462 to branch 5.11 - SimpleIndexingProgressMonitor stops logging as soon as the number of indexed document stops being a multiple of the logging period

yrodiere:
* [HSEARCH-3622](https://hibernate.atlassian.net//browse/HSEARCH-3622): Backport to 5.x: SimpleIndexingProgressMonitor stops logging as soon as the number of indexed document stops being a multiple of the logging period

It turns out the bug does not show up in 5.11, because, for a number of reasons, we always increment the number of indexed documents by one, and never more.

I will backport the fix to 5.11 regardless, in order to avoid future regressions, but I won't backport to 5.10 or earlier.


yrodiere:
Merged, thanks

PR #2036, opened 2019-07-08
HSEARCH-3621 Clarify the error message when the Elasticsearch cluster cannot be reached upon startup

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3621

~Based on #2038, which should be merged first.~ => Done

In short, this:

```
org.hibernate.search.util.common.SearchException: HSEARCH000520: Hibernate Search bootstrap failed. Failures:

    ORM mapping: 
        type 'com.acme.model.Type1': 
            index 'IndexName1': 
                failures: 
                  - HSEARCH400080: Failed to detect the Elasticsearch version running on the cluster.
        type 'com.acme.model.Type2': 
            index 'IndexName2': 
                failures: 
                  - HSEARCH400080: Failed to detect the Elasticsearch version running on the cluster.
	at org.hibernate.search.engine.reporting.impl.RootFailureCollector.checkNoFailure(RootFailureCollector.java:56)
	at org.hibernate.search.engine.common.impl.SearchIntegrationBuilderImpl.prepareBuild(SearchIntegrationBuilderImpl.java:170)
```

is now reported like this:

```
org.hibernate.search.util.common.SearchException: HSEARCH000520: Hibernate Search bootstrap failed. Failures:

    backend 'BackendName': 
        failures: 
          - HSEARCH400080: Failed to detect the Elasticsearch version running on the cluster: HSEARCH400007: Elasticsearch request failed: java.net.ConnectException: Connection refused
Request: GET  with parameters {}
Response: (no response)
	at org.hibernate.search.engine.reporting.impl.RootFailureCollector.checkNoFailure(RootFailureCollector.java:56)
	at org.hibernate.search.engine.common.impl.SearchIntegrationBuilderImpl.prepareBuild(SearchIntegrationBuilderImpl.java:172)
```

PR #2037, opened 2019-07-09
HSEARCH-3440 Finalize the lucene DirectoryProvider SPI

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3440

~Based on #2036 , which should be merged first.~ => Done


documentation/src/main/asciidoc/backend-lucene.asciidoc
```diff
@@ -14,14 +12,184 @@ but the defaults might not suit everyone.
...
+
+[NOTE]
+====
+Local filesystem directories really are designed to be *local* to one server and one application.
+
+In particular, they should not be shared between multiple Hibernate Search instances
+(example: network share).
+
+If you need to share indexes between multiple Hibernate Search instances,
+the Elasticsearch backend will be a better choice.
```
fax4ever: Is the Elasticsearch backend the only option to have a working cluster of Hibernate Search instances? What about Infinispan directory provider? Maybe we could add a `so far`...

documentation/src/main/asciidoc/backend-lucene.asciidoc
```diff
@@ -14,14 +12,184 @@ but the defaults might not suit everyone.
...
+In order to write to an index, Lucene needs to acquire a lock to ensure no other application instance
+writes to the same index concurrently.
+Each directory type comes with a default locking strategy that should work well enough in most situations.
+
+For those (very) rare situations where a different locking strategy is needed,
+Hibernate Search exposes a configuration property at the backend level:
+
+[source]
+----
+hibernate.search.backends.<backend-name>.directory.locking.strategy = /path/to/my/root
```
fax4ever: Does `/path/to/my/root` make sense here?

documentation/src/main/asciidoc/backend-lucene.asciidoc
```diff
@@ -14,14 +12,184 @@ but the defaults might not suit everyone.
...
+----
 
-// TODO HSEARCH-3440 document this
+The following strategies are available:
+
+* `simple-filesystem`:
+Locks the index by creating a marker file and checking it before write operations.
+This implementation is very simple and based Java's File API.
+If for some reason an application ends abruptly,
+the marker file will stay on the filesystem and will need to be removed manually.
```
fax4ever: maybe we can point out that this option is available only for local-filesystem directory type

documentation/src/main/asciidoc/backend-lucene.asciidoc
```diff
@@ -14,14 +12,184 @@ but the defaults might not suit everyone.
...
++
+See `org.apache.lucene.store.SimpleFSLockFactory`.
+* `native-filesystem`:
+Similarly to `simple-filesystem`, locks the index by creating a marker file,
+but using native OS file locks instead of Java's File API,
+so that locks will be cleaned up if the application ends abruptly.
++
+This implementation has known problems with NFS: it should be avoided on network shares.
```
fax4ever: Didn't we say that network file systems are to be avoided in any cases?

documentation/src/main/asciidoc/backend-lucene.asciidoc
```diff
@@ -14,14 +12,184 @@ but the defaults might not suit everyone.
...
++
+This is the default strategy for the `local-filesystem` directory type.
```
fax4ever: As above. Maybe we can point out that this option is available only for local-filesystem directory type

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/lowlevel/directory/LuceneLocalFileSystemDirectoryIT.java
```diff
@@ -0,0 +1,149 @@
...
+		assertThat( indexDirectory )
+				.isDirectory();
+
+		long contentSizeBeforeIndexing = directorySize( indexDirectory );
+
+		checkIndexingAndQuerying();
+
+		long contentSizeAfterIndexing = directorySize( indexDirectory );
+
+		assertThat( contentSizeAfterIndexing )
```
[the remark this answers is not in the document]
yrodiere: I'm not sure it's useful, and besides, the index is not guaranteed to decrease in size when we delete documents. They may just be marked as deleted using a flag, and actually removed from disk when segments are next merged.
Let's not bother for now.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cfg/LuceneBackendSettings.java
```diff
@@ -40,11 +42,55 @@ private LuceneBackendSettings() {
...
+	 * See the reference documentation for a list of available values.
+	 * <p>
+	 * Defaults to {@link Defaults#DIRECTORY_TYPE}.
+	 */
+	public static final String DIRECTORY_TYPE = DIRECTORY_PREFIX + DirectoryRadicals.TYPE;
+
+	/**
+	 * The filesystem root for directories.
+	 * <p>
+	 * Only available for the "local-filesystem" directory type.
```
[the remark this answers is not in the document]
yrodiere: The full key would become `hibernate.search.backend.myBackend.directory.local_filesystem.root`.
I've already had complaints that property keys are too long.

Also, we've decided to go another route in other places. Backends, for example: even though the "directory" only makes sense for Lucene, we don't use `hibernate.search.backend.myBackend.*lucene.*directory`.

The reasoning is that, if there's no risk of ambiguity, we may as well keep the property key as short as possible.

I updated the description of https://hibernate.atlassian.net/browse/HSEARCH-859, so we can discuss it if you feel strongly about it.

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/lowlevel/directory/AbstractDirectoryIT.java
```diff
@@ -0,0 +1,100 @@
...
+	public final TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+	@Rule
+	public final SearchSetupHelper setupHelper = new SearchSetupHelper();
+
+	protected SearchIntegration searchIntegration;
+	protected IndexMapping indexMapping;
+	protected StubMappingIndexManager indexManager;
+
+	protected abstract Object getDirectoryType();
```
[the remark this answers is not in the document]
yrodiere: Right, I just added one.

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/lowlevel/directory/LuceneLocalHeapDirectoryIT.java
```diff
@@ -0,0 +1,64 @@
...
+		LuceneIndexManagerImpl luceneIndexManager = indexManager.unwrapForTests( LuceneIndexManagerImpl.class );
+		assertThat( luceneIndexManager.getIndexAccessorForTests().getDirectoryForTests() )
+				.isInstanceOf( ByteBuffersDirectory.class );
+	}
+
+	@Override
+	@Test
+	@Ignore // Cannot use this locking strategy with a non-FS directory
+	public void lockingStrategy_nativeFilesystem() throws IOException {
+		super.lockingStrategy_nativeFilesystem();
```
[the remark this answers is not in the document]
yrodiere: We do... I just added tests.

documentation/src/main/asciidoc/backend-lucene.asciidoc
yrodiere: The Infinispan directory provider 1) does not exist for Search 6 yet and 2) is not enough to cluster Hibernate Search, since we also need to route reindexing events from slave nodes to the master node.

Support for clustering is not planned until 6.1, and even then I'm not sure we will address it for the Lucene backend, just the Elasticsearch one, at least for a start. We may extend that to Lucene if it's really easy, but we just don't have the manpower necessary to maintain our own clustered search engine solution.

Infinispan query is another story entirely, since they handle clustering themselves.

documentation/src/main/asciidoc/backend-lucene.asciidoc
yrodiere: It was a copy/paste error. Fixed.

documentation/src/main/asciidoc/backend-lucene.asciidoc
yrodiere: Right, that was clumsy. NFS can be used, but not to share data between multiple instances of Search. It could still be used to have Search run on one machine and the indexes be stored on another (though I doubt performance would be great).

I changed wording a bit to make this clearer.

yrodiere:
Thanks, I addressed most of your comments and rebased. Waiting for CI, then I'll merge.

PR #2043, opened 2019-07-18
Search 6 groundwork - Support (single-valued) projections on fields within nested object fields

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3062

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/impl/ElasticsearchIndexModel.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/impl/ElasticsearchIndexModel.java
@@ -75,15 +75,6 @@ public class ElasticsearchIndexModel {
 		return fieldNodes.get( absoluteFieldPath );
 	}
 
-	public String getNestedDocumentPath(String absoluteFieldPath) {
-		ElasticsearchIndexSchemaFieldNode<?> fieldNode = fieldNodes.get( absoluteFieldPath );
-		if ( fieldNode == null ) {
-			return null;
-		}
-
-		return fieldNode.getNestedPath();
-	}
-
 	@Override
 	public String toString() {
 		return new StringBuilder( getClass().getSimpleName() )
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/impl/ElasticsearchIndexModel.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/impl/ElasticsearchIndexModel.java
@@ -75,6 +75,15 @@ public class ElasticsearchIndexModel {
 		return fieldNodes.get( absoluteFieldPath );
 	}
 
+	public String getNestedDocumentPath(String absoluteFieldPath) {
+		ElasticsearchIndexSchemaFieldNode<?> fieldNode = fieldNodes.get( absoluteFieldPath );
+		if ( fieldNode == null ) {
+			return null;
+		}
+
+		return fieldNode.getNestedPath();
+	}
+
 	@Override
 	public String toString() {
 		return new StringBuilder( getClass().getSimpleName() )
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
@@ -7,7 +7,6 @@
 package org.hibernate.search.backend.elasticsearch.scope.model.impl;
 
 import java.lang.invoke.MethodHandles;
-import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.Objects;
@@ -156,7 +155,7 @@ public class ElasticsearchScopeModel {
 		}
 	}
 
-	public Set<String> getNestedDocumentPaths(String absoluteFieldPath) {
+	public String getNestedDocumentPath(String absoluteFieldPath) {
 		Optional<String> nestedDocumentPath = indexModels.stream()
 				.map( indexModel -> indexModel.getFieldNode( absoluteFieldPath ) )
 				.filter( Objects::nonNull )
@@ -171,6 +170,6 @@ public class ElasticsearchScopeModel {
 				} )
 				.orElse( Optional.empty() );
 
-		return ( nestedDocumentPath.isPresent() ) ? Collections.singleton( nestedDocumentPath.get() ) : Collections.emptySet();
+		return nestedDocumentPath.orElse( null );
 	}
 }
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
@@ -7,9 +7,11 @@
 package org.hibernate.search.backend.elasticsearch.scope.model.impl;
 
 import java.lang.invoke.MethodHandles;
-import java.util.HashSet;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -155,13 +157,20 @@ public class ElasticsearchScopeModel {
 	}
 
 	public Set<String> getNestedDocumentPaths(String absoluteFieldPath) {
-		HashSet<String> nestedDocumentPaths = new HashSet<>();
-		for ( ElasticsearchIndexModel indexModel : indexModels ) {
-			String nestedFieldPath = indexModel.getNestedDocumentPath( absoluteFieldPath );
-			if ( nestedFieldPath != null ) {
-				nestedDocumentPaths.add( nestedFieldPath );
-			}
-		}
-		return nestedDocumentPaths;
+		Optional<String> nestedDocumentPath = indexModels.stream()
+				.map( indexModel -> indexModel.getFieldNode( absoluteFieldPath ) )
+				.filter( Objects::nonNull )
+				.map( fieldNode -> Optional.ofNullable( fieldNode.getNestedPath() ) )
+				.reduce( (nestedDocumentPath1, nestedDocumentPath2) -> {
+					if ( Objects.equals( nestedDocumentPath1, nestedDocumentPath2 ) ) {
+						return nestedDocumentPath1;
+					}
+
+					throw log.conflictingNestedDocumentPathsForProjection(
+							absoluteFieldPath, nestedDocumentPath1.orElse( null ), nestedDocumentPath2.orElse( null ), getIndexesEventContext() );
+				} )
+				.orElse( Optional.empty() );
+
+		return ( nestedDocumentPath.isPresent() ) ? Collections.singleton( nestedDocumentPath.get() ) : Collections.emptySet();
 	}
 }
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
@@ -7,6 +7,7 @@
 package org.hibernate.search.backend.elasticsearch.scope.model.impl;
 
 import java.lang.invoke.MethodHandles;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.Set;
@@ -151,5 +152,16 @@ public class ElasticsearchScopeModel {
 			}
 			throw log.unknownFieldForSearch( absoluteFieldPath, getIndexesEventContext() );
 		}
+	}
+
+	public Set<String> getNestedDocumentPaths(String absoluteFieldPath) {
+		HashSet<String> nestedDocumentPaths = new HashSet<>();
+		for ( ElasticsearchIndexModel indexModel : indexModels ) {
+			String nestedFieldPath = indexModel.getNestedDocumentPath( absoluteFieldPath );
+			if ( nestedFieldPath != null ) {
+				nestedDocumentPaths.add( nestedFieldPath );
+			}
+		}
+		return nestedDocumentPaths;
 	}
 }
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchDistanceToFieldProjection.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchDistanceToFieldProjection.java
@@ -13,11 +13,13 @@ import java.util.regex.Pattern;
 import org.hibernate.search.backend.elasticsearch.gson.impl.JsonAccessor;
 import org.hibernate.search.backend.elasticsearch.gson.impl.JsonArrayAccessor;
 import org.hibernate.search.backend.elasticsearch.gson.impl.JsonObjectAccessor;
+import org.hibernate.search.backend.elasticsearch.search.projection.util.impl.SloppyMath;
 import org.hibernate.search.engine.search.loading.spi.LoadingResult;
 import org.hibernate.search.engine.search.loading.spi.ProjectionHitMapper;
 import org.hibernate.search.engine.spatial.DistanceUnit;
 import org.hibernate.search.engine.spatial.GeoPoint;
 
+import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 
@@ -29,8 +31,25 @@ class ElasticsearchDistanceToFieldProjection implements ElasticsearchSearchProje
 
 	private static final Pattern NON_DIGITS_PATTERN = Pattern.compile( "\\D" );
 
+	private static final String DISTANCE_PROJECTION_SCRIPT =
+		// Use ".size() != 0" to check whether this field has a value. ".value != null" won't work on ES7+
+		" Object result;" +
+		" if (doc[params.fieldPath].size() != 0) {" +
+			" result = doc[params.fieldPath].arcDistance(params.lat, params.lon);" +
+		" } else {" +
+			" for (int i=0; i<params.nestedPaths.length; i++) {" +
+				" String nestedPath = params.nestedPaths[i];" +
+				" String relativeFieldPath = params.relativeFieldPaths[i];" +
+				" if (params['_source'][nestedPath] == null) continue;" +
+				" if (params['_source'][nestedPath][relativeFieldPath] == null) continue;" +
+				" return params['_source'][nestedPath][relativeFieldPath]" +
+			" }" +
+		" }" +
+		" return result;";
+
 	private final Set<String> indexNames;
 	private final String absoluteFieldPath;
+	private final Set<String> nestedPaths;
 
 	private final GeoPoint center;
 
@@ -38,9 +57,10 @@ class ElasticsearchDistanceToFieldProjection implements ElasticsearchSearchProje
 
 	private final String scriptFieldName;
 
-	ElasticsearchDistanceToFieldProjection(Set<String> indexNames, String absoluteFieldPath, GeoPoint center, DistanceUnit unit) {
+	ElasticsearchDistanceToFieldProjection(Set<String> indexNames, String absoluteFieldPath, Set<String> nestedPaths, GeoPoint center, DistanceUnit unit) {
 		this.indexNames = indexNames;
 		this.absoluteFieldPath = absoluteFieldPath;
+		this.nestedPaths = nestedPaths;
 		this.center = center;
 		this.unit = unit;
 		this.scriptFieldName = createScriptFieldName( absoluteFieldPath, center, unit );
@@ -53,7 +73,7 @@ class ElasticsearchDistanceToFieldProjection implements ElasticsearchSearchProje
 			SCRIPT_FIELDS_ACCESSOR
 					.property( scriptFieldName ).asObject()
 					.property( "script" ).asObject()
-					.set( requestBody, createScript( absoluteFieldPath, center ) );
+					.set( requestBody, createScript( absoluteFieldPath, nestedPaths, center ) );
 		}
 	}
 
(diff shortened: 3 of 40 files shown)


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/scope/model/impl/LuceneScopeModel.java
```diff
@@ -148,4 +148,13 @@ public void checkNestedField(String absoluteFieldPath) {
 			throw log.unknownFieldForSearch( absoluteFieldPath, getIndexesEventContext() );
 		}
 	}
+
+	public boolean isNestedField(String absoluteFieldPath) {
```
yrodiere: Why didn't you simply include this information in the object returned by `getSchemaNodeComponent`? Since we already call that method every time we need this information.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchResultExtractorImpl.java
```diff
@@ -70,13 +79,17 @@
 		}
 
 		List<Object> extractedData = new ArrayList<>( topDocs.scoreDocs.length );
+		Map<Integer, Set<Integer>> nestedDocs = fetchNestedDocs( indexSearcher, topDocs.scoreDocs );
```
yrodiere: Good idea.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchResultExtractorImpl.java
```diff
@@ -88,23 +101,57 @@
...
-		TermQuery query = new TermQuery( new Term( LuceneFields.rootIdFieldName(), id ) );
+	private Map<String, Set<Integer>> fetchChildren(IndexSearcher indexSearcher, Set<String> parentIds) {
+		BooleanQuery.Builder builder = new BooleanQuery.Builder();
+		for ( String parentId : parentIds ) {
+			TermQuery query = new TermQuery( new Term( LuceneFields.rootIdFieldName(), parentId ) );
+			builder.add( query, BooleanClause.Occur.SHOULD );
+		}
+		builder.setMinimumNumberShouldMatch( 1 );
+
+		BooleanQuery booleanQuery = builder.build();
```
yrodiere: If you're going to use a query, to fetch children, you'd better use `ToChildBlockJoinQuery`, which might be more optimized. This seems to be an analog to what we already use for `nested` predicates.

We could do something like this:

```
new ToChildBlockJoinQuery(
    new MatchAllDocsQuery(),
    new QueryBitSetProducer( luceneQuery ),
    ScoreMode.None
)
```

Where `luceneQuery` is the query that was used to retrieve the parents.

As you can see, it requires to pass a `BitSetProducer` to specify the parents we're interested in, and the only available implementation is based on a query (`QueryBitSetProducer`), so we use that.
But in our case, we just executed a query to retrieve the parents, so it's a bit ridiculous.

We should think about it... I'm sure there's a better way. Maybe we could turn the `TopDocs` into a `BitSet`, or collect a `DocIdSet` in `org.hibernate.search.backend.lucene.search.extraction.impl.LuceneCollectors#collect` when we know we will need to load data from nested documents.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchResultExtractorImpl.java
```diff
@@ -88,23 +101,57 @@
...
+		Map<String, Set<Integer>> result = new HashMap<>();
```
yrodiere: On a related note, it's a shame that we are collecting *all* children, whereas we really only need *some* of them. Namely, we need the nested documents that hold the data of the fields we're projecting on.

There could be many "nested object fields" in a single document, with many different paths: "authors", "categories", "authors.publishers", ... If we only project on "authors.name", why should we also fetch "categories" and "authors.publishers"?

By switching to `ToChildBlockJoinQuery` as mentioned above, we could use a query to filter child documents by their path : something like `new TermQuery( new Term( LuceneFields.nestedDocumentPathFieldName(), "<the path of the nested document we're interested in" )`. Which could potentially reduce the number of fetched children in many cases.

Of course, that would require to know the paths we're interested in. But that shouldn't be too hard: currently we store the information "do we need nested documents", as a boolean; we only have to turn this into a set of paths we're interested in. These paths could be passed through `org.hibernate.search.backend.lucene.search.projection.impl.LuceneSearchProjection#contributeFields`, I suspect.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchResultExtractorImpl.java
```diff
@@ -88,23 +101,57 @@
...
 
 		try {
-			return indexSearcher.search( query, MAX_CHILDREN ).scoreDocs;
+			TopDocs topDocs = indexSearcher.search( booleanQuery, MAX_CHILDREN );
```
yrodiere: You don't really need topDocs here, which involves sorting by relevance. `indexSearcher.search( booleanQuery, <some collector>`, where the collector only collects the document IDs in a `DocIdSet`, would probably be better.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchResultExtractorImpl.java
fax4ever: cool

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneChildrenCollector.java
```diff
@@ -0,0 +1,75 @@
...
+		}
+
+		@Override
+		public void setScorer(Scorable scorer) throws IOException {
+			// we don't need any scorer
+		}
+
+		@Override
+		public void collect(int doc) throws IOException {
+			// FIXME tried to use BinaryDocValues without success. The field isn't nether binary nor numeric.
```
yrodiere: It's a good idea to use doc values here. Currently we don't store doc values for this field, which is why BinaryDocValues didn't work, but we definitely should.

In `org.hibernate.search.backend.lucene.document.impl.LuceneNestedObjectDocumentBuilder#contribute`, you will have to use something else than a `StringField` for the "rootId" field.

You can just use `Field` instead of `StringField`, and pass your own field type, defined as a private static constant. See in `StringField` how field types are constructed.

Then doc values should work as expected. See `org.hibernate.search.backend.lucene.search.extraction.impl.DistanceCollector.DistanceLeafCollector` to get an idea of how doc values are used (it's not the same type of doc values, but usage should be similar).

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchResultExtractorImpl.java
```diff
@@ -71,4 +95,58 @@
...
+
+	private Map<Integer, Set<Integer>> fetchNestedDocs(IndexSearcher indexSearcher, ScoreDoc[] scoreDocs) throws IOException {
+		// if the projection does not need any nested object skip their fetching
+		if ( storedFieldVisitor.getNestedDocumentPaths().isEmpty() ) {
+			return new HashMap<>();
+		}
+
+		Map<String, Integer> parentIds = new HashMap<>();
+		for ( ScoreDoc hit : scoreDocs ) {
+			Document doc = indexSearcher.doc( hit.doc, ID_FIELD_SET );
```
yrodiere: This could be avoided if we simply added a collector to the initial search that would collect the id using doc values.
Now that I think about it, it could be a great improvement in general: if we could collect IDs during the query execution rather than after it, we may gain some performance, especially for large indexes.
No need to address this now; I created https://hibernate.atlassian.net/browse/HSEARCH-3657

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneFieldCodec.java
```diff
@@ -39,8 +40,8 @@
 	 * @param absoluteFieldPath The path of the field whose value is assumed to be stored.
 	 * @param collector A collector of absolute field paths to be extracted from the document.
 	 */
-	default void contributeStoredFields(String absoluteFieldPath, Consumer<String> collector) {
-		collector.accept( absoluteFieldPath );
+	default void contributeStoredFields(String absoluteFieldPath, Set<String> nestedDocumentPaths, BiConsumer<String, Set<String>> collector) {
```
yrodiere: I'm not sure why you pass a `nestedDocumentPaths` here and expected the callback to pass it back to the collector.
Wouldn't it be simpler to just not mention the `nestedDocumentPaths` here, and expect the called to add them to the field visitor builder through a dedicated, separate method (e.g. `addNestedDocumentPaths`)?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/scope/model/impl/LuceneScopeModel.java
```diff
@@ -148,4 +149,15 @@ public void checkNestedField(String absoluteFieldPath) {
 			throw log.unknownFieldForSearch( absoluteFieldPath, getIndexesEventContext() );
 		}
 	}
+
+	public Set<String> getNestedDocumentPaths(String absoluteFieldPath) {
+		HashSet<String> nestedDocumentPaths = new HashSet<>();
```
yrodiere: Why did you allow multiple document paths? That would mean that some targeted indexes have a different structure, and currently, for every projection, we throw an exception when the structure differs between indexes.

I'm really not sure it's a good idea to try and support that kind of things immediately. It would be safer to just throw an exception.
Whatever you choose... please add a test?

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/projection/SearchProjectionIT.java
```diff
@@ -608,6 +609,7 @@ private void initData() {
 
 		final IndexObjectFieldReference nestedObject;
 		final FieldModel<String> nestedField;
+		final FieldModel<GeoPoint> nestedGeoPoint;
```
yrodiere: It looks like you're not doing anything with that new field?

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/search/spatial/LuceneNestedDocumentDistanceProjectionIT.java
```diff
@@ -0,0 +1,141 @@
...
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import org.assertj.core.api.Assertions;
+import org.assertj.core.data.Offset;
+
+/**
+ * Tests distance projetions on nested document.
```
yrodiere: `projetions` => `projections`

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/impl/ElasticsearchIndexModel.java
```diff
@@ -75,6 +75,15 @@ public ElasticsearchIndexSchemaObjectNode getObjectNode(String absolutePath) {
 		return fieldNodes.get( absoluteFieldPath );
 	}
 
+	public String getNestedDocumentPath(String absoluteFieldPath) {
```
yrodiere: Same comment as for Lucene: I don't think you need this method.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
```diff
@@ -152,4 +153,15 @@ public void checkNestedField(String absoluteFieldPath) {
 			throw log.unknownFieldForSearch( absoluteFieldPath, getIndexesEventContext() );
 		}
 	}
+
+	public Set<String> getNestedDocumentPaths(String absoluteFieldPath) {
```
yrodiere: Same comment as for the Lucene backend: I don't think you should support conflicting nested paths in multiple indexes.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchDistanceToFieldProjection.java
```diff
@@ -29,18 +31,36 @@
...
+		" Object result;" +
+		" if (doc[params.fieldPath].size() != 0) {" +
+			" result = doc[params.fieldPath].arcDistance(params.lat, params.lon);" +
+		" } else {" +
+			" for (int i=0; i<params.nestedPaths.length; i++) {" +
+				" String nestedPath = params.nestedPaths[i];" +
+				" String relativeFieldPath = params.relativeFieldPaths[i];" +
+				" if (params['_source'][nestedPath] == null) continue;" +
+				" if (params['_source'][nestedPath][relativeFieldPath] == null) continue;" +
+				" return params['_source'][nestedPath][relativeFieldPath]" +
```
yrodiere: Maybe add a comment here to explain why we have to go to such lengths?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/util/SloppyMath.java
```diff
@@ -0,0 +1,295 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.elasticsearch.search.projection.util;
```
yrodiere: Please put this class in an `impl` package.

And maybe you could remove all the dead code? We only use part of this class, I think.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/document/model/impl/LuceneIndexModel.java
```diff
@@ -68,6 +68,15 @@ public ScopedAnalyzer getScopedAnalyzer() {
 		return scopedAnalyzer;
 	}
 
+	public String getNestedDocumentPath(String absoluteFieldPath) {
```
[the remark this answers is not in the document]
fax4ever: Generally speaking, I would prefer always to delegate as much as possible to the server class, keeping the client call as simple as possible. So that the code could be spread among different classes. 

Following the principle of [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter).

But actually in this case we already expose the `LuceneIndexSchemaFieldNode` to the outside. So I'll put aside that principle :)

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/projection/SearchProjectionIT.java
```diff
@@ -317,6 +320,60 @@ public void mixed() {
...
+		SearchQuery<List<?>> query;
+
+		query = scope.query()
+				.asProjection( f ->
+						f.composite(
+								f.field( indexMapping.string1Field.relativeFieldName, String.class ),
+								f.documentReference(),
+								f.field( "nested." + indexMapping.nestedField.relativeFieldName, String.class ),
+								f.field( "nested.nested." + indexMapping.nestedNestedField.relativeFieldName, String.class ),
+								f.field( "nested.flattened." + indexMapping.flattenedField.relativeFieldName, String.class )
```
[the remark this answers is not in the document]
fax4ever: Probably I did get something. Sorry

> the direct parent could be stored as an object, but have a parent that is stored as nested

I interpreted it as:

> try to project a field that is contained in a not-nested object, which in turn is container in a nested object

Then I added a test case for the field `IndexMapping#flattenedField` on `SearchProjectionIT`.
This field is contained in an object that is not directly nested, but it has a nested ancestor.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneChildrenCollector.java
fax4ever: Very clear! Thank you for the explanation

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/document/model/impl/LuceneIndexModel.java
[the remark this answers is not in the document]
yrodiere: Yes, it's not relevant in this case.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/projection/SearchProjectionIT.java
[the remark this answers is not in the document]
yrodiere: You're right, sorry. the test on "nested.flattened.*" is what we need. I'll mark this as resolved.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchResultExtractorImpl.java
fax4ever: I will add a comment over here with the Jira key.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneFieldCodec.java
fax4ever: You're definitely right. Done!

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/scope/model/impl/LuceneScopeModel.java
fax4ever: Indeed I was not sure to allow that.
Let's raise an exception for.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/projection/SearchProjectionIT.java
fax4ever: That's right! I would have used it to test nested distance projection, but then I created `NestedDocumentDistanceProjectionIT.java` for that.

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/search/spatial/LuceneNestedDocumentDistanceProjectionIT.java
fax4ever: I think it has been already solved by a subsequent commit.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchDistanceToFieldProjection.java
fax4ever: I will

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/util/SloppyMath.java
fax4ever: Sure!

fax4ever:
@yrodiere thanks for the review.
I think I addressed the comments.

Moreover, I added these 2 extra commits:

> HSEARCH-3062 Make Lucene nested paths single valued
HSEARCH-3062 Make ES nested paths single valued

As a followup of:

> HSEARCH-3062 Check nested document paths field compatibility

It seems that once we have verified the incompatibilities we don't need a set anymore.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/projection/impl/LuceneFieldProjection.java
```diff
@@ -42,6 +44,7 @@ public void contributeCollectors(LuceneCollectorsBuilder luceneCollectorBuilder)
 	@Override
 	public void contributeFields(LuceneDocumentStoredFieldVisitorBuilder builder) {
 		codec.contributeStoredFields( absoluteFieldPath, builder::add );
+		codec.contributeNestedDocumentPaths( nestedDocumentPaths, builder::addNestedDocumentPaths );
```
yrodiere: `builder.addNestedDocumentPaths( nestedDocumentPaths )` would work just as well.


`codec.contributeStoredFields` exists because there's an actual reason for codec to contribute "low-level" fields, in particular in the case of geopoints. For nested document paths, codecs don't have anything to do, so there's no reason to get them involved. See the implementations of `contributeNestedDocumentPaths`: there's only one, and it's dead simple.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/document/impl/LuceneNestedObjectDocumentBuilder.java
```diff
@@ -28,6 +52,7 @@ void contribute(String rootIndexName, MultiTenancyStrategy multiTenancyStrategy,
 		document.add( new StringField( LuceneFields.typeFieldName(), LuceneFields.TYPE_CHILD_DOCUMENT, Store.YES ) );
 		document.add( new StringField( LuceneFields.rootIndexFieldName(), rootIndexName, Store.YES ) );
 		document.add( new StringField( LuceneFields.rootIdFieldName(), rootId, Store.YES ) );
+		document.add( new Field( LuceneFields.rootIdBinaryFieldName(), new BytesRef( rootId ), TYPE_STORED_BINARY ) );
```
yrodiere: You don't need to add a new field, you can just replace the one above with this one. Here you're storing two values in each document for the root ID field name, which is not useful.
Just replace this:

```
		document.add( new StringField( LuceneFields.rootIdFieldName(), rootId, Store.YES ) );
		document.add( new Field( LuceneFields.rootIdBinaryFieldName(), new BytesRef( rootId ), TYPE_STORED_BINARY ) ); 
```

... with this:

```
		document.add( new Field( LuceneFields.rootIdFieldName(), new BytesRef( rootId ), TYPE_STORED_BINARY ) ); 
```

And remove the method `LuceneFields.rootIdBinaryFieldName`.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/projection/impl/LuceneFieldProjection.java
fax4ever: makes sense, thanks!

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/document/impl/LuceneNestedObjectDocumentBuilder.java
fax4ever: I had figured out that it could have been replaced. But I wasn't sure about.

fax4ever:
Merged thanks!

PR #2044, opened 2019-07-19
HSEARCH-3642 Merge the DslConverter and ProjectionConverter enums into one

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3642


fax4ever:
Merged thanks!

PR #2045, opened 2019-07-24
HSEARCH-3643 Entities should be unproxied before getting field values when indexing

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3643


fax4ever:
Merged thanks!

PR #2046, opened 2019-07-25
HSEARCH-3644 ORM warning on bootstrap: HHH000450: Encountered request for Service by non-primary service role 

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3644

We used to set `hibernate.service.allow_crawling` to `false` in our integration tests, requiring that Hibernate ORM behaves in a very strict way when instantiating services. This was a good idea a long time ago when the way services are instantiated changed in Hibernate ORM, because it allowed us to fail the build if we somehow made a mistake in our (ORM) services declaration.

But very recently, we discovered that disabling this "crawling" also disables some warnings, which for some arcane reason have started appearing on every startup when Hibernate Search is enabled.

This PR:

* Removes the forced `allow_crawling` setting. As a result, Hibernate ORM is now more permissive, but will log a warning when we rely on lenient behavior.
* Adds a test checking that there are no warnings on startup, apart from those we expect considering the test setup.
* Fixes the problem in Hibernate Search that led to Hibernate ORM logging a warning when it retrieves Hibernate Search services. This commit is rather complex, but the core is rather simple: I just tried to avoid having service interfaces that extend each other. The complexity mostly comes from the changes that were required to avoid cyclic inter-package dependencies.


yrodiere:
@fax4ever This PR was related to a problem reported by @gsmet , but he didn't have time to review before his PTO. Would you mind reviewing this? I'll add some context to the description.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/spi/DifferentSessionFactoriesIT.java
```diff
@@ -28,7 +29,7 @@
 
 /**
  * Tests the creation of a {@link HibernateOrmSearchSession}
- * using a {@link HibernateOrmSearchSessionContextProvider}
+ * using a {@link HibernateOrmSearchSessionMappingContext}
```
fax4ever: or using a `HibernateSearchContextProviderService`?

yrodiere:

PR #2047, opened 2019-07-29
[5.11] HSEARCH-3647 Search result lost after Session.merge on a child entity when using Multi-Tenancy

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3647

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/spi/AbstractDocumentBuilder.java
+++ b/engine/src/main/java/org/hibernate/search/engine/spi/AbstractDocumentBuilder.java
@@ -152,13 +152,14 @@ public abstract class AbstractDocumentBuilder {
 	}
 
 	/**
-	 * If we have a work instance we have to check whether the instance to be indexed is contained in any other indexed entities.
-	 *
 	 * @see #appendContainedInWorkForInstance(Object, WorkPlan, ContainedInRecursionContext, String)
 	 * @param instance the instance to be indexed
 	 * @param workPlan the current work plan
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/hibernate/search/engine/spi/AbstractDocumentBuilder.java
```diff
@@ -152,12 +152,12 @@ public void postInitialize(IndexedTypeSet indexedClassesSet) {
...
-	 * If we have a work instance we have to check whether the instance to be indexed is contained in any other indexed entities.
-	 *
 	 * @see #appendContainedInWorkForInstance(Object, WorkPlan, ContainedInRecursionContext, String)
 	 * @param instance the instance to be indexed
 	 * @param workPlan the current work plan
 	 * @param currentRecursionContext the current {@link org.hibernate.search.engine.spi.ContainedInRecursionContext} object used to check the graph traversal
+	 * @deprecated Use {@link #appendContainedInWorkForInstance(Object, WorkPlan, ContainedInRecursionContext, String)}
+	 * instead, and pass the appropriate tenant ID.
 	 */
 	public void appendContainedInWorkForInstance(Object instance, WorkPlan workPlan, ContainedInRecursionContext currentRecursionContext) {
```
gsmet: Should we add this method to the ones we don't want the codebase to use?

yrodiere: Good idea, done.

yrodiere:
Merged, thanks

PR #2048, opened 2019-07-30
HSEARCH-3646 Upgrade to Lucene 8.2

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3646

I had a look at the release notes: no backward incompatible change that would affect us, some performance improvements. The upgrade is safe.


yrodiere:
Tests passed on CI; merging.

PR #2050, opened 2019-07-30
HSEARCH-3447 Replace uses of deprecated Hibernate criteria with JPA criteria

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3447

### DIFF
--- a/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/BatchIndexingWorkspace.java
+++ b/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/BatchIndexingWorkspace.java
@@ -6,13 +6,14 @@
  */
 package org.hibernate.search.mapper.orm.massindexing.impl;
 
-import java.io.Serializable;
 import java.lang.invoke.MethodHandles;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Future;
 import java.util.concurrent.ThreadPoolExecutor;
+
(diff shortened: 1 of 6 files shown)


mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/BatchIndexingWorkspace.java
```diff
@@ -29,19 +30,19 @@
  *
  * @author Sanne Grinovero
  */
-public class BatchIndexingWorkspace extends ErrorHandledRunnable {
+public class BatchIndexingWorkspace<E, I> extends ErrorHandledRunnable {
```
fax4ever: Very nice having these generics here. Even if their meaning should be very compressible, maybe we could depict them in the java doc. The same you did for `IdentifierConsumerDocumentProducer`.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/BatchIndexingWorkspace.java
```diff
@@ -73,11 +74,10 @@
 		this.sessionContext = sessionContext;
 
 		this.indexedType = type;
+		this.idAttributeOfIndexedType = idAttributeOfIndexedType;
 		this.idFetchSize = idFetchSize;
 		this.transactionTimeout = transactionTimeout;
 
-		this.idNameOfIndexedType = sessionFactory.getMetamodel().entity( indexedType ).locateIdAttribute().getName();
```
fax4ever: 2 empty lines here

yrodiere:

PR #2051, opened 2019-07-30
HSEARCH-3551 Hibernate Search doesn't boot when declared as a dependency in a module-info.java

yrodiere:
https://hibernate.atlassian.net//browse/HSEARCH-3551

Note that changes related to the `AggregatedClassLoader`/`ServiceResolver` are directly inspired by my other PR on ORM: https://github.com/hibernate/hibernate-orm/pull/2890 See that PR for a detailed explanation.

Most other changes are only to accommodate this `AggregatedClasLoader`/`ServiceResolver` change and to fix problems in tests.


engine/src/main/java/org/hibernate/search/engine/environment/classpath/spi/ServiceResolver.java
```diff
@@ -0,0 +1,27 @@
...
+	 * @param serviceContract The java type defining the service contract
+	 * @param <T> The type of the service contract
+	 *
+	 * @return The ordered set of discovered services.
+	 */
+	<T> Iterable<T> loadJavaServices(Class<T> serviceContract);
+}
+
+
+
```
fax4ever: I think that there are too many empty lines here.

engine/src/main/java/org/hibernate/search/engine/environment/classpath/spi/AggregatedClassLoader.java
```diff
@@ -22,7 +29,31 @@
...
+
+		// finally the system classloader
+		final ClassLoader sysClassLoader = locateSystemClassLoader();
+		if ( sysClassLoader != null ) {
+			orderedClassLoaderSet.add( sysClassLoader );
+		}
+
+		// now build the aggregated class loader...
+		return new AggregatedClassLoader(
+				orderedClassLoaderSet.toArray( new ClassLoader[0] )
```
fax4ever: Actually I'm not sure about, but maybe could have sense to pre-assign a size to the array.

engine/src/main/java/org/hibernate/search/engine/environment/classpath/spi/DefaultResourceResolver.java
```diff
@@ -0,0 +1,80 @@
...
+			}
+		}
+
+		return null;
+	}
+
+}
+
+
+
```
fax4ever: Same here. It seems there are too many empty lines.

engine/src/main/java/org/hibernate/search/engine/environment/classpath/spi/DefaultClassResolver.java
```diff
@@ -0,0 +1,47 @@
...
+		}
+		catch (Exception | LinkageError e) {
+			throw log.unableToLoadTheClass( className, e );
+		}
+	}
+
+}
+
+
+
```
fax4ever: same here?

engine/src/main/java/org/hibernate/search/engine/environment/classpath/spi/DefaultServiceResolver.java
```diff
@@ -0,0 +1,192 @@
...
+							services.add( provider.get() );
+						}
+					} );
+			return services;
+		}
+	}
+}
+
+
+
```
fax4ever: same here?

engine/src/main/java/org/hibernate/search/engine/environment/classpath/spi/AggregatedClassLoader.java
yrodiere: Passing `new ClassLoader[0]` is about as efficient as passing an array with the right size, at least it was last time I checked. I'll change that if it turns out there really is an efficiency problem :)

yrodiere:

PR #2052, opened 2019-08-01
HSEARCH-3612 Normalize the terms to match in wildcard/prefix predicates

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3612

Tested locally for all Elasticsearch supported profiles: the behavior seems consistent among all of them.

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/predicate/impl/LuceneTextFieldPredicateBuilderFactory.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/predicate/impl/LuceneTextFieldPredicateBuilderFactory.java
@@ -61,7 +61,7 @@ public final class LuceneTextFieldPredicateBuilderFactory<F>
 	public WildcardPredicateBuilder<LuceneSearchPredicateBuilder> createWildcardPredicateBuilder(
 			String absoluteFieldPath) {
 		checkSearchable( absoluteFieldPath );
-		return new LuceneTextWildcardPredicateBuilder( absoluteFieldPath );
+		return new LuceneTextWildcardPredicateBuilder( absoluteFieldPath, analyzerOrNormalizer );
 	}
 
 	@Override
--- a/documentation/src/main/asciidoc/search-dsl.asciidoc
+++ b/documentation/src/main/asciidoc/search-dsl.asciidoc
@@ -510,30 +510,27 @@ include::{sourcedir}/org/hibernate/search/documentation/searchdsl/predicate/Pred
 ----
 ====
 
-// TODO HSEARCH-3612 When the ticket is fixed, rephrase this warning to state that only ES 6.7 and below is affected by the normalization problem.
-
 [IMPORTANT]
 ====
-Patterns used in wildcard predicates are not analyzed or normalized,
-and they are expected to match a *single* indexed token, not a sequence of tokens.
+If a normalizer has been defined on the field, the patterns used in wildcard predicates
+will be normalized.
 
-For example, a pattern such as `Cat*` will not match anything
-when targeting a field that applies a lowercase filter when indexing.
-`cat*` may match, since it's already lowercased.
+If an analyzer has been defined on the field:
 
-Similarly, a pattern such as `john gr*` will not match anything
+* when using the Elasticsearch backend, the patterns won't be analyzed nor normalized,
+and will be expected to match a *single* indexed token, not a sequence of tokens.
+* when using the Lucene backend the patterns will be normalized, but not tokenized:
+the pattern will still be expected to match a *single* indexed token, not a sequence of tokens.
+
+For example, a pattern such as `Cat*` could match `cat`
+when targeting a field having a normalizer that applies a lowercase filter when indexing.
+
+A pattern such as `john gr*` will not match anything
 when targeting a field that tokenizes on spaces.
 `gr*` may match, since it doesn't include any space.
 
-These limitations generally make the wildcard predicate unadvisable for high-level,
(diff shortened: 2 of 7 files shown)


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/predicate/impl/LuceneTextFieldPredicateBuilderFactory.java
```diff
@@ -61,7 +61,7 @@ public LuceneTextFieldPredicateBuilderFactory( boolean searchable,
 	public WildcardPredicateBuilder<LuceneSearchPredicateBuilder> createWildcardPredicateBuilder(
 			String absoluteFieldPath) {
 		checkSearchable( absoluteFieldPath );
-		return new LuceneTextWildcardPredicateBuilder( absoluteFieldPath );
+		return new LuceneTextWildcardPredicateBuilder( absoluteFieldPath, analyzerOrNormalizer );
```
yrodiere: There's also a "prefix" syntax in the simple query string predicate. Could you please implement the fix and test it for that predicate, too?

documentation/src/main/asciidoc/search-dsl.asciidoc
```diff
@@ -462,30 +462,24 @@ include::{sourcedir}/org/hibernate/search/documentation/searchdsl/predicate/Pred
...
+
+If an analyzer has been defined on the field,
+using Elasticsearch backend the patterns won't be analyzed nor normalized,
+using Lucene backend the patters will be normalized, but not tokenized.
 
-For example, a pattern such as `Cat*` will not match anything
-when targeting a field that applies a lowercase filter when indexing.
-`cat*` may match, since it's already lowercased.
+For example, a pattern such as `Cat*` could match `cat`
+when targeting a field having a normalizer applies a lowercase filter when indexing.
```
yrodiere: `normalizer applies` => `normalizer that applies`

documentation/src/main/asciidoc/search-dsl.asciidoc
```diff
@@ -462,30 +462,24 @@ include::{sourcedir}/org/hibernate/search/documentation/searchdsl/predicate/Pred
...
 [IMPORTANT]
 ====
-Patterns used in wildcard predicates are not analyzed or normalized,
-and they are expected to match a *single* indexed token, not a sequence of tokens.
+If a normalizer has been defined on the field, the patterns used in wildcard predicates
+will be normalized.
+
+If an analyzer has been defined on the field,
+using Elasticsearch backend the patterns won't be analyzed nor normalized,
+using Lucene backend the patters will be normalized, but not tokenized.
```
yrodiere: `patters` => `patterns`

yrodiere: I would rephrase:

```
If an analyzer has been defined on the field:

* when using the Elasticsearch backend, the patterns won't be analyzed nor normalized,
and will be expected to match a *single* indexed token, not a sequence of tokens.
* when using the Lucene backend the patterns will be normalized, but not tokenized:
the pattern will still be expected to match a *single* indexed token, not a sequence of tokens.
```

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/predicate/impl/LuceneTextFieldPredicateBuilderFactory.java
fax4ever: Sorry I missed completely the `prefix` part. Adding now...

fax4ever:
@yrodiere According to the last test I added it seems that the terms are already normalized, even without the `analyze_wildcard` flag enabled. Addressed the comment so far

yrodiere:
@fax4ever To be honest, I didn't think prefix queries would perform tokenization at all. I expected them to only work with a single term, just like the wildcard queries. I was surprised that your tests passed...

fax4ever:
Indeed it doesn't work :) 

So let's just stick tot:
> Test normalization of prefix within simple query terms 

Thanks

yrodiere:
> Indeed it doesn't work :)

I don't get it, what doesn't work?

I was talking about prefix queries using multiple terms, which I didn't think would work, e.g. : `"a few words"*`. I only expected single-term prefix queries to work, e.g. `word*`.

fax4ever:
Me too. I had tried a tess, such as: 
> "prefix1 prefix2 prefix3"*.

And it didn't work. For that reason I settled for normalization only.

yrodiere:
But your test contains this:

```
		assertThat( createQuery.apply( "\"" + PREFIX_1_DIFFERENT_CASE + "\"*" ) )
				.hasDocRefHitsAnyOrder( INDEX_NAME, DOCUMENT_1, DOCUMENT_2, DOCUMENT_3, DOCUMENT_4 ); 
```

The query string evaluates to `"SmILE on oUr LOves, and while thou dRAWest thE"*`, and the targeted field is analyzed, so the field itself is tokenized.

If the prefix query didn't tokenize the input, it would try to find a term starting with `SmILE on oUr LOves, and while thou dRAWest thE`, and it would fail, because the only terms in the index are `smile`, `on`, `our`, `loves`, etc., not the complete sentence.

So it *does* tokenize input. Which is nice, I suppose?

I launched a full build, let's see how it goes, but if everything passes, +1 to merge. Thanks!

fax4ever:
@yrodiere yes It is. Thank you. 
Good idea to lunch a full build to test the different version of Elasticsearch.

fax4ever:
Merged thanks!

PR #2054, opened 2019-08-06
HSEARCH-3360 Correctly handle the flush() and clear() events in the ORM integration

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3360
https://hibernate.atlassian.net/browse/HSEARCH-1350


fax4ever:
It needs work because we need to: 

1. Add extensive tests for all possible types of flush. Such as: automatic flushes happening on a query, single-entity flush, etc.

2. Add a functional test for: [HSEARCH-1350](https://hibernate.atlassian.net/browse/HSEARCH-1350)

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/event/impl/HibernateSearchEventListener.java
```diff
@@ -145,6 +148,15 @@ public void onFlush(FlushEvent event) {
 //		}
 	}
 
+	@Override
+	public void onClear(ClearEvent event) {
+		EventSource session = event.getSession();
+		// skip the clearNotPrepared operation in case of rollback
```
yrodiere: Note for later: this is a piece of code that should be changed in another PR related to adding support for events outside of transactions.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/session/SearchSessionFlushIT.java
```diff
@@ -0,0 +1,115 @@
...
+
+	@Before
+	public void before() {
+		backendMock.expectAnySchema( IndexedEntity.INDEX_NAME );
+		sessionFactory = ormSetupHelper.start().setup( IndexedEntity.class );
+		backendMock.verifyExpectationsMet();
+	}
+
+	@Test
+	public void onManualFlush() {
```
yrodiere: This is a bit confusing, as it could mean `FlushMode.MANUAL`, but it doesn't. Maybe `onExplicitFlush`?

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/work/impl/PojoWorkPlanImpl.java
```diff
@@ -39,6 +39,8 @@
 	private final Map<Class<?>, PojoIndexedTypeWorkPlan<?, ?, ?>> indexedTypeDelegates = new LinkedHashMap<>();
 	private final Map<Class<?>, PojoContainedTypeWorkPlan<?>> containedTypeDelegates = new HashMap<>();
 
+	private volatile boolean isPreparing = false;
```
yrodiere: This class is not thread safe and should only be used in a single thread. You can remove the `volatile` keyword.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/work/impl/PojoWorkPlanImpl.java
```diff
@@ -108,14 +110,24 @@ public void purge(Class<?> clazz, Object providedId) {
 
 	@Override
 	public void prepare() {
-		for ( PojoContainedTypeWorkPlan<?> delegate : containedTypeDelegates.values() ) {
-			delegate.resolveDirty( this::updateBecauseOfContained );
+		if ( isPreparing ) {
```
yrodiere: This looks... odd. Can you tell me how to reproduce the issue you're working around? I'm concerned we may need to do more than just ignoring the call...

integrationtest/showcase/library/src/test/java/org/hibernate/search/integrationtest/showcase/library/functional/LessMemoryEntityService.java
```diff
@@ -0,0 +1,84 @@
...
+ * {@link org.hibernate.search.integrationtest.showcase.library.service.DocumentService}.
+ * <p>
+ * But it applies also a <b>flush</b> and a <b>clear</b> on {@link EntityManager} and an <b>evictAll</b> on {@link EntityManagerFactory#getCache()}
+ * before every commit.
+ * <p>
+ * This strategy can be useful for a batch processing to release memory.
+ */
+@Service
+@Transactional
+public class LessMemoryEntityService {
```
yrodiere: If we really want a functional IT, and if you really want to put it in the showcase, I think we need to be closer to the use case.

Unless I'm mistaken, the use case was batch inserts. So, something like a `BatchInsertService` or a `DatabaseInitializationService` a `batchInsert(Iterator<BookToInsert>)` method in `DocumentService` (`BookToInsert` being a bean created from an external source) would seem to make more sense.

The alternative, of course, would be to move this test to `org.hibernate.search.integrationtest.mapper.orm.nonregression.automaticindexing`. Then, you could just have all the flush/clear code in a single method, and the entity classes in the same `.java` file. It would probably be easier.

integrationtest/showcase/library/src/test/java/org/hibernate/search/integrationtest/showcase/library/functional/LibraryFunctionalIT.java
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.TestPropertySource;
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+@TestPropertySource(properties = { "automatic_indexing.strategy=session" })
+@ActiveProfiles(resolver = TestActiveProfilesResolver.class)
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
+public class LibraryFunctionalIT {
```
yrodiere: This seem to imply other tests are not functional, which is wrong. Maybe `LibraryBatchInsertIT`?

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/session/SearchSessionWritePlanBaseIT.java
```diff
@@ -210,6 +210,87 @@ public void earlyProcess() {
 		backend1Mock.verifyExpectationsMet();
 	}
 
+	@Test
+	public void sessionClear() {
```
yrodiere: Since you're not using the `SearchSessionWritePlan` API, I think this test is about automatic indexing, and should be moved to `AutomaticIndexingBasicIT`, or a new test class in the test package if you deem it necessary.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/session/SearchSessionFlushIT.java
```diff
@@ -0,0 +1,115 @@
...
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+/**
+ * Tests the behaviour of different kinds of {@link Session#flush()}.
+ * Each of ones has to trigger a {@link PojoWorkPlan#prepare()} exactly when the flush is expected.
+ */
+@TestForIssue( jiraKey = "HSEARCH-3360" )
+public class SearchSessionFlushIT {
```
yrodiere: Same here, I think you should move this to `AutomaticIndexingBasicIT` or move this class to the same package.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/work/impl/PojoWorkPlanImpl.java
fax4ever: You're right. Actually I was also skeptical about that.

Without the check, we get a `StackOverflow` running `AutomaticIndexingBridgeExplicitReindexingFunctionalIT.test`.

In the stack trace we can see an infinite loop of `HibernateSearchEventListener.onAutoFlush`. Where each of ones invokes a prepare on the same `PojoWorkPlanImpl` instance.

integrationtest/showcase/library/src/test/java/org/hibernate/search/integrationtest/showcase/library/functional/LessMemoryEntityService.java
fax4ever: I would opt the latter. But would it still a functional?

fax4ever:
Thanks for review @yrodiere. Changes are applied here.

To reproduce the recursion on `PojoWorkPlanImpl#prepare` run the test case `AutomaticIndexingBridgeExplicitReindexingFunctionalIT.test`.

The recursion is triggered by `HibernateSearchEventListener.onAutoFlush` method.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/work/impl/PojoWorkPlanImpl.java
yrodiere: So... I investigated and it turns out that, in most cases, this recursion shouldn't happen.

In this test, the recursion only happens because we execute a search query inside a bridge, and that search query triggers an automatic flush. But, since we flush just before we execute the bridges, there shouldn't be anything to flush.

So, I added a few commits:

1. Throw an exception when such recursion happens, because it's definitely not normal.
2. Disable automatic flushes while processing entities for indexing.
3. I'll get back to the third one.

Unfortunately, this is not enough. It turns out the `AutoFlushEvent` is fired even when no flush happens: it is actually used to tell the `DefaultAutoFlushEventListener` "please check whether an automatic flush is required, and if so, do it".

This means that our current behavior of considering a flush occurred every time we get an `AutoFlushEvent` is wrong: it just means a flush was requested, but it may not have occurred.

So we need to check whether the flush actually occurred. I added a third commit that tries to leverage `AutoFlushEvent#isFlushRequired`. Unfortunately this method always return `false`, because of what I think is a bug in ORM.

I'll open a ticket for ORM, but it likely will take a few weeks before the fix is released...

In the meantime, maybe you could explore an alternative solution: `org.hibernate.SessionEventListener#partialFlushEnd`. This method is called after an automatic flush is actually performed, and only if it is actually performed. Ignore the "partial" in the name, the flush is not partial at all.
So you could make `HibernateSearchEventListener` extend `BaseSessionEventListener` and move the implementation of the auto flush listener to `partialFlushEnd`.
The problem will be how to register the session listener, and unfortunately I don't have an answer for that... Maybe you can investigate?

yrodiere: Created https://hibernate.atlassian.net/browse/HHH-13592

integrationtest/showcase/library/src/test/java/org/hibernate/search/integrationtest/showcase/library/functional/LessMemoryEntityService.java
yrodiere: To me it is, since it tests a feature. Thanks.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/work/impl/PojoWorkPlanImpl.java
fax4ever: Thanks @yrodiere. 

From my investigation trying to register the listener at `SessionFactory` level there is an hibernate property `hibernate.session.events.auto` to set the fully qualified class name implementing the `SessionEventListener` interface.
It seems that there is no service to invoke to register it, like for other events.
Moreover, it is supposed to exist only one listener. So if we registered our listener, the user couldn't do the same for his/hers.

Trying to register the listener at `Session` level, that it means do it for each session we create...
It seems to be add later session listeners. I know it's no very nice, but I've not found a better solution.

I'm following the latter. Let's see if it will bring us to a solution.

fax4ever: @yrodiere it seems that switching from `onAutoFlush` to `partialFlushEnd` produces the same results. Looking at Hibernate ORM source code it seems that `autoflush` always triggers a `partialFlushEnd` event.

What would I do? For now I would make idempotent the `PojoWorkPlan` instance again, waiting for the completion of HHH-13592. Opening a follow-up of HHH-13592 on Search side.

On the other hand, any other ideas are more than welcome.

fax4ever:
@yrodiere I've dropped the commit `Avoid recursive flushes...`. I will merge it as soon as we get the green light from the CI.

fax4ever:
Merged thanks

yrodiere:
Thanks!

PR #2056, opened 2019-08-14
HSEARCH-3068 Restore support for "simulated" transaction using a Flush listener

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3068

Based on #2054

### DIFF
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/AutomaticIndexingOutOfTransactionIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/AutomaticIndexingOutOfTransactionIT.java
@@ -11,9 +11,10 @@ import static org.hibernate.search.util.impl.integrationtest.orm.OrmUtils.within
 import javax.persistence.Entity;
 import javax.persistence.Id;
 
-import org.hibernate.FlushMode;
 import org.hibernate.SessionFactory;
 import org.hibernate.cfg.AvailableSettings;
+import org.hibernate.search.mapper.orm.Search;
+import org.hibernate.search.mapper.orm.session.SearchSessionWritePlan;
 import org.hibernate.search.mapper.pojo.mapping.definition.annotation.GenericField;
 import org.hibernate.search.mapper.pojo.mapping.definition.annotation.Indexed;
 import org.hibernate.search.util.impl.integrationtest.common.rule.BackendMock;
@@ -49,7 +50,6 @@ public class AutomaticIndexingOutOfTransactionIT {
 	public void add() {
 		withinSession( sessionFactory, session -> {
 			IndexedEntity entity1 = new IndexedEntity( 1, "number1" );
-			session.setHibernateFlushMode( FlushMode.AUTO );
 			session.persist( entity1 );
 
 			backendMock.expectWorks( IndexedEntity.INDEX_NAME )
@@ -61,6 +61,30 @@ public class AutomaticIndexingOutOfTransactionIT {
 			// during the flush.
 			session.flush();
 			backendMock.verifyExpectationsMet();
+		} );
+	}
+
+	@Test
+	public void clear() {
+		withinSession( sessionFactory, session -> {
+			SearchSessionWritePlan writePlan = Search.session( session ).writePlan();
+
+			IndexedEntity entity1 = new IndexedEntity( 1, "number1" );
+			// working directly on the work plan to add works immediately (not at flush time!)
+			writePlan.addOrUpdate( entity1 );
+
+			// clearing the update of entity 1
+			session.clear();
+
+			IndexedEntity entity2 = new IndexedEntity( 2, "number2" );
+			writePlan.addOrUpdate( entity2 );
+
+			// only entity 2 is supposed to be flushed here
+			backendMock.expectWorks( IndexedEntity.INDEX_NAME )
+					.update( "2", b -> b.field( "text", "number2" ) )
+					.preparedThenExecuted();
+
+			session.flush();
 		} );
 	}
 
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/AutomaticIndexingOutOfTransactionIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/AutomaticIndexingOutOfTransactionIT.java
@@ -1,0 +1,82 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.integrationtest.mapper.orm.automaticindexing;
+
+import static org.hibernate.search.util.impl.integrationtest.orm.OrmUtils.withinSession;
+
+import javax.persistence.Entity;
+import javax.persistence.Id;
+
+import org.hibernate.FlushMode;
+import org.hibernate.SessionFactory;
+import org.hibernate.cfg.AvailableSettings;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.GenericField;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.Indexed;
+import org.hibernate.search.util.impl.integrationtest.common.rule.BackendMock;
+import org.hibernate.search.util.impl.integrationtest.orm.OrmSetupHelper;
+import org.hibernate.search.util.impl.test.annotation.TestForIssue;
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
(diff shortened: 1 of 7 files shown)


util/common/src/main/java/org/hibernate/search/util/common/concurrent/ConcurrentReferenceHashMap.java
```diff
@@ -0,0 +1,1922 @@
...
+ * This class is a member of the <a href="{@docRoot}/../technotes/guides/collections/index.html">
+ * Java Collections Framework</a>.
+ *
+ * @param <K> the type of keys maintained by this map
+ * @param <V> the type of mapped values
+ *
+ * @author Doug Lea
+ * @author Jason T. Greene
+ */
+class ConcurrentReferenceHashMap<K, V> extends AbstractMap<K, V>
```
yrodiere: Another reason not to use a `Synchronization` map: this class has proven problematic in the past and I've worked hard to get rid of it in Search 6, be it only because it's too complex for its own good. Please don't restore it.

util/common/src/main/java/org/hibernate/search/util/common/concurrent/ConcurrentReferenceHashMap.java
```diff
@@ -20,6 +20,7 @@
 import java.util.AbstractCollection;
 import java.util.AbstractMap;
```
yrodiere: When you restore classes that were obviously copy/pasted from another project, please don't refactor them. It makes it very hard to track differences between projects.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/event/impl/HibernateSearchEventListener.java
```diff
@@ -139,16 +144,15 @@ public void onPostUpdateCollection(PostCollectionUpdateEvent event) {
...
-//		Synchronization synchronization = flushSynch.get( session );
-//		if ( synchronization != null ) {
-//			//first cleanup
-//			flushSynch.remove( session );
-//			log.debug( "flush event causing index update out of transaction" );
-//			synchronization.beforeCompletion();
-//			synchronization.afterCompletion( Status.STATUS_COMMITTED );
-//		}
+		Session session = event.getSession();
+		Synchronization synchronization = flushSynch.get( session );
```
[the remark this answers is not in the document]
fax4ever: I don't know if it a good idea always calling execute on the work plan. In such way we could invoke execute on flush even if there is no update outside transactions on the current session.

It would nice to have if there was a way to detect if there are some out of transaction update operations pending on the current session.

Currently we rely on the weird synchronization map to detect them. If there was another way, I would happy about it.

fax4ever:
Made all the changes here. It looks much better now.
Thanks

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/event/impl/HibernateSearchEventListener.java
```diff
@@ -131,16 +133,32 @@ public void onPostUpdateCollection(PostCollectionUpdateEvent event) {
...
+	public void onAutoFlush(AutoFlushEvent event) throws HibernateException {
+		getCurrentWorkPlan( state.getContextProvider(), event.getSession() ).prepare();
+	}
+
+	@Override
+	public void onClear(ClearEvent event) {
+		EventSource session = event.getSession();
+		// skip the clearNotPrepared operation in case of rollback
+		if ( session.isTransactionInProgress() ) {
+			getCurrentWorkPlan( state.getContextProvider(), session ).clearNotPrepared();
```
yrodiere: I get why you originally added this, but I think it's wrong now that you support flush events outside of transactions. What happens when you change entities, clear, then flush, outside of a transaction? It shoudn't do anything, yet with this code I suspect it will update the index, or result in a LazyInitializationException, because the clear is ignored.

What was the problem with executing this code outside of a transaction, exactly?

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/session/SearchSessionWritePlanOutOfTransactionIT.java
```diff
@@ -0,0 +1,89 @@
...
+	@PortedFromSearch5(original = "TransactionTest#testTransactionCommit")
+	public void add() {
+		withinSession( sessionFactory, session -> {
+			IndexedEntity entity1 = new IndexedEntity( 1, "number1" );
+			session.setHibernateFlushMode( FlushMode.AUTO );
+			session.persist( entity1 );
+
+			// TODO HSEARCH-3068 support it
+			SubTest.expectException( () -> {
+				/*backendMock.expectWorks( IndexedEntity.INDEX_NAME )
```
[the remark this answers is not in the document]
yrodiere: Ok, so you squashed the test together with the fix. That's not what I meant, but to each his own.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/AutomaticIndexingOutOfTransactionIT.java
```diff
@@ -0,0 +1,84 @@
...
+		backendMock.expectAnySchema( IndexedEntity.INDEX_NAME );
+		sessionFactory = ormSetupHelper
+				.start()
+				.withProperty( AvailableSettings.ALLOW_UPDATE_OUTSIDE_TRANSACTION, true )
+				.setup( IndexedEntity.class );
+		backendMock.verifyExpectationsMet();
+	}
+
+	@Test
+	@PortedFromSearch5(original = "TransactionTest#testTransactionCommit")
```
yrodiere: This is not at all the same test as `TransactionTest#testTransactionCommit`, though, since the original was inside a transaction and this one is not.

Also, if you add `@PortedFromSearch5` annotations, please use the fully qualified name of the class, and mark the original method with `@Category(PortedToSearch6.class)`. This will help later when we'll try to port more tests from Search 5 to 6.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/event/impl/HibernateSearchEventListener.java
```diff
@@ -137,18 +133,18 @@ public void onPostUpdateCollection(PostCollectionUpdateEvent event) {
...
-//			synchronization.beforeCompletion();
-//			synchronization.afterCompletion( Status.STATUS_COMMITTED );
-//		}
+		HibernateOrmListenerContextProvider contextProvider = state.getContextProvider();
+		EventSource session = event.getSession();
+		PojoWorkPlan currentWorkPlan = getCurrentWorkPlan( contextProvider, session );
+		currentWorkPlan.prepare();
+
+		// flush within a transaction should trigger only the prepare phase,
+		// since the execute phase is supported to be triggered by the transaction commit
```
yrodiere: `is supported to be triggered` => `is supposed to be triggered`?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/event/impl/HibernateSearchEventListener.java
fax4ever: We would have a small issue just on rollback. Since the rollback triggers a clear event out of transaction. The issue is just about the creation of a work plan for nothing,
But we can change the predicate into:
``` java
// skip the clearNotPrepared operation in case of rollback
if ( !TransactionStatus.ROLLED_BACK.equals( session.getTransaction().getStatus() ) ) {
  getCurrentWorkPlan( state.getContextProvider(), session ).clearNotPrepared();
}
```

And I have a **change entities, clear, then flush, outside of a transaction** test to add

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/AutomaticIndexingOutOfTransactionIT.java
fax4ever: You're right tests are not equals. So I removed the annotation from there.

fax4ever:
Thanks @yrodiere. Rebased and made the changes here.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/AutomaticIndexingOutOfTransactionIT.java
```diff
@@ -64,6 +64,30 @@ public void add() {
...
 	}
 
+	@Test
+	public void clear() {
+		withinSession( sessionFactory, session -> {
+			SearchSessionWritePlan writePlan = Search.session( session ).writePlan();
+
+			IndexedEntity entity1 = new IndexedEntity( 1, "number1" );
+			// working directly on the work plan to add works immediately (not at flush time!)
+			writePlan.addOrUpdate( entity1 );
```
yrodiere: Right, that's definitely an edge case. But it's nice to have it covered :)

fax4ever:
Thanks!

PR #2057, opened 2019-08-15
HSEARCH-899 Check for illegal usage of @IndexedEmbedded

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-899

### DIFF
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/mapping/definition/AnnotationMappingDiscoveryIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/mapping/definition/AnnotationMappingDiscoveryIT.java
@@ -18,11 +18,11 @@ import javax.persistence.Embeddable;
 import javax.persistence.Embedded;
 import javax.persistence.Entity;
 import javax.persistence.Id;
+import javax.persistence.MappedSuperclass;
 
 import org.hibernate.search.engine.backend.document.DocumentElement;
 import org.hibernate.search.engine.backend.document.IndexObjectFieldReference;
 import org.hibernate.search.mapper.orm.cfg.HibernateOrmMapperSettings;
-import org.hibernate.search.mapper.orm.mapping.HibernateOrmMappingConfigurationContext;
 import org.hibernate.search.mapper.orm.mapping.HibernateOrmSearchMappingConfigurer;
 import org.hibernate.search.mapper.pojo.bridge.PropertyBridge;
 import org.hibernate.search.mapper.pojo.bridge.binding.PropertyBindingContext;
@@ -37,6 +37,7 @@ import org.hibernate.search.mapper.pojo.bridge.runtime.PropertyBridgeWriteContex
 import org.hibernate.search.mapper.pojo.mapping.definition.annotation.GenericField;
 import org.hibernate.search.mapper.pojo.mapping.definition.annotation.Indexed;
 import org.hibernate.search.mapper.pojo.mapping.definition.annotation.IndexedEmbedded;
+import org.hibernate.search.mapper.pojo.mapping.definition.programmatic.ProgrammaticMappingConfigurationContext;
 import org.hibernate.search.mapper.pojo.model.PojoModelProperty;
 import org.hibernate.search.util.impl.integrationtest.common.rule.BackendMock;
 import org.hibernate.search.util.impl.integrationtest.orm.OrmSetupHelper;
@@ -69,6 +70,7 @@ public class AnnotationMappingDiscoveryIT {
 						.objectField( "annotatedProperty", b3 -> {
 							// We do not expect any particular property in the object field added by the bridge
 						} )
+						.field( "alwaysPresent", String.class )
 				)
 				.objectField( "nonAnnotationMappedEmbedded", b2 -> b2
 						/*
@@ -76,20 +78,20 @@ public class AnnotationMappingDiscoveryIT {
 						 * which has not been registered explicitly.
 						 */
 						.field( "text", String.class )
+						.field( "alwaysPresent", String.class )
 				)
 		);
 
 		ormSetupHelper.start()
 				.withProperty(
 						HibernateOrmMapperSettings.MAPPING_CONFIGURER,
-						new HibernateOrmSearchMappingConfigurer() {
-							@Override
-							public void configure(HibernateOrmMappingConfigurationContext context) {
-								context.programmaticMapping()
-										.type( IndexedEntity.class )
-										.property( "nonAnnotationMappedEmbedded" )
-										.indexedEmbedded();
-							}
+						(HibernateOrmSearchMappingConfigurer) context -> {
+							context.programmaticMapping()
+									.type( IndexedEntity.class )
+									.property( "nonAnnotationMappedEmbedded" )
+									.indexedEmbedded();
+
+							mapAlwaysPresentProperty( context.programmaticMapping() );
 						}
 				)
 				.setup(
@@ -106,15 +108,17 @@ public class AnnotationMappingDiscoveryIT {
 		backendMock.expectSchema( IndexedEntity.INDEX, b -> b
 				.objectField( "annotationMappedEmbedded", b2 -> {
 					/*
-					 * This object field should be empty because
-					 * the annotation mapping for the embedded type has *NOT* been automatically discovered.
+					 * This object field should contain only the property mapped using the programmatic API,
+					 * because the annotation mapping for the embedded type has *NOT* been automatically discovered.
 					 */
+					b2.field( "alwaysPresent", String.class );
 				} )
 				.objectField( "nonAnnotationMappedEmbedded", b2 -> {
 					/*
-					 * This object field should be empty because
-					 * the annotation mapping for the embedded type has *NOT* been automatically discovered.
+					 * This object field should contain only the property mapped using the programmatic API,
+					 * because the annotation mapping for the embedded type has *NOT* been automatically discovered.
 					 */
+					b2.field( "alwaysPresent", String.class );
 				} )
 		);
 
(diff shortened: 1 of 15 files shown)


integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/mapping/definition/AnnotationMappingDiscoveryIT.java
```diff
@@ -103,21 +105,7 @@ public void configure(HibernateOrmMappingConfigurationContext context) {
 
 	@Test
 	public void discoveryDisabled() {
-		backendMock.expectSchema( IndexedEntity.INDEX, b -> b
```
yrodiere: You're completely changing the purpose of this test. Please don't do that.

This test is about checking that type annotation mapping discovery is effectively disabled. We used to check that by asserting that no mapping was contributed for the two `@IndexedEmbedded`. We can no longer do that because of the check you added.

One solution would be to add a programmatic mapping for the embedded types that adds some fields to these types, different from the ones configured through annotations. We would check that the programatically defined fields are added, but not the ones defined through annotations.

integrationtest/mapper/pojo/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/AnnotationMappingDiscoveryIT.java
```diff
@@ -93,22 +95,8 @@ public void discoveryEnabled() {
 
 	@Test
 	public void discoveryDisabled() {
-		backendMock.expectSchema( IndexedEntity.INDEX, b -> b
```
yrodiere: Same comment as `integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/mapping/definition/AnnotationMappingDiscoveryIT.java`. Please don't repurpose existing tests.

integrationtest/mapper/pojo/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/IndexedEmbeddedBaseIT.java
```diff
@@ -657,18 +657,12 @@ public String getInvalid() {
...
 
-		backendMock.expectSchema( INDEX_NAME, b -> b
-				.field( "text", String.class )
-				// TODO HSEARCH-899 this isn't supposed to be legal => throw an exception for
-				.objectField( "invalid", c -> {
-				} )
-		);
-
-		JavaBeanMapping mapping = setupHelper.start()
+		SubTest.expectException( () -> setupHelper.start()
```
yrodiere: I think it would be clearer if you set these expectations right from the start, in the commit that adds the test. But to each his own, I guess.

integrationtest/mapper/pojo/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/IndexedEmbeddedBaseIT.java
```diff
@@ -657,18 +657,12 @@ public String getInvalid() {
...
 				.withAnnotatedEntityTypes( IndexedEntity.class )
-				.setup();
-
-		backendMock.verifyExpectationsMet();
+				.setup() )
+				.assertThrown()
+				.isInstanceOf( SearchException.class )
+				.hasMessageContaining( "HSEARCH700060: The indexed embedded property 'invalid' of type 'java.lang.String' does not provide any valid field." );
```
yrodiere: Please don't mention the error code ("HSEARCH700060") in these assertions, they may change before the release (upon merging the PR, for example) and to be honest we don't really care about the exact code.

integrationtest/mapper/pojo/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/IndexedEmbeddedBaseIT.java
```diff
@@ -708,18 +702,13 @@ public EmptyNested getInvalid() {
...
-		JavaBeanMapping mapping = setupHelper.start()
+		SubTest.expectException( () -> setupHelper.start()
 				.withAnnotatedEntityTypes( IndexedEntity.class )
-				.setup();
-
-		backendMock.verifyExpectationsMet();
+				.setup() )
+				.assertThrown()
+				.isInstanceOf( SearchException.class )
+				.hasMessageContaining( "HSEARCH700060: The indexed embedded property 'invalid' of type" )
```
yrodiere: Same comment here about the error code and context of the error.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/processing/building/impl/PojoIndexingProcessorTypeNodeBuilder.java
```diff
@@ -135,13 +140,16 @@ public void closeOnFailure() {
...
+			if ( immutableBridgeHolders.isEmpty() && immutablePropertyNodes.isEmpty() ) {
+				if ( parentIndexObjectReferences.isEmpty() ) {
+					/*
+					 * If this node doesn't create any object in the document, and it doesn't have any bridge,
+					 * nor any property node, then it is useless and we don't need to build it.
+					 */
+					return Optional.empty();
+				}
+
+				throw log.invalidIndexedEmbedded( modelPath.getParent().getParent().getPropertyModel().getName(), modelPath.getTypeModel().getName() );
```
yrodiere: This is not the right place for this error. I explicitly discussed the right way to implement this fix with you, and I mentioned why we needed to do it that way. I guess I wasn't clear enough.

Fields can be filtered out. Consider this:

```
@Entity 
@Indexed
public class Level1 {
    @IndexedEmbedded(includePaths = "level3b.text")
    private Level2 level2;
}

@Entity 
@Indexed
public class Level2 {
    @IndexedEmbedded
    private Level3A level3a;

    @IndexedEmbedded
    private Level3B level3b;
}

@Entity 
public class Level3A {
    @GenericField
    private String text;
}

@Entity 
public class Level3B {
    @GenericField
    private String text;
}
```

When building the indexing processor for `Level1`, it's entirely possible that a type node is created for `level3a`, but it won't define any field, because the `includePaths` clause filters it out.
Now this specific setup might not fail because there are optimizations in place which may trigger an early stop and prevent the type node from being created, but you get the idea: I'm sure there are other cases where an empty type node can legitimately be created.

The right way to address this problem, as we discussed previously, is to:

1. Replace `org.hibernate.search.engine.mapper.mapping.building.spi.TypeMetadataContributorProvider#forEach` with a method that takes a type model as a paremeter, and returns a set of `C`. Essentially `Set<C> get(MappableTypeModel typeModel)`.
2. In `org.hibernate.search.mapper.pojo.processing.building.impl.PojoIndexingProcessorValueNodeBuilderDelegate#indexedEmbedded`, test the returned set before doing anything: if it's empty, it means there isn't any mapping defined for the embedded type, and we should throw the exception.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/logging/impl/Log.java
```diff
@@ -403,4 +403,7 @@ SearchException cannotPurgeNonIndexedContainedType(
 			value = "The bind() method of binder '%1$s' is not implemented correctly:"
 					+ " it did not call context.setMarker(...).")
 	SearchException missingMarkerForBinder(Object binder);
+
+	@Message(id = ID_OFFSET_2 + 60, value = "The indexed embedded property '%1$s' of type '%2$s' does not provide any valid field.")
+	SearchException invalidIndexedEmbedded(String propertyName, String propertyType);
```
yrodiere: Following the change in the implementation mentioned in my other comment, I'd rephrase the error like this:

```
	@Message(id = ID_OFFSET_2 + 60, value = "Type '%1$s' cannot be indexed-embedded, because no index mapping (@GenericField, @FullTextField, ...) is defined for that type.")
	SearchException invalidIndexedEmbedded(@FormatWith(PojoTypeModelFormatter.class) PojoTypeModel<?> typeModel); 
```

No need to mention the property name, it will be automatically included in the failure report (you can check that).

integrationtest/mapper/pojo/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/IndexedEmbeddedBaseIT.java
yrodiere: Also, you'll probably want to check the context of the error (type, property name) in the failure report. See how we define expectations in `error_includePaths_nonMatched`.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/mapping/definition/AnnotationMappingDiscoveryIT.java
fax4ever: Good idea. Let's do that!

integrationtest/mapper/pojo/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/IndexedEmbeddedBaseIT.java
fax4ever: Squashed with the subsequent one.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/processing/building/impl/PojoIndexingProcessorTypeNodeBuilder.java
fax4ever: Sorry if I did remember that. Actually I don't have a good memory.
Not a problem to move the place where to address it.

Just for my very curiosity. How does it possible that:
> it won't define any field, because the includePaths clause filters it out.

if `Level3B` defines the generic field `text`?
Thanks ;)

fax4ever: @yrodiere sorry. I'm trying to apply the strategy I think you pointed out here.

Probably I didn't grasp something. But the list is not empty, even if we have a String as a nested model type.

My attempt resides in the commit named:
`HSEARCH-899 Move check to PojoIndexingProcessorValueNodeBuilderDelegate`.

Thanks

fax4ever:
@yrodiere thanks! 
Fixed the cases. Removing the odd '<no value extractors>' from the expected path context....

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/processing/building/impl/PojoIndexingProcessorTypeNodeBuilder.java
yrodiere: Looks like we solved this.

Regarding your question:

> Just for my very curiosity. How does it possible that:
>
> >    it won't define any field, because the includePaths clause filters it out.
>
> if Level3B defines the generic field text?

It's `level3a` which may end up empty, because it is not mentioned in `includePaths`. `level3b` and `level3b.text` will definitely be included, since they are explicitly mentioned in `includePaths`.

yrodiere:
Looks good, merging. Thanks!

PR #2058, opened 2019-08-18
HSEARCH-3564 Type cast too strong

carstenartur:
https://hibernate.atlassian.net/browse/HSEARCH-3564


fax4ever:
Hi @carstenartur. Thanks for the contribution.

It seems that the commit does not compile:
~~~
[ERROR] COMPILATION ERROR : 
[ERROR] /home/fercoli/code/hibernate-search/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchSearchProjection.java:[66,65] incompatible types: java.lang.Object cannot be converted to capture#1 of ?
[INFO] 1 error
~~~

Moreover, what's the purpose of the contribution?
Many thanks
Fabio

yrodiere:
Hi @carstenartur ,

I appreciate the contribution, and some of these suggestions are legitimate clean-ups, but some of them are wrong, like the one on `new Appendable()` in the tests, or the one that breaks the build.

There may be strong reasons, beyond everyday maintenance, to perform this cleanup, such as the project not compiling in a specific IDE. If that's the case, please create a ticket on JIRA so we can discuss how to best address these problems: https://hibernate.atlassian.net/projects/HSEARCH.

In the meantime, I'll close this PR.

carstenartur:
No problem.
missed that error, sorry, 
May have missunderstood HSEARCH-3564

@Fabio:
This is about fixes of type casts, which are too strict and casting to a more general type would be sufficient for the actual expression. Sometimes, an instance is casted to a specific subtype, but methods invoked or fields used from the casted instance are actually defined by some supertype. In that case cast to too specific type introduces an unnecessary coupling to the code and limits its extensibility.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchSearchProjection.java
```diff
@@ -63,7 +63,7 @@ E extract(ProjectionHitMapper<?, ?> projectionHitMapper,
 	@SuppressWarnings({ "unchecked", "rawtypes" })
 	static <Z> Z transformUnsafe(ElasticsearchSearchProjection<?, Z> projection, LoadingResult<?> loadingResult,
 			Object extractedData, SearchProjectionTransformContext context) {
-		return (Z) ( (ElasticsearchSearchProjection) projection ).transform( loadingResult, extractedData, context );
```
yrodiere: This cast is necessary to bypass generics limitations of the compiler. We can't remove it.

engine/src/main/java/org/hibernate/search/engine/cfg/spi/ConvertUtils.java
```diff
@@ -310,7 +310,7 @@ public static BigInteger convertBigInteger(Object value) {
 					.collect( Collectors.toList() );
 		}
 		else if ( value instanceof String ) {
-			return separatorPattern.splitAsStream( (String) value )
```
yrodiere: This isn't consistent with the `instanceof` above, nor with the other `convert` methods we have in the same class. I'd be in favor of leaving this as-is.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/work/impl/LuceneSearchWork.java
```diff
@@ -89,7 +89,7 @@ else if ( limit == 0L ) {
 			return 0;
 		}
 		else {
-			return Math.min( (int) ( offset + limit ), reader.maxDoc() );
```
yrodiere: This cast should indeed be removed, but the new code doesn't follow the Hibernate code style. There should be a space after an opening parenthesis (missing here), and a space before a closing parenthesis.

See http://hibernate.org/search/contribute/, in particular see the code style configuration for [Eclipse](https://github.com/hibernate/hibernate-ide-codestyles/blob/master/eclipse/hibernate-code-template-search.xml) or [IntelliJ IDEA](https://github.com/hibernate/hibernate-ide-codestyles/blob/master/intellij-14/hibernate_noorm.xml).

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/query/SearchQueryFetchIT.java
```diff
@@ -84,7 +84,7 @@ public void fetch_noArg() {
 
 	@Test
 	public void fetch_limit() {
-		assertThat( matchAllQuery().fetch( (Integer) null ) )
```
yrodiere: Code style: there should be a space before and after `null`.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/query/SearchQueryFetchIT.java
```diff
@@ -152,7 +152,7 @@ public void fetchHits_noArg() {
 
 	@Test
 	public void fetchHits_limit() {
-		assertThat( matchAllQuery().fetchHits( (Integer) null ) )
```
yrodiere: Code style: there should be a space before and after `null`.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/smoke/usertype/OptionalStringUserType.java
```diff
@@ -59,7 +59,7 @@ public void nullSafeSet(PreparedStatement st, Object value, int index, SharedSes
 			delegateType.nullSafeSet( st, null, index, session );
 		}
 		else {
-			delegateType.nullSafeSet( st, (String) optional.get(), index, session );
```
yrodiere: Code style: there should be a space before `st`.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/model/impl/HibernateOrmPathFilterFactory.java
```diff
@@ -305,7 +306,7 @@ else if ( collectionValue instanceof org.hibernate.mapping.Map ) {
 				 * Do not let ORM confuse you: getKey() doesn't return the value of the map key,
 				 * but the value of the foreign key to the targeted entity...
 				 */
-				return ( (org.hibernate.mapping.Map) collectionValue ).getIndex();
```
yrodiere: This is inconsistent with the `instanceof` above. I would not change it.

util/common/src/test/java/org/hibernate/search/util/common/impl/ToStringTreeBuilderTest.java
```diff
@@ -142,8 +142,8 @@ public String toString() {
 					.startObject( "emptyChild" )
 						.endObject()
 					.attribute( "appendable", new Appendable() )
-					.attribute( "appendableAsObject", (Object) new Appendable() )
-					.attribute( "nullAppendable", (Appendable) null )
```
yrodiere: Code style: there should be a space after an opening parenthesis and before a closing parenthesis.

util/common/src/test/java/org/hibernate/search/util/common/impl/ToStringTreeBuilderTest.java
```diff
@@ -142,8 +142,8 @@ public String toString() {
 					.startObject( "emptyChild" )
 						.endObject()
 					.attribute( "appendable", new Appendable() )
-					.attribute( "appendableAsObject", (Object) new Appendable() )
```
yrodiere: This cast allows us to check that the method accepts an `Object`. I'd keep it?

yrodiere:
Fixed the few remaining style issues, squashed, added the ticket number to the commit message, and merged to master.
Thanks!

PR #2061, opened 2019-08-21
HSEARCH-1270 Test nested transaction flush and rollback

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-1270

### DIFF
--- a/integrationtest/showcase/library/src/test/java/org/hibernate/search/integrationtest/showcase/library/service/InnerTransactionRollbackService.java
+++ b/integrationtest/showcase/library/src/test/java/org/hibernate/search/integrationtest/showcase/library/service/InnerTransactionRollbackService.java
@@ -1,0 +1,72 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.integrationtest.showcase.library.service;
+
+import javax.persistence.EntityManager;
+
+import org.hibernate.search.integrationtest.showcase.library.model.Book;
+
+import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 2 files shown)


fax4ever:
@yrodiere You're definitely right!
In the new version I used `Propagation.REQUIRES_NEW` to reproduce the case. Moreover I check better the exception :P

integrationtest/showcase/library/src/test/java/org/hibernate/search/integrationtest/showcase/library/service/InnerTransactionRollbackService.java
```diff
@@ -0,0 +1,72 @@
...
+		Book book = new Book( 731, isbn, "The Prince", "Niccolò Machiavelli", "Describes the human people behaviour...", "politics" );
+		entityManager.persist( book );
+		entityManager.flush();
+		throw new MakeRollbackException( "Rollback the inner transaction" );
+	}
+
+	public void doOuter(String outerIsbn, String nestedIsbn) {
+		// This data is supposed to be pushed to the index:
+		Book book = new Book( 739, outerIsbn, "The Late Mattia Pascal", "Luigi Pirandello", "Describes the human life conditions...", "misfortune" );
+		entityManager.persist( book );
```
yrodiere: This looks great. Just one more thing: could you please duplicate this method, add a flush just here, and add a test to check the behavior is the same?

Just trying to rule out a potential bug where the rollback of the inner transaction would trigger a session clear, which would clear the works of the outer transaction... I don't think it's the case, but since you've discovered weird things around clear recently, I have a doubt.

fax4ever: Done. The session clear is triggered by the inner transaction rollback, but it does not affect the already prepared works. Only the rollback of the outer could do that. But the rollback clear cannot access to a work plan owned by a transaction.

yrodiere:
Merged, thanks!

fax4ever:
Thank you!

PR #2063, opened 2019-08-22
HSEARCH-3564 Protected member in final class

carstenartur:
any instances of members being declared protected in classes that are
declared final. Such members may be declared private or package-visible
instead.


yrodiere:
I will revert the change about `BindingUsingPropertyMarkerAccessIT`, because the particular class you changed was specifically about testing `protected` fields.

yrodiere:
Removed the bit about `BindingUsingPropertyMarkerAccessIT`, added the ticket number to the commit message, and merged to master.
Thanks!

PR #2064, opened 2019-08-23
HSEARCH-3656 Update mapper ORM documentation chapter: writeplan API process is no longer necessary

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3656

@fax4ever Could you review this today, please? I'd like to put it in today's release.

### DIFF
--- a/documentation/src/main/asciidoc/mapper-orm-indexing-automatic.asciidoc
+++ b/documentation/src/main/asciidoc/mapper-orm-indexing-automatic.asciidoc
@@ -7,6 +7,9 @@ By default, every time an entity is changed through a Hibernate ORM Session,
 if that entity is <<mapper-orm-entityindexmapping,mapped to an index>>,
 Hibernate Search updates the relevant index.
 
+To be precise, index updates happen on transaction commit or,
+if working outside of a transaction, on session flush.
+
 [[mapper-orm-indexing-automatic-configuration]]
 == Configuration
 
@@ -15,7 +18,7 @@ or if you update it regularly through <<mapper-orm-indexing-explicit,explicit in
 You can enable or disable automatic indexing by setting the configuration property
 `hibernate.search.automatic_indexing.strategy`:
 
-* when set to `session`, each change to an indexed entity
+* when set to `session` (the default), each change to an indexed entity
 (persist, update, delete) through a Hibernate ORM Session/EntityManager
 will automatically lead to a similar modification to the index.
 * when set to `none`, changes to entities are ignored,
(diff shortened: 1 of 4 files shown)


mapper/orm/src/main/java/org/hibernate/search/mapper/orm/session/SearchSessionWritePlan.java
```diff
@@ -6,16 +6,27 @@
...
  * <p>
- * When a transaction is active, operations are queued internally,
- * and are only applied on transaction commit.
+ * This class is stateful: it queues operations internally to apply them at a later time.
  * <p>
- * {@link #process()} and {@link #execute()} allow to control explicitly
- * how operations are applied, without relying on transaction commits,
- * which can be useful when indexing large amounts of data in batches.
+ * When {@link #process()} is called,
+ * or when {@link org.hibernate.search.mapper.orm.cfg.HibernateOrmMapperSettings#AUTOMATIC_INDEXING_STRATEGY automatic indexing is enabled}
```
fax4ever: I don't think that the AUTOMATIC_INDEXING_STRATEGY is mandatory here. I would write in case of session flush (explicit, automatic, dirty check, query flush, entity flush).

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/session/SearchSessionWritePlan.java
```diff
@@ -6,16 +6,27 @@
...
+ * and a Hibernate ORM Session {@code flush()} happens,
+ * the entities will be processed and index documents will be built
+ * and stored in an internal buffer.
+ * <p>
+ * When {@link #execute()} is called,
+ * or when {@link org.hibernate.search.mapper.orm.cfg.HibernateOrmMapperSettings#AUTOMATIC_INDEXING_STRATEGY automatic indexing is enabled}
```
fax4ever: Same here. I don't think strategy really matters. I would write that the execution of the prepared works will be triggered by a commit or by an out of transaction flush.

documentation/src/main/asciidoc/mapper-orm-indexing-automatic.asciidoc
```diff
@@ -35,22 +38,40 @@ are not detected by Hibernate Search.
...
-See <<mapper-orm-indexing-explicit-writeplan-process-execute>> for a few solutions.
+However, if you perform a batch process inside a transaction,
+and perform flush/clear, regularly to save memory,
+be aware that Hibernate Search's internal buffer holding documents to index
+will grow on each flush, and will not be cleared until the transaction is committed or rolled back.
+If you encounter memory issues because of that,
+see <<mapper-orm-indexing-explicit-writeplan-process-execute>> for a few solutions.
+Outside of transactions, indexing happens on session flush::
+When entity changes happen outside of any transaction (not recommended),
+indexes are updated immediately upon session `flush()`.
```
fax4ever: This is perfect. But I would add something like:
> Without that flush the changes will be lost.

documentation/src/main/asciidoc/mapper-orm-indexing-explicit.asciidoc
```diff
@@ -150,36 +110,42 @@ See <<mapper-orm-indexing-explicit-writeplan-writes>>.
...
-However, if enough memory is available,
-this is the safest pattern, leaving both the database and indexes unaffected
-if an error happens during the batch process.
+With this pattern, if an exception is thrown,
+part of the data will already be in the index, with no way to roll back the changes,
+while the database changes will have been rolled back.
+The index will thus be inconsistent with the database.
+
+To recover from that situation, you will have to either cancel your database changes manually,
+or wipe the indexes clean and <<mapper-orm-indexing-explicit-massindexer,reindex>>.
```
fax4ever: Maybe it is too scary. Usually the user can just resume the batch process from the first not committed chunk. I would mention this option too.

fax4ever: In a recovery of consistency process a forward compensation can be easier and safer than a backward one.

yrodiere: This is scary on purpose, because this strategy is a bad idea. But I'm sure some people will want it anyway, so I mention it for completeness.

That being said, you're right that there's something wrong here: there are no database changes to cancel, since they were rolled back.
I'll change this part.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/session/SearchSessionWritePlan.java
yrodiere: It does matter, since it controls whether automatic indexing is enabled or not.

I think you're confusing the automatic indexing strategy with the synchronization strategy.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/session/SearchSessionWritePlan.java
yrodiere: Same answer.

yrodiere:
Fixed a few problems and merged. Thanks!

PR #2067, opened 2019-08-26
HSEARCH-3667 Envers compatibility

dcdh:
https://hibernate.atlassian.net/browse/HSEARCH-3667

### DIFF
--- a/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/model/impl/HibernateOrmBootstrapIntrospector.java
+++ b/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/model/impl/HibernateOrmBootstrapIntrospector.java
@@ -16,6 +16,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.hibernate.AssertionFailure;
@@ -56,9 +57,12 @@ public class HibernateOrmBootstrapIntrospector extends AbstractPojoHCAnnBootstra
 	public static HibernateOrmBootstrapIntrospector create(Metadata metadata,
 			ReflectionManager ormReflectionManager,
 			ConfigurationPropertySource propertySource) {
-		Collection<PersistentClass> persistentClasses = metadata.getEntityBindings();
+		Collection<PersistentClass> persistentClasses = metadata.getEntityBindings()
+				.stream()
+				.filter( PersistentClass::hasPojoRepresentation )
+				.collect( Collectors.toList() );
 		Map<Class<?>, HibernateOrmBasicTypeMetadata> typeMetadata = new HashMap<>();
-		collectPersistentTypes( typeMetadata, metadata.getEntityBindings() );
+		collectPersistentTypes( typeMetadata, persistentClasses );
 		for ( PersistentClass persistentClass : persistentClasses ) {
 			collectEmbeddedTypesRecursively( typeMetadata, persistentClass.getIdentifier() );
 			collectEmbeddedTypesRecursively( typeMetadata, persistentClass.getPropertyIterator() );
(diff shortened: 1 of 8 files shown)


dcdh:
I've create this pull request to fix an issue when using Hibernate Search with Envers.

However I've got two remarks to do about my PR:
1. I did not find a way to implement tests from your one;
2. No dependency is present on Envers into Search which is logic. I add to hard code the full qualified name for **DefaultRevisionEntity** which may vary in time and introduce a specific code which I am not fan...

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/model/impl/HibernateOrmBootstrapIntrospector.java
```diff
@@ -56,7 +57,11 @@
 	public static HibernateOrmBootstrapIntrospector create(Metadata metadata,
 			ReflectionManager ormReflectionManager,
 			ConfigurationPropertySource propertySource) {
-		Collection<PersistentClass> persistentClasses = metadata.getEntityBindings();
+		Collection<PersistentClass> persistentClasses = metadata.getEntityBindings()
+				.stream()
+				.filter( PersistentClass::hasPojoRepresentation )
+				.filter( persistentClass -> !"org.hibernate.envers.DefaultRevisionEntity".equals( persistentClass.getClassName() ) )
```
yrodiere: I'm not sure why you need to explicitly exclude `DefaultRevisionEntity`? From what I can see it doesn't have the same issue as the other "non-pojo" persistent classes, i.e. it won't trigger an NPE, and it's not annotated with `@Indexed` so Hibernate Search should ignore it in the end (it will just collect metadata and then forget about it).

yrodiere:
@dcdh Hello!

Do you think you'll be able to add a test to this PR, so we can merge it?

In case you want to do it, I provided some guidance in my earlier comment regarding how to add the test.

If you don't have the time, please let us know so we can plan accordingly.

Thanks.

dcdh:
Hello Yoann,

I'm sorry for my delay to response. I was quite busy.

I will give you tonight some remarks about it.

Regards,

Damien Clément d'Huart

Le jeu. 5 sept. 2019 à 09:15, Yoann Rodière <notifications@github.com> a
écrit :

dcdh:
Hello Yoann,

Sorry for the late reply. I'm working on my free time to integrate hibernate ORM + Envers + Search under Quarkus.

Regarding Hibernate it seems that Hibernate Search is decoupled from Hibernate Core. The link is made in Quarkus via the `HibernateSearchElasticsearchRecorder` which initialize Search after Core.
Regarding tests, firstly I was thinking that I should made it as a bootstrap. However it was before I understand about integration in Quarkus and the use of HibernateSearchElasticsearchRecorder.

So I admit I am stuck on testing this PR...

yrodiere:
Ok, thanks for the answer. We'll give it a try.

yrodiere:
Alright, I implemented a few (simple) tests, ported from Search 5.

Goods news is: your patch does solve the problem, and Envers seems to be working now. Thank you very much for bringing this to our attention and taking the time to fix it!

I just launched a full build of this PR, testing in all supported environments (multiple databases, in particular). Let's see how it turns out, and if it succeeds I will merge.

Regarding Quarkus, one thing to keep in mind is that Quarkus takes Hibernate Search and performs extra steps to integrate it and make it compatible with all of its improvements (native images, compile-time boot, ...). The integration tests in the Hibernate Search project only run on a "traditional" hotspot JVM without any of the Quarkus bits. So in the context of this PR, Quarkus is not relevant.

That being said, there will be a need to test the Envers integration in Quarkus. I think you should be able to discuss that in https://github.com/quarkusio/quarkus/issues/3690 . @gsmet is the person to talk to if anything doesn't work.

yrodiere:
Merged! This fix will be part of the next release. Thanks again!

PR #2074, opened 2019-09-04
HSEARCH-3690 Add a default identifier bridge for type String

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3690


yrodiere:
Merged, thanks

PR #2084, opened 2019-09-11
HSEARCH-3671 Make SearchScope session-independent

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3671


documentation/src/test/java/org/hibernate/search/documentation/gettingstarted/withhsearch/withoutanalysis/GettingStartedWithoutAnalysisIT.java
```diff
@@ -100,11 +101,11 @@ public void test() {
...
 			// tag::searching-objects[]
 			// Not shown: get the entity manager and open a transaction
-			SearchSession searchSession = Search.session( entityManager ); // <1>
+			SearchMapping searchMapping = Search.mapping( entityManager.getEntityManagerFactory() ); // <1>
 
-			SearchScope<Book> scope = searchSession.scope( Book.class ); // <2>
+			SearchScope<Book> scope = searchMapping.scope( Book.class ); // <2>
 
-			SearchResult<Book> result = scope.search() // <3>
+			SearchResult<Book> result = scope.search( entityManager ) // <3>
```
gsmet: Is this mandatory? I must admit I'm not excited by the fact that we have to passe the EMF *and* the EM. It makes things far less fluid IMHO.

yrodiere: Eventually, people who don't want to load entities will be allowed to call `scope.search()` (if using single-tenancy) or `scope.search(tenantId)` (if using multi-tenancy). That's not implemented yet.

For people who actually want to load entities, well, we need an entity manager. Yes, we have a factory, but I don't think creating the entity manager ourselves silently would be a good idea. The loaded entities would not be usable anyway.

I could provide a way to pass the scope to a `SearchSession`, e.g. `searchSession.search( scope )`, but then you need an entity manager to create the `SearchSession`, so it's just more verbose.

Also, please keep in mind this is just the syntax with scopes, which is not really the one I want to promote. Let's say it's an alternate, lower-level syntax. The first syntax, with a `SearchSession`, is much better in most cases.

yrodiere:
I just pushed a few other commits to not require passing an entity manager when creating a `SearchWriter` or a `MassIndexer`. I'm beginning to think that, maybe, for those two cases, we should promote the mapping + scope syntax rather than the session syntax. I.e.: 

```
MassIndexer indexer = Search.mapping( entityManagerFactory ).scope( MyEntity.class ).massIndexer();
```

Rather than:

```
// This entity manager is created for the only purpose of calling our API. It's not really useful.
EntityManager entityManager = entityManagerFactory.createEntityManager();
MassIndexer indexer = Search.session( entityManager ).massIndexer( MyEntity.class );
```

... but I suppose it's too verbose? :)

documentation/src/test/java/org/hibernate/search/documentation/gettingstarted/withhsearch/withoutanalysis/GettingStartedWithoutAnalysisIT.java
yrodiere: If the only problem you have is with the example itself, and the appearance of code for users who just don't want lambdas, I suppose I could also restore the `scope()` methods on the session. Then users will be able to trigger the search and create scope from a single object that was created from an entity manager.

However, the scopes will still be session-independent, so the `search()` method on those will still take an `EntityManager` in parameter. This method generally won't be useful though, since people can just call `search` on the session.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/Search.java
```diff
@@ -6,29 +6,55 @@
...
+	public static SearchMapping mapping(SessionFactory sessionFactory) {
+		SessionFactoryImplementor sessionFactoryImpl =
+				HibernateOrmUtils.toSessionFactoryImplementor( sessionFactory );
+		return getSearchMapping( sessionFactoryImpl );
+	}
+
+	/**
+	 * Retrieve the {@link SearchMapping} from a JPA {@link EntityManagerFactory}.
+	 *
+	 * @param entityManagerFactory A Hibernate ORM session factory.
```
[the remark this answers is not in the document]
yrodiere: Right, done

yrodiere:
I reverted some of the changes to keep the previous syntax `searchSession.scope( ... )` and introduce a new `searchSession.search( scope )` method.

So the syntax with object (which, keep that in mind, is an _alternative_ syntax) will be:

```
SearchSession searchSession = Search.session( entityManager );
SearchScope<Book> scope = searchSession.scope( Book.class );
return searchSession.search( scope )
    .predicate( scope.predicate().matchAll().toPredicate() )
    .fetchHits( 20 );
```

... with the possibility to create the scope from the mapping directly, if need be (no example in the documentation yet, I'll introduce some when it's actually useful).

It is, I think, fairly similar to what we used to have:

```
SearchSession searchSession = Search.session( entityManager );
SearchScope<Book> scope = searchSession.scope( Book.class );
return scope.search()
    .predicate( scope.predicate().matchAll().toPredicate() )
    .fetchHits( 20 );
```

yrodiere:
@fax4ever Actually never mind, I think you already have enough review work. I addressed your comments with the last few commits, so I'll merge once the CI build succeeds.

fax4ever:
Anyway @yrodiere. seeing the API, not handling a full review of the changes, I can say we have reached a good compromise.

yrodiere:
Thanks @gsmet and @fax4ever ! Merged.

PR #2088, opened 2019-09-13
HSEARCH-3704 Use less alarming wording in warnings about 6 being work in progress

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3704


fax4ever:
Merged thanks!

PR #2089, opened 2019-09-13
HSEARCH-3606 Revisit the names of "SearchWriter" and "SearchSessionWritePlan"

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3606

~Based on #2084 , which should be merged first.~ => Done

I renamed `SearchSessionWritePlan` to `SearchIndexingPlan`. It's both simpler, and more to the point: this is about indexing.
Also, it doesn't mention the session, which could prove useful in the future if we want to use it in APIs allowing users to implement their own mass indexer

I renamed `SearchWriter` to `SearchWorkspace`. The new name is still quite generic, but it makes more sense: a workspace is where work is done, it delimits what part of the indexes will affected.
Also, this name would be suitable for a future zero-downtime indexing solution: by creating a workspace with specific parameters, a user could trigger the creation of a temporary workspaces targeting temporary indexes, which could ultimately replace the main indexes.

### DIFF
--- a/mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/work/impl/PojoIndexedTypeIndexingPlan.java
+++ b/mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/work/impl/PojoIndexedTypeIndexingPlan.java
@@ -89,7 +89,7 @@ public class PojoIndexedTypeIndexingPlan<I, E, D extends DocumentElement> extend
 		}
 	}
 
-	void prepare() {
+	void process() {
 		sendCommandsToDelegate();
 		getDelegate().prepare();
 	}
@@ -107,7 +107,7 @@ public class PojoIndexedTypeIndexingPlan<I, E, D extends DocumentElement> extend
 		delegate.discard();
 	}
 
-	void clearNotPrepared() {
+	void discardNotProcessed() {
 		this.indexingPlansPerId.clear();
 	}
 
--- a/mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/work/impl/PojoIndexedTypeIndexingPlan.java
+++ b/mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/work/impl/PojoIndexedTypeIndexingPlan.java
@@ -91,7 +91,7 @@ public class PojoIndexedTypeIndexingPlan<I, E, D extends DocumentElement> extend
 
 	void process() {
 		sendCommandsToDelegate();
-		getDelegate().prepare();
+		getDelegate().process();
 	}
 
 	CompletableFuture<?> execute() {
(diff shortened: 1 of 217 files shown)


mapper/orm/src/main/java/org/hibernate/search/mapper/orm/session/impl/HibernateOrmSearchSessionSession.java
```diff
@@ -50,23 +50,24 @@
 /**
  * The actual implementation of {@link SearchSession}.
  */
-public class HibernateOrmSearchSession extends AbstractPojoSearchSession
-		implements SearchSession, HibernateOrmScopeSessionContext, ReferenceHitMapper<EntityReference> {
+public class HibernateOrmSearchSessionSession extends AbstractPojoSearchSession
```
fax4ever: I know that name will be reverted by a subsequent commit. But is there a reason to have it in the history?

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/work/impl/PojoIndexedTypeIndexingPlan.java
```diff
@@ -44,58 +44,58 @@ public PojoIndexedTypeWorkPlan(PojoWorkIndexedTypeContext<I, E, D> typeContext,
...
-		for ( IndexedEntityWorkPlan workPerDocument : workPlansPerId.values() ) {
-			workPerDocument.resolveDirty( containingEntityCollector );
+		for ( IndexedEntityIndexingPlan plan : indexingPlansPerId.values() ) {
+			plan.resolveDirty( containingEntityCollector );
 		}
 	}
 
-	void prepare() {
-		sendWorksToDelegate();
+	void process() {
```
fax4ever: Is there a reason to use `process` at mapper level and `prepare` at engine/backend level?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/session/impl/HibernateOrmSearchSessionSession.java
yrodiere: I squashed a fix-up commit into the wrong commit. Fixed, thanks.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/work/impl/PojoIndexedTypeIndexingPlan.java
yrodiere: Well... I didn't want to make this a huge refactoring, so I left out most of the backend side. But since I'm not alone to think the wording in the backend is inconsistent... I added a few commits to address this, and a few more problems.

yrodiere:
Merged, thanks!

PR #2090, opened 2019-09-13
HSEARCH-3669 + HSEARCH-3707 Rename fetch()/fetchHits() to fetchAll/fetchAllHits and swap the parameters of fetch(Integer,Integer)

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3669
https://hibernate.atlassian.net/browse/HSEARCH-3707

@gsmet @fax4ever I'd be interested in your opinion...


yrodiere:
Ok, if it's only intuitive to me, I guess it's not intuitive at all. I reverted the name change (`fetchFrom` is back to `fetch`).

I'll leave the parameter swapping, though, because it looks like it really is the general convention (Spring Data's `PageRequest`, Panache's `Page.of`, ...). It will be an annoying backward-incompatible change, but that was to be expected from an Alpha release.

PR #2092, opened 2019-09-17
HSEARCH-3670 Handle sql.Date in DefaultJavaUtilDateValueBridge

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3670

### DIFF
--- a/mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/bridge/builtin/impl/DefaultJavaUtilDateValueBridge.java
+++ b/mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/bridge/builtin/impl/DefaultJavaUtilDateValueBridge.java
@@ -23,7 +23,12 @@ public final class DefaultJavaUtilDateValueBridge implements ValueBridge<Date, I
 
 	@Override
 	public Instant toIndexedValue(Date value, ValueBridgeToIndexedValueContext context) {
-		return value == null ? null : value.toInstant();
+		if ( value == null ) {
+			return null;
+		}
+
+		// java.sql.* types do not support toInstant(). See HSEARCH-3670
+		return Instant.ofEpochMilli( value.getTime() );
 	}
 
 	@Override
--- a/documentation/src/main/asciidoc/mapper-orm-mapping.asciidoc
+++ b/documentation/src/main/asciidoc/mapper-orm-mapping.asciidoc
@@ -489,10 +489,10 @@ or <<backend-elasticsearch-field-types,Elasticsearch field types>> depending on 
 |java.time.MonthDay|- | `MonthDay.parse(String)`
 |java.util.UUID|`toString()` as a `java.lang.String` | `UUID.fromString(String)`
 |java.util.Calendar|A `java.time.ZonedDateTime` representing the same date/time and timezone; see <<mapper-orm-legacy-date-time-apis>> | `ZonedDateTime.parse(String)`
-|java.util.Date|`toInstant()` as a `java.time.Instant`; see <<mapper-orm-legacy-date-time-apis>> | `Instant.parse(String)`
-|java.sql.Timestamp|`toInstant()` as a `java.time.Instant`; see <<mapper-orm-legacy-date-time-apis>> | `Instant.parse(String)`
-|java.sql.Date|`toInstant()` as a `java.time.Instant`; see <<mapper-orm-legacy-date-time-apis>> | `Instant.parse(String)`
-|java.sql.Time|`toInstant()` as a `java.time.Instant`; see <<mapper-orm-legacy-date-time-apis>> | `Instant.parse(String)`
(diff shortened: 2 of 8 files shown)


mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/bridge/builtin/impl/DefaultJavaUtilDateValueBridge.java
```diff
@@ -23,7 +23,13 @@ public String toString() {
...
 	@Override
 	public Instant toIndexedValue(Date value, ValueBridgeToIndexedValueContext context) {
-		return value == null ? null : value.toInstant();
+		if ( value == null ) {
+			return null;
+		}
+
+		// java.sql.Date does not support toInstant(). See HSEARCH-3670.
+		// We're using here the same strategy applied on DefaultJavaSqlDateValueBridge to handle them.
+		return value instanceof java.sql.Date ? Instant.ofEpochMilli( value.getTime() ) : value.toInstant();
```
yrodiere: Couldn't we always use `Instant.ofEpochMilli( value.getTime() )`? That would spare us one `instanceof`.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/bridge/builtin/impl/DefaultJavaUtilDateValueBridge.java
```diff
@@ -27,9 +27,10 @@ public Instant toIndexedValue(Date value, ValueBridgeToIndexedValueContext conte
 			return null;
 		}
 
-		// java.sql.Date does not support toInstant(). See HSEARCH-3670.
-		// We're using here the same strategy applied on DefaultJavaSqlDateValueBridge to handle them.
-		return value instanceof java.sql.Date ? Instant.ofEpochMilli( value.getTime() ) : value.toInstant();
+		// java.sql.Date and java.sql.Time do not support toInstant(). See HSEARCH-3670.
+		// We're using here the same strategy applied on DefaultJavaSql*ValueBridge to handle them.
+		return ( value instanceof java.sql.Date ) | ( value instanceof java.sql.Time ) ?
```
yrodiere: You used a bitwise `OR` here (single pipe). Please avoid that.

Also, same comment as before: you could just always use `Instant.ofEpochMilli( value.getTime() )`.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/bridge/builtin/impl/DefaultJavaUtilDateValueBridge.java
```diff
@@ -27,9 +27,9 @@ public Instant toIndexedValue(Date value, ValueBridgeToIndexedValueContext conte
 			return null;
 		}
 
-		// java.sql.Date and java.sql.Time do not support toInstant(). See HSEARCH-3670.
+		// java.sql.* types do not support toInstant(). See HSEARCH-3670.
 		// We're using here the same strategy applied on DefaultJavaSql*ValueBridge to handle them.
-		return ( value instanceof java.sql.Date ) | ( value instanceof java.sql.Time ) ?
+		return ( value instanceof java.sql.Date ) || ( value instanceof java.sql.Time ) || ( value instanceof java.sql.Timestamp ) ?
```
yrodiere: Same comment as before about always using `Instant.ofEpochMilli( value.getTime() )`.

Also, you didn't test `Timestamp`?

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/bridge/builtin/impl/DefaultJavaUtilDateValueBridge.java
fax4ever: I would have used only as workaround for `java.sql` legacy types. But actually `java.util.Date` is legacy too. Done

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/bridge/builtin/impl/DefaultJavaUtilDateValueBridge.java
fax4ever: It seems that it is not possible to add a test for a timestamp without a significant change. 

The `FieldDefaultBridgeIT#projection` method test would fail, since `DefaultJavaUtilDateValueBridge#fromIndexedValue` method cannot distinguish the cases, it converts the indexed `Instant` value to a `util.Date` instance and, unlike `java.sql.Date` and  `java.sql.Time`, the equals  of `Timestamp` check the instance type.

Does such a test worth the effort to handle it?

fax4ever:
Thanks @yrodiere for reviewing it.
I think I've addressed the commit.
Is there anything else we can do here?

fax4ever:
I forgot to update the doc. I've done with the last commit.

yrodiere:
Thanks. I pushed a test for `java.sql.Timestamp`. Let's merge if the build passes.

fax4ever:
Nice to have it. Thanks

PR #2095, opened 2019-09-17
HSEARCH-3708 Simplify the syntax of the range predicate DSL by leveraging the Range class

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3708

The main purpose is to remove the gazillion interfaces that allow to write `.from(..).excludeLimit().to(...).excludeLimit()` or `.from(..).excludeLimit().to(...)` and then chain with other calls, because if we leave them, solving HSEARCH-3687 will be a nightmare.

But I think the new APIs are sane, and maybe even better, considering they are more consistent with the syntax introduced for the range aggregation DSL.


integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/bool/BooleanSortAndRangePredicateIT.java
```diff
@@ -40,7 +40,7 @@
  * Tests sorting and ranging behaviour querying a boolean type field
  *
  * @see FieldSearchSortIT
- * @see RangeSearchPredicateIT
+ * @see RangeSearchPredicateLegacySyntaxIT
```
fax4ever: Since this class is using the new API, maybe it should keep to reference `RangeSearchPredicateIT` and not the legacy one.

yrodiere:
Addressed your comment and merged. Thanks!

PR #2096, opened 2019-09-18
HSEARCH-3687 Add self types to DSL interfaces to make overrides by backends easier

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3687

~Based on #2095 , which should be merged first.~ => Done

Note this does not add any feature. This is merely an early API change which will hopefully spare us breaking changes when we do need to add new features.


yrodiere:
Thanks! I merged the other PR and rebased this one on master. Waiting for CI.

PR #2097, opened 2019-09-18
HSEARCH-3691 Remove the ability to use @GeoPointBinding on properties of type GeoPoint

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3691

Documentation seems to be already updated.

IMHO we don't need to test that having a such binding does not work anymore. Usually I'm a big fan of having a test for everything, but not in this case (we have removed some code, not added it).
Anyway, we could add it anytime.


fax4ever:
Merged thanks!

PR #2104, opened 2019-09-26
HSEARCH-3684 @IndexedEmbedded.includePaths includes fields one level too deep in some cases

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3684

This fixes the problem with a rather extreme solution (see commit " HSEARCH-3684 Fix deeply nested indexed-embedded fields being incorrectly included"), but frankly after a few days on this I've wasted enough time.

In short, instead of pre-computing the set of allowed paths each time we nest an `@IndexedEmbedded`, we just remember each `@IndexedEmbedded` exactly as defined, and when determining whether a path is included or not, we will dynamically "ask" the whole `@IndexedEmbedded` chain.

It's a bit less efficient, but it's just bootstrap anyway, and it's much easier to debug (you can "see" exactly which indexed-embedded caused a path to be included/excluded). More importantly, it's the only solution I could find.

Note I had to make some additional changes to how encountered paths are tracked. If you're interested in "why", try to run `IndexedEmbeddedIncludePathsAndDepthIT` before and after these fixes. But it's... complex.


mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/building/impl/PojoMapper.java
```diff
@@ -172,35 +179,41 @@ private void prepareEntityOrIndexedType(PojoRawTypeModel<?> rawTypeModel,
...
+				contextProvider.getIndexManagerBuildingState(
 						indexedTypeMetadata.getBackendName(),
 						indexedTypeMetadata.getIndexName()
 								.orElse( entityTypeMetadata.getEntityName() ),
 						multiTenancyEnabled
 				);
-
+		IndexedEntityBindingContext bindingContext = contextProvider.createIndexedEntityBindingContext(
+				this, indexManagerBuildingState.getSchemaRootNodeBuilder()
+		);
```
fax4ever: IMHO, collaborations here have become too complex. The readability is not the very good.
I think that is very important having code that is very understandable.

If the goal here is to create an instance of `PojoIndexedTypeManagerBuilder`, maybe it is better to do all the job in `indexManagerBuildingState`. 
In general, I like to have responsible objects, instead of having object providers which unique responsibility is to provide other instances.
As a rule of thumb, try to write sequence diagram. These never lie. If these are simple, collaborations are good.

Do you think we could simplify a bit the collaborations here? Thanks

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/spi/IndexedEntityBindingContextProvider.java
```diff
@@ -8,9 +8,15 @@
 
 import java.util.Optional;
 
-public interface IndexManagerBuildingStateProvider {
+import org.hibernate.search.engine.backend.document.model.dsl.spi.IndexSchemaRootNodeBuilder;
+
+public interface IndexedEntityBindingContextProvider {
```
fax4ever: This is continuing to be a provider of `IndexManagerBuildingState`, why shouldn't it be reflected in its name? But I think that the naming problem here is the reflex of a design issue. See the other comment.

mapper/pojo/src/main/java/org/hibernate/search/mapper/pojo/mapping/building/impl/PojoMapper.java
yrodiere: > If the goal here is to create an instance of PojoIndexedTypeManagerBuilder, maybe it is better to do all the job in indexManagerBuildingState.

Well as the names indicate, `PojoIndexedTypeManagerBuilder` is implemented in the POJO module. `indexManagerBuildingState`, on the other hand, is implemented in the engine and mainly delegates to the backends.
There are multiple components at work here, all of which can have multiple implementations, and each requiring a different execution context. So I cannot just stuff everything into a single component to "simplify" collaboration.

I agree this code is complex. It is probably *too* complex for what it does.
I'm not sure what you would have me do about that, though.
I'll try to discuss this with you over chat.

yrodiere: @fax4ever I pushed a few commits that replace the `IndexedEntityBindingContextProvider` with a `MappedIndexManagerFactory`. Overall it's more consistent, and a bit better than it was before this PR, which should be good enough considering improving the overall design of the bootstrap is out of scope.

In short, this:

```java
		PojoTypeAdditionalMetadata metadata = typeAdditionalMetadataProvider.get( indexedEntityType );
		// This metadata is guaranteed to exist; see prepareEntityOrIndexedType()
		PojoIndexedTypeAdditionalMetadata indexedTypeMetadata = metadata.getIndexedTypeMetadata().get();
		PojoEntityTypeAdditionalMetadata entityTypeMetadata = metadata.getEntityTypeMetadata().get();

		IndexManagerBuildingState<?> indexManagerBuildingState =
				contextProvider.getIndexManagerBuildingState(
						indexedTypeMetadata.getBackendName(),
						indexedTypeMetadata.getIndexName()
								.orElse( entityTypeMetadata.getEntityName() ),
						multiTenancyEnabled
				);
		IndexedEntityBindingContext bindingContext = contextProvider.createIndexedEntityBindingContext(
				this, indexManagerBuildingState.getSchemaRootNodeBuilder()
		);
		PojoIndexedTypeManagerBuilder<?, ?> builder = createIndexedTypeManagerBuilder(
				indexedEntityType, indexManagerBuildingState, bindingContext
		);
		// Put the builder in the map before anything else, so it will be closed on error
		indexedTypeManagerBuilders.put( indexedEntityType, builder );

		PojoMappingCollectorTypeNode collector = builder.asCollector();

		contributorProvider.get( indexedEntityType )
				.forEach( c -> c.contributeMapping( collector ) );
```

... became this:

```java
	private void mapIndexedType(PojoRawTypeModel<?> indexedEntityType,
			MappedIndexManagerFactory indexManagerFactory) {
		PojoTypeAdditionalMetadata metadata = typeAdditionalMetadataProvider.get( indexedEntityType );
		// This metadata is guaranteed to exist; see prepareEntityOrIndexedType()
		PojoIndexedTypeAdditionalMetadata indexedTypeMetadata = metadata.getIndexedTypeMetadata().get();
		PojoEntityTypeAdditionalMetadata entityTypeMetadata = metadata.getEntityTypeMetadata().get();

		MappedIndexManagerBuilder<?> indexManagerBuilder = indexManagerFactory.createMappedIndexManager(
				this,
				indexedTypeMetadata.getBackendName(),
				indexedTypeMetadata.getIndexName()
						.orElse( entityTypeMetadata.getEntityName() ),
				multiTenancyEnabled
		);
		PojoIndexedTypeManagerBuilder<?, ?> builder = createIndexedTypeManagerBuilder(
				indexedEntityType, indexManagerBuilder
		);
		// Put the builder in the map before anything else, so it will be closed on error
		indexedTypeManagerBuilders.put( indexedEntityType, builder );

		PojoMappingCollectorTypeNode collector = builder.asCollector();

		contributorProvider.get( indexedEntityType )
				.forEach( c -> c.contributeMapping( collector ) );
	}
```

So `IndexManagerBuildingState` and `IndexedEntityBindingContext` were merged into a single component, `MappedIndexManagerBuilder`. The `PojoIndexedTypeManagerBuilder` will have to collaborate with this component to create a type manager.

Does this look good (enough) to you?

fax4ever: Thank you @yrodiere. Let me check...

fax4ever: IMHO, it looks much better now.

yrodiere:
Merged, thanks!

fax4ever:
Thank you!

PR #2105, opened 2019-09-26
HSEARCH-3640 Support get indexes/backend from mapper orm

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3640

### DIFF
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/mapping/SearchMappingIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/mapping/SearchMappingIT.java
@@ -1,0 +1,120 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.integrationtest.mapper.orm.mapping;
+
+import javax.persistence.Entity;
+import javax.persistence.Id;
+
+import org.hibernate.SessionFactory;
+import org.hibernate.search.engine.backend.index.IndexManager;
+import org.hibernate.search.mapper.orm.Search;
+import org.hibernate.search.mapper.orm.mapping.SearchMapping;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.GenericField;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.Indexed;
+import org.hibernate.search.util.impl.integrationtest.common.rule.BackendMock;
+import org.hibernate.search.util.impl.integrationtest.orm.OrmSetupHelper;
+
+import org.hibernate.testing.TestForIssue;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import org.assertj.core.api.Assertions;
+
+@TestForIssue( jiraKey = "HSEARCH-3640" )
+public class SearchMappingIT {
+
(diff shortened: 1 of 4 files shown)


integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/mapping/SearchMappingIT.java
```diff
@@ -0,0 +1,120 @@
...
+	@Test
+	public void getBackendByName() {
+		SearchMapping target = Search.mapping( sessionFactory );
+
+		Object backend = target.getBackend( BACKEND_1_NAME );
+		Assertions.assertThat( backend ).isNotNull();
+
+		// TODO HSEARCH-3640
+		//		backend = target.getBackend( BACKEND_2_NAME );
+		//		Assertions.assertThat( backend ).isNotNull();
```
fax4ever: Preview because it doesn't work yet

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/mapping/SearchMappingIT.java
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	@Entity
```
yrodiere: You should set the entity name too, otherwise it will be set to `SearchMappingIT$Person` or something similar, and depending on the name of the test class, this might be too long for some databases (Oracle has a particularly low limit, IIRC).
Generally I try to simply define a `NAME` constant and use that for both the entity name and index name. Or just the entity name, since the index name will default to that.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/mapping/SearchMappingIT.java
yrodiere: You have to target these backends from the mapper. Unused backends are not created.

Either:
*  use `@Indexed(backend = ...)` in both entities.
* or rely on `@Indexed(backend = ...)` in one entity and on the default backend in the other (the default backend is the first one passed to `OrmSetupHelper.withBackendMocks`).

 See `org.hibernate.search.integrationtest.mapper.orm.session.SearchIndexingPlanBaseIT.IndexedEntity1`.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/mapping/SearchMappingIT.java
fax4ever: I would instead use a different name, such as `PersonEntity`. To mark the fact that for this entity the index name has been taken from the @Index annotation. WDYT?

fax4ever:
Thanks @yrodiere. I addressed the comments.
Anything else can we improve here before we merge it?

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/mapping/SearchMappingIT.java
fax4ever: I moved the Pet entity to the second backend. It worked!

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/mapping/SearchMappingIT.java
yrodiere: I think that's not the purpose of this test and we should keep things simple when they are not relevant. But it's not very important.

yrodiere:
Merged, thanks!

fax4ever:
Thanks!

PR #2111, opened 2019-10-03
HSEARCH-3725 IllegalStateException thrown by Gson while encoding Elasticsearch requests

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3725

After this fix, running the reproducer more than 1000 times didn't show the bug. But then last time I had to run it 825 times, so... I guess we'll just have to content with the fact that the workaround is effective *in theory*.


fax4ever:
@yrodiere I started [a new build](https://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-2111/2/) with all supported versions.

fax4ever:
Merged thanks!

yrodiere:
Thanks!

PR #2112, opened 2019-10-04
HSEARCH-3727 Elasticsearch orchestrators do not wait for single-work worksets to finish with ElasticsearchParallelWorkProcessor

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3727

Mainly useful for HSEARCH-3110 (restore support for ErrorHandler).


fax4ever:
Good catch!
It was odd that the behavior was different for a single element workset.
Merged thanks!

PR #2115, opened 2019-10-09
HSEARCH-3733 BatchExecutor sometimes forgets to schedule the processor since HSEARCH-3084

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3733

This should fix the problem where integration tests are stuck in an infinite wait.

I couldn't find a way to write a non-regression test, but locally executing `FieldSearchSortIT` with the Elasticsearch backend, execution hanged after ~200 executions before the batch, and not at all even after ~1000 executions after the patch.


yrodiere:
These stuck builds are really too annoying, I'll merge directly to avoid more problems on CI.

PR #2116, opened 2019-10-09
HSEARCH-3734 Configuration property hibernate.search.reflection.strategy is ignored

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3734

Turns out we were passing a `propertySource` that was not masked ("hibernate.search." removed) to the introspector, which expected it to be masked. This PR solves the problem.

Not adding tests because I simply don't know how to check that method handles are not used.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/common/impl/SearchIntegrationBuilderImpl.java
+++ b/engine/src/main/java/org/hibernate/search/engine/common/impl/SearchIntegrationBuilderImpl.java
@@ -85,6 +85,11 @@ public class SearchIntegrationBuilderImpl implements SearchIntegrationBuilder {
 	}
 
 	@Override
+	public ConfigurationPropertySource getMaskedPropertySource() {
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/hibernate/search/engine/common/impl/SearchIntegrationBuilderImpl.java
```diff
@@ -84,6 +84,11 @@ public SearchIntegrationBuilderImpl(ConfigurationPropertySource propertySource,
 		propertyChecker.beforeBoot();
 	}
 
+	@Override
+	public ConfigurationPropertySource getMaskedPropertySource() {
+		return propertySource;
```
fax4ever: Nice to reuse the field `SearchIntegrationBuilderImpl#propertySource`.

yrodiere:
Merged, thanks

PR #2117, opened 2019-10-11
HSEARCH-3736 Lucene startup creates an unnecessary IndexWriter

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3736

By not setting `hasUncommittedWorks = true` when we ensure the index exists, we skip the end-of-batch commit and thus avoid the creation of an IndexWriter.


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/orchestration/impl/LuceneWriteWorkProcessor.java
```diff
@@ -81,6 +81,20 @@ void beforeWorkSet(DocumentCommitStrategy commitStrategy, DocumentRefreshStrateg
 				|| DocumentRefreshStrategy.FORCE.equals( refreshStrategy );
 	}
 
+	void ensureIndexExists() {
```
fax4ever: I would add a comment over here, explaining the reason we need this method bypassing the normal submit. That would be what you explain in the pull request description, ie the method do not set hasUncommittedWorks to true, in order to not provoke a commit.

yrodiere:
Addressed your comment and merged. Thanks!

fax4ever:
I read the comment it is very clear. Thank you

PR #2120, opened 2019-10-11
HSEARCH-3724 Upgrade to Jackson 2.9.10

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3724


yrodiere:
The CI build passed: merging.

PR #2121, opened 2019-10-11
HSEARCH-1353 + HSEARCH-1375 + HSEARCH-3110 Restore support for ErrorHandler (now FailureHandler)

yrodiere:
* [HSEARCH-1353](https://hibernate.atlassian.net/browse/HSEARCH-1353): Document the expected behaviour of ErrorHandler
* [HSEARCH-1375](https://hibernate.atlassian.net/browse/HSEARCH-1375): Improve design of ErrorHandler
* [HSEARCH-3110](https://hibernate.atlassian.net/browse/HSEARCH-3110): Restore support for ErrorHandler

~Based on #2117 ([HSEARCH-3736](https://hibernate.atlassian.net/browse/HSEARCH-3736)), which should be merged first.~ => Done

This moves the ErrorHandler to API, renames it to FailureHandler (see commit messages for an explanation), and brings a few changes aimed at improving the API, as well as removing a few bugs.

There are still a few issues which will be addressed in follow-up tickets; see the tickets linked from [HSEARCH-1375](https://hibernate.atlassian.net/browse/HSEARCH-1375).


documentation/src/test/java/org/hibernate/search/documentation/reporting/failurehandler/MyFailureHandler.java
```diff
@@ -0,0 +1,57 @@
...
+	public void handle(FailureContext context) { // <1>
+		try {
+			String failingOperationDescription = context.getFailingOperation().toString(); // <2>
+			Throwable throwable = context.getThrowable(); // <3>
+
+			// ... report the failure ... // <4>
+		}
+		catch (Throwable t) {
+			// ... handle the failure and do *not* propagate the throwable ... // <5>
+		}
```
fax4ever: Why leave the burden of capturing the exception to the user?

documentation/src/main/asciidoc/configuration.asciidoc
```diff
@@ -237,3 +237,54 @@ for beans injected into Hibernate Search.
...
+but in some cases, such as Lucene segment merging failures,
+the exception in background threads cannot be propagated.
+By default, when that happens, the failure is logged at the `ERROR` level.
+
+To customize background failure handling, you will need to:
+
+* Define a class that implements the `org.hibernate.search.engine.reporting.FailureHandler` interface.
+* Configure the backend to use that implementation by setting the configuration property
+`hibernate.search.background_failure_handler`
+to a <<configuration-property-types,bean reference>> pointing to the implementation.
```
fax4ever: Are we sure that `Type of configuration properties` is the right chapter to link here?

engine/src/main/java/org/hibernate/search/engine/reporting/IndexFailureContext.java
```diff
@@ -0,0 +1,24 @@
...
+public interface IndexFailureContext extends FailureContext {
+
+	/**
+	 * @return The list of index operations that weren't committed yet when the failure occurred.
+	 * Never {@code null}, but may be empty.
+	 * These operations may not have been applied to the index.
+	 * Use {@link Object#toString()} to get a textual representation of each operation.
+	 */
+	List<?> getUncommittedOperations();
+
```
fax4ever: why do we not add a method to get the index name?

documentation/src/main/asciidoc/configuration.asciidoc
yrodiere: Yes, it explains what a bean reference is exactly.

documentation/src/test/java/org/hibernate/search/documentation/reporting/failurehandler/MyFailureHandler.java
yrodiere: Yes, I suppose I could wrap the handler with a "safe" one we implement ourselves. I'll do that.

engine/src/main/java/org/hibernate/search/engine/reporting/IndexFailureContext.java
yrodiere: Mostly because I can't get it easily in the case of Elasticsearch... I'll see what I can do.

yrodiere: Created https://hibernate.atlassian.net/browse/HSEARCH-3741 , because it will be easier to address after some of the pending PRs have been merged.

yrodiere:
Addressed your comments and merged, thanks!

PR #2122, opened 2019-10-11
HSEARCH-3726 Exceptions are not propagated from the Elasticsearch backend to the mapper when using IndexingPlan or automatic indexing

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3726

~Based on #2121 , which should be merged first.~ => Done

### DIFF
--- a/backend/elasticsearch/src/test/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/ElasticsearchSerialWorkProcessorTest.java
+++ b/backend/elasticsearch/src/test/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/ElasticsearchSerialWorkProcessorTest.java
@@ -22,6 +22,7 @@ import org.hibernate.search.backend.elasticsearch.work.impl.ElasticsearchWorkAgg
 import org.junit.Before;
 import org.junit.Test;
 
+import org.easymock.EasyMock;
 import org.easymock.EasyMockSupport;
 import org.easymock.IAnswer;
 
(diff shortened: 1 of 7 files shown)


backend/elasticsearch/src/test/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/ElasticsearchSerialWorkProcessorTest.java
```diff
@@ -110,7 +111,51 @@ public void simple_multipleWorksInWorkSet() {
...
+		assertThat( futureAll ).isSuccessful();
+	}
+
+	@Test
+	public void simple_sequenceFailure() {
+		ElasticsearchWork<Object> work = work( 1 );
+
+		CompletableFuture<Void> sequenceFuture = new CompletableFuture<>();
+
+		replayAll();
```
fax4ever: I think there is no mock to replay. Maybe we can remove it.

backend/elasticsearch/src/test/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/ElasticsearchSerialWorkProcessorTest.java
```diff
@@ -110,7 +111,51 @@ public void simple_multipleWorksInWorkSet() {
...
+		ElasticsearchSerialWorkProcessor processor =
+				new ElasticsearchSerialWorkProcessor( sequenceBuilderMock, bulkerMock );
+		verifyAll();
```
fax4ever: I think there is no mock to verify. Maybe we can remove it.

yrodiere: There are mocks involved just below. Calling `replayAll` and `verifyAll` allows us to check that no method is called on those mocks.
Even if we didn't care, mocks simply shouldn't be used by the processor in their default "record" state: at best it won't do anything, at worst it will mess with the expectations of the next assertions.

yrodiere:
Merged, thanks!

PR #2123, opened 2019-10-15
HSEARCH-3735 Report failures in work executions only once, at the mapper level

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3735

In short this makes sure that exceptions occurring during indexing are propagated from the backend to the mapper using `CompletableFutures`, and that the mapper will properly report the failures with the appropriate context (entity reference, in particular).

~Based on #2121 and #2122 , which should be merged first.~ => Done


documentation/src/main/asciidoc/mapper-orm-indexing-automatic.asciidoc
```diff
@@ -120,13 +120,18 @@ i.e. to control the minimum progress of indexing before the application thread i
...
-* when set to `queued`, the application thread will be resumed as soon as
+* [[mapper-orm-indexing-automatic-synchronization-queued]] when set to `queued`, the application thread will be resumed as soon as
 the index changes are queued in the backend.
 +
 This strategy offers no guarantee as to whether indexing will be performed successfully,
 or even whether indexing will be performed at all:
 the local JVM may crash before the works are executed, in which case the indexing requests will be forgotten,
 or indexing may simply fail.
++
+With this strategy, failures to extract data from entities will will lead to an exception being thrown in the application thread,
```
fax4ever: will 2 times

yrodiere:
Addressed your comments and merged. Thanks!

PR #2124, opened 2019-10-15
HSEARCH-3721 Remove deprecated methods that were already present in 6.0.0.Beta1

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3721


yrodiere:
Merged, thanks

PR #2125, opened 2019-10-16
HSEARCH-3739 Test and fix MassIndexer interruption/cancellation

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3739

Based on #2123 , which should be merged first.


integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingFailureIT.java
```diff
@@ -420,23 +456,48 @@ private void doMassIndexing(SessionFactory sessionFactory,
...
 		}
+		catch (AssertionError e) {
+			assertionError = e;
+			throw e;
+		}
 		finally {
 			Book.failOnBook2GetId.set( false );
 			Book.failOnBook2GetTitle.set( false );
+
+			if ( assertionError == null ) {
```
fax4ever: I don't think `assertionError` may be different from null here. assertionError could have been changed only into the catch clause and the latter throws an exception. Maybe we can remove the check.

util/common/src/main/java/org/hibernate/search/util/common/impl/ExecutorCompletableFuture.java
```diff
@@ -0,0 +1,49 @@
...
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.util.common.impl;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+
+class ExecutorCompletableFuture<T> extends CompletableFuture<T> {
```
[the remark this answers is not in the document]
yrodiere: Alright, renamed to `CancelableExecutionCompletableFuture`.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingFailureIT.java
yrodiere: The finally block gets executed whatever happens in catch blocks, even if they re-throw an exception. This `assertionError` may be null.

PR #2143, opened 2019-11-05
HSEARCH-3755 Configuration property checking displays the wrong name for the property controlling its behavior

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3755


fax4ever:
Merged thanks

PR #2144, opened 2019-11-06
Support "exists" predicate for @IndexedEmbedded

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-2389

~~Preview, need to check incompatibility model of multi-indexes.~~ done


fax4ever:
I handled one alert of LGTM: the one about Dereferenced variable may be null.
I left the other since the boxed variable is used in the equals.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/ExistsPredicateIT.java
```diff
@@ -4,7 +4,7 @@
  * License: GNU Lesser General Public License (LGPL), version 2.1 or later
  * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
  */
-package org.hibernate.search.integrationtest.backend.elasticsearch.tmp;
+package org.hibernate.search.integrationtest.backend.tck.search.predicate;
```
yrodiere: So now we have an `ExistsPredicateIT` and an `ExistsSearchPredicateIT`... That's messy :/

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/ExistsPredicateIT.java
```diff
@@ -30,6 +34,8 @@
 public class ExistsPredicateIT {
 
 	private static final String INDEX_NAME = "IndexName";
+	private static final String COMPATIBLE_INDEX_NAME = "CompatibleIndexName";
+	private static final String INCOMPATIBLE_INDEX_NAME = "IncompatibleIndexName";
```
yrodiere: There are many ways for a field to have an incompatible definition in another index:

* object on one side, non-object on the other
* nested on one side, flattened on the other
* flattened with a given set of children on one side, object with a different set of children on the other
* flattened with same children on both side BUT the children themselves have incompatible exists predicates (e.g. doc values are enabled on one side but not the other).

I think that you should have at least one test for each of the cases above, even if you don't test all the possible combinations for the last case.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
```diff
@@ -130,15 +130,32 @@ public EventContext getIndexesEventContext() {
...
-			if ( objectNode != null ) {
-				return true;
+			Boolean hasObjectNode = ( objectNode != null );
+
+			if ( result == null ) {
+				result = hasObjectNode;
+				selectedIndexName = indexName;
+				selectedObjectNode = objectNode;
+			}
+			else if ( !result.equals( hasObjectNode ) ) {
```
yrodiere: So you're throwing an exception in the following cases:

1. index A has an object field, index B has a non-object field
2. index A has an object field, index B doesn't have any field with this name (object or non-object).

I agree with 1, but 2 should not throw an exception.

You probably need more tests?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/scope/model/impl/LuceneScopeModel.java
```diff
@@ -76,26 +76,44 @@ public EventContext getIndexesEventContext() {
...
 		for ( LuceneIndexModel indexModel : indexModels ) {
-			// TODO HSEARCH-2389 check multi-indexes model incompatibility
+			String indexName = indexModel.getIndexName();
+
 			LuceneIndexSchemaObjectNode objectNode = indexModel.getObjectNode( absoluteFieldPath );
+
+			if ( selectedIndexName != null ) {
+				if ( ( selectedObjectNode == null && objectNode != null ) ||
+						( selectedObjectNode != null && objectNode == null ) ) {
+					throw log.conflictingObjectFieldModel( absoluteFieldPath,
```
yrodiere: Same problem as for Elasticsearch: you will throw an exception when one of the indexes simply doesn't declare the field, but you shouldn't.

Also, and it's a problem in the case of Lucene, you are completely ignoring two types of incompatibilities:

* the object nodes are stored differently (nested vs. flattened)
* the object nodes have a different set of children

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/tmp/ExistsPredicateIT.java
```diff
@@ -0,0 +1,228 @@
...
+
+	// this document has only first level fields: string and numeric
+	private static final String DOCUMENT_1 = "1";
+
+	// this document has also an empty nested and an empty flattened objects
+	private static final String DOCUMENT_2 = "2";
+
+	// this document has also second level fields:
+	// string and numeric within both nested and flattened objects
+	private static final String DOCUMENT_3 = "3";
```
yrodiere: I think there's another interesting case that you forgot: a document with an object field whose only content is another object field, which contains a string field with a value.

E.g.:

```
{
   "nested": {
      "nested2": {
         "string": "foo"
      }
   }
}
```

What if we use the `exists` predicate on `nested` with this data? The document should match, but I'm not sure it would with the current implementation.

We should also test this with a flattened object field.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/document/model/impl/LuceneIndexSchemaObjectNode.java
```diff
@@ -25,17 +30,22 @@ public static LuceneIndexSchemaObjectNode root() {
 
 	private final String nestedDocumentPath;
 
+	private final List<String> childrenPaths;
```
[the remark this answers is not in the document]
fax4ever: Right. Originally they were relative path. I forgot to change the name.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/scope/model/impl/LuceneScopeModel.java
```diff
@@ -74,6 +74,30 @@ public EventContext getIndexesEventContext() {
...
+		for ( LuceneIndexModel indexModel : indexModels ) {
+			// TODO HSEARCH-2389 check multi-indexes model incompatibility
+			LuceneIndexSchemaObjectNode objectNode = indexModel.getObjectNode( absoluteFieldPath );
+			if ( objectNode == null ) {
+				continue;
+			}
+
+			LuceneScopedIndexObjectComponent<T> scopedIndexObjectComponent = new LuceneScopedIndexObjectComponent<>();
+			for ( String childPath : objectNode.getChildrenPaths() ) {
+				LuceneIndexSchemaFieldNode<?> schemaNode = indexModel.getFieldNode( childPath );
```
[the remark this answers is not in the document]
fax4ever: You definitely right. In case of exists targeting an object field containing another object, it will load a null schemaNode I think.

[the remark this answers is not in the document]
fax4ever: Instead it seems it works, because we have a null check block just above this code line. Anyway I updated the test case.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/scope/model/impl/LuceneScopeModel.java
```diff
@@ -74,6 +74,30 @@ public EventContext getIndexesEventContext() {
...
+				if ( schemaNode == null ) {
+					continue;
+				}
+
+				T component = componentRetrievalStrategy.extractComponent( schemaNode );
+				scopedIndexObjectComponent.addFieldComponent( childPath, component );
+			}
+			return scopedIndexObjectComponent;
```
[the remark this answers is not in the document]
fax4ever: Yes compatibility checks are handled in subsequent commits.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/predicate/impl/LuceneExistsCompositePredicateBuilder.java
```diff
@@ -0,0 +1,55 @@
...
+
+	@Override
+	public Query build(LuceneSearchPredicateContext context) {
+		// if exists at least one not-null field, exists on object field should match
+		BooleanQuery.Builder builder = new BooleanQuery.Builder();
+		for ( LuceneSearchPredicateBuilder child : children ) {
+			builder.add( child.build( context ), BooleanClause.Occur.SHOULD );
+		}
+		builder.setMinimumNumberShouldMatch( 1 );
+		return builder.build();
```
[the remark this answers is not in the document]
fax4ever: In this case we should have a Boolean query with no child, I guess it could return false. Anyway let's test it.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/tmp/ExistsPredicateIT.java
fax4ever: I think you're right: current implementation does not. Let's handle the recursive case.

fax4ever:
Added all the tests. Currently they do not work, let me know if this is the expected behavior.

fax4ever:
Handled compatibility check for Lucene. I'm working on Elasticsearch.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/ObjectExistsSearchPredicateIT.java
```diff
@@ -227,13 +253,15 @@ private void checkDocumentsCreation() {
 			nested = nestedObject.toReference();
 			nestedString = nestedObject.field( "string", f -> f.asString() ).toReference();
 			nestedNumeric = nestedObject.field( "numeric", f -> f.asInteger() ).toReference();
-			nestedX2 = nestedObject.objectField( "nestedX2" ).toReference();
+			nestedX2 = nestedObject.objectField( "nestedX2", ObjectFieldStorage.NESTED ).toReference();
+			nestedNoChild = root.objectField( "nestedNoChild", ObjectFieldStorage.NESTED ).toReference();
```
yrodiere: Shouldn't you add a value for this object in `initData` in order for the test to make sense?

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/ObjectExistsSearchPredicateIT.java
```diff
@@ -115,6 +119,7 @@ public void nested() {
 				.fetchAllHits();
 
 		// DOCUMENT_2 won't be matched either, since it hasn't any not-null field
+		// DOCUMENT_4 won't be matched either, since we use a nested storage for the inner object field
```
yrodiere: Even if I understand why it works that way, it's arguably more a bug than a feature. We should automatically add a join query ("nested" predicate) in that case. Maybe create a ticket related to the one that you created recently about implicit "nested" predicates, but specifically for the "exists" predicate? We may need specific handling.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/ObjectExistsSearchPredicateIT.java
```diff
@@ -210,7 +217,8 @@ public void flattened() {
 				.fetchAllHits();
 
 		// DOCUMENT_2 won't be matched either, since it hasn't any not-null field
-		assertThat( docs ).hasDocRefHitsAnyOrder( INDEX_NAME, DOCUMENT_3 );
+		// DOCUMENT_4 will match, even if the matching field is not a direct field of the targeted path
```
yrodiere: `direct field` => `direct child`? Not sure I understood correctly, though.

Same below.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/ObjectExistsSearchPredicateIT.java
```diff
@@ -448,4 +488,26 @@ private void checkDocumentsCreation() {
...
+			nestedObject.field( "string", f -> f.asBigDecimal().decimalScale( 3 ) ).toReference();
+			nestedObject.field( "numeric", f -> f.asInteger() ).toReference();
+
+			nestedObject.objectField( "nestedX2", ObjectFieldStorage.NESTED ).toReference();
+
+			IndexSchemaObjectField flattenedObject = root.objectField( "flattened", ObjectFieldStorage.FLATTENED );
+			flattenedObject.toReference();
+
+			flattenedObject.field( "string", f -> f.asString() ).toReference();
+			// field has same name, but with an incompatible exists predicates: unSortable vs Sortable
```
yrodiere: Great idea.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/ObjectExistsSearchPredicateIT.java
fax4ever: You're right. Done

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/ObjectExistsSearchPredicateIT.java
fax4ever: yeap. it sounds better

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/ObjectExistsSearchPredicateIT.java
fax4ever: thanks

fax4ever:
Addressed the comments so far. Is there anything else we can improve here? Thanks

fax4ever:
I've started the [build 12](https://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-2144/12/) testing all supported environments.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/ObjectExistsSearchPredicateIT.java
yrodiere: I couldn't find the ticket?

fax4ever: Sorry I didn't do that. I did now: https://hibernate.atlassian.net/browse/HSEARCH-3762.

fax4ever:
Merged thanks!

PR #2147, opened 2019-11-08
HSEARCH-3662 Define aggregations from JSON with Elasticsearch

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3662


yrodiere:
Addressed the double blank line issue and merged. Thanks!

PR #2150, opened 2019-11-13
HSEARCH-3761 Use GSON rather than strings to pass JSON objects around in Elasticsearch APIs

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3761


yrodiere:
Merged, thanks

PR #2151, opened 2019-11-13
HSEARCH-3496 Customization of Elasticsearch search requests

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3496

Based on #2150 ([HSEARCH-3761](https://hibernate.atlassian.net/browse/HSEARCH-3761)), which should be merged first.


integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/search/query/ElasticsearchSearchQueryRequestTransformerIT.java
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	@Test
+	public void path() {
+		StubMappingScope scope = indexManager.createScope();
+
+		SearchQuery<DocumentReference> query = scope.query().extension( ElasticsearchExtension.get() )
+				.predicate( f -> f.matchAll() )
+				.requestTransformer( context -> {
+					assertThat( context.getPath() ).isEqualTo( "/" + INDEX_NAME.toLowerCase( Locale.ROOT ) + "/_search" );
```
fax4ever: It seems that constant `INDEX_NAME` is already lowercase.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/ElasticsearchExtensionIT.java
```diff
@@ -167,6 +174,10 @@ public void query() {
 				.hasDocRefHitsAnyOrder( INDEX_NAME, FIRST_ID, SECOND_ID, THIRD_ID, FOURTH_ID, FIFTH_ID, EMPTY_ID )
 				.hasTotalHitCount( 6 );
 
+		Assertions.assertThat( result.getResponseBody() )
+				.isNotNull()
+				.extracting( body -> body.get( "_shards" ) ).isInstanceOf( JsonObject.class );
```
fax4ever: Maybe I miss something, but it seems that we already test it in the method `query_gsonResponseBody`.

documentation/src/test/java/org/hibernate/search/documentation/search/query/QueryDslIT.java
```diff
@@ -378,6 +382,78 @@ public void fetchSize() {
...
+					.extension( ElasticsearchExtension.get() ) // <1>
+					.predicate( f -> f.match()
+							.field( "title" )
+							.matching( "robot" ) )
+					.requestTransformer( context -> { // <2>
+						Map<String, String> parameters = context.getParametersMap(); // <3>
+						parameters.put( "search_type", "dfs_query_then_fetch" );
+
+						JsonObject body = context.getBody(); // <4>
+						body.addProperty( "min_score", 0.5f );
```
fax4ever: That's nice. It looks very readable.

documentation/src/test/java/org/hibernate/search/documentation/search/query/QueryDslIT.java
```diff
@@ -378,6 +382,78 @@ public void fetchSize() {
...
+						JsonObject body = context.getBody();
+						body.add( "suggest", jsonObject( suggest -> { // <3>
+							suggest.add( "my-suggest", jsonObject( mySuggest -> {
+								mySuggest.addProperty( "text", "robt" );
+								mySuggest.add( "term", jsonObject( term -> {
+									term.addProperty( "field", "title" );
+								} ) );
+							} ) );
+						} ) );
+					} )
```
fax4ever: IMHO that seems too hard to read, and maybe to write as well. Why do not we write the JSON a string instead? Maybe we can do it even if here we're merging objects...

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/ElasticsearchExtensionIT.java
yrodiere: Right, I made a mistake during a rebase

documentation/src/test/java/org/hibernate/search/documentation/search/query/QueryDslIT.java
yrodiere: We could write the JSON as a string in this specific cases, but users will, in most cases, want to generate the JSON dynamically, and then concatenating strings opens the door to injection vulnerabilities. Which is why I didn't want to give an example that would give the impression that using strings is a good idea.

And yes, I agree it's still not very readable even with the utility method. It'll have to do; it's an advanced feature, after all.

yrodiere:
Merged, thanks!

PR #2152, opened 2019-11-15
HSEARCH-3764 Allow disabling conversion in the ID predicate

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3764


fax4ever:
Nice work. I have nothing to say about it.
Merged thanks

PR #2154, opened 2019-11-19
HSEARCH-3767 Elasticsearch date/time bridges throw NPE on projections

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3767


fax4ever:
Merged thanks.

PR #2159, opened 2019-12-04
HSEARCH-1401 Allow integrators to declare entities whose structure is defined programmatically ("free-form")

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-1401

@fax4ever This should allow you to make progress regarding protobuf values in the Infinspan integration.

I checked everything works as intended by implementing support for dynamic-map entities. It's incomplete, but that's no big deal in my opinion.


yrodiere:
Thanks! I addressed your comments and force-pushed. Will merge as soon as the build passes.

PR #2160, opened 2019-12-05
HSEARCH-3780 Fix random failures in ElasticsearchClientFactoryImplIT

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3780


yrodiere:
Merging immediately as these failures are starting to be annoying.

PR #2161, opened 2019-12-06
HSEARCH-3634 Clarify the error message when a @Transient property is missing a @IndexingDependency annotation

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3634


fax4ever:
Merged thanks!

PR #2166, opened 2019-12-06
HSEARCH-3783 query.explain() with the Elasticsearch backend fails when using projections or multi-tenancy

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3783


yrodiere:
Merged, thanks

PR #2167, opened 2019-12-10
HSEARCH-3657 + HSEARCH-3784 Lucene/Elasticsearch metadata storage improvements

yrodiere:
* [HSEARCH-3657](https://hibernate.atlassian.net/browse/HSEARCH-3657): Switch to docvalues instead of stored fields for Hibernate Search metadata in Lucene/ES documents
* [HSEARCH-3784](https://hibernate.atlassian.net/browse/HSEARCH-3784): Avoid storing the index name in the index for each document

I thought I had to do this as a prerequisite for my work on HSEARCH-3765 (Add minimal support for Index Aliases), but it turns out it may not be necessary after all... Anyway, we would have had to do it at some point, so I didn't waste my time.


yrodiere:
Addressed the two bugs detected by sonar and force-pushed.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/reader/impl/IndexReaderMetadataResolver.java
```diff
@@ -0,0 +1,47 @@
...
+		this.indexNameByDirectoryReader = indexNameByDirectoryReader;
+	}
+
+	public String resolveIndexName(LeafReaderContext context) {
+		return indexNameByDirectoryReader.get( getDirectoryReader( context ) );
+	}
+
+	public static Object get(Directory directory) {
+		return directory;
+	}
```
fax4ever: I'm not sure this method is used

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/reader/impl/IndexReaderMetadataResolver.java
```diff
@@ -0,0 +1,47 @@
...
+import org.apache.lucene.index.IndexReaderContext;
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.store.Directory;
+
+public final class IndexReaderMetadataResolver {
+
+	private final Map<DirectoryReader, String> indexNameByDirectoryReader;
+
+	IndexReaderMetadataResolver(Map<DirectoryReader, String> indexNameByDirectoryReader) {
+		this.indexNameByDirectoryReader = indexNameByDirectoryReader;
```
fax4ever: Maybe we can wrap with `Collections#unmodifiableMap()`

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearcherImpl.java
```diff
@@ -160,20 +158,29 @@ else if ( limit.equals( 0 ) ) {
...
+				Set<Integer> nestedDocsForHit = topDocIdsToNestedDocIds.get( hit.doc );
+				if ( nestedDocsForHit != null ) {
+					for ( Integer child : nestedDocsForHit ) {
+						indexSearcher.doc( child, storedFieldVisitor );
+					}
 				}
+				document = storedFieldVisitor.getDocumentAndReset();
+			}
+			else {
+				document = null;
```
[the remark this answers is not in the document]
yrodiere: Right, good catch

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/reader/impl/IndexReaderMetadataResolver.java
yrodiere: Doesn't seem useful here, since the map is not exposed publicly.

If the goal was to prevent the map from being modified by the caller, then we'd have to copy the map. But in this case I'm not worried: the caller can be trusted.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/reader/impl/IndexReaderMetadataResolver.java
yrodiere: Right, good catch

yrodiere:
@fax4ever I rebased on your PR (#2158), please do not merge until that one is merged.

yrodiere:
Merged. Thanks for the review.

PR #2172, opened 2019-12-17
HSEARCH-3461 Do not optimize the indexes by default after mass indexing and document the drawbacks

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3461

### DIFF
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingFailureIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingFailureIT.java
@@ -98,7 +98,7 @@ public class MassIndexingFailureIT {
 						)
 						.hasCauseInstanceOf( SimulatedFailure.class ),
 				expectIndexScopeWork( StubIndexScopeWork.Type.PURGE, ExecutionExpectation.SUCCEED ),
-				expectIndexScopeWork( StubIndexScopeWork.Type.OPTIMIZE, ExecutionExpectation.SUCCEED ),
+				expectIndexScopeWork( StubIndexScopeWork.Type.FORCE_MERGE, ExecutionExpectation.SUCCEED ),
 				expectIndexingWorks( ExecutionExpectation.FAIL ),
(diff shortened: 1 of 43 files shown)


integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingFailureIT.java
```diff
@@ -529,6 +526,21 @@ private void doMassIndexingWithFailure(SessionFactory sessionFactory,
...
 		doMassIndexingWithFailure(
 				sessionFactory,
+				searchSession -> searchSession.massIndexer(),
+				threadExpectation,
+				thrownExpectation,
+				expectationSetters
+		);
+	}
+
+	private void doMassIndexingWithFailure(SessionFactory sessionFactory,
```
fax4ever: IMHO too many indirections on helper methods make the test code less readable. If it is possible, I would not create a new helper method here.

yrodiere: While I agree indirection should be avoided if possible, here the alternative is copy/pasting 70 lines of code into a dozen different tests. That would hardly be more readable, and certainly not more maintainable.

That being said, I made even more changes to this test in another PR, and I suspect you won't like these changes either. Let's discuss this there, and I'll consider changing this if you can provide a reasonable alternative.

fax4ever: I think there is something in the middle between avoid indirection at all and the current solution.
I would reorganize the sub-calls so that we do not never reach a depth of 3.
Basically one main method that every test method eventually would use, using an optional  indirection method at most to reach it.
This comment was a nice to have, it is not critical.

yrodiere:
Good catch for `unableToOptimizeIndex`. I renamed the method and changed the message.

Regarding the name of the operation, I must admit I do not like to stray from the official Lucene/Elasticsearch wording, but in this particular case I'll make an exception. But it's just because `mergeSegments` is indeed a better name and this operation should only be use very rarely, and most likely by advanced users.

I rebased and added a commit, let's wait for CI.

yrodiere:
Merged, thanks!

PR #2173, opened 2019-12-17
HSEARCH-3729 Give more information about failures to MassIndexer users

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3729

Based on #2172 , which should be merged first.


yrodiere:
Thanks. I just rebased on master. I will merge as soon as CI goes green.

PR #2174, opened 2019-12-17
HSEARCH-3795 Get rid of Version.logVersion()

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3795


fax4ever:
Merged thanks!

PR #2176, opened 2019-12-18
HSEARCH-3418 Document that Instants are truncated to the millisecond when indexed

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3418

### DIFF
--- a/documentation/src/main/asciidoc/mapper-orm-mapping.asciidoc
+++ b/documentation/src/main/asciidoc/mapper-orm-mapping.asciidoc
@@ -543,24 +543,24 @@ or <<backend-elasticsearch-field-types,Elasticsearch field types>> depending on 
 |java.time.MonthDay|-|-|`MonthDay.parse(String)`
 |java.util.UUID|`toString()` as a `java.lang.String`|-|`UUID.fromString(String)`
 |java.util.Calendar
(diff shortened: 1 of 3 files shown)


documentation/src/main/asciidoc/mapper-orm-mapping.asciidoc
```diff
@@ -479,47 +479,108 @@ see <<backend-lucene-field-types,Lucene field types>>
...
+ |<<mapper-orm-directfieldmapping-supported-types-date-time>>
+ |`Year.parse(String)`
+|java.time.YearMonth
+ |-
+ |<<mapper-orm-directfieldmapping-supported-types-date-time>>
+ |`YearMonth.parse(String)`
+|java.time.MonthDay|-|-|`MonthDay.parse(String)`
+|java.util.UUID|`toString()` as a `java.lang.String`|-|`UUID.fromString(String)`
+|java.util.Calendar
+ |A `java.time.ZonedDateTime` representing the same date/time and timezone; see <<mapper-orm-legacy-date-time-apis>>.
```
fax4ever: Maybe we can put the `<<mapper-orm-legacy-date-time-apis>>`s under the limitations column.

yrodiere:
Yes, the limitations are _currently_ the same. I am not sure it will stay that way, though, so I'd rather document them separately. It's not like many people will need to use both backends anyway ;)

I addressed your comment and rebased. Will merge as soon as CI goes green.

PR #2177, opened 2019-12-18
HSEARCH-1828 Clarify documentation about ways to disable Hibernate Search

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-1828


yrodiere:
@Sanne Merged, thanks. Nice to see you here ;)

PR #2178, opened 2019-12-18
HSEARCH-3793 Document usage of the "exists" predicate on object fields

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3793


fax4ever:
Merged thanks!

PR #2180, opened 2020-01-13
HSEARCH-3801 Hibernate Search blocks loading of non hibernate batch jobs

zeitiger:
In reference of `org.jberet.spi.JobXmlResolver` it's defined that a resolver that is not responsible should return `null` instead of throwing a runtime exception. The exception prevent that further available resolver will be consider.
```
    /**
     * Locates the job XML and creates a stream to the contents.
     *
     * @param jobXml      the name of the job XML with a {@code .xml} suffix
     * @param classLoader the class loader for the application
     *
     * @return a stream of the job XML or {@code null} if the job XML content was not found
     *
     * @throws java.io.IOException if an error occurs creating the stream
     */
```

### DIFF
--- a/jsr352/jberet/src/main/java/org/hibernate/search/jsr352/jberet/impl/HibernateSearchJobXmlResolver.java
+++ b/jsr352/jberet/src/main/java/org/hibernate/search/jsr352/jberet/impl/HibernateSearchJobXmlResolver.java
@@ -15,7 +15,6 @@ import java.util.Map;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


yrodiere:
Thanks for taking the time of contributing!

Before this is merged, we will need a few adjustments to follow our best practices:

1. Remove the (now useless) import of `SearchException` in the file you changed.
2. Create a JIRA ticket for your issue: https://hibernate.atlassian.net/
3. Prefix the commit message with the key of your ticket.

yrodiere:
Merged and backported to 5.10. Thanks!

PR #2181, opened 2020-01-14
HSEARCH-3629 Move search query loading options to a better place in the search query DSL

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3629


fax4ever:
Didn't find anything to say.
Good idea to group the minor options!
Merged thanks

PR #2182, opened 2020-01-16
HSEARCH-3802 query.fetchTotalHitCounts always fails when setting a timeout with Elasticsearch

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3802


fax4ever:
Thank you, merged!

PR #2183, opened 2020-01-17
HSEARCH-3803 Pass id bridge for implicit provided id

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3803


mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/runtime/impl/ProvidedStringIdentifierMapping.java
```diff
@@ -9,21 +9,36 @@
...
+	private final BeanHolder<? extends IdentifierBridge<Object>> bridgeHolder;
 
 	@SuppressWarnings("unchecked") // This class is bivariant in E
-	public static <E> IdentifierMappingImplementor<String, E> get() {
-		return (IdentifierMappingImplementor<String, E>) INSTANCE;
+	public static <E> IdentifierMappingImplementor<String, E> get(BeanHolder<? extends IdentifierBridge<Object>> bridgeHolder) {
+		return (IdentifierMappingImplementor<String, E>) new ProvidedStringIdentifierMapping( bridgeHolder );
+	}
+
+	public ProvidedStringIdentifierMapping(BeanHolder<? extends IdentifierBridge<Object>> bridgeHolder) {
```
yrodiere: This should probably be private?

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/runtime/impl/ProvidedStringIdentifierMapping.java
```diff
@@ -9,21 +9,36 @@
...
+import org.hibernate.search.engine.environment.bean.BeanHolder;
+import org.hibernate.search.mapper.pojo.bridge.IdentifierBridge;
 import org.hibernate.search.mapper.pojo.bridge.runtime.spi.BridgeMappingContext;
 import org.hibernate.search.mapper.pojo.bridge.runtime.spi.BridgeSessionContext;
 import org.hibernate.search.mapper.pojo.logging.impl.Log;
+import org.hibernate.search.util.common.impl.Closer;
 import org.hibernate.search.util.common.logging.impl.LoggerFactory;
 
 
 public class ProvidedStringIdentifierMapping implements IdentifierMappingImplementor<String, Object> {
```
yrodiere: The type arguments are wrong here. It should be `<Object, Object>`. See the javadoc of the interface.

Also, consequently, you should probably remove the class to `ProvidedIdentifierMapping`: it's no longer exclusively about string identifiers.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/runtime/impl/ProvidedStringIdentifierMapping.java
```diff
@@ -41,12 +56,12 @@ public String getIdentifier(Object providedId, Supplier<?> entityProvider) {
```
yrodiere: You should also change `getIdentifier(Object)` above. Just remove `String.valueOf()` and return `providedId` directly.
Please keep the "if" block: it's necessary.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/building/impl/PojoIdentityMappingCollectorImpl.java
```diff
@@ -48,11 +50,11 @@
 			PojoEntityTypeAdditionalMetadata entityTypeMetadata,
 			PojoMappingHelper mappingHelper,
 			IndexedEntityBindingContext bindingContext,
-			boolean implicitProvidedId) {
+			BeanHolder<? extends IdentifierBridge<Object>> identifierBridge) {
 		this.typeModel = typeModel;
 		this.mappingHelper = mappingHelper;
 		this.bindingContext = bindingContext;
-		this.implicitProvidedId = implicitProvidedId;
+		this.identifierBridge = identifierBridge;
```
yrodiere: `providedIdentifierBridge` would be less ambiguous. For other mappers, the identifier bridge is set up differently.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/building/impl/PojoIndexedTypeManagerBuilder.java
```diff
@@ -51,7 +53,7 @@
 			PojoMappingHelper mappingHelper,
 			MappedIndexManagerBuilder<D> indexManagerBuilder,
 			PojoIndexedTypeExtendedMappingCollector extendedMappingCollector,
-			boolean implicitProvidedId) {
+			BeanHolder<? extends IdentifierBridge<Object>> identifierBridge) {
```
yrodiere: Same here, I would rename this parameter to `providedIdentifierBridge`.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/building/impl/PojoMapper.java
```diff
@@ -67,7 +70,7 @@
 
 	private final ContextualFailureCollector failureCollector;
 	private final TypeMetadataContributorProvider<PojoTypeMetadataContributor> contributorProvider;
-	private final boolean implicitProvidedId;
+	private final BeanHolder<? extends IdentifierBridge<Object>> identifierBridge;
```
yrodiere: Same here, I would rename this field to `providedIdentifierBridge`.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/building/impl/PojoMapper.java
```diff
@@ -93,19 +96,21 @@ public PojoMapper(MappingBuildContext buildContext,
 			TypeMetadataContributorProvider<PojoTypeMetadataContributor> contributorProvider,
 			PojoBootstrapIntrospector introspector,
 			ContainerExtractorRegistry containerExtractorRegistry,
-			boolean implicitProvidedId,
+			BeanReference<? extends IdentifierBridge<Object>> identifierBridge,
```
yrodiere: Same here, I would rename this parameter to `providedIdentifierBridgeReference`.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/building/impl/PojoMapper.java
```diff
@@ -93,19 +96,21 @@ public PojoMapper(MappingBuildContext buildContext,
...
-		this.implicitProvidedId = implicitProvidedId;
 		this.multiTenancyEnabled = multiTenancyEnabled;
 
 		this.failureHandler = buildContext.getFailureHandler();
 		this.threadPoolProvider = buildContext.getThreadPoolProvider();
 
 		this.delegate = delegate;
 
+		this.identifierBridge = ( identifierBridge == null ) ? null :
+				buildContext.getBeanResolver().resolve( identifierBridge );
```
yrodiere: A `BeanHolder` must be closed. If you instantiate the bridge here, you must close it in case of failure.

Personally, I would just store the reference here, pass it along to `new PojoIndexedTypeManagerBuilder` in `createIndexedTypeManagerBuilder`, and only instantiate the bridge in `PojoIdentityMappingCollectorImpl#applyDefaults` *if necessary*. You really don't need to instantiate the bridge so early.

This is actually what was explained in the ticket, so I assume you did this because you didn't have access to the bean resolver in `PojoIdentityMappingCollectorImpl`. If that's the problem, well, I don't have any problem with passing the bean resolver or the `MappingBuildContext` alongside the bean reference.

It's also more correct, given that the APIs allow users to override the provided ID on a per-entity basis, and the map the ID from a property instead. Though that's not really a feature, more a side-effect of how the APIs are designed.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/runtime/impl/ProvidedStringIdentifierMapping.java
fax4ever: You're definitely right. Do you mean `you should probably` **rename** `the class`?

yrodiere: Yes, that was a typo.

fax4ever:
Thank you @yrodiere for the review. I agree with all your comments.
I put the change on bean resolving and closing on failure in different commit.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/building/impl/PojoIdentityMappingCollectorImpl.java
```diff
@@ -69,6 +77,7 @@ void closeOnFailure() {
 		try ( Closer<RuntimeException> closer = new Closer<>() ) {
 			closer.push( IdentifierMappingImplementor::close, identifierMapping );
 			closer.push( RoutingKeyProvider::close, routingKeyProvider );
+			closer.push( BeanHolder::close, providedIdentifierBridgeHolder );
```
yrodiere: I don't think you need this, as `identifierMapping.close()` (above) will effectively close the bridge, and it's very unlikely (impossible?) that a failure occurs between the assignment of `providedIdentifierBridgeHolder` and the assignment of `identifierMapping`.

So, to clarify: you can remove `providedIdentifierBridgeHolder` and use a local variable instead, then remove this line.

fax4ever: Yes, I also think that it is impossible.
I applied the changes, let's wait for the CI, then I will merge it.
Thanks

yrodiere:
Thanks!

PR #2184, opened 2020-01-20
HSEARCH-3798 Test sorting on numeric fields without specifying missing value behavior

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3798


fax4ever:
Good to have these tests. Merged, thanks!

PR #2185, opened 2020-01-20
HSEARCH-3171 Test using CDI beans referenced by their name

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3171


fax4ever:
Nice job! Merged thanks!

PR #2186, opened 2020-01-21
HSEARCH-3810 + HSEARCH-3811 + HSEARCH-3812 Performance and logging improvements

yrodiere:
* [HSEARCH-3810](https://hibernate.atlassian.net/browse/HSEARCH-3810): Rework JBoss logging formatters: execute computationally intensive operations in toString rather than the constructor
* [HSEARCH-3811](https://hibernate.atlassian.net/browse/HSEARCH-3811): Various attempts at improving bootstrap performance
* [HSEARCH-3812](https://hibernate.atlassian.net/browse/HSEARCH-3812): Be more consistent with timeout messages in search queries across backends (Lucene/ES)

This does not seem to improve performance noticeably, probably because the impacted code is already quick to execute. Regardless, I had to work on it to test it, and it looks better, so... let's merge?


fax4ever:
Have you used some performance test?
Anyway changes look good. So I merged.

yrodiere:
> Have you used some performance test?

It's about bootstrap performance, so I didn't really need a specific test. I used [async-profiler](https://github.com/jvm-profiling-tools/async-profiler) for sampling, VisualVM for profiling, applied to [Guillaume's demo](https://github.com/hibernate/hibernate-demos/).
Got 2.2s boot time with his demo, and approximately 4s if I add another 300 simple indexed entities. So it's slower than we'd like, but at least it scales rather well for simple entities.

As I said, there's wasn't much improvement with this PR. A lot of the bootstrap time was spent loading classes, and the rest seemed legitimate overall.

The only obvious problems were how the Elasticsearch REST client [instantiates an SSLContext for no reason](https://github.com/elastic/elasticsearch/issues/51225) and GSON's slow startup (we'll try to address that, but that can't be done in just a few hours).

PR #2196, opened 2020-01-29
HSEARCH-3819 JsonSyntaxException when validating or migrating an Elasticsearch schema with a geo_point field with null_value

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3819

I encountered the problem while working on something else.

I checked this solves the problem on my other branch, but I don't think it's worth an automated test. A lot of tests will fail in case of regression anyway, at least once my other branch is merged.


fax4ever:
I think the same, a lot of tests would fail in case of regression.
Thanks I'm merging it.

PR #2199, opened 2020-02-04
HSEARCH-3791 Configurable alias-based reads/writes for Elasticsearch indexes

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3791

Essentially, this changes the way we access Elasticsearch indexes:

* Before this PR, for a HSearch index named `myIndex`, we used to create an Elasticsearch index named `myindex`. Then, search queries and writes are sent directly to that index.
* After this PR, for a HSearch index named `myIndex`, we create an Elasticsearch index named `myindex-000001` with two aliases pointing to that index: `myindex-write` and `myindex-read`. Then, search queries are sent to `myindex-read` and writes are sent to `myindex-write`.

During normal operation of an application, this does not change anything. When reindexing, however, we now have a way to keep a functioning search while we are busy building a new copy of the index. See commit "HSEARCH-3791 Demonstrate the feasibility user-managed zero-downtime reindexing" as well as the documentation I added.

We'll add something to automate the alias switching when mass indexing when we address another ticket.

I ran a full build here, it passed: https://ci.hibernate.org/job/hibernate-search-personal-yoann/job/HSEARCH-3791/35/


documentation/src/main/asciidoc/backend-elasticsearch.asciidoc
```diff
@@ -265,6 +265,67 @@ Elasticsearch disables sharding by default.
...
+2. Configure the backend to use that implementation by setting the configuration property
+`hibernate.search.backends.<backend name>.layout.strategy`
+to a <<configuration-property-types,bean reference>> pointing to the implementation.
+
+For example, the implementation below will lead to the following layout for an index named `myIndex`:
+
+* Write operations (indexing, purge, ...) will target the alias `myindex-write`.
+* Read operations (searching, explaining, ...) will target the alias `myindex` (no suffix).
+* If Hibernate Search <<backend-elasticsearch-index-lifecycle,creates the index automatically>>
+at exactly 19:19:00 on November 6th, 2017,
```
fax4ever: /2017/2020/s ?

yrodiere:
> I've just one concern, is it possible that there exists some use cases could not work well having the read/write aliases?

I'm not aware of any setup where creating aliases may be impossible. As far as I can tell, it's a best practice.

> In other word, does make sense let the user to use Hibernate Search without them (as in beta4)?

It could, but the cost of maintaining both approaches in our codebase is, in my opinion, too high to be worth it. Worst case, if someone complains and provides a valid use case, we might introduce a configuration option to revert to the beta4 behavior. It would have to be a damn good use case, though.

>  Anyway +1 to merge from me.

Merged, thanks!

PR #2200, opened 2020-02-10
HSEARCH-3775 Restore advanced Reader/Writer handling in the Lucene backend

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3775

Complete explanation in the documentation added in the very last commit.


documentation/src/main/asciidoc/backend-lucene.asciidoc
```diff
@@ -543,3 +542,136 @@ link:http://wiki.apache.org/solr/AnalyzersTokenizersTokenFilters[Solr Wiki]
...
+[[backend-lucene-io-commit]]
+=== Commit
+
+In Lucene terminology, a _commit_ is when changes buffered in an index writer
+are pushed to the index itself,
+so that a crash or power loss will no longer result in data loss.
+
+Performance-wise, committing may be an expensive operation,
+which is why Hibernate Search tries not to commit too often.
+Changes are processed in batches of a few hundred writes,
```
fax4ever: Maybe we would eventually make the value configurable.

documentation/src/main/asciidoc/backend-lucene.asciidoc
```diff
@@ -543,3 +542,136 @@ link:http://wiki.apache.org/solr/AnalyzersTokenizersTokenFilters[Solr Wiki]
...
+then the index reader will not be refreshed, even though it may be out-of-date.
+
+The refresh interval is set at the index level:
+
+[source]
+----
+hibernate.search.backends.<backend name>.indexes.<index name>.io.refresh_interval = 0 (default)
+# OR
+hibernate.search.backends.<backend name>.index_defaults.io.refresh_interval = 0 (default)
+----
```
fax4ever: I would mention the relationship between the attributes `refresh_interval ` and `synchronization.strategy`. If I understand well, with the `searchable` `synchronization.strategy` `refresh_interval` has no meaning.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/reader/impl/NearRealTimeIndexReaderProvider.java
```diff
@@ -0,0 +1,155 @@
...
+		/**
+		 * @return {@code true} if the reader is still fresh enough to be used,
+		 * i.e. if it is completely up-to-date with the state of the index writer
+		 * OR is out-of-date by less than the configured refresh interval,
+		 * and refresh wasn't forced by a previous write.
+		 * @throws IOException If an I/O failure occurs.
+		 */
+		boolean isFresh() throws IOException {
+			if ( refreshForced ) {
+				return false;
```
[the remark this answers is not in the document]
yrodiere: Expiration happens in the background without user interaction, so we try to optimize things.

This is a forced refresh, requested explicitly by the user, so I'd rather not try to be smart and just do as we are told.

engine/src/main/java/org/hibernate/search/engine/backend/orchestration/spi/BatchingExecutor.java
```diff
@@ -216,48 +214,112 @@ private void processBatch() {
...
+		// Nothing more to do, end the batch and terminate
+		CompletableFuture<?> batchFuture = processor.endBatch();
+
+		/*
+		 * Wait for works to complete before trying to handle the next batch.
+		 * Note: timeout is expected to be handled by the processor
+		 * (the Elasticsearch client adds per-request timeouts, in particular),
+		 * so this "join" will not last forever
+		 */
+		Futures.unwrappedExceptionJoin( batchFuture );
```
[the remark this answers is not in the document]
yrodiere: That's unrelated; the fact that we return a `CompletableFuture` is mainly for the benefit of the Elasticsearch backend.

For what it's worth, yes, we could block inside `endBatch` even in the Elasticsearch backend and skip returning a `CompletableFuture`, but that would require a lot of changes in the tests of the Elasticsearch work processor, so I'd rather not do that in this PR (if at all).

documentation/src/main/asciidoc/backend-lucene.asciidoc
yrodiere: Yes, [that's the plan](https://hibernate.atlassian.net/browse/HSEARCH-3575) :)

documentation/src/main/asciidoc/backend-lucene.asciidoc
yrodiere: Yes, it is useful when mass indexing. But I'll cover that in [HSEARCH-3808](https://hibernate.atlassian.net/browse/HSEARCH-3808): I need to update that part of the documentation for other reasons, too.

yrodiere:
Merged, thanks!

PR #2201, opened 2020-02-12
HSEARCH-3827 Restore simple backend performance tests

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3827

~Based on #2200, which should be merged first.~ => Done

I'm getting the following numbers of writes per second for Lucene.
Note that absolute numbers don't mean anything, since they heavily depend on hardware, the number of threads, etc. What's important is that we actually see an improvement when we set a commit interval.

Mass indexing (without a DB, generated content):

``` 
configuration                                       Score           Score Error (99.9%)
(default)                                           85,802.99       3,425.56
io.commit_interval=1000                             236,790.43      6,360.17
io.commit_interval=1000&io.refresh_interval=1000    238,705.66      10,294.92
io.strategy=debug                                   89,021.94       3,367.94
```

Automatic indexing (without a DB, generated content):

```
configuration                                     commitStrategy  Score        Score Error (99.9%)
(default)                                         FORCE           7,569.90     157.26
                                                  NONE            99,333.15    2,313.56
io.commit_interval=1000                           FORCE           9,155.50     223.43
                                                  NONE            664815.0894  40274.47602
io.commit_interval=1000&io.refresh_interval=1000  FORCE           9359.757108  215.647766
                                                  NONE            643673.987   25927.36115
io.strategy=debug                                 FORCE           7753.374738  227.715543
                                                  NONE            103691.1606  4578.391557
```

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/cfg/impl/SystemConfigurationPropertySource.java
+++ b/engine/src/main/java/org/hibernate/search/engine/cfg/impl/SystemConfigurationPropertySource.java
@@ -1,0 +1,57 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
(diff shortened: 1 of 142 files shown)


fax4ever:
Great work! Merged thanks

PR #2202, opened 2020-02-13
HSEARCH-3460 Remove the confusion between flush and refresh

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3460

~Based on #2200, which should be merged first.~ => Done

Essentially, this introduces a "refresh" operation to make changes to the index visible in search, and removes the implicit refresh from the "flush" operation. The goal is to make the API more sensible and to give more control over what's going on in the background, in particular for advanced users.

See the commit "HSEARCH-3460 Document APIs for explicit index refresh" for more information.


fax4ever:
It seems to be a compilation error on `IndexWorkspaceRefreshIT`, that is visible only if we rebase this branch on top of master.

fax4ever:
About last commit IMHO index scope is not so confusing with search scope. Because even if the word scope is the same, there are intended differently: the scope of a search vs an operation that is scoped on an index.
Anyway, +1 to merge when we it is rebased manually on top of the master.

yrodiere:
Merged, thanks

PR #2206, opened 2020-02-17
SearchSortFactory documentation update

bobbyphilip:
This is related to  HSEARCH-3721 changed by e9aac57256e8138d1ebdf97fdf4cbc0dea124755
The documentation was not updated at that point


yrodiere:
Good catch! Rebased and merged, thanks.

PR #2208, opened 2020-02-19
HSEARCH-3833 Writing on different local-heap indexes will provoke LockObtainFailedException

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3833

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/directory/impl/DirectoryProviderInitializationContextImpl.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/directory/impl/DirectoryProviderInitializationContextImpl.java
@@ -7,6 +7,7 @@
 package org.hibernate.search.backend.lucene.lowlevel.directory.impl;
 
 import java.util.Optional;
+import java.util.function.Supplier;
 
 import org.hibernate.search.backend.lucene.cfg.LuceneBackendSettings;
 import org.hibernate.search.backend.lucene.lowlevel.directory.LockingStrategyName;
@@ -69,10 +70,10 @@ public class DirectoryProviderInitializationContextImpl implements DirectoryProv
 	}
 
 	@Override
-	public Optional<LockFactory> createConfiguredLockFactory() {
+	public Optional<Supplier<LockFactory>> createConfiguredLockFactorySupplier() {
 		// TODO HSEARCH-3635 Restore support for configuring a custom LockFactory
 		return LOCKING_STRATEGY.get( configurationPropertySource )
-				.map( DirectoryProviderInitializationContextImpl::createLockFactory );
+				.map( DirectoryProviderInitializationContextImpl::createLockFactorySupplier );
 	}
 
 	public BeanHolder<? extends DirectoryProvider> createDirectoryProvider() {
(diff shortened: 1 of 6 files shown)


integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/search/LuceneSearchMultiIndexIT.java
```diff
@@ -63,9 +71,19 @@
...
+
+	public LuceneSearchMultiIndexIT(String directoryType) {
+		this.directoryType = directoryType;
+	}
+
 	@Before
 	public void setup() {
-		setupHelper.start()
+		setupHelper.start( BACKEND_1_NAME )
+				.withBackendProperty( BACKEND_1_NAME, "directory.type", directoryType )
```
yrodiere: You no longer need to define and pass a backend name. You can call `start()` and `withBackendProperty( "directory.type", directoryType )`.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/directory/impl/LocalFileSystemDirectoryProvider.java
```diff
@@ -58,7 +58,7 @@ public void initialize(DirectoryProviderInitializationContext context) {
 		this.directoryRoot = ROOT.get( propertySource ).toAbsolutePath();
 		FileSystemAccessStrategyName accessStrategyName = FILESYSTEM_ACCESS_STRATEGY.get( propertySource );
 		this.accessStrategy = FileSystemAccessStrategy.get( accessStrategyName );
-		this.lockFactory = context.createConfiguredLockFactory().orElseGet( FSLockFactory::getDefault );
```
yrodiere: Why do you call `get` here? You'll end up having the exact same problem as with the local heap directory, if the user selects the single-instance lock factory explicitly through configuration.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/directory/impl/DirectoryProviderInitializationContextImpl.java
```diff
@@ -93,16 +94,16 @@ public ConfigurationPropertySource getConfigurationPropertySource() {
 		}
 	}
 
-	private static LockFactory createLockFactory(LockingStrategyName name) {
```
yrodiere: This should probably be renamed to `createLockFactorySupplier`.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/directory/impl/DirectoryProviderInitializationContextImpl.java
```diff
@@ -69,7 +70,7 @@ public ConfigurationPropertySource getConfigurationPropertySource() {
 	}
 
 	@Override
-	public Optional<LockFactory> createConfiguredLockFactory() {
```
yrodiere: This should probably be renamed to `createConfiguredLockFactorySupplier`.

fax4ever:
Merged thanks!

PR #2209, opened 2020-02-19
HSEARCH-3808 + HSEARCH-3831 Automatic indexing synchronization strategy improvements

yrodiere:
* [HSEARCH-3808](https://hibernate.atlassian.net/browse/HSEARCH-3808): Rename and clarify automatic indexing synchronization strategies
* [HSEARCH-3831](https://hibernate.atlassian.net/browse/HSEARCH-3831): Allow setting the automatic indexing sync strategy to a custom bean in configuration


yrodiere:
Thanks. I kept the name `sync` for two reasons:

1. to keep the symmetry with `async`
2. We want to follow the principle of least surprise. According to this principle, if a users sets the strategy to `sync`, we're expected to be _fully_ synchronous anyway.

PR #2210, opened 2020-02-19
HSEARCH-3825 Purging a document from the POJO mapper will lead to NPE

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3825

Turns out the NPE would only happen when a routing key bridge was defined.

Still, a fix is necessary.


mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/work/spi/PojoIndexingPlan.java
```diff
@@ -90,12 +90,12 @@
 	 * Entities to reindex as a result from this operation will not be resolved.
 	 * <p>
 	 * No effect on the index if the entity is not in the index.
-	 *
-	 * @param typeIdentifier The identifier of the entity type.
+	 *  @param typeIdentifier The identifier of the entity type.
```
fax4ever: There is an extra space before `@param`

yrodiere:
Rebased and merged. Thanks!

PR #2220, opened 2020-02-26
HSEARCH-3841 new version_check.enabled setting added

acucumel:
https://hibernate.atlassian.net/browse/HSEARCH-3841

new settings added to get rid of the version check at start in some cases.

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchLinkImpl.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchLinkImpl.java
@@ -6,10 +6,7 @@
  */
 package org.hibernate.search.backend.elasticsearch.impl;
 
-import java.io.IOException;
-import java.lang.invoke.MethodHandles;
-import java.util.Optional;
-
+import com.google.gson.GsonBuilder;
 import org.hibernate.search.backend.elasticsearch.ElasticsearchVersion;
 import org.hibernate.search.backend.elasticsearch.client.impl.ElasticsearchClientUtils;
 import org.hibernate.search.backend.elasticsearch.client.spi.ElasticsearchClient;
@@ -31,7 +28,9 @@ import org.hibernate.search.util.common.AssertionFailure;
 import org.hibernate.search.util.common.impl.Closer;
 import org.hibernate.search.util.common.logging.impl.LoggerFactory;
 
-import com.google.gson.GsonBuilder;
+import java.io.IOException;
+import java.lang.invoke.MethodHandles;
+import java.util.Optional;
 
 class ElasticsearchLinkImpl implements ElasticsearchLink {
 	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
@@ -42,6 +41,7 @@ class ElasticsearchLinkImpl implements ElasticsearchLink {
 	private final boolean logPrettyPrinting;
 	private final ElasticsearchDialectFactory dialectFactory;
 	private final Optional<ElasticsearchVersion> configuredVersionOptional;
+	private final boolean versionCheckEnabled;
 
 	private ElasticsearchClientImplementor clientImplementor;
 	private ElasticsearchVersion elasticsearchVersion;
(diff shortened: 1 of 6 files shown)


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchLinkImpl.java
```diff
@@ -111,12 +114,26 @@ void onStart(ConfigurationPropertySource propertySource) {
...
+					if ( !configuredVersion.matches( elasticsearchVersion ) ) {
+						throw log.unexpectedElasticsearchVersion( configuredVersion, elasticsearchVersion );
+					}
+				}
+			}
+			else if ( configuredVersionOptional.isPresent() ) {
 				ElasticsearchVersion configuredVersion = configuredVersionOptional.get();
-				if ( !configuredVersion.matches( elasticsearchVersion ) ) {
-					throw log.unexpectedElasticsearchVersion( configuredVersion, elasticsearchVersion );
+				if ( configuredVersion.getMinor().isPresent() && configuredVersion.getMicro().isPresent() ) {
```
yrodiere: Ideally this (and the case "no version is configured") should be checked earlier, in `ElasticsearchBackendFactory`. This is especially important for Quarkus, where the configuration is checked at compile-time.

Also, I think you can relax the constraint and allow the micro to be missing: the Elasticsearch team would have to mess up big time for us to have to behave differently based on the micro.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -80,8 +81,8 @@
 			value = "Elasticsearch request failed: %3$s\nRequest: %1$s\nResponse: %2$s"
 	)
 	SearchException elasticsearchRequestFailed(
-			@FormatWith( ElasticsearchRequestFormatter.class ) ElasticsearchRequest request,
-			@FormatWith( ElasticsearchResponseFormatter.class ) ElasticsearchResponse response,
+			@FormatWith(ElasticsearchRequestFormatter.class) ElasticsearchRequest request,
```
yrodiere: I agree these changes are correct, but would you mind moving them to a separate commit dedicated to formatting changes?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchLinkImpl.java
acucumel: thanks for the tips, i will do that in few minutes

acucumel:
changes made

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
acucumel: i tried to remove my first commit but it didn't workout ^^

yrodiere:
Looks you added a test for the case where no version is configured; which is nice, I didn't think about that. But you forgot to add the test where everything works :)

As explained above:

> Copy/paste explicitModelDialect to a new test method that checks what happens when the version check is disabled and a correct version is passed. You should only need to set the new property with a call to withBackendProperty, and to remove the call to elasticsearchClientSpy.expectNext.

Also, the build fails because of style issues: see https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2220/2/pipeline#step-38-log-958

You can run `mvn checkstyle:check -fn` locally before you commit to check the code style: that way, you won't have to wait for the CI build.

acucumel:
mvn checkstyle:check -fn is ok this time and i add the `explicitProtocolDialect_noVersionCheck_completeVersion` :)

yrodiere:
I rebased on master, squashed your commits and removed the formatting changes in `Log.java`.

I also changed the "version" documentation (including pre-existing documentation) to focus more on the use case where the version check is disabled; after some though I think it makes more sense.

Merging, thanks!

PR #2221, opened 2020-02-27
HSEARCH-3814 Mark experimental APIs/SPIs with an @Incubating annotation

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3814

I've adapted the one from Hibernate Validator.


mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/model/dependency/PojoOtherEntityIndexingDependencyConfigurationContext.java
```diff
@@ -23,9 +21,8 @@
 	 * @return {@code this}, for method chaining.
 	 * @throws org.hibernate.search.util.common.SearchException If the given path cannot be applied to the entity type.
 	 * @see #use(PojoModelPathValueNode)
-	 * @hsearch.experimental This feature is under active development.
-	 *    Usual compatibility policies do not apply: incompatible changes may be introduced in any future release.
 	 */
+	@Incubating
```
yrodiere: Looks like you forgot the method just below.

fax4ever: Yes i did. Fixed

fax4ever:
Merged! Thanks for the review

PR #2222, opened 2020-02-27
HSEARCH-2859 Test multivalued spacial field

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-2859

### DIFF
--- a/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/spatial/MultivaluedSpatialIT.java
+++ b/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/spatial/MultivaluedSpatialIT.java
@@ -1,0 +1,93 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.integrationtest.backend.tck.search.spatial;
+
+import static org.hibernate.search.util.impl.integrationtest.common.assertion.SearchResultAssert.assertThat;
+import static org.hibernate.search.util.impl.integrationtest.mapper.stub.StubMapperUtils.referenceProvider;
+
+import org.hibernate.search.engine.backend.common.DocumentReference;
+import org.hibernate.search.engine.backend.document.DocumentElement;
+import org.hibernate.search.engine.backend.document.IndexFieldReference;
+import org.hibernate.search.engine.backend.document.model.dsl.IndexSchemaElement;
+import org.hibernate.search.engine.backend.work.execution.spi.IndexIndexingPlan;
+import org.hibernate.search.engine.search.query.SearchQuery;
+import org.hibernate.search.engine.spatial.GeoBoundingBox;
+import org.hibernate.search.engine.spatial.GeoPoint;
+import org.hibernate.search.integrationtest.backend.tck.testsupport.util.rule.SearchSetupHelper;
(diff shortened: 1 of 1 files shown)


integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/spatial/MultivaluedSpacialIT.java
```diff
@@ -0,0 +1,147 @@
...
+import org.hibernate.search.engine.spatial.GeoPolygon;
+import org.hibernate.search.integrationtest.backend.tck.testsupport.util.rule.SearchSetupHelper;
+import org.hibernate.search.util.impl.integrationtest.mapper.stub.StubMappingIndexManager;
+import org.hibernate.search.util.impl.integrationtest.mapper.stub.StubMappingScope;
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class MultivaluedSpacialIT {
```
yrodiere: It's spatial, with a `T`. Same for the commit message.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/spatial/MultivaluedSpacialIT.java
```diff
@@ -0,0 +1,147 @@
...
+
+		query = scope.query()
+				.where( f -> f.spatial().within().field( "geoPoint" ).circle( PARIS, 1000 ) )
+				.toQuery();
+
+		assertThat( query ).hasDocRefHitsAnyOrder( INDEX_NAME, "1", "3" );
+	}
+
+	@Test
+	public void boundingBox() {
```
yrodiere: I'm not sure this tests the bug described in HSEARCH-2859.

In Search 5, latitude and longitude were stored in two separate fields.
As a result, as soon as you had multiple values, there wasn't any way to know which latitude matched which longitude anymore, and we could end up having documents that match even though they shouldn't.

To test for that, you should craft your coordinates carefully.

For example:

* Doc 1: `[0.0, 0.0]`, `[1.0, 1.0]`
* Doc 2: `[0.0, 1.0]`, `[1.0, 0.0]`

If you search for documents in the bounding box `([-0.5, -0.5], [0.5, 0.5])`, only document 1 should match, because only point `[0.0, 0.0]` is in the bounding box. If document 2 matches, it means the coordinates are all mixed up, and the bug I described above occurred.

Here with Rome and Paris, I don't think you can get any kind of false positive by using the latitude of Rome with the longitude of Paris, and vice-versa. That's because your predicates would filter just as well if they filtered exclusively on the latitude or exclusively on the longitude.

fax4ever:
Thanks for the review. I didn't get that issue was about merging longitude and latitude of different values. I think I fixed the test.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/spatial/MultivaluedSpatialIT.java
```diff
@@ -0,0 +1,93 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+
+@TestForIssue(jiraKey = "HSEARCH-2859")
+public class MultivaluedSpatialIT {
+
+	private static final String INDEX_NAME = "IndexName";
+
+	// latitude increases from south to north
+	// longitude increases from west to est
```
yrodiere: Typo: it's "east", not "est". Same below.

fax4ever: Right! I updated the name of the constant as well.

yrodiere:
Merged, thanks

fax4ever:
Thanks

PR #2223, opened 2020-02-27
HSEARCH-3844 Exception: null: java.lang.NullPointerException when use simpleQueryString()

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3844


fax4ever:
Merged thanks!

PR #2225, opened 2020-02-28
Solved the SimpleQueryString problem HSEARCH-3845

wklaczynski:
Solved the SimpleQueryString problem
and added support for operation flags
HSEARCH-3845

I have also added support for operation flags. It is now possible to set in the predicate what operations are possible in the query. By default, if you do not call "flags()", all operations are set.

```
                select = scope.predicate().simpleQueryString()
                        .field("shoppingName").boost(0.5f)
                        .matching(searchFullText)
                        .flags().all()
                        .disable(SimpleQueryFlag.PHRASE)
                        .defaultOperator(BooleanOperator.AND);

```
https://hibernate.atlassian.net/projects/HSEARCH/issues/HSEARCH-3845?filter=reportedbyme&orderby=created%20DESC


yrodiere:
Thank you for your work. That's appreciated, however:

* I solved the NPE in another pull request, which has been merged. If no analyzer is defined, we ought to use the `KeywordAnalyzer` (= no analysis).
* The flags you added are nice, but are a completely separate feature and should have their own ticket.
* Any new feature should come with tests.

Could you please open a JIRA ticket for the "flags" feature, and add the appropriate test methods to `org.hibernate.search.integrationtest.backend.tck.search.predicate.SimpleQueryStringSearchPredicateIT`? Thanks.

I'll close this pull request for now, but please do open another one when you have tests for the "flags" feature.

yrodiere:
Ok, I was mistaken, you opened a dedicated ticket for this feature (HSEARCH-3845), it's just that you also solved another one while you were at it (HSEARCH-3844). Can you please remove the fix for HSEARCH-3844 (the NPE) and address exclusively HSEARCH-3845 in your PR?

wklaczynski:
The patch for HSEARCH-3844 was tiny. Based on the creation of StandardAnalyzer if the analyzerOrNormalizer is null.

` if ( analyzerOrNormalizer == null ) {
analyzerOrNormalizer = new StandardAnalyzer();
}`

The second patch is also very small. Relying on a change to the original SimpleQueryParser. Because the solution unnecessarily caused problems.

`Map<String, Float> weights = new LinkedHashMap<>();
 for ( String fieldName : fields.keySet() ) {
		LuceneSimpleQueryStringPredicateBuilderFieldState state = fields.get( fieldName );
		Float boost = state.getBoost();
			if ( boost == null ) {
				boost = 1f;
			}
			weights.put( fieldName, boost );
		}
		SimpleQueryParser queryParser = new SimpleQueryParser( analyzer, weights, -1 );`

Practically the second solution has opened the way to handling the option flag. 
I mixed it up a bit giving ready solutions for option flags as a solution. I will make a new branch only as a solution for HSEARCH-3845 and HSEARCH-3844. And the solution to include flags will stay in this branch. I mixed it up a bit giving ready solutions for option flags as a solution. I will make a new branch only as a solution for HSEARCH-3845 and HSEARCH-3844. And the solution to include flags will stay in this branch. But I will make a new "pull request" for this problem.

PR #2226, opened 2020-02-28
HSEARCH-3845 Prefix query in SimpleQueryString predicate is case sensitive

wklaczynski:
https://hibernate.atlassian.net/browse/HSEARCH-3845

HSEARCH-3845 SimpleQueryParser setting, and field bost activation.

`Map<String, Float> weights = new LinkedHashMap<>();
 for ( String fieldName : fields.keySet() ) {
		LuceneSimpleQueryStringPredicateBuilderFieldState state = fields.get( fieldName );
		Float boost = state.getBoost();
			if ( boost == null ) {
				boost = 1f;
			}
			weights.put( fieldName, boost );
		}
		SimpleQueryParser queryParser = new SimpleQueryParser( analyzer, weights, -1 );`


yrodiere:
Thanks!

As I explained somewhere else, HSEARCH-3844 has already been solved. Can you rebase your work on master? You won't need the change that adds a `StandardAnalyzer` anymore.

yrodiere:
Also, can you please follow [the guidelines for contributing](https://github.com/hibernate/hibernate-search/blob/master/CONTRIBUTING.md#commit)? In particular, commit message should start with the ticket key (here it's `HSEARCH-3845`).

yrodiere:
I removed changes related to HSEARCH-2844, rebased on master, and removed old code that is no longer necessary after your changes.

I'm fine with merging this, because this makes the code simpler, but I still don't understand what this is supposed to fix. The example you gave in the ticket already has a test and does work correctly: see `org.hibernate.search.integrationtest.backend.tck.search.predicate.SimpleQueryStringSearchPredicateIT#prefix_normalizePrefixTerm`.

So, can you provide more information about the problem you encountered?

* The code of your entity (with the annotations such as `@GenericField`, `@FullTextField`, etc.)
* The indexed content (the name of the shop you're trying to find)
* The *exact* code used to create the query

This will allow me to write a reproducer for the problem, then to understand what didn't work, then to check that there aren't any other, related problems that we should try to fix.

yrodiere:
I was finally able to isolate the problem. As I initially suspected, it was caused by a lack of normalization in the prefix query.

I added a test; let's wait for the CI build to finish, then I'll merge.

yrodiere:
Merged. @wklaczynski Thanks for the report and for the fix!

PR #2229, opened 2020-03-02
HSEARCH-3752 Implicit nested predicates

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3752

~~Preview: only match predicate and ElasticSearch backend are done.
If you don't want to have an early look, please ignore this pull request until it isn't a preview anymore.~~
now is ready for review

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchDifferentNestedObjectCompatibilityChecker.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchDifferentNestedObjectCompatibilityChecker.java
@@ -1,0 +1,57 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.elasticsearch.scope.model.impl;
+
+import java.lang.invoke.MethodHandles;
+import java.util.List;
+
+import org.hibernate.search.backend.elasticsearch.logging.impl.Log;
+import org.hibernate.search.util.common.logging.impl.LoggerFactory;
+
+public class ElasticsearchDifferentNestedObjectCompatibilityChecker {
+
+	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
+
+	private final ElasticsearchScopeModel scopeModel;
+
+	private String absoluteFieldPath;
+	private List<String> nestedObjectPath;
+
+	public ElasticsearchDifferentNestedObjectCompatibilityChecker(ElasticsearchScopeModel scopeModel) {
+		this.scopeModel = scopeModel;
+	}
+
+	public void combineAndCheck(String anotherAbsoluteFieldPath) {
+		if ( absoluteFieldPath == null ) {
+			absoluteFieldPath = anotherAbsoluteFieldPath;
+			nestedObjectPath = scopeModel.getNestedPathHierarchy( absoluteFieldPath );
+			return;
+		}
+
+		List<String> anotherNestedObjectPath = scopeModel.getNestedPathHierarchy( anotherAbsoluteFieldPath );
+		if ( !nestedObjectPath.equals( anotherNestedObjectPath ) ) {
+			throw log.simpleQueryStringSpanningMultipleNestedPaths( absoluteFieldPath, getLastPath( nestedObjectPath ), anotherAbsoluteFieldPath,
+					getLastPath( anotherNestedObjectPath )
+			);
+		}
+	}
+
+	public List<String> getNestedObjectPath() {
+		return nestedObjectPath;
+	}
+
+	public boolean isEmpty() {
+		return nestedObjectPath == null || nestedObjectPath.isEmpty();
+	}
+
+	private static String getLastPath(List<String> hierarchy) {
+		if ( hierarchy.isEmpty() ) {
+			return "<<root>>";
+		}
+		return hierarchy.get( hierarchy.size() - 1 );
+	}
+}
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchDifferentNestedObjectCompatibilityChecker.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchDifferentNestedObjectCompatibilityChecker.java
@@ -14,38 +14,40 @@ import org.hibernate.search.util.common.logging.impl.LoggerFactory;
 
 public class ElasticsearchDifferentNestedObjectCompatibilityChecker {
 
+	public static ElasticsearchDifferentNestedObjectCompatibilityChecker empty(ElasticsearchScopeModel scopeModel) {
+		return new ElasticsearchDifferentNestedObjectCompatibilityChecker( scopeModel, null, null );
+	}
+
 	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
 
 	private final ElasticsearchScopeModel scopeModel;
+	private final String fieldPath;
+	private final List<String> nestedPathHierarchy;
 
-	private String absoluteFieldPath;
-	private List<String> nestedObjectPath;
-
-	public ElasticsearchDifferentNestedObjectCompatibilityChecker(ElasticsearchScopeModel scopeModel) {
+	private ElasticsearchDifferentNestedObjectCompatibilityChecker(ElasticsearchScopeModel scopeModel, String fieldPath, List<String> nestedPathHierarchy) {
 		this.scopeModel = scopeModel;
+		this.fieldPath = fieldPath;
+		this.nestedPathHierarchy = nestedPathHierarchy;
 	}
 
-	public void combineAndCheck(String anotherAbsoluteFieldPath) {
-		if ( absoluteFieldPath == null ) {
-			absoluteFieldPath = anotherAbsoluteFieldPath;
-			nestedObjectPath = scopeModel.getNestedPathHierarchy( absoluteFieldPath );
-			return;
+	public ElasticsearchDifferentNestedObjectCompatibilityChecker combineAndCheck(String incomingFieldPath) {
+		List<String> incomingNestedPathHierarchy = scopeModel.getNestedPathHierarchyForField( incomingFieldPath );
+		if ( fieldPath == null ) {
+			return new ElasticsearchDifferentNestedObjectCompatibilityChecker( scopeModel, incomingFieldPath, incomingNestedPathHierarchy );
 		}
 
-		List<String> anotherNestedObjectPath = scopeModel.getNestedPathHierarchy( anotherAbsoluteFieldPath );
-		if ( !nestedObjectPath.equals( anotherNestedObjectPath ) ) {
-			throw log.simpleQueryStringSpanningMultipleNestedPaths( absoluteFieldPath, getLastPath( nestedObjectPath ), anotherAbsoluteFieldPath,
-					getLastPath( anotherNestedObjectPath )
-			);
+		if ( !nestedPathHierarchy.equals( incomingNestedPathHierarchy ) ) {
+			throw log.simpleQueryStringSpanningMultipleNestedPaths( fieldPath, getLastPath( nestedPathHierarchy ), incomingFieldPath, getLastPath( incomingNestedPathHierarchy ) );
 		}
+		return this;
 	}
 
-	public List<String> getNestedObjectPath() {
-		return nestedObjectPath;
+	public List<String> getNestedPathHierarchy() {
+		return nestedPathHierarchy;
 	}
 
 	public boolean isEmpty() {
-		return nestedObjectPath == null || nestedObjectPath.isEmpty();
+		return nestedPathHierarchy == null || nestedPathHierarchy.isEmpty();
 	}
 
 	private static String getLastPath(List<String> hierarchy) {
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
@@ -197,7 +197,7 @@ public class ElasticsearchScopeModel {
 						return nestedDocumentPath1;
 					}
 
-					throw log.conflictingNestedDocumentPathsForProjection(
+					throw log.conflictingNestedDocumentPaths(
 							absoluteFieldPath, nestedDocumentPath1.orElse( null ), nestedDocumentPath2.orElse( null ), getIndexesEventContext() );
 				} )
 				.orElse( Optional.empty() );
@@ -215,7 +215,7 @@ public class ElasticsearchScopeModel {
 						return nestedDocumentPath1;
 					}
 
-					throw log.conflictingNestedDocumentPathHierarchyForProjection(
+					throw log.conflictingNestedDocumentPathHierarchy(
 							absoluteFieldPath, nestedDocumentPath1.orElse( null ), nestedDocumentPath2.orElse( null ), getIndexesEventContext() );
 				} )
 				.orElse( Optional.empty() );
@@ -233,7 +233,7 @@ public class ElasticsearchScopeModel {
 						return nestedDocumentPath1;
 					}
 
-					throw log.conflictingNestedDocumentPathHierarchyForProjection(
+					throw log.conflictingNestedDocumentPathHierarchy(
 							absoluteObjectPath, nestedDocumentPath1.orElse( null ), nestedDocumentPath2.orElse( null ), getIndexesEventContext() );
 				} )
 				.orElse( Optional.empty() );
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
@@ -205,11 +205,11 @@ public class ElasticsearchScopeModel {
 		return nestedDocumentPath.orElse( null );
 	}
 
-	public List<String> getNestedPathHierarchy(String absoluteFieldPath) {
+	public List<String> getNestedPathHierarchyForField(String absoluteFieldPath) {
 		Optional<List<String>> nestedDocumentPath = indexModels.stream()
 				.map( indexModel -> indexModel.getFieldNode( absoluteFieldPath ) )
 				.filter( Objects::nonNull )
-				.map( fieldNode -> Optional.ofNullable( fieldNode.getNestedPathHierarchy() ) )
+				.map( node -> Optional.ofNullable( node.getNestedPathHierarchy() ) )
 				.reduce( (nestedDocumentPath1, nestedDocumentPath2) -> {
 					if ( Objects.equals( nestedDocumentPath1, nestedDocumentPath2 ) ) {
 						return nestedDocumentPath1;
@@ -227,7 +227,7 @@ public class ElasticsearchScopeModel {
 		Optional<List<String>> nestedDocumentPath = indexModels.stream()
 				.map( indexModel -> indexModel.getObjectNode( absoluteObjectPath ) )
 				.filter( Objects::nonNull )
-				.map( fieldNode -> Optional.ofNullable( fieldNode.getNestedPathHierarchy() ) )
+				.map( node -> Optional.ofNullable( node.getNestedPathHierarchy() ) )
 				.reduce( (nestedDocumentPath1, nestedDocumentPath2) -> {
 					if ( Objects.equals( nestedDocumentPath1, nestedDocumentPath2 ) ) {
 						return nestedDocumentPath1;
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
@@ -222,4 +222,22 @@ public class ElasticsearchScopeModel {
 
 		return nestedDocumentPath.orElse( null );
 	}
+
+	public List<String> getNestedPathHierarchyForObject(String absoluteObjectPath) {
+		Optional<List<String>> nestedDocumentPath = indexModels.stream()
+				.map( indexModel -> indexModel.getObjectNode( absoluteObjectPath ) )
+				.filter( Objects::nonNull )
+				.map( fieldNode -> Optional.ofNullable( fieldNode.getNestedPathHierarchy() ) )
+				.reduce( (nestedDocumentPath1, nestedDocumentPath2) -> {
+					if ( Objects.equals( nestedDocumentPath1, nestedDocumentPath2 ) ) {
+						return nestedDocumentPath1;
+					}
+
+					throw log.conflictingNestedDocumentPathHierarchyForProjection(
+							absoluteObjectPath, nestedDocumentPath1.orElse( null ), nestedDocumentPath2.orElse( null ), getIndexesEventContext() );
+				} )
+				.orElse( Optional.empty() );
+
+		return nestedDocumentPath.orElse( null );
+	}
 }
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchSearchNestedPredicateBuilder.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchSearchNestedPredicateBuilder.java
@@ -6,7 +6,6 @@
  */
 package org.hibernate.search.backend.elasticsearch.search.predicate.impl;
 
-import java.util.LinkedList;
 import java.util.List;
 
 import org.hibernate.search.backend.elasticsearch.gson.impl.JsonAccessor;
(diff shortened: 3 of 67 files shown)


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchSearchPredicateBuilder.java
```diff
@@ -48,7 +62,31 @@ public final JsonObject build(ElasticsearchSearchPredicateContext context) {
 		}
 
 		JsonObject result = doBuild( context, outerObject, innerObject );
-		return ( withConstantScore ) ? applyConstantScore( result ) : result;
+		if ( withConstantScore ) {
+			result = applyConstantScore( result );
+		}
+
+		if ( !context.isExplicitNested() && nestedPathHierarchy != null && !nestedPathHierarchy.isEmpty() ) {
```
yrodiere: I believe this could be more fine-grained: instead of checking whether there's explicit nesting, compare the closest nested predicate with the nested path hierarchy of this predicate.

For example, let's assume you have a field with two levels of nesting: `nested1.nested2.stringField`. If the user applies a nested predicate on `nested1` only, and forgets about `nested2` (which could be understandable in some situations) ... then we could add the nesting on `nested2` automatically when we detect the "closest nested predicate" doesn't match the last path in `nestedPathHierarchy`.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchSearchPredicateBuilder.java
```diff
@@ -6,22 +6,36 @@
...
+import java.util.List;
+
 import org.hibernate.search.backend.elasticsearch.gson.impl.JsonAccessor;
 import org.hibernate.search.engine.search.predicate.spi.SearchPredicateBuilder;
 
 import com.google.gson.JsonObject;
 
 
-
 public abstract class AbstractElasticsearchSearchPredicateBuilder
```
yrodiere: Some predicates, such as the `id` and `bool` predicate, have nothing to do with fields.

I believe it would make more sense to have a dedicated abstract class for predicates that support nested. Just like you did with `AbstractElasticsearchSearchNestedSortBuilder`, which extends `AbstractElasticsearchSearchSortBuilder` and adds nesting support.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchNestedPredicateBuilder.java
```diff
@@ -35,6 +35,8 @@ public void nested(ElasticsearchSearchPredicateBuilder nestedBuilder) {
 	@Override
 	protected JsonObject doBuild(ElasticsearchSearchPredicateContext context,
 			JsonObject outerObject, JsonObject innerObject) {
+		context.explicitNested();
```
yrodiere: I think we took another path in `LuceneSearchPredicateContext`: the context is immutable, and we create another one where needed. Which seems safer to me.

Here you need to forward some internal state too, so you could have something like this instead:

```
ElasticsearchSearchPredicateContext childContext = context.nest( absoluteFieldPath ); // Returns a new context
```

And then below:

```
QUERY_ACCESSOR.set( innerObject, nestedBuilder.build( childContext ) ); 
```

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/predicate/impl/AbstractLuceneSearchPredicateBuilder.java
```diff
@@ -52,4 +54,22 @@ public final Query build(LuceneSearchPredicateContext context) {
...
+			Function<LuceneSearchPredicateContext, Query> baseBuild) {
+		if ( furtherImplicitNestedSteps.isEmpty() ) {
+			return baseBuild.apply( context );
+		}
+
+		if ( furtherImplicitNestedSteps.size() == 1 ) {
+			String lastStep = furtherImplicitNestedSteps.get( 0 );
+			LuceneSearchPredicateContext childContext = new LuceneSearchPredicateContext( lastStep );
+			// super.build( childContext ) must be called only from the very last nest step
+			return LuceneNestedPredicateBuilder.doBuild( context, lastStep, baseBuild.apply( context ) );
```
yrodiere: You don't use the child context. This seems wrong, and there's probably a missing test, otherwise this would have been detected by a failing tests.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchSearchNestedPredicateBuilder.java
```diff
@@ -0,0 +1,58 @@
...
+		return result;
+	}
+
+	private JsonObject applyImplicitNested(JsonObject partialResult, List<String> nestedPathHierarchy,
+			ElasticsearchSearchPredicateContext context) {
+		if ( nestedPathHierarchy.isEmpty() ) {
+			return partialResult;
+		}
+
+		String lastPath = nestedPathHierarchy.remove( 0 );
```
yrodiere: This line and the ones below could be converted to a simple loop instead of using recursion + a mutable collection. It would probably be easier to read, and there would be one fewer LinkedList allocation.

E.g.:


```
	JsonObject result = partialResult;
	for ( String nestedPath : nestedPathHierarchy ) {
		if ( context.isNested( nestedPath ) ) {
			// skip this implicit nested
			continue;
		}
    
		JsonObject innerObject = new JsonObject();

		PATH_ACCESSOR.set( innerObject, lastPath );
		QUERY_ACCESSOR.set( innerObject, applyImplicitNested( partialResult, nestedPathHierarchy, context ) );

		JsonObject outerObject = new JsonObject();
		outerObject.add( "nested", innerObject );

		result = outerObject; 
	}
```

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchSearchPredicateContext.java
```diff
@@ -6,18 +6,46 @@
...
+		HashSet<String> paths = new HashSet<>( explicitNestedPaths );
+		paths.add( path );
+		return new ElasticsearchSearchPredicateContext( sessionContext, paths );
+	}
+
+	public boolean isExplicitNested(String path) {
+		// Checking explicitNestedPaths#contins( path ) could be not enough
+		// for cases where explicit nest is used only for some nested level of the hierarchy:
+		// see NestedSearchPredicateIT#search_nestedOnTwoLevels_onlySecondLevel.
+		for ( String explicitNestedPath : explicitNestedPaths ) {
```
yrodiere: You don't need to keep a set of all nested paths. Just keeping the last one will be enough, since it's supposed to be prefixed with the previous ones anyway.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
```diff
@@ -222,4 +222,22 @@ public String getNestedDocumentPath(String absoluteFieldPath) {
...
+		Optional<List<String>> nestedDocumentPath = indexModels.stream()
+				.map( indexModel -> indexModel.getObjectNode( absoluteObjectPath ) )
+				.filter( Objects::nonNull )
+				.map( fieldNode -> Optional.ofNullable( fieldNode.getNestedPathHierarchy() ) )
+				.reduce( (nestedDocumentPath1, nestedDocumentPath2) -> {
+					if ( Objects.equals( nestedDocumentPath1, nestedDocumentPath2 ) ) {
+						return nestedDocumentPath1;
+					}
+
+					throw log.conflictingNestedDocumentPathHierarchyForProjection(
```
yrodiere: You should probably rename this method and adapt the message.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/predicate/impl/AbstractLuceneSearchNestedPredicateBuilder.java
```diff
@@ -6,23 +6,59 @@
...
+		if ( furtherImplicitNestedSteps.size() == 1 ) {
+			String lastStep = furtherImplicitNestedSteps.get( 0 );
+			LuceneSearchPredicateContext childContext = new LuceneSearchPredicateContext( lastStep );
+			// super.build( childContext ) must be called only from the very last nest step
+			return LuceneNestedPredicateBuilder.doBuild( context, lastStep, super.build( childContext ) );
+		}
+
+		String step = furtherImplicitNestedSteps.remove( 0 );
+		LuceneSearchPredicateContext childContext = new LuceneSearchPredicateContext( step );
+		return LuceneNestedPredicateBuilder.doBuild( context, step, applyImplicitNestedSteps( furtherImplicitNestedSteps, childContext ) );
```
yrodiere: Same comment as for the Elasticsearch backend: a loop would be much clearer than recursion here.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/scope/model/impl/LuceneDifferentNestedObjectCompatibilityChecker.java
```diff
@@ -0,0 +1,57 @@
...
+	private String absoluteFieldPath;
+	private List<String> nestedObjectPath;
+
+	public LuceneDifferentNestedObjectCompatibilityChecker(LuceneScopeModel scopeModel) {
+		this.scopeModel = scopeModel;
+	}
+
+	public void combineAndCheck(String anotherAbsoluteFieldPath) {
+		if ( absoluteFieldPath == null ) {
+			absoluteFieldPath = anotherAbsoluteFieldPath;
```
yrodiere: I'm pretty sure introducing mutability here will come back and bite us... Can't you make it immutable, like the other checkers?

documentation/src/main/asciidoc/search-dsl-predicate.asciidoc
```diff
@@ -287,7 +287,8 @@ on object fields that are known to have at least one sub-field that is never nul
 [IMPORTANT]
 ====
 For object fields with <<mapper-orm-indexedembedded-storage-nested,`NESTED` storage>>,
-the `exists` predicate needs to be wrapped in a <<search-dsl-predicate-nested,`nested` predicate>>.
+the `exists` predicate needs to be wrapped in a <<search-dsl-predicate-nested,`nested` predicate>>,
+or the <<search-dsl-predicate-nested-implicit,`nested` implicit form>> can be used.
 ====
```
yrodiere: I would simply remove this admonition. It's not important information if it works out of the box without the user even thinking about it.

documentation/src/main/asciidoc/search-dsl-predicate.asciidoc
```diff
@@ -549,6 +550,24 @@ and *another* of which has a last name matching `asimov`,
...
+----
+<1> The nested predicate is created implicitly, since target fields here belong to a nested object.
+<2> The author must have a first name matching `isaac`.
+<3> The author must have a last name matching `asimov`.
+<4> All returned hits will be books for which at least one author
+has a first name matching `isaac` and a last name matching `asimov`.
+Books that happen to have multiple authors,
+one of which has a first name matching `isaac`
+and *another* of which has a last name matching `asimov`,
+will *not* match.
```
yrodiere: This is wrong, though. They will match, because we apply the nested predicate separately to each `match` predicate.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
yrodiere: Never mind, you did it in a later commit :)

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchSearchPredicateBuilder.java
yrodiere: I believe this comment still stands?

fax4ever: I think that the test `ImplicitNestedSearchPredicateIT.javanested_X2_explicit_implicit` covers the case. Doesn't it?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/predicate/impl/AbstractLuceneSearchPredicateBuilder.java
fax4ever: Oh yes, I don't. Thank you. I'm going to add a test for that.

fax4ever: Ok, I checked it...

Basically, the passed (child) context would be used by another implicit or explicit nested predicates.
So if there were a subsequent nest, it would benefit from it.
I think it's not impossible to have any nested after (nested) an implicit nested, because implicit nested always reaches the leaf (field) of the tree.
But let me know if you agree with that.

Anyway, I have to remove the unused local variable, so I'll do it.

fax4ever: That is the same reason why `AbstractElasticsearchSearchNestedPredicateBuilder#applyImplicitNestedStep` does not need to update the context.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/ImplicitNestedSearchPredicateIT.java
```diff
@@ -0,0 +1,254 @@
...
+		plan.add( referenceProvider( DOCUMENT_1 ), document -> {
+			DocumentElement nestedDocument = document.addObject( indexMapping.nested );
+			nestedDocument.addValue( indexMapping.nestedString, ANY_STRING );
+			nestedDocument.addValue( indexMapping.nestedNumeric, ANY_INTEGER );
+			nestedDocument.addValue( indexMapping.nestedText, SOME_PHRASE_TEXT );
+			nestedDocument.addValue( indexMapping.nestedGeo, G11 );
+
+			DocumentElement nestedDocumentX2 = nestedDocument.addObject( indexMapping.nestedX2 );
+			nestedDocumentX2.addValue( indexMapping.nestedX2Numeric, ANY_INTEGER );
+		} );
```
yrodiere: You may want to index another document, just to be sure that predicates are actually taken into account... You would only expect document 1 to turn up in search hits.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchSearchPredicateBuilder.java
yrodiere: Looks like it. Thanks.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchSearchNestedPredicateBuilder.java
fax4ever: The example runs `applyImplicitNested( partialResult, nestedPathHierarchy, context ) ` inside the loop. Of course it always possible to convert a recursion with a loop but I don't think that the resulting code would be simplified by that. Do you?

fax4ever: Maybe traversing the `nestedPathHierarchy` in the inverse order would work. I'm trying now...

yrodiere: > Do you?

Yes. If I compare this recursion solution with what you did in `org.hibernate.search.backend.elasticsearch.search.sort.impl.AbstractElasticsearchSearchNestedSortBuilder#enrichInnerObject`, I must say that other looping solution is much clearer.

I suppose the problem is more complex, though... I was just hoping there would be a clearer solution :)

fax4ever: It seems to be working:

``` java
private static JsonObject applyImplicitNestedStepIt(JsonObject partialResult, List<String> nestedPathHierarchy, ElasticsearchSearchPredicateContext context) {
	JsonObject result = partialResult;

	// traversing the nestedPathHierarchy in the inverted order
	for ( int i = 0; i < nestedPathHierarchy.size(); i++ ) {
		String path = nestedPathHierarchy.get( nestedPathHierarchy.size() - 1 - i );
		if ( context.isExplicitNested( path ) ) {
			// skip this implicit nested
			continue;
		}

		JsonObject innerObject = new JsonObject();

		PATH_ACCESSOR.set( innerObject, path );
		QUERY_ACCESSOR.set( innerObject, result );

		JsonObject outerObject = new JsonObject();
		outerObject.add( "nested", innerObject );
		result = outerObject;
	}

	return result;
}
```

fax4ever: The trick is the inverted traversing order... So that we can reuse the nested JSON object

documentation/src/main/asciidoc/search-dsl-predicate.asciidoc
fax4ever: It was the first thing I had done, then I changed it :)

documentation/src/main/asciidoc/search-dsl-predicate.asciidoc
fax4ever: You're right. Let's update the statement.

fax4ever:
Thank you @yrodiere.
Comments addressed.
Is there anything else I can do here?

fax4ever:
I ran a built for all support environments: https://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-2229/10/

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/ExistsSearchPredicateIT.java
```diff
@@ -224,6 +225,7 @@ public void inFlattenedObject_withDocValues() {
 	 * Querying fields of a nested object without a nested predicate should result in no result at all.
 	 */
 	@Test
+	@Ignore // TODO HSEARCH-3752 restore the test with the new assumptions
```
yrodiere: TODO?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/predicate/impl/AbstractLuceneSearchNestedPredicateBuilder.java
```diff
@@ -0,0 +1,47 @@
...
+		this.nestedPathHierarchy = nestedPathHierarchy;
+	}
+
+	@Override
+	public final Query build(LuceneSearchPredicateContext context) {
+		List<String> nestedSteps = implicitNestedSteps( context, nestedPathHierarchy );
+		return applyImplicitNestedSteps( nestedSteps, context, super::build );
+	}
+
+	public static List<String> implicitNestedSteps(LuceneSearchPredicateContext context, List<String> nestedPathHierarchy) {
```
yrodiere: I'm not sure I get the value of this method, and in particular the value of copying the list of nested paths.
You managed to do without copying the list in the Elasticsearch backend, by simply looping over *all* paths and skipping the ones that are not relevant. Can't we do the same here?
Deep down, it boils down to simply using code similar to the Elasticsearch backend.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchSearchNestedPredicateBuilder.java
```diff
@@ -0,0 +1,58 @@
...
+		return result;
+	}
+
+	public static JsonObject applyImplicitNested(JsonObject partialResult, List<String> nestedPathHierarchy, ElasticsearchSearchPredicateContext context) {
+		JsonObject result = partialResult;
+
+		// traversing the nestedPathHierarchy in the inverted order
+		for ( int i = 0; i < nestedPathHierarchy.size(); i++ ) {
+			String path = nestedPathHierarchy.get( nestedPathHierarchy.size() - 1 - i );
+			if ( context.isExplicitNested( path ) ) {
```
yrodiere: Knowing that the nested path hiarchy, in reverse order, will look like this:

* `nested1.nested2.nested3`
* `nested1.nested2`
* `nested1`

... and that `isExplicitNested()` checks `theExplicitPath.startsWith(theHierarchyComponent)`.

You will have something like that:

* First iteration: `"nested1".startsWith("nested1.nested2.nested3")` => `false`
* Second iteration: `"nested1".startsWith("nested1.nested2")` => `false`
* Second iteration: `"nested1".startsWith("nested1")` => `true`

It seems to me that you could just make `isExplicitNested()` check `theExplicitPath.equals(theHierarchyComponent)`. From what I understand, this will be strictly equivalent, and simpler.

Better yet, you could make the code more similar to the Lucene backend: remove `isExplicitNested()` and just write `context.getNestedPath().equals( path )` here.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/predicate/impl/AbstractLuceneSearchPredicateBuilder.java
yrodiere: Agreed, it makes sense.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/scope/model/impl/LuceneDifferentNestedObjectCompatibilityChecker.java
yrodiere: Looks like this wasn't addressed?

fax4ever: You're right. Doing now...

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/ImplicitNestedSearchPredicateIT.java
fax4ever: I added the document `DOCUMENT_2` that basically does not match any predicate, with the exception of `matchAll` and `exists` predicates. See the commit:

> HSEARCH-3752 Add a non-matching document to the test case

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/ExistsSearchPredicateIT.java
fax4ever: I forgot it!

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchSearchNestedPredicateBuilder.java
fax4ever: Yes, we got this additional benefit by processing in reverse order. I will add the test case you described above.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/predicate/impl/AbstractLuceneSearchNestedPredicateBuilder.java
fax4ever: You're right, we can just use the context to skip any explicit already performed nested step.

fax4ever:
@yrodiere I think I've done the changes.
I spent some time adding some extra test for complex cases.
If there is something else to improve, I'll be happy to do that.

fax4ever:
Started a full build: https://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-2229/12/

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchDifferentNestedObjectCompatibilityChecker.java
```diff
@@ -0,0 +1,57 @@
...
+	private List<String> nestedObjectPath;
+
+	public ElasticsearchDifferentNestedObjectCompatibilityChecker(ElasticsearchScopeModel scopeModel) {
+		this.scopeModel = scopeModel;
+	}
+
+	public void combineAndCheck(String anotherAbsoluteFieldPath) {
+		if ( absoluteFieldPath == null ) {
+			absoluteFieldPath = anotherAbsoluteFieldPath;
+			nestedObjectPath = scopeModel.getNestedPathHierarchyForField( absoluteFieldPath );
```
yrodiere: Well... Same as Lucene: this probably shouldn't be mutable :)

fax4ever:
Done. Let's wait for the CI than I'll merge it. Thanks @yrodiere

PR #2231, opened 2020-03-03
HSEARCH-3849 Create-or-validate schema management by default

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3849

Based on #2230, which should be merged first.

### DIFF
--- a/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/schema/management/SchemaManagementStrategyName.java
+++ b/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/schema/management/SchemaManagementStrategyName.java
@@ -54,6 +54,22 @@ public enum SchemaManagementStrategyName {
 
 	/**
 	 * A strategy that creates missing indexes and their schema on startup,
+	 * and validates the schema of existing indexes.
+	 * <p>
+	 * Note that creating indexes and their schema will not populate the indexed data:
+	 * newly created indexes will always be empty.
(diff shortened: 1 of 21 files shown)


mapper/orm/src/main/java/org/hibernate/search/mapper/orm/schema/management/SchemaManagementStrategyName.java
```diff
@@ -0,0 +1,134 @@
...
+	 * newly created indexes will always be empty.
+	 * To populate indexes with pre-existing data, use mass indexing.
+	 * <p>
+	 * With Elasticsearch only, an exception will be thrown on startup if some indexes already exist
+	 * but their schema does not match the requirements of the Hibernate Search mapping:
+	 * missing fields, fields with incorrect type, missing analyzer definitions or normalizer definitions, ...
+	 * <p>
+	 * With Lucene, the schema of indexes is not validated.
+	 */
+	CREATE_OR_VALIDATE( "create-or-validate" ),
```
fax4ever: Maybe it's not 100% clear that with Lucene any pre-existing schema will provoke an exception.

yrodiere: It will, in fact, always pass as long as the index exists. I'll clarify.

yrodiere:
Merged, thanks!

PR #2240, opened 2020-03-09
HSEARCH-3857 In some cases the SearchIndexingPlan generate a Concurre…

acucumel:
…ntModificiationException if an embedded entities is not listed in the addOrUpdate or on specific order (if only the last entity is concern this case)


acucumel:
> Thanks a lot for working on this!
> 
> I just pushed a few changes:
> 
> * Split your commit in two, one for the test and one for the fix. I needed that in order to (temporarily) revert the fix and reproduce the problem on my machine.
> * Simplified a few things in the test (not strictly necessary, but will certainly help if one day we have to debug a regression)
> * Added a fix and test for the same problem in `PojoIndexingPlanImpl`.
> * Removed an unnecessary collection copy in `PojoContainedTypeIndexingPlan` (not strictly necessary either, but it will spare us a few memory allocations).
> 
> I'll wait for the CI build to succeed and will merge.

Perfect, thanks :)

yrodiere:
Merged, thanks!

PR #2224, opened 2020-02-27
Resolve Sorting MultiValues Problem HSEARCH-3839

wklaczynski:

PR #2244, opened 2020-03-16
HSEARCH-3103 Sorts on multi-valued fields

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3103

Based on #2243, which should be merged first.

This supersedes #2224. I kept the core of the original solution by @wklaczynski , with the following additions:

* Rewrote the tests:
  * They can now be run from the commandline: they used to only work when run from within the IDE)
  * They actually test the feature: they used to always pass, because we expected the documents to be returned in the exact order they were indexed, which is what happens when the sort doesn't do anything at all... and incidentally, that's exactly the case we were in, because a small bug in the implementation.
  * They test all field types (not just integer/long/double/float).
* Cleaned up the code: removed some dead code, made use of pre-existing code instead of rewriting things like multi-valued doc values classes.
* Fixed several bugs in the implementation that I noticed thanks to the tests
* Added more explicit error messages in some edge cases (attempt to sort on a median of a field in a nested document, to sort on the sum of a temporal field, ...)
* Changed the default sort mode from "always MIN" to "MIN for ascending sort, MAX for descending sort", to match Elasticsearch's behavior.
* Implemented sorts on multi-valued string fields
* Implemented distance sorts on multi-valued geo-point fields
* Documented the feature


documentation/src/main/asciidoc/search-dsl-sort.asciidoc
```diff
@@ -208,3 +193,132 @@ include::{sourcedir}/org/hibernate/search/documentation/search/sort/SortDslIT.ja
...
+
+The order controlled explicitly through the following options:
+
+* `.asc()` for an ascending order.
+* `.desc()` for a descending order.
+* `.order(...)` for an order defined by the given argument: `SortOrder.ASC`/`SortOrder.DESC`.
+
+Below are a few examples with the <<search-dsl-sort-field,field sort>>.
+
+.Sorting by field values in explicitly ascending order with `asc()
```
fax4ever: I think that there is a missing apostrophe at the end

documentation/src/main/asciidoc/search-dsl-sort.asciidoc
```diff
@@ -208,3 +193,132 @@ include::{sourcedir}/org/hibernate/search/documentation/search/sort/SortDslIT.ja
...
+|`SortMode.AVG`
+|Computes the https://en.wikipedia.org/wiki/Arithmetic_mean[arithmetic mean] of all values for each document
+and picks that average for comparison with other documents.
+|Numeric and temporal fields (`long`, `LocalDate`, ...), <<search-dsl-sort-distance,distance>>.
+|Text fields (`String`, ...).
+|`SortMode.MEDIAN`
+|Computes the https://en.wikipedia.org/wiki/Median[median] of all values for each document,
+and picks that median for comparison with other documents.
+|Numeric and temporal fields (`long`, `LocalDate`, ...), <<search-dsl-sort-distance,distance>>.
+|Text fields (`String`, ...).
```
fax4ever: Maybe in the future we could support the median of a string keyword field.

yrodiere: It's not supported by Elasticsearch. I didn't check, but the reason is probably about performance. But yeah, maybe, in the future, if someone wants it.

yrodiere:
Thanks. I addressed your comments and force-pushed. I will merge after we merge #2229 , because I expect some conflicts.

yrodiere:
Rebased on master just now and started a build for all supported environments: https://ci.hibernate.org/job/hibernate-search/job/PR-2244/9/

yrodiere:
... and there's a bug in ES 5.6. I disabled some tests for ES 5.6. Second attempt: https://ci.hibernate.org/job/hibernate-search/job/PR-2244/11/

PR #2248, opened 2020-03-25
HSEARCH-3869 java.lang.IndexOutOfBoundsException when an Elasticsearch bulk work fails

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3869


yrodiere:
Merged, thanks

PR #2255, opened 2020-03-31
HSEARCH-3846 Filtering multi values for sort and aggregation rebase

wklaczynski:
https://hibernate.atlassian.net/browse/HSEARCH-3846

Fixing the problem of aggregating multiple HSEARCH-3846 values. Added filtering of nested filter minister values for sorting and aggregation. A solution for sorting multiple values in user provided mode. 

Rebase has been fixed to the master branch.


yrodiere:
Thanks for the rebase, but this PR breaks almost every test related to sorts or aggregations: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2255/5/tests

yrodiere:
@wklaczynski This looks a bit complex, and is apparently a bit hard to get in working shape.

May I suggest that you just remove all the changes related to aggregations, and focus on sorts as a first step? If you reduce the scope of this PR to **just sorts**, the changes to low-level classes related to docvalues become unnecessary (it's just a matter of filtering children, after all) and the PR will be much simpler to debug.

Then as a second step we can consider changes to aggregations, in a second PR addressing a separate ticket.

The aggregation changes are unlikely to get merged anyway. First because they are out of scope for HSEARCH-3846, but also because the solution you chose cannot be implemented easily and reliably on Elasticsearch. As I already mentioned before, Elasticsearch does not pick *one* value per document for aggregations, but instead considers all values of each document. That's a sensible default, and it'd be better to have the same default on the Lucene backend.

wklaczynski:
Aggregation is together with sorting. Filtering is a tiny problem. I've solved all problems before. I just have to get the tests done. These are only minor errors, I think that it should start working and meet all test conditions in a moment.

wklaczynski:
I have implemented the default condition none for aggregation if none is set to mode. I saw that you did there max. If no value for multivalue is set for aggregation, it will return the count of all possible values without calculating any function. Unfortunately, elasticsearch does not support the aggregation filter. But since Lucene can do it after modifications. I suspect that they will make some extra money, we will be the first in this matter.

wklaczynski:
I saw that you transferred the type name to SortMode. Aggregation is also a type for many values, now virtually duplicated. I didn't change it, I left it for SortMode, and for aggregation is MultiValue. Maybe it would be good to consider to establish a common type name?

wklaczynski:
It defaults to multiple values for aggregation. Of course, after previous filtering, if the filter has been defined. Just selecting the function causes that the values will be converted to one value according to the functions MIN, MAX, AVG, etc. ...

wklaczynski:
Myślę, że problem gotowy do zamknięcia.

wklaczynski:
Although it now automatically builds nested questions, sometimes it is necessary to define such a query. Especially when using native lucene questions. In order not to build separate queries for sorting, aggregation and search, this query when using a nested predicate should be ignored when building the same nested context.

```java
            PermissionQuery allowQuery = new PermissionQuery("prices.permissions", operation, user, true);
            PermissionQuery dannyQuery = new PermissionQuery("prices.permissions", operation, user, false);

            BooleanQuery query = new BooleanQuery.Builder()
                    .add(allowQuery, BooleanClause.Occur.MUST)
                    .add(dannyQuery, BooleanClause.Occur.MUST_NOT)
                    .build();

            filter = predicate.nested().objectField("prices")
                    .nest(f -> f.extension(LuceneExtension.get()).fromLuceneQuery(query))
                    .toPredicate();

            sort.add(sorter.field("prices.bruttoPrice_sort")
                    .asc()
                    .mode(SortMode.MIN)
                    .filter(filter));

            SearchQuery<Assortment> query = session.search(scope)
                    .where((f) -> f.bool().must(select)
                    .filter(filter))
                    .sort(sort.toSort())
                    .toQuery();

```
Using the same filter for sort will be possible because nesting will be ignored. But nesting is recognized when using the same filter for search.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/ValueSortField.java
```diff
@@ -0,0 +1,79 @@
...
+import org.apache.lucene.search.FieldComparatorSource;
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.SortField;
+import org.apache.lucene.search.join.ScoreMode;
+
+/**
+ *
+ * @author Waldemar Kłaczyński
+ */
+public class ValueSortField extends SortField {
```
yrodiere, line 18: This is unrelated to the ticket and we can expose filtered sorts without this.

I will not merge this as part of this PR, but you can find the corresponding changeset [here](https://github.com/yrodiere/hibernate-search/commit/e0f771fa20f372632a03ab48ed2bb0420a241f25) should you ever need it.

engine/src/main/java/org/hibernate/search/engine/search/aggregation/dsl/RangeAggregationOptionsStep.java
```diff
@@ -15,6 +20,44 @@
...
+	extends AggregationFinalStep<A> {
 
+	/**
+	 * Start describing the behavior of this sort when a document do have
+	 * multi values for the targeted field.
+	 *
+	 * @param mode The renge.
+	 * @return The next step.
+	 */
+	S mode(MultiValue mode);
```
yrodiere, line 32: 1. Exposing a "mode" is not necessary to implement filters on aggregations. Those are two unrelated problems, which should be addressed in two separate tickets and pull requests.
2. Exposing a "mode" for aggregations is not a good idea, as I explained multiple times. It cannot be supported for Elasticsearch, in particular, so if we ever expose it, it will have to be in Lucene-specific extensions to the DSL.
3. Correctly implementing aggregations on multi-valued fields is indeed necessary, but not necessary for this PR. I will keep your code around and submit another PR later, with tests.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/docvalues/impl/DoubleMultiValuesToSingleValuesSource.java
```diff
@@ -176,21 +224,85 @@ public boolean advanceExact(int parentDoc) throws IOException {
...
+					result.add( values.nextValue() );
+				}
+			}
+		}
+
+		Collections.sort( result );
+		return result;
+	}
+
+	protected NumericDoubleValues replaceScores(NumericDoubleValues values, DoubleValues scores) {
```
yrodiere, line 264: This is not necessary and it's actually the source of your troubles related to missing values on sorts. You just don't need this hack, and without this hack you don't need your commits "Fix sort missing values".

I will remove this before submitting a new PR.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/docvalues/impl/DoubleMultiValuesToSingleValuesSource.java
```diff
@@ -155,11 +197,17 @@ protected NumericDoubleValues select(SortedNumericDoubleDocValues values, final
...
 
 			int lastSeenParentDoc = -1;
-			double lastEmittedValue = -1;
+			private Iterator<Double> value;
+			private List<Double> all;
 
 			@Override
 			public double doubleValue() {
-				return lastEmittedValue;
+				return value != null ? value.next() : -1;
```
yrodiere, line 205: NumericDoubleValues are not supposed to be multi-valued, on contrary to SortedNumericDocValues.
This hijacks and interface that was not intended for this purpose at all.

Worse, in aggregations with a multi-value mode of none, this solution takes a `SortedNumericDocValues` (multi-valued), converts it to a `NumericDoubleValues` (single-valued) thanks to `DoublieMultiValuesToSingleValuesSource`, and then uses this hack to actually extract multiple values and expose them as multi-valued in `LongMultiDoubleValuesSource`...

This is clearly sub-optimal, and a (way) better solution would be to simply use `SortedNumericDocValues` directly in `LongMultiDoubleValuesSource`.

I will work on removing this hack before I submit a PR for aggregations on multi-valued fields.

wklaczynski: It is not necessary for sorting, it always assumes cout = 1. But for aggregation, when it should select all non-redundant values. This simple method eliminates code redundancy. Earlier I had a different solution, but ultimately it turned out to be optimal. Look at the aggregation problem when multivalue is none. Currently, Lucene does not solve this problem. multivalue is new at all. I suspect that over time they will solve, so far a simple modification allows aggregation of many values. This is not a great momentary modification. So far, it makes us ahead of lucene, with time lucene will probably find the optimal solution and it will be possible in higher versions of lucene to go back and use the original solution.

wklaczynski: Also, I was looking for a different solution, but as of today, it required the least effort, copying the overall implementation for aggregation and adding loops for many values. Currently, thanks to this I am working on a slightly modified package with the "guy" lucene. Maybe in the future lucene will solve the problem, so far not much is happening in 8.5. And the patch bypasses the problem and makes aggregation work properly. Removing the patch will prevent aggregation of many values without converting it to SUM, AVG, MIN, MAX as NONE aggregation.

wklaczynski: One could try to modify "FacetsCollector" for non-converted multiple values accordingly. Then it could have been done without doing and modifying the classes from the lucene "fecets" package. But it would be rather difficult to do it, maybe easier for nested values, but for flat relations it would be a big problem.

PR #2259, opened 2020-04-06
HSEARCH-3883 NullPointerException when building a SimpleQueryString predicate

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3883

I didn't add a test because it turns out that the code triggering the NPE [was relying on unsupported behavior](https://discourse.hibernate.org/t/hibernate-search-6-0-0-beta6-simplequerydsl-breaking/3977/6?u=yrodiere), but I think this cleanup is a good think regardless. We can ask ourselves whether we want to officially support that use case later.

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
@@ -7,6 +7,7 @@
 package org.hibernate.search.backend.elasticsearch.scope.model.impl;
 
 import java.lang.invoke.MethodHandles;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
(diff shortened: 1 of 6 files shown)


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/scope/model/impl/LuceneScopeModel.java
```diff
@@ -250,7 +251,7 @@ public String getNestedDocumentPath(String absoluteFieldPath) {
 				} )
 				.orElse( Optional.empty() );
 
-		return nestedDocumentPath.orElse( null );
+		return nestedDocumentPath.orElse( Collections.emptyList() );
```
fax4ever, line 254: Good. Maybe we can do the same change on the next method as well.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
```diff
@@ -220,7 +221,7 @@ public String getNestedDocumentPath(String absoluteFieldPath) {
 				} )
 				.orElse( Optional.empty() );
 
-		return nestedDocumentPath.orElse( null );
+		return nestedDocumentPath.orElse( Collections.emptyList() );
```
fax4ever, line 224: Same here. We can make the subsequent method return an empty list in case of no nested parent.

yrodiere:
Thanks Fabio, I applied your suggestions. Will merge as soon as the CI build succeeds.

PR #2260, opened 2020-04-06
HSEARCH-3886 Missing values should be last by default in field sorts

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3886


fax4ever:
good choice to choose the `MISSING_LAST` as default.
Merging now...

PR #2261, opened 2020-04-06
HSEARCH-3846 Filtering for sorts on multi-valued fields in nested documents

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3846

Extracted from #2255 by @wklaczynski , with some clean-up, fixes, tests and documentation.

Based on #2260, which should be merged first.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ atlassian-ide-plugin.xml
 .nbattrs
 nb-configuration.xml
 licenseheader.txt
+nbactions.xml
 
 # Sublime
 *.sublime-*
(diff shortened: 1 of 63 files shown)


fax4ever:
I this there is commit related to this request that is present only on the #2262:
> HSEARCH-3846 Test backend-specific extensions in sort filters.

I'll review it in #2262

yrodiere:
@fax4ever Good catch, it seems I forgot to push it to this branch. I just rebased on master and force-pushed. I'll merge when CI is happy.

wklaczynski:
The aggregation problem has not yet been resolved. Can you rebase aggregation to the current master?

PR #2262, opened 2020-04-07
HSEARCH-3885 Throw exceptions when attempting to use predicate with incompatible nested path hierarchy in a nested predicate

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3885

Based on #2261, which should be merged first.

This simply makes sure that something like this:

```
.where( f -> f.nested().objectField( "nestedObject" )
        .nested( f.match().field( "anotherNestedObject.someText" ).matching( ... ) )
```

... will throw an exception, because predicates inside a nested predicate should be scoped on the corresponding nested object, and not a parent or sibling.

Perhaps more importantly, this makes sure that filters in sorts will check that the given predicate actually targets the right nested document, e.g. this will throw an exception:

```
.sort( f -> f.field( "nestedObject.someText" )
        .filter( pf -> pf.match().field( "anotherNestedObject.someCategory" ).matching( ... ) )
```


yrodiere:
Merged, thanks!

PR #2264, opened 2020-04-12
Unexpected fields when use nested hsearch 3889

wklaczynski:
Resolve Exception Predicate targets unexpected fields when use nested() in sort filter


yrodiere:
As I explained in [HSEARCH-3889](https://hibernate.atlassian.net/browse/HSEARCH-3889), I'm concerned this will allow some illegal predicates, such as the one below, which will probably not work:

```
            SearchQuery<Assortment> query = session.search(scope)
                    .where(f -> f.nested().objectField("prices").nest(
                                        f.nested().objectField("prices").nest(
                                                             f.nested().objectField("prices").nest(
                                                                                  f.match().field("prices.thing").matching("foo")
                                                             )
                                        )
                    ))
                    .toQuery();
```

I provided a solution to your problem in the ticket. Closing this for now.

wklaczynski:
I know you can do that. But then you need to submit the filter for sorting and searching separately. The idea is to use the same filtering query and not worry about how to use it when sorting and searching differently. This is important for very complex filters, where it is necessary to use the nesting predicate once and in other cases it should not be used. PR should be able to adapt himself. Do not throw exceptions, require some complex actions.

```java
PermissionQuery allowQuery = new PermissionQuery("prices.permissions", operation, user, true);

            filter = createComplexFilter(...);

            sort.add(sorter.field("prices.bruttoPrice_sort")
                    .asc()
                    .mode(SortMode.MIN)
                    .filter(filter));

            SearchQuery<Assortment> query = session.search(scope)
                    .where((f) -> f.bool().must(select)
                    .filter(filter)
                    .sort(sort.toSort())
                    .toQuery();
.....                    
                    
public static SearchPredicate createComplexFilter(...) {

            PermissionQuery allowQuery = new PermissionQuery("prices.prmissions", "share", user, true);
            PermissionQuery dannyQuery = new PermissionQuery("prices.prmissions", "share", user, false);

            BooleanQuery query = new BooleanQuery.Builder()
                    .add(allowQuery, BooleanClause.Occur.MUST)
                    .add(dannyQuery, BooleanClause.Occur.MUST_NOT)
                    .build();

            SearchPredicate permfilter = predicate
                    .nested().objectField("prices")
                    .nest(predicate
                            .extension(LuceneExtension.get())
                            .fromLuceneQuery(query))
                    .toPredicate();

            SearchPredicate activefilter = ctx.predicate()
                    .match().field("prices.active")
                    .matching(true)
                    .toPredicate();

            SearchPredicate filter = predicate.bool()
                    .must(activefilter)
                    .must(permfilter)
                    .toPredicate();

            return filter;   
}            
```

In this case it would not be possible. My patch makes it possible. Just when building a query, it decides by context how to build a filter.

wklaczynski:
The user could define filters as you presented. Maybe it is better to allow more if it is forbidden. Now the code is getting too complicated. Dividing queries later, submitting them according to the use case is inefficient and complicates the code. You can't prepare filtering patterns, querying dynamic models. Building separate filters for sorting and searching, and in the future and aggregation will be too complicated.

PR #2265, opened 2020-04-14
HSEARCH-1927 + HSEARCH-1929 Aggregations on multi-valued numeric fields for the Lucene backend

yrodiere:
* [HSEARCH-1927](https://hibernate.atlassian.net/browse/HSEARCH-1927): Range faceting on multiple numeric values does not work
* [HSEARCH-1929](https://hibernate.atlassian.net/browse/HSEARCH-1929): @Facet on multi value field breaks indexing (be it for discrete or range faceting)

Based on code extracted from #2255 by @wklaczynski , with a bit of clean-up and a few fixes.

PR #2266, opened 2020-04-15
HSEARCH-3892 Aggregations on fields in nested documents

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3892

~This is based on #2265, which should be merged first.~ => Done


yrodiere:
Full build: https://ci.hibernate.org/job/hibernate-search/job/PR-2266/2/

yrodiere:
The tests pass. I'll go ahead and merge. The code is mostly taken from Lucene with a few changes, so I don't think there would be much value in reviewing this in depth.

PR #2267, opened 2020-04-15
HSEARCH-3893 + HSEARCH-3894 Bugfixes for aggregations on fields in nested documents

yrodiere:
* [HSEARCH-3893](https://hibernate.atlassian.net/browse/HSEARCH-3893): Elasticsearch aggregations return the count of nested (leaf) documents
* [HSEARCH-3894](https://hibernate.atlassian.net/browse/HSEARCH-3894): Lucene text aggregations count nested documents twice when they include the same term multiple times


fax4ever:
Changes make sense to me!
Merging thanks!

PR #2268, opened 2020-04-16
HSEARCH-3881 Filtering for aggregations on multi-valued fields in nested documents

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3881

Based on #2267, which should be merged first.

Initial commits extracted from #2255 by @wklaczynski. I cleaned up a few things, added an Elasticearch implementation, added tests, and added some documentation.


fax4ever:
great work +1 to merge it after that the conflict is solved

yrodiere:
Merged, thanks

PR #2271, opened 2020-04-19
Correcting jms.password property name to provide login credentials 

arslnmsd:


yrodiere:
Merged, thanks! Will backport this right now.

PR #2272, opened 2020-04-23
HSEARCH-3896 Verify multiple loading with the DefaultProjectionHitMapper on the same query instance

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3896

### DIFF
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/search/SearchQueryBaseIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/search/SearchQueryBaseIT.java
@@ -120,6 +120,39 @@ public class SearchQueryBaseIT {
 	}
 
 	@Test
+	@TestForIssue(jiraKey = "HSEARCH-3896")
+	public void target_byClass_singleType_reuseQueryInstance() {
+		OrmUtils.withinSession( sessionFactory, session -> {
+			SearchSession searchSession = Search.session( session );
+
+			SearchQuery<Book> query = searchSession.search( Book.class )
+					.where( f -> f.matchAll() )
+					.toQuery();
+
+			// reuse three times the same query instance
+			for ( int i = 0; i < 3; i++ ) {
+				backendMock.expectSearchObjects(
+						Arrays.asList( Book.NAME ),
+						b -> {
+						},
+						StubSearchWorkBehavior.of(
+								3L,
+								reference( Book.NAME, "1" ),
+								reference( Book.NAME, "2" ),
+								reference( Book.NAME, "3" )
+						)
+				);
+
+				Assertions.assertThat( query.fetchAllHits() ).containsExactly(
+						session.load( Book.class, 1 ),
+						session.load( Book.class, 2 ),
+						session.load( Book.class, 3 )
+				);
+			}
+		} );
+	}
+
+	@Test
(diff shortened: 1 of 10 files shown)


integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/search/SearchQueryBaseIT.java
```diff
@@ -130,7 +163,8 @@ public void target_byClass_multipleTypes() {
 
 			backendMock.expectSearchObjects(
 					Arrays.asList( Book.NAME, Author.NAME ),
-					b -> { },
+					b -> {
+					},
```
yrodiere, line 167: Please select the code you want to reformat before launching the automatic formatter next time, this will avoid this kind of noise :)

mapper/javabean/src/main/java/org/hibernate/search/mapper/javabean/search/loading/context/impl/JavaBeanLoadingContext.java
```diff
@@ -21,15 +21,15 @@
...
+	private final DocumentReferenceConverter<EntityReference> documentReferenceConverter;
 
-	private JavaBeanLoadingContext(ProjectionHitMapper<EntityReference, Void> projectionHitMapper) {
-		this.projectionHitMapper = projectionHitMapper;
+	private JavaBeanLoadingContext(DocumentReferenceConverter<EntityReference> documentReferenceConverter) {
+		this.documentReferenceConverter = documentReferenceConverter;
 	}
 
 	@Override
 	public ProjectionHitMapper<EntityReference, Void> getProjectionHitMapper() {
```
yrodiere, line 31: Could you please also rename this method to `createProjectionHitMapper`? That would clarify that a new one is returned every time the method is called.

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/rule/SearchWorkCall.java
```diff
@@ -92,9 +93,15 @@ protected boolean isSimilarTo(SearchWorkCall<?> other) {
...
 			actualProjectionContext.reset();
-			extractedElements.add(
-					actualRootProjection.extract( actualProjectionHitMapper, rawHit, actualProjectionContext )
-			);
+			Object extract = actualRootProjection.extract( actualProjectionHitMapper, rawHit, actualProjectionContext );
+			extractedElements.add( extract );
+
+			// Check for HSEARCH-3896
+			if ( extract instanceof Integer ) {
+				if ( ( (int) extract ) >= rawHits.size() ) {
```
yrodiere, line 101: I wouldn't add this check. Keys do not necessary match the index of hits.

For example you could have two requested loads for a single hit. That may be the reason this checks fails?

Currently that's a bit dumb, since we would load the same entity twice, but in the future one can imagine loading associated entities from their ID specified in a field, for example.

fax4ever: dropped this commit

fax4ever:
Thank you @yrodiere for the review. Anything else here?

yrodiere:
Merged, thanks!

PR #2273, opened 2020-04-24
HSEARCH-3898 Change the explain API so that the type name is used

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3897
https://hibernate.atlassian.net/browse/HSEARCH-3898

I created the follow up (not solved here): HSEARCH-3899.
To use the entity id instead of the document id for the same API.

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
@@ -7,6 +7,7 @@
 package org.hibernate.search.backend.elasticsearch.scope.model.impl;
 
 import java.lang.invoke.MethodHandles;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
@@ -33,30 +34,34 @@ public class ElasticsearchScopeModel {
 	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
 
 	private final Set<ElasticsearchIndexModel> indexModels;
-	private final Map<String, URLEncodedString> hibernateSearchIndexNamesToIndexReadNames;
-	private final Set<String> mappedTypeNames;
+	private final Set<String> hibernateSearchIndexNames;
+	private final Map<String, URLEncodedString> mappedTypeToElasticsearchIndexNames;
 
 	public ElasticsearchScopeModel(Set<ElasticsearchIndexModel> indexModels) {
 		this.indexModels = indexModels;
 		// Use LinkedHashMap/LinkedHashSet to ensure stable order when generating requests
-		this.hibernateSearchIndexNamesToIndexReadNames = new LinkedHashMap<>();
-		this.mappedTypeNames = new LinkedHashSet<>();
+		this.hibernateSearchIndexNames = new LinkedHashSet<>();
+		this.mappedTypeToElasticsearchIndexNames = new LinkedHashMap<>();
 		for ( ElasticsearchIndexModel model : indexModels ) {
-			hibernateSearchIndexNamesToIndexReadNames.put( model.getHibernateSearchIndexName(), model.getNames().getRead() );
-			mappedTypeNames.add( model.getMappedTypeName() );
+			hibernateSearchIndexNames.add( model.getHibernateSearchIndexName() );
+			mappedTypeToElasticsearchIndexNames.put( model.getMappedTypeName(), model.getNames().getRead() );
 		}
 	}
 
-	public Set<String> getHibernateSearchIndexNames() {
-		return hibernateSearchIndexNamesToIndexReadNames.keySet();
-	}
-
-	public Map<String, URLEncodedString> getHibernateSearchIndexNamesToIndexReadNames() {
-		return hibernateSearchIndexNamesToIndexReadNames;
-	}
-
 	public Set<String> getMappedTypeNames() {
-		return mappedTypeNames;
+		return mappedTypeToElasticsearchIndexNames.keySet();
+	}
+
+	public Set<String> getHibernateSearchIndexNames() {
+		return hibernateSearchIndexNames;
+	}
+
+	public Collection<URLEncodedString> getElasticsearchIndexNames() {
+		return mappedTypeToElasticsearchIndexNames.values();
+	}
+
+	public Map<String, URLEncodedString> getMappedTypeToElasticsearchIndexNames() {
+		return mappedTypeToElasticsearchIndexNames;
 	}
 
 	public EventContext getIndexesEventContext() {
--- a/integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/ElasticsearchExtensionIT.java
+++ b/integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/ElasticsearchExtensionIT.java
@@ -314,7 +314,7 @@ public class ElasticsearchExtensionIT {
 		)
 				.isInstanceOf( SearchException.class )
 				.hasMessageContaining(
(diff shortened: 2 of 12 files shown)


integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/LuceneExtensionIT.java
```diff
@@ -270,12 +276,12 @@ public void query_explain_multipleIndexes_invalidIndexName() {
...
 		Assertions.assertThatThrownBy(
-				() -> query.explain( "NotAnIndexName", FIRST_ID )
+				() -> query.explain( "NotAMappedTypeName", FIRST_ID )
 		)
 				.isInstanceOf( SearchException.class )
 				.hasMessageContaining(
-						"index name 'NotAnIndexName' is not among the indexes targeted by this query: ["
-								+ INDEX_NAME + ", " + OTHER_INDEX_NAME + "]"
+						"mapped type name 'NotAMappedTypeName' is not among the mapped types targeted by this query: [",
+						TYPE_NAME, OTHER_INDEX_NAME, "]"
```
yrodiere, line 284: `OTHER_INDEX_NAME` => `OTHER_TYPE_NAME`?

I'm surprised that the test passed, by the way...

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
```diff
@@ -34,16 +33,16 @@
 
 	private final Set<ElasticsearchIndexModel> indexModels;
 	private final Map<String, URLEncodedString> hibernateSearchIndexNamesToIndexReadNames;
```
yrodiere, line 35: I think that, after your change, this can be turned into a `Set<URLEncodedString` and renamed to `indexReadNames`.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/ElasticsearchExtensionIT.java
```diff
@@ -310,12 +316,12 @@ public void query_explain_multipleIndexes_invalidIndexName() {
...
 		Assertions.assertThatThrownBy(
-				() -> query.explain( "NotAnIndexName", FIRST_ID )
+				() -> query.explain( "NotAMappedName", FIRST_ID )
 		)
 				.isInstanceOf( SearchException.class )
 				.hasMessageContaining(
-						"type name 'NotAnIndexName' is not among the mapped type targeted by this query: ["
-						+ INDEX_NAME + ", " + OTHER_INDEX_NAME + "]"
+						"type name 'NotAMappedName' is not among the mapped type targeted by this query: [",
+						INDEX_NAME, OTHER_INDEX_NAME
```
yrodiere, line 324: Shouldn't this be `TYPE_NAME`, `OTHER_TYPE_NAME`?

Also, why did you have to remove the comma and square bracket? Something looks wrong here.

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/LuceneExtensionIT.java
fax4ever: it was my fault, I always thought that `hasMessageContaining` takes more message terms as a parameter, but it isn't so.
To check more terms we have to invoke `hasMessageContaining` more times.
Fixing it...

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/ElasticsearchExtensionIT.java
fax4ever: Yes they should. Same issue as above :)

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/scope/model/impl/ElasticsearchScopeModel.java
fax4ever: Right! We don't need a map for that anymore.

fax4ever:
Thank you for the review @yrodiere. I think I've addressed the comments here.

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/LuceneExtensionIT.java
```diff
@@ -255,10 +261,9 @@ public void query_explain_multipleIndexes_missingIndexName() {
 				() -> query.explain( FIRST_ID )
 		)
 				.isInstanceOf( SearchException.class )
-				.hasMessageContaining( "explain(String id) cannot be used when the query targets multiple indexes" )
-				.hasMessageContaining(
-						"pass one of [" + INDEX_NAME + ", " + OTHER_INDEX_NAME + "]"
-				);
+				.hasMessageContaining( "explain(String id) cannot be used when the query targets multiple mapped types" )
+				.hasMessageContaining( TYPE_NAME )
+				.hasMessageContaining( OTHER_TYPE_NAME );
```
yrodiere, line 266: I still don't get why you're no longer checking whether the string contains `"pass one of [" + INDEX_NAME + ", " + OTHER_INDEX_NAME + "]"`?

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/LuceneExtensionIT.java
```diff
@@ -270,13 +275,12 @@ public void query_explain_multipleIndexes_invalidIndexName() {
...
+				() -> query.explain( "NotAMappedTypeName", FIRST_ID )
 		)
 				.isInstanceOf( SearchException.class )
-				.hasMessageContaining(
-						"index name 'NotAnIndexName' is not among the indexes targeted by this query: ["
-								+ INDEX_NAME + ", " + OTHER_INDEX_NAME + "]"
-				);
+				.hasMessageContaining( "mapped type name 'NotAMappedTypeName' is not among the mapped types targeted by this query" )
+				.hasMessageContaining( TYPE_NAME )
+				.hasMessageContaining( OTHER_TYPE_NAME );
```
yrodiere, line 283: Same here

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/LuceneExtensionIT.java
fax4ever: Differently from Elasticsearch, the class `LuceneScopeModel` stores `typeNames` and `indexNames` as `HashSet`s and not as `LinkedHashSet`. So I don't think the order with which they appear is predictable. May I have to change their implementations?

yrodiere: > May I have to change their implementations?

Yes, definitely. You should change the implementation to use `LinkedHashSet` and behave predictably, not change the tests to relax the constraints :)

fax4ever: I will do

fax4ever: changed

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/LuceneExtensionIT.java
fax4ever: changed

yrodiere:
Merged, thanks!

PR #2274, opened 2020-04-30
HSEARCH-3273 Dynamic fields

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3273

@fax4ever So once again there are many changes. Mostly because of tests, but still. Could you please just have a look at the documentation and API, see if they are clear?


yrodiere:
Full build: https://ci.hibernate.org/job/hibernate-search/job/PR-2274/2/

documentation/src/main/asciidoc/mapper-orm-bridge-index-field-dsl.asciidoc
```diff
@@ -165,3 +165,105 @@ and pass it to the bridge for later use.
...
+their path and type are known on bootstrap.
+
+In some very specific cases, the path of a field is not known until you actually index it;
+for example, you may want to index a `Map<String, Integer>` by using the map keys as field names.
+The field, then, is considered _dynamic_.
+
+Dynamic fields are not declared on bootstrap,
+but need to match a field _template_ that is declared on bootstrap.
+The template includes the field type and structural information (multi-valued or not, ...),
+but omits the field name.
```
fax4ever, line 182: I would use the term names instead of name here, to make clear that for each template there could be more fields.

documentation/src/main/asciidoc/mapper-orm-bridge-index-field-dsl.asciidoc
```diff
@@ -165,3 +165,105 @@ and pass it to the bridge for later use.
...
+
+Field templates are declared with the `fieldTemplate` methods, as shown below.
```
fax4ever, line 184: Reading the chapter I think that it is not clear which interface the following `bind` method belongs to.
Maybe we should write something like:
> The template is activated on a give property using a property binding annotation.

Moreover, ideally I think that we should use a different annotation to bind dynamic fields.

documentation/src/test/java/org/hibernate/search/documentation/mapper/orm/bridge/document/model/dsl/dynamic/UserMetadataBinder.java
```diff
@@ -0,0 +1,68 @@
...
+import org.hibernate.search.mapper.pojo.bridge.PropertyBridge;
+import org.hibernate.search.mapper.pojo.bridge.binding.PropertyBindingContext;
+import org.hibernate.search.mapper.pojo.bridge.mapping.programmatic.PropertyBinder;
+import org.hibernate.search.mapper.pojo.bridge.runtime.PropertyBridgeWriteContext;
+
+public class UserMetadataBinder implements PropertyBinder {
+
+	//tag::bind[]
+	@Override
+	public void bind(PropertyBindingContext context) {
```
fax4ever, line 24: Maybe we should include the class statement in the code sample.

documentation/src/test/java/org/hibernate/search/documentation/mapper/orm/bridge/document/model/dsl/dynamic/UserMetadataBinder.java
```diff
@@ -0,0 +1,68 @@
...
+		private Bridge(IndexObjectFieldReference userMetadataFieldReference) {
+			this.userMetadataFieldReference = userMetadataFieldReference;
+		}
+
+		//tag::write[]
+		@Override
+		public void write(DocumentElement target, Object bridgedElement, PropertyBridgeWriteContext context) {
+			Map<String, String> userMetadata = (Map<String, String>) bridgedElement;
+
+			DocumentElement indexedUserMetadata = target.addObject( userMetadataFieldReference ); // <1>
```
fax4ever, line 58: I think that is not clear where the field `userMetadataFieldReference` comes from. Maybe we should include at the least the constructor in the code sample, even if I would add the class statement as well.

documentation/src/test/java/org/hibernate/search/documentation/mapper/orm/bridge/document/model/dsl/dynamic/Book.java
```diff
@@ -0,0 +1,53 @@
...
+@Indexed
+public class Book {
+
+	@Id
+	private Integer id;
+
+	@ElementCollection
+	@SortNatural
+	@PropertyBinding(binder = @PropertyBinderRef(type = UserMetadataBinder.class))
+	private Map<String, String> userMetadata = new TreeMap<>();
```
fax4ever, line 31: I was thinking about possible use cases.
I don't know if it is very useful to generate more fields having the exact same mapping and modeling.

documentation/src/main/asciidoc/mapper-orm-bridge-index-field-dsl.asciidoc
```diff
@@ -165,3 +165,105 @@ and pass it to the bridge for later use.
...
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/mapper/orm/bridge/document/model/dsl/dynamic/MultiTypeUserMetadataBinder.java[tags=bind]
+----
+<1> Declare an object field with `objectField`.
+<2> Declare a field template for integer with `fieldTemplate`.
+<3> Pass the *template* name.
+<4> Define the field type as integer, sortable.
+<5> Assign a path pattern to the template, so that only fields ending with `_int` will be considered as integers.
```
fax4ever, line 242: I like the idea of having templates. I don't know but maybe we should find a way to make the association to the field more dynamic, providing the template within the field name, e.g. having in the map key something like `fieldName:templateName`.

Moreover, I would let the user to define templates globally, for the entire application, not only for a given property.

documentation/src/main/asciidoc/mapper-orm-bridge-index-field-dsl.asciidoc
yrodiere: > Reading the chapter I think that it is not clear which interface the following bind method belongs to.

Right. I added something to clarify that.

> Moreover, ideally I think that we should use a different annotation to bind dynamic fields.

In order to do that, we would need one annotation per type: `@FullTextFieldTemplate`, `@KeywordFieldTemplate`, ... Which is a bit annoying.

But more importantly, dynamic fields are... well, dynamic. The annotation-based mapping is intrinsically static: all the fields are declared from the start. Typically, they are not useful for POJOs that are mapped "directly" with annotations.

They become useful when you need to map, say, a `Map<String, ?>`, or a `JsonObject`. Then you can't declare the fields upfront, but you may be able to define templates. However, when you do that, it's likely that you will need some custom code. Maybe the maps contain other maps as values and you need some recursion; maybe you don't want to index *everything* in the map.

My point is, you will need custom code (a bridge) at some point, so users who need templates already have a class ready to declare them. With that in mind, annotations feel a bit pointless, in my opinion.

I suppose we could introduce annotations for very specific use cases in the future. `@DynamicJsonFields` would be an idea. But in any case, that doesn't make the current API obsolete, as it will still be needed when we implement support for `@DynamicJsonFields` (for example).

documentation/src/test/java/org/hibernate/search/documentation/mapper/orm/bridge/document/model/dsl/dynamic/UserMetadataBinder.java
yrodiere: I was trying to avoid that, as I wanted this section to focus on the DSL and not on the bridge... But I can see how that's confusing. I'll see what I can do.

documentation/src/main/asciidoc/mapper-orm-bridge-index-field-dsl.asciidoc
yrodiere: > I like the idea of having templates. I don't know but maybe we should find a way to make the association to the field more dynamic, providing the template within the field name, e.g. having in the map key something like `fieldName:templateName`.

Passing a reference to the template would indeed be nice, and that's I wanted to do initially, but that's simply not possible with Elasticsearch. You can't specify the type at indexing time, you have to rely on a template with a name pattern.

Eventually we might consider introducing an extension for Lucene only, I suppose. But I'd rather wait until we get a real use case.

> Moreover, I would let the user to define templates globally, for the entire application, not only for a given property.

I don't think it would be a good idea to do that for the entire application. Templates are defined per-index in Elasticsearch, and I'd rather stick to that so that changing templates will only affect the indexes that actually need this template.

We should keep in mind that dynamic fields are not generally necessary. They are not necessary when you directly map properties of your entities, because then you can simply use `@*Field` annotations and create "static" fields. They are mostly necessary for advanced use cases such as mapping a `Map` or a JSON object whose schema is not known in advance.

If you need to define templates for the whole index, that's totally possible, as explained below:

> Templates declared on a given schema element can be matched in children of that element.
> For example, if you declare templates at the root of your entity (through a <<mapper-orm-bridge-typebridge,type bridge>>),
> these templates will be implicitly available in every single property bridge of that entity.
> In such cases, templates declared in property bridges will take precedence over those declared in the type bridge.

documentation/src/test/java/org/hibernate/search/documentation/mapper/orm/bridge/document/model/dsl/dynamic/Book.java
yrodiere: I'm not sure what you mean by this?

> having the exact same mapping and modeling

yrodiere: Ah, you mean using a single template. Well yes it's very limited, but can be useful if all you want to index is text.

A more advanced use case is presented below.

yrodiere:
Thanks! I applied your suggestions and rebased. Waiting for CI, then I'll merge.

PR #2278, opened 2020-05-05
HSEARCH-3911 + HSEARCH-3914 Expose TopDocs and Sort in Lucene queries

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3911
https://hibernate.atlassian.net/browse/HSEARCH-3914


fax4ever:
Let's put this pull request on ice since we don't sure we really need the document id to rebuild score and top docs. 

We have to test if we can use to the index in the query hits instead, when we pass the reconstructed top docs to `org.apache.lucene.search.TopDocs#merge(int, TopDocs[])` Lucene low level API.

fax4ever:
@yrodiere I only added a test for documentation.

yrodiere:
@fax4ever I pushed the commits addressing HSEARCH-3914 and also some tests that reproduce what Infinispan currently does. Waiting for CI, then I'll mege.

fax4ever:
that's great! thank you

yrodiere:
Merged, thanks

PR #2279, opened 2020-05-06
HSEARCH-3901 Expose metadata about the annotated type in the context passed to PropertyMappingAnnotationProcessor/TypeMappingAnnotationProcessor

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3901

Turns out we cannot use markers, or the existing `PojoModelElement` which relies on markers, because markers are derived from mapping annotations; essentially it's a problem of circular dependency.

Anyway, I managed to expose the annotations directly, which should be good enough.


fax4ever:
I think there is all we need to support the HS5 annotations.
Merged, thanks!

PR #2280, opened 2020-05-07
HSEARCH-3584 Support ORM bytecode enhancement when embeddables, mapped superclasses or entity hierarchies are involved

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3584


yrodiere:
Merged, thanks

PR #2283, opened 2020-05-12
HSEARCH-3097 Deprecate the IndexedEmbedded "prefix" feature

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3097


yrodiere:
Merged, thanks

PR #2284, opened 2020-05-13
HSEARCH-3902 Expose addOrUpdate/delete/purge on IndexIndexer/PojoIndexer

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3902

@fax4ever With this PR, I believe you'll have everything you need to get rid of PojoIndexingPlan in Infinispan and rely on a singleton PojoIndexer instead.


yrodiere:
Merged, thanks

PR #2288, opened 2020-05-14
HSEARCH-3842 Expose a global option to use ReindexOnUpdate.NO as a default

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3842

@fax4ever Thanks to this, you shouldn't have to use `ReindexOnUpdate` in Infinispan at all, except in one specific place to set the default to `NO`.


fax4ever:
Actually restoring all the ISPN query tests I found a case for it.
This test has a bidirectional relationship between the root entity and the owned embedded and I had to add manually:

``` java
@IndexingDependency(reindexOnUpdate = ReindexOnUpdate.NO)
```

There are few cases, but it is of course better to make it the default.

yrodiere:
I'll take that as a "yes". Merged, thanks!

PR #2291, opened 2020-05-20
HSEARCH-962 HibernateOrmCriteriaEntityLoader needs to execute multiple queries in cases where the underlying database has a limit on the total amount of query parameters

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-962

This simply changes the way `fetchSize` is handled for criteria queries: we now take care of not requesting more than `fetchSize` elements for each query execution, execute the query multiple times if necessary.
That's about the same behavior as in `HibernateOrmByIdEntityLoader` and should solve the problem in most cases as the default fetch size is `100`.


fax4ever:
Merged thanks!

PR #2293, opened 2020-05-25
HSEARCH-3628 Finer loading options in search queries (fetch graph, ...)

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3628

### DIFF
--- a/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/loading/impl/MutableEntityLoadingOptions.java
+++ b/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/loading/impl/MutableEntityLoadingOptions.java
@@ -6,10 +6,16 @@
  */
 package org.hibernate.search.mapper.orm.search.loading.impl;
 
+import java.util.ArrayList;
+import java.util.List;
+
+import org.hibernate.persister.entity.EntityPersister;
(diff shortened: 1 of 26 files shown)


mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/loading/impl/MutableEntityLoadingOptions.java
```diff
@@ -6,11 +6,17 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import org.hibernate.persister.entity.EntityPersister;
 import org.hibernate.search.util.common.impl.Contracts;
 
 public class MutableEntityLoadingOptions {
 	private int fetchSize;
 
+	private List<EntityGraphHint> entityGraphHints;
```
fax4ever, line 18: This is just my curiosity, is it possible to define multiple graphs on the same query instance?

yrodiere: No, but theoretically you could target multiple, unrelated entity types with the same query, in which case you could define multiple graph, one for each entity type.

yrodiere:
Merged, thanks

PR #2296, opened 2020-05-28
HSEARCH-1231 Inherit @Indexed annotations by default

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-1231


yrodiere:
Merged, thanks!

PR #2297, opened 2020-05-29
HSEARCH-3933 Fix reset expectations and verify expectations met for the stub backend

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3933


fax4ever:
Merged thanks

PR #2298, opened 2020-06-04
HSEARCH-3936 Projectable fields in the Elasticsearch schema should have "store = false"

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3936


fax4ever:
Merged thanks!

PR #2299, opened 2020-06-05
HSEARCH-3937 + HSEARCH-3938 Fix bootstrap when using an EnvironmentSynchronizer

yrodiere:
* [HSEARCH-3937](https://hibernate.atlassian.net/browse/HSEARCH-3937): Error messages on boot/shutdown are hidden or unclear when using an EnvironmentSynchronizer
* [HSEARCH-3938](https://hibernate.atlassian.net/browse/HSEARCH-3938): HibernateOrmIntegrationBooterImpl confuses the startup event for a shutdown event, and vice-versa


yrodiere:
Done and merged. Thanks!

PR #2305, opened 2020-06-11
HSEARCH-3323 Search 6 groundwork - Restore support for scrolling

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3323

~~@yrodiere here is the draft for the feature, fell free to complete / fix it if you want provide this feature before July.~~
~~if the CI passes, this is ready for review~~

draft again


integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/search/loading/SearchQueryEntityLoadingScrollingIT.java
```diff
@@ -0,0 +1,46 @@
...
+						hitDocumentReferences.size(),
+						hitDocumentReferences
+				)
+		);
+
+		backendMock.expectCloseScroll( targetIndexes );
+
+		try ( SearchScroll<T> scroll = query.scroll( hitDocumentReferences.size() ) ) {
+			SearchScrollResult<T> next = scroll.next();
+			return next.getHits();
```
yrodiere, line 43: While this may be useful, I don't think it's enough.

I think we need another test to check that chunk processing works correctly, something like this:

```
	try (Session session = ...) {
		SearchQuery<T> query = ...;
		try ( SearchScroll<T> scroll = query.scroll( hitDocumentReferences.size() ) ) {
			for ( SearchScrollResult<T> next = scroll.next(); next.hasHits(); next = scroll.next() ) {
				// Do something with the hits...
				// (in tests, maybe just update some property)
				session.flush();
				session.clear();
			}
		}
	}
	
	// (in tests, check that changes above had an effect)
```

Maybe you can adapt this test class, maybe you need a separate test class. In any case, we need a test.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryImpl.java
```diff
@@ -149,7 +146,21 @@ public long fetchTotalHitCount() {
 
 	@Override
 	public SearchScroll<H> scroll(Integer pageSize) {
-		throw new UnsupportedOperationException( "Not yet implemented" );
+		NonBulkableWork<ElasticsearchLoadableSearchResult<H>> work = getSearchWorkBuilder()
+				.scrolling( pageSize, "60s" )
```
yrodiere: This hardcoded value here needs a TODO :)

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryImpl.java
```diff
@@ -149,7 +146,21 @@ public long fetchTotalHitCount() {
...
+				.build();
+
+		final ElasticsearchSearchResult<H> searchResult = Futures.unwrappedExceptionJoin( queryOrchestrator.submit( work ) )
+				.loadBlocking();
```
yrodiere, line 155: I think this should be done on the first call to `next()`, not here. Otherwise you're pre-fetching entities, which may be unexpected to some users: what if they do a `session.clear()` before the call to `scroll()` and the call to `next()`?

I understand this will complicate the implementation of `ElasticsearchSearchScroll`, but I really think it's the only way to comply with the principle of least surprise.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchScroll.java
```diff
@@ -0,0 +1,57 @@
...
+	@Override
+	public void close() {
+		Futures.unwrappedExceptionJoin( queryOrchestrator.submit( workFactory.clearScroll( scrollId ).build() ) );
+	}
+
+	@Override
+	public SearchScrollResult<H> next() {
+		if ( firstNext ) {
+			firstNext = false;
+			return new SimpleSearchScrollResult<>( !firstNextHits.isEmpty(), firstNextHits );
```
yrodiere, line 53: You can't use `!firstNextHits.isEmpty()` to determine whether there were hits or not.

The point of `hasHits()` in `ScrollResult` is to serve as a loop condition: if there were hits on this page, then there may be more pages and I'll iterate one more time.

Imagine that a page of results contains *only* entities that were deleted from the database. Then the hits will be empty. Now imagine there are actually more hits after that, which were *not* delete from the database.
With `!firstNextHits.isEmpty()`, `hasHits()` will return `false` on the first page of results, the client loop will terminate, and it will never even request the second page.

What you really need here is to check that the hits *before loading* are non-empty. I think adding a getter to `ElasticsearchLoadableSearchResult` that returns `extractedHits.size()` would do the trick.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/Elasticsearch7SearchResultExtractor.java
```diff
@@ -73,6 +76,8 @@
 		Map<AggregationKey<?>, ?> extractedAggregations = aggregations.isEmpty() ?
 				Collections.emptyMap() : extractAggregations( extractContext, responseBody );
 
+		String scrollId = extractScrollId( responseBody );
```
yrodiere, line 79: Either solution looks fine to me. No opinion :)

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/cfg/ElasticsearchBackendSettings.java
```diff
@@ -239,6 +239,17 @@ private ElasticsearchBackendSettings() {
 	 */
 	public static final String THREAD_POOL_SIZE = "thread_pool.size";
 
+	/**
+	 * Property for specifying the maximum duration {@code ScrollableResults} will be usable if no
```
yrodiere, line 243:
```suggestion
	 * Property for specifying the maximum duration a {@code Scroll} will be usable if no
```

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchBackendFactory.java
```diff
@@ -154,6 +160,8 @@ public BackendImplementor create(String name, BackendBuildContext buildContext,
 
 			indexLayoutStrategyHolder = createIndexLayoutStrategy( buildContext, propertySource );
 
+			Integer scrollTimeout = SCROLL_TIMEOUT.get( propertySource );
```
yrodiere, line 163: By checking out this property when the backend is created, you're effectively making it a build-time property in Quarkus. I think it would make sense for users to change this property at deployment time, so I'd recommend checking it out when the backend starts, instead.

Currently all such "deployment-time" properties are related to the client (hosts, connection timeout, etc.). I think the easiest way to introduce the scroll timeout as a "deployment-time" property would be to move it to `ElasticsearchLinkImpl`. You can initialize it in `onStart` and add a getter similar to the other ones we already have there.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/logging/impl/Log.java
```diff
@@ -703,4 +703,12 @@ SearchException invalidFieldValueType(@FormatWith(ClassFormatter.class) Class<?>
 					+ " Make sure to call '.multi()' when you create the projection.")
 	SearchException invalidSingleValuedProjectionOnMultiValuedField(String absolutePath, @Param EventContext context);
 
+	@Message(id = ID_OFFSET_2 + 132,
+			value = "An IOException happened while scrolling data for query '%1$s'.")
+	SearchException ioExceptionOnScrollNextExecution(Query luceneQuery, @Param EventContext context, @Cause IOException e);
+
+	@Message(id = ID_OFFSET_2 + 133,
+			value = "An IOException happened while closing scrolling for query '%1$s'.")
```
yrodiere, line 711:
```suggestion
			value = "An IOException happened while closing scroll for query '%1$s'.")
```

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/logging/impl/Log.java
```diff
@@ -703,4 +703,12 @@ SearchException invalidFieldValueType(@FormatWith(ClassFormatter.class) Class<?>
...
+	SearchException ioExceptionOnCloseScrollExecution(String query, @Cause IOException e);
```
yrodiere, line 712: You could just use `unableToCloseIndexReader(...)`.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneLoadableSearchResult.java
```diff
@@ -49,6 +51,8 @@
 		this.rootProjection = rootProjection;
 		this.hitCount = hitCount;
 		this.extractedData = extractedData;
+		// compute this value before the loading
+		this.hasHits = !extractedData.isEmpty();
```
yrodiere, line 55: See, that's what I was talking about :)

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchQueryImpl.java
```diff
@@ -148,6 +153,26 @@ public Sort luceneSort() {
...
 
+	SearchScrollResult<H> next(HibernateSearchMultiReader indexReader, Integer offset, Integer limit) {
+		timeoutManager.start();
+
+		ReadWork<LuceneLoadableSearchResult<H>> work = workFactory.search( searcher, offset, limit );
+		LuceneLoadableSearchResult<H> search = doSubmitWithIndexReader( work, indexReader );
+		LuceneSearchResult<H> result = search
+				/*
+				 * WARNING: the following call must run in the user thread.
+				 * If we introduce async processing, we will have to add a loadAsync method here,
```
yrodiere, line 164:
```suggestion
				 * If we introduce async processing, we will have to add a nextAsync method here,
```

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchQueryImpl.java
```diff
@@ -148,6 +153,26 @@ public Sort luceneSort() {
 		return luceneSort;
 	}
 
+	SearchScrollResult<H> next(HibernateSearchMultiReader indexReader, Integer offset, Integer limit) {
```
yrodiere, line 156: I'd move this method to scroll. If feels weird to have this back-reference to the query.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchQueryImpl.java
```diff
@@ -148,6 +153,26 @@ public Sort luceneSort() {
...
+		timeoutManager.start();
+
+		ReadWork<LuceneLoadableSearchResult<H>> work = workFactory.search( searcher, offset, limit );
```
yrodiere, line 159: As far as I can tell, you didn't re-implement the exponentially-increasing behavior of Search 5. Is there a specific reason? I'm worried this will result in bad performance. For 100 pages of 100 elements, you will execute the query 100 times, while the Search 5 solution would just execute it 5 times...

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryImpl.java
fax4ever: Agreed... even if it will complicate a bit the alignment between the scroll instances client side (ours) and the server side (ES), it will be better to involve the server side scroll at the first `next()` time

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchScroll.java
fax4ever: right, it makes sense

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchQueryImpl.java
fax4ever: I simply forgot to do that. I'll do it in a different commit.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryImpl.java
fax4ever: right! I didn't put the TODO, anyway the row is overwritten by a further commit.

engine/src/main/java/org/hibernate/search/engine/search/query/SearchScroll.java
```diff
@@ -0,0 +1,32 @@
...
+
+	/**
+	 * Returns the next page, with at most {@code pageSize} hits.
+	 * <p>
+	 * May return a result with less than {@code pageSize} elements if only that many hits are left.
+	 *
+	 * @return The next {@link SearchScrollResult}.
+	 * @see SearchFetchable#scroll(Integer)
+	 */
+	SearchScrollResult<H> next();
```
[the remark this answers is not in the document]
yrodiere: Turned out this couldn't be done with Elasticsearch.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchScroll.java
```diff
@@ -72,9 +78,19 @@ public void close() {
...
+		SearchScrollResult<H> cached = prefetchResults.poll();
+		if ( cached != null ) {
+			return cached;
+		}
+
 		timeoutManager.start();
+		int fetchSize = pageSize * fetchFactor;
 
-		ReadWork<LuceneLoadableSearchResult<H>> work = workFactory.search( searcher, scrollIndex, pageSize );
+		ReadWork<LuceneLoadableSearchResult<H>> work = workFactory.search( searcher, scrollIndex, fetchSize );
```
yrodiere, line 93: As we already discussed, hits of one particular page should only be loaded when `next()` is called, so that users can safely use `session.flush()`/`session.clear()` and perform chunk processing while keeping memory usage low.

Here you're loading hits in advance each time you execute the query, including hits of the next page.
(Also note you're re-loading hits of the previous pages, which is completely pointless and will hurt performance). 

Note **this should have been detected by your tests**. Maybe have a **look at your tests first**, to determine why this implementation passed the tests even though it loads all entities in advance?

Then, to fix this... Your algorithm doesn't need to produce pages immediately. When it executes the query, it should cache only the `LuceneLoadableSearchResult`.

Then, each time someone calls `next()`, you need to:

* Determine the first and last index of the page to load.
* Check whether the loadable results include all of them:
  * If so, `LuceneLoadableSearchResult.loadBlocking( firstIndexIncluded, endIndexExcluded )` to load the hits, and return them.
  * If not, double the fetch size and execute the query again, then call `LuceneLoadableSearchResult.loadBlocking( firstIndexIncluded, endIndexExcluded )` to load the hits, and return them.
  * Note there's an edge case when the end of the scroll was reached. You obviously don't want to re-execute the query if you know you won't get any more hits.

This is basically the algorithm from Search 5. The name of the classes involved and the code structure changed, but the principles remain the same.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/search/loading/SearchQueryEntityChangingScrollingIT.java
```diff
@@ -0,0 +1,114 @@
...
+			backendMock.expectNextScroll( targetIndexes, StubSearchWorkBehavior.of( 0, documentReferences() ) );
+			backendMock.expectCloseScroll( targetIndexes );
+
+			int index = 0;
+
+			try ( SearchScroll<SimpleEntity> scroll = query.scroll( 3 ) ) {
+				for ( SearchScrollResult<SimpleEntity> next = scroll.next(); next.hasHits(); next = scroll.next() ) {
+					assertThat( next.hits() ).hasHitsAnyOrder( new SimpleEntity( index++ ), new SimpleEntity( index++ ), new SimpleEntity( index++ ) );
+					changeNames( next.hits() );
+				}
```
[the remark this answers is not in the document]
fax4ever: I had written these lines of code, but I removed them by mistake, sorry :)

fax4ever:
Hi @yrodiere I hadn't finished with the changes actually :P I'll let you know when I done.

yrodiere:
Hum ok, I got a notification with the description being edited to tell me it's ready for review... Something must have gone wrong. Sorry.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchScroll.java
fax4ever: About the fact that

> this should have been detected by your tests

I’m not very sure about that.
I think that you are probably talking about backend tests, since the exponential query execution is something that is done by the Lucene backend under the hood.

If so, we have 2 tests for backend targeting the scroll behavior:
`SearchQueryScrollIT`
`SearchQueryResultLoadingOrTrasformingIT`

About the former, I think that it doesn’t inspect how a backend fetches/queries the data. From my understanding, it looks only at the final scroll-result instances.
About the latter, I think that there are no expectations scoped on a single `next()`, they are scoped on the whole loading process of a given scroll.
All in all, i’m not sure we should update our tests.

yrodiere: > All in all, i’m not sure we should update our tests.

Well you just gave two problems you noticed in these tests you just created...

We expect `next` to have a specific behavior, to load only the entities that are returned. Yet we don't check that it only loads these entities.

So I'd say tests are incomplete and should be fixed? Or at least there should be another test method to test this? Something where you set easy mock's expectations just before each call to `next()`.

fax4ever: Right, we can add a test to `SearchQueryResultLoadingOrTrasformingIT` to be sure that entities are loaded at each fetch and not pre-fetched/cached by the backend, scoping the expectations at `next` level. But I would not add a specific test for the `exponential query execution`. Does is sound good?

yrodiere: Right, sounds good.

fax4ever: I'll do that, thanks for the explanation

fax4ever:
@yrodiere I've updated the branch adding a test the loading behavior on scrolling... it fails now for the Lucene backend for the reasons you expressed above... Next days I'll finish it, doing all the staff we agreed to do.

fax4ever:
Thank you @yrodiere for completing the job.
+1 to merge from me

yrodiere:
Merged, thanks!

PR #2320, opened 2020-07-08
HSEARCH-3963 make DefaultElasticsearchAnalyzerDefinitionTranslator extensible

Rokko11:
Refactoring for adding custom analyzer definitions


yrodiere:
The code you're changing (and projecting to override in your application) is not part of the API. It's not intended to be used in applications, and can be changed in incompatible ways (or even removed) at any time. See the [compatibility policy](https://hibernate.org/community/compatibility-policy/).

If you have specific filters in mind, maybe you should rather modify the code to handle those filters? If they require a specific dependency, you can change `LuceneAnalyzerImplementationTranslationMapBuilder` to use class names (Strings) instead of classes as the keys in the translation map, and add a methid `add(String luceneClassName, String typeName)` to handle them.
Then you can add things like this, and your filters will be translated correctly:

```
 				.add( "org.apache.lucene.analysis.phonetic.PhoneticFilterFactory", "phonetic" ) 
```

yrodiere:
Closing as explained in [HSEARCH-3963](https://hibernate.atlassian.net/browse/HSEARCH-3963): there are simpler solutions.

PR #2326, opened 2020-07-22
HSEARCH-3636 + HSEARCH-3967 Restore the ability to configure Lucene directories on a per-index or per-shard basis

yrodiere:
* [HSEARCH-3636](https://hibernate.atlassian.net/browse/HSEARCH-3636): Restore the ability to configure Lucene directories on a per-index or per-shard basis
* [HSEARCH-3967](https://hibernate.atlassian.net/browse/HSEARCH-3967): Specify defaults for index-level configuration properties at the backend level

Nothing much to review here, we've already validated the approach on Zulip.

Will wait for the Sonar report and merge.


yrodiere:
The "bug" is a false positive. Merging.

PR #2328, opened 2020-07-22
HSEARCH-3899 Change the explain API so that the entity id is used, instead of document id

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3899

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/ElasticsearchSearchQuery.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/ElasticsearchSearchQuery.java
@@ -16,27 +16,30 @@ public interface ElasticsearchSearchQuery<H>
 	/**
 	 * Explain score computation of this query for the document with the given id.
 	 * <p>
-	 * This is a shorthand for {@link #explain(String, String)} when the query only targets one index.
+	 * This is a shorthand for {@link #explain(String, Object)} when the query only targets one index.
 	 *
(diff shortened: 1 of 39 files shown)


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/ElasticsearchSearchQuery.java
```diff
@@ -16,27 +16,31 @@
...
+	 * @param id The id of the entity whose score should be explained.
+	 * This is the entity ID, which may be of any type ({@code long}, ...),
+	 * not the document ID which is always a string.
 	 * @return A {@link JsonObject} describing the score computation for the hit.
 	 * @throws org.hibernate.search.util.common.SearchException If the given mapped type name does not refer to a mapped name targeted by this query,
 	 * or if the explain request fails.
 	 */
 	// TODO HSEARCH-3899 avoiding merging index concepts (e.g.: documentId) and entity concepts (e.g.: mapped type)
-	JsonObject explain(String typeName, String id);
+	JsonObject explain(String typeName, Object id);
```
fax4ever, line 44: I think that thanks to this issue we can safely remove the todo here

fax4ever: as you already did for the similar Lucene interface

yrodiere:
Fixed, thanks. Will merge as soon as CI passes.

PR #2330, opened 2020-07-24
HSEARCH-3311 Test (and document) alternatives to analyzer discriminators in Search 6

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3311

I also included an experimental implementation, annotated with `@Incubating`. I think it's worth it, be it only to reduce the amount of code users need to write.

I believe the solution will evolve in the future, in particular if we address [HSEARCH-3903](https://hibernate.atlassian.net/browse/HSEARCH-3903). But for the time being... it works.


fax4ever:
Merged, thank you!

PR #2333, opened 2020-08-04
HSEARCH-3787 Support hard timeout on Hibernate ORM entity loading

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3787
~~https://hibernate.atlassian.net/browse/HSEARCH-3774~~ removed from this PR's scope

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/Elasticsearch7SearchResultExtractor.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/Elasticsearch7SearchResultExtractor.java
@@ -81,8 +81,8 @@ class Elasticsearch7SearchResultExtractor<H> implements ElasticsearchSearchResul
 		Integer took = TOOK_ACCESSOR.get( responseBody ).get();
 		Boolean timedOut = TIMED_OUT_ACCESSOR.get( responseBody ).get();
 
-		Integer remainingTimeToHardTimeout = ( hardTimeoutInMilliseconds == null ) ? null :
-				Math.toIntExact( hardTimeoutInMilliseconds ) - took;
+		Long remainingTimeToHardTimeout = ( hardTimeoutInMilliseconds == null ) ? null :
+				hardTimeoutInMilliseconds - took;
 
 		return new ElasticsearchLoadableSearchResult<>(
 				extractContext,
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/Elasticsearch7SearchResultExtractor.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/Elasticsearch7SearchResultExtractor.java
@@ -90,7 +90,7 @@ class Elasticsearch7SearchResultExtractor<H> implements ElasticsearchSearchResul
 				extractedHits,
 				extractedAggregations,
 				took, timedOut, scrollId,
-				timeoutManager.remainingTimeToHardTimeout()
+				timeoutManager
 		);
 	}
 
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/Elasticsearch7SearchResultExtractor.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/Elasticsearch7SearchResultExtractor.java
@@ -17,6 +17,7 @@ import org.hibernate.search.backend.elasticsearch.gson.impl.JsonObjectAccessor;
 import org.hibernate.search.backend.elasticsearch.search.aggregation.impl.ElasticsearchSearchAggregation;
 import org.hibernate.search.backend.elasticsearch.search.projection.impl.ElasticsearchSearchProjection;
 import org.hibernate.search.backend.elasticsearch.search.projection.impl.SearchProjectionExtractContext;
+import org.hibernate.search.backend.elasticsearch.search.timeout.impl.ElasticsearchTimeoutManager;
 import org.hibernate.search.backend.elasticsearch.work.impl.ElasticsearchSearchResultExtractor;
 import org.hibernate.search.engine.search.aggregation.AggregationKey;
 import org.hibernate.search.engine.search.loading.spi.ProjectionHitMapper;
@@ -63,7 +64,8 @@ class Elasticsearch7SearchResultExtractor<H> implements ElasticsearchSearchResul
 	}
 
 	@Override
-	public ElasticsearchLoadableSearchResult<H> extract(JsonObject responseBody, Long hardTimeoutInMilliseconds) {
+	public ElasticsearchLoadableSearchResult<H> extract(JsonObject responseBody,
+			ElasticsearchTimeoutManager timeoutManager) {
 		ElasticsearchSearchQueryExtractContext extractContext = requestContext.createExtractContext(
 				responseBody
 		);
@@ -81,9 +83,6 @@ class Elasticsearch7SearchResultExtractor<H> implements ElasticsearchSearchResul
 		Integer took = TOOK_ACCESSOR.get( responseBody ).get();
 		Boolean timedOut = TIMED_OUT_ACCESSOR.get( responseBody ).get();
 
-		Long remainingTimeToHardTimeout = ( hardTimeoutInMilliseconds == null ) ? null :
-				hardTimeoutInMilliseconds - took;
-
 		return new ElasticsearchLoadableSearchResult<>(
 				extractContext,
 				rootProjection,
@@ -91,7 +90,7 @@ class Elasticsearch7SearchResultExtractor<H> implements ElasticsearchSearchResul
 				extractedHits,
 				extractedAggregations,
 				took, timedOut, scrollId,
-				remainingTimeToHardTimeout
+				timeoutManager.remainingTimeToHardTimeout()
 		);
 	}
 
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/Elasticsearch7SearchResultExtractor.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/Elasticsearch7SearchResultExtractor.java
@@ -63,7 +63,7 @@ class Elasticsearch7SearchResultExtractor<H> implements ElasticsearchSearchResul
 	}
 
 	@Override
-	public ElasticsearchLoadableSearchResult<H> extract(JsonObject responseBody) {
+	public ElasticsearchLoadableSearchResult<H> extract(JsonObject responseBody, Long hardTimeoutInMilliseconds) {
 		ElasticsearchSearchQueryExtractContext extractContext = requestContext.createExtractContext(
 				responseBody
 		);
@@ -81,13 +81,17 @@ class Elasticsearch7SearchResultExtractor<H> implements ElasticsearchSearchResul
 		Integer took = TOOK_ACCESSOR.get( responseBody ).get();
 		Boolean timedOut = TIMED_OUT_ACCESSOR.get( responseBody ).get();
 
+		Integer remainingTimeToHardTimeout = ( hardTimeoutInMilliseconds == null ) ? null :
+				Math.toIntExact( hardTimeoutInMilliseconds ) - took;
+
 		return new ElasticsearchLoadableSearchResult<>(
 				extractContext,
 				rootProjection,
 				hitCount,
 				extractedHits,
 				extractedAggregations,
-				took, timedOut, scrollId
+				took, timedOut, scrollId,
+				remainingTimeToHardTimeout
 		);
 	}
 
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryImpl.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryImpl.java
@@ -109,11 +109,13 @@ public class ElasticsearchSearchQueryImpl<H> extends AbstractSearchQuery<H, Elas
 
 	@Override
 	public ElasticsearchSearchResult<H> fetch(Integer offset, Integer limit) {
+		timeoutManager.start();
 		NonBulkableWork<ElasticsearchLoadableSearchResult<H>> work = searchWorkBuilder()
 				.paging( defaultedLimit( limit, offset ), offset )
 				.build();
 
-		return Futures.unwrappedExceptionJoin( queryOrchestrator.submit( work ) )
+		ElasticsearchSearchResultImpl<H> result = Futures.unwrappedExceptionJoin(
+				queryOrchestrator.submit( work ) )
 				/*
 				 * WARNING: the following call must run in the user thread.
 				 * If we introduce async query execution, we will have to add a loadAsync method here,
@@ -122,10 +124,14 @@ public class ElasticsearchSearchQueryImpl<H> extends AbstractSearchQuery<H, Elas
 				 * so we may choose to throw exceptions for those.
 				 */
 				.loadBlocking();
+		timeoutManager.stop();
+		return result;
 	}
 
 	@Override
 	public long fetchTotalHitCount() {
+		timeoutManager.start();
+
 		JsonObject filteredPayload = new JsonObject();
 		Optional<JsonObject> querySubTree = JsonAccessor.root().property( "query" ).asObject().get( payload );
 		if ( querySubTree.isPresent() ) {
@@ -143,7 +149,9 @@ public class ElasticsearchSearchQueryImpl<H> extends AbstractSearchQuery<H, Elas
 						ElasticsearchSearchRequestTransformerContextImpl.createTransformerFunction( requestTransformer )
 				);
 		NonBulkableWork<Long> work = builder.build();
-		return Futures.unwrappedExceptionJoin( queryOrchestrator.submit( work ) );
+		Long result = Futures.unwrappedExceptionJoin( queryOrchestrator.submit( work ) );
+		timeoutManager.stop();
+		return result;
 	}
 
 	@Override
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryImpl.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryImpl.java
@@ -158,9 +158,8 @@ public class ElasticsearchSearchQueryImpl<H> extends AbstractSearchQuery<H, Elas
 	public ElasticsearchSearchScroll<H> scroll(int chunkSize) {
 		String scrollTimeoutString = this.scrollTimeout + "s";
 
-		NonBulkableWork<ElasticsearchLoadableSearchResult<H>> firstScroll = searchWorkBuilder()
-				.scrolling( chunkSize, scrollTimeoutString )
-				.build();
+		SearchWorkBuilder<ElasticsearchLoadableSearchResult<H>> firstScroll = searchWorkBuilder()
+				.scrolling( chunkSize, scrollTimeoutString );
 
 		return new ElasticsearchSearchScrollImpl<>( queryOrchestrator, workFactory, searchResultExtractor,
 				scrollTimeoutString, firstScroll, timeoutManager );
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryImpl.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryImpl.java
@@ -157,7 +157,11 @@ public class ElasticsearchSearchQueryImpl<H> extends AbstractSearchQuery<H, Elas
 				.scrolling( chunkSize, scrollTimeoutString )
 				.build();
 
-		return new ElasticsearchSearchScrollImpl<>( queryOrchestrator, workFactory, searchResultExtractor, scrollTimeoutString, firstScroll );
+		Long hardTimeoutInMilliseconds = ( exceptionOnTimeout && timeoutUnit != null && timeoutValue != null ) ?
+				timeoutUnit.toMillis( timeoutValue ) : null;
+
+		return new ElasticsearchSearchScrollImpl<>( queryOrchestrator, workFactory, searchResultExtractor,
+				scrollTimeoutString, firstScroll, hardTimeoutInMilliseconds );
 	}
 
 	@Override
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryImpl.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryImpl.java
@@ -21,6 +21,7 @@ import org.hibernate.search.backend.elasticsearch.search.query.ElasticsearchSear
 import org.hibernate.search.backend.elasticsearch.search.query.ElasticsearchSearchRequestTransformer;
 import org.hibernate.search.backend.elasticsearch.search.query.ElasticsearchSearchResult;
 import org.hibernate.search.backend.elasticsearch.search.query.ElasticsearchSearchScroll;
+import org.hibernate.search.backend.elasticsearch.search.timeout.impl.ElasticsearchTimeoutManager;
 import org.hibernate.search.backend.elasticsearch.util.spi.URLEncodedString;
 import org.hibernate.search.backend.elasticsearch.work.builder.factory.impl.ElasticsearchWorkBuilderFactory;
 import org.hibernate.search.backend.elasticsearch.work.builder.impl.CountWorkBuilder;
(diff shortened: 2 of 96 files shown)


integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/search/loading/SearchQueryEntityLoadingBaseIT.java
```diff
@@ -62,6 +63,22 @@ public void simple() {
...
+
+		persistThatManyEntities( entityCount );
+
+		testLoadingThatManyEntities(
+				session -> { }, // No particular session setup
+				o -> { }, // No particular loading option
+				entityCount,
+				// Only one entity type means only one statement should be executed, even if there are multiple hits
+				c -> c.assertStatementExecutionCount().isEqualTo( 1 ),
+				1, TimeUnit.SECONDS
```
yrodiere, line 78: You can't just put a timeout and not make any assertion related to that timeout. The test, as it is, is just a smoke test.

You need to test at least these cases:

* `failAfter` with a very large timeout: expect everything to work fine.
* `failAfter` with a small-ish timeout (say, 1ms) and very slow loading: expect an exception to be thrown.
* `failAfter` with a large timeout (say, 500ms), very slow index response (say, 499ms) and normal loading: expect an exception to be thrown.
* And maybe more. Have a look at your code and see if some obvious edge cases are not covered by the above?

To create the right conditions for your tests (slow index response time, slow loading), you have two options.

The cleanest approach would be to stub the `TimingSource`. However, it might not be very easy, and if done incorrectly you may end up not testing what you want.

The admittedly more dirty, but more straightforward approach is to voluntarily slow down your components:

* To simulate slow index response time, just pass a custom behavior to `org.hibernate.search.util.impl.integrationtest.common.rule.BackendMock#expectSearchObjects(java.util.Collection<java.lang.String>, java.util.function.Consumer<org.hibernate.search.util.impl.integrationtest.common.stub.backend.search.StubSearchWork.Builder>, org.hibernate.search.util.impl.integrationtest.common.rule.StubSearchWorkBehavior<org.hibernate.search.engine.backend.common.DocumentReference>)`, and in that behavior call `Thread.sleep()`.
* To simulate slow loading, you can add a `SessionEventListener` through `SessionBuilderImpl#eventListeners`, and in `SessionEventListener#jdbcPrepareStatementEnd()` call `Thread.sleep()`.

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/rule/SearchWorkCall.java
```diff
@@ -122,4 +123,16 @@ public String toString() {
...
+	static Integer getLoadingTimeout(StubSearchWork work) {
+		Long timeout = work.getFailAfterTimeout();
+		TimeUnit timeUnit = work.getFailAfterTimeUnit();
+		if ( timeout == null || timeUnit == null ) {
+			return null;
+		}
+
+		long nanos = timeUnit.toNanos( timeout );
+		int millis = Math.toIntExact( nanos / 1000000 );
+		return ( nanos % 1000000 == 0 ) ? millis : millis + 1;
```
yrodiere, line 135: I believe this code was copy/pasted from the Lucene backend? Maybe it would be a good idea to move it to some shared util class.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/Elasticsearch7SearchResultExtractor.java
```diff
@@ -81,13 +81,17 @@
 		Integer took = TOOK_ACCESSOR.get( responseBody ).get();
 		Boolean timedOut = TIMED_OUT_ACCESSOR.get( responseBody ).get();
 
+		Integer remainingTimeToHardTimeout = ( hardTimeoutInMilliseconds == null ) ? null :
+				Math.toIntExact( hardTimeoutInMilliseconds ) - took;
```
yrodiere: That's hardly a good way to compute the remaining time. `took` is the execution time on the server side. It ignores network transfer time completely.

The "right" way to handle these things would be to keep track of when the query started executing on the client side. Then here you would be able to know exactly how much time elapsed. That's exactly what we do in the Lucene backend, and I believe that would be the only correct solution here too.

On a side note, this solution would also allow us to solve https://hibernate.atlassian.net/browse/HSEARCH-3774 (in another PR, obviously).

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/loading/impl/HibernateOrmByTypeEntityLoader.java
```diff
@@ -47,7 +47,8 @@
 				referencesByDelegate.entrySet() ) {
 			HibernateOrmComposableEntityLoader<? extends T> delegate = entry.getKey();
 			List<EntityReference> referencesForDelegate = entry.getValue();
-			delegate.loadBlocking( referencesForDelegate, objectsByReference );
+			// FIXME Use a deadline instead of a timeout
```
yrodiere: Yep. Pass the `TimeoutManager`, for instance.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/loading/impl/HibernateOrmNonEntityIdPropertyEntityLoader.java
```diff
@@ -113,13 +114,16 @@ public void loadBlocking(List<EntityReference> references, Map<? super EntityRef
...
-	private Query<? extends E> createQuery(int fetchSize) {
+	private Query<? extends E> createQuery(int fetchSize, Integer timeout) {
 		Query<? extends E> query = (Query<? extends E>) HibernateOrmUtils.createQueryForLoadByUniqueProperty(
 				session, entityPersister, documentIdSourcePropertyName,
 				DOCUMENT_ID_SOURCE_PROPERTY_PARAMETER_NAME
 		);
 
 		query.setFetchSize( fetchSize );
+		if ( timeout != null ) {
+			query.setTimeout( HibernateOrmComposableEntityLoader.getTimeoutInSeconds( timeout ) );
```
yrodiere, line 125: Rounding up the timeout to the next second is definitely not acceptable for search queries. People need sub-second response times for the overwhelming majority of use cases.

Use `query.setHint( QueryHints.SPEC_HINT_TIMEOUT, <value in milliseconds> )` instead.

engine/src/main/java/org/hibernate/search/engine/common/impl/DelegatingBuildContext.java
```diff
@@ -44,4 +46,12 @@ public ThreadPoolProvider threadPoolProvider() {
 	public FailureHandler failureHandler() {
 		return delegate.getFailureHandler();
 	}
+
+	public EngineThreads getEngineThreads() {
```
yrodiere, line 50: I think you mean `engineThreads()`? :)

engine/src/main/java/org/hibernate/search/engine/common/impl/SearchIntegrationBuilderImpl.java
```diff
@@ -194,11 +199,14 @@ public SearchIntegrationPartialBuildState prepareBuild() {
 
 			threadProviderHolder = THREAD_PROVIDER.getAndTransform( propertySource, beanResolver::resolve );
 			ThreadPoolProviderImpl threadPoolProvider = new ThreadPoolProviderImpl( threadProviderHolder );
+			engineThreads = new EngineThreads( threadPoolProvider );
+			timingSource = new DefaultTimingSource( engineThreads );
```
yrodiere, line 203: There's a `SuppressingCloser` below; you should probably close these with that closer?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/impl/LuceneBackendImpl.java
```diff
@@ -19,14 +19,14 @@
 import org.hibernate.search.backend.lucene.multitenancy.impl.MultiTenancyStrategy;
 import org.hibernate.search.backend.lucene.orchestration.impl.LuceneSyncWorkOrchestratorImpl;
 import org.hibernate.search.backend.lucene.resources.impl.BackendThreads;
-import org.hibernate.search.backend.lucene.common.timing.impl.TimingSource;
```
yrodiere, line 22: You should probably merge the commit `HSEARCH-3787 Remove lucene-level TimingSource ` with the previous one. Otherwise you're missing out on git's ability to track renamed files and to provide a commit history for a particular file even beyond its renaming.

engine/src/main/java/org/hibernate/search/engine/common/timing/impl/TimingSource.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.engine.common.timing.impl;
```
yrodiere, line 7: If you're going to expose `TimingSource` to other modules (e.g. the backends), you will need to put this interface in an `spi` package.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/timeout/impl/LuceneTimeoutManager.java
```diff
@@ -0,0 +1,89 @@
...
+		timingSource.ensureInitialized();
+	}
+
+	public Counter createCounter() {
+		return new LuceneCounterAdapter( timingSource );
+	}
+
+	@Override
+	protected void onTimedOut() {
+		if ( hasHardTimeout() ) {
```
yrodiere, line 67: I think this should be moved to the superclass, too. Be it only because it's simpler that way, or because we also support "soft" timeouts with the Elasticsearch backend. You could get rid of a few class members in the Elasticsearch backend if it was able to query `hasHardTimeout` directly.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/spi/PojoMappingDelegate.java
```diff
@@ -30,6 +31,8 @@
 
 	FailureHandler failureHandler();
 
+	TimingSource timingSource();
```
yrodiere, line 34: Wouldn't it be simpler to just expose the `TimeoutManager` to the mapper, and not the `TimingSource`? I thought that was the reason you moved the `TimeoutManager` to the engine... I'm a bit confused.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/loading/impl/HibernateOrmByTypeEntityLoader.java
```diff
@@ -23,13 +25,21 @@
...
-	HibernateOrmByTypeEntityLoader(Map<String, HibernateOrmComposableEntityLoader<? extends T>> delegatesByEntityName) {
+	HibernateOrmByTypeEntityLoader(Map<String, HibernateOrmComposableEntityLoader<? extends T>> delegatesByEntityName,
+			TimingSource timingSource) {
 		this.delegatesByEntityName = delegatesByEntityName;
+		this.timingSource = timingSource;
 	}
 
 	@Override
 	public List<T> loadBlocking(List<EntityReference> references, Integer timeout) {
+		TimeoutManager timeoutManager = null;
```
yrodiere, line 38: Here, you could just pass the `TimeoutManager` to this method instead of the `Integer timeout`. I expect this will make a lot of code simpler and safer.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/Elasticsearch7SearchResultExtractor.java
```diff
@@ -81,8 +81,8 @@
 		Integer took = TOOK_ACCESSOR.get( responseBody ).get();
 		Boolean timedOut = TIMED_OUT_ACCESSOR.get( responseBody ).get();
 
-		Integer remainingTimeToHardTimeout = ( hardTimeoutInMilliseconds == null ) ? null :
-				Math.toIntExact( hardTimeoutInMilliseconds ) - took;
+		Long remainingTimeToHardTimeout = ( hardTimeoutInMilliseconds == null ) ? null :
```
yrodiere, line 84: Yes, using `long` all the way is safer.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/stub/MapperEasyMockUtils.java
```diff
@@ -91,4 +101,22 @@ private MapperEasyMockUtils() {
...
+				EasyMockUtils.collectionAnyOrderMatcher( new ArrayList<>( context.loadingMap.keySet() ) ), isNull() ) )
+				.andAnswer( () -> ( (List<R>) EasyMock.getCurrentArguments()[0] ).stream()
+						.map( context.loadingMap::get )
+						.collect( Collectors.toList() ) );
+	}
+
+	private static <R, E> void loadWithTimeout(EntityLoader<R, E> objectLoaderMock,
+			HitMappingDefinitionContext<R, E> context) {
+		expect( objectLoaderMock.loadBlocking(
+				EasyMockUtils.collectionAnyOrderMatcher( new ArrayList<>( context.loadingMap.keySet() ) ), notNull() ) )
```
yrodiere, line 117: Right, good idea. You could even make this more powerful and pass a `Capture<Long>`, so that in your tests, you can check that the number is within a given range.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/search/SearchQueryBaseIT.java
```diff
@@ -178,6 +180,34 @@ public void target_byClass_multipleTypes() {
 		} );
 	}
 
+	@Test
+	public void target_byClass_multipleTypes_entityLoadingTimeout() {
```
yrodiere, line 184: See my comment on your other commit adding mapper-level tests: you need to simulate slow loading, at least.

Also, I'm sure this test currently passes (from time to time) only because your implementation rounds the timeout value to the nearest second.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/search/loading/SearchQueryEntityLoadingMultipleTypesIT.java
```diff
@@ -227,6 +231,52 @@ public void mixedHierarchies() {
 		);
 	}
 
+	@Test
+	public void mixedHierarchies_entityLoadingTimeout() {
```
yrodiere, line 235: Same here: you need to simulate slow loading, at least.

Also, I'm sure this test currently passes (from time to time) only because your implementation rounds the timeout value to the nearest second.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/Elasticsearch7SearchResultExtractor.java
fax4ever: Agreed. I assigned the issue to myself.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/loading/impl/HibernateOrmNonEntityIdPropertyEntityLoader.java
fax4ever: Right! I forgot we have this option

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/timeout/impl/LuceneTimeoutManager.java
fax4ever: Sure. Originally I didn't think that Elasicsearch could use it.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/stub/MapperEasyMockUtils.java
fax4ever: yep makes sense to use a spy for that

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/rule/SearchWorkCall.java
fax4ever: In Lucene backend we're using

``` java
timeoutUnit.toMillis( timeoutValue )
```
Probably it would be better doing the same here.

fax4ever: Or maybe not, since the method handles null inputs and always perform a round up of the values.
Moving to an util class...

``` java
@Test
public void test() {
	long l = TimeUnit.NANOSECONDS.toMillis( 1L );
	Assertions.assertThat( l ).isEqualTo( 0 );

	Long aLong = ConvertUtils.toMilliseconds( 1L, TimeUnit.NANOSECONDS );
	Assertions.assertThat( aLong ).isEqualTo( 1L );
}
```

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/loading/impl/HibernateOrmNonEntityIdPropertyEntityLoader.java
fax4ever: Anyway I'm not sure that with `QueryHints.SPEC_HINT_TIMEOUT` we can have more precision. See `AbstractProducedQuery#setHint`, in particular the if branch when `SPEC_HINT_TIMEOUT.equals( hintName )`.

fax4ever: Probably it is better to use `HINT_TIMEOUT`.

engine/src/main/java/org/hibernate/search/engine/common/impl/SearchIntegrationBuilderImpl.java
fax4ever: Yes the idea was to have the variables declared in the outer block for that... but I forget to add the code lines to close them on error on startup...

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/loading/impl/HibernateOrmNonEntityIdPropertyEntityLoader.java
yrodiere: > Anyway I'm not sure that with QueryHints.SPEC_HINT_TIMEOUT we can have more precision. See AbstractProducedQuery#setHint, in particular the if branch when SPEC_HINT_TIMEOUT.equals( hintName ).

Well, that's underwhelming... Looks like JDBC only allows second precision. I guess at least we'll be ready for a potential future version of the spec, if they ever have a look at that...

What's more annoying is how are we going to test this... We need a query that takes significantly more than one second to execute, or we need to mock the `TimingSource` (which has its own disadvantage, as I mentioned over the chat). Let's discuss this on Zulip if you need to.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/spi/PojoMappingDelegate.java
fax4ever: In the current solution the engine can provide the timing source, this instance is shared with all the modules that need it. On the other hand the timeout manager is created each time a module needs it by the module itself. So that the module can control its lifecycle. Another benefit is the fact that a client module can choose its own subclass implementation, e.g.: `LuceneTimeoutManager`.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/search/loading/SearchQueryEntityLoadingBaseIT.java
fax4ever: I added extensive (I think) for the first and second cases. About the last case you mentioned. IMO this would test the time capability of the stub backend to track time and to pass the right difference to the mapper. I don't thing it would make much sense.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/loading/impl/HibernateOrmByTypeEntityLoader.java
```diff
@@ -47,7 +48,12 @@
 				referencesByDelegate.entrySet() ) {
 			HibernateOrmComposableEntityLoader<? extends T> delegate = entry.getKey();
 			List<EntityReference> referencesForDelegate = entry.getValue();
-			delegate.loadBlocking( referencesForDelegate, objectsByReference );
+
+			delegate.loadBlocking( referencesForDelegate, objectsByReference, timeoutManager );
+		}
+
+		if ( timeoutManager != null ) {
+			timeoutManager.stop();
```
yrodiere, line 56: I believe this should be called by whoever called `start`, not by this class.

util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/TimeoutLoadingListener.java
```diff
@@ -0,0 +1,24 @@
...
+package org.hibernate.search.util.impl.integrationtest.mapper.orm;
+
+import org.hibernate.BaseSessionEventListener;
+import org.hibernate.QueryTimeoutException;
+import org.hibernate.Session;
+import org.hibernate.engine.spi.SessionImplementor;
+
+public class TimeoutLoadingListener extends BaseSessionEventListener {
+
+	public static void registerSlowerLoadingListener(Session session) {
```
yrodiere, line 16: Maybe `registerTimingOutLoadingListener`?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/timeout/impl/ElasticsearchTimeoutManager.java
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	@Override
+	protected void onTimedOut() {
+		if ( hasHardTimeout() ) {
+			throw log.clientSideTimedOut( Duration.ofNanos( timeoutUnit.toNanos( timeoutValue ) ), query.toString() );
+		}
+	}
+
+	public boolean exceptionOnTimeout() {
```
yrodiere, line 57: There's already a method that does that: `hasHardTimeout`.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/search/loading/impl/HibernateOrmByTypeEntityLoader.java
fax4ever: Good catch. Right, the timeout lifecycle is handled by the backends.

fax4ever:

As you pointed out, I removed the HSEARCH-3774 from the scope of this pull request and did the three changes you suggested above. I'm going to create a new pull request to handle HSEARCH-3774  properly.
Meanwhile, is there anything else we can improve here?

fax4ever:
Here is a build with all supported evns: https://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-2333/21/

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryImpl.java
```diff
@@ -222,6 +234,7 @@ private Integer defaultedLimit(Integer limit, Integer offset) {
 	}
 
 	private JsonObject doExplain(ElasticsearchSearchIndexContext index, Object id) {
+		timeoutManager.start();
```
yrodiere, line 237: I just noticed this... you're not supposed to use the timeout manager during `explain()` (cf. documentation). At least we made sure not to use it in the Lucene backend.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryImpl.java
```diff
@@ -141,23 +144,32 @@ public long fetchTotalHitCount() {
...
 		NonBulkableWork<Long> work = builder.build();
-		return Futures.unwrappedExceptionJoin( queryOrchestrator.submit( work ) );
+		Long result = Futures.unwrappedExceptionJoin( queryOrchestrator.submit( work ) );
+		timeoutManager.stop();
+		return result;
 	}
 
 	@Override
 	public ElasticsearchSearchScroll<H> scroll(int chunkSize) {
+		timeoutManager.start();
```
yrodiere, line 159: I just noticed this... you're not supposed to use the timeout manager during the initial call to `scroll()`, since we don't send a request to Elasticsearch.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryImpl.java
fax4ever: I agree, I'm removing it...

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryImpl.java
fax4ever: Actually with the initial call of scroll() on Elasticsearch we got the first result page of next() you know. So, IMO makes sense to keep it.

fax4ever: Done and more important this remove discover an inner bug solved by this commit https://github.com/hibernate/hibernate-search/pull/2333/commits/6e95d93879d1d99a0b53e26bfe2c1b7154bd608d.

fax4ever:
Address the comments, thanks. When a full CI build passes I'll merge it.

PR #2335, opened 2020-08-06
HSEARCH-3974 Backend-specific extensions for scroll

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3974


yrodiere:
Thanks, merging. Regarding the follow-up, we're discussing this on Zulip.

PR #2336, opened 2020-08-06
HSEARCH-3975 Expose scrolling operations through JPA/ORM queries

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3975


yrodiere:
Thanks; I added a few tests. Will merge as soon as CI passes.

PR #2339, opened 2020-08-17
HSEARCH-3517 Optimize the computation of total hit count

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3517

~Draft since the last point of the description of the issue is missing. The one to stop the search when it's certain we cannot find any more hits.~

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryBuilder.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryBuilder.java
@@ -72,6 +72,7 @@ public class ElasticsearchSearchQueryBuilder<H>
 	private Long timeoutValue;
 	private TimeUnit timeoutUnit;
 	private boolean exceptionOnTimeout;
+	private Integer totalHitCountMinimum;
 	private ElasticsearchSearchRequestTransformer requestTransformer;
 
 	public ElasticsearchSearchQueryBuilder(
@@ -121,6 +122,11 @@ public class ElasticsearchSearchQueryBuilder<H>
 		this.timeoutValue = timeout;
 		this.timeoutUnit = timeUnit;
 		this.exceptionOnTimeout = true;
+	}
+
+	@Override
+	public void totalHitCountMinimum(int totalHitCountMinimum) {
+		this.totalHitCountMinimum = totalHitCountMinimum;
 	}
 
 	@Override
@@ -229,7 +235,7 @@ public class ElasticsearchSearchQueryBuilder<H>
 				payload, requestTransformer,
 				searchResultExtractor,
 				timeoutManager,
-				scrollTimeout
+				scrollTimeout, totalHitCountMinimum
 		);
 	}
 }
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryBuilder.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryBuilder.java
@@ -72,7 +72,7 @@ public class ElasticsearchSearchQueryBuilder<H>
 	private Long timeoutValue;
 	private TimeUnit timeoutUnit;
 	private boolean exceptionOnTimeout;
-	private Integer totalHitCountMinimum;
+	private Integer totalHitsThreshold;
 	private ElasticsearchSearchRequestTransformer requestTransformer;
 
 	public ElasticsearchSearchQueryBuilder(
@@ -125,8 +125,8 @@ public class ElasticsearchSearchQueryBuilder<H>
 	}
 
 	@Override
-	public void totalHitCountMinimum(int totalHitCountMinimum) {
-		this.totalHitCountMinimum = totalHitCountMinimum;
+	public void totalHitsThreshold(int totalHitsThreshold) {
+		this.totalHitsThreshold = totalHitsThreshold;
 	}
 
 	@Override
@@ -235,7 +235,7 @@ public class ElasticsearchSearchQueryBuilder<H>
 				payload, requestTransformer,
 				searchResultExtractor,
 				timeoutManager,
-				scrollTimeout, totalHitCountMinimum
+				scrollTimeout, totalHitsThreshold
 		);
 	}
 }
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryBuilder.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryBuilder.java
@@ -72,7 +72,7 @@ public class ElasticsearchSearchQueryBuilder<H>
 	private Long timeoutValue;
 	private TimeUnit timeoutUnit;
 	private boolean exceptionOnTimeout;
-	private Integer totalHitsThreshold;
+	private Integer totalHitCountThreshold;
 	private ElasticsearchSearchRequestTransformer requestTransformer;
 
 	public ElasticsearchSearchQueryBuilder(
@@ -125,8 +125,8 @@ public class ElasticsearchSearchQueryBuilder<H>
 	}
 
 	@Override
-	public void totalHitsThreshold(int totalHitsThreshold) {
-		this.totalHitsThreshold = totalHitsThreshold;
+	public void totalHitCountThreshold(int totalHitCountThreshold) {
+		this.totalHitCountThreshold = totalHitCountThreshold;
 	}
 
 	@Override
@@ -235,7 +235,7 @@ public class ElasticsearchSearchQueryBuilder<H>
 				payload, requestTransformer,
 				searchResultExtractor,
 				timeoutManager,
-				scrollTimeout, totalHitsThreshold
+				scrollTimeout, totalHitCountThreshold
 		);
 	}
 }
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryBuilder.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryBuilder.java
@@ -72,7 +72,7 @@ public class ElasticsearchSearchQueryBuilder<H>
 	private Long timeoutValue;
 	private TimeUnit timeoutUnit;
 	private boolean exceptionOnTimeout;
-	private Integer totalHitCountThreshold;
+	private Long totalHitCountThreshold;
 	private ElasticsearchSearchRequestTransformer requestTransformer;
 
 	public ElasticsearchSearchQueryBuilder(
@@ -125,7 +125,7 @@ public class ElasticsearchSearchQueryBuilder<H>
 	}
 
 	@Override
-	public void totalHitCountThreshold(int totalHitCountThreshold) {
+	public void totalHitCountThreshold(long totalHitCountThreshold) {
 		this.totalHitCountThreshold = totalHitCountThreshold;
 	}
 
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/impl/SearchWork.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/impl/SearchWork.java
@@ -78,7 +78,7 @@ public class SearchWork<R> extends AbstractNonBulkableWork<R> {
 		private final Set<URLEncodedString> indexes = new HashSet<>();
 
 		private Boolean trackTotalHits;
-		private Integer totalHitCountMinimum;
+		private Integer totalHitsThreshold;
 		private Integer from;
 		private Integer size;
 		private Integer scrollSize;
@@ -138,11 +138,11 @@ public class SearchWork<R> extends AbstractNonBulkableWork<R> {
 		}
 
 		@Override
-		public SearchWorkBuilder<R> totalHitCountMinimum(Integer totalHitCountMinimum) {
+		public SearchWorkBuilder<R> totalHitsThreshold(Integer totalHitsThreshold) {
 			// setting trackTotalHits to false only if this parameter was already set,
 			// the parameter is not supported by the older Elasticsearch server
 			if ( trackTotalHits != null && trackTotalHits ) {
-				this.totalHitCountMinimum = totalHitCountMinimum;
+				this.totalHitsThreshold = totalHitsThreshold;
 			}
 			return this;
 		}
@@ -173,9 +173,9 @@ public class SearchWork<R> extends AbstractNonBulkableWork<R> {
 			}
 
 			if ( trackTotalHits != null ) {
-				if ( trackTotalHits && totalHitCountMinimum != null ) {
+				if ( trackTotalHits && totalHitsThreshold != null ) {
 					// total hits is tracked but a with a limited precision
-					builder.param( "track_total_hits", totalHitCountMinimum );
+					builder.param( "track_total_hits", totalHitsThreshold );
 				}
 				else {
 					builder.param( "track_total_hits", trackTotalHits );
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/impl/SearchWork.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/impl/SearchWork.java
@@ -78,6 +78,7 @@ public class SearchWork<R> extends AbstractNonBulkableWork<R> {
 		private final Set<URLEncodedString> indexes = new HashSet<>();
 
 		private Boolean trackTotalHits;
+		private Integer totalHitCountMinimum;
 		private Integer from;
 		private Integer size;
 		private Integer scrollSize;
@@ -137,6 +138,16 @@ public class SearchWork<R> extends AbstractNonBulkableWork<R> {
 		}
 
 		@Override
+		public SearchWorkBuilder<R> totalHitCountMinimum(Integer totalHitCountMinimum) {
+			// setting trackTotalHits to false only if this parameter was already set,
+			// the parameter is not supported by the older Elasticsearch server
+			if ( trackTotalHits != null && trackTotalHits ) {
+				this.totalHitCountMinimum = totalHitCountMinimum;
+			}
+			return this;
+		}
+
+		@Override
 		protected ElasticsearchRequest buildRequest() {
 			ElasticsearchRequest.Builder builder =
 					ElasticsearchRequest.post()
@@ -162,7 +173,13 @@ public class SearchWork<R> extends AbstractNonBulkableWork<R> {
 			}
 
 			if ( trackTotalHits != null ) {
-				builder.param( "track_total_hits", trackTotalHits );
+				if ( trackTotalHits && totalHitCountMinimum != null ) {
+					// total hits is tracked but a with a limited precision
+					builder.param( "track_total_hits", totalHitCountMinimum );
+				}
+				else {
+					builder.param( "track_total_hits", trackTotalHits );
+				}
 			}
 
 			if ( timeoutManager.defined() ) {
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/impl/SearchWork.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/impl/SearchWork.java
@@ -69,16 +69,15 @@ public class SearchWork<R> extends AbstractNonBulkableWork<R> {
 		}
 
 		public static <T> Builder<T> forElasticsearch7AndAbove(JsonObject payload, ElasticsearchSearchResultExtractor<T> resultExtractor) {
-			// TODO HSEARCH-3517 disable track_total_hits when possible
 			return new Builder<>( payload, resultExtractor, true, false );
 		}
 
 		private final JsonObject payload;
 		private final ElasticsearchSearchResultExtractor<R> resultExtractor;
-		private final Boolean trackTotalHits;
 		private final boolean allowPartialSearchResultsSupported;
 		private final Set<URLEncodedString> indexes = new HashSet<>();
 
+		private Boolean trackTotalHits;
 		private Integer from;
 		private Integer size;
 		private Integer scrollSize;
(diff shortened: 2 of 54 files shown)


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/impl/SearchWork.java
```diff
@@ -127,6 +126,14 @@ private Builder(JsonObject payload, ElasticsearchSearchResultExtractor<R> result
 			return this;
 		}
 
+		@Override
+		public SearchWorkBuilder<R> disableTrackTotalHits() {
+			if ( trackTotalHits != null && trackTotalHits ) {
```
yrodiere, line 131: Why this check? Someone is asking to disable track total hits, we should just do it. No need for complex logic here.

engine/src/main/java/org/hibernate/search/engine/search/query/SearchResult.java
```diff
@@ -19,7 +19,7 @@
 	/**
 	 * @return The total number of matching entities, ignoring pagination settings.
 	 */
-	long totalHitCount();
+	Long totalHitCount();
```
yrodiere, line 22: As discussed, this shouldn't be. I believe you only had to do that because you use `SearchResult` as a return even when fetching just hits, even though that interface only makes sense when we actually have a total hit count.

I can see two solutions:

1. Use a `long`, but make it default to `0` when we don't know. Dirty but effective. Note it will make more sense and will be cleaner when we're able to mark the number as a "lower bound": `0` is always a lower bound of the total hit count, so the method would actually be telling the truth.
2. Change the flow of information so that there are two different paths:
  * When loading the whole result (including the total hit count, call `ElasticsearchLoadableSearchResult#loadBlocking()` as we used to. (Same for Lucene)
  * When loading just hits, call a new method: call `ElasticsearchLoadableSearchResult#loadHitsBlocking()`, which only returns a `List<H>`. (Same for Lucene). It shouldn't be too hard since `loadBlocking` is already mostly just about loading hits; I'm sure you'll be able to factorize code easily.

engine/src/main/java/org/hibernate/search/engine/search/query/spi/AbstractSearchQuery.java
```diff
@@ -54,21 +54,18 @@ public R fetch(Integer limit) {
...
-		List<H> hits = result.hits();
-		int fetchedHitCount = result.hits().size();
+		List<H> hits = fetchHits( 2 );
+		int fetchedHitCount = hits.size();
 		if ( fetchedHitCount == 0 ) {
 			return Optional.empty();
 		}
 		else if ( fetchedHitCount > 1 ) {
-			throw log.nonSingleHit( result.totalHitCount() );
+			throw log.nonSingleHit( fetchedHitCount );
```
yrodiere, line 68: This value will always be 2, while the `totalHitCount()` we used previously was actually informational. Please remove this parameter and change the log message accordingly: `Multiple hits when a single hit was expected.`.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/query/SearchQueryTimeoutIT.java
```diff
@@ -118,7 +118,7 @@ public void timeout_slowQuery_smallTimeout_limitFetching() {
 				.truncateAfter( 1, TimeUnit.NANOSECONDS )
 				.fetchAll();
 
-		assertThat( result.totalHitCount() ).isLessThan( TOTAL_DOCUMENT_COUNT );
+		assertThat( result.totalHitCount() ).isNull();
```
yrodiere, line 121: That's... surprising. What's going on here?

yrodiere:
Looks good so far, except for a few comments. You might want to also add a test for Elasticsearch using a client spy to verify the `track_total_hits` parameter is set correctly, e.g. in `org.hibernate.search.integrationtest.backend.elasticsearch.search.query.ElasticsearchSearchQueryIT`?

fax4ever:
@yrodiere I didn't address your comments so fat, I pushed just a clean up of the preexisting work. I'll address it with the next push. Thanks for the review!

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/impl/SearchWork.java
fax4ever: In the current (master) behavior `trackTotalHits` is null in case the version of Elasticsearch is strictly less than 7.
If we tried to set `track_total_hits` in ES 6, we got an error, since the parameter is not supposed to be set.
I just kept the preexisting convention, setting the parameter to `false` only if the parameter was previously set. 
I agree with you that it is not very expressive.

fax4ever: I'm adding a comment, since other changes could have a larger impact...

engine/src/main/java/org/hibernate/search/engine/search/query/SearchResult.java
fax4ever: I fixed it before reading this comment using the solution (1). Anyway this is only temporary, since as you wrote we're going to change the API supporting the lower bound hit counts.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/query/SearchQueryTimeoutIT.java
fax4ever: Already fixed

engine/src/main/java/org/hibernate/search/engine/search/query/spi/AbstractSearchQuery.java
fax4ever: Good catch! Since we don't have the hit count anymore, we don't know the exact repetition number! Makes sense to be generic in the message!

fax4ever:
I launched a full build: https://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-2339/16/

yrodiere:
Added a commit to disable some tests on ES 6.x and below, since `track_total_hits` is only supported in ES 7.0+.

documentation/src/main/asciidoc/reference/search-dsl-query.asciidoc
```diff
@@ -210,6 +211,57 @@ include::{sourcedir}/org/hibernate/search/documentation/search/query/QueryDslIT.
...
+
+.Define a total hits threshold
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/query/QueryDslIT.java[tags=fetching-totalHitsThreshold]
+----
+<1> Define a `totalHitsThreshold` for the current query
+<2> Extract the `SearchResultTotal` from the `SearchResult`.
+<3> Take the lower bound total hit count, it will return the exact hit count if the count is exact.
```
yrodiere, line 258:
```suggestion
<3> Retrieve a lower-bound estimate of the total hit count. This will return the exact hit count if available.
```

documentation/src/main/asciidoc/reference/search-dsl-query.asciidoc
```diff
@@ -210,6 +211,57 @@ include::{sourcedir}/org/hibernate/search/documentation/search/query/QueryDslIT.
...
+<4> Test if the count is exact.
+<5> Test if the count is a lower bound.
```
yrodiere, line 260:
```suggestion
<5> Test if the count is a lower-bound estimate.
```

documentation/src/main/asciidoc/reference/search-dsl-query.asciidoc
```diff
@@ -210,6 +211,57 @@ include::{sourcedir}/org/hibernate/search/documentation/search/query/QueryDslIT.
...
+====
+
+As we said, the count here may not be `exact` but a lower bound.
```
yrodiere, line 263:
```suggestion
```

documentation/src/main/asciidoc/reference/search-dsl-query.asciidoc
```diff
@@ -210,6 +211,57 @@ include::{sourcedir}/org/hibernate/search/documentation/search/query/QueryDslIT.
...
+
+.Working with the result total
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/query/QueryDslIT.java[tags=fetching-resultTotal]
+----
+<1> Extract the `SearchResultTotal` from the `SearchResult`.
+<2> Take the exact total hit count, this call will raise an exception if the hit count is not exact.
+<3> Take the lower bound total hit count, it will return the exact hit count if the count is exact.
```
yrodiere, line 239:
```suggestion
<3> Retrieve a lower-bound estimate of the total hit count. This will return the exact hit count if available.
```

documentation/src/main/asciidoc/reference/search-dsl-query.asciidoc
```diff
@@ -210,6 +211,57 @@ include::{sourcedir}/org/hibernate/search/documentation/search/query/QueryDslIT.
...
+For these cases, the count may not be `exact` but a lower bound.
+
+.Working with the result total
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/query/QueryDslIT.java[tags=fetching-resultTotal]
+----
+<1> Extract the `SearchResultTotal` from the `SearchResult`.
+<2> Take the exact total hit count, this call will raise an exception if the hit count is not exact.
```
yrodiere, line 238:
```suggestion
<2> Retrieve the exact total hit count. This call will raise an exception if the only available hit count is a lower-bound estimate.
```

documentation/src/test/java/org/hibernate/search/documentation/search/query/QueryDslIT.java
```diff
@@ -417,6 +418,49 @@ public void fetchSize() {
...
+	}
+
+	@Test
+	public void resultTotal_totalHitsThreshold() {
+		OrmUtils.withinJPATransaction( entityManagerFactory, entityManager -> {
+			SearchSession searchSession = Search.session( entityManager );
+			// tag::fetching-totalHitsThreshold[]
+			SearchResult<Book> result = searchSession.search( Book.class )
+					.where( f -> f.matchAll() )
+					.totalHitsThreshold( 2 ) // <1>
```
yrodiere, line 451: Maybe use a more realistic value here, e.g. `1000`? It's not a big deal if we don't reach the threshold in this test: we already have tests elsewhere.

yrodiere (changes requested):
Nice job overall. I have a few comments, and a few suggestions to make the wording of the documentation feel "smoother".

documentation/src/main/asciidoc/reference/search-dsl-query.asciidoc
```diff
@@ -210,6 +211,57 @@ include::{sourcedir}/org/hibernate/search/documentation/search/query/QueryDslIT.
...
+to the current page or not. For pagination see <<search-dsl-query-fetching-results-pagination>>.
+The count can be either exact or a lower bound of it.
+
+The total hit count is always exact, except the cases:
+
+* We got a limiting fetch timeout.
+See <<limit_the_number_of_results_when_the_time_limit_is_reached>>.
+* We're using the Elasticsearch backend and the total hit count is higher than `10.000`.
+See https://www.elastic.co/guide/en/elasticsearch/reference/current/search-your-data.html#track-total-hits[Elasticsearch documentation].
+* If we have defined a `totalHitsThreshold` for the current query. See <<search-dsl-query-total-hits-threshold>>.
```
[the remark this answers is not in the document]
yrodiere: Actually, is the part about Elasticsearch even true? I thought we set `track_total_hits` to `true` by default, precisely to avoid this problem?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/query/impl/ElasticsearchSearchQueryBuilder.java
```diff
@@ -125,8 +125,8 @@ public void failAfter(long timeout, TimeUnit timeUnit) {
 	}
 
 	@Override
-	public void totalHitCountMinimum(int totalHitCountMinimum) {
-		this.totalHitCountMinimum = totalHitCountMinimum;
+	public void totalHitsThreshold(int totalHitsThreshold) {
```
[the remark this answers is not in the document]
fax4ever: `totalHitsThreshold` is the Lucene name. but I agree with you, it is more `totalHitCountThreshold`, since the target of the threshold is the count and not the hits.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchScrollImpl.java
```diff
@@ -85,7 +85,7 @@ public void close() {
 		}
 
 		// no more results check
-		if ( scrollIndex >= search.totalHitCount() ) {
+		if ( search.totalHitCount() == null || scrollIndex >= search.totalHitCount() ) {
```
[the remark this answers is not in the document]
fax4ever: Yes, in case of scroll the count is always required to be exact (no disbled - no threshold).
Even Though in case of a limit fetching timeout the hit count can be a lower bound.
That is the case tested by `SearchQueryScrollIT#all_truncateAfter`, in which we got a lower bounded hit count.
I think that for this very case we should not rely on hit count to skip the extraction and go over instead, 
so I would change it. What do you think?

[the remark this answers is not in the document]
yrodiere: Ok. So you want to take the total hit count threshold into account for scrolls. Not sure it's absolutely necessary, but let's try.

Yes, if you want to allow setting a threshold, we should not rely on the total hit count here.

However, we can't just skip this condition: that's basically the terminal condition of user loops when scrolling, which will switch `SearchScrollResult#hasHits()` to `false`.

I think you should mirror what you did for Elasticsearch: don't rely on the total hit count to determine whether it's the last scroll result, but rather on the number of extracted results (before loading), i.e. `hasHits() == !LuceneLoadableSearchResult.extractedData.isEmpty()`. Then you can get rid of the use of totalHitCount here entirely.

By the way, maybe we should always set the totalHitCountThreshold to `0` for scrolls, since we won't use the total hit count at all?
And maybe we should do the same for Elasticsearch, set `track_total_hits` to `false` for scrolls?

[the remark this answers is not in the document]
fax4ever: I think that currently even the Elasticsearch backend uses total hit count to skip the hit extraction if, see `Elasticsearch7SearchResultExtractor#extract` row 65:
``` java
final List<Object> extractedHits = hitCount > 0 ?
  extractHits( extractContext ) : Collections.emptyList();
```
So probably, even if indirectly, `ElasticsearchLoadableSearchResult#hasHits` depends on total hit count too.

Anyway, I agree with you that there is no reason to use hit count to skip, in case it is exactly zero, the extraction. It is better to use the current next/page hitSize. E.g.:

``` java
public boolean hasHits() {
  TopDocs topDocs = luceneCollectors.getTopDocs();
  return ( topDocs == null ) ? false : topDocs.scoreDocs.length > 0;
}
```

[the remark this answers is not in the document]
yrodiere: > I think that currently even the Elasticsearch backend uses total hit count to skip the hit extraction if this is 0, see Elasticsearch7SearchResultExtractor#extract row 65:
> 
> ```
> final List<Object> extractedHits = hitCount > 0 ?
>   extractHits( extractContext ) : Collections.emptyList();
> ```

Not really, this is just an optimization. If the hit count is 0, `extractHits()` would have returned an empty list anyway, so we skip it.

documentation/src/main/asciidoc/reference/search-dsl-query.asciidoc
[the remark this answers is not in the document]
fax4ever: Right. Reading better the ES documentation it seems that passing `true` to `track_total_hits`, as we do, the hit count is always exact.

> When set to true the search response will always track the number of hits that match the query accurately

I'm removing the ES item from the list

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/query/SearchQueryOptionsIT.java
```diff
@@ -91,6 +91,26 @@ public void failAfter() {
...
 	}
 
+	@Test
+	public void totalHitsThreshold() {
+		try ( SearchSession session = mapping.createSession() ) {
+			SearchQuery<EntityReference> query = session.search( IndexedEntity.class )
+					.selectEntityReference()
+					.where( f -> f.matchAll() )
+					// define a total hits threshold
+					.totalHitsThreshold( 100 )
```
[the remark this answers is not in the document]
fax4ever: Removed the entire commit

documentation/src/test/java/org/hibernate/search/documentation/gettingstarted/withhsearch/withoutanalysis/GettingStartedWithoutAnalysisIT.java
```diff
@@ -109,7 +109,7 @@ public void test() {
 					)
```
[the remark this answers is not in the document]
fax4ever: I tried to add a test targeting the deprecated API.

``` java
@Test
public void totalHitCount_deprecated() {
	// we can remove this test when the deprecated API is removed
	initData( 5000 );
	StubMappingScope scope = index.createScope();

	SearchResult<DocumentReference> fetch = scope.query()
			.where( f -> f.matchAll() )
			.toQuery()
			.fetch( 10 );

	assertThat( fetch.totalHitCount() ).isEqualTo( 5000 );
}
```

but it seems that it provokes a compilation errorr

``` bash
[ERROR] COMPILATION ERROR : 
[ERROR] /home/fabio/code/hibernate-search/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/query/SearchQueryBaseIT.java: warnings found and -Werror specified
```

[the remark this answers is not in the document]
yrodiere: Just add `@SuppressWarnings("deprecated")` to the test method.

fax4ever:
@yrodiere thanks for the review. I should have applied all the suggestions.
Here is a new full build: https://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-2339/47/

documentation/src/main/asciidoc/reference/search-dsl-query.asciidoc
```diff
@@ -210,6 +212,57 @@ include::{sourcedir}/org/hibernate/search/documentation/search/query/QueryDslIT.
...
+----
+<1> Extract the `SearchResultTotal` from the `SearchResult`.
+<2> Retrieve the exact total hit count. This call will raise an exception if the only available hit count is a lower-bound estimate.
+<3> Retrieve a lower-bound estimate of the total hit count. This will return the exact hit count if available.
+<4> Test if the count is exact.
+<5> Test if the count is a lower bound.
+====
+
+[[search-dsl-query-total-hits-threshold]]
+=== `totalHitsThreshold(...)`: optimizing total hit count computation
```
yrodiere, line 240:
```suggestion
=== `totalHitCountThreshold(...)`: optimizing total hit count computation
```

documentation/src/main/asciidoc/reference/search-dsl-query.asciidoc
```diff
@@ -210,6 +212,57 @@ include::{sourcedir}/org/hibernate/search/documentation/search/query/QueryDslIT.
...
+
+When working with large result sets, counting the number of hits exactly can be very resource-consuming.
+
+When sorting by score (the default) and retrieving the result through `fetch(...)`, it is possible to yield significant performance improvements by allowing Hibernate Search to return a lower-bound estimate of the total hit count, instead of the exact total hit count. In that case, the underlying engine (Lucene or Elasticsearch) will be able to skip large chunks of non-competitive hits, leading to fewer index scans and thus better performance.
+
+To enable this performance optimization, call `totalHitsThreshold(...)` when building the query, as shown in the example below.
```
yrodiere, line 246:
```suggestion
To enable this performance optimization, call `totalHitCountThreshold(...)` when building the query, as shown in the example below.
```

documentation/src/main/asciidoc/reference/search-dsl-query.asciidoc
```diff
@@ -210,6 +212,57 @@ include::{sourcedir}/org/hibernate/search/documentation/search/query/QueryDslIT.
...
+
+[NOTE]
+====
+This optimization has no effect on calls to `fetchHits(...)` (which is already optimized by default) or to `fetchTotalHitCount()` (which always returns an exact hit count).
+====
+
+.Defining a total hits threshold
```
yrodiere, line 253:
```suggestion
.Defining a total hit count threshold
```

yrodiere (commented):
We're getting there... Please restore the test of `totalHitCount()` (see below) with `@SuppressedWarnings("deprecated")` and finish the `totalHitsThreshold` => `totalHitCountThreshold` renaming, and then let's merge!

documentation/src/test/java/org/hibernate/search/documentation/gettingstarted/withhsearch/withoutanalysis/GettingStartedWithoutAnalysisIT.java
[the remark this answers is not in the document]
fax4ever: I'm restoring it now

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchScrollImpl.java
```diff
@@ -85,9 +85,8 @@ public void close() {
 		}
 
 		// no more results check
-		if ( scrollIndex >= search.totalHitCount() ) {
-			return new LuceneSearchScrollResultImpl<>( false, Collections.emptyList(), Duration.ZERO,
-					false );
+		if ( scrollIndex >= search.hitSize() ) {
```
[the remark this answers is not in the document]
fax4ever: Yep basically when we double, we gather again all the `TopDocs` from the beginning.
We're not supposed to have many doubles, at most a logarithmic times on the input size, so the amortized cost is not so much,, but sure we can improve skipping the hits we have already provided.

fax4ever:
Thanks @yrodiere, I'll merge it as soon the CI passes.

fax4ever:
Merged thanks!

yrodiere:
Thanks!

PR #2340, opened 2020-08-19
HSEARCH-3309 Allow per-index analyzer definitions in Elasticsearch

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3309

I chose a simpler approach than what was described in the ticket: I just moved the analysis configurer to the index level for Elasticsearch. That way, users are free to configure at the backend level or at the index level, and the way to configure at each level is similar (unlike what was suggested in the ticket).

### DIFF
--- a/documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
+++ b/documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
@@ -712,6 +712,23 @@ include::{resourcesdir}/analysis/elasticsearch-simple.properties[]
 <1> Assign the configurer to the backend using a Hibernate Search configuration property.
 ====
 
+[NOTE]
+====
+A different analysis configurer can be assign to each index:
+
+[source]
+----
+# To set the default configurer for all indexes:
+hibernate.search.backend.analysis.configurer com.mycompany.MyAnalysisConfigurer
+# To assign a specific configurer to a specific index:
+hibernate.search.backend.indexes.<index name>.analysis.configurer com.mycompany.MySpecificAnalysisConfigurer
+----
+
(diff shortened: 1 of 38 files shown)


fax4ever:
Right! Moving the definition of analyzers at index level is the more natural approach, since it is the way this backend works.

documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
```diff
@@ -712,6 +712,23 @@ include::{resourcesdir}/analysis/elasticsearch-simple.properties[]
...
+----
+# To set the default configurer for all indexes:
+hibernate.search.backend.analysis.configurer com.mycompany.MyAnalysisConfigurer
+# To assign a specific configurer to a specific index:
+hibernate.search.backend.indexes.<index name>.analysis.configurer com.mycompany.MySpecificAnalysisConfigurer
+----
+
+If a specific configurer is assigned to an index,
+the default configurer will be ignored for that index:
+only definitions from the specific configurer will be taken into account.
```
fax4ever, line 729: If the user needs to reuse definitions in the backend configurer to a specific index configurer, class inheritance can be used. Or cannot it?

fax4ever (approved):
Nice job. Just a naive question. Anyway +1 to merge it

documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
yrodiere: Sure. Inheritance, composition, whatever. It's just code.

yrodiere:
Merged, thanks!

PR #2343, opened 2020-08-27
HSEARCH-3988 Using search(...).select(...) drops the type of loading options and prevents their definition

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3988


fax4ever:
Merged thanks!

PR #2344, opened 2020-09-01
HSEARCH-3980 Expose multiValuedInRoot in the metamodel

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3980

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/backend/metamodel/IndexFieldDescriptor.java
+++ b/engine/src/main/java/org/hibernate/search/engine/backend/metamodel/IndexFieldDescriptor.java
@@ -68,4 +68,10 @@ public interface IndexFieldDescriptor {
 		return multiValued();
 	}
 
+	/**
+	 * @return {@code true} if this field can have multiple values in the same root document
+	 * or if it is contained, directly or indirectly, in an object field that can have multiple values.
+	 */
+	boolean multiValuedInRoot();
+
 }
--- a/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/metamodel/IndexFieldDescriptorIT.java
+++ b/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/metamodel/IndexFieldDescriptorIT.java
@@ -98,7 +98,11 @@ public class IndexFieldDescriptorIT {
 		assertThat( fieldDescriptor )
 				.returns( getRelativeFieldName(), IndexFieldDescriptor::relativeName )
 				.returns( getAbsoluteFieldPath(), IndexFieldDescriptor::absolutePath )
-				.returns( fieldStructure.isMultiValued(), IndexFieldDescriptor::multiValued );
+				.returns( fieldStructure.isMultiValued(), IndexFieldDescriptor::multiValued )
+				// In this specific test, we decided that nested fields are multi-valued, while flattened fields are not.
+				// That's just a detail though: the two characteristics are not usually related.
+				.returns( fieldStructure.isMultiValued() || fieldStructure.isInNested(),
+						IndexFieldDescriptor::multiValuedInRoot );
 
 		// Type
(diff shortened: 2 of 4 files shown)


engine/src/main/java/org/hibernate/search/engine/backend/metamodel/IndexFieldDescriptor.java
```diff
@@ -68,4 +68,10 @@ default boolean isMultiValued() {
 		return multiValued();
 	}
 
+	/**
+	 * @return {@code true} if this field can have multiple values in the same parent document
```
yrodiere, line 72:
```suggestion
	 * @return {@code true} if this field can have multiple values in the same root document
```

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/util/TestedFieldStructure.java
```diff
@@ -89,4 +89,9 @@ public boolean isSingleValued() {
 	public boolean isMultiValued() {
 		return !isSingleValued();
 	}
+
+	public boolean isMultiValuedInRoot() {
+		// all nested object mappings made by tests are multivalued
+		return isMultiValued() || isInNested();
```
yrodiere, line 95: That's true in `IndexFieldDescriptorIT`, but not everywhere. See the comment in that test:

```
				// In this specific test, we decided that nested fields are multi-valued, while flattened fields are not.
				// That's just a detail though: the two characteristics are not usually related.
```

Maybe you shouldn't change `IndexFieldLocation`/`TestFieldStructure`, and move the condition to `IndexFieldDescriptorIT`?

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/util/IndexFieldLocation.java
```diff
@@ -17,6 +17,10 @@
 	IN_FLATTENED,
 	/**
 	 * The field is in a nested object field (nested document).
+	 * <p>
+	 * The nested object is defined as multivalued,
+	 * so that this field will be multivalued in root,
+	 * even if it isn't multivalued.
```
yrodiere, line 23: Same comment as in `TestFieldStructure`: this may not always be true.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/metamodel/IndexFieldDescriptorIT.java
```diff
@@ -98,7 +98,8 @@ public void valueField() {
 		assertThat( fieldDescriptor )
 				.returns( getRelativeFieldName(), IndexFieldDescriptor::relativeName )
 				.returns( getAbsoluteFieldPath(), IndexFieldDescriptor::absolutePath )
-				.returns( fieldStructure.isMultiValued(), IndexFieldDescriptor::multiValued );
+				.returns( fieldStructure.isMultiValued(), IndexFieldDescriptor::multiValued )
+				.returns( fieldStructure.isMultiValuedInRoot(), IndexFieldDescriptor::multiValuedInRoot );
```
yrodiere, line 102: Maybe you could test this for object fields too (i.e. in the `parent()` test method)?

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/util/TestedFieldStructure.java
fax4ever: Right, it does make sense to not change `TestedFieldStructure` but only `TestedFieldStructure`.

fax4ever:
@yrodiere thanks for the good suggestions. I think I've applied all of them.

yrodiere:
Merged, thanks!

PR #2345, opened 2020-09-02
HSEARCH-2505 Lower the default read timeout with Elasticsearch

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-2505

~~Draft since I have to check if we have to change something in our doc...~~

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/cfg/ElasticsearchBackendSettings.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/cfg/ElasticsearchBackendSettings.java
@@ -106,7 +106,7 @@ public final class ElasticsearchBackendSettings {
 	 * Expects a positive Integer value in milliseconds, such as 60000,
 	 * or a String that can be parsed into such Integer value.
 	 * <p>
-	 * Defaults to {@link Defaults#REQUEST_TIMEOUT}.
+	 * Defaults to no request timeout.
 	 */
 	public static final String REQUEST_TIMEOUT = "request_timeout";
 
@@ -267,7 +267,6 @@ public final class ElasticsearchBackendSettings {
 
 		public static final List<String> HOSTS = Collections.singletonList( "localhost:9200" );
 		public static final String PROTOCOL = "http";
-		public static final int REQUEST_TIMEOUT = 60000;
 		public static final int READ_TIMEOUT = 30000;
 		public static final int CONNECTION_TIMEOUT = 1000;
 		public static final int MAX_CONNECTIONS = 20;
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/cfg/ElasticsearchBackendSettings.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/cfg/ElasticsearchBackendSettings.java
@@ -268,8 +268,8 @@ public final class ElasticsearchBackendSettings {
 		public static final List<String> HOSTS = Collections.singletonList( "localhost:9200" );
 		public static final String PROTOCOL = "http";
 		public static final int REQUEST_TIMEOUT = 60000;
-		public static final int READ_TIMEOUT = 60000;
-		public static final int CONNECTION_TIMEOUT = 3000;
+		public static final int READ_TIMEOUT = 30000;
+		public static final int CONNECTION_TIMEOUT = 1000;
 		public static final int MAX_CONNECTIONS = 20;
 		public static final int MAX_CONNECTIONS_PER_ROUTE = 10;
 		public static final boolean DISCOVERY_ENABLED = false;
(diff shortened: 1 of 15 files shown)


fax4ever:
Here is a full build: https://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-2345/4/.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/cfg/ElasticsearchIndexSettings.java
```diff
@@ -117,7 +117,7 @@ private Defaults() {
 		}
 
 		public static final IndexStatus SCHEMA_MANAGEMENT_MINIMAL_REQUIRED_STATUS = IndexStatus.GREEN;
-		public static final int SCHEMA_MANAGEMENT_MINIMAL_REQUIRED_STATUS_WAIT_TIMEOUT = 10_000;
+		public static final int SCHEMA_MANAGEMENT_MINIMAL_REQUIRED_STATUS_WAIT_TIMEOUT = 60_000;
```
yrodiere, line 120: While I think some users will need this value, I don't think we should changedefault.

Imagine a user just starting with the library. By default we require status "green" for all indexes. Imagine the user has only one Elasticsearch node: their indexes cannot be replicated, so they will never reach status "green" and be stuck at "yellow".

Now, without any configuration, startup with block waiting for this "green" status that will never happen; then it will issue an error stating that the status "green" was expected but wasn't reached in 10 seconds.

If you raise the default to 60s, the application will block for **one whole minute** instead... I'd say that's more than enough for most people to consider something went wrong and to kill the application. They'll never see the error message and will be left wondering why the hell everything is stuck.

I think it's better to have a smaller timeout by default, so that inexperienced users will get some feedback on what's wrong exactly. More experienced users that know their cluster is slow can raise the timeout, and *then* all the work you've done in this ticket will be very useful (not to mention it's also useful for long-running search queries anyway).

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/cfg/ElasticsearchBackendSettings.java
```diff
@@ -267,7 +267,6 @@ private Defaults() {
 
 		public static final List<String> HOSTS = Collections.singletonList( "localhost:9200" );
 		public static final String PROTOCOL = "http";
-		public static final int REQUEST_TIMEOUT = 60000;
```
yrodiere, line 270: You should also update the javadoc on `ElasticsearchBackendSettings#REQUEST_TIMEOUT`:

```
	 * Defaults to {@link Defaults#REQUEST_TIMEOUT}.
```

=>

```
	 * Defaults to no request timeout.
```

documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
```diff
@@ -201,9 +201,9 @@ Timeouts::
 +
 [source]
 ----
-hibernate.search.backend.request_timeout = 60000 (default)
-hibernate.search.backend.connection_timeout = 3000 (default)
-hibernate.search.backend.read_timeout = 60000 (default)
+// hibernate.search.backend.request_timeout = (not defined by default)
```
yrodiere, line 204:
```suggestion
# hibernate.search.backend.request_timeout = (not defined by default)
```

Comments in `*.properties` files start with a hash character (`#`).

yrodiere (commented):
Thanks! Just a few details to iron out and then I think we can merge.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/ElasticsearchClientImpl.java
```diff
@@ -136,9 +136,10 @@ public void onFailure(Exception exception) {
...
 				);
 
-		long currentTimeoutValue = ( elasticsearchRequest.timeoutManager() == null ) ?
-				requestTimeout : elasticsearchRequest.timeoutManager().remainingTimeToHardTimeout();
-		TimeUnit currentTimeoutUnit = ( elasticsearchRequest.timeoutManager() == null ) ?
+		RequestDeadline requestDeadline = elasticsearchRequest.deadline();
+		long currentTimeoutValue = ( requestDeadline == null ) ?
+				requestTimeout : requestDeadline.remainingTimeToHardTimeout();
+		TimeUnit currentTimeoutUnit = ( requestDeadline == null ) ?
 				timeoutUnit : TimeUnit.MILLISECONDS;
```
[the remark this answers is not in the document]
fax4ever: You're right we can simplify here the collaborations

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/cfg/ElasticsearchIndexSettings.java
fax4ever: the problem with that is we're using the same `SCHEMA_MANAGEMENT_MINIMAL_REQUIRED_STATUS_WAIT_TIMEOUT` property for the request (socket) timeout of health check calls.
In the master I think we have 60 seconds for this very value, so I reused here this value.

fax4ever:
Here is another [full build-n.6](https://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-2345/6/)

yrodiere (approved):
Looks good to me! Feel free to merge once the build succeeds.
Thanks!

fax4ever:
Merged thanks!

PR #2346, opened 2020-09-02
HSEARCH-3613 Allow field access to property with the JavaBean mapper

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3613

This was necessary in order to ~implement~ (EDIT: actually, to **test**) the partial compatibility layer from Search 5 APIs to Search 6 APIs (migration helper).

I didn't rename the artifact yet; we'll do that in HSEARCH-1800.


fax4ever:
Nice to have this feature. Merged thanks

PR #2347, opened 2020-09-03
HSEARCH-3993 Add default identifier bridges for types character, boolean, byte, float, double

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3993


fax4ever:
Looks perfect! Merged thanks

PR #2349, opened 2020-09-04
HSEARCH-3864 java.lang.Error thrown during mass indexing are swallowed and not reported to the user

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3864

There is some code duplication.
I tried also to make [Make *MassIndexingFailureIT parametric] (https://github.com/fax4ever/hibernate-search/commit/0988a4efa58a8b5164c845f552cf0b0c5a683a86) as alternative...

### DIFF
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/AbstractMassIndexingErrorIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/AbstractMassIndexingErrorIT.java
@@ -27,8 +27,6 @@ import org.hibernate.search.mapper.orm.massindexing.MassIndexer;
 import org.hibernate.search.mapper.orm.massindexing.MassIndexingFailureHandler;
 import org.hibernate.search.mapper.pojo.mapping.definition.annotation.GenericField;
 import org.hibernate.search.mapper.pojo.mapping.definition.annotation.Indexed;
-import org.hibernate.search.util.common.SearchException;
-import org.hibernate.search.util.impl.integrationtest.common.FailureReportUtils;
 import org.hibernate.search.util.impl.integrationtest.common.rule.BackendMock;
 import org.hibernate.search.util.impl.integrationtest.common.rule.ThreadSpy;
 import org.hibernate.search.util.impl.integrationtest.common.stub.backend.index.StubIndexScaleWork;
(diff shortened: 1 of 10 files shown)


integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingFailureDefaultBackgroundErrorHandlerIT.java
```diff
@@ -0,0 +1,131 @@
...
+import org.hibernate.search.mapper.orm.massindexing.MassIndexingFailureHandler;
+import org.hibernate.search.util.common.SearchException;
+import org.hibernate.search.util.impl.test.ExceptionMatcherBuilder;
+import org.hibernate.search.util.impl.test.rule.ExpectedLog4jLog;
+
+import org.junit.Rule;
+
+import org.apache.log4j.Level;
+
+public class MassIndexingFailureDefaultBackgroundErrorHandlerIT extends AbstractMassIndexingErrorIT {
```
yrodiere, line 17: The name of this class looks wrong. It tests behavior when *errors* occur during mass indexing, with the default background *failure* handler. So it should be `MassIndexingErrorDefaultBackgroundFailureHandlerIT`.

yrodiere (changes requested):
Thanks.

There seems to be some problems in the tests. Please fix that first?

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/AbstractMassIndexingErrorIT.java
```diff
@@ -650,7 +610,7 @@ private SessionFactory setup() {
 	private enum ExecutionExpectation {
 		SUCCEED,
 		FAIL,
-		SKIP;
+		ERROR;
```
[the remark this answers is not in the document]
fax4ever: This is the current behavior:

|                               | AbstractMassIndexingErrorIT | AbstractMassIndexingFailureIT |
|-------------------------------|-----------------------------|-------------------------------|
| indexingAndFlush              | FAIL                        | FAIL                          |
| indexingAndRefresh            | FAIL                        | FAIL                          |
| flush                         | FAIL                        | FAIL                          |
| getId                         | ERROR                       | SKIP                          |
| purge                         | FAIL                        | FAIL                          |
| indexing                      | FAIL                        | FAIL                          |
| refresh                       | FAIL                        | FAIL                          |
| dropAndCreateSchema_exception | FAIL                        | FAIL                          |
| mergeSegmentsAfter            | FAIL                        | FAIL                          |
| mergeSegmentsBefore           | FAIL                        | FAIL                          |
| getTitle                      | ERROR                       | SKIP                          |

`AbstractMassIndexingErrorIT` behaves as `AbstractMassIndexingFailureIT` for all the cases with the exception of `getId` and `getTitle`.
That sounds reasonable to me. If it doesn't sound good for you, what would be the expected behavior?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/MassIndexingFailureContext.java
```diff
@@ -80,6 +82,24 @@ public Object getFailingOperation() {
 		return failingOperation();
 	}
 
+	@Override
+	public boolean equals(Object o) {
```
yrodiere, line 86: Uh no, I don't think it's a good idea to add this just for tests... I don't quite see a use case. Plus you forgot to override equals/hashcode in subclasses.

No, really, let's not do this. If you want to check that `failingOperation` and `throwable` are the same in your test, just do that explicitly? `assertThat( context1.failingOperation() ).isEqualTo( ...)`?

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingErrorDefaultBackgroundFailureHandlerIT.java
```diff
@@ -52,25 +51,19 @@ protected void assertEntityIndexingFailureHandling(String entityName, String ent
...
+				ExceptionMatcherBuilder.isException( SimulatedError.class )
 						.withMessage( exceptionMessage )
 						.build(),
-				failingOperationAsString,
-				"Entities that could not be indexed correctly:",
-				entityReferenceAsString
+				failingOperationAsString
 		)
-				.once();
+				.times( 2 );
```
yrodiere: I wonder why the error is reported twice? There may be a problem here. I'll investigate.

yrodiere (changes requested):
See my comments below.

1. The equals/hashcode are probably not a great idea.
2. The duplicated report of the error looked weird. I pushed a few commits to simplify reporting of errors: they are just propagated as-is. Please let me know what you think.

yrodiere:
I added a few commits to simplify how Errors are reported: we will just propagate them and won't attempt to report them to the failure handlers.
After some discussion on Zulip, we'll go that way.
Let's merge once the build passes.

yrodiere (approved):

yrodiere:
Merged, thanks!

fax4ever:
Thank you!

PR #2351, opened 2020-09-07
HSEARCH-3122 Restore support for configuring "strict" mapping in Elasticsearch

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3122

~~Draft:~~
~~1. We need more tests: invalid name && propagation on nested object mapping~~
~~2. Documentation~~

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/dsl/impl/AbstractElasticsearchIndexSchemaObjectNodeBuilder.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/dsl/impl/AbstractElasticsearchIndexSchemaObjectNodeBuilder.java
@@ -111,14 +111,8 @@ public abstract class AbstractElasticsearchIndexSchemaObjectNodeBuilder implemen
 
 	abstract String getAbsolutePath();
 
-	final DynamicType resolveSelfDynamicType() {
-		if ( templates.isEmpty() ) {
-			// TODO HSEARCH-3122 allow configuring this at index level (configuration properties)
-			return DynamicType.STRICT;
-		}
-		else {
-			return DynamicType.TRUE;
-		}
+	final DynamicType resolveSelfDynamicType(DynamicType defaultDynamicType) {
+		return ( templates.isEmpty() ) ? defaultDynamicType : DynamicType.TRUE;
 	}
 
 	private void putField(String name, ElasticsearchIndexSchemaNodeContributor contributor) {
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
@@ -652,4 +652,8 @@ public interface Log extends BasicLogger {
 	SearchTimeoutException clientSideTimedOut(@FormatWith(DurationInSecondsAndFractionsFormatter.class) Duration timeout,
 			String query);
 
+	@Message(id = ID_OFFSET_3 + 121, value = "Invalid dynamic type: '%1$s'."
+			+ " Valid values are: %2$s.")
+	SearchException invalidDynamicType(String invalidRepresentation, List<String> validRepresentations);
+
 }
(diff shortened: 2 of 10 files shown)


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -652,4 +652,8 @@ SearchException differentImplementationClassForQueryElement(@FormatWith(ClassFor
 	SearchTimeoutException clientSideTimedOut(@FormatWith(DurationInSecondsAndFractionsFormatter.class) Duration timeout,
 			String query);
 
+	@Message(id = ID_OFFSET_3 + 121, value = "Invalid dynamic type: '%1$s'."
+			+ " Valid statuses are: %2$s.")
```
yrodiere, line 656:
```suggestion
			+ " Valid values are: %2$s.")
```

documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
```diff
@@ -612,6 +612,33 @@ contains the actual index name (e.g. `myindex-000001`), not the alias (e.g. `myi
 If indexes do not follow the default naming scheme `<hibernateSearchIndexName>-<6 digits>`,
 a custom <<backend-elasticsearch-indexlayout,index layout>> must be configured.
 
+=== Dynamic types mapping
+
+By default, Hibernate Search set the Elasticsearch mapping dynamic properties to `strict`.
+See dynamic property mapping on
+https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic.html[Elasticsearch documentation].
+Which means that Elasticsearch server will throw an exception
+if we should try to work on fields that do not belong to the schema defined at bootstrap time.
```
yrodiere, line 621:
```suggestion
By default, Hibernate Search sets the link:{elasticsearchDocUrl}/dynamic.html[`dynamic`] property in Elasticsearch index mappings to `strict`.
This means that attempting to index documents with fields that are not present in the mapping will lead to an indexing failure.
```

documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
```diff
@@ -612,6 +612,33 @@ contains the actual index name (e.g. `myindex-000001`), not the alias (e.g. `myi
...
+hibernate.search.backend.dynamic_mapping = strict (default)
+# To configure a specific index:
+hibernate.search.backend.indexes.<index name>.dynamic_mapping = strict (default)
+----
+
+We said that Hibernate Search **usually** works on declared schema fields. More precisely, it always does
+if no <<mapper-orm-bridge-index-field-dsl-dynamic>> are defined.
+In the case of the root object or a nested object containing some dynamic fields,
+the mapping dynamic properties will be always set to `true` instead,
+regardless the value of dynamic_mapping property.
```
yrodiere, line 640:
```suggestion
When field templates are defined, `dynamic` will be forced to `true`, in order to allow for dynamic fields. In that case, the value of the `dynamic_mapping` property is ignored.
```

yrodiere (changes requested):
Thanks! Just a few comments.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/dsl/impl/AbstractElasticsearchIndexSchemaObjectNodeBuilder.java
```diff
@@ -12,7 +12,7 @@
 import java.util.Map;
 
 import org.hibernate.search.backend.elasticsearch.document.model.impl.AbstractElasticsearchIndexSchemaFieldNode;
-import org.hibernate.search.backend.elasticsearch.lowlevel.index.mapping.impl.DynamicType;
+import org.hibernate.search.backend.elasticsearch.index.DynamicType;
```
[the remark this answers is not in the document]
fax4ever: Yes move a `Gson` annotated object to an API wasn't a good idea. Fixed!

fax4ever:
We are changing an Elasticsearch parameter, so, even if maybe we shouldn't have issue with the older ES versions, here is a full build: https://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-2351/9/.

yrodiere:
Merged, thanks!

PR #2354, opened 2020-09-09
HSEARCH-4007 Throw an exception when the document ID turns out to be non-unique when loading entities during a search query

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4007


fax4ever:
Merged thanks!

PR #2357, opened 2020-09-10
HSEARCH-4011 Hibernate Search ScrollableResults should throw a JPA QueryTimeoutException instead of SearchTimeoutException

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4011


fax4ever:
Merged thanks!

PR #2364, opened 2020-09-16
HSEARCH-4015 Remove legacy dependecies from parent pom

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4015


yrodiere:
@fax4ever Did you try building with `-Dlegacy.skip=false`? If the legacy module relies on this dependency management, you cannot remove it. At most you can move the dependency management to `legacy/pom.xml`.

fax4ever:
I didn't know that we could compile, build and install the legacy module... There is any utility on doing that?

yrodiere:
Just to make sure the code we're copying (from time to time) still works. It runs on CI only.

yrodiere:
Merged, thanks

PR #2367, opened 2020-09-17
HSEARCH-3986 Test that Hibernate Search releases all file handles on close

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3986

### DIFF
--- a/integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/lifecycle/LuceneCleanupIT.java
+++ b/integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/lifecycle/LuceneCleanupIT.java
@@ -1,0 +1,160 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.integrationtest.backend.lucene.lifecycle;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.hibernate.search.util.impl.integrationtest.mapper.stub.StubMapperUtils.referenceProvider;
+
+import org.hibernate.search.backend.lucene.cfg.LuceneIndexSettings;
+import org.hibernate.search.backend.lucene.lowlevel.directory.spi.DirectoryProvider;
+import org.hibernate.search.engine.backend.document.model.dsl.IndexSchemaElement;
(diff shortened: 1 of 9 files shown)


util/internal/integrationtest/backend/lucene/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/lucene/directory/TrackingDirectoryProvider.java
```diff
@@ -0,0 +1,33 @@
...
+package org.hibernate.search.util.impl.integrationtest.backend.lucene.directory;
+
+import org.hibernate.search.backend.lucene.lowlevel.directory.spi.DirectoryCreationContext;
+import org.hibernate.search.backend.lucene.lowlevel.directory.spi.DirectoryHolder;
+import org.hibernate.search.backend.lucene.lowlevel.directory.spi.DirectoryProvider;
+
+import org.apache.lucene.store.SingleInstanceLockFactory;
+
+
+public class TrackingDirectoryProvider implements DirectoryProvider {
```
fax4ever, line 16: Actually the scope of this class is not clear to me. I was thinking if this class is supposed to be used by some user should be documented somehow. If not, maybe should be moved to the tests.

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/lifecycle/LuceneCleanupIT.java
```diff
@@ -0,0 +1,161 @@
...
+						(BeanReference<TrackingDirectoryProvider>) beanResolver -> {
+							BeanHolder<DirectoryProvider> delegateHolder = beanResolver.resolve( DirectoryProvider.class,
+									"local-filesystem" );
+							return BeanHolder.of( new TrackingDirectoryProvider( delegateHolder.get(), tracker ) )
+									.withDependencyAutoClosing( delegateHolder );
+						} )
+				.setup();
+	}
+
+
```
fax4ever, line 142: 2 spaces

fax4ever (commented):
just a question

util/internal/integrationtest/backend/lucene/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/lucene/directory/TrackingDirectoryProvider.java
yrodiere: It is in test code. More precisely, utils for integration tests:

```
package org.hibernate.search.util.impl.integrationtest.backend.lucene.directory
```

I'll add a short comment.

yrodiere:
Addressed the comments and merged. Thanks.

PR #2368, opened 2020-09-17
HSEARCH-4016 Avoid call to loader.loadBlocking(<empty list>) when there are no entities to load

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4016


fax4ever:
Merged thanks!

PR #2369, opened 2020-09-18
HSEARCH-3992 Add support for a default analyzer configured at the analysis configurer level

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3992


fax4ever:
Merged thanks!

PR #2370, opened 2020-09-18
HSEARCH-4017 Don't scan index leaves that don't contain topdocs when applying projections

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4017


fax4ever:
Merged thanks!

PR #2372, opened 2020-09-22
HSEARCH-3997 Mapping arrays of primitive types (int[]/float[]/etc.) leads to ClassCastException upon indexing

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3997


fax4ever:
Merged thanks!

PR #2373, opened 2020-09-22
HSEARCH-3998 Mass indexing a (non-abstract) parent class in a type hierarchy with a subclass annotated with @Indexed(enabled = false) still indexes that subclass

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3998

### DIFF
--- a/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/MassIndexingIndexedTypeGroup.java
+++ b/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/MassIndexingIndexedTypeGroup.java
@@ -1,0 +1,117 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.mapper.orm.massindexing.impl;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.ListIterator;
+import java.util.Set;
+import javax.persistence.metamodel.EntityType;
+import javax.persistence.metamodel.SingularAttribute;
+
+import org.hibernate.persister.entity.EntityPersister;
+import org.hibernate.search.mapper.orm.common.impl.HibernateOrmUtils;
+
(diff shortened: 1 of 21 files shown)


mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/MassIndexingIndexedTypeGroup.java
```diff
@@ -0,0 +1,131 @@
...
+	 * @return One or more type groups that are guaranteed to be disjoint.
+	 */
+	public static List<MassIndexingIndexedTypeGroup<?>> disjoint(
+			Set<? extends HibernateOrmMassIndexingIndexedTypeContext<?>> indexedTypeContexts) {
+		List<MassIndexingIndexedTypeGroup<?>> typeGroups = new ArrayList<>();
+		for ( HibernateOrmMassIndexingIndexedTypeContext<?> typeContext : indexedTypeContexts ) {
+			MassIndexingIndexedTypeGroup<?> typeGroup = MassIndexingIndexedTypeGroup.single( typeContext );
+			// First try to merge this new type group with an existing one
+			ListIterator<MassIndexingIndexedTypeGroup<?>> iterator = typeGroups.listIterator();
+			while ( iterator.hasNext() ) {
```
fax4ever, line 43: I don't know if we could use a `Union-Find` here to improve the complexity. See https://en.wikipedia.org/wiki/Disjoint-set_data_structure.

fax4ever (approved):
Very nice job! Just a comment.
+1 to merge

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/MassIndexingIndexedTypeGroup.java
yrodiere: I don't know either. This code is less than 20 lines long though, so I'd say let's merge it and get back to it if we ever come by a library that provides an on-the-shelf algorithm that does exactly what we want...

yrodiere: Oh, you meant time complexity. Yeah, I don't think it's that important here. It's only executed once before mass indexing, and worst case we'll have what? 200 entity types? Even if this takes one second of CPU, mass indexing will take much, much longer due to heavy I/O. Not worth spending time on, at least not until we actually see a problem in performance tests.

yrodiere:
Merged, thanks

PR #2374, opened 2020-09-23
HSEARCH-4020 Add the entity reference to the message of exceptions occurring during automatic indexing

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4020


yrodiere:
@fax4ever Actually this is not very important or dangerous, I'll merge as soon as CI passes, so you can take a break from reviewing PRs :)

PR #2376, opened 2020-09-23
HSEARCH-3999 Deletion of contained entity whose elementCollection is indexed-embedded in the containing entity leads to LazyInitializationException

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3999

I ended up disabling resolution of entities to reindex when a contained entity is deleted. It's a much smaller change than it sounds, since:

* In most cases, resolution of entities to reindex from a deleted entity will not work correctly (you'll get a `LazyInitializationException`).
* Even when it works, you'll need to update the other side of the relevant associations in order for the deleted entity to no longer appear in documents of index-embedding entities... so the index-embedding entities would have been reindexed anyway (because the association changed).

As a matter of fact, all our tests still pass even after this change. And we no longer have a `LazyInitializationException` when deleting contained entities.

### DIFF
--- a/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/work/impl/AbstractPojoTypeIndexingPlan.java
+++ b/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/work/impl/AbstractPojoTypeIndexingPlan.java
@@ -82,8 +82,8 @@ abstract class AbstractPojoTypeIndexingPlan<I, E, S extends AbstractPojoTypeInde
 		final I identifier;
 		Supplier<E> entitySupplier;
 
-		boolean delete;
-		boolean add;
+		boolean deleted;
+		boolean added;
 
 		boolean shouldResolveToReindex;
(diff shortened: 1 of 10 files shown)


mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/work/impl/AbstractPojoTypeIndexingPlan.java
```diff
@@ -6,24 +6,165 @@
...
+
+		private void addDirtyPath(String dirtyPath) {
+			if ( dirtyPaths == null ) {
+				dirtyPaths = new HashSet<>();
+			}
+			dirtyPaths.add( dirtyPath );
+		}
+	}
+
+
```
fax4ever, line 169: 2 new empty lines

fax4ever (approved):
Looks very good. +1 to merge

yrodiere:
Merged, thanks

PR #2379, opened 2020-09-28
HSEARCH-3985 Test Hibernate Search 6 indexing when hibernate.use_identifier_rollback=true

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3985

### DIFF
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/AutomaticIndexingIdentiferRollbackIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/AutomaticIndexingIdentiferRollbackIT.java
@@ -1,0 +1,165 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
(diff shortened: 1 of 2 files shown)


integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/AutomaticIndexingIdentiferRollbackIT.java
```diff
@@ -0,0 +1,168 @@
...
+
+			entity1Id.set( entity1.getId() );
+
+			backendMock.expectWorks( EntityWithJpaIdAsDocumentId.NAME )
+					.add( entity1.getId().toString(), b -> { } )
+					.processedThenExecuted();
+		} );
+		backendMock.verifyExpectationsMet();
+
+		AtomicReference<EntityWithJpaIdAsDocumentId> entity1FromTransaction = new AtomicReference<>();
```
fax4ever, line 73: it seems that this reference is set in the lambda but never get after that

fax4ever (commented):
Looks good, I've just a comment here...

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/AutomaticIndexingIdentiferRollbackIT.java
yrodiere: Right, this was a remnant of a former attempt at testing this. It's not necessary, I removed it.

yrodiere:
Merged, thanks!

PR #2384, opened 2020-10-02
Changing the string null check to improve the performance.

sann3:
isEmpty() gives **20-40%** more performance than "".equals().

String equals source code, 
```
public boolean equals(Object anObject) {
        if (this == anObject) {
            return true;
        }
        if (anObject instanceof String) {
            String aString = (String)anObject;
            if (coder() == aString.coder()) {
                return isLatin1() ? StringLatin1.equals(value, aString.value)
                                  : StringUTF16.equals(value, aString.value);
            }
        }
        return false;
    }
```

Operator instanceOf, class casting, 3 if statements, and all this is just to start the check if the string is actually empty. 

Now compare it with `String.isEmpty()`:
```
public boolean isEmpty() {
    return value.length == 0;
}
```

If we increase the performance in the framework, automatically multiple projects get the benefit.

#hacktoberfest

### DIFF
--- a/integrationtest/v5migrationhelper/orm/src/test/java/org/hibernate/search/test/embedded/doubleinsert/Address.java
+++ b/integrationtest/v5migrationhelper/orm/src/test/java/org/hibernate/search/test/embedded/doubleinsert/Address.java
@@ -109,7 +109,7 @@ public class Address implements Serializable {
 	}
 
 	public String getAddress2() {
-		if ( null == this.address2 || "".equals( this.address2 ) ) {
+		if ( null == this.address2 || this.address2.isEmpty() ) {
 			return "N/A";
 		}
 		return address2;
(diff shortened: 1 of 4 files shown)


yrodiere:
Hello,

Thank you for your contribution.

I don't know about improving performance (we'd need actual tests for that), but I agree that the code is better after your change.

However, this is test code. It is not consumed by any project. So these changes will not improve performance of Hibernate Search.

What's more, this is *legacy* test code, which is there to test backward compatibility with Hibernate Search 5 APIs. So I don't want to change anything in there unless we really need to. Eventually, this code will be removed.

If you want to contribute other changes to the project, feel free to pop out on [our chat](https://hibernate.zulipchat.com/#narrow/stream/132092-hibernate-search-dev) and we can discuss what would be interesting topics for a new contributor. Always happy to see new faces :)

PR #2397, opened 2020-10-08
HSEARCH-2836 Timeouts start when the request is submitted to the client, leading to systematic timeout when client is under heavy load

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-2836

As it turns out:

* The Elasticsearch client no longer enforces a request timeout.
* Our own request timeout is only enforced when the user asks for it explicitly, typically for search queries.

As a result, the problem mentioned in this ticket no longer exists.

I added a few tests to demonstrate that even when the REST client is clogged up with many pending requests, the new requests won't incur a read timeout (because the read timeout starts when the request is actually sent).

They will incur a request timeout, but that's working as intended, since the request timeout is opt-in.


fax4ever:
Merged thanks!

PR #2399, opened 2020-10-09
HSEARCH-4046 Expose a map of static children by name in the metamodel of object fields

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4046

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/backend/metamodel/IndexCompositeElementDescriptor.java
+++ b/engine/src/main/java/org/hibernate/search/engine/backend/metamodel/IndexCompositeElementDescriptor.java
@@ -7,6 +7,7 @@
 package org.hibernate.search.engine.backend.metamodel;
 
 import java.util.Collection;
+import java.util.Map;
 import java.util.function.Function;
 
 import org.hibernate.search.engine.backend.document.model.dsl.IndexSchemaElement;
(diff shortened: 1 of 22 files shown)


engine/src/main/java/org/hibernate/search/engine/backend/metamodel/IndexCompositeElementDescriptor.java
```diff
@@ -44,8 +45,19 @@
 	 * fields created dynamically through {@link IndexSchemaElement#fieldTemplate(String, Function) templates}
 	 * are not included in the collection.
 	 *
-	 * @return A collection containing all fields.
+	 * @return A collection containing all static child fields.
 	 */
 	Collection<? extends IndexFieldDescriptor> staticChildren();
 
+	/**
+	 * Get all statically-defined, direct child fields for this element, mapped by their relative name.
```
fax4ever, line 53: Is relative name `IndexFieldDescriptor#relativeName`? Is would be not critical, but if so, maybe we could add the link to the method in the Javadoc.

fax4ever (approved):
Nice job. +1 to merge it

engine/src/main/java/org/hibernate/search/engine/backend/metamodel/IndexCompositeElementDescriptor.java
yrodiere: Good idea. Done.

yrodiere:
Merged, thanks!

PR #2400, opened 2020-10-12
HSEARCH-4047 + HSEARCH-4048 Search DSL improvements and fixes for object fields

yrodiere:
* [HSEARCH-4048](https://hibernate.atlassian.net/browse/HSEARCH-4048): Relative precedence of object field templates vs. value field templates is not always taken into account
* [HSEARCH-4047](https://hibernate.atlassian.net/browse/HSEARCH-4047): More explicit error messages when attempting to use features specific to value fields on an object field

PR #2401, opened 2020-10-12
HSEARCH-3905 exists() predicate ignores dynamic fields among children of the targeted object field with the Lucene backend

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3905

Based on #2400 , which must be merged first.


fax4ever (approved):
It looks good!

yrodiere:
Merged, thanks

PR #2403, opened 2020-10-13
More explicit failure message when two types target the same index

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4006


yrodiere:
Perfect! Merged, thanks

PR #2408, opened 2020-10-19
HSEARCH-4023 Expose the total hit count in scroll results

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4023

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchScrollImpl.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchScrollImpl.java
@@ -93,7 +93,7 @@ public class LuceneSearchScrollImpl<H> implements LuceneSearchScroll<H> {
 
 		// no more results check
 		if ( scrollIndex >= search.hitSize() ) {
-			return new LuceneSearchScrollResultImpl<>( false, Collections.emptyList(),
+			return new LuceneSearchScrollResultImpl<>( search.total(), false, Collections.emptyList(),
 					timeoutManager.tookTime(), timeoutManager.isTimedOut() );
 		}
 
@@ -119,7 +119,8 @@ public class LuceneSearchScrollImpl<H> implements LuceneSearchScroll<H> {
 
 		// increasing the index for further next(s)
 		scrollIndex += chunkSize;
-		return new LuceneSearchScrollResultImpl<>( true, result.hits(), result.took(), result.timedOut() );
+		return new LuceneSearchScrollResultImpl<>( search.total(), true, result.hits(),
+				result.took(), result.timedOut() );
 	}
 
 	private <T> T doSubmitWithIndexReader(ReadWork<T> work, HibernateSearchMultiReader indexReader) {
--- a/engine/src/main/java/org/hibernate/search/engine/search/query/spi/SimpleSearchScrollResult.java
+++ b/engine/src/main/java/org/hibernate/search/engine/search/query/spi/SimpleSearchScrollResult.java
@@ -9,20 +9,29 @@ package org.hibernate.search.engine.search.query.spi;
 import java.time.Duration;
 import java.util.List;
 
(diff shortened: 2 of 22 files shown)


engine/src/main/java/org/hibernate/search/engine/search/query/spi/SimpleSearchScrollResult.java
```diff
@@ -9,22 +9,38 @@
...
+	public SimpleSearchScrollResult(SearchResultTotal resultTotal, boolean hasHits, List<H> hits,
+			Duration took, Boolean timedOut) {
+		this.resultTotal = resultTotal;
 		this.hasHits = hasHits;
 		this.hits = hits;
 		this.took = took;
 		this.timedOut = timedOut;
 	}
 
+	public SimpleSearchScrollResult(boolean hitExact, long hitCount, boolean hasHits, List<H> hits,
```
yrodiere, line 32: Looks like it would be better to have just one constructor here (the one above) and to add a method `of(boolean hitExact, long hitCount)` in `SimpleSearchResultTotal`?

At least you would avoid the code duplication from `SimpleSearchResult`.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchScrollImpl.java
```diff
@@ -93,7 +98,12 @@ public void close() {
 
 		// no more results check
 		if ( scrollIndex >= search.hitSize() ) {
-			return new LuceneSearchScrollResultImpl<>( false, Collections.emptyList(),
+			if ( cachedTotal == null ) {
+				cachedTotal = ( timeoutManager.isTimedOut() ) ? SimpleSearchResultTotal.lowerBound( 0L ) :
+						SimpleSearchResultTotal.exact( 0L );
```
yrodiere, line 103: That would be `search.hitSize()`, not `0`. Same above.

That being said... you could expose the `SearchResultTotal` in `LuceneExtractableSearchResult`, and use that. It's already there, accessible from `luceneCollectors.getResultTotal()`, and you know it's correct, so why not use it?

Going one step further, you could even rethink this method: if we're going beyond the end, in this block you would just set a boolean `hasHits = false`, and let this method continue to the last line, where you pass the boolean to the constructor of the search result... I think it may work as expected?

yrodiere (commented):
Thanks! Just a few comments. Feel free to merge once they've been addressed.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchScrollImpl.java
fax4ever: All good ides... I'm trying to applying them.

fax4ever: I applied the first two. About the the last one, the good fact of the current solution is that will be sure to avoid any `#loadBlocking` if we know in advance that we don't need it.

fax4ever:
@yrodiere Thanks for the review. I think I've applied all the changes, with the exception of the one described here: https://github.com/hibernate/hibernate-search/pull/2408#discussion_r507635439.
Let me know if you agree with merge it.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchScrollImpl.java
yrodiere: > I applied the first two

Thanks

> About the the last one, the good fact of the current solution is that will be sure to avoid any #loadBlocking if we know in advance that we don't need it.

`loadBlocking` won't do anything if there is nothing to load, so that's not a very good reason. Anyway, it's not that important, so let's forget about it.

yrodiere:
Merged, thanks

fax4ever:
Thanks!

PR #2409, opened 2020-10-19
HSEARCH-4053 + HSEARCH-4061 Remove legacy code and clean up the POM

yrodiere:
* [HSEARCH-4053](https://hibernate.atlassian.net/browse/HSEARCH-4053): Remove the "legacy" directory from the codebase
* [HSEARCH-4061](https://hibernate.atlassian.net/browse/HSEARCH-4061): Clean up a few obsolete entries in the POM


fax4ever:
Trying to force Spring to use an upgraded version of `junit-jupiter` was a nice try, I didn't think about that.
Anyway, it seems that we have to keep an insecure JUnit version for our Spring integration test.
Eventually we won't have it anymore, when we upgrade to JUnit 5 all the tests.
Merging... Thanks!

PR #2410, opened 2020-10-20
HSEARCH-3990 Optimize the retrieval of topDocs in very long Lucene scrolls

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3990

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchScrollImpl.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchScrollImpl.java
@@ -44,6 +44,7 @@ public class LuceneSearchScrollImpl<H> implements LuceneSearchScroll<H> {
 	private int scrollIndex = 0;
 	private int queryFetchSize;
 	private LuceneExtractableSearchResult<H> search;
+	private int offset = 0;
 
 	public LuceneSearchScrollImpl(LuceneSyncWorkOrchestrator queryOrchestrator,
 			LuceneWorkFactory workFactory, LuceneSearchContext searchContext,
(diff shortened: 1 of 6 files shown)


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/work/impl/ScrollWork.java
```diff
@@ -21,9 +21,12 @@
 
 	private final LuceneSearcher<?, ER> searcher;
 
+	// TODO HSEARCH-3990 Use the offset
```
yrodiere, line 24: You left this TODO. It doesn't seem necessary?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchScrollImpl.java
```diff
@@ -88,21 +89,24 @@ public void close() {
 			if ( search != null ) {
```
yrodiere, line 89: I believe the condition `scrollIndex + chunkSize > queryFetchSize` needs to be updated? It would be `( scrollIndex + chunkSize ) > ( offset + queryFetchSize )`.

Maybe we need an additional test, because something should have failed...

This is getting very confusing... I suggest the following renamings:

* `search` => `currentPage`
* `offset` => `currentPageOffset`
* `queryFetchSize` => `currentPageLimit`
* `index` => `currentChunkStartIndexInPage`
* `endIndexExclusive` => `currentChunkEndIndexInPage`
* `scrollIndex` => `nextChunkOffset`

The condition would become:

```
( nextChunkOffset + chunkSize ) > ( currentPageOffset + currentPageLimit )
```

yrodiere (changes requested):

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearchScrollImpl.java
fax4ever: you're right... going to fix it. Thanks

fax4ever: Done. About the additional test, I think that we don't need it. 
Because the external behavior would be the same and what we change with your fix is to avoid doing more scroll works than the ones we really need.

fax4ever:
Thanks @yrodiere, I've applied the changes.

yrodiere (approved):

fax4ever:
Merged thanks!

PR #2411, opened 2020-10-20
HSEARCH-2584 Introduce some commonly used Analyzer definitions by default

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-2584

### DIFF
--- a/documentation/src/main/asciidoc/reference/backend-lucene.asciidoc
+++ b/documentation/src/main/asciidoc/reference/backend-lucene.asciidoc
@@ -572,7 +572,7 @@ both when indexing (document processing)
 and when searching (query processing).
 
 The Lucene backend comes with some <<backend-lucene-analysis-builtin,default analyzers>>,
-but analysis also be configured explicitly.
+but analysis can also be configured explicitly.
 
 To configure analysis in a Lucene backend, you will need to:
 
--- a/documentation/src/main/asciidoc/reference/backend-lucene.asciidoc
+++ b/documentation/src/main/asciidoc/reference/backend-lucene.asciidoc
@@ -592,9 +592,25 @@ See below for examples.
 Built-in analyzers and normalizers are available out-of-the-box and don't require explicit configuration.
 If necessary, they can be overridden by defining your own analyzer/normalizer with the same name.
 
-The Lucene backend comes with a single built-in analyzer, named `default`.
-It is used by default with <<mapper-orm-directfieldmapping-annotations-fulltextfield,`@FullTextField`>>.
-Unless overridden explicitly, this analyzer uses a `standard` tokenizer and a `lowercase` filter.
+The Lucene backend comes with a series of built-in analyzer:
+
+[cols="l,1",options="header"]
+.Built-in analyzers provided out-of-the box by the Lucene backend
+|====
+|Analyzer name|Description
+|default|It is used by default with <<mapper-orm-directfieldmapping-annotations-fulltextfield,`@FullTextField`>>.
+Uses a `standard` tokenizer and a `lowercase` filter.
+|standard|Uses a `standard` tokenizer and a `lowercase` filter.
+|simple|Tokenize the text into tokens whenever it encounters a character which is not a letter,
+then lowercase each token.
+|whitespace|Tokenize the text into tokens whenever it encounters a character which is not a white space,
+without changing the tokens.
+|stop|Tokenize the text into tokens whenever it encounters a character which is not a letter,
+then it removes english stop words and finally lowercase each token.
+|keyword|Do not change in any way the text.
+With this analyzer a full text field would behave exactly as it was a keyword field.
+Maybe you should consider using a <<mapper-orm-directfieldmapping-annotations-keywordfield,`@KeywordField`>> instead.
+|====
 
 The Lucene backend does not provide any built-in normalizer.
 
--- a/documentation/src/main/asciidoc/reference/backend-lucene.asciidoc
+++ b/documentation/src/main/asciidoc/reference/backend-lucene.asciidoc
@@ -594,41 +594,45 @@ If necessary, they can be overridden by defining your own analyzer/normalizer wi
 
 The Lucene backend comes with a series of built-in analyzer:
 
-[cols="l,2,1",options="header"]
-.Built-in analyzers provided out-of-the-box by the Lucene backend
-|====
-|Analyzer name|Description|Is an instance of
-|default|
-It is used by default with <<mapper-orm-directfieldmapping-annotations-fulltextfield,`@FullTextField`>>.
-First, tokenize using the standard tokenizer, which follows Word Break rules from the
+`default`::
+The analyzer used by default with <<mapper-orm-directfieldmapping-annotations-fulltextfield,`@FullTextField`>>.
++
+Default implementation: `org.apache.lucene.analysis.standard.StandardAnalyzer`.
++
+Default behavior: first, tokenize using the standard tokenizer, which follows Word Break rules from the
 Unicode Text Segmentation algorithm, as specified in http://unicode.org/reports/tr29/[Unicode Standard Annex #29].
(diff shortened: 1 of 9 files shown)


engine/src/main/java/org/hibernate/search/engine/backend/analysis/AnalyzerNames.java
```diff
@@ -19,14 +19,59 @@ private AnalyzerNames() {
...
 	 *     <li>Then, lowercase each token.</li>
 	 * </ul>
 	 */
-	public static final String DEFAULT = "default";
+	public static final String SIMPLE = "simple";
+
+	/**
+	 * The whitespace analyzer.
+	 * <ul>
+	 *     <li>Tokenize the text into tokens whenever it encounters a character which is not a white space.</li>
```
yrodiere, line 51: `which is not a white space` => `which is a white space`?

engine/src/main/java/org/hibernate/search/engine/backend/analysis/AnalyzerNames.java
```diff
@@ -19,14 +19,59 @@ private AnalyzerNames() {
...
+	 *     <li>Finally, lowercase each token.</li>
+	 * </ul>
+	 */
+	public static final String STOP = "stop";
+
+	/**
+	 * The keyword analyzer.
+	 * <p>
+	 * Do not change in any way the text.
+	 * With this analyzer a full text field would behave exactly as it was a keyword field.
```
yrodiere, line 71: `With this analyzer a full text field would behave similarly to a keyword field, but with fewer features: no terms aggregations, for example.`

engine/src/main/java/org/hibernate/search/engine/backend/analysis/AnalyzerNames.java
```diff
@@ -19,14 +19,59 @@ private AnalyzerNames() {
 	 * <p>
 	 * This analyzer will generally be used for full-text field that don't require specific analysis.
 	 * <p>
-	 * Unless overridden by explicit analysis configuration, the default analyzer will be the standard analyzer:
+	 * Unless overridden by explicit analysis configuration,
+	 * the default analyzer will be the {@link #STANDARD} analyzer.
```
yrodiere, line 23: That's not really what you implemented, though. If a user overrides the analyzer named "standard", the analyzer named "default" will still be the `StandardAnalyzer`, not the user defined analyzer for "standard".

Maybe keep the previous documentation for the default analyzer, to be more correct?

documentation/src/main/asciidoc/reference/backend-lucene.asciidoc
```diff
@@ -592,9 +592,25 @@ See below for examples.
 Built-in analyzers and normalizers are available out-of-the box and don't require explicit configuration.
 If necessary, they can be overridden by defining your own analyzer/normalizer with the same name.
 
-The Lucene backend comes with a single built-in analyzer, named `default`.
-It is used by default with <<mapper-orm-directfieldmapping-annotations-fulltextfield,`@FullTextField`>>.
-Unless overridden explicitly, this analyzer uses a `standard` tokenizer and a `lowercase` filter.
+The Lucene backend comes with a series of built-in analyzer:
+
+[cols="l,1",options="header"]
+.Built-in analyzers provided out-of-the box by the Lucene backend
```
yrodiere, line 598: `out-of-the box` => `out-of-the-box`

documentation/src/main/asciidoc/reference/backend-lucene.asciidoc
```diff
@@ -592,9 +592,25 @@ See below for examples.
...
+|====
+|Analyzer name|Description
+|default|It is used by default with <<mapper-orm-directfieldmapping-annotations-fulltextfield,`@FullTextField`>>.
+Uses a `standard` tokenizer and a `lowercase` filter.
```
yrodiere, line 602: The previous documentation was wrong, I copy-pasted from the Elasticsearch documentation without taking care.

I think it would be better if you copied the definition from the `AnalyzerNames` javadoc, like you did for the other built-in analyzers?
Maybe also append:

```
This analyzer is an instance of `org.apache.lucene.analysis.standard.StandardAnalyzer`.
```

documentation/src/main/asciidoc/reference/backend-lucene.asciidoc
```diff
@@ -592,9 +592,25 @@ See below for examples.
...
+|standard|Uses a `standard` tokenizer and a `lowercase` filter.
```
yrodiere, line 603: Same here.

documentation/src/main/asciidoc/reference/backend-lucene.asciidoc
```diff
@@ -592,9 +592,25 @@ See below for examples.
...
+|simple|Tokenize the text into tokens whenever it encounters a character which is not a letter,
```
yrodiere, line 604: Same here and for the other analyzers below: it would be a good idea to mention the actual Lucene Analyzer class.

documentation/src/main/asciidoc/reference/backend-lucene.asciidoc
```diff
@@ -592,9 +592,25 @@ See below for examples.
...
+then lowercase each token.
+|whitespace|Tokenize the text into tokens whenever it encounters a character which is not a white space,
+without changing the tokens.
+|stop|Tokenize the text into tokens whenever it encounters a character which is not a letter,
+then it removes english stop words and finally lowercase each token.
+|keyword|Do not change in any way the text.
+With this analyzer a full text field would behave exactly as it was a keyword field.
+Maybe you should consider using a <<mapper-orm-directfieldmapping-annotations-keywordfield,`@KeywordField`>> instead.
```
yrodiere, line 612: To be updated to match the javadoc (I suggested a change there).

yrodiere (commented):
Thanks! Looks good overall, here are a few suggestions to clarify the documentation.

gsmet:
I might have missed something but I see relatively low value to this if you don't introduce the ascii folding filter.

It shouldn't be in the way for English and makes a huge difference for other languages.

Agreed it might not exactly be what you want in all cases but, for a default, I think we should have it.

yrodiere:
For non-english languages, there are per-language analyzers. But we decided against including them because there are so many, and we don't want to:

1. Be responsible for maintaining that list.
2. Force user to load 100 analyzers in their application when they will only use one or two.

Now, about including an analyzer with an ascii-folding filter... The thing is, there is no such analyzer in Elasticsearch.
I'm definitely not going to pollute ES indexes with Hibernate Search-specific analyzer definitions, so we could introduce one for Lucene, but we'd need a separate set of constants for Lucene-only analyzers. That's... questionable.

I'm okay with these definitions being only useful for demo applications, because I firmly believe that there's no silver bullet and that every non-trivial application needs to define its own analyzers matching its own requirements.

engine/src/main/java/org/hibernate/search/engine/backend/analysis/AnalyzerNames.java
fax4ever: You're right, if the user overrode the `standard` analyzer, this sentence would no longer be true!
I'm restoring the old documentation...

documentation/src/main/asciidoc/reference/backend-lucene.asciidoc
fax4ever: There were already 2 other `out-of-the box`  in the documentation. I will fix them in a different commit.

fax4ever:
I applied the changes.
I also introduced a column for the analyzer classes... I don't know if it was a good idea.

yrodiere:
>  I also introduced a column for the analyzer classes... I don't know if it was a good idea.

It renders pretty badly on smaller screens, so I don't think it was. I added additional commits to use a description list instead (and fix some issues I just noticed). I'll merge. Thanks!

PR #2412, opened 2020-10-21
HSEARCH-4021 Move the tests to log4j2

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4021


util/internal/integrationtest/v5migrationhelper/src/main/java/org/hibernate/search/test/util/impl/ExpectedLog4jLog.java
```diff
@@ -10,25 +10,24 @@
...
 import org.hamcrest.Matcher;
 import org.hamcrest.StringDescription;
 import org.hamcrest.TypeSafeMatcher;
-import org.junit.rules.TestRule;
-import org.junit.runners.model.Statement;
 
-/**
- * @author Yoann Rodiere
- */
 public class ExpectedLog4jLog implements TestRule {
```
yrodiere, line 31: I think you replaced the code with the other implementation, `org.hibernate.search.util.impl.test.rule.ExpectedLog4jLog`? If so, I'd say we may as well remove this class and replace it with the other implementation everywhere. I can do it if you want?

yrodiere (commented):
Looks good, just one comment. And also, well, the tests don't pass anymore :)

fax4ever:
Now it is really ready for the review... Sorry for the noise.

util/internal/test/src/main/java/org/hibernate/search/util/impl/test/rule/log4j/LogProgrammatic.java
```diff
@@ -0,0 +1,53 @@
...
+package org.hibernate.search.util.impl.test.rule.log4j;
+
+import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.core.Appender;
+import org.apache.logging.log4j.core.LoggerContext;
+import org.apache.logging.log4j.core.config.Configuration;
+import org.apache.logging.log4j.core.config.LoggerConfig;
+
+public class LogProgrammatic {
```
yrodiere, line 16: This is a really unusual class name... Maybe rename to something more "to the point", such as `Log4j2AppenderHelper` or `Log4j2ConfigurationAccessor`?

util/internal/integrationtest/v5migrationhelper/src/main/java/org/hibernate/search/test/util/impl/ExpectedLog4jLog.java
```diff
@@ -10,21 +10,21 @@
...
+import org.apache.logging.log4j.core.LogEvent;
 import org.hamcrest.CoreMatchers;
 import org.hamcrest.Description;
 import org.hamcrest.Matcher;
 import org.hamcrest.StringDescription;
 import org.hamcrest.TypeSafeMatcher;
-import org.junit.rules.TestRule;
-import org.junit.runners.model.Statement;
 
 /**
```
yrodiere, line 29: You didn't replace this version of `ExpectedLog4jLog` with the other one, as we discussed during the previous review?

util/internal/integrationtest/v5migrationhelper/src/main/resources/log4j2.properties
```diff
@@ -0,0 +1,40 @@
+### direct log messages to stdout ###
```
yrodiere, line 1: You can safely remove this file. This module depends on the main test utils module, which already provides Log4j configuration, and while different, that configuration is enough for our needs here.

integrationtest/jdk/java-modules/src/test/resources/log4j2.properties
```diff
@@ -0,0 +1,35 @@
...
+appender.file.filename=hibernate.log
+appender.file.layout.type=PatternLayout
+appender.file.layout.pattern=%d{ABSOLUTE} (%t) %5p %c{1}:%L - %m%n
+
+### direct messages to socket - chainsaw ###
+appender.socket.type=Socket
+appender.socket.name=SOCKET_APPENDER
+appender.socket.host=localhost
+appender.socket.port=4560
+appender.socket.locationInfo=true
```
yrodiere, line 19: Feel free to remove this appender: we don't need it.

integrationtest/jdk/java-modules/src/test/resources/log4j2.properties
```diff
@@ -0,0 +1,35 @@
...
+appender.stdout.name=STDOUT
+appender.stdout.layout.type=PatternLayout
+appender.stdout.layout.pattern=%d{ABSOLUTE} (%t) %5p %c{1}:%L - %m%n
+
+### direct messages to file hibernate.log ###
+appender.file.type=File
+appender.file.name=LOGFILE
+appender.file.filename=hibernate.log
+appender.file.layout.type=PatternLayout
+appender.file.layout.pattern=%d{ABSOLUTE} (%t) %5p %c{1}:%L - %m%n
```
yrodiere, line 12: Feel free to remove this appender: we don't need it.

yrodiere (commented):
Looks good overall. Just a few comments; feel free to merge once they've been addressed. Thanks!

util/internal/integrationtest/v5migrationhelper/src/main/java/org/hibernate/search/test/util/impl/ExpectedLog4jLog.java
fax4ever: I've thought about it. I would prefer to change the `v5 migration helper` modules as little as possible. If there was something to make sharing, I would do it as a different contribution/issue.

util/internal/integrationtest/v5migrationhelper/src/main/resources/log4j2.properties
fax4ever: Since there was a `log4j.properties` here, I'll do in a different commit.

fax4ever:
Thanks for the review. If you agree, I would merged it as soon as the CI has passed.

util/internal/integrationtest/v5migrationhelper/src/main/resources/log4j2.properties
yrodiere: My point was, you changed it already, and what's more, you changed it to match the main code...

fax4ever:
Merged, thanks!

PR #2414, opened 2020-10-22
HSEARCH-4070 Lucene scroll ignores totalHitCountThreshold

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4070

PR #2415, opened 2020-10-22
HSEARCH-4068 Skip non-competitive hits when running a MatchAllDocsQuery with descending score sort with Lucene

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4068

This PR is based on #2414 , which should be merged first.

After this patch, when a user runs a `MatchAllDocsQuery` (and there are no nested documents), we automatically set the `totalHitCountThreshold` to 0 and compute the total hit count in constant time.

What's interesting about this? Well, if the user requested a sort by descending score, we are now able to skip non-competitive hits. Which means for example that if the index contains 500,000 documents and the user requested the first 10, we will go through 10 hits instead of 500,000.

Judging from performance tests in Infinispan, the impact on performance is not huge, but still not negligible.

Search 5.11:

```
[Done in 60.010000s] 1 node(s), 10 thread(s) per node: Query 90th: 3.850239, QPS: 743.742710
```

Search 6.0 before this patch:

```
[Done in 60.012000s] 1 node(s), 10 thread(s) per node: Query 90th: 3.473407, QPS: 775.578218
```

Search 6.0 after this patch:

```
[Done in 60.011000s] 1 node(s), 10 thread(s) per node: Query 90th: 1.261567, QPS: 899.818367
```

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/extraction/impl/ExtractionRequirements.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/extraction/impl/ExtractionRequirements.java
@@ -19,6 +19,7 @@ import org.hibernate.search.backend.lucene.search.timeout.impl.LuceneTimeoutMana
 
 import org.apache.lucene.search.Collector;
 import org.apache.lucene.search.IndexSearcher;
+import org.apache.lucene.search.MatchAllDocsQuery;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.Sort;
 import org.apache.lucene.search.SortField;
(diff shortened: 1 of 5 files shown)


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/extraction/impl/ExtractionRequirements.java
```diff
@@ -43,16 +44,31 @@ private ExtractionRequirements(Builder builder) {
...
 
-	public LuceneCollectors createCollectors(IndexSearcher indexSearcher, Query luceneQuery, Sort sort,
+	public LuceneCollectors createCollectors(IndexSearcher indexSearcher, Query originalLuceneQuery, Sort sort,
 			IndexReaderMetadataResolver metadataResolver, int maxDocs, LuceneTimeoutManager timeoutManager,
-			int totalHitCountThreshold)
+			int requestedTotalHitCountThreshold)
 			throws IOException {
+		// Necessary to unwrap boolean queries with a single clause, in particular:
+		// we have optimizations in place when there is a single query and this query is a MatchAllDocsQuery.
+		Query rewrittenLuceneQuery = indexSearcher.rewrite( originalLuceneQuery );
```
fax4ever, line 53: Instead of rewrite the query here, wouldn't be better to this job when we create the `definitiveLuceneQuery` `LuceneSearchQueryBuilder#build`?

fax4ever (commented):
Nice job. If was interesting to know if ES does the same...
Anyway, just a doubt.
But as usual, feel free to ignore it.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/extraction/impl/ExtractionRequirements.java
yrodiere: We can't, because we don't have the `indexSearcher` at that point. Rewriting a query requires access to the index searcher (and its underlying index reader).

yrodiere:
Merged, thanks!

PR #2416, opened 2020-10-22
HSEARCH-4056 Expose the index size from LuceneIndexManager

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4056

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/schema/management/impl/LuceneIndexSchemaManager.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/schema/management/impl/LuceneIndexSchemaManager.java
@@ -8,6 +8,7 @@ package org.hibernate.search.backend.lucene.schema.management.impl;
 
 import java.util.Collection;
 import java.util.concurrent.CompletableFuture;
+import java.util.function.BinaryOperator;
 
 import org.hibernate.search.backend.lucene.orchestration.impl.LuceneParallelWorkOrchestrator;
 import org.hibernate.search.backend.lucene.work.impl.IndexManagementWork;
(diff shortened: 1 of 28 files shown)


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/schema/management/impl/LuceneIndexSchemaManager.java
```diff
@@ -60,9 +61,21 @@ public LuceneIndexSchemaManager(LuceneWorkFactory luceneWorkFactory,
 		return doSubmit( luceneWorkFactory.validateIndexExists() );
 	}
 
+	public CompletableFuture<Long> computeSizeInBytes() {
+		IndexManagementWork<Long> computeSizeWork = luceneWorkFactory.computeSizeInBytes();
+		BinaryOperator<Long> add = (left, right) -> left + right;
+
+		CompletableFuture<Long> totalSizeFuture = CompletableFuture.completedFuture( 0L );
+		for ( LuceneParallelWorkOrchestrator orchestrator : indexManagerContext.allManagementOrchestrators() ) {
```
fax4ever, line 69: If I don't get wrong, using the management orchestrators means that we're going to use more threads (one per shard) to compute the index size? 
But I think that I didn't get well that. :P
In case I did, I think we shouldn't split a constant-time-per-item computation among more threads.

fax4ever (commented):
Nice job. Very clean.
I've just a doubt... but probably I didn't get things right :P

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/schema/management/impl/LuceneIndexSchemaManager.java
yrodiere: I agree it's an unnecessary optimization, given that computing index size will probably be relatively fast. It just happens to also be the cleanest way to implement this, so I thought I might as well do it.

To answer your questions:

1. Yes, we will parallelize this computation if there are more than one shard. Though in most cases there will only be one shard, so it will not matter.
2. Computing the size of a shard is not constant-time, we iterate over every file in the index and compute its size. So it's closer to `O(n)` where `n` is the number of files.
3. Computing the size of a shard is a blocking operation, so executing the operation of each shard in parallel *might* be faster (we would "wait" for the I/O to finish in parallel for all shards).

yrodiere:
Merged, thanks

PR #2420, opened 2020-10-28
HSEARCH-4077 + HSEARCH-4078 Avoid extra logging when using ExpectedLog4jLog

yrodiere:
* [HSEARCH-4078](https://hibernate.atlassian.net/browse/HSEARCH-4078): Fix transient failure in LibraryShowcaseMassIndexingIT
* [HSEARCH-4077](https://hibernate.atlassian.net/browse/HSEARCH-4077): Avoid extra logging when using ExpectedLog4jLog

This should fix the build for the master branch, in particular.

### DIFF
--- a/util/internal/test/src/main/java/org/hibernate/search/util/impl/test/rule/log4j/LogChecker.java
+++ b/util/internal/test/src/main/java/org/hibernate/search/util/impl/test/rule/log4j/LogChecker.java
@@ -53,7 +53,7 @@ public class LogChecker {
 				if ( extraEvents == null ) {
 					extraEvents = new ArrayList<>();
 				}
-				extraEvents.add( event );
+				extraEvents.add( event.toImmutable() );
 			}
 		}
 	}
--- a/util/internal/test/src/main/java/org/hibernate/search/util/impl/test/rule/log4j/LogChecker.java
+++ b/util/internal/test/src/main/java/org/hibernate/search/util/impl/test/rule/log4j/LogChecker.java
@@ -43,7 +43,8 @@ public class LogChecker {
 		}
 	}
 
-	void process(LogEvent event) {
+	// This must be synchronized to avoid problems when multiple threads issue log events concurrently
+	synchronized void process(LogEvent event) {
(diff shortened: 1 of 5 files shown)


yrodiere:
The failure is unrelated. Putting on hold until #2423 get merged.

yrodiere:
I merged the PR this was depending on, and addressed Sonar/LGTM issues. Ready to review!

util/internal/test/src/main/java/org/hibernate/search/util/impl/test/rule/log4j/LogChecker.java
```diff
@@ -43,7 +43,8 @@ public void appendFailure(Description description, String newline) {
 		}
 	}
 
-	void process(LogEvent event) {
+	// This must be synchronized to avoid problems when multiple threads issue log events concurrently
+	synchronized void process(LogEvent event) {
```
fax4ever, line 47: It seems that this method only appends data to `extraEvents`, there is no remove.
Moreover, possible interleaving of `add` on `extraEvents` and count increment would not affect the consistency, or, at least, I think so.
I was asking myself if we could remove the `synchronized` here, with some changes:
 
1. Create the event list in the constructor
2. Use an atomic integer for the counter
3. Use a concurrent friendly data structure for the event lists, such as `ConcurrentLinkedQueue`. 
 
I know that this is test code (not production code), but maybe it is worth trying to make them more efficient.

fax4ever (commented):
Trying to reduce the scope of the test appender is good idea, maybe we can filter by log severity as well.
I've just one comment.
Anyway, as usual, +1 to merge it as soon as we get a working CI.

util/internal/test/src/main/java/org/hibernate/search/util/impl/test/rule/log4j/LogChecker.java
yrodiere: I don't think this is worth optimizing, especially considering the additional complexity the optimization would introduce.

yrodiere:
Merged, thanks!

PR #2422, opened 2020-10-28
HSEARCH-4079 + HSEARCH-4080 Query timeout fixes

yrodiere:
* [HSEARCH-4080](https://hibernate.atlassian.net/browse/HSEARCH-4080): Lucene total hit count is wrongly considered exact on timeout
* [HSEARCH-4079](https://hibernate.atlassian.net/browse/HSEARCH-4079): Search query always times out when timeout is lower than 1ms

~Based on #2421 , which should be merged first.~ => Done


fax4ever (approved):
It looks good!
We have just to change the author of the commit https://github.com/hibernate/hibernate-search/pull/2422/commits/75707376d4c6ff0e4bc9b56102cc3b0314c84ba4, that it is strangely marked as if I authored it.

yrodiere:
Merged, thanks

PR #2427, opened 2020-10-29
HSEARCH-4064 + HSEARCH-4076 Document limitations and troubleshooting

yrodiere:
* [HSEARCH-4076](https://hibernate.atlassian.net/browse/HSEARCH-4076): Document loggers
* [HSEARCH-4064](https://hibernate.atlassian.net/browse/HSEARCH-4064): Add warnings in the documentation about concurrent updates to the same entity in a clustered application

### DIFF
--- a/documentation/src/main/asciidoc/reference/limitations.asciidoc
+++ b/documentation/src/main/asciidoc/reference/limitations.asciidoc
@@ -1,0 +1,79 @@
+[[limitations]]
+= Limitations
+// Search 5 anchors backward compatibility
+[[elasticsearch-limitations]]
+
+[[limitations-parallel-embedded-update]]
+== In rare cases, automatic indexing involving `@IndexedEmbedded` may lead to out-of sync indexes
+
+=== Description
+
+If two entity instances are <<mapper-orm-indexedembedded,indexed-embedded>> in the same "index-embedding" entity,
+and these two entity instance are updated in parallel transactions,
+there is a small risk that the transaction commits happen in just the wrong way,
+leading to the index-embedding entity being reindexed with only part of the updates.
+
+For example, consider indexed entity A, which index-embeds B and C.
+The following course of events involving two parallel transactions (T1 and T2)
+will lead to an out of date index:
+
+* T1: Load B.
+* T1: Change B in a way that will require reindexing A.
+* T2: Load C.
+* T2: Change C in a way that will require reindexing A.
+* T2: Request the transaction commit.
+  Hibernate Search builds the document for A.
+  While doing so, it automatically loads B. B appears unmodified, as T1 wasn't committed yet.
+* T1: Request the transaction commit.
+  Hibernate Search builds documents to index.
+  While doing so, it automatically loads C. C appears unmodified, as T2 wasn't committed yet.
+* T1: Transaction is committed.
+  Hibernate Search automatically sends the updated A to the index.
+  In this version, B is updated, but C is not.
+* T2: Transaction is committed.
+  Hibernate Search automatically sends the updated A to the index.
+  In this version, C is updated, but B is not.
+
+This chain of events ends with the index containing a version of A where C is updated, but B is not.
+
+=== Workaround
+
+The following solutions can help circumvent this limitation:
+
+1. Avoid parallel updates to entities that are indexed-embedded in the same indexed entity.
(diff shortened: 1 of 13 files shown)


documentation/src/main/asciidoc/reference/limitations.asciidoc
```diff
@@ -0,0 +1,79 @@
...
+This chain of events ends with the index containing a version of A where C is updated, but B is not.
+
+=== Workaround
+
+The following solutions can help circumvent this limitation:
+
+1. Avoid parallel updates to entities that are indexed-embedded in the same indexed entity.
+This is only possible in very specific setups.
+2. OR schedule a <<mapper-orm-indexing-massindexer,full reindexing>> of your database periodically (e.g. every night)
+to get the index back in sync with the database.
```
fax4ever, line 46: I would like to add a third option here (if it makes sense):
Remove a re-indexing trigger. E.g.: from C -> A.
But I'm aware that many to one and many to many are very common with a relational data store.

documentation/src/main/asciidoc/reference/troubleshooting.asciidoc
```diff
@@ -0,0 +1,103 @@
...
+Alternatively, rely on the logs:
+<<troubleshooting-logging-query,`org.hibernate.search.query`>> will log every query at the `DEBUG` level
+before it is executed.
+
+For more general information about what is being executed, rely on loggers:
+
+* <<troubleshooting-logging-lucene-infostream,`org.hibernate.search.backend.lucene.infostream`>>
+for Lucene.
+* <<troubleshooting-logging-elasticsearch-request,`org.hibernate.search.elasticsearch.request`>>
+for Elasticsearch.
```
fax4ever, line 18: I don't know if it makes sense to have this wrap up chapter if we have the same information just above.

documentation/src/main/asciidoc/reference/troubleshooting.asciidoc
```diff
@@ -0,0 +1,103 @@
...
+Logs every single search query at the `DEBUG` level, before its execution.
+[[troubleshooting-logging-elasticsearch-request]]`org.hibernate.search.elasticsearch.request`::
+Available for Elasticsearch backends only.
++
+Logs every single request sent to Elasticsearch at the `DEBUG` or `TRACE` level after its execution.
+Use the `DEBUG` level for compact information (method, path, execution time, status code, ...),
+or the `TRACE` level to also get the full request and response.
++
+You can enable pretty-printing (line breaks and indentation) for the request and response
+using a <<backend-elasticsearch-configuration-logging,configuration property>>.
```
fax4ever, line 37: /using a/using the/s ?

documentation/src/main/asciidoc/reference/troubleshooting.asciidoc
```diff
@@ -0,0 +1,103 @@
...
+[[troubleshooting-logging-lucene-infostream]]`org.hibernate.search.backend.lucene.infostream`::
+Available for Lucene backends only.
++
+Logs low level trace information about Lucene's internal components, at the `TRACE` level.
++
+Enabling the `TRACE` level on this logger is not enough:
+you must also enable the infostream using a <<backend-lucene-io-writer-infostream,configuration property>>.
```
fax4ever, line 44: /using a/using the/s ?

fax4ever (approved):
Nice job! Just few comments

documentation/src/main/asciidoc/reference/troubleshooting.asciidoc
yrodiere: No, using a configuration property. There are many.

documentation/src/main/asciidoc/reference/troubleshooting.asciidoc
yrodiere: No, using a configuration property. There are many.

documentation/src/main/asciidoc/reference/limitations.asciidoc
yrodiere: While it would eliminate the problem, this solution would also lead to out-of-date indexes, which is what we want to avoid here. So I don't think it's worth mentioning.

yrodiere:
Merged, thanks!

PR #2435, opened 2020-11-05
HSEARCH-4094 JSR-352 documentation should mention that an additional dependency is necessary

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4094

### DIFF
--- a/documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
+++ b/documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
@@ -179,7 +179,7 @@ You will need to add this dependency:
 [source, XML, subs="+attributes"]
 ----
 <dependency>
-   <groupId>org.hibernate</groupId>
+   <groupId>org.hibernate.search</groupId>
    <artifactId>hibernate-search-backend-elasticsearch-aws</artifactId>
    <version>{hibernateSearchVersion}</version>
 </dependency>
--- a/documentation/src/main/asciidoc/reference/mapper-orm-indexing-jsr352.asciidoc
+++ b/documentation/src/main/asciidoc/reference/mapper-orm-indexing-jsr352.asciidoc
@@ -16,6 +16,30 @@ implementation (see <<jsr-352-emf-jberet,how to configure it here>>).
 As for other implementations, they can also be used, but will require
 <<jsr-352-emf-other-implementation,a bit more configuration on your side>>.
 
+If the runtime is JBeret, you need to add the following dependency:
+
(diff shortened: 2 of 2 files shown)


documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
```diff
@@ -179,7 +179,7 @@ You will need to add this dependency:
 [source, XML, subs="+attributes"]
 ----
 <dependency>
-   <groupId>org.hibernate</groupId>
+   <groupId>org.hibernate.search</groupId>
```
yrodiere, line 182: Nice catch.

documentation/src/main/asciidoc/reference/mapper-orm-indexing-jsr352.asciidoc
```diff
@@ -399,8 +423,7 @@ an entity manager factory by its persistence unit name when using the mass index
 [[jsr-352-emf-other-implementation]]
 === Other DI-enabled JSR-352 implementations
 
-If you want to use a different JSR-352 implementation that happens to allow dependency injection,
-you can use `hibernate-search-mapper-orm-batch-jsr352-core` under the following conditions:
+If you want to use a different JSR-352 implementation:
```
yrodiere, line 426:
```suggestion
If you want to use a different JSR-352 implementation that happens to allow dependency injection:
```

This is really only necessary if the batch runtime supports DI.

documentation/src/main/asciidoc/reference/mapper-orm-indexing-jsr352.asciidoc
```diff
@@ -30,6 +54,8 @@ include::{sourcedir}/org/hibernate/search/documentation/mapper/orm/indexing/Hibe
 
 == Job Parameters
 
+The following table will contain all the job parameters we can use to customize the reindexing batch processes.
```
yrodiere, line 57:
```suggestion
The following table contains all the job parameters you can use to customize the mass-indexing job.
```

Not sure this is really necessary, but as you wish.

yrodiere (changes requested):
Thanks! Only a few comments. Feel free to merge after addressing them.

documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
fax4ever: thank you

fax4ever:
Applied the suggestions and merged. Thanks!

PR #2436, opened 2020-11-05
HSEARCH-4069 Measure time more precisely when computing execution time of Lucene queries

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4069

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/search/timeout/spi/TimeoutManager.java
+++ b/engine/src/main/java/org/hibernate/search/engine/search/timeout/spi/TimeoutManager.java
@@ -73,20 +73,25 @@ public class TimeoutManager {
 		this.type = type;
 		this.deadline = timeoutMs == null ? null : new DynamicDeadline();
 
-		timingSource.ensureTimeEstimateIsInitialized();
+		if ( requireMonotonicTimeEstimate() ) {
+			timingSource.ensureTimeEstimateIsInitialized();
+		}
 	}
 
 	/**
 	 * we start counting from this method call (if needed)
 	 */
 	public void start() {
-		this.monotonicTimeEstimateStart = timingSource.monotonicTimeEstimate();
-		this.nanoTimeStart = timingSource.nanoTime();
+		if ( requireMonotonicTimeEstimate() ) {
+			monotonicTimeEstimateStart = timingSource.monotonicTimeEstimate();
+		}
+
+		nanoTimeStart = timingSource.nanoTime();
 	}
 
 	public void stop() {
-		this.monotonicTimeEstimateStart = null;
-		this.nanoTimeStart = null;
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/hibernate/search/engine/search/timeout/spi/TimeoutManager.java
```diff
@@ -138,12 +141,18 @@ public boolean hasHardTimeout() {
...
+	 * This method may be performance expensive.
+	 *
+	 * @return high precision duration of took time.
+	 */
 	public Duration tookTime() {
-		return Duration.ofMillis( elapsedTimeInMilliseconds() );
+		return Duration.ofMillis( timingSource.nanoTime() - nanoTimeStart );
 	}
 
 	protected long elapsedTimeInMilliseconds() {
```
yrodiere, line 154: Maybe the reason for having two methods with a similar purpose would be more obvious if you renamed this to `elapsedTimeEstimateMillis()`?

engine/src/main/java/org/hibernate/search/engine/search/timeout/spi/TimeoutManager.java
```diff
@@ -79,11 +80,13 @@ public TimeoutManager(TimingSource timingSource, Long timeoutValue, TimeUnit tim
 	 * we start counting from this method call (if needed)
 	 */
 	public void start() {
-		this.start = timingSource.monotonicTimeEstimate();
+		this.monotonicTimeEstimateStart = timingSource.monotonicTimeEstimate();
```
yrodiere, line 83: Actually, if no timeout was defined, this line becomes unnecessary.

That in itself wouldn't be worth discussing, but... if you only execute this line when there actually is a timeout, then you can do the same for `timingSource.ensureInitialized();`, in the constructor above. And *that* method can be costly, as it triggers the creation of a thread pool on its first execution.

yrodiere (commented):
Looks good! Just a few suggestions.

engine/src/main/java/org/hibernate/search/engine/search/timeout/spi/TimeoutManager.java
fax4ever: Good idea!

engine/src/main/java/org/hibernate/search/engine/search/timeout/spi/TimeoutManager.java
fax4ever: Right, we don't need it anymore to compute the took time!

fax4ever:
Thanks for the reviews and the ideas. I did some changes.

yrodiere:
Merged, thanks!

PR #2438, opened 2020-11-10
HSEARCH-3934 Custom index settings for Elasticsearch indexes created by Hibernate Search

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3934

Draft since we don't have a branch for the 6.1 yet.
So this contribution mustn't be merged with the current master as soon as we release the `6.0.0.Final`.


yrodiere (changes requested):
Thanks, and sorry for the late review.

While this will work, I think there's room for improvement:

* Custom settings are only taken into account when creating indexes, not when validating them. That's a shame, everything else gets created *and* validated.
* The fact that we erase all other settings as soon as there are custom settings is also a bit annoying. I can see people already having an analyzer configurer at the backend level, and wanting to add some settings to a specific index. They will have to put all their analyzer configuration into the settings file all over again...

To address this, I think the following changes are required:

* Add "extra attributes" to `org.hibernate.search.backend.elasticsearch.lowlevel.index.settings.impl.IndexSettings` and `org.hibernate.search.backend.elasticsearch.lowlevel.index.settings.impl.Analysis`. The point is to push all attributes that are not modeled in the class into a `Map<String, JsonElement>`, so that deserialization degrades gracefully. See how it's done in `org.hibernate.search.backend.elasticsearch.lowlevel.index.aliases.impl.IndexAliasDefinition` for example: `extraAttributes` is annotated with `@SerializableExtraProperties` and the class has its own JSON adapter (`@JsonAdapter(IndexAliasDefinitionJsonAdapterFactory.class)`).
* When you deserialize the settings file given by the user, deserialize directly into an `org.hibernate.search.backend.elasticsearch.lowlevel.index.settings.impl.IndexSettings`, not into a JsonObject.
* When you build the settings, push everything from the custom `IndexSettings` to the `IndexSettings` built by Hibernate Search. You can assume the "extra properties" in the settings built by Hibernate Search are empty, and for analysis definitions you can pick whatever's simpler in case of conflict: throw an exception, or override the definition from the analysis configurer.

Then you will be able to add validation for the "extra attributes". Maybe you can do something similar to `org.hibernate.search.backend.elasticsearch.validation.impl.IndexAliasDefinitionValidator#validate`? Though in this case, I would use `validateAllIgnoreUnexpected` instead of `validateAllIncludingUnexpected`: we don't want to fail when we encounter an unknown setting, because some people might chose to configure their indexes through index templates.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/cfg/ElasticsearchIndexSettings.java
```diff
@@ -57,8 +57,8 @@ private ElasticsearchIndexSettings() {
 	 * <p>
 	 * Expects a string representing the path to a file in the classpath, encoded in UTF-8. The file must contain index settings expressed in JSON format, with the exact same syntax as expected by the Elasticsearch server under the "settings" property <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-index.html#create-index-settings">when creating an index</a>. For example, if the file content is <code>{"index.codec": "best_compression"}</code>, it will set {@literal  index.codec} to {@literal best_compression}.
 	 * <p>
-	 * Note that the settings generated by Hibernate Search won't be applied.
-	 * This means for instance that any analyzer defined by the {@link #ANALYSIS_CONFIGURER} will be ignored.
+	 * Note that the settings generated by Hibernate Search will be overridden in case of conflict of some definitions.
+	 * This means for instance that any analyzer defined by the {@link #ANALYSIS_CONFIGURER} could be overridden (or not) by the provided custom settings.
```
yrodiere, line 61: I understood this as "maybe it will work, maybe it won't".

Maybe this will be less ambiguous:

```suggestion
 	 * For instance, if an analyzer "myAnalyzer" is defined by the {@link #ANALYSIS_CONFIGURER} <strong>and</strong> this settings file, the definition from the settings file will win. If it is only defined in either the analysis configurer or the settings file, but not both, it will be preserved as-is.
```

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/impl/CreateIndexWork.java
```diff
@@ -94,8 +94,11 @@ public Builder settings(IndexSettings settings, IndexSettings customSettings) {
 			 * We better not include the null fields.
 			 */
 			Gson gson = gsonProvider.getGsonNoSerializeNulls();
-			// TODO HSEARCH-3934 Merge customSettings into settings
-			payload.add( "settings", gson.toJsonTree( ( customSettings == null ) ? settings : customSettings ) );
+
+			// if customSettings are present, merge them with the ones created by Search
+			settings.merge( customSettings );
```
yrodiere, line 99: Since this is a mutating method, it would be better to execute it during metadata building, don't you think? It feels a bit weird to alter the metadata stored in the `IndexManager` just because someone requested to create the index in the Elasticsearch cluster.

fax4ever: right! it is better to merge them earlier

fax4ever:
There is also a full build: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2438/7/pipeline.
It does not include the last commit `HSEARCH-3934 Fix sonar issue`, but I don't think this is relevant.

yrodiere:
@fax4ever Could you add some tests for the migration/validation of custom settings, please? :) You can probably create test classes following the same pattern as the ones about analysis in `org.hibernate.search.integrationtest.backend.elasticsearch.schema.management`: `ElasticsearchIndexSchemaManagerCreationCustomSettingsIT`, `ElasticsearchIndexSchemaManagerUpdateCustomSettingsIT`, `ElasticsearchIndexSchemaManagerValidationCustomSettingsIT`.

fax4ever:
Adding the tests I discovered many issues, that I think I have fixed them all.

yrodiere:
@fax4ever Sonar reports an unused method. Please remove it?

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/schema/management/ElasticsearchIndexSchemaManagerUpdateCustomSettingsIT.java
```diff
@@ -0,0 +1,216 @@
...
+				"       } " +
+				"     } " +
+				"   } " +
+				" } "
+		);
+
+		assertThatThrownBy( () -> setupAndUpdateIndex() )
+				.isInstanceOf( SearchException.class )
+				.hasMessageContainingAll(
+						"Unable to update settings", "setting [index.number_of_shards], not updateable" );
```
yrodiere, line 193: I'd be surprised if all Elasticsearch versions returned the same error message. Did you run a full build?

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/schema/management/ElasticsearchIndexSchemaManagerCreationCustomSettingsIT.java
```diff
@@ -0,0 +1,232 @@
...
+				.isEqualTo( "\"3\"" );
+	}
+
+	@Test
+	public void success_mergeWithOverlapping() {
+		elasticsearchClient.index( index.name() )
+				.ensureDoesNotExist().registerForCleanup();
+
+		// merge the default analysis configurer with the custom settings,
+		// there are no overlapping of their definitions
```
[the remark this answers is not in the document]
fax4ever: yes there is

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/schema/management/impl/ElasticsearchSchemaMigratorImpl.java
```diff
@@ -92,10 +94,15 @@ public ElasticsearchSchemaMigratorImpl(ElasticsearchSchemaAccessor schemaAccesso
 		return schemaAccessor.updateAliases( indexName, aliases );
 	}
 
-	private CompletableFuture<?> doMigrateSettings(URLEncodedString indexName, IndexSettings settings) {
+	private CompletableFuture<?> doMigrateSettings(URLEncodedString indexName, IndexSettings expectedSettings,
+			IndexSettings actualSettings) {
+
+		// remove all already present settings extra attributes
+		expectedSettings.diff( actualSettings.getExtraAttributes() );
```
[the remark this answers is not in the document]
fax4ever: you're right... it would make sense if we were in the builder. Fixed

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/schema/management/ElasticsearchIndexSchemaManagerUpdateCustomSettingsIT.java
fax4ever: You're right. Testing the Elasticsearch generated error text is not a good idea in case, let's remove this assertion.

fax4ever:
Run a full build: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2438/12/pipeline

yrodiere (approved):
Ready to merge as soon as we start working on 6.1.

Thanks!

fax4ever:
Thank you!!

PR #2440, opened 2020-11-10
HSEARCH-3821 Improve test coverage for Elasticsearch schema validation

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3821

I also spent some time searching for some cases to cover [JsonElementEquivalence#isObjectEquivalent](https://sonarcloud.io/code?id=org.hibernate.search%3Ahibernate-search-parent&selected=org.hibernate.search%3Ahibernate-search-parent%3Abackend%2Felasticsearch%2Fsrc%2Fmain%2Fjava%2Forg%2Fhibernate%2Fsearch%2Fbackend%2Felasticsearch%2Fvalidation%2Fimpl%2FJsonElementEquivalence.java).
But I didn't find a case for which we have to compare JSON objects, according to the use cases we have had so far.

### DIFF
--- a/integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/schema/management/ElasticsearchIndexSchemaManagerValidationMappingAttributeIT.java
+++ b/integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/schema/management/ElasticsearchIndexSchemaManagerValidationMappingAttributeIT.java
@@ -212,6 +212,26 @@ public class ElasticsearchIndexSchemaManagerValidationMappingAttributeIT {
 	}
 
 	@Test
+	public void attribute_index_valid() {
+		StubMappedIndex index = StubMappedIndex.ofNonRetrievable(
+				root -> root.field( "myField", f -> f.asInteger().searchable( Searchable.YES ) ).toReference()
+		);
+
+		elasticSearchClient.index( index.name() ).deleteAndCreate();
+		elasticSearchClient.index( index.name() ).type().putMapping(
+				simpleMappingForInitialization(
+						"'myField': {"
+								+ "'type': 'integer',"
+								+ "'index': true"
+						+ "}"
+				)
+		);
+
+		// the expected value true is the default
+		setupAndValidate( index );
+	}
+
+	@Test
 	public void attribute_index_invalid() {
 		StubMappedIndex index = StubMappedIndex.ofNonRetrievable(
 				root -> root.field( "myField", f -> f.asInteger() ).toReference()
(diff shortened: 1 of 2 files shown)


integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/schema/management/ElasticsearchIndexSchemaManagerValidationMappingAttributeIT.java
```diff
@@ -211,6 +211,26 @@ public void attribute_index_missing() {
 		setupAndValidate( index );
 	}
 
+	@Test
```
yrodiere, line 214: Commit message: you meant "missing *test*"?

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/schema/management/ElasticsearchIndexSchemaManagerValidationMappingBaseIT.java
```diff
@@ -217,6 +217,28 @@ public void property_attribute_leniency() {
 		setupAndValidate( index );
 	}
 
+	@Test
+	public void floatAndDouble() {
```
yrodiere, line 221: Adding fields 'float'/'double' type is not enough to reach the validation code of `float`/`double` values: the JSON representation of the schema still doesn't include any `float`/`double` values, so we won't have to validate such values. You need to also add an `indexNullAs` so that we'll have to validate the `null_value` attribute in the schema.

yrodiere (changes requested):
My two cents.

Regarding `JsonElementEquivalence#isObjectEquivalent`, the only way to reach this code is through `JsonElementValidator`, which is used in `IndexAliasDefinitionValidator` to validate extra attributes in alias definitions, and also in `AnalysisDefinitionValidator` to validate parameters passed to an analyzer/token filter/tokenizer/etc.
To execute this code, you will have to find a valid schema that includes a JSON object as a parameter to an alias definition, or to an analyzer/token filter/tokenizer/etc.

fax4ever:
@yrodiere thanks for the review. I did the changes to tests on double and float.
About to test JSON objects as a parameter, I didn't find a valid Elasticsearch setting that would allow me to do that.
From my understanding the issue with alias definition is that we don't support filtering on them, so we cannot compare filters.
About custom analyzers/normalyzers, I tried for instance to define in place filters, but it seems that ES doesn't support that.

yrodiere:
>  From my understanding the issue with alias definition is that we don't support filtering on them, so we cannot compare filters.

Can't we create an index with a filtered alias, and check that validation fails?

fax4ever:
Hi @yrodiere, from my understanding we already have [such a test](https://github.com/hibernate/hibernate-search/blob/master/integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/schema/management/ElasticsearchIndexSchemaManagerValidationAliasesIT.java#L161).

yrodiere:
Ok, I suppose this validation code is only invoked when comparing two non-null objects. Seems like we can't reach it for now.

Merging, thanks!

fax4ever:
Exactly! Thanks

PR #2442, opened 2020-11-12
HSEARCH-4099 Allow option for setting Elasticsearch connection information via full URIs

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4099

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/ServerUris.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/ServerUris.java
@@ -74,6 +74,10 @@ final class ServerUris {
 	}
 
 	private static ServerUris fromStrings(String protocol, List<String> hostAndPortStrings) {
+		if ( hostAndPortStrings.isEmpty() ) {
+			throw log.emptyListOfHosts();
+		}
+
 		HttpHost[] hosts = new HttpHost[hostAndPortStrings.size()];
 		// Note: protocol and URI scheme are not the same thing,
 		// but for HTTP/HTTPS both the protocol and URI scheme are named HTTP/HTTPS.
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/ServerUris.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/ServerUris.java
@@ -65,7 +65,7 @@ final class ServerUris {
 			if ( https == null ) {
 				https = currentHttps;
 			}
-			else if ( ( currentHttps && !https || ( !currentHttps && https ) ) ) {
+			else if ( currentHttps != https ) {
 				throw log.differentProtocolsOnUris( serverUrisStrings );
 			}
 		}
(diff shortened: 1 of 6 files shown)


documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
```diff
@@ -105,6 +105,28 @@ The default for this property is `http`.
 
 This property may be set to either `http` or `https`.
 
+==== Provide complete URIs
+
+Alternatively to `hosts` and `protocol`, it is possible to define both the terms using the same property:
```
yrodiere, line 110:
```suggestion
 Alternatively, it is possible to define both the protocol and hosts as one or more URIs using a single property:
```

documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
```diff
@@ -105,6 +105,28 @@ The default for this property is `http`.
...
+hibernate.search.backend.uris = http://localhost:9200
+----
+
+This property may be set to a String representing an URI such as `http://localhost` or `https://es.mycompany.com:4400`,
+or a String containing multiple such URI strings separated by commas,
+or a `Collection<String>` containing such URI strings.
+
+[WARNING]
+====
+Note that there atrre some constraints we need to respect in order to use this property:
```
yrodiere, line 123:
```suggestion
There are some constraints regarding the use of this property:
```

documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
```diff
@@ -105,6 +105,28 @@ The default for this property is `http`.
...
+
+* All the uris must have the same protocol.
+* Cannot be used if `hosts` or `protocol` are set.
+* It must contain at least a valid URI.
```
yrodiere, line 127:
```suggestion
* The provided list of URIs must not be empty.
```

yrodiere (commented):
Nice, here are a few comments.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/ServerUris.java
```diff
@@ -27,7 +29,51 @@ private ServerUris(HttpHost[] hosts, boolean sslEnabled) {
...
+
+		if ( hostAndPortStrings.isPresent() ) {
+			throw log.uriAndHosts( uris.get(), hostAndPortStrings.get() );
+		}
+
+		return fromStrings( uris.get() );
+	}
+
+	private static ServerUris fromStrings(List<String> serverUrisStrings) {
+		if ( serverUrisStrings.isEmpty() ) {
```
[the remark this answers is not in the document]
fax4ever: I don't think it either, I'm adding the test...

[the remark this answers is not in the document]
fax4ever: At the moment we get an error from the ES client.
``` bash
java.lang.AssertionError: 
Expecting:
  <java.lang.IllegalArgumentException: hosts must not be null nor empty>
to be an instance of:
  <org.hibernate.search.util.common.SearchException>
but was:
  <"java.lang.IllegalArgumentException: hosts must not be null nor empty

...
```

fax4ever:
Thanks for review.
I might have applied the suggestions.

yrodiere:
Merged, thanks

PR #2444, opened 2020-11-14
HSEARCH-4106 + HSEARCH-4107 + HSEARCH-4108 Session proxy fixes

yrodiere:
* [HSEARCH-4106](https://hibernate.atlassian.net/browse/HSEARCH-4106): Force tests to execute with JUnit 4
* [HSEARCH-4108](https://hibernate.atlassian.net/browse/HSEARCH-4108): Search.session() does not work with session proxies 
* [HSEARCH-4107](https://hibernate.atlassian.net/browse/HSEARCH-4107): Session cast fails when creating a FullTextSession with Spring 2.4.0

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -250,14 +250,6 @@
         <version.log4j>2.13.3</version.log4j>
         <version.junit>4.13.1</version.junit>
 
-        <!--
-            junit-vintage-engine 5.5.2 version does not support junit 4.13.1.
-            If we try to use it, a parsing error is thrown.
-            So we downgrade JUnit, so that junit-vintage-engine will handle it.
-            See https://stackoverflow.com/a/64371503/1812965
-        -->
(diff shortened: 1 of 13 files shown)


pom.xml
```diff
@@ -282,7 +274,7 @@
         <version.org.google.guava>30.0-jre</version.org.google.guava>
 
         <!-- >>> Spring integration tests -->
-        <version.org.springframework.boot>2.3.4.RELEASE</version.org.springframework.boot>
+        <version.org.springframework.boot>2.4.0-RC1</version.org.springframework.boot>
```
andrei-ivanov, line 277: `2.4.0` is out

yrodiere: Right, I was looking for 2.4.0.RELEASE and couldn't find it, but it seems they dropped the `.RELEASE` suffix.
I updated the PR, thanks for the heads-up.

PR #2445, opened 2020-11-16
HSEARCH-4088 Type safe PropertyBridge/TypeBridge

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4088

There some stuff I'm not very sure about. For instance I added some extra method to `PropertyBindingContext` to bind list and map typed property bridges. Let's use the review to discuss about it.

### DIFF
--- a/documentation/src/main/asciidoc/reference/mapper-orm-bridge-index-field-dsl.asciidoc
+++ b/documentation/src/main/asciidoc/reference/mapper-orm-bridge-index-field-dsl.asciidoc
@@ -33,10 +33,9 @@ include::{sourcedir}/org/hibernate/search/documentation/mapper/orm/bridge/docume
 This is done through a lambda taking advantage of another DSL.
 See <<mapper-orm-bridge-index-field-type-dsl>> for more information.
 <5> Get a reference to the declared field.
-<6> Pass the reference to the bridge for later use.
-Optionally you can provide the expected type for the bridged property,
-using the method `<P> bridge(Class<P>, PropertyBridge<P>)`,
-to statically define the type and so avoid casting on the bridge.
+<6> Call `context.bridge(...)` to define the bridge to use.
+<7> Pass the expected type of values.
+<8> Pass the bridge instance.
 
 [source, JAVA, indent=0, subs="+callouts"]
 ----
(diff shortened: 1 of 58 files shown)


yrodiere (changes requested):
Thanks.

First, I think some tests are missing, in particular regarding bridges implementing `PropertyBridge` or `TypeBridge` without generic type arguments. We need to check backward compatibility. I added comments in the relevant places in the diff.

Second, I don't think it's a good idea to introduce the `mapPropertyBrige`/`listPropertyBridge` methods:

1. If we do introduce these methods, we'll need to check the type of elements upon binding the bridge so that they are type-safe. Your current implementation just ignores the parameters, so it gives the impression that it's type-safe, but actually it's not. It would be possible to implement these checks, but...
2. The problem is much broader than just `List`/`Map`. There are other collection types, obviously. We obviously don't want to add a new method every time we support another collection type. Worse, there are custom parameterized types (`MyEntity<T>`), for which we cannot possibly exposed one method per type. So we would need a different approach, perhaps similar to Google's `TypeToken` (see `com.google.gson.reflect.TypeToken#getParameterized`).
3. We have similar problems in other places. Obviously we have the same problem with`ValueBridge`, and we didn't address it there. But there's also `BridgesConfigurationContext`: currently you can only declare a default bridge for `List`, but it would be nice to be able to declare a default bridge for `List<String>`.
4. That's a weaker argument, but I think we should at least test covariance: what if the user requests `List<Number>` but the property is of type `List<Integer>` or `List<? extends Number>`? Do we allow it?

In short: implications are much broader than it seems. Let's not commit to an API until we have a solution that we're confident will correctly address the use cases we have in mind, and is flexible enough to consistently address future use cases too.

There's already a relevant ticket, maybe we should expand the description to also cover `PropertyBinder`/`TypeBinder`: https://hibernate.atlassian.net/browse/HSEARCH-3355

yrodiere (approved):
*WARNING*: I pushed commits to your branch to fix a few remaining issues.

Apart from that, +1 to merge as soon as CI agrees.

fax4ever:
I missed them. Thank you!

yrodiere:
Merged, thanks!

PR #2447, opened 2020-11-17
HSEARCH-3479 + HSEARCH-4096 Spring Boot hangs while trying to retrieve (built-in) beans for Hibernate Search

yrodiere:
* [HSEARCH-4096](https://hibernate.atlassian.net/browse/HSEARCH-4096): Spring Boot hangs while trying to retrieve (built-in) beans for Hibernate Search
* [HSEARCH-3479](https://hibernate.atlassian.net/browse/HSEARCH-3479): Migrate Search 5 CDI and Spring integration tests to Search 6

The core of the issue is that Spring Boot, in its default configuration, is unable to retrieve beans while Hibernate ORM is starting. If we try that, a deadlock occurs.
There are ways around that (see the comments on the ticket), but none that we can implement from within Hibernate Search.

So, if the user relies on custom beans in their Hibernate Search mapping, they'll have to configure Spring appropriately to work around the bug.

Now, if the user does *not* rely on custom beans, on the other hand... There's no reason Hibernate Search should try to retrieve beans from Spring. And yet, it does, and the deadlock happens. The reason is that we always try Spring/CDI first when retrieving beans, even when retrieving *built-in* beans (e.g. annotation processors, container extractors, GeoPointBinder, ...).

This PR changes the default behavior of bean retrieval from Hibernate Search so that we rely on the internal registry or reflection for all built-in beans.

The main side effect of those changes is that newly introduced new built-in beans in future versions of Hibernate Search will take precedence over user beans. So for example if a user had an analysis configurer named `simple`, and in Search 6.1 we introduce a built-in analysis configurer named `simple`, when the user upgrades all of a sudden their own bean will be ignored.

This is annoying, but there's not much I can do to avoid that. I can offer workaround, however: by introducing a new syntax for bean references, we offer the ability to explicitly skip some bean sources. For example `bean:myBeanName` will only look for beans in Spring, and won't try built-in beans or reflection. `class:org.mycompany.MyClass` will try Spring and reflection, but will not try built-in beans.

Since this is all a bit complicated, I made sure that using these prefixes is not mandatory, even for custom beans. So existing configuration where you set the property `hibernate.search.backend.analysis.configurer` to `com.mycompany.MyAnalysisConfigurer` will continue to work: you won't have to change it to `class:com.mycompany.MyAnalysisConfigurer`. As long as there is no naming conflict between a built-in bean and a custom bean, you will be alright.

I also added integration tests for Spring bean resolution. I didn't really migrate Search 5 tests in the end, but rather took inspiration from what we already had for CDI, because that test looked more complete.

Note I did not migrate Spring-JTA integration tests, because it does not work. I'll handle that in https://hibernate.atlassian.net/browse/HSEARCH-4112


documentation/src/main/asciidoc/reference/configuration.asciidoc
```diff
@@ -187,36 +185,110 @@ This includes, but may not be limited to:
...
 
-* If a dependency injection framework is integrated into Hibernate ORM,
+* If the given reference matches a built-in bean, that bean is used.
++
+Example: the name `simple`,
+when used as the value of the property `hibernate.search.backend.layout.strategy`
+to configure the <<backend-elasticsearch-indexlayout,Elasticsearch index layout strategy>>,
+resolves to the built-in `simple` strategy.
+* Otherwise, if a dependency injection framework is integrated into Hibernate ORM,
 the reference is first requested to the DI framework.
```
fax4ever, line 254: I don't know if `first` still makes here. Even if we are clearly in the `Otherwise` branch, I wouldn't use the term `first`, to be make 100% clear that this check is performed only after we check the `built-in` bean repository.

integrationtest/mapper/orm-spring/src/test/java/org/hibernate/search/integrationtest/spring/beanresolution/SpringBeanResolutionIT.java
```diff
@@ -0,0 +1,277 @@
...
+			assertThat( StaticCounters.get().get( counterKeys.instantiate ) ).isEqualTo( expectedInstances );
+			assertThat( StaticCounters.get().get( counterKeys.postConstruct ) ).isEqualTo( expectedInstances );
+			assertThat( StaticCounters.get().get( counterKeys.preDestroy ) ).isEqualTo( 0 );
+		}
+
+		switch ( expectedScope ) {
+			case PROTOTYPE:
+				// Spring does not call @PreDestroy on prototype beans
+				// See https://docs.spring.io/spring-framework/docs/5.3.1/reference/html/core.html#beans-factory-scopes-prototype
+				assertThat( StaticCounters.get().get( counterKeys.preDestroy ) ).isEqualTo( 0 );
```
fax4ever, line 156: Maybe if the framework expected behavior is not respected, we should assert nothing about it.

fax4ever (approved):
Really a nice job!
I have some comments that you can ignore.

integrationtest/mapper/orm-spring/src/test/java/org/hibernate/search/integrationtest/spring/beanresolution/SpringBeanResolutionIT.java
yrodiere: I'd like to know if they change the behavior.

engine/src/main/java/org/hibernate/search/engine/environment/bean/impl/BeanConfigurationContextImpl.java
```diff
@@ -33,8 +32,8 @@
 		configuredBeans( exposedType ).add( name, reference );
 	}
 
-	Map<Class<?>, BeanReferenceRegistryForType<?>> configuredBeans() {
-		return Collections.unmodifiableMap( configuredBeans );
+	ConfigurationBeanRegistry buildRegistry() {
+		return new ConfigurationBeanRegistry( new HashMap<>( configuredBeans ) );
```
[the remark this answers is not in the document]
yrodiere: It's not really useful here, since `ConfigurationBeanRegistry` "owns" the map and never changes it.

engine/src/main/java/org/hibernate/search/engine/environment/bean/impl/BeanSource.java
```diff
@@ -0,0 +1,29 @@
...
+
+	/**
+	 * The beans defined using
+	 * {@link org.hibernate.search.engine.environment.bean.spi.BeanConfigurer}s.
+	 */
+	CONFIGURATION,
+	/**
+	 * The bean manager, e.g. CDI, Spring, ...
+	 */
+	BEAN_MANAGER,
```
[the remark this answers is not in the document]
yrodiere: While I agree this name is an approximation, we need to say *something* about the source in error messages, and I'd rather keep consistent wording everywhere.

Right now I feel that "bean manager" will be the clearest way to say "CDI/Spring" to users without saying it explicitly. And it's currently accurate since the only `BeanProvider` implementations we currently use are based on CDI/Spring.

yrodiere:
Thanks. I updated the PR, I will merge soon.

PR #2450, opened 2020-11-18
HSEARCH-4112 Hibernate Search + Spring-JTA fails because of closed entity manager

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4112


mapper/orm/src/main/java/org/hibernate/search/mapper/orm/session/impl/HibernateOrmSearchSession.java
```diff
@@ -249,27 +245,23 @@ public PojoRuntimeIntrospector runtimeIntrospector() {
...
-
-		TransientReference<Map<Transaction, PojoIndexingPlan<EntityReference>>> reference =
-				(TransientReference<Map<Transaction, PojoIndexingPlan<EntityReference>>>) sessionImplementor.getProperties()
-				.get( INDEXING_PLAN_PER_TRANSACTION_MAP_KEY );
-		Map<Transaction, PojoIndexingPlan<EntityReference>> planPerTransaction = reference == null ? null : reference.get();
-		if ( planPerTransaction == null ) {
-			planPerTransaction = new HashMap<>();
-			reference = new TransientReference<>( planPerTransaction );
-			sessionImplementor.setProperty( INDEXING_PLAN_PER_TRANSACTION_MAP_KEY, reference );
+		HibernateOrmSearchSessionHolder properties =
```
fax4ever, line 249: This is not critical, but I think that `properties` is not a very clear name for the holder instance.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/session/impl/HibernateOrmSearchSessionHolder.java
```diff
@@ -0,0 +1,118 @@
...
+		if ( holder != null ) {
+			return holder;
+		}
+		boolean closedSessionAndTransaction = session.isClosed() && session.isTransactionInProgress();
+		if ( closedSessionAndTransaction ) {
+			// This can happen when using JTA.
+			// Only do this as a fallback: if somehow the session holder was added to the session properties
+			// before the session was closed, we definitely want to use it and avoid the static map.
+			holder = holderPerClosedSessionTransaction.get( session.accessTransaction() );
+		}
```
fax4ever, line 49: This is not critical either, the code looks good in any case, but maybe we can introduce 2 sub classes: one for JTA and the other for the non-JTA case.

fax4ever (approved):
Really a nice job! +1 to merge it
Just very optional comments.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/session/impl/HibernateOrmSearchSessionHolder.java
yrodiere: This is a static method, no polymorphism here.

Also, we will not always end up in this branch, even with JTA: if a `SearchSession` was retrieved before the commit, we'll proceed with the "normal" behavior (which is better).

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/session/impl/HibernateOrmSearchSession.java
yrodiere: Right, will do

yrodiere:
Merged, thanks

PR #2455, opened 2020-11-20
HSEARCH-4116 Ignore empty, blank or null configuration properties when detecting unused properties

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4116

### DIFF
--- a/util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/ElasticsearchTestHostConnectionConfiguration.java
+++ b/util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/ElasticsearchTestHostConnectionConfiguration.java
@@ -29,7 +29,7 @@ public class ElasticsearchTestHostConnectionConfiguration {
 	private final String protocol;
 	private final String username;
 	private final String password;
-	private final boolean awsSigningEnabled;
+	private final Boolean awsSigningEnabled;
 	private final String awsRegion;
 	private final String awsCredentialsType;
 	private final String awsCredentialsAccessKeyId;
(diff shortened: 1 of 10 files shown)


util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/ElasticsearchTestHostConnectionConfiguration.java
```diff
@@ -53,20 +54,20 @@ private ElasticsearchTestHostConnectionConfiguration() {
...
 		properties.put( "username", username );
 		properties.put( "password", password );
-		properties.put( "aws.signing.enabled", String.valueOf( awsSigningEnabled ) );
+		properties.put( "aws.signing.enabled", awsSigningEnabled == null ? null : awsSigningEnabled.toString() );
 		properties.put( "aws.region", awsRegion );
 		properties.put( "aws.credentials.type", awsCredentialsType );
 		properties.put( "aws.credentials.access_key_id", awsCredentialsAccessKeyId );
 		properties.put( "aws.credentials.secret_access_key", awsCredentialsSecretAccessKey );
-		if ( awsSigningEnabled ) {
+		if ( awsSigningEnabled != null && awsSigningEnabled ) {
```
fax4ever, line 70: Maybe you can reuse `isAws()`. It looks the same.

fax4ever (approved):
Looks good. Just a comment

util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/ElasticsearchTestHostConnectionConfiguration.java
yrodiere: Right; it should actually be just `awsSigningEnabled != null`.

yrodiere: Actually I was looking in the wrong place; you're absolutely right.

yrodiere:
Merged, thanks!

PR #2463, opened 2020-12-18
HSEARCH-4128 Close client in case initially fails

bostko:
https://hibernate.atlassian.net/browse/HSEARCH-4128


yrodiere:
Thanks! The patch looks good and I'll merge it if tests pass, but did you check that this actually solves your application hanging problem? Forgetting to close the client is for sure a waste of resources, but I'm not sure how it could lead to the application hanging... If you checked and its works though, I'll believe you.

bostko:
@yrodiere I rebuild locally `5.11.5.Final` with the change and it did work.
Screenshot from 2020-12-18 18-20-01

yrodiere (approved):

yrodiere:
Merged, thanks!

PR #2468, opened 2021-01-18
HSEARCH-4137 Reindexing is skipped when a contained entity change follows a change that is irrelevant to indexing

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4137


fax4ever:
Merged thanks!

PR #2469, opened 2021-01-21
HSEARCH-4138 + HSEARCH-4139 Avoid unnecessary memory allocations on entity updates

yrodiere:
* [HSEARCH-4139](https://hibernate.atlassian.net/browse/HSEARCH-4139): Rely on BitSet rather than Set<String> to track updated properties
* [HSEARCH-4138](https://hibernate.atlassian.net/browse/HSEARCH-4138): Avoid unnecessary memory allocations on entity updates

This may improve performance a bit when entity changes do *not* require reindexing; we'll see.

Even if it doesn't improve performance, the formalization of conversion between the string representation of dirty paths and another representation will be useful when we work on [HSEARCH-4133](https://hibernate.atlassian.net/browse/HSEARCH-4133) and [HSEARCH-4141](https://hibernate.atlassian.net/browse/HSEARCH-4141), so I would merge it anyway.

Creating as draft, since this targets 6.1.


fax4ever:
It looks good. +1 to merge when we're working on 6.1 branch

yrodiere:
Merged, thanks!

PR #2471, opened 2021-01-26
HSEARCH-3771 + HSEARCH-4144 Mass indexing for ORM's dynamic-map entity types

yrodiere:
* [HSEARCH-4144](https://hibernate.atlassian.net/browse/HSEARCH-4144): Refactor ORM entity loading classes and move them to a dedicated package
* [HSEARCH-3771](https://hibernate.atlassian.net/browse/HSEARCH-3771): Mass indexing for ORM's dynamic-map entity types


fax4ever:
Great job! Probably we can use that for the Batch JSR-352 module too. Merging as it is...

PR #2475, opened 2021-02-05
HSEARCH-4150 Entity loading fails when using FetchMode.SUBSELECT

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4150

It's just a workaround for now, the real fix needs to be implemented in ORM.


fax4ever:
Looks good. Merged thanks!

PR #2479, opened 2021-02-09
HSEARCH-3880 Add query cache and query cache policy definition

wklaczynski:
### Defining QueryCache and QueryCachingPolicy for hibernate search.

https://hibernate.atlassian.net/browse/HSEARCH-3880

The extension allows you to define a custom implementation for QueryCache and QueryCachingPolicy. 

**Definitions are possible:** 

- Define with JPA parametr:

```
hibernate.search.backends.portal.cache.query_cache_provider=PortalQueryCachingPolicyProvider.class
hibernate.search.backends.portal.cache.query_caching_policy_provider=PortalQueryCachingPolicyProvider.class
```
Exemple custom implementation for QueryCache and QueryCachingPolicy:

```java
public class PortalQueryCacheProvider  implements QueryCacheProvider {

    public QueryCache getQueryCache(BackendBuildContext context, ConfigurationPropertySource properties, Version luceneVersion) {
        final int maxCachedQueries = 1000;
        // min of 32MB or 5% of the heap size
        final long maxRamBytesUsed = Math.min(1L << 25, Runtime.getRuntime().maxMemory() / 20);
        
        return new LRUQueryCache(maxCachedQueries, maxRamBytesUsed, new MinSegmentSizePredicate(0, .03f), 250);
    }

    static class MinSegmentSizePredicate implements Predicate<LeafReaderContext> {

        private final int minSize;
        private final float minSizeRatio;

        MinSegmentSizePredicate(int minSize, float minSizeRatio) {
            this.minSize = minSize;
            this.minSizeRatio = minSizeRatio;
        }

        @Override
        public boolean test(LeafReaderContext context) {
            final int maxDoc = context.reader().maxDoc();
            if (maxDoc < minSize) {
                return false;
            }
            final IndexReaderContext topLevelContext = ReaderUtil.getTopLevelContext(context);
            final float sizeRatio = (float) context.reader().maxDoc() / topLevelContext.reader().maxDoc();
            return sizeRatio >= minSizeRatio;
        }
    }
}
```
```java
public class PortalQueryCachingPolicyProvider implements QueryCachingPolicyProvider {

    public QueryCachingPolicy getQueryCachingPolicy(BackendBuildContext context, ConfigurationPropertySource properties, Version luceneVersion) {
        return new PortalQueryCachingPolicy();
    }

    public static class PortalQueryCachingPolicy extends UsageTrackingQueryCachingPolicy {

        @Override
        protected int minFrequencyToCache(Query query) {
            if (isCostly(query)) {
                return 2;
            } else {
                return super.minFrequencyToCache(query);
            }
        }

        private boolean isCostly(Query query) {
            return true;
        }
    }
}
```

- Define with Hibernate ORM parametr with StrategyRegistrationProvider:
```
hibernate.search.backend.cache.query_cache_class=portal
hibernate.search.backend.cache.query_caching_policy_class=portal
```
```java
@MetaInfServices(StrategyRegistrationProvider.class)
public class StrategyRegistrationProviderImpl implements StrategyRegistrationProvider {

    @Override
    public Iterable<StrategyRegistration> getStrategyRegistrations() {
      List<StrategyRegistration> strategyRegistrations = new ArrayList<>();
      
      strategyRegistrations.add(
            new SimpleStrategyRegistrationImpl<>(
                  QueryCachingPolicy.class,
                  PortalQueryCachingPolicy.class,
                  "portal",
                  PortalQueryCachingPolicy.class.getName(),
                  PortalQueryCachingPolicy.class.getSimpleName()
            )
      );
      
      strategyRegistrations.add(
            new SimpleStrategyRegistrationImpl<>(
                  QueryCache.class,
                  PortalQueryCache.class,
                  "portal",
                  PortalQueryCache.class.getName(),
                  PortalQueryCache.class.getSimpleName()
            )
      );
      
      return strategyRegistrations;
    }
}
```
Exemple custom implementation for QueryCache and QueryCachingPolicy:

```java
public class PortalQueryCache implements QueryCache {

    private final LRUQueryCache delegate;

    public PortalQueryCache(ConfigurationPropertySource properties, Version luceneVersion) {
        final int maxCachedQueries = 1000;
        // min of 32MB or 5% of the heap size
        final long maxRamBytesUsed = Math.min(1L << 25, Runtime.getRuntime().maxMemory() / 20);

        delegate = new LRUQueryCache(maxCachedQueries, maxRamBytesUsed, new MinSegmentSizePredicate(0, .03f), 250);
    }

    @Override
    public Weight doCache(Weight weight, QueryCachingPolicy policy) {
        return delegate.doCache(weight, policy);
    }

    static class MinSegmentSizePredicate implements Predicate<LeafReaderContext> {

        private final int minSize;
        private final float minSizeRatio;

        MinSegmentSizePredicate(int minSize, float minSizeRatio) {
            this.minSize = minSize;
            this.minSizeRatio = minSizeRatio;
        }

        @Override
        public boolean test(LeafReaderContext context) {
            final int maxDoc = context.reader().maxDoc();
            if (maxDoc < minSize) {
                return false;
            }
            final IndexReaderContext topLevelContext = ReaderUtil.getTopLevelContext(context);
            final float sizeRatio = (float) context.reader().maxDoc() / topLevelContext.reader().maxDoc();
            return sizeRatio >= minSizeRatio;
        }
    }
}
```

```java
public class PortalQueryCachingPolicy implements QueryCachingPolicy {

    private final UsageTrackingQueryCachingPolicy delegate;

    public PortalQueryCachingPolicy(ConfigurationPropertySource properties, Version luceneVersion) {
        delegate = new UsageTrackingQueryCachingPolicy();
    }

    @Override
    public void onUse(Query query) {
        delegate.onUse(query);
    }

    @Override
    public boolean shouldCache(Query query) throws IOException {
        return delegate.shouldCache(query);
    }
}
```

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cache/QueryCachingConfigurationContext.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cache/QueryCachingConfigurationContext.java
@@ -1,0 +1,33 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.lucene.cache;
+
+import org.apache.lucene.search.QueryCache;
+import org.apache.lucene.search.QueryCachingPolicy;
+import org.apache.lucene.util.Version;
+
+/**
+ * A context allowing the configuration of query caching in a Lucene backend.
+ */
+public interface QueryCachingConfigurationContext {
+
+	/**
+	 * @return The Lucene version in use in the configured backend.
+	 */
+	Version luceneVersion();
+
+	/**
+	 * @param cache The {@link QueryCache} to use when searching.
+	 */
+	void queryCache(QueryCache cache);
+
+	/**
+	 * @param policy The {@link QueryCachingPolicy} to use when searching.
+	 */
+	void queryCachingPolicy(QueryCachingPolicy policy);
+
+}
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cfg/LuceneBackendSettings.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cfg/LuceneBackendSettings.java
@@ -66,6 +66,18 @@ public final class LuceneBackendSettings {
 	public static final String ANALYSIS_CONFIGURER = "analysis.configurer";
 
 	/**
+	 * The analysis configurer to use.
+	 * <p>
+	 * Expects a reference to a bean of type {@link LuceneAnalysisConfigurer}.
+	 * <p>
+	 * Defaults to no value.
+	 *
+	 * @see org.hibernate.search.engine.cfg The core documentation of configuration properties,
+	 * which includes a description of the "bean reference" properties and accepted values.
+	 */
+	public static final String QUERY_CACHING_CONFIGURER = "caching.configurer";
+
+	/**
 	 * The size of the thread pool assigned to the backend.
 	 * <p>
 	 * Expects a strictly positive integer value,
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cfg/LuceneBackendSettings.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cfg/LuceneBackendSettings.java
@@ -7,6 +7,7 @@
 package org.hibernate.search.backend.lucene.cfg;
 
 import org.hibernate.search.backend.lucene.analysis.LuceneAnalysisConfigurer;
+import org.hibernate.search.backend.lucene.cache.QueryCachingConfigurer;
 import org.hibernate.search.backend.lucene.multitenancy.MultiTenancyStrategyName;
 
 import org.apache.lucene.util.Version;
@@ -66,9 +67,9 @@ public final class LuceneBackendSettings {
 	public static final String ANALYSIS_CONFIGURER = "analysis.configurer";
 
 	/**
-	 * The analysis configurer to use.
+	 * The query caching configurer to use.
 	 * <p>
-	 * Expects a reference to a bean of type {@link LuceneAnalysisConfigurer}.
+	 * Expects a reference to a bean of type {@link QueryCachingConfigurer}.
 	 * <p>
 	 * Defaults to no value.
 	 *
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cfg/LuceneBackendSettings.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cfg/LuceneBackendSettings.java
@@ -76,7 +76,7 @@ public final class LuceneBackendSettings {
 	 * @see org.hibernate.search.engine.cfg The core documentation of configuration properties,
 	 * which includes a description of the "bean reference" properties and accepted values.
 	 */
-	public static final String QUERY_CACHING_CONFIGURER = "caching.configurer";
+	public static final String QUERY_CACHING_CONFIGURER = "query.caching.configurer";
 
 	/**
 	 * The size of the thread pool assigned to the backend.
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/impl/LuceneBackendFactory.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/impl/LuceneBackendFactory.java
@@ -8,6 +8,7 @@ package org.hibernate.search.backend.lucene.impl;
 
 import java.lang.invoke.MethodHandles;
 import java.text.ParseException;
+import java.util.List;
 import java.util.Locale;
 import java.util.Optional;
 
@@ -39,8 +40,9 @@ import org.hibernate.search.util.common.logging.impl.LoggerFactory;
 import org.hibernate.search.util.common.reporting.EventContext;
 
 import org.apache.lucene.util.Version;
-
-
+import org.hibernate.search.backend.lucene.cache.QueryCachingConfigurationContext;
+import org.hibernate.search.backend.lucene.cache.QueryCachingConfigurer;
+import org.hibernate.search.backend.lucene.cache.impl.LuceneQueryCachingContext;
 
 public class LuceneBackendFactory implements BackendFactory {
 
@@ -62,6 +64,11 @@ public class LuceneBackendFactory implements BackendFactory {
 					.asBeanReference( LuceneAnalysisConfigurer.class )
 					.build();
 
+	private static final OptionalConfigurationProperty<BeanReference<? extends QueryCachingConfigurer>> QUERY_CACHING_CONFIGURER
+			= ConfigurationProperty.forKey( LuceneBackendSettings.QUERY_CACHING_CONFIGURER )
+					.asBeanReference( QueryCachingConfigurer.class )
+					.build();
+
 	@Override
 	public BackendImplementor create(EventContext eventContext, BackendBuildContext buildContext,
 			ConfigurationPropertySource propertySource) {
@@ -78,11 +85,17 @@ public class LuceneBackendFactory implements BackendFactory {
 					buildContext, propertySource, luceneVersion
 			);
 
+			LuceneQueryCachingContext cachingContext
+					= new LuceneQueryCachingContext( luceneVersion );
+
+			configureQueryCache( buildContext, propertySource, cachingContext );
+
 			return new LuceneBackendImpl(
 					eventContext,
 					backendThreads,
 					new LuceneWorkFactoryImpl( multiTenancyStrategy ),
 					analysisDefinitionRegistry,
+					cachingContext,
 					multiTenancyStrategy,
 					buildContext.timingSource(),
 					buildContext.failureHandler()
@@ -157,6 +170,31 @@ public class LuceneBackendFactory implements BackendFactory {
 		}
 		catch (Exception e) {
 			throw log.unableToApplyAnalysisConfiguration( e.getMessage(), e );
+		}
+	}
+
+	private void configureQueryCache(
+			BackendBuildContext buildContext, ConfigurationPropertySource propertySource,
+			QueryCachingConfigurationContext context) {
+		try {
+			final BeanResolver beanResolver = buildContext.beanResolver();
+			try ( BeanHolder<List<QueryCachingConfigurer>> implicitConfigurersHolder =
+					beanResolver.resolve( beanResolver.allConfiguredForRole( QueryCachingConfigurer.class ) ) ) {
+				for ( QueryCachingConfigurer configurer : implicitConfigurersHolder.get() ) {
+					configurer.configure( context );
+				}
+			}
+
+			// Apply the user-provided query cache configurer if necessary
+			QUERY_CACHING_CONFIGURER.getAndMap( propertySource, beanResolver::resolve )
+					.ifPresent( holder -> {
+						try ( BeanHolder<? extends QueryCachingConfigurer> configurerHolder = holder ) {
+							configurerHolder.get().configure( context );
+						}
+					} );
+		}
+		catch (Exception e) {
+			throw log.unableToApplyQueryCacheConfiguration( e.getMessage(), e );
 		}
 	}
 
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/logging/impl/Log.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/logging/impl/Log.java
@@ -650,4 +650,8 @@ public interface Log extends BasicLogger {
 	@Message(id = ID_OFFSET + 145,
 		value = "Unknown nmaed predicate definition '%1$s'.")
 	SearchException unknownNamedPredicateForSearch(String name, @Param EventContext context);
+
+	@Message(id = ID_OFFSET + 146,
+			value = "Unable to apply query caching configuration: %1$s")
+	SearchException unableToApplyQueryCacheConfiguration(String errorMessage, @Cause Exception e);
 }
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cache/QueryCachingConfigurer.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cache/QueryCachingConfigurer.java
@@ -1,0 +1,25 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.lucene.cache;
+
+import org.hibernate.search.backend.lucene.cfg.LuceneBackendSettings;
+
+/**
+ * A configurer for query caching.
+ * <p>
+ * Users can select a configurer through the
+ * {@link LuceneBackendSettings#QUERY_CACHING_CONFIGURER configuration properties}.
+ */
+public interface QueryCachingConfigurer {
+
+	/**
+	 * Configures query caching as necessary using the given {@code context}.
+	 * @param context A context exposing methods to configure caching.
+	 */
+	void configure(QueryCachingConfigurationContext context);
+
+}
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cache/impl/LuceneQueryCachingContext.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cache/impl/LuceneQueryCachingContext.java
@@ -1,0 +1,48 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.lucene.cache.impl;
+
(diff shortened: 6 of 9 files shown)


yrodiere (changes requested):
Thanks! That's a nice feature. That being said, I think we could improve how it is exposed.

I'll start with the problems I noticed, so that you can understand why I'm making a suggestion, and I'll suggest a solution at the end.

## Problems

### Mixing mappers and backends

`HibernateOrmQueryCacheProvider` in particular is problematic, because it introduces a dependency from the ORM mapper to the Lucene backend.

We've worked very hard to remove that kind of dependency in Hibernate Search 6. I will not merge anything that reintroduces that dependency.

Fortunately that was just one of many ways to configure caching, so we can just drop this one ;) 

Which brings me to my next point...

### Do one thing, and do it well

There are many ways to configure caching. I see at least four different ways:

* Setting `hibernate.search.backend.cache.query_cache_class`
* OR setting `hibernate.search.backend.cache.query_cache_provider`
* OR registering your own `StrategyRegistrationProvider` to create caches or caching policies
* OR registering a `QueryCacheProvider`/`QueryCachingPolicyProvider` as a `ServiceLoader`.

While choice is nice, I'm not sure it's useful in this case. Worse, the more options we expose, the harder it becomes to test all possible options and their combinations.

Let's focus and offer a single way to configure caching? If there really is a compelling reason to offer configuration through the `ServiceLoader`, for example, we could address this in a separate pull request.

Which brings me to...

### Consistency of bean retrieval

There are other pluggable components in Hibernate Search. And while each type of components brings its own set of challenges,  being able to reference the pluggable components in multiple ways is definitely something that's already been addressed.

There are already [many ways to reference pluggable components in Hibernate Search](https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#configuration-bean-reference-parsing): by class name, by CDI/Spring bean name, or even by [a name you assign yourself in Hibernate Search](https://github.com/hibernate/hibernate-search/blob/074572c814c3481571d476bcaff03b78a952df39/engine/src/main/java/org/hibernate/search/engine/environment/bean/spi/BeanConfigurer.java#L10) (this last one is an SPI and not API, though).

You can benefit from that very easily, without adding any extra code. No need to deal with `ServiceLoaders` or ORM `StrategyRegistrations`.

I doubt we really need to add support for `ServiceLoaders` or `StrategyRegistrations`, but if you think we do, feel free to create a ticket and explain what they bring compared to existing solutions. Then we'll be able to address this... in a different ticket/pull request, and globally for **all pluggable components**, not just caching.

### Consistency of configuration APIs

Finally, there are other places in Hibernate Search 6 where users can provide custom pluggable components. They generally follow a common approach: the user provides a component which acts as a callback, which gets passed contextual information and gets a chance to set various configuration options.

See for example:

* `BeanConfigurer`
* `LuceneAnalysisConfigurer`
* `ElasticsearchAnalysisConfigurer`
* `HibernateOrmSearchMappingConfigurer`
* `ElasticsearchHttpClientConfigurer`

I think the same approach would make sense for caching, so let's try to be consistent and use it?

## Solution

Fortunately, most of my comments above are about doing less, not more.

I'm suggesting the following API.

### Configuration

```
hibernate.search.backend.query.caching.configurer = com.mycompany.MyCachingConfigurer
```

OR (if it's a CDI/Spring bean, or if it's defined in a `BeanConfigurer`):

```
hibernate.search.backend.query.caching.configurer = my-caching-configurer-name
```

### API

```java
/**
 * A configurer for query caching.
 * <p>
 * Users can select a configurer through the
 * {@link LuceneBackendSettings#QUERY_CACHING_CONFIGURER configuration properties}.
 */
public interface QueryCachingConfigurer {

	/**
	 * Configures query caching as necessary using the given {@code context}.
	 * @param context A context exposing methods to configure caching.
	 */
	void configure(QueryCachingConfigurationContext context);

}
```

```java
/**
 * A context allowing the configuration of query caching in a Lucene backend.
 */
public interface QueryCachingConfigurationContext {

	/**
	 * @return The Lucene version in use in the configured backend.
	 */
	Version luceneVersion();

	/**
	 * @param cache The {@link QueryCache} to use when searching.
	 */
	void queryCache(QueryCache cache);

	/**
	 * @param policy The {@link QueryCachingPolicy} to use when searching.
	 */
	void queryCachingPolicy(QueryCachingPolicy policy);

}
```

### Usage in your specific case

```java
@MetaInfServices(BeanConfigurer.class)
public class PortalBeanConfigurer implements BeanConfigurer {
	@Override
	public void configure(BeanConfigurationContext context) {
		context.define(
				QueryCachingConfigurer.class,
				"portal",
				BeanReference.ofInstance((QueryCachingConfigurer) context -> {
					context.queryCache(new PortalQueryCache());
					context.queryCachingPolicy(new PortalQueryCachingPolicy());
				})
		);
	}
}
```

```
hibernate.search.backend.query.caching.configurer=portal
```

Thoughts?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cfg/LuceneBackendSettings.java
```diff
@@ -65,6 +69,54 @@ private LuceneBackendSettings() {
...
+	 * <p>
+	 * Expects a reference to a bean of type {@link QueryCachingPolicy}.
+	 * <p>
+	 * Defaults to no value.
+	 *
+	 * @see org.hibernate.search.engine.cfg The core documentation of configuration properties,
+	 * which includes a description of the "bean reference" properties and accepted values.
+	 */
+	public static final String QUERY_CACHING_POLICY_CLASS = "cache.query_caching_policy_class";
+
```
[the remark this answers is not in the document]
wklaczynski: Of course, this can be simplified. I'll take care of it.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/impl/LuceneBackendFactory.java
```diff
@@ -160,6 +173,25 @@ private LuceneAnalysisDefinitionRegistry getAnalysisDefinitionRegistry(
 		}
 	}
 
+	private void configureQueryCashe(
+			BackendBuildContext buildContext, ConfigurationPropertySource propertySource,
+			QueryCachingConfigurationContext context) {
+		try {
+			LuceneCachingServiceConfigurer.configure( buildContext, context );
```
yrodiere, line 180: I don't think we discussed this? You're essentially applying all caching configurers you can find, just because they are in the classpath, and without having to insert anything in the configuration properties.

Personally, I think this kind of "magic" brings more headaches than it helps and I don't think we should do it.

But... if you really, really want such a feature, please do it the Hibernate Search way:

```java
try ( BeanHolder<List<QueryCachingConfigurer>> implicitConfigurersHolder =
        beanResolver.resolve( beanResolver.allConfiguredForRole( QueryCachingConfigurer.class ) ) ) {
    for ( QueryCachingConfigurer configurer : implicitConfigurersHolder.get() ) {
        configurer.configure( context );
    }
}
```

Developers could then register their configurers this way, and they would be applied automatically:

```java
@MetaInfServices(BeanConfigurer.class)
public class PortalBeanConfigurer implements BeanConfigurer {
	@Override
	public void configure(BeanConfigurationContext context) {
		context.define(
				QueryCachingConfigurer.class,
				BeanReference.ofInstance((QueryCachingConfigurer) context -> {
					context.queryCache(new PortalQueryCache());
					context.queryCachingPolicy(new PortalQueryCachingPolicy());
				})
		);
	}
}
```

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/logging/impl/Log.java
```diff
@@ -636,4 +636,8 @@ SearchException cannotUseQueryElementForObjectFieldBecauseCreationException(Stri
 			value = "Unable to create instance of analysis component '%1$s': %2$s")
 	SearchException unableToCreateAnalysisComponent(@FormatWith(ClassFormatter.class) Class<?> type, String causeMessage,
 			@Cause Exception cause);
+
+	@Message(id = ID_OFFSET + 143,
+			value = "Error while applying query cache configuration: %1$s")
```
yrodiere, line 641: To be consistent with `unableToApplyAnalysisConfiguration()`:

```suggestion
			value = "Unable to apply query caching configuration: %1$s")
```

engine/src/main/java/org/hibernate/search/engine/backend/spi/BackendBuildContext.java
```diff
@@ -22,6 +23,8 @@
 
 	ResourceResolver resourceResolver();
 
+	ServiceResolver serviceResolver();
```
yrodiere, line 26: If you follow my recommendation from other comments, you can remove this.

yrodiere (changes requested):

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cache/QueryCachingContext.java
```diff
@@ -0,0 +1,30 @@
...
+package org.hibernate.search.backend.lucene.cache;
+
+import java.util.Optional;
+import org.apache.lucene.search.QueryCache;
+import org.apache.lucene.search.QueryCachingPolicy;
+
+/**
+ * A context allowing the configuration of query caching in a Lucene backend.
+ */
+public interface QueryCachingContext {
```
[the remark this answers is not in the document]
wklaczynski: I'll remove it.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cache/QueryCachingConfigurationContext.java
```diff
@@ -0,0 +1,33 @@
...
+package org.hibernate.search.backend.lucene.cache;
+
+import org.apache.lucene.search.QueryCache;
+import org.apache.lucene.search.QueryCachingPolicy;
+import org.apache.lucene.util.Version;
+
+/**
+ * A context allowing the configuration of query caching in a Lucene backend.
+ */
+public interface QueryCachingConfigurationContext extends QueryCachingContext {
```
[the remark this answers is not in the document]
wklaczynski: For code aesthetics, not to define other class LuceneQueryCachingConfigurationContext by constructor parameters. I will remove "QueryCachingContext".

Maybe change "LuceneQueryCachingConfigurationContext" to "LuceneQueryCachingContext", which will implement "QueryCachingConfigurationContext", and define "LuceneQueryCachingContext" through parameters?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/impl/LuceneBackendFactory.java
wklaczynski: I'll change it.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/logging/impl/Log.java
wklaczynski: I'll change it.

engine/src/main/java/org/hibernate/search/engine/backend/spi/BackendBuildContext.java
wklaczynski: I'll remove it.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cache/QueryCachingConfigurationContext.java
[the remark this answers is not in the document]
yrodiere: You could simply have `LuceneQueryCachingConfigurationContext` extend both `QueryCachingConfigurationContext` and `QueryCachingContext`? Though personally, I'd use two separate implementations, in order to guarantee immutability of the `QueryCachingContext`.

yrodiere:
@wklaczynski Don't worry about the failing build, the failure is unrelated: https://hibernate.atlassian.net/browse/HSEARCH-4164

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cache/QueryCachingConfigurationContext.java
```diff
@@ -0,0 +1,46 @@
...
+	/**
+	 * @param cache The {@link QueryCache} to use when searching.
+	 */
+	void queryCache(QueryCache cache);
+
+	/**
+	 * @return The corresponding {@link QueryCache}, or {@link Optional#empty()} if it doesn't exist.
+	 *
+	 */
+	Optional<QueryCache> queryCache();
```
yrodiere, line 33: I'm still not following why you need this here :) Implementors are expected to define the cache and policy: why would they need to read it?
If you need this for implementation purposes, then it has nothing to do here in the API. Just expose it in the implementation class, but not in the interface.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/impl/LuceneBackendFactory.java
```diff
@@ -160,6 +173,31 @@ private LuceneAnalysisDefinitionRegistry getAnalysisDefinitionRegistry(
 		}
 	}
 
+	private void configureQueryCashe(
```
yrodiere, line 176:
```suggestion
	private void configureQueryCache(
```

yrodiere (commented):
Thanks! The API and implementation look great now, though I added a few minor comments.

Could you please add some tests?

You can put the tests in `integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/cache`.

I'm thinking you could:

* Test the most common error cases. To that end, you can copy `integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/analysis/LuceneAnalysisConfigurerIT.java` and adapt it to the query cache configurer (some tests may not be relevant, especially the one about parameters).
* Test that the cache and cache policy are actually used by Hibernate Search. To that end, you can copy `integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/analysis/AnalysisCustomIT.java` and adapt it to the query cache configurer. I would simply use a custom cache configurer that sets up a stub cache and a stub caching policy. Then run a query, and check that the cache and caching policy were called. You can use Mockito if it's convenient: it's already in the dependencies.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cache/QueryCachingConfigurationContext.java
```diff
@@ -0,0 +1,46 @@
...
+	/**
+	 * @param policy The {@link QueryCachingPolicy} to use when searching.
+	 */
+	void queryCachingPolicy(QueryCachingPolicy policy);
+
+	/**
+	 * @return The corresponding {@link QueryCachingPolicy}, or {@link Optional#empty()} if it doesn't exist.
+	 *
+	 */
+	Optional<QueryCachingPolicy> queryCachingPolicy();
```
yrodiere, line 44: Same comment as for `queryCache()`:
> I'm still not following why you need this here :) Implementors are expected to define the cache and policy: why would they need to read it?
> If you need this for implementation purposes, then it has nothing to do here in the API. Just expose it in the implementation class, but not in the interface.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cache/QueryCachingConfigurationContext.java
wklaczynski: In service it can load successive implementations recursively. While defining, it was already added, you can make a delegating implementation that covers the existing one. Thanks to these parameters, the programmer knows that the object is already defined and has the ability to cover it by covering some of its functions. 

In other library:
```java 
@MetaInfServices(QueryCachingConfigurer.class)
public class OtherQueryCachingConfigurer implements QueryCachingConfigurer {

    @Override
    public void configure(QueryCachingConfigurationContext context) {
        context.queryCache(new GlobalQueryCache(context.luceneVersion()));
        context.queryCachingPolicy(new GlobalQueryCachingPolicy(context.luceneVersion()));
    }
}
```
In custom project:
```java
@MetaInfServices(QueryCachingConfigurer.class)
public class MyAppQueryCachingConfigurer implements QueryCachingConfigurer {

    @Override
    public void configure(QueryCachingConfigurationContext context) {
        
        context.queryCache().ifPresentOrElse((previous) -> {
            context.queryCache(new WrappedQueryCache(context.luceneVersion()));
        }, () -> {
            context.queryCache(new MyQueryCache(context.luceneVersion()));
        });
    }    
}
```

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/impl/LuceneBackendFactory.java
wklaczynski: I'll change it.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/cache/QueryCachingConfigurationContext.java
yrodiere: The main problem I see with this is that there is not guarantee about the order in which the configurers are applied, apart from "implicit ones are applied before the one specified in the configuration".

If you have multiple caching configurers defined through the ServiceLoader, they can be applied in a different order depending on the JARs they come from, or even depending on the number of registered beans (due to `HashMap` rebalancings).

So yes, I think it's better to remove this until we have an actual use case.

If someone wants to apply multiple query caches, they can write their own "composite" configurer, which would delegate to other configurers in a well-defined order. And if that's not enough, they can submit a ticket and we can think better how to solve their problem once we have more information about the use case :)

wklaczynski: I'll change it.

yrodiere (approved):
Thanks. I added a few fix-up commits for the last remaining problems.

I'll wait for the CI build to pass, then I will merge.

PR #2480, opened 2021-02-09
HSEARCH-4142 Identifier projection

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4142

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchIdProjection.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchIdProjection.java
@@ -64,7 +64,7 @@ public class ElasticsearchIdProjection<I> extends AbstractElasticsearchProjectio
 					ValueConvert.YES );
 
 			// check expected identifier type:
-			identifierValueConverter.requiresType( identifierType );
+			identifierValueConverter.checkSourceTypeAssignableTo( identifierType );
 			@SuppressWarnings("uncheked") // just checked
 			DocumentIdentifierValueConverter<? extends I> casted = (DocumentIdentifierValueConverter<? extends I>) identifierValueConverter;
 
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchIdProjection.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchIdProjection.java
@@ -1,0 +1,79 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.elasticsearch.search.projection.impl;
+
+import org.hibernate.search.backend.elasticsearch.search.impl.ElasticsearchSearchContext;
+import org.hibernate.search.engine.backend.types.converter.spi.DocumentIdentifierValueConverter;
+import org.hibernate.search.engine.search.common.ValueConvert;
+import org.hibernate.search.engine.search.loading.spi.LoadingResult;
+import org.hibernate.search.engine.search.loading.spi.ProjectionHitMapper;
+import org.hibernate.search.engine.search.projection.SearchProjection;
+import org.hibernate.search.engine.search.projection.spi.IdProjectionBuilder;
+
+import com.google.gson.JsonObject;
+
+public class ElasticsearchIdProjection<I> extends AbstractElasticsearchProjection<String, I> {
+
+	private final ProjectionExtractionHelper<String> extractionHelper;
+	private final DocumentIdentifierValueConverter<? extends I> identifierValueConverter;
+
+	private ElasticsearchIdProjection(ElasticsearchSearchContext searchContext,
+			ProjectionExtractionHelper<String> extractionHelper,
+			DocumentIdentifierValueConverter<? extends I> identifierValueConverter) {
+		super( searchContext );
+		this.extractionHelper = extractionHelper;
+		this.identifierValueConverter = identifierValueConverter;
+	}
+
+	@Override
+	public String toString() {
+		return getClass().getSimpleName();
+	}
+
+	@Override
+	public void request(JsonObject requestBody, SearchProjectionRequestContext context) {
+		extractionHelper.request( requestBody, context );
+	}
+
+	@Override
+	public String extract(ProjectionHitMapper<?, ?> projectionHitMapper, JsonObject hit,
+			SearchProjectionExtractContext context) {
+		return extractionHelper.extract( hit, context );
+	}
+
+	@Override
+	public I transform(LoadingResult<?, ?> loadingResult, String extractedData,
+			SearchProjectionTransformContext context) {
+		return identifierValueConverter.convertToSource(
+				extractedData, context.fromDocumentIdentifierValueConvertContext() );
+	}
+
+	static class Builder<I> extends AbstractBuilder<I> implements IdProjectionBuilder<I> {
+
+		private final ElasticsearchIdProjection<I> projection;
+
+		Builder(ElasticsearchSearchContext searchContext, ProjectionExtractionHelper<String> extractionHelper,
+				Class<I> identifierType) {
+			super( searchContext );
+
+			DocumentIdentifierValueConverter<?> identifierValueConverter = searchContext.indexes().idDslConverter(
+					ValueConvert.YES );
+
+			// check expected identifier type:
+			identifierValueConverter.requiresType( identifierType );
+			@SuppressWarnings("uncheked") // just checked
+			DocumentIdentifierValueConverter<? extends I> casted = (DocumentIdentifierValueConverter<? extends I>) identifierValueConverter;
+
+			this.projection = new ElasticsearchIdProjection<>( searchContext, extractionHelper, casted );
+		}
+
+		@Override
+		public SearchProjection<I> build() {
+			return projection;
+		}
+	}
+}
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/projection/impl/LuceneIdProjection.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/projection/impl/LuceneIdProjection.java
@@ -61,7 +61,7 @@ public class LuceneIdProjection<I> extends AbstractLuceneProjection<String, I> {
 					.idDslConverter( ValueConvert.YES );
 
 			// check expected identifier type:
-			identifierValueConverter.requiresType( identifierType );
+			identifierValueConverter.checkSourceTypeAssignableTo( identifierType );
 			@SuppressWarnings("uncheked") // just checked
 			DocumentIdentifierValueConverter<? extends I> casted = (DocumentIdentifierValueConverter<? extends I>) identifierValueConverter;
 
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/projection/impl/LuceneIdProjection.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/projection/impl/LuceneIdProjection.java
@@ -1,0 +1,76 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.lucene.search.projection.impl;
+
+import org.hibernate.search.backend.lucene.lowlevel.collector.impl.IdentifierCollector;
+import org.hibernate.search.backend.lucene.search.extraction.impl.LuceneResult;
+import org.hibernate.search.backend.lucene.search.impl.LuceneSearchContext;
+import org.hibernate.search.engine.backend.types.converter.spi.DocumentIdentifierValueConverter;
+import org.hibernate.search.engine.search.common.ValueConvert;
+import org.hibernate.search.engine.search.loading.spi.LoadingResult;
+import org.hibernate.search.engine.search.loading.spi.ProjectionHitMapper;
+import org.hibernate.search.engine.search.projection.SearchProjection;
+import org.hibernate.search.engine.search.projection.spi.IdProjectionBuilder;
+
+public class LuceneIdProjection<I> extends AbstractLuceneProjection<String, I> {
+
+	private final DocumentIdentifierValueConverter<? extends I> identifierValueConverter;
+
+	private LuceneIdProjection(LuceneSearchContext searchContext,
+			DocumentIdentifierValueConverter<? extends I> identifierValueConverter) {
+		super( searchContext );
+		this.identifierValueConverter = identifierValueConverter;
+	}
+
+	@Override
+	public String toString() {
+		return getClass().getSimpleName();
+	}
+
+	@Override
+	public void request(SearchProjectionRequestContext context) {
+		context.requireCollector( IdentifierCollector.FACTORY );
+	}
+
+	@SuppressWarnings("unchecked")
+	@Override
+	public String extract(ProjectionHitMapper<?, ?> mapper, LuceneResult documentResult,
+			SearchProjectionExtractContext context) {
+		return context.getCollector( IdentifierCollector.KEY ).get( documentResult.getDocId() );
+	}
+
+	@Override
+	public I transform(LoadingResult<?, ?> loadingResult, String extractedData,
+			SearchProjectionTransformContext context) {
+		return identifierValueConverter.convertToSource(
+				extractedData, context.fromDocumentIdentifierValueConvertContext() );
+	}
+
+	public static class Builder<I> extends AbstractBuilder<I> implements IdProjectionBuilder<I> {
+
+		private final LuceneIdProjection<I> projection;
+
+		public Builder(LuceneSearchContext searchContext, Class<I> identifierType) {
+			super( searchContext );
+
+			DocumentIdentifierValueConverter<?> identifierValueConverter = searchContext.indexes()
+					.idDslConverter( ValueConvert.YES );
+
+			// check expected identifier type:
+			identifierValueConverter.requiresType( identifierType );
+			@SuppressWarnings("uncheked") // just checked
+			DocumentIdentifierValueConverter<? extends I> casted = (DocumentIdentifierValueConverter<? extends I>) identifierValueConverter;
+
+			projection = new LuceneIdProjection<>( searchContext, casted );
+		}
+
+		@Override
+		public SearchProjection<I> build() {
+			return projection;
+		}
+	}
+}
--- a/documentation/src/main/asciidoc/reference/search-dsl-projection.asciidoc
+++ b/documentation/src/main/asciidoc/reference/search-dsl-projection.asciidoc
@@ -87,6 +87,30 @@ include::{sourcedir}/org/hibernate/search/documentation/search/projection/Projec
 NOTE: The entity does not get loaded as part of the projection.
 If you want the actual entity instance, use the <<search-dsl-projection-entity,`entity` projection>>
 
+== `id`: return ids to matched entities
+
+The `id` projection returns the id to the matched entity.
+The API allows to express the expected entity type of the returning ids.
+
+.Returning ids to matched entities, providing the identity type.
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/projection/ProjectionDslIT.java[tags=id-int]
+----
+====
+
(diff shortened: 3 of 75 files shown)


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/projection/impl/LuceneIdProjection.java
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	@Override
+	public String toString() {
+		return getClass().getSimpleName();
+	}
+
+	@Override
+	public void request(SearchProjectionRequestContext context) {
+		context.requireCollector( DocumentReferenceCollector.FACTORY );
```
yrodiere, line 34: Similarly to my comment on the Elasticsearch backend, I would use a different collector here: one that does not collect document references, but document IDs only (Strings). You could copy/paste `DocumentReferenceCollector` and remove everything related to the entity type (`metadataResolver`, `currentLeafMappedTypeName`, ...).

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/projection/SearchProjectionIT.java
```diff
@@ -122,20 +122,23 @@ public void references() {
...
 				.toQuery();
 		assertThatQuery( query ).hasListHitsAnyOrder( b -> {
-			b.list( document1Reference, document1Reference, document1Reference );
-			b.list( document2Reference, document2Reference, document2Reference );
-			b.list( document3Reference, document3Reference, document3Reference );
-			b.list( emptyReference, emptyReference, emptyReference );
+			b.list( document1Reference, document1Reference, document1Reference, DOCUMENT_1 );
+			b.list( document2Reference, document2Reference, document2Reference, DOCUMENT_2 );
+			b.list( document3Reference, document3Reference, document3Reference, DOCUMENT_3 );
+			b.list( emptyReference, emptyReference, emptyReference, EMPTY );
```
yrodiere, line 141: I think this feature it warrants additional testing, especially if you follow my recommendation of implementing it similarly to the ID predicate.

I think we'd need tests inspired from these, at least:

* `org.hibernate.search.integrationtest.backend.tck.search.projection.FieldSearchProjectionSingleValuedBaseIT#simple`
* `org.hibernate.search.integrationtest.backend.tck.search.projection.FieldSearchProjectionSingleValuedBaseIT#noClass`
* `org.hibernate.search.integrationtest.backend.tck.search.projection.FieldSearchProjectionSingleValuedBaseIT#duplicated`
* `org.hibernate.search.integrationtest.backend.tck.search.projection.FieldSearchProjectionTypeCheckingAndConversionIT#validSuperClass`
* `org.hibernate.search.integrationtest.backend.tck.search.projection.FieldSearchProjectionTypeCheckingAndConversionIT#invalidProjectionType_projectionConverterEnabled`
* `org.hibernate.search.integrationtest.backend.tck.search.projection.FieldSearchProjectionTypeCheckingAndConversionIT#multiIndex_withCompatibleIndex_projectionConverterEnabled`
* `org.hibernate.search.integrationtest.backend.tck.search.projection.FieldSearchProjectionTypeCheckingAndConversionIT#multiIndex_withIncompatibleIndex_projectionConverterEnabled`
* `org.hibernate.search.integrationtest.backend.tck.search.projection.FieldSearchProjectionTypeIndependentIT#nullClass` (this one doesn't need its own test class in the case of the ID projection, obviously)

I'd suggest you keep the organization with one `*IT` class per concern, since it's done that way in many other tests.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/work/IndexingFieldTypesIT.java
```diff
@@ -102,6 +102,19 @@ public void withReference() {
...
+					.select( f -> f.composite(
+							(id, val) -> new IdAndValue<>( id, val ),
+							f.id( String.class ),
+							f.field( absoluteFieldPath, typeDescriptor.getJavaType() )
+					) )
+					.where( f -> f.matchAll() )
+					.toQuery();
+
+			assertThatQuery( query ).hasHitsAnyOrder( expectedDocuments );
+		}
```
yrodiere, line 117: I think you can remove the block above with the entity reference projection. This test is not about testing the entity reference projection, we're just using it because it's convenient.

But it's a good idea to use the ID projection instead, now that we have it.

v5migrationhelper/orm/src/main/java/org/hibernate/search/scope/impl/V5MigrationOrmSearchScopeAdapter.java
```diff
@@ -56,7 +56,7 @@ public SearchSortFactory sort() {
 	@Override
 	public SearchProjection<Object> idProjection() {
 		SearchProjectionFactory<EntityReference, ?> factory = delegate.projection();
-		return factory.composite( EntityReference::id, factory.entityReference() ).toProjection();
+		return factory.id().toProjection();
```
yrodiere, line 59: Nice work replacing `entityReference` projection with `id` projections everywhere!

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/search/SearchQueryBaseIT.java
```diff
@@ -38,6 +38,7 @@
 import org.hibernate.search.mapper.pojo.mapping.definition.annotation.GenericField;
 import org.hibernate.search.mapper.pojo.mapping.definition.annotation.Indexed;
 import org.hibernate.search.mapper.pojo.mapping.definition.annotation.IndexedEmbedded;
+import org.hibernate.search.util.common.SearchException;
 import org.hibernate.search.util.common.SearchTimeoutException;
 import org.hibernate.search.util.impl.integrationtest.common.rule.BackendMock;
 import org.hibernate.search.util.impl.integrationtest.common.rule.StubSearchWorkBehavior;
```
yrodiere, line 44: Unrelated to this commit, but I need to put this comment somewhere.

Could you please also add tests for all ID types? I'm thinking of copy/pasting `org.hibernate.search.integrationtest.mapper.pojo.mapping.definition.DocumentIdDefaultBridgeBaseIT#projection` into a new method that uses the `id()` projection instead of the `entityReference()` projection.

Same with `org.hibernate.search.integrationtest.mapper.pojo.mapping.definition.DocumentIdDefaultBridgeOverridingIT#projection`.

documentation/src/main/asciidoc/reference/search-dsl-projection.asciidoc
```diff
@@ -87,6 +87,30 @@ include::{sourcedir}/org/hibernate/search/documentation/search/projection/Projec
 NOTE: The entity does not get loaded as part of the projection.
 If you want the actual entity instance, use the <<search-dsl-projection-entity,`entity` projection>>
 
+== `id`: return ids to matched entities
+
+The `id` projection returns the id to the matched entity.
```
yrodiere, line 92:
```suggestion
The `identifier` projection returns the identifier of the matched entity.
```

documentation/src/main/asciidoc/reference/search-dsl-projection.asciidoc
```diff
@@ -87,6 +87,30 @@ include::{sourcedir}/org/hibernate/search/documentation/search/projection/Projec
...
+The API allows to express the expected entity type of the returning ids.
```
yrodiere, line 93:
```suggestion
NOTE: If the provided identifier type does not match the type of identifiers for targeted entity types,
an exception will be thrown.
See also <<search-dsl-projected-value-type>>.
```

documentation/src/main/asciidoc/reference/search-dsl-projection.asciidoc
```diff
@@ -87,6 +87,30 @@ include::{sourcedir}/org/hibernate/search/documentation/search/projection/Projec
...
+
+.Returning ids to matched entities, providing the identity type.
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/projection/ProjectionDslIT.java[tags=id-int]
+----
+====
+
```
yrodiere, line 102: Please use a structure similar to the documentation of the `field` projection:

```suggestion
You can omit the "identifier type" argument, but then you will get projections of type `Object`:

```

documentation/src/main/asciidoc/reference/search-dsl-projection.asciidoc
```diff
@@ -87,6 +87,30 @@ include::{sourcedir}/org/hibernate/search/documentation/search/projection/Projec
 NOTE: The entity does not get loaded as part of the projection.
 If you want the actual entity instance, use the <<search-dsl-projection-entity,`entity` projection>>
 
+== `id`: return ids to matched entities
```
yrodiere, line 90:
```suggestion
== `id`: return identifiers of matched entities 
```

documentation/src/main/asciidoc/reference/search-dsl-projection.asciidoc
```diff
@@ -87,6 +87,30 @@ include::{sourcedir}/org/hibernate/search/documentation/search/projection/Projec
...
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/projection/ProjectionDslIT.java[tags=id-object]
+----
+====
+
+NOTE: If the provided identity type does not match the current returned ids,
+an exception will be thrown.
+
```
yrodiere, line 113:
```suggestion
```

yrodiere (commented):
Thanks! My comments below. Mainly, I think the implementation should be more similar to what we do for ID predicates.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchIdProjection.java
```diff
@@ -0,0 +1,68 @@
...
+import org.hibernate.search.engine.search.loading.spi.LoadingResult;
+import org.hibernate.search.engine.search.loading.spi.ProjectionHitMapper;
+import org.hibernate.search.engine.search.projection.SearchProjection;
+import org.hibernate.search.engine.search.projection.spi.IdProjectionBuilder;
+
+import com.google.gson.JsonObject;
+
+public class ElasticsearchIdProjection<I> extends AbstractElasticsearchProjection<DocumentReference, I> {
+
+	private final DocumentReferenceExtractionHelper helper;
```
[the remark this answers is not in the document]
fax4ever: I'm not sure that the document ids are enough to load the entity ids. I think that we still need `DocumentReference`s.

engine/src/main/java/org/hibernate/search/engine/backend/common/spi/DocumentReferenceConverter.java
```diff
@@ -12,4 +12,6 @@
 
 	R fromDocumentReference(DocumentReference reference);
 
+	Object entityId(DocumentReference reference);
```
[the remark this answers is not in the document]
fax4ever: Sure. It is better to check at projection time creation!

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/projection/impl/LuceneIdProjection.java
fax4ever: > I'm not sure that the document ids are enough to load the entity ids. I think that we still need DocumentReferences.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchIdProjection.java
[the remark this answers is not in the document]
yrodiere: Let's agree to disagree. As I said, have a look at how it's done for ID predicates.

All you need to convert a document ID to an entity ID is a converter, which will be backed by `org.hibernate.search.mapper.pojo.bridge.IdentifierBridge#fromDocumentIdentifier`. The type name is not involved there.

You have an example of how it can work with the `IdDslConverter`, which does the opposite job (converts from entity ID to document ID):

* At mapping time, `org.hibernate.search.engine.backend.document.model.dsl.spi.IndexSchemaRootNodeBuilder#idDslConverter` allows the mapper to register an ID converter for a given mapped type.
* The implementation of that converter for the POJO mapper is `org.hibernate.search.mapper.pojo.bridge.runtime.impl.PojoIdentifierBridgeToDocumentIdentifierValueConverter`, and it's backed by an `IdentifierBridge`.
* When creating an ID predicate, the identifier is retrieved in `org.hibernate.search.backend.elasticsearch.search.predicate.impl.ElasticsearchMatchIdPredicate.Builder#value`
* As you can see, `org.hibernate.search.backend.elasticsearch.scope.model.impl.ElasticsearchScopeSearchIndexesContext#idDslConverter` retrieves the converter for the targeted indexes, checking that all targeted indexes have the same converter.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/projection/impl/LuceneIdProjection.java
yrodiere: Same answer.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchIdProjection.java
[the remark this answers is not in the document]
fax4ever: you're right!

yrodiere:
@fax4ever There are conflicts with the master branch, but I'd recommend you address my comments *before* rebasing. I think most of the conflicts will just go away after you change how the projection is implemented.

fax4ever:
Thank you for the review @yrodiere!
Doing the changes takes some time, since now the id projections do not rely anymore on document reference conversion, but they use the bridge directly with the document id.
I have some doubts about some choices, they sound good to me, but I'm not sure they are good to you as well.
So I'm expecting at least another series of comments and I'll be here to address them.

documentation/src/main/asciidoc/reference/search-dsl-projection.asciidoc
```diff
@@ -87,6 +87,30 @@ include::{sourcedir}/org/hibernate/search/documentation/search/projection/Projec
 NOTE: The entity does not get loaded as part of the projection.
 If you want the actual entity instance, use the <<search-dsl-projection-entity,`entity` projection>>
 
+== `id`: return identifiers of matched entities 
+
+The `identifier` projection returns the identifier of the matched entity.
+NOTE: If the provided identifier type does not match the type of identifiers for targeted entity types,
+an exception will be thrown.
+See also <<search-dsl-projected-value-type>>.
```
yrodiere, line 95:
```suggestion
```

documentation/src/main/asciidoc/reference/search-dsl-projection.asciidoc
```diff
@@ -87,6 +87,30 @@ include::{sourcedir}/org/hibernate/search/documentation/search/projection/Projec
...
+
+.Returning ids to matched entities, providing the identity type.
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/projection/ProjectionDslIT.java[tags=id-int]
+----
+====
+You can omit the "identifier type" argument, but then you will get projections of type `Object`:
```
yrodiere, line 104:
```suggestion
If the provided identifier type does not match the type of identifiers for targeted entity types,
an exception will be thrown. See also <<search-dsl-projected-value-type>>. 

You can omit the "identifier type" argument, but then you will get projections of type `Object`:
```

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/projection/IdentifierSearchProjectionSingleValuedIT.java
```diff
@@ -0,0 +1,205 @@
...
+import org.hibernate.search.util.impl.integrationtest.mapper.stub.BulkIndexer;
+import org.hibernate.search.util.impl.integrationtest.mapper.stub.SimpleMappedIndex;
+import org.hibernate.search.util.impl.integrationtest.mapper.stub.StubMappedIndex;
+import org.hibernate.search.util.impl.integrationtest.mapper.stub.StubMappingScope;
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class IdentifierSearchProjectionSingleValuedIT {
```
yrodiere, line 32: You should rename this test to `IdentifierSearchProjectionBaseIT`. Identifiers are always single-valued.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/search/SearchQueryBaseIT.java
```diff
@@ -474,6 +482,21 @@ public void select_searchProjection_multiple() {
...
+			assertThatThrownBy( () -> searchSession.search( Book.class )
+					.select( f -> f.id( String.class ) ) // wrong identifier type
+			)
+					.isInstanceOf( SearchException.class )
+					.hasMessageContainingAll(
+							"The required identifier type 'java.lang.String' does not match the actual identifier type 'java.lang.Integer'"
+					);
+		} );
+	}
+
```
yrodiere, line 499: I think you can remove this.

Invalid types are already tested on the backend side in `IdentifierSearchProjectionSingleValuedIT#invalidProjectionType`.

On the mapper side, I don't think you need a dedicated test for ORM; you're already dealing with this in `integrationtest/mapper/pojo-base`.

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/projection/IdProjectionIT.java
```diff
@@ -0,0 +1,68 @@
...
+		try ( SearchSession session = mapping.createSession() ) {
+			assertThatThrownBy( () -> session.search( IndexedEntity.class )
+					.select( f -> f.id( String.class ) )
+			)
+					.isInstanceOf( SearchException.class )
+					.hasMessageContainingAll(
+							"The required identifier type 'java.lang.String' does not match the actual identifier type 'java.lang.Integer'"
+					);
+		}
+	}
```
yrodiere, line 55: I think you can remove this.

On the backend side, you're already testing invalid projection types in `IdentifierSearchProjectionSingleValuedIT#invalidProjectionType`.

On the mapper side, it does indeed make sense to test this too, but IMO the only reason it makes sense is because we have many different bridges with different types. So I'd suggest you test this for every default bridge.

Fortunately, there's an easy way to do it: add a test method in `DocumentIdDefaultBridgeBaseIT` instead. You can take inspiration from `org.hibernate.search.integrationtest.mapper.pojo.mapping.definition.FieldDefaultBridgeBaseIT#indexToProjectionConverter`.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/projection/IdentifierSearchProjectionSingleValuedIT.java
```diff
@@ -0,0 +1,205 @@
...
+	@Test
+	public void multiIndex_withCompatibleIndex() {
+		StubMappingScope scope = index.createScope( compatibleIndex );
+		SearchQuery<String> query = scope.query()
+				.select( f -> f.id( String.class ) )
+				.where( f -> f.matchAll() )
+				.toQuery();
+
+		List<String> result = query.fetchHits( 30 );
+		assertThat( result ).containsExactlyInAnyOrder( ids );
```
yrodiere, line 130: This test would make more sense if the compatible index actually contained documents, don't you think?

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/projection/IdentifierSearchProjectionSingleValuedIT.java
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	private void initData() {
+		BulkIndexer indexer = index.bulkIndexer();
+		for ( int i = 0; i < 7; i++ ) {
+			int finalI = i;
+			indexer.add( ids[i], f -> f.addValue( index.binding().name, names[finalI] ) );
+		}
+
+		indexer.join();
```
yrodiere, line 176: FYI you can also do this:

```suggestion
		index.bulkIndexer()
				.add( 7, i -> documentProvider(
						ids[i],
						document -> document.addValue( index.binding().name, names[i] )
				) )
				.join();
```

fax4ever:
Thanks for the second review here.
I think that all the comments ( with which I 100% agree ) have been addressed.
I ran [this build](https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2480/7/pipeline) to test all the supported environments.

yrodiere (approved):

PR #2482, opened 2021-02-09
HSEARCH-4155 Deleting then re-inserting the same entity in a transaction does not result in an update

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4155

Best reviewed commit by commit, since one of the commits is a large (internal) renaming.


fax4ever (approved):
Looks good. +1 merge and backport as soon as the Ci gives us a positive feedback

yrodiere:
Merged, thanks!

PR #2484, opened 2021-02-09
HSEARCH-1302 Test match on stopwords-only full-text field

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-1302

I just added a test for it

### DIFF
--- a/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/analysis/StopwordAnalyzerIT.java
+++ b/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/analysis/StopwordAnalyzerIT.java
@@ -1,0 +1,102 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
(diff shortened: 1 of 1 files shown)


yrodiere:
@fax4ever How is this test different from `org.hibernate.search.integrationtest.backend.tck.search.predicate.AbstractPredicateSimpleAnalysisIT#noTokenAfterAnalysis`?

fax4ever:
In new test we have a document with a field having all stopwords... So the stopwords are provided at writing time.

yrodiere:
> In new test we have a document with a field having all stopwords... So the stopwords are provided at writing time.

I believe in `org.hibernate.search.integrationtest.backend.tck.search.predicate.AbstractPredicateSimpleAnalysisIT#noTokenAfterAnalysis` too, there is an indexed document containing the stopword we're looking for. See `org.hibernate.search.integrationtest.backend.tck.search.predicate.AbstractPredicateSimpleAnalysisIT.DataSet#analyzedFieldValues`.

If there really is a problem with the existing test, I'd recommend improving it rather than creating a new one. Backend tests are notoriously slow to start...

If there's no problem with the existing test, we can simply drop this PR :) The best code is the one you don't have to write!

fax4ever:
You're right we already cover this case with this test! Let's close this PR!

PR #2485, opened 2021-02-09
HSEARCH-4153 Automatic loading of entities passed by type + id in PojoIndexingPlan

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4153

~Based on #2482 , which should be merged first.~ => Done


fax4ever:
I have a shallow review and it looks good! Merged thanks!

PR #2486, opened 2021-02-10
HSEARCH-3884 Support Regexp queries having a regex expression predicate

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3884

### DIFF
--- a/documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
+++ b/documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
@@ -563,11 +563,12 @@ include::{sourcedir}/org/hibernate/search/documentation/search/predicate/Predica
 [IMPORTANT]
 ====
 The provided regular expressions used in regexp predicates
-will never be normalized or transformed in any way,
-even if a normalizer/analyzer has been defined on the field.
+may be normalized or not, it depends on the backend, if a normalizer has been defined on the field.
+The same will never be tokenized
+even if a analyzer with a proper tokenizer has been defined on the field.
 
 This means that if a normalizer/analyzer has been defined on the field,
-the user needs to be aware about how this normalizer/analyzer processes the data
+the user sometimes needs to be aware about how this normalizer/analyzer processes the data
 in order to provide a meaningful regular expression for using the predicate.
 ====
 
--- a/documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
+++ b/documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
@@ -547,6 +547,44 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
 either on a per-field basis with a call to `.boost(...)` just after `.field(...)`/`.fields(...)`
 or for the whole predicate with a call to `.boost(...)` after `.matching(...)`.
 
+[[search-dsl-predicate-regexp]]
+== `regexp`: match a regular expression pattern
+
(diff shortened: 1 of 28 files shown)


integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/RegexpPredicateBaseIT.java
```diff
@@ -0,0 +1,445 @@
...
+			return f.regexp().field( field0Path ).field( field1Path )
+					.matching( dataSet.values.matchingArg( matchingDocOrdinal ) );
+		}
+
+		@Override
+		protected String predicateNameInErrorMessage() {
+			return "predicate:regexp";
+		}
+	}
+}
```
yrodiere, line 445: These tests are only using very exotic regular expressions, bascially just literals. That's fine, because these tests are not about regexps but all the other features. But what about testing actual regexps?

Please add some tests with a few *real* regular expressions. You can copy `org.hibernate.search.integrationtest.backend.tck.search.predicate.WildcardPredicateSpecificsIT`, which by the way may have some tests that are relevant for the regexp predicate too, and add a few test methods.

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -547,6 +547,30 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+====
+The provided regular expressions used in regexp predicates
+will never be normalized or transformed in any way,
+even if a normalizer/analyzer has been defined on the field.
+
+This means that if a normalizer/analyzer has been defined on the field,
+the user needs to be aware about how this normalizer/analyzer processes the data
+in order to provide a meaningful regular expression for using the predicate.
+====
+
```
yrodiere, line 573: Please follow the same format as the documentation of other predicates. See for examples the sections "Targeting multiple fields" or "Other options" for the wildcard predicate.

yrodiere (changes requested):
Thanks! My comments below. In short, the implementation seems great, I just think we need more tests and documentation.

fax4ever:
Thank you for the review @yrodiere!
I don't know if we are done or there is something else to do.

fax4ever:
I ran a full build to check if have the same behaviour on all supported Elasticsearch versions: https://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-2486/6.

yrodiere (commented):
Thanks! Looks great, I'm just not sure the documentation about the behavior with analyzed/normalized fields is clear enough. I added a suggestion. Feel free to merge after applying it!

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -547,6 +547,45 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java[tags=regexp]
+----
+====
+
+[IMPORTANT]
+====
+The provided regular expressions used in regexp predicates
+may be normalized or not, it depends on the backend, if a normalizer has been defined on the field.
+The same will never be tokenized
```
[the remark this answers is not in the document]
fax4ever: I don't know if it is correct. With `the same` here I was trying to refer to the subject of the other sentence, that is `The provided regular expressions`.

fax4ever:
Merged thanks!

PR #2491, opened 2021-02-15
HSEARCH-4157 Raise the timeout on test

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4157


yrodiere:
Uh, 1 hour is a bit extreme. If commits are really that slow, I'd say Hibernate Search is not even useful on MacOS. How long are those commits taking exactly?

yrodiere:
Merged, thanks.

PR #2497, opened 2021-02-18
HSEARCH-4159 Hibernate Search doesn't properly execute the backend's preStop() operation on shutdown

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4159


fax4ever:
Merged thanks!

PR #2501, opened 2021-02-22
HSEARCH-4164 Fix transient failure in AbstractAutomaticIndexingMultiAssociationIT

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4164

I wasn't able to reproduce the issue. However, I was able to imagine a situation in which it can theoretically happen:

* Transaction 1 happens in an application thread, creating entity instance A1.
* The `BatchingExecutor` receives the event, and delegates to `LocalHeapQueueProcessor`.
* `LocalHeapQueueProcessor` starts a batch, opens a session, loads A1, indexes it, and ends the batch.
* Meanwhile, transaction 2 happens in an application thread, updating entity instance A1.
* The `BatchingExecutor` receives the event *before it noticed that the event queue was empty.
  As a result it doesn't call `LocalHeapQueueProcessor.complete()` (which would release the session), but it starts another batch.
* `LocalHeapQueueProcessor` re-uses the session from the previous batch.
  It starts a batch, opens a session, tries to loads A1, but it's already in the first-level cache.
  So it reindexes A1 with the same value as in the previous batch, incorrectly.

Simply clearing the session between each batch should resolve the problem.


yrodiere:
Merging, since there isn't much to review :) Let's see if it works.

PR #2507, opened 2021-02-25
HSEARCH-4172 Ignore relative ordering of works on different documents in BackendMock

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4172

Should also fix https://hibernate.atlassian.net/browse/HSEARCH-4164

### DIFF
--- a/util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/rule/BackendMock.java
+++ b/util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/rule/BackendMock.java
@@ -16,6 +16,7 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
+import java.util.stream.Stream;
 
 import org.hibernate.search.engine.backend.common.DocumentReference;
 import org.hibernate.search.engine.backend.spi.BackendBuildContext;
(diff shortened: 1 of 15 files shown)


util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/rule/BackendMock.java
```diff
@@ -421,9 +402,13 @@ public DocumentWorkCallListContext discarded() {
 			log.debugf( "Expecting %d works to be discarded", works.size() );
 			works.stream()
 					.map( work -> new DocumentWorkCall( indexName, DocumentWorkCall.WorkPhase.DISCARD, work ) )
-					.forEach( expectationConsumer );
+					.forEach( this::expect );
 			return this;
 		}
+
+		private void expect(DocumentWorkCall call) {
+			backendBehavior().getDocumentWorkCalls( call.documentKey() ).expectInOrder( call );
```
fax4ever, line 410: Does still it make sense to have a method named `expectInOrder`?

fax4ever (commented):
It looks nice, I've just a doubt.

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/rule/BackendMock.java
yrodiere: `expectInOrder`/`expectInAnyOrder` are defined in `CallQueue`, which is ised for for other purposes, in particular schema expectations where we need `expectInAnyOrder`.

So yes, it makes sense.

fax4ever: OK +1 to merge it

PR #2508, opened 2021-02-25
HSEARCH-4170 Upgrade to Elasticsearch 7.11

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4170

Full build here: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search-personal-yoann/detail/HSEARCH-4170/9/pipeline

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -204,7 +204,7 @@ The following profiles are available:
  * `elasticsearch-7.7` for 7.7
  * `elasticsearch-7.8` for 7.8 to 7.9
  * `elasticsearch-7.10` for 7.10 (the default)
- * `elasticsearch-7.11` for 7.11+
+ * `elasticsearch-7.11` for 7.11+ ([not open-source](https://opensource.org/node/1099))
 
 A list of available versions for `test.elasticsearch.connection.version` can be found on
 [Maven Central](https://search.maven.org/search?q=g:org.elasticsearch%20AND%20a:elasticsearch&core=gav).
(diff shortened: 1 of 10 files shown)


gsmet:
Isn't 7.11 the first non-Open Source version?

Do we really want to do that and make a non Open Source software our default setup? I know the low level client is still Open Source but I'm not sure it's a good idea to continue supporting non-OSS versions of Elasticsearch (and in any case, I wouldn't make it the default test setup).

gsmet:
My rationale is that if no Open Source tools support 7.11+, at some point, they might have to reconsider things.

yrodiere:
> Isn't 7.11 the first non-Open Source version?

It is.

> Do we really want to do that and make a non Open Source software our default setup?

Well, I mainly want to test it now that it's out. As to making it the default... I guess I can revert that. I'll add a commit.

> My rationale is that if no Open Source tools support 7.11+, at some point, they might have to reconsider things.

In an ideal world, yes. I don't share your optimism, though.

I could be persuaded not to advertise 7.11 support, but I don't think not supporting it at all would be a good thing for Hibernate Search.

For now I'll make 7.10 the default for testing and I will add warnings in the documentation about 7.11 not being open-source.

yrodiere:
> For now I'll make 7.10 the default for testing and I will add warnings in the documentation about 7.11 not being open-source.

Done. Does it look like a good compromise, @gsmet?

gsmet (approved):

PR #2512, opened 2021-03-02
HSEARCH-4132 Use entity change capture SPI to send entity change events into a "synthetic" table

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4132

### DIFF
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/outbox/CustomRoutingBridge.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/outbox/CustomRoutingBridge.java
@@ -1,0 +1,56 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.integrationtest.mapper.orm.outbox;
+
+import org.hibernate.search.mapper.pojo.bridge.RoutingBridge;
+import org.hibernate.search.mapper.pojo.bridge.binding.RoutingBindingContext;
+import org.hibernate.search.mapper.pojo.bridge.mapping.programmatic.RoutingBinder;
+import org.hibernate.search.mapper.pojo.bridge.runtime.RoutingBridgeRouteContext;
+import org.hibernate.search.mapper.pojo.route.DocumentRoutes;
+
+public class CustomRoutingBridge implements RoutingBridge<RoutedIndexedEntity> {
+
+	public static final int NUMBER_OF_SHARD = 7;
+
+	@Override
+	public void route(DocumentRoutes routes, Object entityIdentifier, RoutedIndexedEntity indexedEntity,
+			RoutingBridgeRouteContext context) {
+		String text = indexedEntity.getText();
+		char selector;
+
+		if ( text == null || text.trim().isEmpty() ) {
+			selector = 'z';
+		}
+		else {
+			selector = text.charAt( 0 );
+		}
+
+		// this is not a really great way to define the route id,
+		// we're using here just for testing
+		int shardId = selector % NUMBER_OF_SHARD;
+		routes.addRoute().routingKey( shardId + "" );
+	}
+
+	@Override
+	public void previousRoutes(DocumentRoutes routes, Object entityIdentifier, RoutedIndexedEntity indexedEntity,
+			RoutingBridgeRouteContext context) {
+		for ( int i = 0; i < 7; i++ ) {
+			// we simply don’t know what the previous route id of the book was,
+			// so we tell Hibernate Search to follow all possible routes
+			routes.addRoute().routingKey( i + "" );
+		}
+	}
+
+	public static final class CustomRoutingBinder implements RoutingBinder {
+
+		@Override
+		public void bind(RoutingBindingContext context) {
+			context.dependencies().use( "text" );
+			context.bridge( RoutedIndexedEntity.class, new CustomRoutingBridge() );
+		}
+	}
+}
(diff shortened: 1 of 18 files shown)


fax4ever:
Run a full build: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2512/4/pipeline

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/outbox/impl/OutboxAdditionalJaxbMappingProducer.java
```diff
@@ -59,20 +57,15 @@
...
 	public Collection<MappingDocument> produceAdditionalMappings(final MetadataImplementor metadata,
 			IndexView jandexIndex, final MappingBinder mappingBinder, final MetadataBuildingContext buildingContext) {
 		ServiceRegistry serviceRegistry = metadata.getMetadataBuildingOptions().getServiceRegistry();
 		ConfigurationService service = serviceRegistry.getService( ConfigurationService.class );
-		Boolean outboxEnabled = FILL_OUTBOX_TABLE.get( ConfigurationPropertySource.fromMap( service.getSettings() ) );
 
-		if ( !outboxEnabled ) {
+		Object customIndexingStrategy = service.getSettings().get( HibernateOrmMapperSettings.AUTOMATIC_INDEXING_STRATEGY );
+		if ( customIndexingStrategy == null || !OutboxTableAutomaticIndexingStrategy.class.getName().equals(
+				customIndexingStrategy ) ) {
```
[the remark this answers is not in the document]
yrodiere: https://hibernate.atlassian.net/browse/HSEARCH-4182

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/outbox/CustomRoutingBridge.java
```diff
@@ -0,0 +1,56 @@
...
+			selector = 'z';
+		}
+		else {
+			selector = text.charAt( 0 );
+		}
+
+		// this is not a really great way to define the route id,
+		// we're using here just for testing
+		int shardId = selector % NUMBER_OF_SHARD;
+		routes.addRoute().routingKey( shardId + "" );
```
yrodiere, line 35: Sharding and the number of shards should not be a concern here. Those are configured on the backend side.

This would work just as well (especially for tests), and it's much simpler:

```suggestion
		routes.addRoute().routingKey( indexedEntity.getSomeEnumProperty().name() );
```

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/outbox/CustomRoutingBridge.java
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	@Override
+	public void previousRoutes(DocumentRoutes routes, Object entityIdentifier, RoutedIndexedEntity indexedEntity,
+			RoutingBridgeRouteContext context) {
+		for ( int i = 0; i < 7; i++ ) {
+			// we simply don’t know what the previous route id of the book was,
+			// so we tell Hibernate Search to follow all possible routes
+			routes.addRoute().routingKey( i + "" );
```
yrodiere, line 44: Same here, you really should use some enum property. Sharding is backend concern.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/route/DocumentRouteDescriptor.java
```diff
@@ -21,6 +21,10 @@ private DocumentRouteDescriptor(String routingKey) {
 		this.routingKey = routingKey;
 	}
 
+	public String routingKey() {
+		return routingKey;
+	}
+
```
yrodiere, line 27: I don't know why you moved this up here, but for the sake of understanding each other I'll clarify why I put it down there in the first place.

I generally try to order methods this way: start with implemented and overridden methods, from the highest classes/interfaces in the hierarchy, then new methods declared by this class.

That's not a rule, but, well, next time you see this and feel the urge to change it, you'll know it wasn't random ;)

yrodiere (changes requested):
Thanks! Looks good overall. I have a few comments.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/route/DocumentRouteDescriptor.java
fax4ever: I used to put hashCode, equals and toString after other methods... sorry if I changed it :P 
Going to revert it...

fax4ever:
Thank you @yrodiere for the review.
I think I addressed all the comments.

fax4ever:
Full build: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2512/8/

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/outbox/impl/OutboxEvent.java
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	private Integer id;
+	private String entityName;
+	private String serializedId;
+	private byte[] serializedRoutes;
+	private Type type;
+
+	@Transient
+	private Object identifier;
```
yrodiere, line 29: I missed that in my original review, but can we make sure the name of these properties are straightforward?

* There's no point saying the values are "serialized", as that's mostly a matter of encoding. We don't say `integerCount` for a property representing a count that happens to be encoded as an integer, so there's no reason to say `serializedId` for a property reprensenting an ID that happens to be encoded as a serialized string.
* Since there's an even id and an entity ID, we should probably make it clear that a given property is about the entity, and another about the even.

So maybe something like this?

```suggestion
	private Integer id;
	private Type type;
	
	private String entityName;
	private String entityId;

	private byte[] documentRoutes;

	@Transient
	private Object originalEntityId;
```

yrodiere (commented):
Looks good to me, thanks. Maybe just one last change for the properties of `OutboxEvent`? It's unlikely we will change that later, so better get it right from the start :)

fax4ever:
Thank you! Updated the contribution...

yrodiere (approved):

fax4ever:
Thank you!

PR #2513, opened 2021-03-02
HSEARCH-4173 Multi-index Elasticsearch query fails (sometimes silently) when field is present in only one index

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4173

We'll need to backport this to 6.0. It shouldn't affect many people, only those who target multiple indexes in the same query. But it's rather bad for the few affected people.


yrodiere:
Full build: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2513/2/pipeline

fax4ever (approved):
The code looks good.
There are few test failures.
+1 to merge as soon the CI passes!

yrodiere:
Merged, thanks!

PR #2519, opened 2021-03-05
HSEARCH-4184 A contained entity with only irrelevant changes may mistakenly trigger reindexing

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4184


fax4ever:
It looks good! Merged, thanks!

PR #2521, opened 2021-03-09
HSEARCH-4134 Implement background polling service for outbox events

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4134

Draft: we need to implement (and to test) a rational retry strategy for indexing events that goes wrong.


fax4ever:
I also wanted to test failures on entity loading, as we did for the mass indexing.
The problem is that here the loaded classes are out of our (test) control, since they are: `OutboxEvent` and `OutboxEventRetry`.
I think that there will be a lot to fix to do here, so let's discuss this missing test too...

fax4ever:
Here is a full build: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2521/10/

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/outbox/impl/OutboxTableIndexerExecutor.java
```diff
@@ -63,11 +63,23 @@ public void start() {
 	}
 
 	private void run() {
-		try ( Session session = mapping.sessionFactory().openSession() ) {
-			processOutboxRetrialsEntities( session );
+		if ( mapping.sessionFactory().isClosed() ) {
```
yrodiere: I'm a bit surprised. Hibernate Search should shut down when `SessionFactoryObserver#sessionFactoryClosing` is closed, so *before* the session factory is actually closed.
When Hibernate Search gets closed, the automatic indexing stategy's `stop` method should be called, which should stop any event processing *before* the session factory is actually closed.

So you shouldn't have such problem... unless the `stop` method doesn't actually stop processing. And I think that's the case here, because of the loop you have in the event processor (`while (!events.isEmpty())`).

When you're done addressing my other comments, please come back to this one and try reverting this change. I don't think it will be necessary anymore.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/logging/impl/Log.java
```diff
@@ -294,4 +296,41 @@ SearchException foundMultipleEntitiesForDocumentId(String entityName, String doc
 	@Message(id = ID_OFFSET + 43, value = "Outbox-generated entity mapping: %1$s")
 	void outboxGeneratedEntityMapping(String xmlMappingDefinition);
 
+	@LogMessage(level = INFO)
+	@Message(id = ID_OFFSET + 44, value = "Session factory is closed. Probably Hibernate Search is shutting down.")
```
yrodiere, line 300:
```suggestion
	@Message(id = ID_OFFSET + 44, value = "Session factory is closed. Hibernate Search is probably shutting down.")
```

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/logging/impl/Log.java
```diff
@@ -294,4 +296,41 @@ SearchException foundMultipleEntitiesForDocumentId(String entityName, String doc
...
+	void sessionFactoryIsClosedOnOutboxProcessing();
+
+	@LogMessage(level = ERROR)
+	@Message(id = ID_OFFSET + 45, value = "There are some failure on processing outbox retrials entities.")
```
yrodiere, line 304: Hum ok, but what are those failures?
I think you should not log this, you should forward the failures to the background failure handler instead.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/logging/impl/Log.java
```diff
@@ -294,4 +296,41 @@ SearchException foundMultipleEntitiesForDocumentId(String entityName, String doc
...
+	void failureOnProcessingOutboxRetrials();
+
+	@LogMessage(level = ERROR)
+	@Message(id = ID_OFFSET + 46, value = "There are some failure on processing outbox entities.")
```
yrodiere, line 308: Hum ok, but what are those failures?
I think you should not log this, you should forward the failures to the background failure handler instead.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/logging/impl/Log.java
```diff
@@ -294,4 +296,41 @@ SearchException foundMultipleEntitiesForDocumentId(String entityName, String doc
...
+	void failureOnProcessingOutbox();
+
+	@LogMessage(level = ERROR)
+	@Message(id = ID_OFFSET + 47, value = "There are some failure on saving outbox retrials entities.")
```
yrodiere, line 312: Hum ok, but what are those failures?
I think you should not log this, you should forward the failures to the background failure handler instead.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/logging/impl/Log.java
```diff
@@ -294,4 +296,41 @@ SearchException foundMultipleEntitiesForDocumentId(String entityName, String doc
...
+	void failureOnSaveOutboxRetrials();
+
+	@LogMessage(level = ERROR)
+	@Message(id = ID_OFFSET + 48, value = "There are some failure on delete outbox entities.")
```
yrodiere, line 316: Hum ok, but what are those failures?
I think you should not log this, you should forward the failures to the background failure handler instead.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/logging/impl/Log.java
```diff
@@ -294,4 +296,41 @@ SearchException foundMultipleEntitiesForDocumentId(String entityName, String doc
...
+	void failureOnDeleteOutbox();
+
+	@LogMessage(level = ERROR)
+	@Message(id = ID_OFFSET + 49, value = "There are some failure on update outbox retrials entities.")
```
yrodiere, line 320: Hum ok, but what are those failures?
I think you should not log this, you should forward the failures to the background failure handler instead.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/logging/impl/Log.java
```diff
@@ -294,4 +296,41 @@ SearchException foundMultipleEntitiesForDocumentId(String entityName, String doc
...
+	void failureOnUpdateOutboxRetrials();
+
+	@LogMessage(level = WARN)
+	@Message(id = ID_OFFSET + 50, value = "Max '%1$s' retries exhausted to process the event '%2$s'." +
```
yrodiere, line 324: I think you should not log this, you should forward the failures to the background failure handler instead.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/outbox/impl/OutboxEventBackgroundExecutor.java
```diff
@@ -48,7 +48,14 @@ public OutboxEventBackgroundExecutor(AutomaticIndexingMappingContext mapping, Sc
...
 
 	public void start() {
-		executor.scheduleAtFixedRate( this::run, 0, pollingInterval, TimeUnit.MILLISECONDS );
+		executor.scheduleAtFixedRate( () -> {
+			try {
+				run();
+			}
+			catch (Throwable throwable) {
+				log.failureOnOutboxBackgroundProcessing( throwable );
+			}
```
yrodiere, line 57: FYI if you switch to `SingletonTask`, you won't need this anymore, it does the reporting for you, and it correctly uses the background failure handler.

yrodiere (changes requested):
Thanks. Here you go, the first batch of comments ;)

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/cfg/HibernateOrmMapperSettings.java
```diff
@@ -83,6 +83,20 @@ private HibernateOrmMapperSettings() {
...
+	 * <p>
+	 * Whether the all relevant indexing events present on the outbox table should be processed and deleted from
+	 * the same table by background processes.
+	 * <p>
+	 * Expects a Boolean value such as {@code true} or {@code false},
+	 * or a string that can be parsed to such Boolean value.
+	 * <p>
+	 * Defaults to {@link Defaults#AUTOMATIC_INDEXING_PROCESS_OUTBOX_TABLE}.
+	 */
+	public static final String AUTOMATIC_INDEXING_PROCESS_OUTBOX_TABLE = PREFIX + Radicals.AUTOMATIC_INDEXING_PROCESS_OUTBOX_TABLE;
```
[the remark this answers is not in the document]
fax4ever: Good idea.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/outbox/impl/OutboxTableIndexerExecutor.java
```diff
@@ -0,0 +1,154 @@
...
+		this.mapping = mapping;
+		this.executor = executor;
+	}
+
+	public void start() {
+		executor.scheduleAtFixedRate( this::run, 0, 8, TimeUnit.MILLISECONDS );
+	}
+
+	public CompletableFuture<?> stop() {
+		executor.shutdown();
```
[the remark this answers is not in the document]
fax4ever: this call is not to dispose the executor resource, but to return something similar to `BatchingExecutor#completition`.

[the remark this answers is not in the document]
yrodiere: The point of returning a `CompletableFuture<?>` is to be non-blocking, so this is not ideal.
Anyway, using `SingletonTask` as I suggested above should make this problem irrelevant, since `SingletonTask` can give you a `CompletableFuture<?>` through its `completion()` method.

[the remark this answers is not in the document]
fax4ever: right! But I just said that to explain the reason why I did it.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/outbox/impl/OutboxTableIndexerExecutor.java
```diff
@@ -0,0 +1,154 @@
...
+
+	private void run() {
+		try ( Session session = mapping.sessionFactory().openSession() ) {
+			List<OutboxEvent> events = findOutboxEntries( session );
+
+			while ( !events.isEmpty() ) {
+				AutomaticIndexingQueueEventProcessingPlan plan = mapping.createIndexingQueueEventProcessingPlan(
+						session );
+
+				events = findOutboxEntries( session );
```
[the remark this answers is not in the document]
fax4ever: it is already solved by a subsequent commit

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/outbox/impl/OutboxTableIndexerExecutor.java
```diff
@@ -0,0 +1,154 @@
...
+				for ( OutboxEvent event : events ) {
+					DocumentRoutesDescriptor routes = getRoutes( event );
+
+					switch ( event.getType() ) {
+						case ADD:
+							plan.add( event.getEntityName(), event.getEntityId(), routes );
```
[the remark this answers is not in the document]
fax4ever: Right! Going to rescope it...

fax4ever:
@yrodiere thank you for the extensive review.
I'm working on the last three comments.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/outboxtable/OutboxTableAutomaticIndexingStrategyEdgeIT.java
```diff
@@ -0,0 +1,173 @@
...
+					}
+
+					context.add( index + "", b -> b.field( "indexedField", "indexed value: " + index ) );
+				}
+				context.createdThenExecuted();
+			} );
+			backendMock.verifyExpectationsMet();
+		}
+	}
+
```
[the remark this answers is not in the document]
fax4ever: I moved the eventual consistency check for events in wrong order to a [dedicated issue](https://hibernate.atlassian.net/browse/HSEARCH-4202).

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/outbox/impl/OutboxTableIndexerExecutor.java
fax4ever: Addressing the other commits the loop in not there anymore, anyway the issue still seems to persist.

fax4ever:
Run [full build](https://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-2521/18/) mainly to test different SQL environment.

fax4ever:
Run [full build](https://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-2521/21/)

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/outbox/impl/OutboxEventBackgroundExecutor.java
```diff
@@ -0,0 +1,283 @@
...
+					List<OutboxEvent> outboxRetries = findOutboxRetries( session );
+					if ( !outboxRetries.isEmpty() ) {
+						// Process the event retries
+						OutboxEventProcessingPlan eventProcessing = new OutboxEventProcessingPlan(
+								mapping, session, outboxRetries );
+						eventProcessing.processEvents();
+						deleteProcessedOutboxRetries( session, outboxRetries, eventProcessing.collectFailedEvents() );
+						updateUnprocessedOutboxRetries( session, eventProcessing );
+					}
+				} );
```
yrodiere, line 102: With just a few changes, you can remove this block of code along with a few methods, then use the same code for retries and initial events, and all the tests will still pass.

I pushed a commit to demonstrate that, please keep it.

These changes also ensure that a retry event always has an ID higher than the initial event, which could prove useful.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/outbox/impl/OutboxEventBase.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.mapper.orm.outbox.impl;
+
+public interface OutboxEventBase {
```
yrodiere, line 9: This interface can be removed.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/outbox/impl/OutboxEventBackgroundExecutor.java
```diff
@@ -0,0 +1,283 @@
...
+	}
+
+	private class HibernateOrmOutboxWorker implements SingletonTask.Worker {
+
+		@Override
+		public CompletableFuture<?> work() {
+			if ( mapping.sessionFactory().isClosed() ) {
+				log.sessionFactoryIsClosedOnOutboxProcessing();
+				return CompletableFuture.completedFuture( null );
+			}
```
yrodiere, line 88: Tests still pass if I remove this block of code. Please do?

As I explained in another comment, I don't expect this method to be called when `mapping.sessionFactory().isClosed()` returns true.

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/rule/BackendWorkThreadingExpectations.java
```diff
@@ -47,7 +47,7 @@ public void awaitIndexingAssertions(ThrowingRunnable assertions) {
 		else {
 			await().pollDelay( Duration.ZERO )
 					.pollInterval( iterative( duration -> duration.multipliedBy( 2 ), Duration.ofMillis( 5 ) ) )
-					.atMost( Duration.ofSeconds( 5 ) )
+					.atMost( Duration.ofSeconds( 15 ) )
```
yrodiere, line 50: I'm curious, why was this necessary exactly? A specific test?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/logging/impl/Log.java
```diff
@@ -294,4 +294,8 @@ SearchException foundMultipleEntitiesForDocumentId(String entityName, String doc
 	@Message(id = ID_OFFSET + 43, value = "Outbox-generated entity mapping: %1$s")
 	void outboxGeneratedEntityMapping(String xmlMappingDefinition);
 
+	@LogMessage(level = INFO)
+	@Message(id = ID_OFFSET + 44, value = "Session factory is closed. Hibernate Search is probably shutting down.")
+	void sessionFactoryIsClosedOnOutboxProcessing();
```
yrodiere, line 299: This can probably be removed after you address my other comments

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/automaticindexing/impl/AutomaticIndexingQueueEventProcessingPlanImpl.java
```diff
@@ -45,4 +46,17 @@ public void delete(String entityName, String serializedId, DocumentRoutesDescrip
...
+
+	@Override
+	public String toSerializedId(String entityName, Object identifier) {
+		return delegate.toSerializedId( entityName, identifier );
+	}
+
+	@Override
+	public EntityReference entityReference(String entityName, String serializedId) {
+		PojoIndexingQueueEventProcessingPlan.EntityReferenceInfo refInfo =
+				delegate.entityReference( entityName, serializedId );
```
yrodiere, line 58: Instead of using an intermediary representation (`EntityReferenceInfo`), you could pass the `entityReferenceFactory` to `delegate.entityReference()` so that it returns the right type of reference right away.

That's how it's done everywhere else, at least :)

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/outbox/impl/OutboxEventBackgroundExecutor.java
```diff
@@ -0,0 +1,283 @@
...
+		);
+	}
+
+	public void start() {
+		status.set( Status.STARTED );
+		processingTask.ensureScheduled();
+	}
+
+	public CompletableFuture<?> completion() {
+		status.set( Status.STOPPED );
```
yrodiere, line 73: Maybe rename this method to `preStop`, if you're going to do this?

yrodiere (changes requested):
We're getting there :) Here are a few more comments, nothing critical. I also pushed a commit (see my comments).

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/outbox/impl/OutboxEventBackgroundExecutor.java
```diff
@@ -0,0 +1,283 @@
...
+		query.setParameter( "ids", ids );
+		query.executeUpdate();
+
+		session.flush();
+		session.clear();
+	}
+
+	private static OutboxEvent mergeOutboxEvents(Session session, OutboxEventReference reference,
+			List<OutboxEvent> events) {
+		if ( events.isEmpty() ) {
```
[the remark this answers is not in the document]
fax4ever: it seems to be solved by your commit

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/outbox/impl/OutboxEventProcessingPlan.java
```diff
@@ -0,0 +1,144 @@
...
+		this.events = events;
+		this.eventsIds = new ArrayList<>( events.size() );
+	}
+
+	List<Integer> processEvents() {
+		try {
+			addEventsToThePlan();
+			reportBackendResult( Futures.unwrappedExceptionGet( processingPlan.executeAndReport() ) );
+		}
+		catch (Throwable throwable) {
```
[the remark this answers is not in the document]
fax4ever: Yes I agree too. I'm fixing...

integrationtest/mapper/orm-spring/src/test/java/org/hibernate/search/integrationtest/spring/outbox/TransactionOutboxIT.java
```diff
@@ -0,0 +1,115 @@
...
+	public BackendMock backendMock;
+
+	@Autowired
+	private HelperService helperService;
+
+	@Test
+	public void persist() {
+		Integer id = 1;
+
+		// Check that outer transaction data is processed.
```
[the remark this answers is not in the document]
fax4ever: Right! It was a copy and paste mistake.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/outboxtable/OutboxTableAutomaticIndexingStrategyBaseIT.java
```diff
@@ -0,0 +1,54 @@
...
+		// Just execute all automatic indexing tests
+		"org.hibernate.search.integrationtest.mapper.orm.automaticindexing..*",
+		// ... except these tests that just cannot work with the outbox table strategy
+		// > Synchronization strategies can only be used with the "session" automatic indexing strategy
+		"!org.hibernate.search.integrationtest.mapper.orm.automaticindexing.session.AutomaticIndexingSynchronizationStrategyIT",
+		// > Sending events outside of transactions, during a flush, doesn't work for some reason;
+		//   entities are only visible from other sessions after the original session is closed.
+		"!org.hibernate.search.integrationtest.mapper.orm.automaticindexing.session.AutomaticIndexingOutOfTransactionIT",
+		// > session.getStatistics().getEntityCount() returns different (higher) values with the outbox table strategy
+		"!org.hibernate.search.integrationtest.mapper.orm.automaticindexing.bridge.AutomaticIndexingBridgeExplicitReindexingFunctionalIT",
```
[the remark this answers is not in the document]
fax4ever: I prefer create a new issue, in order to not increase the size of the scope of this one.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/outboxtable/OutboxTableAutomaticIndexingStrategyBaseIT.java
```diff
@@ -0,0 +1,54 @@
...
+		// > already executed with the outbox table strategy
+		"!org.hibernate.search.integrationtest.mapper.orm.automaticindexing.outboxtable..*"
+})
+public class OutboxTableAutomaticIndexingStrategyBaseIT {
```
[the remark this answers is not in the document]
fax4ever: Right. Let's uniforme them, using `outbox polling` (long version) or just `outbox` everywhere.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/outboxtable/OutboxTableAutomaticIndexingStrategyLifecycleIT.java
```diff
@@ -0,0 +1,209 @@
...
+		// verify some entities have not been processed
+		assertThat( outboxEntries ).isNotEmpty();
+		sessionFactory.close();
+
+		// process the entities restarting Search:
+		startSearchWithOutboxPolling();
+		BackendMock.DocumentWorkCallListContext context = backendMock.expectWorks( IndexedEntity.NAME );
+		for ( OutboxEvent event : outboxEntries ) {
+			context.add( event.getEntityId(), b -> b.field( "indexedField", "value for the field" ) );
+		}
```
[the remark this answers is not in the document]
fax4ever: If I try to remove the `resetExpectations` and the code above, the test doesn't work anymore.
Going to move this before I start Search.

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/rule/BackendWorkThreadingExpectations.java
fax4ever: There are some random failing tests. Most of the time: `AutomaticIndexingPolymorphicInverseSideAssociationIT` or `AutomaticIndexingEmbeddedBridgeIT`.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/outbox/impl/OutboxEventBackgroundExecutor.java
fax4ever: Running `OutboxTableAutomaticIndexingStrategyBaseIT` we got several of these calls.
Screenshot 2021-03-29 at 14 19 33

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/logging/impl/Log.java
fax4ever: I don't think we can address the related comment.

fax4ever:
Thanks for the second review @yrodiere.
The second series of comment have been handled (I think).
If there is anything else to improve, I'll be happy to do that.

fax4ever:
Another [full build](https://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-2521/).

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/rule/BackendWorkThreadingExpectations.java
yrodiere: That's quite odd.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/outbox/impl/OutboxEventBackgroundExecutor.java
yrodiere: Indeed. I see there are stacktraces if I remove this code, though the tests don't seem to fail.

From what I can see, the session factory marks itself as closed before it calls `sessionFactoryClosing` on its observers, which leads to all queries throwing exceptions.

I opened https://hibernate.atlassian.net/browse/HHH-14541. Please add a comment referencing it? We really need to fix that bug, because this check won't save us if the session factory gets closed when we're right in the middle of an execution.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/outboxtable/OutboxTableAutomaticIndexingStrategyBaseIT.java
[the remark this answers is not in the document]
yrodiere: Duplicating tests isn't a great solution... but I guess it will do.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/outboxtable/OutboxTableAutomaticIndexingStrategyLifecycleIT.java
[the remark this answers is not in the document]
yrodiere: If you also remove `backendMock.verifyExpectationsMet();` in `startSearchWithoutOutboxPolling`, you can remove `resetExpectations()` and this hack, and the test passes.

yrodiere (commented):
@fax4ever Thanks, LGTM. I'd add a comment on the `sessionFactory.isClosed()` hack, and I'd change `OutboxPollingAutomaticIndexingStrategyLifecycleIT` to avoid re-creating expectations (see my comments above), but feel free to merge once you're done.

fax4ever:
Thank you @yrodiere.
I addressed the last two things...
Going to merge it as soon as we get some positive feedback from the CI.

yrodiere:
@fax4ever Thanks!

PR #2522, opened 2021-03-10
HSEARCH-4173 Multi-index Elasticsearch query fails (sometimes silently) when field is present in only one index

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4173

Follow-up on #2513 ; as mentioned on [the ticket](https://hibernate.atlassian.net/browse/HSEARCH-4173?focusedCommentId=107928), I forgot to handle implicit nested predicates, which for some reason were using some copy/pasted code to do the same thing as explicit nested predicate. This PR fixes that.

Note: This should be backported to 6.0; see #2523


fax4ever:
It looks good! Merged thanks!

PR #2532, opened 2021-03-17
HSEARCH-2589 Introduce the terms predicate

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-2589

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/ElasticsearchStringIndexFieldTypeOptionsStep.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/ElasticsearchStringIndexFieldTypeOptionsStep.java
@@ -188,7 +188,7 @@ class ElasticsearchStringIndexFieldTypeOptionsStep
 			builder.queryElementFactory( PredicateTypeKeys.PHRASE, new ElasticsearchTextPhrasePredicate.Factory() );
 			builder.queryElementFactory( PredicateTypeKeys.WILDCARD, new ElasticsearchTextWildcardPredicate.Factory() );
 			builder.queryElementFactory( PredicateTypeKeys.REGEXP, new ElasticsearchTextRegexpPredicate.Factory() );
-			builder.queryElementFactory( PredicateTypeKeys.TERMS, new ElasticsearchTermsPredicate.Factory( codec ) );
+			builder.queryElementFactory( PredicateTypeKeys.TERMS, new ElasticsearchTermsPredicate.Factory<>( codec ) );
 			builder.queryElementFactory( PredicateTypeKeys.SIMPLE_QUERY_STRING,
 					new ElasticsearchSimpleQueryStringPredicateBuilderFieldState.Factory() );
 		}
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/ElasticsearchStringIndexFieldTypeOptionsStep.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/ElasticsearchStringIndexFieldTypeOptionsStep.java
@@ -22,6 +22,7 @@ import org.hibernate.search.backend.elasticsearch.search.projection.impl.Project
 import org.hibernate.search.backend.elasticsearch.search.sort.impl.SortTypeKeys;
 import org.hibernate.search.backend.elasticsearch.types.codec.impl.ElasticsearchStringFieldCodec;
 import org.hibernate.search.backend.elasticsearch.types.predicate.impl.ElasticsearchSimpleQueryStringPredicateBuilderFieldState;
+import org.hibernate.search.backend.elasticsearch.types.predicate.impl.ElasticsearchTermsPredicate;
 import org.hibernate.search.backend.elasticsearch.types.predicate.impl.ElasticsearchTextMatchPredicate;
 import org.hibernate.search.backend.elasticsearch.types.predicate.impl.ElasticsearchTextPhrasePredicate;
 import org.hibernate.search.backend.elasticsearch.types.predicate.impl.ElasticsearchTextRegexpPredicate;
@@ -187,6 +188,7 @@ class ElasticsearchStringIndexFieldTypeOptionsStep
 			builder.queryElementFactory( PredicateTypeKeys.PHRASE, new ElasticsearchTextPhrasePredicate.Factory() );
 			builder.queryElementFactory( PredicateTypeKeys.WILDCARD, new ElasticsearchTextWildcardPredicate.Factory() );
 			builder.queryElementFactory( PredicateTypeKeys.REGEXP, new ElasticsearchTextRegexpPredicate.Factory() );
+			builder.queryElementFactory( PredicateTypeKeys.TERMS, new ElasticsearchTermsPredicate.Factory( codec ) );
 			builder.queryElementFactory( PredicateTypeKeys.SIMPLE_QUERY_STRING,
 					new ElasticsearchSimpleQueryStringPredicateBuilderFieldState.Factory() );
 		}
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/predicate/impl/ElasticsearchTermsPredicate.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/predicate/impl/ElasticsearchTermsPredicate.java
@@ -7,6 +7,7 @@
 package org.hibernate.search.backend.elasticsearch.types.predicate.impl;
 
 import java.lang.invoke.MethodHandles;
+import java.util.Collection;
 
 import org.hibernate.search.backend.elasticsearch.gson.impl.JsonAccessor;
 import org.hibernate.search.backend.elasticsearch.gson.impl.JsonArrayAccessor;
@@ -125,15 +126,15 @@ public class ElasticsearchTermsPredicate extends AbstractElasticsearchSingleFiel
 		}
 
 		@Override
-		public void matchingAny(Object term, Object... terms) {
+		public void matchingAny(Collection<?> terms) {
 			allMatch = false;
-			fillTerms( term, terms );
+			fillTerms( terms );
 		}
 
 		@Override
-		public void matchingAll(Object term, Object... terms) {
+		public void matchingAll(Collection<?> terms) {
 			allMatch = true;
-			fillTerms( term, terms );
+			fillTerms( terms );
 		}
 
 		@Override
@@ -141,22 +142,22 @@ public class ElasticsearchTermsPredicate extends AbstractElasticsearchSingleFiel
 			return new ElasticsearchTermsPredicate( this );
 		}
 
-		private void fillTerms(Object term, Object[] terms) {
-			if ( terms == null || terms.length == 0 ) {
-				this.term = encode( term );
+		private void fillTerms(Collection<?> terms) {
+			if ( terms.size() == 1 ) {
+				this.term = encode( terms.iterator().next() );
 				this.terms = null;
 				return;
 			}
 
 			this.term = null;
-			this.terms = encode( term, terms );
+			this.terms = encode( terms );
 		}
 
-		private JsonElement[] encode(Object term, Object[] terms) {
-			JsonElement[] result = new JsonElement[terms.length + 1];
-			result[0] = encode( term );
-			for ( int i = 0; i < terms.length; i++ ) {
-				result[i + 1] = encode( terms[i] );
+		private JsonElement[] encode(Collection<?> terms) {
+			JsonElement[] result = new JsonElement[terms.size()];
+			int i = 0;
+			for ( Object term : terms ) {
+				result[i++] = encode( term );
 			}
 
 			return result;
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/predicate/impl/ElasticsearchTermsPredicate.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/predicate/impl/ElasticsearchTermsPredicate.java
@@ -111,7 +111,7 @@ public class ElasticsearchTermsPredicate extends AbstractElasticsearchSingleFiel
 
 	private static class Builder<F> extends AbstractBuilder implements TermsPredicateBuilder {
 
-		private final DslConverter<?, F> dslConverter;
+		private final ElasticsearchSearchValueFieldContext<F> field;
 		private final ElasticsearchFieldCodec<F> codec;
 
 		private JsonElement term;
@@ -121,20 +121,20 @@ public class ElasticsearchTermsPredicate extends AbstractElasticsearchSingleFiel
 		private Builder(ElasticsearchFieldCodec<F> codec, ElasticsearchSearchContext searchContext,
 				ElasticsearchSearchValueFieldContext<F> field) {
 			super( searchContext, field );
-			this.dslConverter = field.type().dslConverter( ValueConvert.NO );
+			this.field = field;
 			this.codec = codec;
 		}
 
 		@Override
-		public void matchingAny(Collection<?> terms) {
+		public void matchingAny(Collection<?> terms, ValueConvert convert) {
 			allMatch = false;
-			fillTerms( terms );
+			fillTerms( terms, convert );
 		}
 
 		@Override
-		public void matchingAll(Collection<?> terms) {
+		public void matchingAll(Collection<?> terms, ValueConvert convert) {
 			allMatch = true;
-			fillTerms( terms );
+			fillTerms( terms, convert );
 		}
 
 		@Override
@@ -142,28 +142,30 @@ public class ElasticsearchTermsPredicate extends AbstractElasticsearchSingleFiel
 			return new ElasticsearchTermsPredicate( this );
 		}
 
-		private void fillTerms(Collection<?> terms) {
+		private void fillTerms(Collection<?> terms, ValueConvert convert) {
+			DslConverter<?, F> dslConverter = field.type().dslConverter( convert );
+
 			if ( terms.size() == 1 ) {
-				this.term = encode( terms.iterator().next() );
+				this.term = encode( terms.iterator().next(), dslConverter );
 				this.terms = null;
 				return;
 			}
 
 			this.term = null;
-			this.terms = encode( terms );
+			this.terms = encode( terms, dslConverter );
 		}
 
-		private JsonElement[] encode(Collection<?> terms) {
+		private JsonElement[] encode(Collection<?> terms, DslConverter<?, F> dslConverter) {
 			JsonElement[] result = new JsonElement[terms.size()];
 			int i = 0;
 			for ( Object term : terms ) {
-				result[i++] = encode( term );
+				result[i++] = encode( term, dslConverter );
 			}
 
 			return result;
 		}
 
-		private JsonElement encode(Object term) {
+		private JsonElement encode(Object term, DslConverter<?, F> dslConverter) {
 			try {
 				F converted = dslConverter.convertUnknown( term, searchContext.toDocumentFieldValueConvertContext() );
 				return codec.encode( converted );
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/predicate/impl/ElasticsearchTermsPredicate.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/predicate/impl/ElasticsearchTermsPredicate.java
@@ -47,7 +47,7 @@ public class ElasticsearchTermsPredicate extends AbstractElasticsearchSingleFiel
 	private final JsonElement[] terms;
 	private boolean allMatch;
 
-	public ElasticsearchTermsPredicate(Builder builder) {
+	public ElasticsearchTermsPredicate(Builder<?> builder) {
 		super( builder );
 		this.term = builder.term;
 		this.terms = builder.terms;
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/predicate/impl/ElasticsearchTermsPredicate.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/predicate/impl/ElasticsearchTermsPredicate.java
@@ -1,0 +1,177 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.elasticsearch.types.predicate.impl;
+
+import java.lang.invoke.MethodHandles;
+
+import org.hibernate.search.backend.elasticsearch.gson.impl.JsonAccessor;
+import org.hibernate.search.backend.elasticsearch.gson.impl.JsonArrayAccessor;
+import org.hibernate.search.backend.elasticsearch.gson.impl.JsonObjectAccessor;
+import org.hibernate.search.backend.elasticsearch.logging.impl.Log;
+import org.hibernate.search.backend.elasticsearch.search.impl.AbstractElasticsearchCodecAwareSearchValueFieldQueryElementFactory;
+import org.hibernate.search.backend.elasticsearch.search.impl.ElasticsearchSearchContext;
+import org.hibernate.search.backend.elasticsearch.search.impl.ElasticsearchSearchValueFieldContext;
+import org.hibernate.search.backend.elasticsearch.search.predicate.impl.AbstractElasticsearchSingleFieldPredicate;
+import org.hibernate.search.backend.elasticsearch.search.predicate.impl.PredicateRequestContext;
+import org.hibernate.search.backend.elasticsearch.types.codec.impl.ElasticsearchFieldCodec;
+import org.hibernate.search.engine.backend.types.converter.spi.DslConverter;
+import org.hibernate.search.engine.reporting.spi.EventContexts;
+import org.hibernate.search.engine.search.common.ValueConvert;
+import org.hibernate.search.engine.search.predicate.SearchPredicate;
+import org.hibernate.search.engine.search.predicate.spi.TermsPredicateBuilder;
+import org.hibernate.search.util.common.logging.impl.LoggerFactory;
+
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+
+public class ElasticsearchTermsPredicate extends AbstractElasticsearchSingleFieldPredicate {
+
+	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
+
+	private static final JsonObjectAccessor TERMS_ACCESSOR = JsonAccessor.root().property( "terms" ).asObject();
+	private static final JsonObjectAccessor TERM_ACCESSOR = JsonAccessor.root().property( "term" ).asObject();
+
+	// for boolean query:
+	private static final JsonObjectAccessor BOOL_ACCESSOR = JsonAccessor.root().property( "bool" ).asObject();
+	private static final JsonArrayAccessor MUST_ACCESSOR = JsonAccessor.root().property( "must" ).asArray();
+
+	private static final JsonAccessor<JsonElement> VALUE_ACCESSOR = JsonAccessor.root().property( "value" );
+
+	private final JsonElement term;
+	private final JsonElement[] terms;
+	private boolean allMatch;
+
+	public ElasticsearchTermsPredicate(Builder builder) {
+		super( builder );
+		this.term = builder.term;
+		this.terms = builder.terms;
+		this.allMatch = builder.allMatch;
+	}
+
+	@Override
+	protected JsonObject doToJsonQuery(PredicateRequestContext context, JsonObject outerObject,
+			JsonObject innerObject) {
+
+		// single term
+		if ( term != null ) {
+			VALUE_ACCESSOR.set( innerObject, term );
+			JsonObject middleObject = new JsonObject();
+			middleObject.add( absoluteFieldPath, innerObject );
+			TERM_ACCESSOR.set( outerObject, middleObject );
+			return outerObject;
+		}
+
+		// multiple terms in OR
+		if ( !allMatch ) {
+			JsonArray jsonArray = new JsonArray( terms.length );
+			for ( JsonElement element : terms ) {
+				jsonArray.add( element );
+			}
+
+			innerObject.add( absoluteFieldPath, jsonArray );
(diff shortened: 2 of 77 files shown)


fax4ever:
Here is a full-build: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2532/3/pipeline

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/predicate/impl/ElasticsearchTermsPredicate.java
```diff
@@ -0,0 +1,177 @@
...
+				jsonArray.add( element );
+			}
+
+			innerObject.add( absoluteFieldPath, jsonArray );
+			TERMS_ACCESSOR.set( outerObject, innerObject );
+			return outerObject;
+		}
+
+		// multiple terms in AND
+		// using a boolean query here:
```
yrodiere, line 82: Can't you just use `minimum_should_match_field`? That seems more efficient, especially if there are many terms.

See https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-terms-set-query.html#terms-set-top-level-params

In fact, if there is a large number of terms, a boolean query simply won't work: you can only stuff a certain number of clauses into a boolean query.

Could you add a test with many terms? By "many" I mean more than what the maximum number of clauses allowed in a boolean query (`BooleanQuery.maxClauseCount`); that's definitely where this predicate will shine.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/predicate/impl/ElasticsearchTermsPredicate.java
```diff
@@ -0,0 +1,177 @@
...
+		if ( !allMatch ) {
+			JsonArray jsonArray = new JsonArray( terms.length );
+			for ( JsonElement element : terms ) {
+				jsonArray.add( element );
+			}
+
+			innerObject.add( absoluteFieldPath, jsonArray );
+			TERMS_ACCESSOR.set( outerObject, innerObject );
+			return outerObject;
+		}
```
yrodiere, line 79: You can merge those two blocks by leveraging `VALUE_ACCESSOR.add( innerObject, term )`. It will create an array automatically if there already is a term.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/predicate/impl/LuceneNumericTermsPredicate.java
```diff
@@ -0,0 +1,107 @@
...
+
+			if ( !allMatch ) {
+				return codec.getDomain().createSetQuery( absoluteFieldPath, terms );
+			}
+
+			BooleanQuery.Builder builder = new BooleanQuery.Builder();
+			for ( E termItem : terms ) {
+				Query query = codec.getDomain().createExactQuery( absoluteFieldPath, termItem );
+				builder.add( query, BooleanClause.Occur.MUST );
+			}
```
yrodiere, line 88: That's not great: there is a (relatively low) limit to how many clauses you can add to a boolean query, and one of the purposes of this new `terms` predicate is to overcome that limitation.

But from what I can see there isn't any alternative, unless we want to roll out our own query... 

Ok, let's go with this. Make sure to test queries with many terms (`> 1024`) for both cases, though (`matchingAll` and `matchingAny`). I suspect you will get an exception for `matchingAll` and will have to add a test exception through `TckBackendFeatures`.

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -630,6 +630,57 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
 either on a per-field basis with a call to `.boost(...)` just after `.field(...)`/`.fields(...)`
 or for the whole predicate with a call to `.boost(...)` after `.matching(...)`.
 
+[[search-dsl-predicate-terms]]
+== `terms`: match a series of term
```
yrodiere, line 634: Maybe "match a set of terms"? Series means "sequence" and might imply order, which we definitely don't support here.

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -630,6 +630,57 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+
+The `terms` predicate matches documents for which a given field contains some terms, any or all of them.
+
+With `matchingAny` we require that at least one of the provided terms matches.
+That is equivalent to having an <<search-dsl-predicate-boolean-or>> applied to the series of single-term predicates.
```
yrodiere, line 639: This implies the boolean OR will contain `match` queries, in which case it's not exactly equivalent,  since we don't analyze terms.

What about this:

```suggestion
Functionally, this is somewhat similar to a <<search-dsl-predicate-boolean-or,boolean `OR`>> with one <<search-dsl-predicate-match,`match`>> predicate per term, but the syntax for a single `terms` predicate is more concise.

NOTE: `matchingAny` expects to be passed *terms*, not just any string. The given terms will not be analyzed/normalized. See <<search-dsl-predicate-terms-analysis>>.
```

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -630,6 +630,57 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+That is equivalent to having an <<search-dsl-predicate-boolean-and>> applied to the series of single-term predicates.
+
+.Matching all of the provided terms
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java[tags=terms-any]
+----
+====
+
```
yrodiere, line 659:
```suggestion

[[search-dsl-predicate-terms-analysis]]
```

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -630,6 +630,57 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+.Matching any of the provided terms
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java[tags=terms-any]
+----
+====
+
+With `matchingAll` we require that all of the provided terms match.
+That is equivalent to having an <<search-dsl-predicate-boolean-and>> applied to the series of single-term predicates.
```
yrodiere, line 650:
```suggestion
Functionally, this is somewhat similar to a <<search-dsl-predicate-boolean-and,boolean `AND`>> with one <<search-dsl-predicate-match,`match`>> predicate per term, but the syntax for a single `terms` predicate is more concise.

NOTE: `matchingAll` expects to be passed *terms*, not just any string. The given terms will not be analyzed/normalized. See <<search-dsl-predicate-terms-analysis>>.
```

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -630,6 +630,57 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+include::{sourcedir}/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java[tags=terms-any]
+----
+====
+
+=== Terms predicates and analysis
+
+NOTE: For an introduction to analysis and how to configure it,
+refer to the <<concepts-analysis>> section.
+
+Differently from other predicates, the provided terms arguments to `matchAny` or `matchAll`
```
yrodiere, line 665:
```suggestion
Differently from other predicates, the terms passed to `matchingAny()` or `matchingAll()`
```

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -630,6 +630,57 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+are never analyzed nor normalized.
```
yrodiere, line 666: Do you have a test for that? That would be nice, especially considering the inconsistent behavior of Elasticsearch.

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -630,6 +630,57 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+This means that in order to match the related documents,
+we need to provide the terms as they were written to the indexes.
```
yrodiere, line 668:
```suggestion
you need to provide the terms as they were written to the indexes.
```

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -630,6 +630,57 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+=== Targeting multiple fields
+
+Optionally, the predicate can target multiple fields.
+In that case, the predicate will match documents for which _any_ of the given fields matches.
+
+See <<search-dsl-predicate-common-multiple-fields>>.
+
+=== Other options
+
+* The score of a `terms` predicate is constant and equal to 1 by default,
```
yrodiere, line 679: Are you sure about this? Did you test it?

yrodiere (changes requested):
Thanks! It looks good overall, but I have a few comments.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/predicate/impl/ElasticsearchTermsPredicate.java
fax4ever: If we merge the two blocks, we would use the `TERMS_ACCESSOR` for both the single and the multiple arguments case. I think that if there exists a term predicate for the single argument it is better to keep using it for this case.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/predicate/impl/ElasticsearchTermsPredicate.java
fax4ever: Reading the documentation, it seems that there isn't a `minimum_should_match` option here. From my understanding `minimum_should_match_field` needs a field on the index to store the `minimum_should_match` value.
There is `minimum_should_match_script` but a script probably needs to be executed somehow, so I'm not sure that would be a great optimization.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/predicate/impl/LuceneNumericTermsPredicate.java
fax4ever: Queries with many terms wouldn't be supported in any case, see my comment above. At most I would check and validate the size of the terms. Going to do that.

fax4ever: Thinking better about that, since we have the same issue ( of validate the max clause count ), maybe it is better to move it to a different issue : https://hibernate.atlassian.net/browse/HSEARCH-4205.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/predicate/impl/ElasticsearchTermsPredicate.java
yrodiere: Right, I misread that part...

Supposedly `painless` scripts are pre-compiled before running the query... So I guess it could perform well?

Anyway, you're right, it seems doubtful this optimization would have a great impact. And considering we have that exact same limitation of accepting a limited number of terms for all `matchingAll` implementations (text and numbers, Lucene and Elasticsearch), I guess it doesn't hurt to keep that limitation here.

yrodiere: We'll need to document the limitation somewhere, though ;)

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/predicate/impl/LuceneNumericTermsPredicate.java
yrodiere: Queries with many terms (significantly more than 1024) should be supported for `matchAny` (when using `TermInSetQuery`). 

I don't remember if you already did, but please make sure you test at least that?

fax4ever: I'm adding now...

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
fax4ever: Is it not already tested by TermsPredicateBaseIT#ScoreIT?

fax4ever:
Thank you for the review @yrodiere.
I should have addressed all the comments here.

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -630,6 +630,71 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+====
+
+With `matchingAll` we require that all of the provided terms match.
+Functionally, this is somewhat similar to a <<search-dsl-predicate-boolean-and,boolean `AND`>> with one <<search-dsl-predicate-match,`match`>> predicate per term, but the syntax for a single `terms` predicate is more concise.
+
+NOTE: `matchingAll` expects to be passed *terms*, not just any string. The given terms will not be analyzed. See <<search-dsl-predicate-terms-analysis>>.
+
+[WARNING]
+====
+With `matchingAll` the given terms must not be more than 1024.
```
yrodiere, line 658: That deserves some more explanation.

```suggestion
By default, `matchingAll` will not accept more than 1024 terms.

It is possible to raise this limit through backend-specific configuration:

* For the Lucene backend, run this code when starting up your application: `org.apache.lucene.search.BooleanQuery.maxClauseCount = <your limit>;`
* For the Elasticsearch backend, see link:{elasticsearchDocUrl}/search-settings.html[the global setting `indices.query.bool.max_clause_count` in the Elasticsearch documentation].

However, keep in mind the limit is there for a reason: attempts to match very large numbers of terms will perform poorly and could lead to crashes.
```

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -630,6 +630,71 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+refer to the <<concepts-analysis>> section.
+
+Differently from other predicates, the terms passed to `matchingAny()` or `matchingAll()`
+are never analyzed nor **usually** normalized.
+If it is true, this means that in order to match the related documents,
+you need to provide the terms as they were written to the indexes.
+
+This is always true with the Lucene backend.
+For the Elasticsearch backend: terms are never analyzed nor normalized if an analyzer is defined on the field.
+While terms are normalized if a normalizer is defined on the field.
```
yrodiere, line 682: A sentence should not start with "while".

Regardless, could you please try to be consistent with existing documentation? We already have a similar warning for the wildcard predicate, and we used an `IMPORTANT` admonition there. So here we should have something like this:

```suggestion
[IMPORTANT]
====
If an analyzer has been defined on the field, the terms will not be analyzed nor normalized.

If a normalizer has been defined on the field:

* when using the Elasticsearch backend, the terms **will** be normalized.
* when using the Lucene backend, the terms **will not** be normalized.

For example, the term `Cat` could match `cat`
when targeting a field having a normalizer that applies a lowercase filter when indexing,
but only when using the Elasticsearch backend.
When using the Lucene backend, only the term `cat` could match `cat`.
====
```

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/predicate/impl/LuceneTextTermsPredicate.java
```diff
@@ -0,0 +1,114 @@
...
+				List<BytesRef> bytesRefs = terms.stream().map( BytesRef::new ).collect( Collectors.toList() );
+				return new TermInSetQuery( absoluteFieldPath, bytesRefs );
+			}
+
+			BooleanQuery.Builder builder = new BooleanQuery.Builder();
+			for ( String termItem : terms ) {
+				Query query = new TermQuery( new Term( absoluteFieldPath, termItem ) );
+				builder.add( query, BooleanClause.Occur.MUST );
+			}
+			return new ConstantScoreQuery( builder.build() );
```
yrodiere, line 97: Same remark as for the Elasticsearch predicate:

> So if someone calls `constantScore()` in the DSL... we will wrap the query in a constant score query twice?
> 
> Could you do something similar to `org.hibernate.search.backend.lucene.types.predicate.impl.LuceneExistsPredicate.AbstractBuilder#AbstractBuilder` instead?
> 
> Additionally, you could also add a safeguard in `applyConstantScore()` that calls a new `isAlwaysConstant()` method, so that it doesn't wrap the query unnecessarily.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/predicate/impl/LuceneNumericTermsPredicate.java
```diff
@@ -0,0 +1,108 @@
...
+			return new ConstantScoreQuery( builder.build() );
```
yrodiere, line 91: Same remark as for the Elasticsearch predicate:

> So if someone calls `constantScore()` in the DSL... we will wrap the query in a constant score query twice?
> 
> Could you do something similar to `org.hibernate.search.backend.lucene.types.predicate.impl.LuceneExistsPredicate.AbstractBuilder#AbstractBuilder` instead?
> 
> Additionally, you could also add a safeguard in `applyConstantScore()` that calls a new `isAlwaysConstant()` method, so that it doesn't wrap the query unnecessarily.

documentation/src/test/java/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java
```diff
@@ -746,6 +746,36 @@ public void regexp() {
...
 	}
 
+	@Test
+	public void terms_any() {
+		withinSearchSession( searchSession -> {
+			// tag::terms-any[]
+			List<Book> hits = searchSession.search( Book.class )
+					.where( f -> f.terms().field( "description" )
+							.matchingAny( "robot", "cab" ) )
+					.fetchHits( 20 );
```
yrodiere, line 756: This predicate will most likely be used on enum fields or embedded id fields.

With that in mind, could you change this example to target the `genre` field? That would be a hint that people don't need to provide values of the field type directly, they can provide values of the property type as well.

I know the other example cannot be changed so easily, so let's leave it as is.

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -630,6 +630,71 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+====
+
+.Matching all of the provided terms
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java[tags=terms-any]
+----
+====
+
```
yrodiere, line 668: Since you added support for argument conversion, could you please add a section named "Expected type of arguments"? You can pretty much copy/paste the one from the `match` predicate (with a few changes). See [here](https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#search-dsl-predicate-match-argument-type).

engine/src/main/java/org/hibernate/search/engine/logging/impl/Log.java
```diff
@@ -408,4 +408,9 @@ SearchException relativeNamedPredicateNameCannotContainDot(String relativeNamedP
 			+ " the required identifier must be a superclass of the actual identifier.")
 	SearchException wrongRequiredIdentifierType(@FormatWith(ClassFormatter.class) Class<?> requiredIdentifierType,
 			@FormatWith(ClassFormatter.class) Class<?> actualIdentifierType);
+
+	@Message(id = ID_OFFSET + 96, value = "Terms predicate with match all does not support '%1$s' different terms "
+			+ "as an argument. The max size of terms is '%2$s'.")
+	SearchException tooManyTermsForMatchAll(int size, int maxSize);
```
yrodiere, line 414: If you follow my other recommendations, this should be removed.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/TermsPredicateMatchingStep.java
```diff
@@ -0,0 +1,101 @@
...
+	 */
+	N matchingAny(Collection<?> terms, ValueConvert convert);
+
+	/**
+	 * Require at least one of the targeted fields to match <b>all</b> of the provided terms.
+	 *
+	 * @param firstTerm The (first) term to match.
+	 * @param otherTerms The others (optional) terms to match.
+	 * The signature of this method defines these parameter as a {@link Object},
+	 * but a specific type is expected depending on the targeted field and on the {@code dslConverter} parameter.
```
yrodiere, line 67:
```suggestion
	 * but a specific type is expected depending on the targeted field.
```

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/TermsPredicateMatchingStep.java
```diff
@@ -0,0 +1,101 @@
...
+	default N matchingAny(Object firstTerm, Object ... otherTerms) {
+		return matchingAny( CollectionHelper.asList( firstTerm, otherTerms ) );
+	}
+
+	/**
+	 * Require at least one of the targeted fields to match <b>any</b> of the provided terms.
+	 *
+	 * @param terms The terms to match.
+	 * The signature of this method defines this parameter as a {@link Collection} of any {@link Object},
+	 * but a specific type is expected depending on the targeted field and on the {@code dslConverter} parameter.
```
yrodiere, line 40:
```suggestion
	 * but a specific type is expected depending on the targeted field.
```

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/TermsPredicateMatchingStep.java
```diff
@@ -0,0 +1,101 @@
...
+ */
+public interface TermsPredicateMatchingStep<N extends TermsPredicateOptionsStep<?>> {
+
+	/**
+	 * Require at least one of the targeted fields to match <b>any</b> of the provided terms.
+	 *
+	 * @param firstTerm The (first) term to match.
+	 * @param otherTerms The others (optional) terms to match.
+	 * The signature of this method defines these parameter as a {@link Object},
+	 * but a specific type is expected depending on the targeted field and on the {@code dslConverter} parameter.
```
yrodiere, line 27:
```suggestion
	 * but a specific type is expected depending on the targeted field.
```

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/TermsPredicateMatchingStep.java
```diff
@@ -0,0 +1,101 @@
...
+	default N matchingAny(Collection<?> terms) {
+		return matchingAny( terms, ValueConvert.YES );
+	}
+
+	/**
+	 * Require at least one of the targeted fields to match <b>any</b> of the provided terms.
+	 *
+	 * @param terms The terms to match.
+	 * The signature of this method defines this parameter as a {@link Collection} of any {@link Object},
+	 * but a specific type is expected depending on the targeted field and on the {@code dslConverter} parameter.
```
yrodiere, line 53:
```suggestion
	 * but a specific type is expected depending on the targeted field and on the {@code convert} parameter.
```

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/TermsPredicateMatchingStep.java
```diff
@@ -0,0 +1,101 @@
...
+	default N matchingAll(Object firstTerm, Object ... otherTerms) {
+		return matchingAll( CollectionHelper.asList( firstTerm, otherTerms ) );
+	}
+
+	/**
+	 * Require at least one of the targeted fields to match <b>all</b> of the provided terms.
+	 *
+	 * @param terms The terms to match.
+	 * The signature of this method defines this parameter as a {@link Collection} of any {@link Object},
+	 * but a specific type is expected depending on the targeted field and on the {@code dslConverter} parameter.
```
yrodiere, line 80:
```suggestion
	 * but a specific type is expected depending on the targeted field.
```

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/TermsPredicateMatchingStep.java
```diff
@@ -0,0 +1,101 @@
...
+	default N matchingAll(Collection<?> terms) {
+		return matchingAll( terms, ValueConvert.YES );
+	}
+
+	/**
+	 * Require at least one of the targeted fields to match <b>all</b> of the provided terms.
+	 *
+	 * @param terms The terms to match.
+	 * The signature of this method defines this parameter as a {@link Collection} of any {@link Object},
+	 * but a specific type is expected depending on the targeted field and on the {@code dslConverter} parameter.
```
yrodiere, line 93:
```suggestion
	 * but a specific type is expected depending on the targeted field and on the {@code convert} parameter.
```

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/ElasticsearchStringIndexFieldTypeOptionsStep.java
```diff
@@ -187,6 +188,7 @@ public ElasticsearchStringIndexFieldTypeOptionsStep aggregable(Aggregable aggreg
 			builder.queryElementFactory( PredicateTypeKeys.PHRASE, new ElasticsearchTextPhrasePredicate.Factory() );
 			builder.queryElementFactory( PredicateTypeKeys.WILDCARD, new ElasticsearchTextWildcardPredicate.Factory() );
 			builder.queryElementFactory( PredicateTypeKeys.REGEXP, new ElasticsearchTextRegexpPredicate.Factory() );
+			builder.queryElementFactory( PredicateTypeKeys.TERMS, new ElasticsearchTermsPredicate.Factory( codec ) );
```
yrodiere, line 191: A diamond operator is missing here and in a few other places.

Please have a look at the sonar report: https://sonarcloud.io/project/issues?id=org.hibernate.search%3Ahibernate-search-parent&open=AXhowV6ZNXwoQfKpUHbp&pullRequest=2532&resolved=false&types=CODE_SMELL

yrodiere (dismissed):

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/impl/TermsPredicateFieldMoreStepImpl.java
```diff
@@ -0,0 +1,125 @@
...
+			}
+			return this;
+		}
+
+		private TermsPredicateOptionsStep<?> matchingAll(Collection<?> terms, ValueConvert convert) {
+			Contracts.assertNotNullNorEmpty( terms, "terms" );
+			Contracts.assertNotNull( convert, "convert" );
+
+			if ( terms.size() > BOOLEAN_QUERIES_MAX_CLAUSE_COUNT ) {
+				throw log.tooManyTermsForMatchAll( terms.size(), BOOLEAN_QUERIES_MAX_CLAUSE_COUNT );
```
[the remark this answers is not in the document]
fax4ever: You're right. We cannot control it. I dropped the entire commit.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/TermsPredicateSpecificsIT.java
```diff
@@ -0,0 +1,200 @@
...
+		);
+
+		StubMappingScope scope = index.createScope();
+
+		SearchQuery<DocumentReference> query = scope.query().where(
+				f -> f.terms().field( "multiValued" ).matchingAny( lotsOfTerms ) ).toQuery();
+
+		assertThatQuery( query )
+				.hasDocRefHitsAnyOrder( c -> c
+						.doc( index.typeName(), DOC_ID )
```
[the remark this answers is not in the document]
fax4ever: Right. I did that for the matchAll.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/ElasticsearchStringIndexFieldTypeOptionsStep.java
fax4ever: The other places are about the `Builder` parameter of the constructors of the predicates.
I'm fixing it, but I think we have the same issue on all predicates (both on ES and Lucene).

fax4ever:
Thanks @yrodiere I agreed and addressed (I think) all the comments here. Anything else to fix/improve?

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/TermsPredicateSpecificsIT.java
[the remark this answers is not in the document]
yrodiere: You forgot about this one?

yrodiere (commented):
Just one comment, feel free to merge. Thanks!

fax4ever:
@yrodiere thanks merged. I think that the comment has been addressed.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/TermsPredicateSpecificsIT.java
```diff
@@ -0,0 +1,188 @@
...
+
+	private final SimpleMappedIndex<IndexBinding> index = SimpleMappedIndex.of( IndexBinding::new );
+	private final List<String> lotsOfTerms = new ArrayList<>();
+
+	public TermsPredicateSpecificsIT() {
+		for ( int i = 0; i < LOT_OF_TERMS_SIZE; i++ ) {
+			if ( i == 7 ) {
+				lotsOfTerms.add( "term" + i );
+			}
+			lotsOfTerms.add( "wrong-term" + i );
```
fax4ever, line 49: see the change here, we match now only one term

PR #2539, opened 2021-04-05
HSEARCH-4130 Use docker containers for Elasticsearch in integration tests

fax4ever:


parents/integrationtest/pom.xml
```diff
@@ -140,33 +140,56 @@
...
                         <skip>${test.elasticsearch.run.skip}</skip>
-                        <flavour>${test.elasticsearch.run.flavour}</flavour>
-                        <clusterName>hsearchEsTestCluster</clusterName>
-                        <httpPort>9200</httpPort>
-                        <version>${test.elasticsearch.connection.version}</version>
-                        <timeout>90</timeout><!-- Make it work on our slow CI -->
-                        <!-- auto-create-index is necessary for non-OSS Elasticsearch -->
-                        <autoCreateIndex>true</autoCreateIndex>
-                        <environmentVariables>
-                            <JAVA_HOME>${test.elasticsearch.run.java_home}</JAVA_HOME>
```
yrodiere, line 156: So you're not using `test.elasticsearch.run.java_home` anymore. That makes sense, of course. But please finish removing it from the `README` and `Jenkinsfile`? Everything related to `elasticsearchJdkTool` in the Jenkinsfile, in particular, is no longer relevant.

parents/integrationtest/pom.xml
```diff
@@ -140,33 +140,56 @@
...
+                                <alias>elasticsearch</alias>
+                                <run>
+                                    <env>
+                                        <discovery.type>single-node</discovery.type>
+                                        <!-- fix Docker images for older versions -->
+                                        <xpack.security.enabled>false</xpack.security.enabled>
+                                    </env>
+                                    <ports>
+                                        <port>9200:9200</port>
+                                        <port>9300:9300</port>
```
yrodiere, line 160: I don't think exposing this port is necessary, as we only use the REST interface.

parents/integrationtest/pom.xml
```diff
@@ -140,33 +140,56 @@
...
+                          Useful to stop processes still running from a previously failed integration test run -->
+                        <allContainers>true</allContainers>
                     </configuration>
-                    <!-- Different executions from 2.x: the "start" goal has been renamed in version 5  -->
                     <executions>
                         <execution>
-                            <id>start-elasticsearch</id>
-                            <phase>pre-integration-test</phase>
+                            <id>docker-start</id>
+                            <phase>process-test-classes</phase>
```
yrodiere, line 185:
```suggestion
                            <phase>pre-integration-test</phase>
```

parents/integrationtest/pom.xml
```diff
@@ -140,33 +140,56 @@
...
-                        </environmentVariables>
+                        <images>
+                            <image>
+                                <name>docker.elastic.co/elasticsearch/elasticsearch:${test.elasticsearch.connection.version}</name>
+                                <alias>elasticsearch</alias>
+                                <run>
+                                    <env>
+                                        <discovery.type>single-node</discovery.type>
+                                        <!-- fix Docker images for older versions -->
+                                        <xpack.security.enabled>false</xpack.security.enabled>
```
yrodiere, line 156: Maybe also this, as it was enabled in all of our custom configuration files? Or do you have a reason to avoid this with containers?

```suggestion
                                        <xpack.security.enabled>false</xpack.security.enabled>
                                        <!-- Prevent swapping
                                              This may trigger warnings upon boot if the system is not correctly set up.
      See https://www.elastic.co/guide/en/elasticsearch/reference/7.5/setup-configuration-                                        memory.html#bootstrap-memory_lock
                                         -->
                                        <bootstrap.memory_lock>true</bootstrap.memory_lock>
```

yrodiere:
Started a full build here: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2539/3/pipeline

parents/integrationtest/pom.xml
```diff
@@ -316,51 +300,17 @@
...
-            <build>
-                <pluginManagement>
-                    <plugins>
-                        <plugin>
-                            <groupId>com.github.alexcojocaru</groupId>
-                            <artifactId>elasticsearch-maven-plugin</artifactId>
-                            <version>${version.com.github.alexcojocaru.elasticsearch.plugin}</version>
-                            <configuration>
-                                <pathConf>${project.build.directory}/elasticsearch-maven-plugin/6.0/configuration/</pathConf>
-                                <pathInitScript>${project.build.directory}/elasticsearch-maven-plugin/6.0/init/init.script</pathInitScript>
```
yrodiere, line 358: So we're no longer creating index templates on startup... I guess that's fine, as it makes the build simpler, and we don't really need the slowlog anyway: we should use performance tests for that.

The templates were also used to reduce the number of replicas, though. That's important in older version (such as 5.6) which default to 5 replicas, which doesn't make sense in a single-node setup. I'll start a full build to see if everything still works even with these default settings.

parents/integrationtest/pom.xml
```diff
@@ -185,45 +208,6 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
             </plugin>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-dependency-plugin</artifactId>
-                <executions>
-                    <execution>
-                        <id>unpack-integrationtest-sharedresources</id>
```
yrodiere, line 193: If you remove this, I think you can remove an (obsolete) comment in `documentation/pom.xml`:

```
<!-- Inherit execution unpack-shared-test-build-resources -->
```

pom.xml
```diff
@@ -477,11 +477,6 @@
         <surefire.environment>default</surefire.environment>
 
         <!-- Elasticsearch tests properties -->
-
-        <!-- Control how an Elasticsearch instance is run automatically during tests -->
-        <test.elasticsearch.run.skip>false</test.elasticsearch.run.skip>
```
yrodiere, line 482: I think this one is still used? Let's keep it here?

yrodiere (commented):
Thanks! Looks good overall, though I have a few comments

parents/integrationtest/pom.xml
fax4ever: Right, pre integration tests seems to be the right phase to trigger the docker start.
Anyway I'm asking why in Quarkus they use `process-test-classes`...
Probably there is some reason...

fax4ever: Maybe they have some non-IT tests that need a container...

yrodiere: Oh, don't get me started about this... They are basically hacking surefire to run integration tests...

fax4ever: I imagined that :)

pom.xml
fax4ever: Yes it is, in order to not change too much the CI configuration. Right. Let's keep it.

fax4ever:
Thanks. I think I made the changes, even if I don't see the `Outdated` labels on the comment. 
Anything else to improve/fix here?

fax4ever:
Run a full build: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2539/6/pipeline

Jenkinsfile
```diff
@@ -800,10 +795,6 @@ class EsLocalBuildEnvironment extends BuildEnvironment {
 	String jdkTool
```
yrodiere, line 795: Looks like this can be removed as well. Don't forget to remove it in the calls to `new EsLocalBuildEnvironment` as well.

yrodiere (commented):
Just one comment. Feel free to merge once it's addressed and the full build passes. Thanks.

fax4ever:
Run another full build (just to be sure it works): https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2539/9/pipeline

fax4ever:
Merged, thanks!

PR #2540, opened 2021-04-08
HSEARCH-3556 Configuration of the max window size in Elasticsearch

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3556

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
@@ -733,4 +733,11 @@ public interface Log extends BasicLogger {
 
 	@Message(id = ID_OFFSET + 139, value = "Unknown filter definition '%1$s'.")
 	SearchException unknownNamedPredicateForSearch(String name, @Param EventContext context);
+
+	@LogMessage(level = Level.WARN)
+	@Message(id = ID_OFFSET + 140, value = "A search query fetching all hits was requested," +
+			" but only '%2$s' hits were retrieved because the maximum result window size forces a limit of '%1$s'" +
+			" hits. Refer to Elasticsearch's 'max_result_window_size' setting for more information." )
+	void defaultedLimitedHits(Integer defaultLimit, long hitCount);
+
 }
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/lowlevel/index/settings/impl/IndexSettings.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/lowlevel/index/settings/impl/IndexSettings.java
@@ -26,6 +26,7 @@ import com.google.gson.annotations.SerializedName;
 public class IndexSettings {
 
 	public static final String MAX_RESULT_WINDOW_ATTRIBUTE = "max_result_window";
+	public static final int MAX_RESULT_WINDOW_DEFAULT = 10_000;
 
 	private Analysis analysis;
 
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/lowlevel/index/settings/impl/IndexSettings.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/lowlevel/index/settings/impl/IndexSettings.java
@@ -17,6 +17,7 @@ import org.hibernate.search.backend.elasticsearch.gson.impl.SerializeExtraProper
 import com.google.gson.GsonBuilder;
 import com.google.gson.JsonElement;
 import com.google.gson.annotations.JsonAdapter;
+import com.google.gson.annotations.SerializedName;
 
 /**
  * Settings for an Elasticsearch index.
@@ -24,7 +25,12 @@ import com.google.gson.annotations.JsonAdapter;
 @JsonAdapter(IndexSettingsJsonAdapterFactory.class)
 public class IndexSettings {
 
+	public static final String MAX_RESULT_WINDOW_ATTRIBUTE = "max_result_window";
+
 	private Analysis analysis;
+
+	@SerializedName(IndexSettings.MAX_RESULT_WINDOW_ATTRIBUTE)
+	private Integer maxResultWindow;
 
 	@SerializeExtraProperties
 	private Map<String, JsonElement> extraAttributes;
@@ -32,8 +38,9 @@ public class IndexSettings {
 	public IndexSettings() {
 	}
 
-	public IndexSettings(Analysis analysis, Map<String, JsonElement> extraAttributes) {
+	public IndexSettings(Analysis analysis, Integer maxResultWindow, Map<String, JsonElement> extraAttributes) {
 		this.analysis = analysis;
+		this.maxResultWindow = maxResultWindow;
 		this.extraAttributes = extraAttributes;
 	}
 
@@ -43,6 +50,10 @@ public class IndexSettings {
 
 	public void setAnalysis(Analysis analysis) {
 		this.analysis = analysis;
+	}
+
+	public Integer getMaxResultWindow() {
+		return maxResultWindow;
 	}
 
 	public Map<String, JsonElement> getExtraAttributes() {
@@ -72,6 +83,10 @@ public class IndexSettings {
 		}
 		else {
 			analysis.merge( overridingIndexSettings.analysis );
+		}
+
+		if ( overridingIndexSettings.maxResultWindow != null ) {
+			maxResultWindow = overridingIndexSettings.maxResultWindow;
 		}
 
 		extraAttributes = overridingIndexSettings.extraAttributes;
@@ -110,6 +125,6 @@ public class IndexSettings {
 		for ( String key : keysToRemove ) {
 			newExtraAttributes.remove( key );
 		}
-		return new IndexSettings( analysis, newExtraAttributes );
+		return new IndexSettings( analysis, maxResultWindow, newExtraAttributes );
 	}
 }
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/validation/impl/IndexSettingsValidator.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/validation/impl/IndexSettingsValidator.java
@@ -28,13 +28,15 @@ public class IndexSettingsValidator implements Validator<IndexSettings> {
 	@Override
 	public void validate(ValidationErrorCollector errorCollector, IndexSettings expected, IndexSettings actual) {
 		Analysis expectedAnalysis = expected.getAnalysis();
-		if ( expectedAnalysis == null ) {
-			// No expectation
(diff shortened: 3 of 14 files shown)


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/lowlevel/index/settings/impl/IndexSettings.java
```diff
@@ -17,23 +17,30 @@
...
  */
 @JsonAdapter(IndexSettingsJsonAdapterFactory.class)
 public class IndexSettings {
 
+	public static final String MAX_RESULT_WINDOW_ATTRIBUTE = "max_result_window";
+
 	private Analysis analysis;
 
+	@SerializedName(IndexSettings.MAX_RESULT_WINDOW_ATTRIBUTE)
+	private Integer maxResultWindow;
```
yrodiere, line 33: I think you forgot to update the `merge` method?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/validation/impl/IndexSettingsValidator.java
```diff
@@ -28,13 +28,14 @@
...
-			// No expectation
-			return;
+		if ( expectedAnalysis != null ) {
+			Analysis actualAnalysis = actual.getAnalysis();
+			validateAnalysisSettings( errorCollector, expectedAnalysis, actualAnalysis );
 		}
-		Analysis actualAnalysis = actual.getAnalysis();
 
-		validateAnalysisSettings( errorCollector, expectedAnalysis, actualAnalysis );
+		LeafValidators.EQUAL.validate( errorCollector, ValidationContextType.OTHER_INDEX_SETTINGS_ATTRIBUTE,
```
yrodiere, line 36: This should use `validateWithDefault`, I think? The settings returned by Elasticsearch will not necessarily include the value of `max_result_window_size`, in which case we should assume `10,000`.

One way to test this would be to create an index manually without specifying the `max_result_window_size`, then executing validation against that index. I'm fairly sure validation would fail with this code, but it should pass.

Probably a test to add to `ElasticsearchIndexSchemaManagerValidationOtherSettingsIT`?

documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
```diff
@@ -891,6 +891,27 @@ refer to the documentation:
...
+----
+The default for this property is 10000.
+
+[NOTE]
+====
+This value will be used to limit the returning hits size if no limit has been defined on the query by the user.
+In this case if Hibernate Search notices there are more results, a warning will be issued.
+====
+
 == Custom index settings
```
yrodiere, line 915: What happens if someone sets the max result window size through custom index settings? Do we take that into account? Is there a test for that?

documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
```diff
@@ -891,6 +891,27 @@ refer to the documentation:
...
+hibernate.search.backend.searching.max_result_window_size = 5000
+# To configure a specific index:
+hibernate.search.backend.indexes.<index name>.searching.max_result_window_size = 5000
+----
+The default for this property is 10000.
+
+[NOTE]
+====
+This value will be used to limit the returning hits size if no limit has been defined on the query by the user.
+In this case if Hibernate Search notices there are more results, a warning will be issued.
```
yrodiere, line 912:
```suggestion
In this case if Hibernate Search notices there are more results, a warning will be logged.
```

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/mapping/ElasticsearchTypeNameMappingSchemaIT.java
```diff
@@ -126,8 +128,17 @@ private JsonObject indexCreationPayload() {
...
 	}
+
+	private JsonObject getSettings() {
+		JsonObject settings = new JsonObject();
+		settings.add(
+				IndexSettings.MAX_RESULT_WINDOW_ATTRIBUTE,
+				new JsonPrimitive( ElasticsearchIndexSettings.Defaults.MAX_RESULT_WINDOW_SIZE )
+		);
+		return settings;
+	}
```
yrodiere, line 143: ... Why is this necessary? You shouldn't have to set the defaults, they are implicit.

util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/rule/TestElasticsearchClient.java
```diff
@@ -107,12 +109,23 @@ public boolean exists() {
...
+		public IndexClient deleteAndCreate(boolean withMaxResultWindow) {
+			TestElasticsearchClient.this.deleteAndCreateIndex( primaryIndexName, writeAlias, readAlias,
+					withMaxResultWindow );
 			return this;
 		}
 
 		public IndexClient deleteAndCreate(String settingsPath, String settings) {
-			JsonObject settingsAsJsonObject = buildStructuredSettings( settingsPath, settings );
+			// by default we add the withMaxResultWindow,
+			// so that we don't have any validation errors for this attribute
```
yrodiere, line 123: Again, we shouldn't have validation errors when the max result window is the default. The validation code should assume that `null` means `10,000`. I left a comment on the validation code with a solution.

util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/rule/TestElasticsearchClient.java
```diff
@@ -486,13 +489,26 @@ private JsonObject buildStructuredSettings(String settingsPath, String settings)
...
+				}
+
 				settingsJsonElement = parent;
 			}
 		}
 
 		return settingsJsonElement.getAsJsonObject();
 	}
 
+	private void addDefaultMaxResultWindow(JsonObject settingsJsonElement) {
```
yrodiere, line 505: Again, we shouldn't have validation errors when the max result window is the default. The validation code should assume that `null` means `10,000`. I left a comment on the validation code with a solution.

documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
yrodiere: Actually... Shouldn't we rely exclusively on custom index settings, and remove the dedicated property? I think it would make sense, don't you?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -733,4 +733,9 @@ SearchException customIndexSettingsJsonSyntaxErrors(String filePath, @Cause Exce
 
 	@Message(id = ID_OFFSET + 139, value = "Unknown filter definition '%1$s'.")
 	SearchException unknownNamedPredicateForSearch(String name, @Param EventContext context);
+
+	@LogMessage(level = Level.WARN)
+	@Message(id = ID_OFFSET + 140, value = "Apply the default limit '%1$s' to the result hits size of '%2$s'." )
```
yrodiere, line 738:
```suggestion
	@Message(id = ID_OFFSET + 140, value = "A search query fetching all hits was requested, but only '%2$s' hits were retrieved because the maximum result window size forces a limit of '%1$s' hits. Refer to Elasticsearch's 'max_result_window_size' setting for more information." )
```

yrodiere (commented):
A few comments... my main question is: do we really want a dedicated configuration property, when we could just retrieve the value of `max_result_window_size` from the custom settings?

fax4ever:
I drastically reduced the scope of this contribution without introducing a new property for the setting.
@yrodiere feel free to review it again... thanks!

fax4ever:
Here is a new full-build:https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2540/4/.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/lowlevel/index/settings/impl/IndexSettings.java
```diff
@@ -49,6 +52,15 @@ public void setAnalysis(Analysis analysis) {
 		return extraAttributes;
 	}
 
+	public Integer getMaxResultWindow() {
+		JsonElement maxResultWindowElement = extraAttributes.get( MAX_RESULT_WINDOW_ATTRIBUTE );
```
yrodiere, line 56: That's weird. Any reason to remove the more explicit `private Integer maxResultWindow`? The point of `extraAttributes` was to hold, well... extra attributes. The ones we don't care about.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/validation/impl/IndexSettingsValidator.java
```diff
@@ -16,7 +19,7 @@
 
 import com.google.gson.JsonElement;
 
-public class IndexSettingsValidator implements Validator<IndexSettings> {
+	public class IndexSettingsValidator implements Validator<IndexSettings> {
```
yrodiere, line 22: Unnecessary tab here.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/lowlevel/index/settings/impl/IndexSettings.java
yrodiere: Also, it seems to me this complicates `IndexSettingsValidator` unnecessarily.

util/internal/integrationtest/common/src/main/resources/log4j2.properties
```diff
@@ -14,7 +14,7 @@ logger.jboss.level=info
 
 ### Search logger
 logger.search.name=org.hibernate.search
-logger.search.level=debug
+logger.search.level=trace
```
yrodiere, line 17: ?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/lowlevel/index/settings/impl/IndexSettings.java
fax4ever: Ok I'm putting back the explicit field....

yrodiere (approved):
LGTM, feel free to merge once CI passes. Thanks!

yrodiere (approved):

fax4ever:
Merged, thanks!

PR #2544, opened 2021-04-13
HSEARCH-4211 Use commits instead of rollbacks in the mass indexer

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4211


fax4ever (commented):
The code looks good.
I have only one doubt, about the fact that now the interceptor may update the data, I think.
But probably you already know about that...

PR #2546, opened 2021-04-16
HSEARCH-4130 Raise the log level of ES container

fax4ever:

Follow-up of #2539 to raise the log level of the container to WARN.


yrodiere:
Merged, thanks

PR #2552, opened 2021-04-20
HSEARCH-4215 Add information about session activity.

wklaczynski:
Session activity information is missing. This is very important if the previously opened session was opened by another code fragment in the same transaction. Elsewhere, the same session is accessed without knowing if a session shutdown was programmatically performed before. 

https://hibernate.atlassian.net/browse/HSEARCH-4215

Use case:
```java
        String key = "some key";
        if (key != null && TransactionHelper.isInTx()) {
            session = SearchTransactionUtil.get(key);
        }

        if (session == null || !session.isOpen()) {
            SearchMapping mapping = createMapping(unit, type, owner);

            session = mapping.createSessionWithOptions().
                    loading((options) -> {
   .........
                    }).build();
            SearchTransactionUtil.put(key, session);
        }
        return session;                        
                        
```

### DIFF
--- a/mapper/javabean/src/main/java/org/hibernate/search/mapper/javabean/session/SearchSession.java
+++ b/mapper/javabean/src/main/java/org/hibernate/search/mapper/javabean/session/SearchSession.java
@@ -27,6 +27,12 @@ public interface SearchSession extends AutoCloseable {
 	void close();
 
 	/**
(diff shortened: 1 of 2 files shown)


yrodiere (approved):

yrodiere:
Merged, thanks

PR #2561, opened 2021-04-26
Reopen of HSEARCH-4130 Use docker for Elasticsearch IT

fax4ever:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -910,8 +910,21 @@ void runBuildOnNode(Closure body) {
 
 void runBuildOnNode(String label, Closure body) {
 	node( label ) {
-		timeout( [time: 1, unit: 'HOURS'], body )
+		pruneDockerContainers()
+        try {
+        	timeout( [time: 1, unit: 'HOURS'], body )
+        }
+        finally {
+        	pruneDockerContainers()
+        }
 	}
+}
+
+void pruneDockerContainers() {
+	sh 'docker container prune -f || true'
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -910,7 +910,19 @@ void runBuildOnNode(Closure body) {
 
 void runBuildOnNode(String label, Closure body) {
 	node( label ) {
-		timeout( [time: 1, unit: 'HOURS'], body )
+		sh 'docker container prune -f'
+        sh 'docker image prune -f'
+        sh 'docker network prune -f'
+        sh 'docker volume prune -f'
```
yrodiere, line 916: Please encapsulate this in a script, like it's done on Quarkus (where you probably took this from)? Especially if you need to run it twice...

yrodiere (changes requested):

Jenkinsfile
fax4ever: I would avoid to create a file for this. I could extract a method. What do you think about it?

yrodiere: I guess it could work, yes.

yrodiere: The build fails. The first failure is caused by Coveralls (it's down today, apparently), but the other one is caused by a docker prune command returning a non-zero status; I assume that happens when there is nothing to prune.

I'm starting to believe that handling this kind of edge cases would be easier in... say... a shell script? :)

yrodiere:
Merged, thanks

PR #2564, opened 2021-05-03
HSEARCH-4203 Improve the mapper-javabean loading API and common code

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4203
EDIT: Fixed https://hibernate.atlassian.net/browse/HSEARCH-4218 too, to address a test failure.

@fax4ever As we discussed, this is mostly refactoring, and tests are already there, so I don't think it's worth your time to review it.

However, I had to change APIs. Mostly for two reasons:

1. I moved entity loading configuration to the mapping, instead of the session. It makes much more sense, but that involves slightly different APIs (in particular so that we can still pass context from the session to the loaders).
2. I tried to look for ways to make interceptors more... accessible to a random users, and I failed completely. So I fell back to a different API that look a lot like what @wklaczynski originally proposed, with a few adjustments which (IMO) should make the API simple yet powerful.

@wklaczynski , could you try out this new API on your application, and tell me if you can still do everything you need? Basically, to configure loading, you will now need to:

* Add a loading strategy to each (root) entity type, using something similar to this:
  ```
  mapperBuilder.addEntityType( MyEntityType.class, c -> c
   			.massLoadingStrategy( new MyMassLoadingStrategy() ) );
  ```
* Implement `MassLoadingStrategy` in `MyMassLoadingStrategy`. Some hints:
  * You can use `MapMassLoadingStrategy` as an example.
  * If there are any resources (sessions, transactions, ...) you need to open for the whole duration of mass indexing, do it in `MassLoadingStrategy#createIdentifierLoader`/`MassLoadingStrategy#createEntityLoader`, and close the resources in the `close` method of the returned loaders.
  * If there are resources you need to open/close for each batch, you can do it with a `try...finally` block in the implementation of the `load` method of the loaders.
  * Entity types will be grouped together for mass indexing if they are part of the same entity hierarchy and their loading strategies are equal (according to `.equals(Object)`)


yrodiere:
@wklaczynski Gentle reminder that this would benefit from you trying it out :)

I can still help you with the upgrade until the end of the week. After that I'll be on PTO for one week, and I'll merge it when I come back if I don't hear any complaint by then.

wklaczynski:
The solution is good. Indeed, interception was unnecessary. 

But the problem is with type casting. I get an exception when compiling "incompatible types: bad return type in lambda expression". 

```java
	private static final Map<String, User> users = new LinkedHashMap<>();

	builder.addEntityType( User.class, (context) -> {
		context.massLoadingStrategy( new MapMassLoadingStrategy( users ) );
		
		context.selectionLoadingStrategy( (includedTypes, options) -> {
			return (identifiers, deadline) -> {
				List<User> result = identifiers.stream()
						.filter( users::containsKey )
						.map( users::get )
						.collect( Collectors.toList());

				//incompatible types;
				return result;
			};
		} );
	} );
```

wklaczynski:
Since there is no way to manage a grouping strategy, is there any other way to block entities to index by super type? 

```java

                //Super Grouping
		builder.addEntityType( Object.class, (context) -> {
			context.massLoadingStrategy( new MapMassLoadingStrategy( db ) );
		} );
		
		builder.addEntityType( User.class);

		builder.addEntityType( Partition.class);

		builder.addEntityType( Topic.class );
		
			db.put( "u1", new User( "u1", "John", "Smith", "London" ) );
			db.put( "u2", new User( "u2", "Martin", "Carter", "Paris" ) );

			db.put( "p1", new Partition( "p1", "Partition 1", users.get( "u1" ) ) );
			db.put( "p2", new Partition( "p2", "Partition 2", users.get( "u2" ) ) );
			db.put( "p3", new Partition( "p3", "Partition 3", users.get( "u2" ) ) );

			db.put( "t1", new Topic( "t1", "Topic 1", partitions.get( "p1" ) ) );
			db.put( "t2", new Topic( "t2", "Topic 2", partitions.get( "p1" ) ) );
			db.put( "t3", new Topic( "t3", "Topic 3", partitions.get( "p2" ) ) );
			db.put( "t4", new Topic( "t4", "Topic 4", partitions.get( "p3" ) ) );

			for ( int i = 0; i < 20; i++ ) {
				publications.put( "p1" + i, new Publication( "p1" + i, "Publication " + i + " in topic 1 ", (Topic) db.get( "t1" )) );
			}

			for ( int i = 0; i < 20; i++ ) {
				publications.put( "p2" + i, new Publication( "p2" + i, "Publication " + i + " in topic 2 ", (Topic) db.get( "t2" )) );
			}

			for ( int i = 0; i < 20; i++ ) {
				publications.put( "p3" + i, new Publication( "p3" + i, "Publication " + i + " in topic 3 ", (Topic) db.get( "t3" )) );
			}

			for ( int i = 0; i < 20; i++ ) {
				publications.put( "p4" + i, new Publication( "p4" + i, "Publication " + i + " in topic 4 ", (Topic) db.get( "t4" )) );
			}

			MassIndexer indexer = session.massIndexer( Object.class );

			MassIndexer monitor = indexer.monitor(new MassIndexingMonitor () {
				@Override
				public void documentsAdded(long increment) {
					System.out.println( "added " + increment );
				}

				@Override
				public void documentsBuilt(long increment) {
					System.out.println( "built " + increment );
				}

				@Override
				public void entitiesLoaded(long increment) {
					System.out.println( "loaded " + increment );
				}

				@Override
				public void addToTotalCount(long increment) {
					System.out.println( "total " + increment );
				}

				@Override
				public void indexingCompleted() {
					System.out.println( "completed" );
				}
			});
```
You could assign the same indexer to each type of entity. But even so, without affecting the grouping strategy, we can't group everything into a super type. Then it indexes each entity separately because the super type is not indexable. 

After all, the information from "MassIndexingMonitor" is completely useless, it is difficult to determine the totalcount of all separate groups to be indexed. 

In this case, instead of one totalcount, we have several and only when the previous group terminates. 

```
INFO: HCANN000001: Hibernate Commons Annotations {5.1.2.Final}
total 2
loaded 2
built 1
built 1
added 2
total 3
loaded 3
built 1
built 1
built 1
added 3
total 4
loaded 4
built 1
built 1
built 1
built 1
added 4
total 0
completed
```
It is now difficult to show the indexing progress bar for such indexer "MassIndexingMonitor". 

The previous solution with sharing a grouping strategy was not elegant, but it allows for some ideas. Maybe, since the grouping strategy is unavailable, then to define the group also pass what is in the "massIndexer (...)" parameters, so that he grouped by the types that the programmer really expects?

wklaczynski:
Maybe it would be easier to define super types? When a type has no strategy, the closest type is searched and implemented over the type that has an indexing strategy. 

```java
                //Super type or interface instance of (User, Dysponent, Agent, ...). 
		builder.addEntityType( IdentityType.class, (context) -> {
			context.massLoadingStrategy( new MapMassLoadingStrategy( db ) );
			context.selectionLoadingStrategy( (includedTypes, options) -> {
				return (identifiers, deadline) -> {
					List result = identifiers.stream()
							.filter( db::containsKey )
							.map( db::get )
					return result;
				};

			} );
		} );

                //Super type or interface instance of (Topic, Publication ...). 
		builder.addEntityType( MessageType.class, (context) -> {
			context.massLoadingStrategy( new MapMassLoadingStrategy( db ) );
			context.selectionLoadingStrategy( (includedTypes, options) -> {
				return (identifiers, deadline) -> {
					List result = identifiers.stream()
							.filter( db::containsKey )
							.map( db::get )
					return result;
				};

			} );
		} );


		builder.addEntityType( User.class );

		builder.addEntityType( Dysponent.class );

		builder.addEntityType( Agent.class );

		builder.addEntityType( Topic.class);

		builder.addEntityType( Publication.class);
```

yrodiere:
> But the problem is with type casting. I get an exception when compiling "incompatible types: bad return type in lambda expression".

Right, there is something wrong with the generics. I'll look into it.

> Maybe it would be easier to define super types? When a type has no strategy, the closest type is searched and implemented over the type that has an indexing strategy.

Yes, that's already how it works. However, since we apply the configurer once to each type, each type will have its own instance of the strategy. If you want to enable grouping, you need to implement `equals`/`hashCode` in your strategies, so that Hibernate Search knows it can group together two types with equal strategies.

See the javadoc of `org.hibernate.search.mapper.javabean.loading.SelectionLoadingStrategy` and `org.hibernate.search.mapper.javabean.loading.MassLoadingStrategy`.

yrodiere:
@wklaczynski I pushed an update. Your "incompatible types" problem should be gone.

As to grouping/inheritance, as I explained above it's just a matter of implementing equals/hashcode in your strategies.

Is everything good apart from that?

wklaczynski:
> @wklaczynski I pushed an update. Your "incompatible types" problem should be gone.
> 
> As to grouping/inheritance, as I explained above it's just a matter of implementing equals/hashcode in your strategies.
> 
> Is everything good apart from that?

Indeed it worked, even equals not necessary in the strategy.

yrodiere:
Fantastic, I'll merge it then. Thank you for your feedback! And of course for the original work :)

wklaczynski:
I checked the hierarchy. Works great. I can add a test later. 

```java
@Indexed
public class User implements IdentityType {....}

@Indexed
public class Topic {...}

@Indexed
public class Publication implements MessageType {...)

@Indexed
public class Partition implements MappedType {...}
......
public class AppDataMassLoadingStrategy<E, I> implements MassLoadingStrategy<E, I> {
..
	@Override
	public MassIdentifierLoader createIdentifierLoader(LoadingTypeGroup<E> includedTypes, MassIdentifierSink<I> sink,
			MassLoadingOptions options) {
		log.logf( Logger.Level.INFO, "Create mass indexing identifier loader for <%s> with types %s", name, includedTypes.toString() );

)

public class AppDataSelectionLoadingStrategy<E, I> implements SelectionLoadingStrategy<E> {
..
	@Override
	public SelectionEntityLoader<E> createEntityLoader(LoadingTypeGroup<E> includedTypes, SelectionLoadingOptions options) {
		log.logf( Logger.Level.INFO, "Create mass selection entity loader for <%s> with types %s", name, includedTypes.toString() );
..
	@Override
	public MassEntityLoader<I> createEntityLoader(LoadingTypeGroup<E> includedTypes, MassEntitySink<E> sink,
			MassLoadingOptions options) {
		log.logf( Logger.Level.INFO, "Create mass indexing entity loader for <%s> with types %s", name, includedTypes.toString() );
```

```java
  	       private static final Map<String, Object> db = new LinkedHashMap<>();


		builder.addEntityType( Object.class, (context) -> {
			context.massLoadingStrategy( new AppDataMassLoadingStrategy( db, "any type" ) );
			context.selectionLoadingStrategy( new AppDataSelectionLoadingStrategy( db, "any type" ) );
		} );

		builder.addEntityType( IdentityType.class, (context) -> {
			context.massLoadingStrategy( new AppDataMassLoadingStrategy( db, "identity type" ) );
			context.selectionLoadingStrategy( new AppDataSelectionLoadingStrategy( db, "identity type" ) );
		} );

		builder.addEntityType( MessageType.class, (context) -> {
			context.massLoadingStrategy( new AppDataMassLoadingStrategy( db, "message type" ) );
			context.selectionLoadingStrategy( new AppDataSelectionLoadingStrategy( db, "message type" ) );
		} );

		builder.addEntityType( MappedType.class, (context) -> {
			context.massLoadingStrategy( new AppDataMassLoadingStrategy( db, "mapped" ) );
			context.selectionLoadingStrategy( new AppDataSelectionLoadingStrategy( db, "mapped type" ) );
		} );

		builder.addEntityType( User.class );

		builder.addEntityType( Partition.class );

		builder.addEntityType( Topic.class );

		builder.addEntityType( Publication.class );
..............		
	       MassIndexer indexer = session.massIndexer( Object.class );
```
Returns:
```
maj 07, 2021 2:32:22 PM org.hibernate.annotations.common.reflection.java.JavaReflectionManager <clinit>
INFO: HCANN000001: Hibernate Commons Annotations {5.1.2.Final}
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <mapped type> with types Partition
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <message type> with types Publication
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <any type> with types Topic
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <identity type> with types User
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <any type> with types Topic
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <identity type> with types User
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <any type> with types Topic
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <message type> with types Publication
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <mapped type> with types Partition
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <message type> with types Publication
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <identity type> with types User
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <identity type> with types User
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <mapped type> with types Partition
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <mapped type> with types Partition
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <message type> with types Publication
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <mapped type> with types Partition
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <message type> with types Publication
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <identity type> with types User
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <any type> with types Topic
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <message type> with types Publication
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <identity type> with types User
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <mapped type> with types Partition
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <any type> with types Topic
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createEntityLoader
INFO: Create mass indexing entity loader for <any type> with types Topic
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createIdentifierLoader
INFO: Create mass indexing identifier loader for <message type> with types Publication
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createIdentifierLoader
INFO: Create mass indexing identifier loader for <identity type> with types User
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createIdentifierLoader
INFO: Create mass indexing identifier loader for <mapped type> with types Partition
maj 07, 2021 2:32:23 PM org.hibernate.javabean.test.search.AppDataMassLoadingStrategy createIdentifierLoader
INFO: Create mass indexing identifier loader for <any type> with types Topic
```

yrodiere:
I thought about it, but unfortunately it's not possible.

The Hibernate ORM mapper has the additional requirement of starting a transaction before entities are loaded, and closing it after they are processed. That requirement cannot be fullfilled with the more limited `SelectionEntityLoader`, since it returns a list of entities and has no idea for how long it will be used, and thus cannot close the transaction.

Of course we could hack our way into making it fit, but frankly I'd rather not.

So you have two separate types, with separate features. If you happen to be able to use a selection loading strategy for mass loading, then you can perfectly well write your own implementation that implements both interfaces. And all will be well :)

yrodiere:
Also, note that while you can currently use lambdas, the interfaces are not annotated with `@FunctionalInterface` on purpose: they are not *intended* to be implemented as lambdas, that just happens to be possible right now. Loading is probably the kind of thing that deserves its own, separate class file.

wklaczynski:
> Also, note that while you can currently use lambdas, the interfaces are not annotated with `@FunctionalInterface` on purpose: they are not _intended_ to be implemented as lambdas, that just happens to be possible right now. Loading is probably the kind of thing that deserves its own, separate class file.

Recently, JAVA has been trying to make things easier. Like TypeScript. It is the developers who should decide how to use it. And since something is possible it should not be forbidden.

wklaczynski:
> @wklaczynski I pushed an update. Your "incompatible types" problem should be gone.
> 
> As to grouping/inheritance, as I explained above it's just a matter of implementing equals/hashcode in your strategies.
> 
> Is everything good apart from that?

It works.

PR #2565, opened 2021-05-03
HSEARCH-4163 Enable multi-tenancy by default if it is defined in ORM

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4163

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchBackendFactory.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchBackendFactory.java
@@ -32,6 +32,7 @@ import org.hibernate.search.engine.backend.spi.BackendFactory;
 import org.hibernate.search.engine.backend.spi.BackendImplementor;
 import org.hibernate.search.engine.cfg.spi.ConfigurationProperty;
 import org.hibernate.search.engine.cfg.ConfigurationPropertySource;
+import org.hibernate.search.engine.cfg.spi.OptionalConfigurationProperty;
 import org.hibernate.search.engine.environment.bean.BeanHolder;
 import org.hibernate.search.engine.environment.bean.BeanReference;
 import org.hibernate.search.engine.environment.bean.BeanResolver;
@@ -45,10 +46,9 @@ import com.google.gson.GsonBuilder;
 
 public class ElasticsearchBackendFactory implements BackendFactory {
 
-	private static final ConfigurationProperty<MultiTenancyStrategyName> MULTI_TENANCY_STRATEGY =
+	private static final OptionalConfigurationProperty<MultiTenancyStrategyName> MULTI_TENANCY_STRATEGY =
 			ConfigurationProperty.forKey( ElasticsearchBackendSettings.MULTI_TENANCY_STRATEGY )
 					.as( MultiTenancyStrategyName.class, MultiTenancyStrategyName::of )
-					.withDefault( ElasticsearchBackendSettings.Defaults.MULTI_TENANCY_STRATEGY )
 					.build();
 
 	private static final ConfigurationProperty<Boolean> LOG_JSON_PRETTY_PRINTING =
@@ -131,7 +131,7 @@ public class ElasticsearchBackendFactory implements BackendFactory {
 					threads, link,
 					typeFactoryProvider,
 					userFacingGson,
-					getMultiTenancyStrategy( propertySource ),
+					getMultiTenancyStrategy( propertySource, buildContext ),
 					indexLayoutStrategyHolder,
 					createTypeNameMapping( propertySource, indexLayoutStrategyHolder.get() ),
 					buildContext.failureHandler(), buildContext.timingSource()
@@ -147,10 +147,16 @@ public class ElasticsearchBackendFactory implements BackendFactory {
 		}
 	}
 
-	private MultiTenancyStrategy getMultiTenancyStrategy(ConfigurationPropertySource propertySource) {
-		MultiTenancyStrategyName multiTenancyStrategyName = MULTI_TENANCY_STRATEGY.get( propertySource );
+	private MultiTenancyStrategy getMultiTenancyStrategy(ConfigurationPropertySource propertySource,
+			BackendBuildContext buildContext) {
+		Optional<MultiTenancyStrategyName> multiTenancyStrategyName = MULTI_TENANCY_STRATEGY.get( propertySource );
+		if ( !multiTenancyStrategyName.isPresent() ) {
+			// the default depends on mapping
+			return buildContext.multiTenancyEnabled() ?
+					new DiscriminatorMultiTenancyStrategy() : new NoMultiTenancyStrategy();
+		}
 
-		switch ( multiTenancyStrategyName ) {
+		switch ( multiTenancyStrategyName.get() ) {
 			case NONE:
 				return new NoMultiTenancyStrategy();
 			case DISCRIMINATOR:
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchBackendFactory.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchBackendFactory.java
@@ -6,6 +6,7 @@
  */
 package org.hibernate.search.backend.elasticsearch.impl;
 
+import java.lang.invoke.MethodHandles;
 import java.util.Locale;
 import java.util.Optional;
 
@@ -17,6 +18,7 @@ import org.hibernate.search.backend.elasticsearch.dialect.impl.ElasticsearchDial
 import org.hibernate.search.backend.elasticsearch.dialect.model.impl.ElasticsearchModelDialect;
 import org.hibernate.search.backend.elasticsearch.gson.spi.GsonProvider;
 import org.hibernate.search.backend.elasticsearch.index.layout.IndexLayoutStrategy;
+import org.hibernate.search.backend.elasticsearch.logging.impl.Log;
 import org.hibernate.search.backend.elasticsearch.mapping.TypeNameMappingStrategyName;
 import org.hibernate.search.backend.elasticsearch.mapping.impl.DiscriminatorTypeNameMapping;
 import org.hibernate.search.backend.elasticsearch.mapping.impl.IndexNameTypeNameMapping;
@@ -38,6 +40,7 @@ import org.hibernate.search.engine.environment.bean.BeanReference;
 import org.hibernate.search.engine.environment.bean.BeanResolver;
 import org.hibernate.search.util.common.AssertionFailure;
 import org.hibernate.search.util.common.impl.SuppressingCloser;
+import org.hibernate.search.util.common.logging.impl.LoggerFactory;
 import org.hibernate.search.util.common.reporting.EventContext;
 
 import com.google.gson.Gson;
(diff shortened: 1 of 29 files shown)


integrationtest/mapper/orm-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/orm/realbackend/multitenant/RealBackendDatabaseMultitenancyIT.java
```diff
@@ -0,0 +1,160 @@
...
+		catch (Exception e) {
+			assertThat( e ).isInstanceOf( SearchException.class )
+					.hasMessageMatching( FailureReportUtils.buildFailureReportPattern()
+							.typeContext( IndexedEntity.class.getName() )
+							.defaultBackendContext()
+							.failure( "Invalid backend configuration: " +
+									"index 'indexed' requires multi-tenancy but no multi-tenancy strategy is set" )
+							.build()
+					);
+		}
```
yrodiere, line 84: You really should use `Assertions.assertThatThrownBy()`.

integrationtest/mapper/orm-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/orm/realbackend/multitenant/RealBackendDatabaseMultitenancyIT.java
```diff
@@ -0,0 +1,160 @@
...
+			// do not catch this kind of exception
+			throw e;
+		}
+		catch (Exception e) {
+			assertThat( e ).isInstanceOf( SearchException.class )
+					.hasMessageMatching( FailureReportUtils.buildFailureReportPattern()
+							.typeContext( IndexedEntity.class.getName() )
+							.defaultBackendContext()
+							.failure( "Invalid backend configuration: " +
+									"index 'indexed' requires multi-tenancy but no multi-tenancy strategy is set" )
```
yrodiere, line 81: I wonder why you didn't remove this check? Now that the multi-tenancy is set at the backend level, it makes no sense to pass the information "I need multi-tenancy" at the index level... The checks, if any, should really be where multi-tenancy support gets decided, i.e. at the backend level.

util/internal/integrationtest/mapper/stub/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/stub/StubMapper.java
```diff
@@ -63,9 +63,9 @@ public void prepareIndexedTypes(Consumer<Optional<String>> backendNameCollector)
 				} );
 	}
 
-	private void prepareType(MappableTypeModel type, Consumer<Optional<String>> backendNameCollector) {
+	private void prepareType(MappableTypeModel type, BackendsInfo backendsInfo) {
 		getMappedIndex( type )
-				.ifPresent( mappedIndex -> backendNameCollector.accept( mappedIndex.backendName() ) );
+				.ifPresent( mappedIndex -> backendsInfo.collect( mappedIndex.backendName(), false ) );
```
yrodiere, line 68: This hard-coded `false` should use the `multiTenancyEnabled` field instead.

yrodiere (changes requested):
I think there are some missing checks (and probably tests). Most of all, I think you should remove the `multiTenancyEnabled` parameter in `org.hibernate.search.engine.backend.spi.BackendImplementor#createIndexManagerBuilder`: it does not make sense now that multi-tenancy support is decided at the backend level.

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/spi/BackendsInfo.java
```diff
@@ -0,0 +1,78 @@
...
+		BackendsInfo result = new BackendsInfo();
+		result.collect( Optional.empty(), false );
+		return result;
+	}
+
+	public static BackendsInfo namedNoMultiTenancy(String name) {
+		BackendsInfo result = new BackendsInfo();
+		result.collect( Optional.of( name ), false );
+		return result;
+	}
```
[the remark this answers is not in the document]
fax4ever: Absolutely, I'm moving them to the test class.

integrationtest/mapper/orm-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/orm/realbackend/multitenant/RealBackendDatabaseMultitenancyIT.java
fax4ever: Using `Assertions.assertThatThrownBy()` I got some errors if the dialect is not H2.
For instance [see this test execution](https://ci.hibernate.org/job/hibernate-search-personal-fabio/job/HSEARCH-4163/3/testReport/junit/org.hibernate.search.integrationtest.mapper.orm.realbackend.multitenant/RealBackendDatabaseMultitenancyIT(database-postgresql-default-elasticsearch)/Non_default_environments___database_postgresql___multiTenancyStrategy_none/).
I don't know if there exists a way to do that.

yrodiere: I see.

You could try something like this:

```
assertThatThrownBy( ... )
        .asInstanceOf( rethrowAssertionFailures() )
        .isInstanceOf( ... )
        ....
```

And put this in `org/hibernate/search/util/impl/test/rule/AssertFactories.java`:

```
public final class AssertFactories {

    private AssertFactories() {
    }

    public static InstanceOfAssertFactory<Throwable, AbstractThrowableAssert<?, ? extends Throwable>> rethrowAssertionFailures() {
        return new InstanceOfAssertFactory( Throwable.class, throwable -> {
            if ( throwable instanceof org.junit.internal.AssumptionViolatedException
                        || throwable instanceof AssertionError ) {
                throw Throwables.toRuntimeException( throwable );
            }
            return Assertions.assertThat( throwable );
        } );
    }

}

```

If you have to stick with this `try`/`catch` for some reason, know that the problem is this test will pass if the startup doesn't throw an exception. So please at least add an Assertions.fail() at the end of the `try` block.

fax4ever: If I try to implement the InstanceOfAssertFactory as you write above, I got this error:
> java.lang.ClassCastException: class org.assertj.core.api.ObjectAssert cannot be cast to class org.assertj.core.api.AbstractThrowableAssert (org.assertj.core.api.ObjectAssert and org.assertj.core.api.AbstractThrowableAssert are in unnamed module of loader 'app')

I'm going to add the `Assertions.fail()` at the end of the try block.

yrodiere: That's just because of a missing diamond operator, I think:

```java
public final class AssertFactories {

    private AssertFactories() {
    }

    public static InstanceOfAssertFactory<Throwable, AbstractThrowableAssert<?, ? extends Throwable>> rethrowAssertionFailures() {
        return new InstanceOfAssertFactory<>( Throwable.class, throwable -> {
            if ( throwable instanceof org.junit.internal.AssumptionViolatedException
                        || throwable instanceof AssertionError ) {
                throw Throwables.toRuntimeException( throwable );
            }
            return Assertions.assertThat( throwable );
        } );
    }

}
```

fax4ever:
@yrodiere thanks for the review.
I should have addressed all the comments.
Feel free to review it again when you're ready.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/impl/ElasticsearchBackendFactory.java
```diff
@@ -147,10 +152,27 @@ public BackendImplementor create(EventContext eventContext, BackendBuildContext
...
+		MultiTenancyStrategyName multiTenancyStrategy = multiTenancyStrategyName.get();
+		// check multiTenancyStrategy mismatch: required by the mapper vs explicitly configured with properties
+		if ( MultiTenancyStrategyName.NONE.equals( multiTenancyStrategy ) &&
+				buildContext.multiTenancyEnabled() ) {
+			throw log.multiTenancyRequiredButExplicitlyDisabledByBackend( eventContext );
+		}
+		if ( MultiTenancyStrategyName.DISCRIMINATOR.equals( multiTenancyStrategy ) &&
+				!buildContext.multiTenancyEnabled() ) {
+			throw log.multiTenancyNotRequiredButExplicitlyEnabledByTheBackend( eventContext );
+		}
```
yrodiere, line 173: Note you can get even better error messages, with the property name in the error message, by moving this check where you get the property value.

And by the way, you should not need to pass context to your log method, there should be some try/catch higher in the call stack that adds the context automatically.

For instance:

```java
Optional<MultiTenancyStrategyName> multiTenancyStrategyName = MULTI_TENANCY_STRATEGY.getAndTransform( propertySource, name -> {
	boolean enabled = !MultiTenancyStrategyName.NONE.equals( multiTenancyStrategy );
	if ( !enabled && buildContext.multiTenancyEnabled() ) {
		throw log.multiTenancyRequiredButExplicitlyDisabledByBackend();
	}
	else if ( enabled && !buildContext.multiTenancyEnabled() ) {
		throw log.multiTenancyNotRequiredButExplicitlyEnabledByTheBackend();
	}
	return name;
} );
```

Then you'll get an error like:

```
default backend:
  failures:
     - Invalid value for property 'hibernate.search.backend.multitenancy.strategy': 'NONE'. The mapping requires multi-tenancy but no multi-tenancy strategy is set.
```

You can find a similar, though more complex, example in `org.hibernate.search.backend.elasticsearch.impl.ElasticsearchLinkImpl#initVersion`.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/impl/LuceneBackendFactory.java
```diff
@@ -129,10 +129,27 @@ private Version getLuceneVersion(EventContext backendContext, ConfigurationPrope
...
+		MultiTenancyStrategyName multiTenancyStrategy = multiTenancyStrategyName.get();
+		// check multiTenancyStrategy mismatch: required by the mapper vs explicitly configured with properties
+		if ( MultiTenancyStrategyName.NONE.equals( multiTenancyStrategy ) &&
+				buildContext.multiTenancyEnabled() ) {
+			throw log.multiTenancyRequiredButExplicitlyDisabledByBackend( eventContext );
+		}
+		if ( MultiTenancyStrategyName.DISCRIMINATOR.equals( multiTenancyStrategy ) &&
+				!buildContext.multiTenancyEnabled() ) {
+			throw log.multiTenancyNotRequiredButExplicitlyEnabledByTheBackend( eventContext );
+		}
```
yrodiere, line 150: Same as Elasticsearch: you can get even better error messages, with the property name in the error message, by moving this check where you get the property value.

yrodiere (commented):
My two cents: some messages are grammatically incorrect, and I suggested a small improvement in how errors on configuration properties are reported.

I also pushed a commit to simplify the exception assertion in RealBackendDatabaseMultitenancyIT.

Feel free to merge after this is solved. Thanks!

fax4ever:
I run again a full build to check non-H2 environments: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2565/10/pipeline

PR #2566, opened 2021-05-03
HSEARCH-4204 Expose the Elasticsearch index names in the metamodel API

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4204

### DIFF
--- a/documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
+++ b/documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
@@ -504,6 +504,23 @@ include::{sourcedir}/org/hibernate/search/documentation/backend/elasticsearch/la
 ----
 ====
 
+=== Retrieving the read and write index names
+
+Index or alias names used to read and write can be retrieved from the <<mapper-orm-mapping-inspect,metamodel>>.
+
+.Retrieving the index names from an Elasticsearch index manager
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/backend/elasticsearch/indexmanager/ElasticsearchIndexManagerIT.java[tags=readWriteName]
+----
+<1> Retrieve the `SearchMapping`.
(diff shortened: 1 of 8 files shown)


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/ElasticsearchIndexDescriptor.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.elasticsearch.document.model;
```
yrodiere, line 7: I think it would make more sense to mirror the engine packages. So this would be:

```suggestion
package org.hibernate.search.backend.elasticsearch.metamodel;
```

Implementations can stay where they are, that doesn't matter. But the API matters.

documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
```diff
@@ -504,6 +504,23 @@ include::{sourcedir}/org/hibernate/search/documentation/backend/elasticsearch/la
 ----
 ====
 
+=== Retrieving the read and write index names
+
+Index or alias names used to read and write can be retrieved from the `LuceneIndexManager`.
```
yrodiere, line 509:
```suggestion
Index or alias names used to read and write can be retrieved from the <<mapper-orm-mapping-inspect,metamodel>>.
```

yrodiere (changes requested):
Thanks! Looks good overall, just a wrong package for the API (IMO) and a copy/paste mistake in the documentation.

yrodiere:
Feel free to merge when they are solved. And don't forget to apply the same changes to the backport :)

PR #2574, opened 2021-05-11
HSEARCH-4198 Transactional outbox event creation with Spring and JTA

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4198

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ nbactions.xml
 .DS_Store
 .attach_pid*
 
+# Bitronix transaction logs
+*.btm
+
--- a/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/outbox/impl/OutboxEventBackgroundExecutor.java
+++ b/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/outbox/impl/OutboxEventBackgroundExecutor.java
@@ -106,6 +106,7 @@ public class OutboxEventBackgroundExecutor {
 					List<OutboxEvent> events = finder.findOutboxEvents( session, batchSize );
 					if ( events.isEmpty() ) {
 						// Nothing to do, try again later (complete() will be called, re-scheduling the polling for later)
+						transactionHelper.commit( session );
 						return CompletableFuture.completedFuture( null );
 					}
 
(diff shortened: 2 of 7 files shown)


mapper/orm/src/main/java/org/hibernate/search/mapper/orm/outbox/impl/OutboxTableSendingPlan.java
```diff
@@ -49,18 +50,43 @@ public void discard() {
...
+					.autoClose( false )
+					.connectionHandlingMode( PhysicalConnectionHandlingMode.DELAYED_ACQUISITION_AND_RELEASE_AFTER_TRANSACTION )
+					.openSession();
+			return sendAndReportOnSession( temporarySession, entityReferenceFactory );
+		}
+		finally {
+			if ( temporarySession != null ) {
+				temporarySession.close();
+			}
+		}
```
yrodiere, line 73: A `try-with-resources` block would be nice here. And that's not just nitpicking:

* with try-with-resources, if `sendAndReportOnSession` throws an exception and `temporarySession.close()` also throws an exception, the second exception will correctly be added as suppressed to the first one.
* with `try/finally`, in the same situation, the first exception will be ignored and only the second one (from `temporarySession.close()`) will be propagated.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/outbox/impl/OutboxEventBackgroundExecutor.java
```diff
@@ -104,6 +104,7 @@ public HibernateOrmOutboxWorker(SessionFactoryImplementor sessionFactory) {
 					List<OutboxEvent> outboxes = finder.findOutboxEvents( session, batchSize );
 					if ( outboxes.isEmpty() ) {
 						// Nothing to do, try again later (complete() will be called, re-scheduling the polling for later)
+						transactionHelper.commit( session );
```
yrodiere, line 107: Nice catch.

.gitignore
```diff
@@ -45,3 +45,7 @@ nbactions.xml
 .DS_Store
 .attach_pid*
 
+# Bitronix transaction logs
+part1.btm
+part2.btm
```
yrodiere, line 50: There may be a `part3` from time to time? Better safe than sorry:

```suggestion
*.btm
```

yrodiere (commented):
Looks good apart from a minor comment. Feel free to merge after it's addressed. Thanks!

fax4ever:
Full build: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2574/3/pipeline

fax4ever:
Thanks @yrodiere, I applied the suggestions.
I'll merge it as soon as the full-CI passes.

fax4ever:
Merged, thanks!

PR #2579, opened 2021-05-19
HSEARCH-4202 Test eventual consistency of outbox events in the wrong order

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4202

Given the three cases provided by the description of the issue, only the second case seems not to be working...
Let's check that on review...

### DIFF
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/outbox/OutboxPollingOutOfOrderIdsIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/outbox/OutboxPollingOutOfOrderIdsIT.java
@@ -33,6 +33,7 @@ import org.hibernate.search.util.impl.integrationtest.mapper.orm.OrmSetupHelper;
 import org.hibernate.search.util.impl.integrationtest.mapper.orm.OrmUtils;
 
 import org.junit.Before;
+import org.junit.Ignore;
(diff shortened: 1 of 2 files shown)


integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/outbox/OutboxPollingOutOfOrderIdsIT.java
```diff
@@ -120,13 +123,111 @@ public void processCreateUpdateDelete() {
...
+
+		backendMock.expectWorks( IndexedEntity.INDEX )
+				.addOrUpdate( "1", b -> b.field( "indexedField", "value for the field" ) )
+				.createdThenExecuted();
+		outboxEventFinder.showAllEventsUpToNow( sessionFactory );
+		backendMock.verifyExpectationsMet();
+	}
+
+	@Test
+	@Ignore("Probably we need to create an issue for it")
```
yrodiere, line 166: Ok, I see what the problem is. I created https://hibernate.atlassian.net/browse/HSEARCH-4228

Please mention the ticket here:

```suggestion
	@Ignore("HSEARCH-4228 Ensure correct processing of events even if their IDs are out of order")
```

yrodiere (commented):
LGTM; please apply my suggestion then feel free to merge. Thanks!

fax4ever:
Merged, thanks!

PR #2580, opened 2021-05-20
HSEARCH-4206 Upgrade to Maven 3.8.1

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4206

Giving up trying to update Maven assembly plugin to version `3.2.0`.


yrodiere:
Thanks, look good. Did you check that all tests are still getting executed? If not, I suppose we can have a look when the CI build completes.

fax4ever:
I've already done several full builds on my personal CI, so I don't expect nothing wrong here.
Anyway here is a full build on PR's CI: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2580/2/.

fax4ever:
I'll wait for it then I merge it.

yrodiere:
> I've already done several full builds on my personal CI, so I don't expect nothing wrong here.

I'm not talking about failing tests, I'm talking about tests that simply don't get executed at all. That has been a problem in the past (in particular for Java module tests).

That being said, looking closer at your patch, it seems you didn't upgrade surefire/failsafe. I suppose they are already at the latest version. If so, then you're right, we don't need to check anything more :)

fax4ever:
Surefire and failsafe plugins are already at `3.0.3-M5`, that seems to be the latest at the very moment.

fax4ever:
Converted to draft, since I want to check the number of executed tests.

fax4ever:
From my check the number of tests have been changed but it seems to be related to do not repeat more times aws ES 5.6 tests.

fax4ever:
Merged, thanks!

PR #2582, opened 2021-05-27
HSEARCH-4230 Preserve the relative order of events for a given entity even in the case of retries

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4230


fax4ever:
Looks good. Merged, thanks!

PR #2583, opened 2021-05-31
HSEARCH-4141 Move the resolution of containing entities to reindex to background processes

yrodiere:
* [HSEARCH-4141](https://hibernate.atlassian.net/browse/HSEARCH-4141): Move the resolution of containing entities to reindex to background processes

~Based on #2582 , which should be merged first.~ => Done

In short, this changes the event-queue-based automatic indexing so that, when a change occurs in a contained entity:

* we create an event for that change (with information about the paths that changed).
* when processing that event, we create one `ADD_OR_UPDATE` event per containing entity that needs to be reindexed.

This will result in more stress on the database, as well as (likely) more latency between when the change occurs and when it's visible in the index.

However, this will also reduce the work needed in the transaction that performs the database changes, which is one of the main goals of asynchronous automatic indexing. That's why this change is definitely appropriate.

As to further work... we may one day offer ways for users to specify some changes to trigger reindexing resolution synchronously in the same transaction where the change occurred, while other changes trigger reindexing resolution asynchronously in the background process. But I'll wait until we have use cases or benchmark results before considering this.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/common/impl/BackendBuildContextImpl.java
+++ b/engine/src/main/java/org/hibernate/search/engine/common/impl/BackendBuildContextImpl.java
@@ -7,18 +7,19 @@
 package org.hibernate.search.engine.common.impl;
 
 import org.hibernate.search.engine.backend.spi.BackendBuildContext;
+import org.hibernate.search.engine.tenancy.spi.TenancyMode;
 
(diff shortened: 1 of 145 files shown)


fax4ever:
I spent some time reviewing the contribution, but I can say that there is nothing I would change here. I'm merging it...

fax4ever:
I just noticed that the CI failed....

yrodiere:
Merged, thanks

PR #2586, opened 2021-06-03
HSEARCH-4229 Use Long for OutboxEvent identifiers

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4229


yrodiere:
Merged, thanks

PR #2589, opened 2021-06-09
HSEARCH-4236 + HSEARCH-4237 No cleanup of threads, if an connection exception during mass indexation happens

yrodiere:
* [HSEARCH-4237](https://hibernate.atlassian.net/browse/HSEARCH-4237): Work around Surefire clearing the interrupt status when forwarding stdout/stderrn
* [HSEARCH-4236](https://hibernate.atlassian.net/browse/HSEARCH-4236): No cleanup of threads, if an connection exception during mass indexation happens


fax4ever (approved):
Great work, really!
+1 to merge it

yrodiere:

PR #2597, opened 2021-06-16
HSEARCH-4244 Migration guide from 6.0 to 6.1

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4244


fax4ever (commented):
It looks good to me.
I would just advise to add link to the Migration guide from 5.x to 6.1,
in case the user needs to move directly from 5.x to 6.1.

yrodiere:
Merged, thanks.

There is no migration guide from 5.x to 6.1, and there won't be. People wanting to migrate from 5.x to 6.1 will have to move to 6.0 first, then to 6.1, as explained in the guide.

fax4ever:
Sorry I meant
> I would just advise to add link to the Migration guide from 5.x to 6.**0**

Just why the user have to move to 6.0 first, then to 6.1, I would add the link.

yrodiere:
There already is such a link:

```
If you're looking to migrate from an earlier version,
you should migrate step-by-step, from one minor version to the next,
following the migration guide of link:https://hibernate.org/search/documentation/[each version].
```

PR #2598, opened 2021-06-16
HSEARCH-4225 Memory leak (SessionImpl) when updating index without transaction

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4225

This bug is not relevant to Hibernate Search 6, since we don't rely on weak-reference hashmaps there (we simply store everything in the session).


fax4ever (approved):
It looks good!

yrodiere:
Merged, thanks

PR #2599, opened 2021-06-16
HSEARCH-4227 Safer work expectations in backend mock

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4227


util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/rule/VerifyingStubBackendBehavior.java
```diff
@@ -58,7 +58,9 @@
 
 	private final Map<String, CallQueue<SchemaManagementWorkCall>> schemaManagementWorkCall = new HashMap<>();
 
-	private final Map<DocumentKey, CallQueue<DocumentWorkCall>> documentWorkCalls = new LinkedHashMap<>();
+	private final Map<DocumentKey, CallQueue<DocumentWorkCreateCall>> documentWorkCreateCalls = new LinkedHashMap<>();
+	private final Map<DocumentKey, CallQueue<DocumentWorkDiscardCall>> documentWorkDiscardCalls = new LinkedHashMap<>();
+	private final Map<DocumentKey, CallQueue<DocumentWorkExecuteCall>> documentWorkExecuteCalls = new LinkedHashMap<>();
```
fax4ever, line 63: I'm asking myself if having three different maps may affect the check of the total order on a sequence of different kind of `DocumentWork*Call`.

fax4ever (approved):
Nice job. I've just a doubt.
+1 to merge it!

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/rule/VerifyingStubBackendBehavior.java
yrodiere: It does, but we don't actually care.

A work can be created then discarded, or created then executed. No other combination makes sense or happens in the tests). Obviously the work cannot be executed or discarded before it was created. So we don't have anything to check :)

fax4ever: makes sense!

PR #2603, opened 2021-06-21
HSEARCH-4239 + HSEARCH-4249 AWS signature testing and fix 

yrodiere:
* [HSEARCH-4249](https://hibernate.atlassian.net/browse/HSEARCH-4249): Test AWS Elasticsearch Service 7.10
* [HSEARCH-4239](https://hibernate.atlassian.net/browse/HSEARCH-4239): Credentials error while mass indexing on AWS ES with static credentials: "The request signature we calculated does not match the signature you provided."

It turns out the problem was in a method we use to generate an InputStream that will be consumed by AWS SDK's signer; it sometimes overwrote previous content, leading to an invalid content hash and thus an invalid signature.

This only happened with specific input of a specific size, which is why we didn't detect this earlier.


yrodiere:
Build including tests against AWS Elasticsearch service: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2603/2/

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/work/IndexIndexerLargeDocumentsIT.java
```diff
@@ -0,0 +1,175 @@
...
+	private void indexAndWait(int count, IntFunction<String> valueProvider, Operation operation) {
+		IndexIndexer indexer = index.createIndexer();
+		CompletableFuture<?>[] tasks = new CompletableFuture<?>[count];
+		for ( int i = 0; i < count; i++ ) {
+			final int id = i;
+			tasks[i] = operation.apply( indexer, referenceProvider( String.valueOf( id ) ),
+					document -> document.addValue( index.binding().content, valueProvider.apply( id ) ) );
+		}
+		CompletableFuture<?> future = CompletableFuture.allOf( tasks );
+		Awaitility.await().until( future::isDone );
```
fax4ever, line 120: This is just a curiosity. Why are we using the `await` method instead of `Future#get` or `join`?

fax4ever (approved):
Just one doubt.
+1 to merge it
I just don't understand why according to the issue GitHub page the line of codes changed here are +8447 -27, but I didn't find those changes.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/work/IndexIndexerLargeDocumentsIT.java
yrodiere: So that if the future never finishes, we are 100% sure the test does finish and throws a nice exception that mentions how much time we waited.

fax4ever:
Screenshot 2021-06-21 at 11 13 06
Where are these changes?

yrodiere:
> Screenshot 2021-06-21 at 11 13 06
> Where are these changes?

At the bottom.

I added a resource file that contains the full text of "Great Expectations" (it's in the public domain), so that's a lot. I won't ask you to review that, though :P

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/work/IndexIndexerLargeDocumentsIT.java
fax4ever: IMO if the almost the same feature is provided by the `java.util.concurrent` library we don't need to use an external library. On the other hand, that is not absolutely critical.

fax4ever:
> > Screenshot 2021-06-21 at 11 13 06
> > Where are these changes?
> 
> At the bottom.
> 
> I added a resource file that contains the full text of "Great Expectations" (it's in the public domain), so that's a lot. I won't ask you to review that, though :P

Right. Sure I didn't

yrodiere:
Merged, thanks

PR #2607, opened 2021-06-23
HSEARCH-4250 Deprecate ElasticsearcAwsCredentialsProvider and make it extend a new ElasticsearchAwsCredentialsProvider

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4250


fax4ever (approved):
Definitely it makes sense to fix the typo in the class name.
But what happens if a user still uses the deprecated interface?

yrodiere:
> But what happens if a user still uses the deprecated interface?

It will work, since the deprecated interface extends the new one.

Merging, thanks.

PR #2608, opened 2021-06-23
HSEARCH-3878 Maximize utilization of database connections during mass indexing

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3878

I didn't have time to run full performance tests, but I did examine how it improves the behavior of backend executors.

Below are Gantt charts of the indexing batches handled by the backend during mass indexing. Each swimlane is an executor, and each "task" is a batch, with the color of each task ranging from dark red (few documents) to light green (many documents).

I generated this data by mass indexing in the [Hibernate Search Wikipedia demo](https://github.com/hibernate/hibernate-demos/tree/main/hibernate-search/hsearch-elasticsearch-wikipedia). The application indexed approximately 35000 entites, a lot of which are quite text-heavy (Wikipedia pages).

(for future reference, here was the command I executed: `logs2gantt --regex "(?P<datetime>[^,]+) TRACE.*((?P<start>Processing)|Processed) (?P<scalar>[0-9]+).*executor '(?P<swimlane>.+)'" --output-format=png --output=lucene-after.png lucene-after.log`)

## Elasticsearch

es-before
es-after

## Lucene

lucene-before
lucene-after

## Analysis

It's pretty clear with Lucene; a bit less so with Elasticsearch. The batch sizes are much more consistent, a sign that we're keeping the backend busy enough that it always has leftover work in the queue. All that while using the exact same amount of database connections.

Regarding throughput: before the change, we indexed about 850 doc/s; after the change, we indexed about 900 doc/s (1000 for Lucene). The throughput change is not significant enough to say there was an improvement, especially considering the setup was far from real-world setups (one laptop with postgres + 2 ES nodes running on the same machine). Nevertheless, I think it's a good sign that the throughput did not decrease.

Conclusion: we got exactly what we wanted, but further testing is required to see if we got a throughput improvement as a bonus.

### DIFF
--- a/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingEntityLoadingRunnable.java
+++ b/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingEntityLoadingRunnable.java
@@ -7,6 +7,7 @@
 package org.hibernate.search.mapper.pojo.massindexing.impl;
 
 import java.lang.invoke.MethodHandles;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


fax4ever:
Looks good, merged thanks!

PR #2611, opened 2021-07-09
HSEARCH-4175 Unify field value converter with identifier value converters

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4175


yrodiere:
@fax4ever This is mostly refactoring, so unless you really want to have a closer look, I'll just wait for tests to pass, then merge.

PR #2613, opened 2021-07-13
HSEARCH-4254 AWS request signing producing an incorrect signature in very specific cases involving non-ascii characters

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4254


fax4ever:
It looks good. Merged thanks!

PR #2615, opened 2021-07-16
HSEARCH-4260 Display Java version during builds

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4260


fax4ever:
Merged, thanks!

PR #2617, opened 2021-07-16
HSEARCH-4242 Test more databases

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4242


fax4ever:
Run a full build: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2617/2/pipeline

integrationtest/mapper/orm-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/orm/realbackend/limitations/ConcurrentEmbeddedUpdateLimitationIT.java
```diff
@@ -61,6 +63,9 @@ public void indexingStrategySession() throws Throwable {
 				.withProperty( HibernateOrmMapperSettings.AUTOMATIC_INDEXING_SYNCHRONIZATION_STRATEGY, "sync" )
 				.setup( Book.class, Author.class, BookEdition.class );
 
+		HibernateORMSkipTestHelper.skipForDialect( sessionFactory, SQLServerDialect.class,
```
yrodiere, line 66: You could perform this check before creating the session factory, which would avoid a bunch of connections and schema creation.

That's what we did for multi-tenancy tests, cf. `MultitenancyTestHelper`.

Just add this to `org.hibernate.search.util.impl.integrationtest.mapper.orm.OrmSetupHelper.SetupContext`:

```java
		public SetupContext skipTestForDialect(Class<? extends Dialect> dialect, String reason) {
			withConfiguration( b -> b.onMetadata( metadataImplementor -> {
				Dialect currentDialect = metadataImplementor.getDatabase().getDialect();
				assumeFalse( "Skipping test for dialect " + dialect.getName() + "; reason: " + reason, dialect.isAssignableFrom( currentDialect.getClass() ) );
			} ) );
			return thisAsC();
		}
```

And call it before `setup( Book.class, ... )` just above.

Jenkinsfile
```diff
@@ -200,12 +200,20 @@ stage('Configure') {
...
-							condition: TestCondition.AFTER_MERGE)
+							condition: TestCondition.AFTER_MERGE, dockerHubImage: "postgres:13.1"),
+					new DatabaseBuildEnvironment(dbName: 'mariadb', mavenProfile: 'ci-mariadb',
+							condition: TestCondition.AFTER_MERGE, dockerHubImage: "mariadb:10.5.8"),
+					new DatabaseBuildEnvironment(dbName: 'mysql', mavenProfile: 'ci-mysql',
+                    		condition: TestCondition.AFTER_MERGE, dockerHubImage: "mysql:8.0.22"),
+                    new DatabaseBuildEnvironment(dbName: 'db2', mavenProfile: 'ci-db2',
+                            condition: TestCondition.AFTER_MERGE, dockerHubImage: "ibmcom/db2:11.5.5.0"),
+                    new DatabaseBuildEnvironment(dbName: 'oracle', mavenProfile: 'ci-oracle',
+                            condition: TestCondition.AFTER_MERGE, dockerHubImage: "gvenzl/oracle-xe:11-slim"),
```
yrodiere, line 214: What do you think of executing Oracle test for every PR build, since they are very likely to fail when we introduce a column whose name is too long?
You would just need to use `TestCondition.BEFORE_MERGE` here.
Maybe we should also add `postgresql`, for good measure. I don't think others are necessary, though.

parents/integrationtest/pom.xml
```diff
@@ -355,7 +524,9 @@
                 </property>
             </activation>
             <properties>
-                <test.elasticsearch.run.skip>true</test.elasticsearch.run.skip>
+                <!-- Skipping to start all containers here -->
+                <!-- When the uris property is set, usually we don't need any container (neither ES nor others). -->
+                <test.containers.run.skip>true</test.containers.run.skip>
```
yrodiere, line 529: If you follow my suggestion regarding the configuration properties, you can revert this particular change.

parents/integrationtest/pom.xml
```diff
@@ -616,38 +787,113 @@
...
+                <jdbc.driver>com.microsoft.sqlserver.jdbc.SQLServerDriver</jdbc.driver>
+                <jdbc.url>jdbc:sqlserver://localhost:1433;databaseName=tempdb</jdbc.url>
+                <jdbc.user>sa</jdbc.user>
+                <jdbc.pass>ActuallyRequired11Complexity</jdbc.pass>
+            </properties>
+        </profile>
+
         <!--
             ###################################################################
             Then, those with commercial drivers
```
yrodiere, line 899: Can you clean this up, please? You obviously moved commercial drivers above this comment.

My suggestion:

* Remove this comment and every profile below this comment.
* Remove the comment higher up that contains "Profiles naming db instances in the Red Hat QA/QE lab".

util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/dialect/HibernateORMSkipTestHelper.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+package org.hibernate.search.util.impl.integrationtest.mapper.orm.dialect;
+
+import static org.junit.Assume.assumeFalse;
+
+import org.hibernate.SessionFactory;
+import org.hibernate.dialect.Dialect;
+import org.hibernate.engine.spi.SessionFactoryImplementor;
+
+public final class HibernateORMSkipTestHelper {
```
yrodiere, line 15: Don't forget to remove this if you follow my suggestion.

yrodiere (commented):
Looks good, thanks. I think we could improve a bit on the Maven configuration side? See my comments.

yrodiere:
Full build: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2617/4/pipeline

integrationtest/mapper/orm-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/orm/realbackend/limitations/ConcurrentEmbeddedUpdateLimitationIT.java
```diff
@@ -61,6 +63,9 @@ public void indexingStrategySession() throws Throwable {
...
+				"The execution could provoke a deadlock on SQLServer that will be solved aborting conflicting requests from client" );
```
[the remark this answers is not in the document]
fax4ever: deadlock is solved only using the `READ_UNCOMMITTED` isolation level, but the same isolation level makes the other tests not working, so we're skipping it as alternative solution

fax4ever:
Thanks @yrodiere for the review.
I applied all the changes. Let's see [the new full build](https://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-2617/8/console).

fax4ever:
It seems that the solution with `failsafe.jvm.args.jdbc` worked. 
Run a new full-build: https://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-2617/17

documentation/pom.xml
```diff
@@ -20,6 +20,8 @@
     <description>Hibernate Search reference documentation</description>
 
     <properties>
+        <test.elasticsearch.run.skip>false</test.elasticsearch.run.skip>
```
yrodiere, line 23: Shouldn't there be a `<test.database.run.skip>false</test.database.run.skip>` too? Or maybe we don't run the documentation tests against each database? If so, maybe we should?

integrationtest/v5migrationhelper/engine/pom.xml
```diff
@@ -20,6 +20,7 @@
     <description>Hibernate Search integration tests for the migration helper for Hibernate Search's Engine</description>
 
     <properties>
+        <test.elasticsearch.run.skip>false</test.elasticsearch.run.skip>
```
yrodiere, line 23: The migration helper doesn't work with Elasticsearch, so we don't test it against Elasticsearch. This line should not be here.

integrationtest/v5migrationhelper/orm/pom.xml
```diff
@@ -20,6 +20,7 @@
     <description>Hibernate Search integration tests for the migration helper for Hibernate ORM</description>
 
     <properties>
+        <test.elasticsearch.run.skip>false</test.elasticsearch.run.skip>
```
yrodiere, line 23: The migration helper doesn't work with Elasticsearch, so we don't test it against Elasticsearch. This line should not be here.

integrationtest/mapper/orm-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/orm/realbackend/limitations/ConcurrentEmbeddedUpdateLimitationIT.java
```diff
@@ -59,6 +60,8 @@ public void indexingStrategySession() throws Throwable {
 		sessionFactory = setupHelper.start()
 				// This is absolutely necessary to avoid false positives in this test
 				.withProperty( HibernateOrmMapperSettings.AUTOMATIC_INDEXING_SYNCHRONIZATION_STRATEGY, "sync" )
+				.skipTestForDialect( SQLServerDialect.class,
+						"The execution could provoke a deadlock on SQLServer that will be solved aborting conflicting requests from client" )
```
yrodiere, line 64:
```suggestion
						"The execution could provoke a deadlock on SQLServer, which will abort our requests upon detecting the deadlock, and will make the tests fail."
						+ " The only known workaround is to use the READ_UNCOMMITTED isolation level, but this makes other tests fail."
						+ " In any case, this looks like a problem in SQLServer, so we're just skipping this test." )
```

integrationtest/jdk/java-modules/pom.xml
```diff
@@ -17,6 +17,8 @@
     <description>Hibernate Search integration tests for (JDK 11+) Java modules</description>
 
     <properties>
+        <test.elasticsearch.run.skip>false</test.elasticsearch.run.skip>
```
yrodiere, line 20: Shouldn't there be a `<test.database.run.skip>false</test.database.run.skip>` too? Just for correctness.

yrodiere (approved):
Looking good. A few minor comments below.

documentation/pom.xml
fax4ever: Yes we don't run, but I think that we should. Let's add `<test.database.run.skip>false</test.database.run.skip>` for consistency.

fax4ever:
Did all the changes and run a new full-build: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2617/20/pipeline

fax4ever:
I'll merge it as soon as the full CI has passed.

fax4ever:
Finally, merged, thanks!

yrodiere:
Great news! Thanks a lot for your work @fax4ever!

PR #2620, opened 2021-07-16
HSEARCH-4248 Remove locks from ProgressiveCharBufferWriter 

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4248


fax4ever:
Merged, thanks!

PR #2622, opened 2021-07-23
HSEARCH-4265 Fix flaky tests with MS-SQL and Oracle

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4265

### DIFF
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/strategy/outbox/OutboxPollingOutOfOrderIdsIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/strategy/outbox/OutboxPollingOutOfOrderIdsIT.java
@@ -108,10 +108,19 @@ public class OutboxPollingOutOfOrderIdsIT {
 		OrmUtils.withinSession( sessionFactory, session -> {
 			List<OutboxEvent> events = outboxEventFinder.findOutboxEventsNoFilterById( session );
 			assertThat( events ).hasSize( 3 );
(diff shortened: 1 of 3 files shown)


fax4ever:
Converted to draft, since I want to inspect more on `findOutboxEvents` results, so that we get a meaningful error in case of wrong order instead of having a generic timeout!

fax4ever:
New full build: https://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-2622/

fax4ever:
Running the tests again, see https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2622/6/pipeline/66, we don't get the error again.
Anyway, I added a commit to log the returned events in case of wrong error: https://github.com/hiberynate/hibernate-search/pull/2622/commits/4c8a1363482371fec4a57b549cc5e6fa132d8b70.
@yrodiere maybe we should merge it and if the error happens again we will have the log.

yrodiere (approved):
@fax4ever Are you sure it's not simply a problem with the resolution of the date/time type we're using with MS-SQL? Or the `current_timestamp` function in MS-SQL? That would explain why it only fails from time to time: it fails when the two events are generated too close to each other and are assigned the same timestamp.

Anyway, feel free to merge if that's not the problem.

fax4ever:
Another build with oracle|mssql https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2622/9/pipeline

fax4ever:
Another specific build: https://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-2622/11/

fax4ever:
Merged, thanks!

PR #2625, opened 2021-07-27
HSEARCH-4269 Fix assertion error on rollback being reported before the assertion error that triggered the rollback

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4269


fax4ever:
Merged, thanks!

PR #2628, opened 2021-07-29
HSEARCH-4151 Reuse the same Query instance when executing multiple queries during entity loading

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4151


fax4ever:
Merged, thanks!

PR #2629, opened 2021-07-29
HSEARCH-4193 Fix tests wrongly using 0-based positional parameters with ORM

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4193


fax4ever:
Merged, thanks!

PR #2638, opened 2021-08-09
HSEARCH-4187 Document the database-polling coordination strategy

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4187

Main relevant sections of the documentations:

* `documentation/target/dist/reference/en-US/html_single/index.html#architecture`
* `documentation/target/dist/reference/en-US/html_single/index.html#mapper-orm-indexing-automatic`
* `documentation/target/dist/reference/en-US/html_single/index.html#coordination`


documentation/src/main/asciidoc/reference/coordination.asciidoc
```diff
@@ -0,0 +1,508 @@
...
+either using the <<mapper-orm-indexing-massindexer,`MassIndexer`>>
+or <<mapper-orm-indexing-manual,manually>>.
+[[coordination-none-indexing-change-filter]]
+Only relevant changes trigger indexing::
+[[mapper-orm-indexing-automatic-concepts-changes-filter]] Hibernate Search is aware of the properties that are accessed when building indexed documents.
+Thanks to that knowledge, it is able to skip reindexing
+when a property is modified, but does not affect the indexed document.
++
+You can control this "dirty checking" by setting
+the <<configuration-property-types,boolean property>> `hibernate.search.automatic_indexing.enable_dirty_check`:
```
fax4ever, line 80: is this property supported?

documentation/src/main/asciidoc/reference/coordination.asciidoc
```diff
@@ -0,0 +1,508 @@
...
+which may represent an overhead for your application threads
+as well as your database.
++
+This extra cost can be mitigated to some extent by:
+* leveraging Hibernate ORM's batch fetching:
+see link:{hibernateDocUrl}#configurations-database-fetch[the `batch_fetch_size` property]
+and link:{hibernateDocUrl}#fetching-batch[the `@BatchSize` annotation].
+* leveraging Hibernate ORM's link:{hibernateDocUrl}#caching[second-level cache],
+especially for immutable entities referenced from indexed entities
+(e.g. for reference data such as countries, cities, ...).
```
fax4ever, line 132: this is not rendered as a bullet list

documentation/src/main/asciidoc/reference/coordination.asciidoc
```diff
@@ -0,0 +1,508 @@
...
+See <<mapper-orm-indexing-automatic-synchronization>> for details.
+[[coordination-none-indexing-association-consistency]]
+Associations must be updated on both sides::
+[[mapper-orm-indexing-automatic-concepts-session-consistency]] Hibernate ORM is able to handle asymmetric updates of associations,
+where only the owning side of association is updated and the other side is ignored.
+The entities in the session will be inconsistent for the duration of the session,
+but upon reloading they will be consistent once again,
+due to how entity loading works.
++
+This dodgy practice of asymmetric updates of associations
```
fax4ever, line 164: I agree with your idea, but I wound't add the adjective `dodgy` here.

documentation/src/main/asciidoc/reference/coordination.asciidoc
```diff
@@ -0,0 +1,508 @@
...
+can cause problems in applications in general,
+but also in Hibernate Search specifically,
+where it may lead to out-of-sync indexes.
+Thus, it must be avoided.
++
+[NOTE]
+====
+For example, let's assume an indexed entity `A` has an `@IndexedEmbedded` association `A.b`to entity `B`,
```
fax4ever, line 172: there are a formatting issue, maybe we should add a space just after the term \`A.b\` here before the `to`

documentation/src/main/asciidoc/reference/coordination.asciidoc
```diff
@@ -0,0 +1,508 @@
...
+The `database-polling` strategy implements coordination through
+<<coordination-database-polling-schema,additional tables>> in the application database.
+
+<<mapper-orm-indexing-automatic,Automatic indexing>> is implemented
+by pushing events to an outbox table within the same transaction as the entity changes,
+and polling this outbox table from background processors which perform indexing.
+
+This strategy is able to provide guarantees that entities will be indexed regardless
+of temporary I/O errors in backend,
+at the cost of being only able to perform this indexing asynchronously.
```
fax4ever, line 213: Maybe we should quote the Outbox pattern

documentation/src/main/asciidoc/reference/coordination.asciidoc
```diff
@@ -0,0 +1,508 @@
...
+The `database-polling` strategy can be enabled with the following settings:
+
+[source]
+----
+hibernate.search.coordination.strategy = database-polling
+----
+
+include::components/dynamic-sharding-not-implemented-warning.asciidoc[]
+
+include::components/coordination-comparison-note.asciidoc[]
```
fax4ever, line 224: I would not repeat this block.

documentation/src/main/asciidoc/reference/coordination.asciidoc
```diff
@@ -0,0 +1,508 @@
...
+when a property is modified, but does not affect the indexed document.
++
+You can control this "dirty checking" by setting
+the <<configuration-property-types,boolean property>> `hibernate.search.automatic_indexing.enable_dirty_check`:
++
+* by default, or when set to `true`,
+Hibernate Search will consider whether modified properties are relevant
+before triggering reindexing.
+* when set to `false`, Hibernate Search will trigger reindexing upon any change,
+regardless of the entity properties that changed.
```
fax4ever, line 256: I would extract these two paragraph to the parent chapter, if these concepts are the same for both configurations.

documentation/src/main/asciidoc/reference/coordination.asciidoc
```diff
@@ -0,0 +1,508 @@
...
+the background process will load that entity in full.
+Depending on your mapping, it may also need to load lazy associations to other entities.
++
+This extra cost can be mitigated to some extent by:
+* leveraging Hibernate ORM's batch fetching;
+see link:{hibernateDocUrl}#configurations-database-fetch[the `batch_fetch_size` property]
+and link:{hibernateDocUrl}#fetching-batch[the `@BatchSize` annotation].
+* leveraging Hibernate ORM's link:{hibernateDocUrl}#caching[second-level cache],
+especially for immutable entities referenced from indexed entities
+(e.g. for reference data such as countries, cities, ...).
```
fax4ever, line 292: Also here I think that the bullet list is not rendered properly

documentation/src/main/asciidoc/reference/coordination.asciidoc
```diff
@@ -0,0 +1,508 @@
...
+where only the owning side of association is updated and the other side is ignored.
+The entities in the session will be inconsistent for the duration of the session,
+but upon reloading they will be consistent once again,
+due to how entity loading works.
++
+This dodgy practice of asymmetric updates of associations
+can cause problems in applications in general,
+but also in Hibernate Search specifically,
+where it may lead to out-of-sync indexes.
+Thus, it must be avoided.
```
fax4ever, line 328: Again I would prefer to move common paragraph to the parent chapter

documentation/src/main/asciidoc/reference/coordination.asciidoc
```diff
@@ -0,0 +1,508 @@
...
+and as such they should be automatically generated when relying on Hibernate ORM's
+link:{hibernateDocUrl}#configurations-hbmddl[automatic schema generation].
+
+If you need to integrate the creation/dropping of these tables to your own script,
+the easiest solution is to have Hibernate ORM generate DDL scripts for your whole schema
+and copy everything related to constructs (tables, sequences, ...) prefixed with `HSEARCH_`.
+See link:{hibernateDocUrl}#configurations-hbmddl[automatic schema generation],
+in particular the Hibernate ORM properties `javax.persistence.schema-generation.scripts.action`,
+`javax.persistence.schema-generation.scripts.create-target`
+and `javax.persistence.schema-generation.scripts.drop-target`.
```
fax4ever, line 367: I know that would be difficult to maintain, but maybe we should provide the sql schemas for each databases.

documentation/src/main/asciidoc/reference/coordination.asciidoc
```diff
@@ -0,0 +1,508 @@
...
+as an <<configuration-property-types,integer value>>.
+This property has no default and must be set explicitly when static sharding is enabled.
+It must be set to the same value on all application nodes with assigned shards.
+* `shards.assigned` defines the shards assigned to the application node
+as an <<configuration-property-types,integer value>>, or multiple comma-separated integer values.
+This property has no default and must be set explicitly when static sharding is enabled.
+Shards are referred to by an index in the range `[0, total_count - 1]` (see above for `total_count`).
+A given application node must be assigned at least one shard
+but may be assigned multiple shards by setting `shards.assigned` to a comma-separated list,
+e.g. `0,4`.
```
fax4ever, line 414: I wouldn't use the `4` in this example, because even if this is unrelated we said that we had 4 nodes just before. Maybe we should use `3`?

documentation/src/main/asciidoc/reference/coordination.asciidoc
```diff
@@ -0,0 +1,508 @@
...
+hibernate.search.backend.processors.indexing.batch_size = 50
+----
+
+* `processors.indexing.enabled` defines whether the background processor responsible for indexing entities is enabled,
+as a <<configuration-property-types,boolean value>>.
+The default for this property is `true`, but it can be set to `false` to disable background processing on some application nodes,
+for example to dedicate some nodes to HTTP request processing and other nodes to background processing.
+* `processors.indexing.polling_interval` defines how long to wait for another query to the outbox events table
+after a query didn't return any event,
+as an <<configuration-property-types,integer value>> in milliseconds.
```
fax4ever, line 504: I would add that this property value may affect the time needed to see changes on indexes

documentation/src/main/asciidoc/reference/coordination.asciidoc
```diff
@@ -0,0 +1,508 @@
...
+The default for this property is `100`.
+* `processors.indexing.batch_size` defines how many outbox events, at most, are processed in a single transaction
+as an <<configuration-property-types,integer value>>.
+The default for this property is `50`.
```
fax4ever, line 508: I would add that this setting can be used for performance tuning.

fax4ever (approved):
All in all is a very good work.
+1 to merge it
Here are my comments.

documentation/src/main/asciidoc/reference/architecture.asciidoc
```diff
@@ -114,45 +182,136 @@ To implement this architecture, use the following Maven dependencies:
...
+[[architecture-examples-database-polling-elasticsearch-pros-and-cons]]
+==== Pros and cons
+
+Pros:
+
+* Safest:
+** the possibility of out-of-sync indexes caused by <<limitations-backend-indexing-error,indexing errors in the backend>>
+that affects other architectures is eliminated here,
+because entity change events <<coordination-database-polling-indexing-guarantee,are persisted in the same transaction as the entity changes>>
+allowing retries for as loeach entity instance can be indexed by one and only one thread ng as necessary.
```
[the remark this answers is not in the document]
yrodiere: `allowing retries for as long as necessary`. Thanks, it seems I moved to something else while editing this line :/

documentation/src/main/asciidoc/reference/coordination.asciidoc
```diff
@@ -0,0 +1,508 @@
...
+
+All <<mapper-orm-indexing-automatic,automatic indexing>> operations
+are executed directly in application threads,
+which gives this strategy the unique ability to provide <<mapper-orm-indexing-automatic-synchronization,synchronous indexing>>,
+at the cost of a few limitations:
+
+* <<limitations-parallel-embedded-update>>
+* <<limitations-backend-indexing-error>>
+
+include::components/coordination-comparison-note.asciidoc[]
```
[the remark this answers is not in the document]
yrodiere: We have several links directly to this section, and I don't think people will bother to go back to the top of the section when the land here.

So I'd rather repeat this, since it's quite short and its only purpose is to give more general information about the coordination strategies.

documentation/src/main/asciidoc/reference/coordination.asciidoc
yrodiere: Yes? See `org.hibernate.search.mapper.orm.session.impl.ConfiguredAutomaticIndexingStrategy#DIRTY_CHECK_ENABLED`.

It's been here for a long time, since Search 5 and probably before. Personally I don't see why someone would use it, but I kept it in 6.0 to ease migration.

documentation/src/main/asciidoc/reference/coordination.asciidoc
yrodiere: I'd rather keep the first one here, because it's very important and it actually is a limitation of the current coordination strategies; one that we'll hopefully solve with Debezium one day.
I agree about the second one, though; I'll move it somewhere else.

yrodiere: After giving it some thought, and considering your other comments: you're right. I'll move everything to their own common section, and leave only links to those sections here (as a reminder).

documentation/src/main/asciidoc/reference/coordination.asciidoc
yrodiere: I really don't want to maintain that.

yrodiere:
Thanks for all these useful comments! I updated my PR accordingly; let's wait for CI then merge.

PR #2644, opened 2021-08-31
HSEARCH-3274 Expose fully defined (non-automatic) Java modules for consumption by users

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-3274
https://hibernate.atlassian.net/browse/HSEARCH-4304

Before to merge it, we need to for a new version [> 1.0.0.Beta2] of [ModiTect](https://github.com/moditect/moditect).
New versions will contain the contributions:
1. https://github.com/moditect/moditect/pull/147
2. https://github.com/moditect/moditect/pull/150

We're temporary testing it apply the main branch of ModiTect source code, then deploy a snapshot locally.


fax4ever:
@gunnarmorling this is the Search contribution to use ModiTect to generate the module-info(s).class if you wanted to look at the use case!

backend/lucene/pom.xml
```diff
@@ -62,45 +62,10 @@
...
-                            </execution>
-                        </executions>
-                    </plugin>
-                </plugins>
-            </build>
-        </profile>
-    </profiles>
+    <build>
+        <plugins>
+            <!-- Not generate module-info with moditect-maven-plugin:
```
yrodiere, line 67:
```suggestion
            <!-- Do not generate module-info with moditect-maven-plugin:
```

backend/elasticsearch-aws/pom.xml
```diff
@@ -41,4 +41,29 @@
...
+                <artifactId>moditect-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>add-module-infos</id>
+                        <configuration>
+                            <module>
+                                <moduleInfo>
+                                    <name>${java.module.name}</name>
+                                    <exports>
+                                        org.hibernate.search.*.impl to org.hibernate.search.util.common, org.hibernate.search.engine, org.hibernate.search.backend.elasticsearch, org.hibernate.search.backend.elasticsearch.aws, org.hibernate.search.mapper.pojo, org.hibernate.search.mapper.javabean, org.hibernate.search.mapper.orm, org.hibernate.search.batch.jsr352.core, org.jboss.logging;
```
yrodiere, line 58: You definitely shouldn't need to export `impl` packages here. It's only necessary in the `hibernate-search-util-common` module.

```suggestion
                                        !org.hibernate.search.*.impl;
```

backend/elasticsearch/pom.xml
```diff
@@ -56,4 +56,29 @@
...
+                <artifactId>moditect-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>add-module-infos</id>
+                        <configuration>
+                            <module>
+                                <moduleInfo>
+                                    <name>${java.module.name}</name>
+                                    <exports>
+                                        org.hibernate.search.*.impl to org.hibernate.search.util.common, org.hibernate.search.engine, org.hibernate.search.backend.elasticsearch, org.hibernate.search.backend.elasticsearch.aws, org.hibernate.search.mapper.pojo, org.hibernate.search.mapper.javabean, org.hibernate.search.mapper.orm, org.hibernate.search.batch.jsr352.core, org.jboss.logging;
```
yrodiere, line 73: You definitely shouldn't need to export `impl` packages here. It's only necessary in the `hibernate-search-util-common` module.

```suggestion
                                        !org.hibernate.search.*.impl;
```

engine/pom.xml
```diff
@@ -36,4 +36,32 @@
...
+                <artifactId>moditect-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>add-module-infos</id>
+                        <configuration>
+                            <module>
+                                <moduleInfo>
+                                    <name>${java.module.name}</name>
+                                    <exports>
+                                        org.hibernate.search.*.impl to org.hibernate.search.util.common, org.hibernate.search.engine, org.hibernate.search.backend.elasticsearch, org.hibernate.search.backend.elasticsearch.aws, org.hibernate.search.mapper.pojo, org.hibernate.search.mapper.javabean, org.hibernate.search.mapper.orm, org.hibernate.search.batch.jsr352.core, org.jboss.logging;
```
yrodiere, line 53: You definitely shouldn't need to export `impl` packages here. It's only necessary in the `hibernate-search-util-common` module.

```suggestion
                                        !org.hibernate.search.*.impl;
```

mapper/javabean/pom.xml
```diff
@@ -40,4 +40,29 @@
...
+                <artifactId>moditect-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>add-module-infos</id>
+                        <configuration>
+                            <module>
+                                <moduleInfo>
+                                    <name>${java.module.name}</name>
+                                    <exports>
+                                        org.hibernate.search.*.impl to org.hibernate.search.util.common, org.hibernate.search.engine, org.hibernate.search.backend.elasticsearch, org.hibernate.search.backend.elasticsearch.aws, org.hibernate.search.mapper.pojo, org.hibernate.search.mapper.javabean, org.hibernate.search.mapper.orm, org.hibernate.search.batch.jsr352.core, org.jboss.logging;
```
yrodiere, line 57: You definitely shouldn't need to export `impl` packages here. It's only necessary in the `hibernate-search-util-common` module.

```suggestion
                                        !org.hibernate.search.*.impl;
```

mapper/orm-batch-jsr352/core/pom.xml
```diff
@@ -72,4 +72,29 @@
...
+                <artifactId>moditect-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>add-module-infos</id>
+                        <configuration>
+                            <module>
+                                <moduleInfo>
+                                    <name>${java.module.name}</name>
+                                    <exports>
+                                        org.hibernate.search.*.impl to org.hibernate.search.util.common, org.hibernate.search.engine, org.hibernate.search.backend.elasticsearch, org.hibernate.search.backend.elasticsearch.aws, org.hibernate.search.mapper.pojo, org.hibernate.search.mapper.javabean, org.hibernate.search.mapper.orm, org.hibernate.search.batch.jsr352.core, org.jboss.logging;
```
yrodiere, line 89: You definitely shouldn't need to export `impl` packages here. It's only necessary in the `hibernate-search-util-common` module.

```suggestion
                                        !org.hibernate.search.*.impl;
```

mapper/orm-batch-jsr352/jberet/pom.xml
```diff
@@ -69,4 +69,12 @@
             <artifactId>jboss-logging-annotations</artifactId>
         </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <!-- Not generate module-info with moditect-maven-plugin:
```
yrodiere, line 75:
```suggestion
            <!-- Do not generate module-info with moditect-maven-plugin:
```

mapper/orm/pom.xml
```diff
@@ -71,69 +71,31 @@
...
+                <artifactId>moditect-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>add-module-infos</id>
+                        <configuration>
+                            <module>
+                                <moduleInfo>
+                                    <name>${java.module.name}</name>
+                                    <exports>
+                                        org.hibernate.search.*.impl to org.hibernate.search.util.common, org.hibernate.search.engine, org.hibernate.search.backend.elasticsearch, org.hibernate.search.backend.elasticsearch.aws, org.hibernate.search.mapper.pojo, org.hibernate.search.mapper.javabean, org.hibernate.search.mapper.orm, org.hibernate.search.batch.jsr352.core, org.jboss.logging;
```
yrodiere, line 87: You definitely shouldn't need to export `impl` packages here. It's only necessary in the `hibernate-search-util-common` module.

```suggestion
                                        !org.hibernate.search.*.impl;
```

mapper/orm/pom.xml
```diff
@@ -71,69 +71,31 @@
...
+                                        *;
+                                    </exports>
+                                    <requires>
+                                        !cdi.api;
```
yrodiere, line 91: Why do you remove this, though? Please add a comment here, at least.

I suspect this means Hibernate Search won't work with CDI when it's in the modulepath?

mapper/pojo-base/pom.xml
```diff
@@ -47,4 +47,29 @@
...
+                <artifactId>moditect-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>add-module-infos</id>
+                        <configuration>
+                            <module>
+                                <moduleInfo>
+                                    <name>${java.module.name}</name>
+                                    <exports>
+                                        org.hibernate.search.*.impl to org.hibernate.search.util.common, org.hibernate.search.engine, org.hibernate.search.backend.elasticsearch, org.hibernate.search.backend.elasticsearch.aws, org.hibernate.search.mapper.pojo, org.hibernate.search.mapper.javabean, org.hibernate.search.mapper.orm, org.hibernate.search.batch.jsr352.core, org.jboss.logging;
```
yrodiere, line 64: You definitely shouldn't need to export `impl` packages here. It's only necessary in the `hibernate-search-util-common` module.
Please remove this?

pom.xml
```diff
@@ -339,7 +339,7 @@
         <version.coveralls.plugin>4.3.0</version.coveralls.plugin>
         <version.com.buschmais.jqassistant.plugin>1.8.0</version.com.buschmais.jqassistant.plugin>
         <version.docker.maven.plugin>0.34.1</version.docker.maven.plugin>
-        <version.moditect.plugin>1.0.0.Beta2</version.moditect.plugin>
+        <version.moditect.plugin>1.0.0-SNAPSHOT</version.moditect.plugin>
```
yrodiere, line 342: Please make sure to change this to a published version before we merge the PR.

util/common/pom.xml
```diff
@@ -37,5 +37,30 @@
...
+                <artifactId>moditect-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>add-module-infos</id>
+                        <configuration>
+                            <module>
+                                <moduleInfo>
+                                    <name>${java.module.name}</name>
+                                    <exports>
+                                        org.hibernate.search.*.impl to org.hibernate.search.util.common, org.hibernate.search.engine, org.hibernate.search.backend.elasticsearch, org.hibernate.search.backend.elasticsearch.aws, org.hibernate.search.mapper.pojo, org.hibernate.search.mapper.javabean, org.hibernate.search.mapper.orm, org.hibernate.search.batch.jsr352.core, org.jboss.logging;
```
yrodiere, line 54: Now *here* it's useful.

v5migrationhelper/engine/pom.xml
```diff
@@ -65,6 +65,8 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-javadoc-plugin</artifactId>
             </plugin>
+            <!-- Not generate module-info with moditect-maven-plugin:
```
yrodiere, line 68:
```suggestion
            <!-- Do not generate module-info with moditect-maven-plugin:
```

v5migrationhelper/orm/pom.xml
```diff
@@ -53,6 +53,8 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-javadoc-plugin</artifactId>
             </plugin>
+            <!-- Not generate module-info with moditect-maven-plugin:
```
yrodiere, line 56:
```suggestion
            <!-- Do not generate module-info with moditect-maven-plugin:
```

yrodiere (commented):
Thanks! My two cents.

I wonder if the moditect plugin configuration should be pulled up to `parents/public/pom.xml`, as it's almost always the same? We can always override it in the few modules that need an override.

yrodiere:
About this:

> Use `MethodHandles.Lookup` instance from Hibernate ORM, so that we don't need to open the client to `org.hibernate.search.mapper.orm` module

I suspect this will take time. And even when it's solved, users will still have to open their module to `org.hibernate.orm.core`.
Let's move this to a different ticket?

pom.xml
fax4ever: This request will always be a draft as long as a new Beta of ModiTect is not released.

fax4ever:
Thanks for the help @yrodiere, I did the changes.
Is there something else we can do here?
If not, let's wait for a new version of ModiTect!

mapper/orm/pom.xml
yrodiere: Bump.

If there is no `cdi.api` module, why do you need to mention it here? Is it automatically detected? In that case, it must exist, right?

Maybe you would rather need some kind of optional dependency? Something like [`requires static`](https://nipafx.dev/java-modules-optional-dependencies/#optional-dependencies-with-requires-static)?

parents/public/pom.xml
```diff
@@ -132,6 +132,37 @@
...
+                        <phase>package</phase>
+                        <goals>
+                            <goal>add-module-info</goal>
+                        </goals>
+                        <configuration>
+                            <jvmVersion>9</jvmVersion>
+                            <module>
+                                <moduleInfo>
+                                    <name>${java.module.name}</name>
+                                    <exports>
```
yrodiere, line 150: Some comments to explain this, maybe? We won't know what the `exports` are for even two weeks from now...

```suggestion
                                    <!-- Exports:
                                             - some impl packages to org.jboss.logging so that it can instantiate
                                               the generated Log/Messages classes
                                             - all impl packages to org.hibernate.engine so that it can instantiate built-in
                                               beans using reflection (e.g. annotation processors, bridges, binders)
                                             - all non-impl (api, spi) packages to users
                                     -->
                                    <exports>
```

util/common/pom.xml
```diff
@@ -37,5 +37,29 @@
...
+                <artifactId>moditect-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>add-module-infos</id>
+                        <configuration>
+                            <module>
+                                <moduleInfo>
+                                    <exports combine.self="override">
+                                        org.hibernate.search.*.impl to org.hibernate.search.util.common, org.hibernate.search.engine, org.hibernate.search.backend.elasticsearch, org.hibernate.search.backend.elasticsearch.aws, org.hibernate.search.mapper.pojo, org.hibernate.search.mapper.javabean, org.hibernate.search.mapper.orm, org.hibernate.search.batch.jsr352.core, org.jboss.logging;
+                                        *;
```
yrodiere, line 54: Why is this not necessary, while it is apparently necessary for other modules?

```suggestion
                                        !org.hibernate.search.*.impl;
                                        *;
```

fax4ever: Actually It is not necessary, because the exact same expression `org.hibernate.search.*.impl` is exported to a set of modules.
I'm going to remove it for the other modules too!

fax4ever:
Thanks to the new review @yrodiere, I made the changes.
I think that is the time to waiting for a new version of ModiTect.

fax4ever:
Full build: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2644/6/pipeline

fax4ever:
Another full build: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2644/8/

fax4ever:
The error on CI was about the EC2 infrastructure:
``` bash
EC2 (hibernate on AWS) - C5XL Containers (sir-zpz68iik) was marked offline: Node is being removed

Unable to create live FilePath for EC2 (hibernate on AWS) - C5XL Containers (sir-zpz68iik)
```

fax4ever:
Another full build: https://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-2644/.
Hoping to no longer have issue with the infrastructure. :P

fax4ever:
Done. Is there everything else to fix or improve here?
I know you approved, but I did other changes meanwhile...
Thanks

backend/elasticsearch/pom.xml
```diff
@@ -56,4 +56,30 @@
...
+                <executions>
+                    <execution>
+                        <id>add-module-infos</id>
+                        <configuration>
+                            <module>
+                                <moduleInfo>
+                                    <exports combine.self="override">
+                                        org.hibernate.search.*.logging.impl to org.jboss.logging;
+                                        org.hibernate.search.*.reporting.impl to org.jboss.logging;
+                                        org.hibernate.search.*.impl to org.hibernate.search.engine, gson;
```
yrodiere, line 74: I suspect that what we need for `gson` is probably an `opens` statements instead of an `exports` statements. Because `gson` may need reflection and access to private fields, which `exports` doesn't provide.

I suspect the only reason it currently works with just an `exports` is that we use our own type adapters (e.g. `PropertyMappingJsonAdapterFactory`) on all the serialized types. 

Would you mind replacing the export to `gson` with something like this? It will probably be more future-proof.

```
<opens>
org.hibernate.search.backend.elasticsearch.lowlevel.index.* to gson
</opens>
```

integrationtest/jdk/java-modules/src/main/java/module-info.java
```diff
@@ -16,7 +16,8 @@
 	requires org.hibernate.search.mapper.orm;
 	requires org.hibernate.search.backend.elasticsearch;
 
-	// TODO HSEARCH-3274 This should be re-exported transitively by org.hibernate.search.mapper.orm
+	// This should be re-exported transitively by org.hibernate.search.mapper.orm
+	// since org.hibernate.search.mapper.orm is still an automatic module
```
yrodiere, line 20:
```suggestion
	// but currently isn't, because org.hibernate.search.mapper.orm
	// is still an automatic module
```

yrodiere (commented):
Only a doubt about the `exports` to `gson`. Feel free to merge once you've dealt with that. Thanks!

backend/elasticsearch/pom.xml
fax4ever: Good idea!

fax4ever:
The last full build: https://ci.hibernate.org/job/hibernate-search/view/change-requests/job/PR-2644/11/

fax4ever:
Thanks @yrodiere I'll merge is as soon as the CI works

fax4ever:
Merged, thanks!

PR #2647, opened 2021-09-06
HSEARCH-4303 No automatic reindexing on insert/delete of contained entity without corresponding update in the containing entity

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4303


yrodiere:
The CI failures are just timeouts; I'll fix that in HSEARCH-4300.

Merging now, because I want this to be in the release.

PR #2649, opened 2021-09-07
HSEARCH-4300 Fix build timeout after 1 hour when testing against DB2

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4300

Also includes a few follow-up fixes for [HSEARCH-4303](https://hibernate.atlassian.net/browse/HSEARCH-4303), since we can now actually test on Oracle.

Some tests are down from 4+ minutes to ~10 seconds on DB2, because we do much fewer DDL operations. Hopefully that will be enough to bring test execution on DB2/Oracle back under 1 hour.


yrodiere:
Test execution on all databases: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2649/2/

yrodiere:
Looks like it was a success; when compared to the last build without timeout (which included fewer tests, that's why db2 still ran in under 1 hour):

* postgresql down from 8min30 to 5min
* mariadb down from 9min to 6min
* mysql down from 13min to 6min30
* mssql down from 10min to 6min
* oracle down from 16min to 8min
* db2 down from 36min to 13min

Merged!

PR #2652, opened 2021-09-13
HSEARCH-4266 Apply type condition outside user conditionals

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4266

It seems that the test is enough and we don't need to inspect the entities loaded by Hibernate ORM, as explained in the code comment.


yrodiere:
Merged, thanks!

fax4ever:
Thanks!

PR #2656, opened 2021-09-21
HSEARCH-4307 Hibernate Search version detection doesn't work on native images

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4307

Backport to 6.0: #2657

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/Version.java
+++ b/engine/src/main/java/org/hibernate/search/engine/Version.java
@@ -22,7 +22,8 @@ public final class Version {
 	 * @return A string representation of the version of Hibernate Search.
 	 */
 	public static String versionString() {
-		return Version.class.getPackage().getImplementationVersion();
+		// This implementation is replaced during the build with another one that returns the correct value.
+		return "UNKNOWN";
 	}
 
 	/**
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -36,4 +36,34 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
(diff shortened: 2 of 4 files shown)


engine/src/main/java/org/hibernate/search/engine/Version.java
```diff
@@ -22,7 +22,8 @@ private Version() {
 	 * @return A string representation of the version of Hibernate Search.
 	 */
 	public static String versionString() {
-		return Version.class.getPackage().getImplementationVersion();
+		// This implementation is replaced during the build with another one that returns the correct value.
+		return "UNKNOWN";
```
fax4ever, line 26: Maybe we can use the API `Version.class.getPackage().getImplementationVersion()` to get the current version for the test. Because I think tests are never executed as natives...

fax4ever (approved):
Just one idea to suggest...
Anyway LGTM

engine/src/main/java/org/hibernate/search/engine/Version.java
yrodiere: That wouldn't be useful for tests: `Version.class.getPackage().getImplementationVersion()` used not to work in tests, either. It was returning `null`.

I don't know why exactly, since it worked fine in real applications. Probably related to some classloading specifics of surefire/failsafe.

Anyway, it doesn't matter anymore, since that implementation is gone.

yrodiere:
Thanks!

The failure was caused by coveralls being down; I'll ignore and merge anyway.

PR #2661, opened 2021-09-28
HSEARCH-4240 Separate artifacts for compatibility with Jakarta EE 9.1 (Jakarta Persistence 3)

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4240

This creates additional Maven artifacts, which are copies of some of the current ones, except their artifact ID is appended with `-jakarta` (e.g. `hibernate-search-mapper-orm-jakarta`), and they use Jakarta APIs instead of Java EE APIs.

This is done through source transformation and recompilation. Dependencies of the new artifacts are different (pointing to Jakarta APIs instead of Java EE ones, as well as Hibernate ORM's `-jakarta` artifacts). References to `javax.` packages in the source code is replaced with references to `jakarta.` packages. Same for XML namespaces.

I used source transformation instead of bytecode transformation because:

1. It seems safer and easier to debug. If something is wrong, it likely won't compile, and then you just look at the generated source code to find out what's wrong exactly.
2. We don't need all the power that bytecode transformation provides, because we're don't have as much coupling to JPA as Hibernate ORM.
3. I hope to use the same solution for ORM6-specific artifacts, which will not only need replacements of packages (since ORM 6 will use Jakarta APIs), but also a few patches to address backwards incompatible SPIs or APIs. With source transformation, I could add patch files to the repository, and have the build apply these patches before compilation.

I also did the source transformation for integration tests, which guarantees that these new artifacts work correctly. To check that the source code actually uses Jakarta APIs, just run `mvn clean install -DskipTests`, then have a look at the code in `target/copied-sources`: this is the one that get compiled, then executed in tests.

Note that I did not add the new artifacts to the distribution ZIP, on purpose: it's too much hassle, and since this is an experimental feature anyway, it's unlikely to be needed by people who still don't use Maven in 2021.

### DIFF
--- a/documentation/src/main/asciidoc/reference/integrations.asciidoc
+++ b/documentation/src/main/asciidoc/reference/integrations.asciidoc
@@ -1,0 +1,56 @@
+[[integrations]]
+= Standards and integrations
+
+[[other-integrations-jakarta]]
+== Jakarta EE
+
+include::components/incubating-warning.asciidoc[]
+
+Hibernate Search includes experimental support for link:{jakartaUrl}[Jakarta EE],
+and it requires only one small change:
+when declaring the dependencies of your project,
+you must add `-jakarta` to some artifact identifiers.
+
+For example, an application using Hibernate ORM and the Elasticsearch backend
+will need to update its dependencies as follows:
+
+[source, XML, subs="+attributes"]
+----
+<dependency>
+   <groupId>org.hibernate</groupId>
+   <artifactId>hibernate-core-jakarta</artifactId> <!--1-->
+   <version>{hibernateVersion}</version>
+</dependency>
+<dependency>
+   <groupId>org.hibernate.search</groupId>
+   <artifactId></artifactId>
+   <artifactId>hibernate-search-mapper-orm-jakarta</artifactId> <!--2-->
+   <version>{hibernateSearchVersion}</version>
+</dependency>
+<dependency>
+   <groupId>org.hibernate.search</groupId>
+   <artifactId>hibernate-search-backend-elasticsearch</artifactId> <!--3-->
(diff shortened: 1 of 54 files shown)


documentation/src/main/asciidoc/reference/integrations.asciidoc
```diff
@@ -0,0 +1,56 @@
...
+<2> Replaces `hibernate-search-mapper-orm`.
+<3> No replacement necessary: this artifact does not rely on Java EE.
+
+All artifacts relying directly or indirectly on Java EE must be replaced with their Jakarta counterpart.
+This includes in particular:
+
+* https://in.relation.to/2021/06/04/hibernate-is-jakarta-jpa-2/#get-it[Hibernate ORM artifacts]
+* `hibernate-search-mapper-orm` => `hibernate-search-mapper-orm-jakarta`
+* `hibernate-search-mapper-orm-batch-jsr352-core` => `hibernate-search-mapper-orm-batch-jsr352-core-jakarta`
+* ...
```
fax4ever, line 46: I think that it is better to list all the artifacts that must be replaced with their Jakarta counterpart.

fax4ever (commented):
LGTM, just one comment

fax4ever:
> I used source transformation instead of bytecode transformation because:

this was great!

documentation/src/main/asciidoc/reference/integrations.asciidoc
yrodiere: True, but we will likely forget to add them here when we add more artifacts, so I'd rather be clear from the start that the list may be incomplete.

yrodiere:
Merged, thanks!

PR #2662, opened 2021-09-28
HSEARCH-4182 Improve detection of whether an OutboxEvent entity mapping is required

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4182

See tests in the last commit. Half of them were not passing before this PR.


engine/src/main/java/org/hibernate/search/engine/common/spi/SearchIntegrationEnvironment.java
```diff
@@ -0,0 +1,168 @@
...
+		this.classResolver = source.classResolver;
+		this.resourceResolver = source.resourceResolver;
+		this.serviceResolver = source.serviceResolver;
+		beanProvider = source.beanProvider;
+		beanResolver = source.beanResolver;
+	}
+
+	@Override
+	public void close() {
+		beanProvider.close();
```
fax4ever, line 88: Is `beanProvider` supposed to be nullable?

fax4ever (approved):
LGMT +1 as soos as Ci si fixed

engine/src/main/java/org/hibernate/search/engine/common/spi/SearchIntegrationEnvironment.java
yrodiere: You are right.

yrodiere:
Merged, thanks

PR #2664, opened 2021-09-29
HSEARCH-4240 Follow-up: fix testing on JDK8

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4240

This creates a separate parent POM for internal, non-integration-test artifacts.

This is needed for internal util modules, in particular:
those modules share a lot of characteristics with integration test
modules, but they definitely should be compiled with -release 8
so that we can freely reuse them when testing against JDK 8.
Integration tests, on the other hand, may be compiled with -release 11.


yrodiere:
Merging now because I want to fix the main build ASAP

PR #2665, opened 2021-09-30
HSEARCH-4253 JSON-format Elasticsearch mapping templates on the application side

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4253


fax4ever:
Full build: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2665/3/pipeline

documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
```diff
@@ -1018,6 +1018,60 @@ Hibernate Search will use this value to limit the returning hits size
...
+====
+
+[IMPORTANT]
+====
+When this property is used Hibernate Search will try to merge the mappings created by the framework
+for the mapped fields with the mappings provided by the user.
+In case of conflicts, which happens when some properties overlap, the mappings provided will take precedence.
+Which means that it is possible to break the mappings for the purposes of Hibernate Search functions.
+====
+
```
yrodiere, line 1057: Maybe also show what the merged mapping could look like in the example above?

```suggestion

In the example above, the resulting, merged mapping could look like this:

.Possible resulting mapping after merging the content of `custom/index-mappings.json` with the Hibernate Search mapping
====
[source, JSON, indent=0, subs="+callouts"]
----
include::{resourcesdir}/custom/index-settings-merged.json[]
----
====
```

**I urge you to check** in tests, but I think the mapping will look something like this:

```
{
  "properties":{
    "_entity_type":{ /* <1> */
      "type":"keyword",
      "index":false,
      "doc_values":true
    },
    "name":{ /* <2> */
      "type":"text",
      "index":true,
      "analyzer":"default",
      "term_vector":"no",
      "norms":true
    },
    "userField":{
      "type":"keyword",
      "index":true,
      "norms":true,
      "doc_values":true
    },
    "userObject":{
      "dynamic":"true",
      "type":"object"
    }
  }
}
```

You can add the following callouts:

```
<1> This property is always added by Hibernate Search for internal implementation purposes.
<2> This is a property generated by Hibernate Search because the indexed entity
has a `String name` property annotated with `@FullTextField`.
```

integrationtest/backend/elasticsearch/src/test/resources/custom-index-mappings/complex-conflicts.json
```diff
@@ -0,0 +1,69 @@
+{
```
yrodiere, line 1: Could you add comments where the conflicts are supposed to be?

I expected to find `searchField` in here, with a type of `object` instead of `text`, so as to create a conflict, but evidently that's not the case. I'm having a hard time finding what is supposed to conflict exactly.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/schema/management/ElasticsearchIndexSchemaManagerUpdateCustomMappingsIT.java
```diff
@@ -0,0 +1,185 @@
...
+				.withSchemaManagement( StubMappingSchemaManagementStrategy.DROP_ON_SHUTDOWN_ONLY )
+				.withIndexProperty( index.name(), ElasticsearchIndexSettings.SCHEMA_MANAGEMENT_MAPPINGS_FILE,
+						"custom-index-mappings/" + customMappingsFile
+				)
+				.withIndex( index )
+				.setup();
+
+		Futures.unwrappedExceptionJoin( index.schemaManager().createOrUpdate() );
+	}
+}
```
yrodiere, line 185: I believe tests (and implementation) of validation are missing?

To be clear, most of validation should already be here; you only need to add code to validate `AbstractTypeMapping#extraAttributes`. And, obviously, to add tests.

You can find an example of how to do that in `org.hibernate.search.backend.elasticsearch.validation.impl.IndexSettingsValidator#validate`. It's important to use `validateAllIgnoreUnexpected` here, because we don't care about extra mapping attributes.

yrodiere (changes requested):
Thanks. Here are a few comments. I think a few changes are required, though mostly in the documentation: I though you hadn't implemented merging, at first, but it turned out you did :)

fax4ever:
I just updated the request with a small fix and the rebase on top of the current master.
I still need to address all the comments.

fax4ever:
A full build to check this few changes: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2665/5/pipeline

PR #2670, opened 2021-09-30
HSEARCH-4282 Configurable transaction timeout for the outbox event processor in JTA environments

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4282


mapper/orm-coordination-database-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/databasepolling/cfg/HibernateOrmMapperDatabasePollingSettings.java
```diff
@@ -178,6 +198,7 @@ private Defaults() {
 		public static final boolean COORDINATION_PROCESSORS_INDEXING_ENABLED = true;
 		public static final int COORDINATION_PROCESSORS_INDEXING_POLLING_INTERVAL = 100;
 		public static final int COORDINATION_PROCESSORS_INDEXING_BATCH_SIZE = 50;
+		public static final Integer COORDINATION_PROCESSORS_INDEXING_TRANSACTION_TIMEOUT = null;
```
yrodiere, line 201: That's not really a default, that's the absence of default. Could you avoid exposing this, and just use an `orElse( null )` when you retrieve the value of the configuration property?

mapper/orm-coordination-database-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/databasepolling/cfg/HibernateOrmMapperDatabasePollingSettings.java
```diff
@@ -131,6 +131,24 @@ private HibernateOrmMapperDatabasePollingSettings() {
 	public static final String COORDINATION_PROCESSORS_INDEXING_BATCH_SIZE =
 			PREFIX + Radicals.COORDINATION_PROCESSORS_INDEXING_BATCH_SIZE;
 
+	/**
+	 * In the background indexing processor, the timeout for transactions processing outbox events.
+	 * <p>
+	 * Only available when {@link HibernateOrmMapperSettings#COORDINATION_STRATEGY} is
+	 * {@link CoordinationStrategyNames#DATABASE_POLLING}.
+	 * <p>
+	 * Only available when JTA transaction manager is configured.
```
yrodiere, line 140:
```suggestion
	 * Only effective when a JTA transaction manager is configured.
```

mapper/orm-coordination-database-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/databasepolling/cfg/HibernateOrmMapperDatabasePollingSettings.java
```diff
@@ -131,6 +131,24 @@ private HibernateOrmMapperDatabasePollingSettings() {
...
+	 * <p>
+	 * Expects a positive Integer value, such as {@code 50},
+	 * or a String that can be parsed into such Integer value.
+	 * <p>
+	 * Defaults to {@link Defaults#COORDINATION_PROCESSORS_INDEXING_BATCH_SIZE}.
+	 * <p>
+	 * the value <code>null</code> means to use the default value, which value depends on the transaction manager in use.
```
yrodiere, line 147:
```suggestion
	 * When using JTA and this property is not set, Hibernate Search will use whatever default transaction timeout configured in the JTA transaction manager.
```

mapper/orm-coordination-database-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/databasepolling/cfg/HibernateOrmMapperDatabasePollingSettings.java
```diff
@@ -131,6 +131,24 @@ private HibernateOrmMapperDatabasePollingSettings() {
...
 
+	/**
+	 * In the background indexing processor, the timeout for transactions processing outbox events.
+	 * <p>
+	 * Only available when {@link HibernateOrmMapperSettings#COORDINATION_STRATEGY} is
+	 * {@link CoordinationStrategyNames#DATABASE_POLLING}.
+	 * <p>
+	 * Only available when JTA transaction manager is configured.
+	 * <p>
+	 * Expects a positive Integer value, such as {@code 50},
```
yrodiere, line 142:
```suggestion
	 * Expects a positive Integer value in seconds, such as {@code 10},
```

yrodiere (commented):
Thanks. Here are a few comments.

mapper/orm-coordination-database-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/databasepolling/cfg/HibernateOrmMapperDatabasePollingSettings.java
```diff
@@ -131,6 +131,24 @@ private HibernateOrmMapperDatabasePollingSettings() {
 	public static final String COORDINATION_PROCESSORS_INDEXING_BATCH_SIZE =
 			PREFIX + Radicals.COORDINATION_PROCESSORS_INDEXING_BATCH_SIZE;
 
+	/**
+	 * In the background indexing processor, the timeout for transactions processing outbox events.
+	 * <p>
+	 * Only available when {@link HibernateOrmMapperSettings#COORDINATION_STRATEGY} is
+	 * {@link CoordinationStrategyNames#DATABASE_POLLING}.
```
yrodiere, line 138: I removed `CoordinationStrategyNames` in a PR that I just merged, so you will need to change this:

```suggestion
	 * {@value #COORDINATION_STRATEGY_NAME}.
```

mapper/orm-coordination-database-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/databasepolling/impl/OutboxEventBackgroundProcessor.java
```diff
@@ -107,7 +110,7 @@ public DatabasePollingOutboxWorker(SessionFactoryImplementor sessionFactory) {
 			}
 
 			try ( SessionImplementor session = (SessionImplementor) mapping.sessionFactory().openSession() ) {
-				transactionHelper.begin( session, null );
+				transactionHelper.begin( session, transactionTimeout );
```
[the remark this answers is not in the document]
fax4ever: > This could do with a test in JTA mode? I think we have some already, we can just copy/paste them and adapt them.
> 
> You could some custom ORM listener that introduces an artificial delay which would make automatic indexing fail with the default transaction timeout, but succeed with the custom one configured though your new configuration property. E.g. something similar to `org.hibernate.search.util.impl.integrationtest.mapper.orm.TimeoutLoadingListener`, but one that would always be registered (you can use `hibernate.session.events.auto` for that).
> 
> Be sure to test both cases, as simply having automatic indexing succeed doesn't prove anything unless you know it would have failed without your setting.

I don't know if we have already without Spring...

fax4ever:
Thanks for the review. I think I've addressed all the comments.
If there is anything else to do, I'm here.

integrationtest/mapper/orm-spring/src/test/java/org/hibernate/search/integrationtest/spring/jta/TransactionTimeoutJtaAndSpringOutboxIT.java
```diff
@@ -0,0 +1,73 @@
...
+		snert.setAge( 99 );
+		snert.setCool( Boolean.TRUE );
+		snertDAO.persist( snert );
+
+		await( "Waiting for indexing assertions" )
+				.untilAsserted( () -> assertThat( TimeoutFailureCollector.EXCEPTIONS ).isNotEmpty() );
+
+		Throwable exception = TimeoutFailureCollector.EXCEPTIONS.iterator().next();
+		Throwable cause = exception.getCause().getCause().getCause();
+		assertThat( cause ).hasMessageContaining( "transaction timed out" );
```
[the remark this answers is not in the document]
fax4ever: I agree with the fact that a user usually uses this setting to increase the timeout so that the transactions will be executed. 
However, according to the [Bitronix documentation](https://github.com/bitronix/btm/wiki/Transaction-manager-configuration#bitronixtmtimerdefaulttransactiontimeout) the default timeout is 60 seconds.
Given that, I don't know if it is the case to introduce a sleep of 60 seconds to trigger a timeout using the default.
To be honest, I think that testing the opposite case, that is having a timeout setting the property to a lower value and not having a timeout with the default, would be the same in terms of verifying the fact that the property works. That is in my opinion what we really need to test here.
Doing that we would need a 1 second sleep that is better.

PR #2675, opened 2021-10-04
HSEARCH-4326 Avoid null values for DirtinessDescriptor.dirtyPath()

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4326


fax4ever (approved):
Thanks, LGTM!
I'm merging it

PR #2680, opened 2021-10-08
HSEARCH-4341 Fix GitHub action build taking 6+ hours to complete

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4341

First attempt. Let's see how it goes.


yrodiere:
Added a potential workaround from here: https://github.com/actions/virtual-environments/issues/1499
It should work better now.

PR #2686, opened 2021-10-18
HSEARCH-4348 Check ORM declared property fields by the ORM member name

olliefreeman:
HSEARCH-4348 Check ORM declared property fields by the ORM member name if not found by the propertyname

When using groovy traits the propertyName will not match with the ORM member name.
This addition will try to find the Member using the propertyName but if it cant then it will try by 
using the ORM metamodel property name.

https://hibernate.atlassian.net/browse/HSEARCH-4348


olliefreeman:
wrt the comment
1. will do, thought main was v5 hence reason i put this onto 6.0
2. at the moment im updating the main grails plugin for HS and using another project to test the full functionality as such its a little hard to provide something to reproduce as everything is in snapshot/in-progress status, which is why i provided the screenshots into the issue ticket.
i'll see if i can manually create a simple grails app with manual implementation of the HS model which should throw out the error.

yrodiere:
@olliefreeman This patch was included in the release of Hibernate Search 6.1.0.Beta1 today. Any luck making Grails work with that? Do you need a backport to 6.0?

PR #2688, opened 2021-10-21
HSEARCH-4351 Avoid unnecessary Thread.sleep in database polling coordination tests

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4351


yrodiere:
Thanks

PR #2697, opened 2021-11-02
HSEARCH-4290 Add an example to the documentation using ".bridgedElement().name()" in a property binder

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4290

### DIFF
--- a/documentation/src/main/asciidoc/reference/mapper-orm-bridge-propertybridge.asciidoc
+++ b/documentation/src/main/asciidoc/reference/mapper-orm-bridge-propertybridge.asciidoc
@@ -263,13 +263,46 @@ include::{sourcedir}/org/hibernate/search/documentation/mapper/orm/bridge/proper
 include::components/incubating-warning.asciidoc[]
 
 The context passed to the property binder's `bind` method
-exposes a `bridgedElement()` method that gives access to metadata about the property being bound,
-in particular its name and type.
+exposes a `bridgedElement()` method that gives access to metadata about the property being bound.
 
-The metadata can also be used to inspect the type of the property in details:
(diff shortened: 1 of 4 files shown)


documentation/src/main/asciidoc/reference/mapper-orm-bridge-propertybridge.asciidoc
```diff
@@ -263,13 +263,46 @@ include::{sourcedir}/org/hibernate/search/documentation/mapper/orm/bridge/proper
...
+getting the property name and using it as a field name.
+
+.Naming a field after the property being bound in a `PropertyBinder`
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/mapper/orm/bridge/propertybridge/bridgedelement/InvoiceLineItemsSummaryBinder.java[tags=include]
+----
+<1> Use the binding context to get the bridged element.
+<2> Use the name of the property as the name of a declared field.
```
fax4ever, line 288: It is almost clear, but I would write something more explicit, like:
> Use the name of the property to get the access to the `objectField`

fax4ever (approved):
LGTM,
just an optional comment

documentation/src/main/asciidoc/reference/mapper-orm-bridge-propertybridge.asciidoc
yrodiere: That's not right, though. We're not really "getting access" to the field, we're defining an index field. It doesn't exist before this line.

I'll clarify that I mean "index" field.

yrodiere:
Merged, thanks

PR #2699, opened 2021-11-04
HSEARCH-4363 NPE on closing SingletonTask still running a work

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4363

### DIFF
--- a/engine/src/test/java/org/hibernate/search/engine/backend/orchestration/spi/SingletonTaskTest.java
+++ b/engine/src/test/java/org/hibernate/search/engine/backend/orchestration/spi/SingletonTaskTest.java
@@ -1,0 +1,97 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.engine.backend.orchestration.spi;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+
+import org.hibernate.search.engine.reporting.EntityIndexingFailureContext;
+import org.hibernate.search.engine.reporting.FailureContext;
+import org.hibernate.search.engine.reporting.FailureHandler;
+import org.hibernate.search.util.impl.test.annotation.TestForIssue;
+
+import org.junit.Test;
+
+import org.awaitility.Awaitility;
+
+import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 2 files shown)


engine/src/test/java/org/hibernate/search/engine/backend/orchestration/spi/SingletonTaskTest.java
```diff
@@ -0,0 +1,101 @@
...
+		testSubject.ensureScheduled();
+
+		// wait a bit to be sure that the task will be scheduled,
+		// but less than the time of the task to be completed
+		Thread.sleep( TEST_WORK_DURATION / 2 );
+
+		testSubject.stop();
+
+		// wait to be sure that the task has been completed
+		Thread.sleep( TEST_WORK_DURATION );
```
yrodiere, line 49: I'm surprised by this from you, who frown on my adding a `Thread.sleep` to tests every single time :)

Yes, the sleep duration is small so it won't affect build time much, but precisely because it's small, it's likely to behave differently from what you want in case the machine slows down for some reason. Hell, even just a garbage collection could block the JVM for more than 10ms.

Fortunately, you're doing a unit test, so you don't need `Thread.sleep` here. You can just have the task block until you unblock it manually (after the `stop`).

E.g.:

```suggestion
		testSubject.ensureScheduled();

		Awaitiltiy.await().until( () -> testBlockingWorker.hasStarted() );

		testSubject.stop();
		
		testBlockingWorker.future.complete( null );
```

```
	private static final class TestBlockingWorker implements SingletonTask.Worker {
		private volatile boolean hasStarted = false;
		private final CompletableFuture<?> future = new CompletableFuture<>();

		@Override
		public CompletableFuture<?> work() {
			boolean hasStarted = true;
			future.join();
			return future;
		}

		@Override
		public void complete() {
			// Nothing to do.
		}
	}
```

engine/src/test/java/org/hibernate/search/engine/backend/orchestration/spi/SingletonTaskTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+	private SingletonTask testSubject = new SingletonTask(
+			"TestSingletonTask",
+			new TestWorker(),
+			new TestScheduler(),
+			new TestFailureHandler()
+	);
+
+	@Test
+	public void test() throws Exception {
```
yrodiere, line 39: Maybe use a more meaningful test name?

```suggestion
	public void stopWhileBlockingWorkInProgress() throws Exception {
```

yrodiere (changes requested):

fax4ever:
As I told you I think that sometimes I have some rush to get things done, even if there is no reason to do so. In fact, the test could be easily and elegantly implemented using a completable future. Sorry. I made the changes here and in the backport PR.

yrodiere (approved):

fax4ever:
Merged, thanks!

PR #2701, opened 2021-11-08
HSEARCH-4273 MassIndexer.cacheMode is not honored

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4273

### DIFF
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingCachingIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingCachingIT.java
@@ -39,7 +39,6 @@ import org.hibernate.stat.Statistics;
 
 import org.junit.Before;
 import org.junit.ClassRule;
-import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.MethodRule;
@@ -135,7 +134,7 @@ public class MassIndexingCachingIT {
 	}
 
 	@Test
-	@Ignore("HSEARCH-4273: MassIndexer.cacheMode is not honored")
+	@TestForIssue(jiraKey = "HSEARCH-4273")
 	public void explicit_get() {
 		setupHolder.runNoTransaction( session -> {
 			SearchSession searchSession = Search.session( session );
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingCachingIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingCachingIT.java
@@ -6,16 +6,23 @@
  */
 package org.hibernate.search.integrationtest.mapper.orm.massindexing;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Fail.fail;
 import static org.assertj.core.api.SoftAssertions.assertSoftly;
 
+import java.util.Arrays;
+import java.util.List;
 import javax.persistence.Entity;
 import javax.persistence.Id;
 import javax.persistence.SharedCacheMode;
 import javax.persistence.Table;
+import javax.persistence.TypedQuery;
 
 import org.hibernate.CacheMode;
+import org.hibernate.Session;
 import org.hibernate.cfg.AvailableSettings;
+import org.hibernate.jpa.QueryHints;
+import org.hibernate.query.Query;
 import org.hibernate.search.engine.backend.work.execution.DocumentCommitStrategy;
 import org.hibernate.search.engine.backend.work.execution.DocumentRefreshStrategy;
 import org.hibernate.search.mapper.orm.Search;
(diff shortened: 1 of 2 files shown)


yrodiere (changes requested):
That's an interesting solution, but:

1. I'm a bit worried that you may have effectively changed the behavior of the mass indexer to cache the results of its own queries (in the query cache). That would be disastrous for performance, as we would put cache one cache entry for every single query executed by the mass indexer, and that entry would never be reused.
   Could you add an assertion to check that the mass indexer doesn't add any entry to the query cache?
   Remember, we're interested in the second level entity cache here, not the query cache.
2. I'm also concerned that your test may be passing only because the query you used to populate the cache is the exact same query used by the mass indexer to load entities. Could you alter that query so that it's different from the one used by the mass indexer? E.g. add `AND true = true` to the `WHERE` clause, or something like that.

documentation/src/main/asciidoc/reference/mapper-orm-indexing-jsr352.asciidoc
```diff
@@ -87,6 +87,9 @@ place after indexing.
 The default is `IGNORE`, and it will be the most efficient choice in most cases,
 but using another mode such as `GET` may be more efficient if many of the entities being indexed
 refer to a small set of other entities.
+Enabling caches has an effect only if the entity has some lazy fields and/or the entity id is the document id too.
```
yrodiere, line 90: I don't think lazy fields have anything to do with this?

```suggestion
Enabling caches has an effect only if the entity id is also the document id, which is the default.
```

documentation/src/main/asciidoc/reference/mapper-orm-indexing-jsr352.asciidoc
```diff
@@ -87,6 +87,9 @@ place after indexing.
...
+`PUT` or `Normal` values may lead to a bad performance, because all the entities are also loaded into
```
yrodiere, line 91:
```suggestion
`PUT` or `NORMAL` values may lead to bad performance, because all the entities are also loaded into
```

documentation/src/main/asciidoc/reference/mapper-orm-indexing-jsr352.asciidoc
```diff
@@ -87,6 +87,9 @@ place after indexing.
 The default is `IGNORE`, and it will be the most efficient choice in most cases,
 but using another mode such as `GET` may be more efficient if many of the entities being indexed
 refer to a small set of other entities.
```
yrodiere, line 89: Looks like this part was wrong:

```suggestion
but using another mode such as `GET` may be more efficient if many of the entities being indexed
are already present in the Hibernate ORM second-level cache before mass indexing.
```

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingCachingIT.java
```diff
@@ -150,19 +166,76 @@ public void explicit_get() {
...
+
+	private Query cachedQuery(Session session, CacheMode cacheMode) {
+		Query<IndexedEntity> query = session.createQuery(
+				"select e from IndexedEntity e where e.id in (:ids)",
+				IndexedEntity.class
+		);
+
+		query.setParameter( "ids", Arrays.asList( 1, 2, 3 ) )
+				.setCacheMode( cacheMode )
+				.setHint( QueryHints.HINT_CACHEABLE, true );
```
yrodiere, line 234: This shouldn't be necessary? We are not supposed to rely the query cache anymore.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/loading/impl/HibernateOrmMassEntityLoader.java
```diff
@@ -62,4 +57,24 @@ public void load(List<I> identifiers) throws InterruptedException {
...
+
+	private List<E> queryByIds(List<I> identifiers) {
+		return typeQueryLoader.createLoadingQuery( session, ID_PARAMETER_NAME )
+				.setParameter( ID_PARAMETER_NAME, identifiers )
+				.setCacheMode( options.cacheMode() )
+				.setLockOptions( LockOptions.NONE )
+				.setCacheable( false )
+				.setHibernateFlushMode( FlushMode.MANUAL )
+				.setFetchSize( identifiers.size() )
+				.setHint( QueryHints.HINT_CACHEABLE, true )
```
yrodiere, line 77: I don't think this is necessary? And it's probably harmful for performance, since we will cache query results but never use the query cache.

Also, it directly contractics the `setCacheable( false )` just above.

```suggestion
```

yrodiere (changes requested):
Just a few corrections, then we can merge. Thanks.

documentation/src/main/asciidoc/reference/mapper-orm-indexing-jsr352.asciidoc
fax4ever: Actually, I think that there could be some benefit, in the case the entity id was not the document id, but the entity had some lazy fields. But let's not confuse the user with that ;)

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingCachingIT.java
fax4ever: You're right. I seems that we don't need it anymore!

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/loading/impl/HibernateOrmMassEntityLoader.java
fax4ever: I'm doing the change, but probably this flag would help in hitting the second level cache in the case I mentioned above:
1. property id != document id
2. entity has some lazy fields
But without a test we cannot say that for sure!
Moreover, we could handle caching in case the document id was the natural id, using the natural id multiloading.

fax4ever:
Thanks for the review @yrodiere!

yrodiere (approved):
LGTM, thanks

fax4ever:
Merged, thanks!

PR #2704, opened 2021-11-19
HSEARCH-4366 Add context to exceptions thrown during processing of POJOs (for indexing or reindexing resolution)

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4366

This simply adds context to exception thrown during processing, i.e. like this:

```
org.hibernate.search.util.common.SearchException: HSEARCH700083: Exception while building document for entity 'RootEntity#1': Simulated failure
Context: path '.containedNoContainer<no value extractors>.containedInContainer'
```

The path representation is far from perfect, but that's a more general problem. I'd say it's better than nothing: at least it gives a general idea of where the problem is in user code.


fax4ever:
LGTM, thank you!

PR #2705, opened 2021-11-22
HSEARCH-4365 remove of entity with IndexedEmbedded collections cause LazyInitializationException during indexing

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4365


fax4ever:
LGTM. Merged, thanks!

PR #2707, opened 2021-11-23
HSEARCH-4369 The outbox event processors apply the polling interval even when there are still events to process

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4369


fax4ever:
LGMT. Thanks, merged!

PR #2710, opened 2021-11-25
HSEARCH-4370 Event processing triggers deadlock with Oracle database

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4370


fax4ever:
Merged, thanks!

PR #2711, opened 2021-11-27
HSEARCH-4373 Use a java service to set the coordinationStrategyExpectations

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4373

### DIFF
--- a/util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/OrmSetupHelper.java
+++ b/util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/OrmSetupHelper.java
@@ -12,9 +12,11 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.ServiceLoader;
 
 import org.hibernate.SessionFactory;
 import org.hibernate.dialect.Dialect;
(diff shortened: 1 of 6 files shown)


fax4ever:
A default|database build: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2711/4/pipeline

fax4ever:
Another build: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2711/5/pipeline

util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/OrmSetupHelper.java
```diff
@@ -24,20 +26,33 @@
...
+	}
+
 	public static void defaultAutomaticIndexingStrategy(
 			CoordinationStrategyExpectations coordinationStrategyExpectations) {
 		OrmSetupHelper.defaultCoordinationStrategyExpectations = coordinationStrategyExpectations;
 	}
 
-	private static CoordinationStrategyExpectations defaultCoordinationStrategyExpectations =
-			CoordinationStrategyExpectations.defaults();
+	private static CoordinationStrategyExpectations defaultCoordinationStrategyExpectations;
```
yrodiere, line 55: I think this can now be `final`?

```suggestion
	private static final CoordinationStrategyExpectations defaultCoordinationStrategyExpectations;
```

And maybe we should adjust the case accordingly (`DEFAULT_COORDINATION_STRATEGY_EXPECTATIONS`).

yrodiere (approved):
Great catch @fax4ever , sorry I didn't notice that when I introduced `ReusableOrmSetupHolder`...

Only a minor comment here, otherwise it looks good to me.

util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/OrmSetupHelper.java
yrodiere: Actually, let's merge as-is; I need to push a related change afterwards, I'll do the change there.

PR #2719, opened 2021-11-30
HSEARCH-4376 Document Elasticsearch normalizing regexps as a bug

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4376


fax4ever:
Merged, thanks!

PR #2774, opened 2021-12-08
HSEARCH-4401 Lucene backend ignores limit when offset + limit is higher than Integer.MAX_VALUE instead of reporting that it cannot handle it

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4401

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearcherImpl.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearcherImpl.java
@@ -7,6 +7,7 @@
 package org.hibernate.search.backend.lucene.search.query.impl;
 
 import java.io.IOException;
+import java.lang.invoke.MethodHandles;
 import java.util.Map;
 
 import org.hibernate.search.backend.lucene.logging.impl.Log;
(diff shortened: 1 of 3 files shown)


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/query/impl/LuceneSearcherImpl.java
```diff
@@ -165,7 +165,7 @@ private LuceneCollectors buildCollectors(IndexSearcher indexSearcher, IndexReade
 	}
 
 	private int getMaxDocs(IndexReader reader, int offset, Integer limit) {
-		if ( limit == null ) {
+		if ( limit == null || (long) offset + limit >= Integer.MAX_VALUE ) {
```
yrodiere, line 168: Why not just `limit.equals( Integer.MAX_VALUE )`? Do you also want to ignore the limit when using `offset = 5` and `limit = Integer.MAX_VALUE - 5`?

If you want to take the offset into account, it would be better to take it into account in the Elasticsearch backend as well.

yrodiere (commented):
I'm not sure I want to encourage such liberal use of magic values in user code, but I guess it won't hurt us.

One thing, though: you are probably doing this to address a problem in the ORM 6 upgrade. Could you explain where the problem was?

Because, we could also just interpret `Integer.MAX_VALUE` as `null` in the ORM mapper.

yrodiere (approved):

fax4ever:
Thanks for the review!

yrodiere:
@fax4ever CI failed. Probably because of the null derereference reported by lgtm (see above).

fax4ever:
> @fax4ever CI failed. Probably because of the null derereference reported by lgtm (see above).

Right! I think I fixed it.

fax4ever:
Merged, thanks!

PR #2820, opened 2022-01-05
HSEARCH-4362 Customizable optional operators for the regexp query

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4362

### DIFF
--- a/util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/stub/backend/types/impl/StubIndexValueFieldType.java
+++ b/util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/stub/backend/types/impl/StubIndexValueFieldType.java
@@ -68,11 +68,14 @@ public final class StubIndexValueFieldType<F>
 					PredicateTypeKeys.EXISTS,
 					PredicateTypeKeys.PHRASE,
 					PredicateTypeKeys.WILDCARD,
-					PredicateTypeKeys.REGEXP,
 					PredicateTypeKeys.TERMS,
 					PredicateTypeKeys.SPATIAL_WITHIN_CIRCLE,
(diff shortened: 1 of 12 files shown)


fax4ever:
Run `default|elasticsearch` build to test more ES versions: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2820/2/pipeline

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/stub/backend/types/impl/StubIndexValueFieldType.java
```diff
@@ -61,7 +61,7 @@ public void apply(StubIndexSchemaDataNode.Builder builder) {
 		public Builder(Class<F> valueClass) {
 			super( valueClass );
 			stubFactories(
-					new StubSearchPredicate.Factory(),
+					new StubSearchPredicate.RegexpFactory(),
 					PredicateTypeKeys.NESTED,
 					PredicateTypeKeys.MATCH,
 					PredicateTypeKeys.RANGE,
```
yrodiere, line 67: You're assigning the regexp predicate factory to all sorts of predicates that are not the regexp predicate.

The right way to do what you want would be to remove PredicateTypeKeys.REGEXP from the list, and add a call to `stubFactories(new StubSearchPredicate.RegexpFactory(), PredicateTypeKeys.REGEXP)`.

Though TBH, this looks like dead code, so it's more for consistency than anything else.

yrodiere (changes requested):
Thanks. Here are a few comments.

fax4ever:
Thanks for the review @yrodiere!
I tried to address the requested changes.

yrodiere (approved):

yrodiere:
Merged, thanks!

PR #2824, opened 2022-01-07
HSEARCH-4428 Add missing classes to org.hibernate.search.backend.elasticsearch.gson.spi.GsonClasses

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4428


fax4ever (approved):
LGTM!

PR #2825, opened 2022-01-07
HSEARCH-4429 Change the limit of the RAM usage by ES containers

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4429

On my local MacOS it worked
Let see if it does on Git actions


yrodiere (approved):
LGTM, feel free to merge once CI is green.

yrodiere:
There seems to be other problems with GH Actions; I'll go ahead and merge, then we can have a look at the remaining problems.

fax4ever:
Makes sense, thank you!

PR #2833, opened 2022-01-10
HSEARCH-4330 Configurable mapping for entities added by the database-polling coordination strategy

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-4330


yrodiere:
Looks good, thanks. As I said above, I pushed an additional commit to simplify how the configuration property source is accessed.

Let's merge once CI agrees.

PR #2838, opened 2022-01-12
HSEARCH-4437 Fix very long Maven download timeouts when building with GitHub Actions

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4437


yrodiere:
Actually, I don't know if changes to GH actions workflow files take effect in pull requests. I suppose I have to merge into main in order to test the fix.

I'll merge.

PR #2841, opened 2022-01-13
HSEARCH-3277 Separate artifacts for compatibility with Hibernate ORM 6 (Jakarta Persistence only, no Java EE)

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3277


yrodiere:
Something not working... but on CI only. I'll re-open the PR when I find out what this is.

PR #2920, opened 2022-03-03
HSEARCH-4498 Fluent definition of composite projections

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4498


yrodiere:
I'll submit another PR with a different API, more appropriate if we want to use it in `nested` projections (HSEARCH-3943).

PR #2933, opened 2022-03-08
HSEARCH-4503 Disable version logging through system properties

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4503

~~Draft until we know for sure which property key we'll want to use; see https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/INFO.20messages.20on.20native-image~~ => Jboss-threads went with `jboss.log-version`, so we're going to go with that. See https://github.com/jbossas/jboss-threads/pull/120/files#diff-f7179ffcce5656c32a94e972f59059723d7d9d1c53bc3274c1d7bdda0b6a212aR61

### DIFF
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/bootstrap/BootstrapLogsIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/bootstrap/BootstrapLogsIT.java
@@ -19,6 +19,8 @@ import org.hibernate.search.mapper.pojo.mapping.definition.annotation.Indexed;
 import org.hibernate.search.mapper.pojo.mapping.definition.annotation.IndexedEmbedded;
 import org.hibernate.search.util.impl.integrationtest.common.rule.BackendMock;
 import org.hibernate.search.util.impl.integrationtest.mapper.orm.OrmSetupHelper;
+import org.hibernate.search.util.impl.test.SystemHelper;
+import org.hibernate.search.util.impl.test.SystemHelper.SystemPropertyRestorer;
 import org.hibernate.search.util.impl.test.annotation.TestForIssue;
 import org.hibernate.search.util.impl.test.rule.ExpectedLog4jLog;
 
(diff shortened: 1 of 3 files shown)


integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/bootstrap/BootstrapLogsIT.java
```diff
@@ -101,6 +103,27 @@ public void version() {
...
+	@TestForIssue(jiraKey = "HSEARCH-4503")
+	public void versionLoggingDisabled() {
+		String propertyKey = "org.hibernate.search.version";
+		String expectedHibernateSearchVersion = System.getProperty( propertyKey );
+		if ( expectedHibernateSearchVersion == null ) {
+			throw new IllegalStateException( "This test cannot be executed, because system property '"
+					+ propertyKey + "' was not defined." );
+		}
+
+		try ( SystemPropertyRestorer systemPropertyChange = SystemHelper.setSystemProperty( "jboss.log-versions", "false" ) ) {
```
gastaldi, line 116:
```suggestion
		try ( SystemPropertyRestorer systemPropertyChange = SystemHelper.setSystemProperty( HibernateOrmMapperSpiSettings.JBOSS_LOG_VERSIONS, "false" ) ) {
```

yrodiere: Thanks, but actually that was on purpose.

We've been bitten in the past by tests using property key constants:

* sometimes we'd update a constant (either intentionally or by mistake) but forget to update documentation or to announce the backwards-incompatible change. The failing test would at least give us a heads-up in such case.
* or in other cases, the constant was not set to the value we intended, but the tests passed anyway, because, well, we'd use the same constant in the test and implementation. If we hard-code the property key in the test, this cannot happen.

So, yeah, I'd rather keep hardcoded strings in tests unless there's a specific reason to avoid that.

PR #2963, opened 2022-03-28
HSEARCH-4519 Update link markup in Getting Started

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4519

- use asciidoc link markup


yrodiere:
Good catch, thanks! I'll backport to older branches as well.

PR #2964, opened 2022-03-28
HSEARCH-4489 Add matchNone() predicate

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4489

- add matchNone() to DSL
- update documentation to mention new predicate

I wasn't able to think of some simple yet, at the same time, meaningful use case for this new predicate, hence if anyone has some suggestions would be happy to add more tests around it.

### DIFF
--- a/documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
+++ b/documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
@@ -67,6 +67,19 @@ include::{sourcedir}/org/hibernate/search/documentation/search/predicate/Predica
 * The score of a `matchAll` predicate is constant and equal to 1 by default,
 but can be <<search-dsl-predicate-common-boost,boosted with `.boost(...)`>>.
 
+[[search-dsl-predicate-match-none]]
+== `matchNone`: match no documents
+
+The `matchNone` predicate is the inverse of `matchAll` and matches no documents.
+
+.Matching no documents
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java[tags=matchNone]
+----
+====
+
 [[search-dsl-predicate-id]]
 == `id`: match a document identifier
 
--- a/engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/SearchPredicateFactory.java
+++ b/engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/SearchPredicateFactory.java
@@ -44,6 +44,14 @@ public interface SearchPredicateFactory {
 	MatchAllPredicateOptionsStep<?> matchAll();
 
 	/**
+	 * Match none of the documents.
+	 *
+	 * @return The initial step of a DSL where the "match none" predicate can be defined.
+	 * @see MatchNonePredicateFinalStep
+	 */
+	MatchNonePredicateFinalStep matchNone();
+
+	/**
 	 * Match documents where the identifier is among the given values.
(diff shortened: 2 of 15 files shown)


documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -67,6 +67,19 @@ include::{sourcedir}/org/hibernate/search/documentation/search/predicate/Predica
...
+The `matchNone` predicate is the inverse of `matchAll` and matches no documents.
+
+.Matching no documents
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java[tags=matchNone]
+----
+====
+
```
yrodiere, line 82: Sorry, I didn't include the original use case in the ticket; here it is: https://discourse.hibernate.org/t/fail-fast-predicate/6062?u=yrodiere

The best solution to avoid running a search query is simply to not call `.fetch()` at all, but in some cases people will have methods dedicated to the creation of inner predicates, and will not be able to prevent the query to run; in that case, they can just create a `matchNone` predicate and the end result will be the same (though it will unfortunately involve some pointless I/O).

So, as you can see, it's a very specific use case, for advanced users. I'm not sure we need further explanation than what you did here, but if you think you can add something clear and concise, feel free to give it a shot :)

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/SearchPredicateFactory.java
```diff
@@ -43,6 +43,14 @@
 	 */
 	MatchAllPredicateOptionsStep<?> matchAll();
 
+	/**
+	 * Match none of the documents.
+	 *
+	 * @return The initial step of a DSL where the "match none" predicate can be defined.
+	 * @see MatchAllPredicateOptionsStep
```
yrodiere, line 50:
```suggestion
	 * @see MatchNonePredicateFinalStep
```

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/MatchNonePredicateSpecificsIT.java
```diff
@@ -0,0 +1,67 @@
...
+		initData();
+	}
+
+	@Test
+	public void matchNone() {
+		assertThatQuery( index.query()
+				.where( SearchPredicateFactory::matchNone ) )
+				.hasNoHits();
+	}
+
```
yrodiere, line 51: Maybe just add another test with a bool predicate containing one predicate that does match a document, and also the `matchNone` predicate, to check that the resulting bool predicate won't match anything?

yrodiere (commented):
Well this is a nice surprise, thanks!

I added a few comments below.

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
marko-bekhta: Hmm, that's an interesting case...  Knew of `match_all` usage within dynamic filters but not the `match_none`. nice :smiley: 
As for the text, I gave it a try ... aaaand the text becomes too cumbersome :smile: but as you've said - it's for advanced users. So those who need it would know its meaning, and others are probably safer without using it :smiley:

marko-bekhta:
haha, well I thought that the links update is not a significant enough change to add, so found this ticket :smile: 

applied the suggested changes and rebased the branch.

PR #2969, opened 2022-03-31
HSEARCH-4522 Use JDK17 by default when building Hibernate Search

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4522


yrodiere:
Build against all supported JDKs: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-2969/3/pipeline

PR #3034, opened 2022-05-04
HSEARCH-4483 Update mass-indexing log message format

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4483

- include the "between-iterations" speed as part of logged info

### DIFF
--- a/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/logging/impl/Log.java
+++ b/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/logging/impl/Log.java
@@ -88,8 +88,8 @@ public interface Log extends BasicLogger {
 	void indexingProgressRaw(long doneCount, long elapsedMs);
 
 	@LogMessage(level = INFO)
-	@Message(id = ID_OFFSET_LEGACY_ENGINE + 31, value = "Mass indexing progress: %2$.2f%% [%1$f documents/second].")
-	void indexingProgressStats(float estimateSpeed, float estimatePercentileComplete);
+	@Message(id = ID_OFFSET_LEGACY_ENGINE + 31, value = "Mass indexing progress: %3$.2f%%. Mass indexing speed: %1$f documents/second since last message, %2$f documents/second since start.")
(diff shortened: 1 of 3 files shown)


mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/logging/impl/Log.java
```diff
@@ -88,8 +88,8 @@ public interface Log extends BasicLogger {
 	void indexingProgressRaw(long doneCount, long elapsedMs);
 
 	@LogMessage(level = INFO)
-	@Message(id = ID_OFFSET_LEGACY_ENGINE + 31, value = "Mass indexing progress: %2$.2f%% [%1$f documents/second].")
-	void indexingProgressStats(float estimateSpeed, float estimatePercentileComplete);
+	@Message(id = ID_OFFSET_LEGACY_ENGINE + 31, value = "Mass indexing progress: %3$.2f%% [%1$f documents/second [current] / %2$f documents/second [since start]].")
```
yrodiere, line 91: Ah, I didn't realize we already had square brackets in this message... it looks a bit weird now, doesn't it?

Maybe something like this?

```suggestion
	@Message(id = ID_OFFSET_LEGACY_ENGINE + 31, value = "Mass indexing progress: %3$.2f%% [%1$f documents/second since last message, %2$f documents/second since start].")
```

yrodiere (changes requested):
Thanks for tackling this! This uncovers a few interesting problems... Seems this is more complex than it looked at first sight :)

My suggestions below.

yrodiere:
Oh, also, I know we don't have many tests, but would you mind updating `org.hibernate.search.integrationtest.showcase.library.LibraryShowcaseMassIndexingIT#testMassIndexingMonitor` to check that we actually dispay the speed since start *and* since last message?

marko-bekhta:
Thanks for spending the time looking at this. `AtomicReference` !!!! That's what I was looking for! :smiley: Test updated as well.

yrodiere (approved):
LGTM, merging. Thanks!

marko-bekhta:
Thanks :smiley:

PR #3054, opened 2022-05-16
HSEARCH-4571 Move out unnecessary tests from showcase

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4571

Not much to add here 😃


integrationtest/showcase/library/src/test/java/org/hibernate/search/integrationtest/showcase/library/LibraryShowcaseMassIndexingIT.java
```diff
@@ -38,6 +37,7 @@ public class LibraryShowcaseMassIndexingIT {
 
 	private static final int NUMBER_OF_BOOKS = 200;
 	private static final int MASS_INDEXING_MONITOR_LOG_PERIOD = 50; // This is the default in the implementation, do not change this value
```
yrodiere, line 39: I think you can remove this and the related checks below.

integrationtest/mapper/orm-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/orm/realbackend/sync/OutOfSyncIndexIT.java
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	private int entityCount() {
+		AtomicLong count = new AtomicLong();
+		withinJPATransaction( entityManagerFactory, entityManager -> count.set(
+				entityManager.createQuery( "select b from book b", Book.class )
+						.getResultList().size()
+		) );
+		return count.intValue();
+	}
```
yrodiere, line 97: Maybe use `OrmUtils.count` instead?

yrodiere (changes requested):
Thanks! A few comments below.

Only the comments about `spring.jpa.properties.` are really important; the rest is more about cleaning up and reusing existing utils.

marko-bekhta:
ohh I haven't even noticed the prefix 🙈  applied the changes and squashed them in the previous commit.

PR #3055, opened 2022-05-16
HSEARCH-4575 Make fields projectable by default for ES

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4575

- change the default behavior of projectable for ES
- extend Tck helper to expose the projectable defaults for backends
- update documentation

Haven't seen anything "updatable" on `SearchProjectionFactory` javadoc.

### DIFF
--- a/documentation/src/main/asciidoc/reference/mapper-orm-mapping-directfieldmapping.asciidoc
+++ b/documentation/src/main/asciidoc/reference/mapper-orm-mapping-directfieldmapping.asciidoc
@@ -159,6 +159,9 @@ Whether the field can be <<search-dsl-projection, projected on>>,
 i.e. whether the field value is stored in the index to allow retrieval later when querying.
 +
 Value: `Projectable.YES`, `Projectable.NO`, `Projectable.DEFAULT`.
++
+The defaults are different for the <<backend-lucene,Lucene>> and <<backend-elasticsearch,Elasticsearch>> backends:
+with Lucene, the default is `Projectable.NO`, while with Elasticsearch it's `Projectable.YES`.
 
 [[mapper-orm-directfieldmapping-aggregable]] `aggregable`::
 Whether the field can be <<search-dsl-aggregation, aggregated>>,
--- a/documentation/src/main/asciidoc/reference/mapper-orm-mapping-directfieldmapping.asciidoc
+++ b/documentation/src/main/asciidoc/reference/mapper-orm-mapping-directfieldmapping.asciidoc
@@ -162,6 +162,12 @@ Value: `Projectable.YES`, `Projectable.NO`, `Projectable.DEFAULT`.
 +
 The defaults are different for the <<backend-lucene,Lucene>> and <<backend-elasticsearch,Elasticsearch>> backends:
 with Lucene, the default is `Projectable.NO`, while with Elasticsearch it's `Projectable.YES`.
++
+[NOTE]
+====
+For <<backend-elasticsearch,Elasticsearch>> if any of `projectable` or `sortable` properties are resolved to `YES`
+on a `GeoPoint` field then this field automatically becomes both `projectable` and `sortable` even if one of them was explicitly set to `NO`.
+====
 
 [[mapper-orm-directfieldmapping-aggregable]] `aggregable`::
 Whether the field can be <<search-dsl-aggregation, aggregated>>,
--- a/documentation/src/test/java/org/hibernate/search/documentation/mapper/orm/indexedentities/SearchMappingIndexedEntitiesIT.java
+++ b/documentation/src/test/java/org/hibernate/search/documentation/mapper/orm/indexedentities/SearchMappingIndexedEntitiesIT.java
@@ -24,6 +24,7 @@ import org.hibernate.search.engine.backend.metamodel.IndexValueFieldTypeDescript
 import org.hibernate.search.mapper.orm.Search;
 import org.hibernate.search.mapper.orm.mapping.SearchMapping;
 import org.hibernate.search.mapper.orm.entity.SearchIndexedEntity;
+import org.hibernate.search.util.impl.integrationtest.common.rule.BackendConfiguration;
 
 import org.junit.Before;
 import org.junit.Rule;
@@ -97,7 +98,7 @@ public class SearchMappingIndexedEntitiesIT {
 				Optional<String> normalizerName = type.normalizerName();
 				// Etc.
 				//end::indexMetamodel[]
-				assertThat( projectable ).isFalse();
+				assertThat( projectable ).isEqualTo( BackendConfiguration.isElasticsearch() ? true : false );
 				assertThat( dslArgumentClass ).isEqualTo( Date.class );
 				assertThat( projectedValueClass ).isEqualTo( Date.class );
 				assertThat( analyzerName ).isEmpty();
--- a/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/projection/DistanceProjectionTypeCheckingAndConversionIT.java
+++ b/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/projection/DistanceProjectionTypeCheckingAndConversionIT.java
@@ -9,6 +9,7 @@ package org.hibernate.search.integrationtest.backend.tck.search.projection;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.hibernate.search.integrationtest.backend.tck.testsupport.types.values.IndexableGeoPointWithDistanceFromCenterValues.CENTER_POINT_1;
 import static org.hibernate.search.util.impl.integrationtest.common.assertion.SearchResultAssert.assertThatQuery;
+import static org.junit.Assume.assumeFalse;
 
 import org.hibernate.search.engine.backend.document.DocumentElement;
 import org.hibernate.search.engine.backend.document.IndexObjectFieldReference;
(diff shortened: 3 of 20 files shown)


yrodiere (changes requested):
Thanks!

That's a nice start, but there are a few missing things:

* Tests are failing?
* You're not testing that the default works correctly when actually doing projections, the only changed test is about internal Hibernate Search metadata!
  * There are existing integration tests for distance/field projections, which explicitly set projectable to `YES`. Those tests probably shouldn't do that anymore with Elasticsearch. See `org.hibernate.search.integrationtest.backend.tck.search.projection.FieldProjectionSingleValuedBaseIT`, `org.hibernate.search.integrationtest.backend.tck.search.projection.FieldProjectionMultiValuedBaseIT`, `org.hibernate.search.integrationtest.backend.tck.search.projection.DistanceProjectionSingleValuedBaseIT`, `org.hibernate.search.integrationtest.backend.tck.search.projection.DistanceProjectionMultiValuedBaseIT` : look for `Projectable.YES` in those files: this will become `projectableByDefault : DEFAULT : YES`.
  * There are existing integration tests to check that projections don't work when they are explicitly disabled: see `org.hibernate.search.integrationtest.backend.tck.search.projection.DistanceProjectionTypeCheckingAndConversionIT#nonProjectable`, `org.hibernate.search.integrationtest.backend.tck.search.projection.FieldProjectionTypeCheckingAndConversionIT#nonProjectable`. Now that the situation is becoming a bit more blurry, we should probably also test that, if the backend disallows projections by default, then trying to project on a field with `Projectable.DEFAULT` fails? I.e. in those test classes, duplicate the field and test method to test `Projectable.DEFAULT` as well (with an `assumeFalse(...)` to skip the test on Elasticsearch).

----

Also, there's one thing I didn't foresee when I created the JIRA ticket: I think this change has a side-effect for GeoPoint fields, they now have doc-values enabled by default. See https://github.com/hibernate/hibernate-search/blob/e1669fd7b3beb04ae4523250337bdb7818e3b627/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/ElasticsearchGeoPointIndexFieldTypeOptionsStep.java#L37 .

That may be a nice improvements for users (I don't know), but that's also a backwards-incompatible change: this will make Elasticsearch mapping validation fail, because Hibernate Search will start requiring that a geopoint field has docvalues enabled in the Elasticsearch mapping, whereas Hibernate Search was completely fine with docvalues being disabled before this change.

According to the [compatibility policy](https://hibernate.org/community/compatibility-policy/#remote-mapping-and-data) that's a change we're allowed to make. I'm on the fence, though... any opinion?

**If** we do introduce this breaking change:

* We must document it in the migration guide (`documentation/src/main/asciidoc/migration/index.asciidoc`, section "Data format and schema changes" -- **please rebase on the main branch first**
* I think we need to change `ElasticsearchGeoPointIndexFieldTypeOptionsStep#complete` so that as soon as doc values are enabled for whatever reason (projectable or sortable), the field is considered projectable *and* sortable. The previous behavior was kind of acceptable if projectable/sortable was opt-in, but if they're the default, it seems weird to consider the field non-sortable by default even though doc values are enabled by default and thus Elasticsearch is perfectly able to sort on that field...

marko-bekhta:
> Tests are failing?

it was an ES variation of a test from the documentation that sneaked in `SearchMappingIndexedEntitiesIT#indexMetamodel`. I think `BackendConfiguration.isElasticsearch() ? true : false` should be enough as we don't have access to `TckBackendFeatures`

>If we do introduce this breaking change:

In the end, I've ended up doing the change for `GeoPoint`. At first, I've tried to "preserve" the logic and make HSEARCH not create doc-values, but that looked a bit too cumbersome.

I've modified some of the tests to use the "updated" `sortable`.

Didn't go with `projectableByDefault : DEFAULT : YES` but with passing an empty consumer as a customizer instead. 

A couple of tests that were expecting exceptions now have `assumeFalse` where applicable. 

I see that you've already merged the migration doc PR so I'll rebase

documentation/src/main/asciidoc/reference/mapper-orm-mapping-directfieldmapping.asciidoc
```diff
@@ -159,6 +159,15 @@ Whether the field can be <<search-dsl-projection, projected on>>,
...
 Value: `Projectable.YES`, `Projectable.NO`, `Projectable.DEFAULT`.
++
+The defaults are different for the <<backend-lucene,Lucene>> and <<backend-elasticsearch,Elasticsearch>> backends:
+with Lucene, the default is `Projectable.NO`, while with Elasticsearch it's `Projectable.YES`.
+
+[IMPORTANT]
+====
+For <<backend-elasticsearch,Elasticsearch>> if any of `projectable` or `sortable` properties are resolved to `YES`
+on a `GeoPoint` field then this field automatically becomes both `projectable` and `sortable` even if one of them was explicitly set to `NO`.
+====
```
yrodiere, line 171: Good idea to mention this, but I wouldn't consider this "important": after all, if a user expects a field not to be sortable but it ends up being sortable, they won't see the difference.

```suggestion
+
[NOTE]
====
For <<backend-elasticsearch,Elasticsearch>> if any of `projectable` or `sortable` properties are resolved to `YES`
on a `GeoPoint` field then this field automatically becomes both `projectable` and `sortable` even if one of them was explicitly set to `NO`.
====
```

documentation/src/test/java/org/hibernate/search/documentation/mapper/orm/indexedentities/SearchMappingIndexedEntitiesIT.java
```diff
@@ -97,7 +98,7 @@ public void indexMetamodel() {
 				Optional<String> normalizerName = type.normalizerName();
 				// Etc.
 				//end::indexMetamodel[]
-				assertThat( projectable ).isFalse();
+				assertThat( projectable ).isEqualTo( BackendConfiguration.isElasticsearch() ? true : false );
```
yrodiere, line 101: :+1:

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/projection/DistanceProjectionTypeCheckingAndConversionIT.java
```diff
@@ -74,7 +76,8 @@ public class DistanceProjectionTypeCheckingAndConversionIT {
 	@BeforeClass
 	public static void setup() {
 		setupHelper.start()
-				.withIndexes( mainIndex, compatibleIndex, rawFieldCompatibleIndex, missingFieldIndex, incompatibleIndex )
+				.withIndexes(
+						mainIndex, compatibleIndex, rawFieldCompatibleIndex, missingFieldIndex, incompatibleIndex )
```
yrodiere, line 80: I see you're using our IDE code style templates; that's good, but in the future please make sure to only apply auto-formatting to portions of files you changed: that way you won't change irrelevant code in your commits.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/spatial/DistanceSearchableSortableIT.java
```diff
@@ -62,6 +64,10 @@ public void searchableSortable() {
 
 	@Test
 	public void searchableNotSortable() {
+		assumeFalse(
+				"Skipping test for ES GeoPoint as those would become sortable by default in this case.",
+				TckConfiguration.get().getBackendFeatures().fieldsProjectableByDefault()
+		);
```
yrodiere, line 70: For this specific test, I'd rather explicitly mark the field as non-projectable, so that we still can check that the Elasticsearch backend behaves correctly for non-sortable fields. Otherwise we wouldn't be testing this behavior at all...

yrodiere (commented):
Almost there :) Tests are passing, and I only have a few comments about documentation. I'll merge right after those are addressed!

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/projection/DistanceProjectionTypeCheckingAndConversionIT.java
marko-bekhta: yes, sure! I remember it took a while to make Idea code style to match the Eclipse one back in the days while working with HV. I've cleaned this formatting up as if it never happened 🙈

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/spatial/DistanceSearchableSortableIT.java
marko-bekhta: I've added a similar test but for a "new" `searchableNotSortableNotProjectable` field, this way it follows that pattern in naming - whatever we set we put in name 😃

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/projection/DistanceProjectionTypeCheckingAndConversionIT.java
yrodiere: Thanks.

FWIW we now publish IDEA code styles as well: https://hibernate.org/community/contribute/intellij-idea/#code-style

PR #3060, opened 2022-05-17
Backport HSEARCH-4515 + HSEARCH-4584 to branch 6.1 - Projection and sort on the same nested field fails with the Lucene Backend

yrodiere:
* [HSEARCH-4584](https://hibernate.atlassian.net/browse/HSEARCH-4584): Projection and sort on the same nested field fails with the Lucene Backend
* [HSEARCH-4515](https://hibernate.atlassian.net/browse/HSEARCH-4515): Simplify NestedDocsProvider and a few projections

Backport of #2955 and  #3059 to branch 6.1.


yrodiere:
I just force-pushed to remove a few commits from [HSEARCH-4515](https://hibernate.atlassian.net/browse/HSEARCH-4515) that do not help fix HSEARCH-4584. The changes should be safer that way.

PR #3064, opened 2022-05-18
HSEARCH-4571 Follow-up to fix tests

yrodiere:
Follows up on #3054 .

@marko-bekhta You may be interested in the first commit. I missed that during my review, otherwise I would have told you Hibernate Search has a built-in feature to make indexing synchronous :) We generally only use that in tests, because performance is bad, but it's *very* useful for tests.


marko-bekhta:
It's funny 😃 yesterday I thought that I should probably point out the usage of await as it was causing the problem with ES tests, but then I thought that you probably saw it and were OK with it 🙈

PR #3068, opened 2022-05-23
HSEARCH-4586 Use clean, consistent utils for all in-session code execution in ORM sessions

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4586

I was also wondering if those `withinTransaction()` methods that are delegating to `with().run...()` methods should go away as well ?

### DIFF
--- a/util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/ReusableOrmSetupHolder.java
+++ b/util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/ReusableOrmSetupHolder.java
@@ -21,7 +21,9 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.BiFunction;
 import java.util.function.Consumer;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import javax.persistence.EntityManagerFactory;
 import javax.persistence.criteria.CriteriaBuilder;
@@ -98,7 +100,7 @@ import org.jboss.logging.Logger;
  * 	}
  * }</pre>
  */
-public class ReusableOrmSetupHolder implements TestRule {
+public class ReusableOrmSetupHolder implements TestRule, PersistenceRunner<Session, Transaction> {
 	private static final Logger log = Logger.getLogger( ReusableOrmSetupHolder.class.getName() );
 
 	/**
@@ -226,12 +228,22 @@ public class ReusableOrmSetupHolder implements TestRule {
 		return sessionFactory;
 	}
 
-	public PersistenceRunner<Session, Transaction> with() {
+	private PersistenceRunner<Session, Transaction> with() {
 		return OrmUtils.with( sessionFactory() );
 	}
 
 	public PersistenceRunner<Session, Transaction> with(String tenantId) {
 		return OrmUtils.with( sessionFactory(), tenantId );
+	}
+
+	@Override
+	public <R> R applyNoTransaction(Function<? super Session, R> action) {
+		return with().applyNoTransaction( action );
+	}
+
+	@Override
+	public <R> R applyInTransaction(BiFunction<? super Session, ? super Transaction, R> action) {
+		return with().applyInTransaction( action );
 	}
 
 	public void runInTransaction(Consumer<? super Session> action) {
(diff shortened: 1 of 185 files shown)


util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/ReusableOrmSetupHolder.java
```diff
@@ -320,7 +321,7 @@ private void setupSessionFactory(Object testInstance) {
 				try {
 					clearAllData( sessionFactory );
 				}
-				catch (RuntimeException e) {
```
yrodiere, line 311: I don't think this change is necessary? Why did you need to make `clearAllData` throw a `Throwable`?

util/internal/test/common/src/main/java/org/hibernate/search/util/impl/test/ThrowingFunction.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.util.impl.test;
```
yrodiere, line 7: I'm sorry, I made a mistake in my comment on JIRA; I meant these function classes should live in package `org.hibernate.search.util.impl.test.function`. Could you please move them there in an additional commit? Using your IDE's refactoring tools, it should be easy :)

yrodiere (commented):
Thanks! A few minor comments below.

> I was also wondering if those withinTransaction() methods that are delegating to with().run...() methods should go away as well ?

Yes! Sorry, I forgot to mention those. Ideally the methods that are actually useful (i.e. not just overloads) should be moved to private methods in `JPAPersistenceRunner`/`NativePersistenceRunner`, and calls to these methods should be replaced with calls to `OrmUtils.with`.

util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/ReusableOrmSetupHolder.java
```diff
@@ -619,7 +620,7 @@ private static class DataClearConfigImpl implements DataClearConfig {
 
 		private final List<Class<?>> entityClearOrder = new ArrayList<>();
 
-		private final List<Consumer<Session>> preClear = new ArrayList<>();
+		private final List<ThrowingConsumer<Session, ?>> preClear = new ArrayList<>();
```
yrodiere, line 633: Ah, I think this is the problem, and the reason you had to turn a `catch (RuntimeException` into a `catch (Throwable` above.

If you apply this suggestion, you'll be able to revert the other changes that add `Throwable` everywhere in this file.

```suggestion
		private final List<ThrowingConsumer<Session, RuntimeException>> preClear = new ArrayList<>();
```

marko-bekhta: yup 😄  that's the culprit 😄. sorry for not replying sooner. I think I've originally done it like you are suggesting here, but then the changes started to span even further, but it might've been in a different place, so I'll double-check once I get a chance.

util/internal/test/common/src/main/java/org/hibernate/search/util/impl/test/ThrowingFunction.java
marko-bekhta: 😏 ok

yrodiere (approved):
Thanks!

You were right to separate the work on `withinTransaction(Session, ...)` from the rest, in the last commit: I didn't think of those (few) use cases, and I think we should keep them as is. I removed that last commit and replaced it with a simple renaming of those methods, to make them more consistent with `PersistenceRunner`. I guess they'll have to stay in `OrmUtils` for now.

If you agree, let's merge as soon as CI goes green :)

marko-bekhta:
fine by me 😃  That call `with(factory).run(session, ...)` did feel a bit awkward, even though I've done it, haha. I did want to have `EntityManager` and `Session` specific logic to live in their own classes. Another thought I just had ... we could add another interface alongside the `PersistenceRunner` that would work with `Session/EntityManager` instead of their factories. And then do something like:
```
with(session).runInTransaction( tx -> {});
...
with(em).runInTransaction( tx -> {});
```
this could make it more "in the same style", but not sure if that's worth it 😄

yrodiere:
> this could make it more "in the same style", but not sure if that's worth it

Yes; I thought about that, but then I noticed there were only two tests that actually needed that :) As you said, probably not worth it.

PR #3140, opened 2022-07-11
HSEARCH-4601 - Syntactic sugar for OR/AND predicates

YannisBres:
https://hibernate.atlassian.net/browse/HSEARCH-4601

### DIFF
--- a/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/BoolPredicateSpecificsIT.java
+++ b/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/BoolPredicateSpecificsIT.java
@@ -109,6 +109,15 @@ public class BoolPredicateSpecificsIT {
 	}
 
 	@Test
+	public void and_function() {
+		assertThatQuery( index.query()
+				.where( f -> f.bool()
+						.and( f2 -> f2.match().field( "field1" ).matching( FIELD1_VALUE1 ) )
+				) )
+				.hasDocRefHitsAnyOrder( index.typeName(), DOCUMENT_1 );
+	}
+
+	@Test
 	public void must_separatePredicateObject() {
 		StubMappingScope scope = index.createScope();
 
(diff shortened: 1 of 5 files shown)


integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/BoolPredicateSpecificsIT.java
```diff
@@ -313,6 +374,30 @@ public void must_should() {
...
+	public void and_or() {
+		// A boolean predicate with and + or clauses:
+		// documents should match regardless of whether or clauses match.
+
+		// Non-matching "and" clauses
+		assertThatQuery( index.query()
+				.where( f -> f.bool()
+						.and( f.match().field( "field1" ).matching( FIELD1_VALUE1 ) )
+						.or( f.match().field( "field2" ).matching( FIELD2_VALUE2 ) )
+						.or( f.match().field( "field3" ).matching( FIELD3_VALUE3 ) )
```
yrodiere, line 387: Hi, and thanks for the contribution.

This doesn't seem to match what was described in the ticket. Can you have another look? We should probably discuss this before (if) you give it another try.

If it's not a misunderstanding and you are simply proposing an alternative solution, I I'm not that enthusiastic about it...   The goal was just to provide a simpler way of building boolean predicates; I don't think this solution is simple.

Just the fact that people can use `and` and `or` in the same predicate will lead to seriously complex behavior... Looking at the example above, wouldn't you say that the syntax `and E1 or E2 or E3` is exotic at best, and straight out confusing at worst?

Also, we end up mixing the old (powerful, but complex) must/should mechanism with the new (limited, but simpler) and/or mechanism, which leads to allowing things like this:

```java

				.where( f -> f.bool()
						.and( f.match().field( "field1" ).matching( FIELD1_VALUE1 ) )
						.should( f.match().field( "field2" ).matching( FIELD2_VALUE2 ) )
						.or( f.match().field( "field3" ).matching( FIELD3_VALUE3 ) )
```

I'm not sure that's a good idea. Even just because of IDE code completion, which will suggest newcomers many confusing methods after `bool`: `and`/`or`/`must`/`should`/`mustNot`/`filter`.

yrodiere (changes requested):

YannisBres:
Arf...  I should have re-read the PR between the time I thought I understood it and the time I started to work on it...
Can I blame the cognitive overload of having to start on a totally brand new code base for such a miserable failure?  ;-)
I should also have followed my instinct, as I was a bit annoyed by these "different ways to do the same thing", which I always find puzzling in APIs...  (Eclipse Collections, I'm looking at you now!)
I'll make another pull request!

PR #3152, opened 2022-07-22
HSEARCH-4601 Syntactic sugar for OR/AND predicates

YannisBres:
https://hibernate.atlassian.net/browse/HSEARCH-4601

Hi!

So, I went back to this PR and I hope I got it right this time after my initial botching.

I have added support for basic n-ary or(...) / and(...) constructs _à la_
```
f.or( f.match().field("foo").matching("val1"),
      f.match().field("bar").matching("val2") )
f.and( f.match().field("foo").matching("val1"),
       f.match().field("bar").matching("val2") )
```

I haven't yet added builder-style constructs with `.add(...)` and `with(...)`.
I guess that should be done.

I haven't updated the documentation yet either.

Before I take the next step, is this going in the right direction?  ;-)

I have added new tests in `BoolPredicateSpecificsIT`, as they are more or less cloned from `bool()` ones.
This makes the file name not very relevant...
Shall I add a new IT file of its own, at the risk of more duplicated code?
If not, shall "duplicated" methods be grouped further, e.g. in `@Nested` test classes?
Also, aren't the tests in `SmokeIT` a bit redundant/pointless?  Maybe they should not be added there...

So far, `AndPredicateClausesStep` and `OrPredicateClausesStep` don't implement `PredicateScoreStep`, which would allow calling `boost(...)` and `constantScore()` on the overall filter.
Shouldn't they?

Similarly, so far, `OrPredicateClausesStep` doesn't support any of the `minimumShouldMatch` constraints.  Shall it?
That looks to me a bit far away from the expected semantic of an `OR` (with `minimumShouldMatch == 1`), so maybe such a feature should be reserved to more complex `bool()`-based predicates.

(If nothing should be added, the Javadoc "@return The initial step of a DSL" should be revised.)

Shall something be done to support `mustNot`?
Again, I'd leave that to `bool()`...

Shall `SearchPredicateFactory.and(...)` and `or(...)` be tagged as `@Incubating` at first?

As usual, using varargs with a parameterized type requires a lot of `@SuppresssWarnings("unchecked")`, including in calling code...  :'-(
Would you happen to have anything better to suggest?
(Besides n overloads for all possible arities. ;-))

Best regards,
          Yannis

### DIFF
--- a/documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
+++ b/documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
@@ -784,11 +784,143 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
 either on a per-field basis with a call to `.boost(...)` just after `.field(...)`/`.fields(...)`
 or for the whole predicate with a call to `.boost(...)` after `.matchingAny(...)` or `.matchingAll(...)`.
 
+[[search-dsl-predicate-and]]
+== `and`: match all clauses
+
+The `and` predicate matches documents that match **all** of its inner predicates, called "clauses".
+
+Matching "and" clauses are taken into account during <<search-dsl-predicate-common-score,score>> computation.
+
+.Matching a document that matches all the multiple given predicates (~`AND` operator)
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java[tags=and]
+----
+<1> The hits *must* have a `title` field matching the text `robot`,
+independently of other clauses in the same boolean predicate.
+<2> The hits *must* have a `description` field matching the text `crime`,
+independently of other clauses in the same boolean predicate.
+<3> All returned hits will match *all* the clauses above:
+they will have a `title` field matching the text `robot`
+*and* they will have a `description` field matching the text `crime`.
+====
+
+[[search-dsl-predicate-and-lambda]]
+=== Adding clauses dynamically with the lambda syntax
+
+It is possible to define the `and` predicate inside a lambda expression.
+This is especially useful when clauses need to be added dynamically to the `and` predicate,
+for example based on user input.
+
+.Easily adding clauses dynamically using `.where(...)` and the lambda syntax
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java[tags=and-dynamicParameters-root]
+----
+<1> Get a custom object holding the search parameters provided by the user through a web form, for example.
+<2> Call `.where(Function)`.
+The function, implemented by a lambda expression, will receive a predicate factory, use it to build an `and` predicate,
+invoke the `with(Consumer)` method and return this predicate. The consumer, implemented by a lambda expression,
+will receive a clause collector for the `and` predicate as an argument, and will add clauses to that collector as necessary.
+<3> By default, a boolean predicate will match nothing if there is no clause.
+To match every document when there is no clause, add an `and` clause that matches everything.
+<4> Inside the lambda, the code is free to use any Java language constructs, such as `if` or `for`,
+to control the addition of clauses.
+In this case, we only add clauses if the relevant parameter was filled in by the user.
+====
+
+Another syntax relying on the method `with(...)` can be useful when the boolean predicate is not the root predicate,
+see <<example-and-or-dynamicParameters-with,this example>>.
+
+[[search-dsl-predicate-and-other]]
+=== Options
+
+* The score of an `and` predicate is variable by default,
+but can be <<search-dsl-predicate-common-constantScore,made constant with `.constantScore()`>>.
+* The score of an `and` predicate can be <<search-dsl-predicate-common-boost,boosted>>
+with a call to `.boost(...)`.
+
+[[search-dsl-predicate-or]]
+== `or`: match any clause
+
+The `or` predicate matches documents that match **any** of its inner predicates, called "clauses".
+
+Matching `or` clauses are taken into account during <<search-dsl-predicate-common-score,score>> computation.
+
+.Matching a document that matches any of multiple given predicates (~`OR` operator)
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java[tags=or]
+----
+<1> The hits *should* have a `title` field matching the text `robot`,
+*or* they should match any other clause in the same boolean predicate.
+<2> The hits *should* have a `description` field matching the text `investigation`,
+*or* they should match any other clause in the same boolean predicate.
+<3> All returned hits will match *at least one* of the clauses above:
+they will have a `title` field matching the text `robot`
+*or* they will have a `description` field matching the text `investigation`.
+====
+
+[[search-dsl-predicate-or-lambda]]
+=== Adding clauses dynamically with the lambda syntax
+
+It is possible to define the `or` predicate inside a lambda expression.
+This is especially useful when clauses need to be added dynamically to the `or` predicate,
+for example based on user input.
+
+.Easily adding clauses dynamically using `.where(...)` and the lambda syntax
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java[tags=or-dynamicParameters-root]
+----
+<1> Get a custom object holding the search parameters provided by the user through a web form, for example.0
+<2> Call `.where(Function)`.
+The function, implemented by a lambda expression, will receive a predicate factory, use it to build an `or` predicate,
+invoke the `with(Consumer)` method and return this predicate. The consumer, implemented by a lambda expression,
+will receive a clause collector for the `or` predicate as an argument, and will add clauses to that collector as necessary.
+<3> Inside the lambda, the code is free to use any Java language constructs, such as `if` or `for`,
+to control the addition of clauses.
+In this case, we only add clauses if the relevant parameter was filled in by the user.
+====
+
+Another syntax relying on the method `with(...)` can be useful when the boolean predicate is not the root predicate:
+
+.[[example-and-or-dynamicParameters-with]]Easily adding clauses dynamically using `with(...)` and the lambda syntax
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java[tags=and-or-dynamicParameters-with]
+----
+<1> Get a custom object holding the search parameters provided by the user through a web form, for example.
+<2> Call `.where(Function)` to create the root predicate, which is not the one we're interested in here.
+<3> Call `.or().with(Consumer)` to create an inner, non-root predicate.
+The consumer, implemented by a lambda expression, will receive a collector as an argument
+and will add clauses to that collector as necessary.
+<4> Inside the lambda, the code is free to use any Java language constructs, such as `if` or `for`,
+to control the addition of clauses.
+In this case, we add one clause per author filter.
+====
+
+[[search-dsl-predicate-or-other]]
+=== Options
+
+* The score of an `or` predicate is variable by default,
+but can be <<search-dsl-predicate-common-constantScore,made constant with `.constantScore()`>>.
+* The score of an `or` predicate can be <<search-dsl-predicate-common-boost,boosted>>
+with a call to `.boost(...)`.
+
 [[search-dsl-predicate-boolean]]
-== [[_combining_queries]] `bool`: combine predicates (or/and/...)
+== `bool`: advanced combinations of predicates (or/and/...)
+
+The `bool` predicate allows combining inner predicates in a more complex fashion than with simpler
+<<search-dsl-predicate-and,`and`>>/<<search-dsl-predicate-or,`or`>> predicates.
 
 The `bool` predicate matches documents that match one or more inner predicates, called "clauses".
-It can be used in particular to build `AND`/`OR` operators.
+It can be used in particular to build `AND`/`OR` operators with additional settings.
 
 Inner predicates are added as clauses of one of the following types:
 
@@ -825,42 +957,14 @@ and are simply used for scoring.
 [[search-dsl-predicate-boolean-or]]
 === Emulating an `OR` operator
 
-A `bool` predicate with only `should` clauses will behave as an `OR` operator.
-
-.Matching a document that matches any of multiple given predicates (~`OR` operator)
-====
-[source, JAVA, indent=0, subs="+callouts"]
-----
-include::{sourcedir}/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java[tags=bool-or]
-----
-<1> The hits *should* have a `title` field matching the text `robot`,
-*or* they should match any other clause in the same boolean predicate.
-<2> The hits *should* have a `description` field matching the text `investigation`,
-*or* they should match any other clause in the same boolean predicate.
-<3> All returned hits will match *at least one* of the clauses above:
-they will have a `title` field matching the text `robot`
-*or* they will have a `description` field matching the text `investigation`.
-====
+A `bool` predicate with only `should` clauses and no `minimumShouldMatch` specification will behave as an `OR` operator.
+In such case, using the simpler <<search-dsl-predicate-or,`or`>> syntax is recommended.
 
 [[search-dsl-predicate-boolean-and]]
 === Emulating an `AND` operator
 
 A `bool` predicate with only `must` clauses will behave as an `AND` operator.
-
-.Matching a document that matches all the multiple given predicates (~`AND` operator)
-====
-[source, JAVA, indent=0, subs="+callouts"]
-----
-include::{sourcedir}/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java[tags=bool-and]
-----
-<1> The hits *must* have a `title` field matching the text `robot`,
-independently of other clauses in the same boolean predicate.
-<2> The hits *must* have a `description` field matching the text `crime`,
-independently of other clauses in the same boolean predicate.
-<3> All returned hits will match *all* the clauses above:
-they will have a `title` field matching the text `robot`
-*and* they will have a `description` field matching the text `crime`.
-====
+In such case, using the simpler <<search-dsl-predicate-and,`and`>> syntax is recommended.
 
 [[search-dsl-predicate-boolean-mustNot]]
 === `mustNot`: excluding documents that match a given predicate
--- a/documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
+++ b/documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
@@ -813,7 +813,7 @@ It is possible to define the `and` predicate inside a lambda expression.
 This is especially useful when clauses need to be added dynamically to the `and` predicate,
 for example based on user input.
 
-.Easily adding clauses dynamically using `.where(...)` and the lambda syntax
+.Easily adding clauses dynamically using `with(...)` and the lambda syntax
 ====
 [source, JAVA, indent=0, subs="+callouts"]
 ----
@@ -830,9 +830,6 @@ To match every document when there is no clause, add an `and` clause that matche
 to control the addition of clauses.
 In this case, we only add clauses if the relevant parameter was filled in by the user.
 ====
-
-Another syntax relying on the method `with(...)` can be useful when the boolean predicate is not the root predicate,
-see <<example-and-or-dynamicParameters-with,this example>>.
 
 [[search-dsl-predicate-and-other]]
 === Options
@@ -871,7 +868,7 @@ It is possible to define the `or` predicate inside a lambda expression.
 This is especially useful when clauses need to be added dynamically to the `or` predicate,
 for example based on user input.
 
-.Easily adding clauses dynamically using `.where(...)` and the lambda syntax
+.Easily adding clauses dynamically using `with(...)` and the lambda syntax
 ====
 [source, JAVA, indent=0, subs="+callouts"]
 ----
@@ -887,23 +884,6 @@ to control the addition of clauses.
 In this case, we only add clauses if the relevant parameter was filled in by the user.
 ====
 
-Another syntax relying on the method `with(...)` can be useful when the boolean predicate is not the root predicate:
-
-.[[example-and-or-dynamicParameters-with]]Easily adding clauses dynamically using `with(...)` and the lambda syntax
-====
-[source, JAVA, indent=0, subs="+callouts"]
-----
-include::{sourcedir}/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java[tags=and-or-dynamicParameters-with]
-----
-<1> Get a custom object holding the search parameters provided by the user through a web form, for example.
-<2> Call `.where(Function)` to create the root predicate, which is not the one we're interested in here.
-<3> Call `.or().with(Consumer)` to create an inner, non-root predicate.
-The consumer, implemented by a lambda expression, will receive a collector as an argument
-and will add clauses to that collector as necessary.
-<4> Inside the lambda, the code is free to use any Java language constructs, such as `if` or `for`,
-to control the addition of clauses.
-In this case, we add one clause per author filter.
-====
 
 [[search-dsl-predicate-or-other]]
 === Options
@@ -914,7 +894,7 @@ but can be <<search-dsl-predicate-common-constantScore,made constant with `.cons
 with a call to `.boost(...)`.
 
 [[search-dsl-predicate-boolean]]
-== `bool`: advanced combinations of predicates (or/and/...)
+== [[_combining_queries]] `bool`: advanced combinations of predicates (or/and/...)
 
 The `bool` predicate allows combining inner predicates in a more complex fashion than with simpler
 <<search-dsl-predicate-and,`and`>>/<<search-dsl-predicate-or,`or`>> predicates.
@@ -957,7 +937,7 @@ and are simply used for scoring.
 [[search-dsl-predicate-boolean-or]]
 === Emulating an `OR` operator
 
-A `bool` predicate with only `should` clauses and no `minimumShouldMatch` specification will behave as an `OR` operator.
+A `bool` predicate with only `should` clauses and no <<search-dsl-predicate-boolean-minimumshouldmatch,`minimumShouldMatch` specification>> will behave as an `OR` operator.
 In such case, using the simpler <<search-dsl-predicate-or,`or`>> syntax is recommended.
 
 [[search-dsl-predicate-boolean-and]]
--- a/documentation/src/test/java/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java
+++ b/documentation/src/test/java/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java
@@ -168,39 +168,118 @@ public class PredicateDslIT {
 	}
 
 	@Test
-	public void bool() {
+	public void and() {
 		withinSearchSession( searchSession -> {
-			// tag::bool-or[]
+			// tag::and[]
 			List<Book> hits = searchSession.search( Book.class )
-					.where( f -> f.bool()
-							.should( f.match().field( "title" )
-									.matching( "robot" ) ) // <1>
-							.should( f.match().field( "description" )
-									.matching( "investigation" ) ) // <2>
+					.where( f -> f.and(
+									f.match().field( "title" )
+											.matching( "robot" ), // <1>
+									f.match().field( "description" )
+											.matching( "crime" ) // <2>
+							)
 					)
 					.fetchHits( 20 ); // <3>
-			// end::bool-or[]
+			// end::and[]
+			assertThat( hits )
+					.extracting( Book::getId )
+					.containsExactlyInAnyOrder( BOOK3_ID );
+		} );
+
+		withinSearchSession( searchSession -> {
+			// tag::and-dynamicParameters-root[]
+			MySearchParameters searchParameters = getSearchParameters(); // <1>
+			List<Book> hits = searchSession.search( Book.class )
+					.where( f -> f.and().with( b -> { // <2>
+						b.add( f.matchAll() ); // <3>
+						if ( searchParameters.getGenreFilter() != null ) { // <4>
+							b.add( f.match().field( "genre" )
+									.matching( searchParameters.getGenreFilter() ) );
+						}
+						if ( searchParameters.getFullTextFilter() != null ) {
+							b.add( f.match().fields( "title", "description" )
+									.matching( searchParameters.getFullTextFilter() ) );
+						}
+						if ( searchParameters.getPageCountMaxFilter() != null ) {
+							b.add( f.range().field( "pageCount" )
+									.atMost( searchParameters.getPageCountMaxFilter() ) );
+						}
+					} ) )
+					.fetchHits( 20 );
+			// end::and-dynamicParameters-root[]
+			assertThat( hits )
+					.extracting( Book::getId )
+					.containsExactlyInAnyOrder( BOOK1_ID, BOOK2_ID );
+		} );
+
+		withinSearchSession( searchSession -> {
+			// tag::and-or-dynamicParameters-with[]
+			MySearchParameters searchParameters = getSearchParameters(); // <1>
+			List<Book> hits = searchSession.search( Book.class )
+					.where( f -> f.and().with( b -> { // <2>
+						b.add( f.matchAll() );
+						if ( searchParameters.getGenreFilter() != null ) {
+							b.add( f.match().field( "genre" )
+									.matching( searchParameters.getGenreFilter() ) );
+						}
+						if ( !searchParameters.getAuthorFilters().isEmpty() ) {
+							b.add( f.or().with( b2 -> { // <3>
+								for ( String authorFilter : searchParameters.getAuthorFilters() ) { // <4>
+									b2.add( f.match().fields( "authors.firstName", "authors.lastName" )
+											.matching( authorFilter ) );
+								}
+							} ) );
+						}
+					} ) )
+					.fetchHits( 20 );
+			// end::and-or-dynamicParameters-with[]
+			assertThat( hits )
+					.extracting( Book::getId )
+					.containsExactlyInAnyOrder( BOOK1_ID, BOOK2_ID, BOOK3_ID );
+		} );
+	}
+
+	@Test
+	public void or() {
+		withinSearchSession( searchSession -> {
+			// tag::or[]
+			List<Book> hits = searchSession.search( Book.class )
+					.where( f -> f.or(
+									f.match().field( "title" )
+											.matching( "robot" ), // <1>
+									f.match().field( "description" )
+											.matching( "investigation" ) // <2>
+							)
+					)
+					.fetchHits( 20 ); // <3>
+			// end::or[]
 			assertThat( hits )
 					.extracting( Book::getId )
 					.containsExactlyInAnyOrder( BOOK1_ID, BOOK2_ID, BOOK3_ID );
 		} );
 
 		withinSearchSession( searchSession -> {
-			// tag::bool-and[]
+			// tag::or-dynamicParameters-root[]
+			MySearchParameters searchParameters = getSearchParameters(); // <1>
 			List<Book> hits = searchSession.search( Book.class )
-					.where( f -> f.bool()
-							.must( f.match().field( "title" )
-									.matching( "robot" ) ) // <1>
-							.must( f.match().field( "description" )
-									.matching( "crime" ) ) // <2>
-					)
-					.fetchHits( 20 ); // <3>
-			// end::bool-and[]
+					.where( f -> f.or().with( b -> { // <2>
+						if ( !searchParameters.getAuthorFilters().isEmpty() ) {
+							for ( String authorFilter : searchParameters.getAuthorFilters() ) { // <3>
+								b.add( f.match().fields( "authors.firstName", "authors.lastName" )
+										.matching( authorFilter ) );
+							}
+						}
+					} ) )
+					.fetchHits( 20 );
+			// end::or-dynamicParameters-root[]
 			assertThat( hits )
 					.extracting( Book::getId )
-					.containsExactlyInAnyOrder( BOOK3_ID );
+					.containsExactlyInAnyOrder( BOOK1_ID, BOOK2_ID, BOOK3_ID );
 		} );
+	}
 
+	@Test
+	public void bool() {
 		withinSearchSession( searchSession -> {
 			// tag::bool-mustNot[]
 			List<Book> hits = searchSession.search( Book.class )
@@ -513,15 +592,16 @@ public class PredicateDslIT {
 		withinSearchSession( searchSession -> {
 			// tag::score-constantScore[]
 			List<Book> hits = searchSession.search( Book.class )
-					.where( f -> f.bool()
-							.must( f.match()
+					.where( f -> f.and(
+							f.match()
 									.field( "genre" )
 									.matching( Genre.SCIENCE_FICTION )
-									.constantScore() )
-							.must( f.match()
+									.constantScore(),
+							f.match()
 									.field( "title" )
 									.matching( "robot" )
-									.boost( 2.0f ) ) )
+									.boost( 2.0f )
+					) )
 					.fetchHits( 20 );
 			// end::score-constantScore[]
 			assertThat( hits )
@@ -532,14 +612,15 @@ public class PredicateDslIT {
 		withinSearchSession( searchSession -> {
 			// tag::score-boost[]
 			List<Book> hits = searchSession.search( Book.class )
-					.where( f -> f.bool()
-							.must( f.match()
+					.where( f -> f.and(
+							f.match()
 									.field( "title" )
 									.matching( "robot" )
-									.boost( 2.0f ) )
-							.must( f.match()
+									.boost( 2.0f ),
+							f.match()
 									.field( "description" )
-									.matching( "self-aware" ) ) )
+									.matching( "self-aware" )
+					) )
 					.fetchHits( 20 );
 			// end::score-boost[]
 			assertThat( hits )
--- a/documentation/src/test/java/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java
+++ b/documentation/src/test/java/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java
@@ -190,18 +190,18 @@ public class PredicateDslIT {
 			// tag::and-dynamicParameters-root[]
 			MySearchParameters searchParameters = getSearchParameters(); // <1>
 			List<Book> hits = searchSession.search( Book.class )
-					.where( f -> f.and().with( b -> { // <2>
-						b.add( f.matchAll() ); // <3>
+					.where( f -> f.and().with( and -> { // <2>
+						and.add( f.matchAll() ); // <3>
 						if ( searchParameters.getGenreFilter() != null ) { // <4>
-							b.add( f.match().field( "genre" )
+							and.add( f.match().field( "genre" )
 									.matching( searchParameters.getGenreFilter() ) );
 						}
 						if ( searchParameters.getFullTextFilter() != null ) {
-							b.add( f.match().fields( "title", "description" )
+							and.add( f.match().fields( "title", "description" )
 									.matching( searchParameters.getFullTextFilter() ) );
 						}
 						if ( searchParameters.getPageCountMaxFilter() != null ) {
-							b.add( f.range().field( "pageCount" )
+							and.add( f.range().field( "pageCount" )
 									.atMost( searchParameters.getPageCountMaxFilter() ) );
 						}
 					} ) )
@@ -237,10 +237,10 @@ public class PredicateDslIT {
 			// tag::or-dynamicParameters-root[]
 			MySearchParameters searchParameters = getSearchParameters(); // <1>
 			List<Book> hits = searchSession.search( Book.class )
-					.where( f -> f.or().with( b -> { // <2>
+					.where( f -> f.or().with( or -> { // <2>
 						if ( !searchParameters.getAuthorFilters().isEmpty() ) {
 							for ( String authorFilter : searchParameters.getAuthorFilters() ) { // <3>
-								b.add( f.match().fields( "authors.firstName", "authors.lastName" )
+								or.add( f.match().fields( "authors.firstName", "authors.lastName" )
 										.matching( authorFilter ) );
 							}
 						}
--- a/documentation/src/test/java/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java
+++ b/documentation/src/test/java/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java
@@ -212,31 +212,6 @@ public class PredicateDslIT {
 					.containsExactlyInAnyOrder( BOOK1_ID, BOOK2_ID );
 		} );
 
-		withinSearchSession( searchSession -> {
-			// tag::and-or-dynamicParameters-with[]
-			MySearchParameters searchParameters = getSearchParameters(); // <1>
-			List<Book> hits = searchSession.search( Book.class )
-					.where( f -> f.and().with( b -> { // <2>
-						b.add( f.matchAll() );
-						if ( searchParameters.getGenreFilter() != null ) {
-							b.add( f.match().field( "genre" )
-									.matching( searchParameters.getGenreFilter() ) );
-						}
-						if ( !searchParameters.getAuthorFilters().isEmpty() ) {
-							b.add( f.or().with( b2 -> { // <3>
-								for ( String authorFilter : searchParameters.getAuthorFilters() ) { // <4>
-									b2.add( f.match().fields( "authors.firstName", "authors.lastName" )
-											.matching( authorFilter ) );
-								}
-							} ) );
-						}
-					} ) )
-					.fetchHits( 20 );
-			// end::and-or-dynamicParameters-with[]
-			assertThat( hits )
-					.extracting( Book::getId )
-					.containsExactlyInAnyOrder( BOOK1_ID, BOOK2_ID, BOOK3_ID );
-		} );
 	}
 
 	@Test
--- a/engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/SearchPredicateFactory.java
+++ b/engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/SearchPredicateFactory.java
@@ -11,6 +11,7 @@ import java.util.function.Consumer;
 
 import org.hibernate.search.engine.backend.types.ObjectStructure;
 import org.hibernate.search.engine.search.common.BooleanOperator;
+import org.hibernate.search.engine.search.predicate.SearchPredicate;
 import org.hibernate.search.util.common.SearchException;
 import org.hibernate.search.util.common.annotation.Incubating;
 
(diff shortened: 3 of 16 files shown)


engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/SearchPredicateFactory.java
```diff
@@ -82,6 +84,64 @@ public interface SearchPredicateFactory {
...
+	 * Match documents if they match a conjunction of boolean clauses.
+	 *
+	 * @return The initial step of a DSL where the "AND" predicate can be defined.
+	 * @see AndPredicateClausesStep
+	 */
+	@SuppressWarnings("unchecked")
+	AndPredicateClausesStep<?> and(
+			Function<? super SearchPredicateFactory, ? extends PredicateFinalStep> firstSearchPredicate,
+			Function<? super SearchPredicateFactory, ? extends PredicateFinalStep>... otherSearchPredicates);
+
```
yrodiere, line 116: I think this overload is worth the trouble.

Passing a function is an exotic use case to begin with (even with `.bool()`), and here it's also problematic because of the generics + varargs.

Can you remove this overload?

We can always add an `add(Function<....>)` method in the builder variant of `and()`/`or()`.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/SearchPredicateFactory.java
```diff
@@ -82,6 +84,64 @@ public interface SearchPredicateFactory {
...
+	 * @see AndPredicateClausesStep
+	 */
+	AndPredicateClausesStep<?> and(
+			SearchPredicate firstSearchPredicate,
+			SearchPredicate... otherSearchPredicates);
+
+	/**
+	 * Match documents if they match a conjunction of boolean clauses.
+	 *
+	 * @return The initial step of a DSL where the "OR" predicate can be defined.
```
yrodiere, line 100:
```suggestion
	 * @return The initial step of a DSL where the "AND" predicate can be defined.
```

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/SearchPredicateFactory.java
```diff
@@ -82,6 +84,64 @@ public interface SearchPredicateFactory {
...
+
+	/**
+	 * Match documents if they match a disjunction of boolean clauses.
+	 *
+	 * @return The initial step of a DSL where the "OR" predicate can be defined.
+	 * @see OrPredicateClausesStep
+	 */
+	@SuppressWarnings("unchecked")
+	OrPredicateClausesStep<?> or(Function<? super SearchPredicateFactory, ? extends PredicateFinalStep> firstSearchPredicate,
+			Function<? super SearchPredicateFactory, ? extends PredicateFinalStep>... otherSearchPredicates);
```
yrodiere, line 143: Same as above, I think we can remove this.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/SmokeIT.java
```diff
@@ -140,6 +140,17 @@ public void where_boolean() {
...
+				.where( f -> f.or(
+								f.match().field( "integer" ).matching( 1 ),
+								f.match().field( "integer" ).matching( 2 )
+						)
+				)
+				.toQuery();
+		assertThatQuery( query )
+				.hasDocRefHitsAnyOrder( index.typeName(), "1", "2" )
+				.hasTotalHitCount( 2 );
+
```
yrodiere, line 153: You can skip updating this tests (`SmokeIT`), I'll remove it because it's no longer useful: https://github.com/hibernate/hibernate-search/pull/3153

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/BoolPredicateSpecificsIT.java
```diff
@@ -99,6 +99,28 @@ public void must() {
...
+				.hasNoHits();
+
+		assertThatQuery( index.query()
+				.where( f -> f.and(
+						f.match().field( "field1" ).matching( FIELD1_VALUE1 ),
+						f.match().field( "field2" ).matching( FIELD2_VALUE1 )
+				) ) )
+				.hasDocRefHitsAnyOrder( index.typeName(), DOCUMENT_1 );
+	}
+
```
yrodiere, line 123: I'd personally prefer that tests for the `AND` predicate be in `AndPredicateSpecificsIT` and tests for the OR predicate be in `OrPredicateSpecificsIT`; it seems more consistent with what we did with other predicates.

Yes it will lead to *some* duplication, but not that much, all things considered. Well worth the cost of a straightforward code structure.

If you disagree, please at least rename the test to make it clear that it's about `bool`, `and` *and* `or`.

yrodiere (commented):
Thanks! It looks good overall so far :)

> I haven't yet added builder-style constructs with .add(...) and with(...).
> I guess that should be done.

Yep, looks like a good next step.

> I haven't updated the documentation yet either.

Ok, it's fine if you do it at the end. It'll spare you some rewriting as you iterate :)

> Before I take the next step, is this going in the right direction? ;-)

I confirm you're back on tracks :)

> I have added new tests in BoolPredicateSpecificsIT, as they are more or less cloned from bool() ones.
> This makes the file name not very relevant...
> Shall I add a new IT file of its own, at the risk of more duplicated code?

That's my preference, yes, but renaming the test is also an option. See my comment below.

> If not, shall "duplicated" methods be grouped further, e.g. in @Nested test classes?

I don't think that's necessary. `@Nested` is used elsewhere, but mostly to reuse tests that we want to apply to many predicates (see `org.hibernate.search.integrationtest.backend.tck.search.predicate.BoolPredicateBaseIT.ScoreIT` for example).

> Also, aren't the tests in SmokeIT a bit redundant/pointless? Maybe they should not be added there...

They are indeed redundant. Please remove the changes to `SmokeIT` from your PR, I'll remove that class (#3153).

> So far, AndPredicateClausesStep and OrPredicateClausesStep don't implement PredicateScoreStep, which would allow calling boost(...) and constantScore() on the overall filter.
> Shouldn't they?

I think they should.

Please also remember to test the feature; you can probably just copy-paste and adapt `BooleanPredicateBaseIT` for each predicate (AND and OR).

> Similarly, so far, OrPredicateClausesStep doesn't support any of the minimumShouldMatch constraints. Shall it?
> That looks to me a bit far away from the expected semantic of an OR (with minimumShouldMatch == 1), so maybe such a feature should be reserved to more complex bool()-based predicates.

I agree with you, that's too exotic for the DSL we're trying to introduce, which is supposed to address simple use cases. Let's not add `minimumShouldMatch`.

> Shall something be done to support mustNot?
> Again, I'd leave that to bool()...

I agree a `NOT` operator would make sense, but there are additional challenges and, besides, it wasn't included in the original ticket description, so we'll address that separately. See [HSEARCH-4645](https://hibernate.atlassian.net/browse/HSEARCH-4645) (not to be addressed in this PR).

> Shall SearchPredicateFactory.and(...) and or(...) be tagged as @Incubating at first?

Good thinking, but no, I don't think that's necessary. There's relatively low risk of having to break backward compatibility here.

> As usual, using varargs with a parameterized type requires a lot of @SuppresssWarnings("unchecked"), including in calling code... :'-(
> Would you happen to have anything better to suggest?

See below, I think we'd better skip that altogether and only allow functions in the builder variant of this DSL.

YannisBres:
Hi !

I guess we agree on everything: thanks for the constructive and professional comments!

I'm a bit sad for the `and( Function<...>... )` (and `or`) overloads, but of course I can't judge [yet?] how not that relevant they are.

DONE:
- Moved new tests in dedicated `AndPredicateSpecificsIT` and `OrPredicateSpecificsIT`, instead of "polluting" `BoolPredicateSpecificsIT`.
Could you please confirm that duplicated code (around test data initialization) is accepted with all its consequences?  Don't we want to share something somewhere (with, conversely, the consequences of stronger coupling)?
- Removed the tests added to SmokeIT, as this file has been deleted in another PR.
- Corrected non-updated copy/paste in Javadoc of `SearchPredicateFactory`.

TODO:
- [ ] 2 code smells reported by Sonar: `S` is indeed not used in `AndPredicateClausesStep` and `OrPredicateClausesStep`. To be removed if chaining / a `self()` method with covariant return type is indeed unnecessary (but I guess it will be necessary).
- [ ] `AndPredicateClausesStep` and `OrPredicateClausesStep` should implement `PredicateScoreStep` in order to support `boost(...)` and `constantScore()`.
- [ ] Rename `AndPredicateClausesStep` to something like `AndPredicateOptionsStep` (see `ExistsPredicateOptionsStep` and `MatchPredicateOptionsStep`).
- [ ] Ditto for `OrPredicateClausesStep`.
- [ ] Remove the overload `AbstractSearchPredicateFactory.and( Function<...>... )` + `or( Function<...>... )`.
- [ ] Clean up copy-pasted code that ended up being unnecessary in `AndPredicateSpecificsIT` and `OrPredicateSpecificsIT`.

Carried-over TODOs:
- [ ] Add builder-style constructs with `.add(...)` and `with(...)`.
- [ ] Documentation

yrodiere:
>  Could you please confirm that duplicated code (around test data initialization) is accepted with all its consequences? Don't we want to share something somewhere (with, conversely, the consequences of stronger coupling)?

I sense you're probing for how well we apply best practices, so I'll detail a bit.

We avoid duplicating code.

In the main code, the only parts where we do duplicate code is APIs (multiple mappers, each with its own API but with subtle differences).

In tests, and actually made a great effort to avoid code duplication. Just have a look at `MatchPredicateBaseIT`, `RangePredicateBaseIT`, etc. A great many tests are shared across multiple predicates, but as you can see, it becomes much harder to understand what's going on in a particular test. For instance, you have to read through three different classes to get a complete view of what happens exactly in `org.hibernate.search.integrationtest.backend.tck.search.predicate.MatchPredicateBaseIT.AnalysisIT#multiIndex_incompatibleAnalyzer`.

So, really, as with most things, it's a trade-off. Avoiding duplication is fine but if it turns the code into a write-only, overly complex mess that external contributors cannot understand anymore (and frankly we're getting close...)... then it's not worth it.

I also generally accept higher complexity for the main code than for tests, because, well, can you trust tests that are more complex than what you're testing?

If you can find a way to "factorize" the duplicated code, have one test class for `bool`, one for `or`, and one for `and`, while keeping the code clear... then sure, I'm all for it. I just don't think it will be easy. 

> 2 code smells reported by Sonar: S is indeed not used in `AndPredicateClausesStep` and `OrPredicateClausesStep`. To be removed if chaining / a self() method with covariant return type is indeed unnecessary (but I guess it will be necessary).

This is about backwards compatibility.

Indeed, we don't need the `S` parameter right now. But if we add methods to `AndPredicateClausesStep` and `OrPredicateClausesStep`, then we might need it. And adding the `S` parameter in a later version may break backwards compatibility for our users: if their code refers to `AndPredicateClausesStep` explicitly (ideally they shouldn't have to, but well, shit happens) at best they'll get raw type warnings, at worst their code won't compile anymore.

Anyway, once you add support for `constantScore`/`boost` by extending `PredicateScoreStep`, you'll see the `S` parameter becomes necessary, so you won't have this problem anymore :)

---

Regarding the code, I'll do another review when it's complete, unless you have specific questions ;)

YannisBres:
DONE:
- Renamed `AndPredicateClausesStep` to `AndPredicateOptionsStep` and `OrPredicateClausesStep` to `OrPredicateOptionsStep`.
- `AndPredicateOptionsStep` and `OrPredicateOptionsStep` now implement `PredicateScoreStep`.
=> Cloned and adapted `BoolPredicateBaseIT` to `AndPredicateBaseIT` and `OrPredicateBaseIT`.

Carried-over TODOs:
- [ ] Add builder-style constructs with `.add(...)` and `with(...)`.
- [ ] Remove the overload `AbstractSearchPredicateFactory.and( Function<...>... )` + `or( Function<...>... )`.
- [ ] Clean up copy-pasted code that ended up being unnecessary in `AndPredicateSpecificsIT` and `OrPredicateSpecificsIT`.
- [ ] Document the new feature.

yrodiere:
> I'm wondering whether it could be relevant to introduce some intermediate abstract class below `AbstractPredicateFinalStep` in order to factorize such concerns. That would probably imply to store the `builder` at that level (=> getter or... `protected final` member [non-]variable?), to add yet another parameterized type for it and to generalize a covariant `self()` method...

That's a lot of effort for two (very) simple methods, but sure, feel free to create a ticket (a Task). It's probably going to be non-trivial, though; see `org.hibernate.search.engine.search.predicate.dsl.impl.AbstractBooleanMultiFieldPredicateCommonState#applyBoostAndConstantScore`.

And yes, renaming `thisAsS` to `self` across the board would be a good idea as well.

yrodiere:
Also:

> to add yet another parameterized type for it

I don't think that's necessary, but we shall see.

YannisBres:
DONE:
- Added builder-style constructs with `.add(...)` and `with(...)`.
`add(...)` supports both `PredicateFinalStep` and `SearchPredicate` (in order to support previously-built predicates): did that make sense?
- Removed the `AbstractSearchPredicateFactory.and( Function<...>... )` + `or( Function<...>... )` overloads.

Carried-over TODOs:
- [ ] Document the new feature.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/SearchPredicateFactory.java
```diff
@@ -82,6 +83,59 @@ public interface SearchPredicateFactory {
...
+
+	/**
+	 * Match documents if they match a conjunction of previously-built {@link SearchPredicate}.
+	 *
+	 * @return The step of a DSL where additional "AND" predicates can be defined and options be set.
+	 * @see AndPredicateOptionsStep
+	 */
+	AndPredicateOptionsStep<?> and(
+			SearchPredicate firstSearchPredicate,
+			SearchPredicate... otherSearchPredicates);
```
yrodiere, line 102: This will allow things such as:

```
f.and( e1, e2 ).add( e3 )
```

I think it's a bit odd?

I'd rather promote consistent syntax, in this case either:

```
f.and( e1, e2, e3 )
```

... or:


```
f.and().add( e1 ).add( e2 ).add( e3 )
```

... but never mixed like in the first example.

Another argument against allowing this is code completion in the IDE: users writing `.and( e1, e2 )` and then using auto-completion to see what's possible on the returned value will see the `with` and `add` methods, which could be confusing. I'd rather not expose them to that, especially considering that we're trying to introduce a simpler API here?

To that end, you could introduce a `AndPredicateClausesStep`, which extends `AndPredicateOptionsStep`, and push all `add`/`with` methods there. `and()` would return `AndPredicateClausesStep`, which allows `add`/`constantScore`/`boost`/`toPredicate` but `and(...)` would return `AndPredicateOptionsStep`, which only allows `constantScore`/`boost`/`toPredicate`.

WDYT?

(Same idea for `or`, of course)

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/OrPredicateOptionsStep.java
```diff
@@ -0,0 +1,46 @@
...
+	/**
+	 * Delegates setting clauses and options to a given consumer.
+	 * <p>
+	 * Best used with lambda expressions.
+	 *
+	 * @param contributor A consumer that will add clauses and options to the collector that it consumes.
+	 * Should generally be a lambda expression.
+	 * @return {@code this}, for method chaining.
+	 */
+	S with(Consumer<? super S> contributor);
```
yrodiere, line 29: This will allow things such as:

```
f.and().with( c -> c.toPredicate() ).add( ... )
```

Which makes no sense, because the lambda has no business calling `toPredicate` to begin with. So we probably shouldn't expose it to the consumer?

I suggest you take more inspiration from the interface hierarchy of `org.hibernate.search.engine.search.predicate.dsl.BooleanPredicateOptionsCollector`: it is complicated, but that complexity **is** solving problems, and you're facing very much the same problems.

yrodiere (commented):
>  add(...) supports both PredicateFinalStep and SearchPredicate (in order to support previously-built predicates): did that make sense?

Yes it does.

You could also add your methods that accept a `Function` (but a single function, so you don't suffer from the same compiler warnings as before) if you wish to remain consistent with `.bool()`. I'm not sure it's really useful, though, so feel free to skip that.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/OrPredicateOptionsStep.java
YannisBres: I thought this was requested in the PR (second set of patterns in "Builder-style syntaxes may be useful as well") and I took inspiration from `BooleanPredicateOptionsCollector`...  I'll have a second look!

YannisBres:
TODO :
- [ ] In `add(...)` methods, add support for `Function<? super SearchPredicateFactory, ? extends PredicateFinalStep>`.
- [ ] Make so that using `add(...)` and `with()` is not "allowed" from the `and(...)`/`or(...)` "constructors" that already allow to specify some predicates.   (Cf. https://github.com/hibernate/hibernate-search/pull/3152#discussion_r934413732.)
- [ ] Prevent support of constructs à la `f.and().with( c -> c.toPredicate() ).add( ... )`.  (Cf. https://github.com/hibernate/hibernate-search/pull/3152#discussion_r934420835.)

Carried-over TODOs:
- [ ] Document the new feature.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/OrPredicateOptionsStep.java
yrodiere: The problem is *only* that the consumer is able to call `toPredicate`; calling `add` like in the JIRA issue, or even `boost()`, in the consumer is fine. But `toPredicate` is a terminal operation, so it doesn't have its place in the `with` call (because the `with` call will return a non-terminal step and allow more changes).

YannisBres: Okay...  I think I got your point!
I reworked the class hierarchy so that it gets more similar to the `BooleanPredicateOptionsCollector` one and `toPredicate()` is now no longer reachable within `with(...)`.
I think I now get better the point of the `GenericBooleanPredicateClausesStep` interface that didn't seem to add much in term of... interface and seemed to mostly repeat the super interface: it actually refines it in order to distinguish which parameterized type is "below" the current level, and what is above it, and why the latter is only used in the `Consumer`: the point is to hide the `PredicateFinalStep` interface there.
Quite tricky but clever!

(I never needed such restrictions even in the worst DSL I ever wrote for building some kind of god object and clarify which settings must be set at instantiation time and which one may be changed later on...)

I end up with 3 interfaces and 2 classes (* 2 for `and` and `or`), as for the beef of `BooleanPredicateOptionsCollector` and couldn't manage to have less...  :'-(
But I do agree that this is worth it, especially as `toPredicate()` is "silently terminal", as it will cause the predicate to be built and all other modifications will not reset the "built state".

I will have to rework that to address https://github.com/hibernate/hibernate-search/pull/3152#discussion_r934413732, though...  :'-(

And now I'm wondering whether we should have regression tests for methods that should not be accessible in some places (=> code that should not compile)...  ;-)
Looks like Javassist also supports working on (simple) Java source code, or the javac API could be used (maybe via https://github.com/OpenHFT/Java-Runtime-Compiler), but that would be for another PR, maybe!  ;-)

yrodiere: > Okay... I think I got your point!

You did, thanks for bearing with me :)

> I end up with 3 interfaces and 2 classes (* 2 for `and` and `or`), as for the beef of `BooleanPredicateOptionsCollector` and couldn't manage to have less... :'-(

That seems fine to me. You *might* be able to take it down to 3*2 interfaces and 2*1 classes by merging the `and`/`or` implementations, with an instance member representing the operator you're working with (the `BooleanOperator` enum will come in handy). But I don't think that's absolutely necessary.

> And now I'm wondering whether we should have regression tests for methods that should not be accessible in some places (=> code that should not compile)... ;-)

That's an interesting idea, but I'm not sure that would be worth the effort: there are so many things that shouldn't compile!

>  Looks like Javassist also supports working on (simple) Java source code, or the javac API could be used (maybe via https://github.com/OpenHFT/Java-Runtime-Compiler), but that would be for another PR, maybe! ;-)

Javassist is a no-no; Hibernate projects have been moving away from it due to significant problems with recent versions of Java; from what I remember there was fundamental problems with the approach Javassist took that prevented it from working perfectly with recent versions of Java.

There are other tools for that (like the one you mentioned), but overall I think a better approach would be automatic API diffs. We've explored that in the past, but never went as far as making the build fail as soon as there's a backward-incompatible change, and certainly not as soon as there's any change, compatible or not. That might be an idea, though :)

YannisBres: > thanks for bearing with me :)

I actually totally appreciate the rigor!

> You might be able to take it down to 3*2 interfaces and 2*1 classes by merging the and/or implementations, with an instance member representing the operator you're working with [...]. But I don't think that's absolutely necessary.

I really think there's too much code that is 99% similar to keep it that way: the only added value I see here is the explicit interface/class names (with `And` / `Or` in them), and that doesn't seem to be worth the trouble.  I'd say that variations around something operator-agnostic like `SimplePredicate` should be clear enough.  Don't you think?
\+ adding support for `mustNot` will make things even worse.

> the `BooleanOperator` enum will come in handy

Hmmm...  I was more thinking about defining my own local `enum` (esp. as `BooleanOperator` doesn't support `mustNot`) in order to attach the proper `builder` method to invoke (`must`, `should`, etc.).  Or did I miss something ?

YannisBres:
DONE :
- In `add(...)` methods, add support for `Function<? super SearchPredicateFactory, ? extends PredicateFinalStep>`.
- Merged the two class hierarchies for `and` and `or` in a single one, as only one method call actually changes and interface/classes names don't seem very relevant (+ support for `mustNot` is coming)...
- Reworked the class hierarchy so that:
  - `add(...)` and `with()` are no longer accessible from the `and(...)`/`or(...)` "constructors" that allow to specify some predicates, but only from the parameterless "constructors".
  - In consumers, `toPredicate()` is no longer accessible.
  - As such, the following code doesn't compile:
  image

FTR, the class hierarchy now looks like this (I wish I had IntelliJ Ultimate...):
SimplePredicate

Carried-over TODOs:
- [ ] Document the new predicates.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/OrPredicateOptionsStep.java
yrodiere: > I'd say that variations around something operator-agnostic like SimplePredicate should be clear enough. Don't you think?

So first, I think `Simple` is a very bad name as it conveys little meaning :) `BooleanJunctionPredicateOptionsStep` might cut it, though I must admit I don't find that name very good either: it's long and not very easy to understand.

If you can find a good name, then sure, you can go a step further and expose a single set of interfaces.

> Hmmm... I was more thinking about defining my own local enum (esp. as BooleanOperator doesn't support mustNot) in order to attach the proper builder method to invoke (must, should, etc.). Or did I miss something ?

When you mention `mustNot`, you are talking about the `not` operator we'll ultimately add?

I would have expected that to be an unary operator, without the builder API. But if you think passing multiple predicates to `not` makes sense, ok, why not. Local enum it is.

yrodiere:
> * support for `mustNot` is coming

I'd say that PR is already more than enough work, let's not bother right now ;)

That's a separate ticket anyway: https://hibernate.atlassian.net/browse/HSEARCH-4645

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/OrPredicateOptionsStep.java
yrodiere: > > I'd say that variations around something operator-agnostic like SimplePredicate should be clear enough. Don't you think?
> 
> So first, I think `Simple` is a very bad name as it conveys little meaning :) `BooleanJunctionPredicateOptionsStep` might cut it, though I must admit I don't find that name very good either: it's long and not very easy to understand.
> 
> If you can find a good name, then sure, you can go a step further and expose a single set of interfaces.

Seeing your code, maybe `ScoredPredicateOptionsStep` + `CompositePredicateClausesStep` would fit?

YannisBres:
FTR, the build failure was due to an attempt to address an IntelliJ warning, which in turn triggered a Javadoc warning...  ;-)
Will undo!

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/OrPredicateOptionsStep.java
YannisBres: > So first, I think Simple is a very bad name as it conveys little meaning :)

Not a fan either.
I reasoned from the documentation, where I think we should first introduce these leaner predicates, which are meant to address many simple use cases, and then move on to more complex, but more powerful, `bool`()-based predicates.
BTW, WDYTAT?

> When you mention `mustNot`, you are talking about the `not` operator we'll ultimately add?

Well, I thought about adding support for `bool().mustNot( predicate1 ).mustNot( predicate2 )...` (and all related syntaxes), with something like `not( predicate1, predicate2, ... )`, which is actually different from what you've described in https://hibernate.atlassian.net/browse/HSEARCH-4645.
Does my proposition make sense, regardless of HSEARCH-4645?
If so, this could be very easy to support: the most complex part would be to recycle existing `mustNot` tests.  ;-)

> Seeing your code, maybe `ScoredPredicateOptionsStep` + `CompositePredicateClausesStep` would fit?

Why not...  We'll miss the opposition between "simple" and "more complex" predicates and this will make less clear that these `...Options...` and `...Clauses...` classes are related but who said that the class hierarchy should be redundantly encoded in class names?  ;-)

yrodiere: >   I reasoned from the documentation, where I think we should first introduce these leaner predicates, which are meant to address many simple use cases, and then move on to more complex, but more powerful, bool()-based predicates.
BTW, WDYTAT?

I'm not sure what you mean by "moving on to ...". You want to make changes to `.bool()`?

>  Does my proposition make sense, regardless of HSEARCH-4645?

It seems very similar to HSEARCH-4645 to me. If you want to do more than what's suggested in that ticket, maybe that's worth discussing on the ticket?

In any case, you've already worked much more than expected on this ticket, I don't think it would be wise to address more problems in the same pull request :)

> Why not... We'll miss the opposition between "simple" and "more complex" predicates and this will make less clear that these ...Options... and ...Clauses... classes are related but who said that the class hierarchy should be redundantly encoded in class names? ;-)

Right, there's a balance to be found between expressiveness and brevity. We can't always put all the information in the name :)

> What about just replacing Simple by Composite (even though more complex predicates are even more composite)? ;-)

You mean `CompositePredicateOptionsStep` + `CompositePredicateClausesStep`? That means we won't be able to use `CompositePredicateOptionsStep` for... non-composite predicates, but since we don't need to at the moment anyway, sure, that's probably best.

YannisBres: > I'm not sure what you mean by "moving on to ...". You want to make changes to `.bool()`?

I was referring to the documentation: I was thinking that we should introduce the new simple predicate combinator _before_ [9.2.11. bool: combine predicates (or/and/…​)](https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#search-dsl-predicate-boolean), and then present `bool()` as a way to combine predicates in a more powerful way, albeit with a slightly more complex syntax.

> In any case, you've already worked much more than expected on this ticket

Well, not that much in term of "CPU" time, but the wall time was really awful...
(But, you know, it's summer and...)

> I don't think it would be wise to address more problems in the same pull request :)

Okay, let's keep that in mind if you ever think it's worth it.

> `...Composite...`

I really think we should find a way to oppose `bool()`-based predicates and these new ones, that only support a single operator, at least at top-level (nesting "simple" predicates remains possible).
Looks to me like the word we are looking for is something around "uniform".  ;-)
So what about keeping the existing names (and esp. the distinction between `Options` and `Clauses` steps) and just replacing `Simple` by `Uniform`?

yrodiere: > > I'm not sure what you mean by "moving on to ...". You want to make changes to `.bool()`?
> 
> I was referring to the documentation: I was thinking that we should introduce the new simple predicate combinator _before_ [9.2.11. bool: combine predicates (or/and/…​)](https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#search-dsl-predicate-boolean), and then present `bool()` as a way to combine predicates in a more powerful way, albeit with a slightly more complex syntax.

Right; this makes sense.

> > `...Composite...`
> 
> I really think we should find a way to oppose `bool()`-based predicates and these new ones, that only support a single operator, at least at top-level (nesting "simple" predicates remains possible). Looks to me like the word we are looking for is something around "uniform". ;-) So what about keeping the existing names (and esp. the distinction between `Options` and `Clauses` steps) and just replacing `Simple` by `Uniform`?

I'm not sure. We're going into philosophical territory at this point, but let's give it a try...

`SimplePredicate*Step` works for `.and()`/`.or()`, but this name could arguably be applied to about any predicate (the `matchAll` predicate is quite "simple" after all), so it's not specific enough.

As you pointed out, `CompositePredicate*Step` suffers from a similar problem: it could apply as much to `.bool()` as to `.and()`/`.or()`.

`UniformPredicate*Step` may be clear to us while this discussion is still fresh, but I doubt it still will be two weeks from now. Besides, `and`/`or` are arguably not the only "uniform" predicates; who's to say the "match" predicate is not "uniform"? :)

`SimpleBooleanJunctionPredicate*Step` would reflect what we mean, more or less, but it's long.

`SimpleBooleanPredicate*Step` is more reasonable (still long, but I've done worse). It could potentially also apply to the `not` predicate (even though we might not expose the `*ClausesStep` for that one, but that's a separate discussion). WDYT?

YannisBres: "Boolean" along with "Predicate" always sounds like a pleonasm to me, but I do understand the will to refer to the `bool()` construct, which is what is ultimately used.

So...  `*SimpleBooleanPredicate*Step*` it is?  ;-)

yrodiere: > "Boolean" along with "Predicate" always sounds like a pleonasm to me, but I do understand the will to refer to the `bool()` construct, which is what is ultimately used.

Well we can go with `SimpleBooleanOperatorPredicate` if you want, but... we're back to a long name :)

> So... *SimpleBooleanPredicate*Step* it is? ;-)

Works for me, at least :) Feel free to use `SimpleBooleanOperatorPredicate` if you prefer that.

YannisBres: Let's go for `SimpleBooleanOperatorPredicate`, then!
Gotta find a way to justify having an `Abstract...FactoryBuilderRepositoryManager`, and we'll be just fine!  ;-)

YannisBres:
DONE:
- Renamed everything to ~`*SimpleBooleanOperatorPredicate*`.

Carried-over TODOs:
- [ ] Document the new predicates.

YannisBres:
DONE:
- First iteration on the documentation:
  - Inserted an "`and`: combine predicates as a conjunction" and an "`or`: combine predicates as a disjunction" section before the initial "`bool`: combine predicates (or/and/...)" section, renamed as "`bool`: advanced combinations of predicates (or/and/…​)".
  (Intros sound a bit verbose to me: WDYT?)
  - Replaced the examples for pure `and` and `or` predicates with "[...] using the simpler `and`/`or` syntax is recommended" and corresponding references.

TODO:
- [ ] Links to the initial `bool` section should be augmented with links to `and` and `or` as appropriate.
- [ ] Some examples that only use AND/OR pattern should be simplified with the simpler syntax: OK?
- [ ] There should be a section "Adding clauses dynamically with the lambda syntax" for the `and` predicate (and a reference to it in order to avoid repetitions in the `or` section) in order to introduce the `add` and `with` syntaxes.  However, I'm wondering whether we'd need specific "overloads" of `SearchQueryWhereStep.where(...)` for that purpose, that clearly state if we want and `and` or an `or` here...  WDYT?

Any other comments?

Here is the generated PDF for easier review: [hibernate_search_reference.pdf](https://github.com/hibernate/hibernate-search/files/9340586/hibernate_search_reference.pdf).

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -784,11 +784,78 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
 either on a per-field basis with a call to `.boost(...)` just after `.field(...)`/`.fields(...)`
 or for the whole predicate with a call to `.boost(...)` after `.matchingAny(...)` or `.matchingAll(...)`.
 
+[[search-dsl-predicate-and]]
+// TODO YB: find references to _combining_queries and add references to _anding_queries
+== [[_anding_queries]] `and`: combine predicates as a conjunction
```
yrodiere, line 789: I'm not sure where this `_anding_queries` anchor comes from? Unless there was such a (generated) anchor in previous versions of the documentation, you don't need that.

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -784,11 +784,78 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+
+The `and` predicate matches documents that match a conjunction of inner predicates, called "clauses".
+`and` clauses are required to match: if they don't match, then the `and` predicate will not match.
```
yrodiere, line 792: WDYT of that? Seems more straightforward to me.

```suggestion
The `and` predicate matches documents that match **all** of its inner predicates, called "clauses".
```

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -784,11 +784,78 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+=== Options
+
+* The score of an `and` predicate is variable by default,
+but can be <<search-dsl-predicate-common-constantScore,made constant with `.constantScore()`>>.
+* The score of an `and` predicate can be <<search-dsl-predicate-common-boost,boosted>>
+with a call to `.boost(...)`.
+
+[[search-dsl-predicate-or]]
+// TODO YB: find references to _combining_queries and add references to _oring_queries
+== [[_oring_queries]] `or`: combine predicates as a disjunction
```
yrodiere, line 821: Same comment about `_oring_queries`: where does that come from?

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -784,11 +784,78 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+
+The `or` predicate matches documents that match at least one inner predicates, called "clauses".
+If none of the clauses match, then the `or` predicate will not match.
```
yrodiere, line 824:
```suggestion
The `or` predicate matches documents that match **any** of its inner predicates, called "clauses".
```

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -784,11 +784,78 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+* The score of an `or` predicate is variable by default,
+but can be <<search-dsl-predicate-common-constantScore,made constant with `.constantScore()`>>.
+* The score of an `or` predicate can be <<search-dsl-predicate-common-boost,boosted>>
+with a call to `.boost(...)`.
+
 [[search-dsl-predicate-boolean]]
-== [[_combining_queries]] `bool`: combine predicates (or/and/...)
+== [[_combining_queries]] `bool`: advanced combinations of predicates (or/and/...)
+
+The `bool` predicate allows to combine inner predicates in a more complex fashion than with simpler
```
yrodiere, line 854: Grammar: it's either "allow <somebody> to -" or "allow -ing".
"allow to -" is incorrect.

```suggestion
The `bool` predicate allows combining inner predicates in a more complex fashion than with simpler
```

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -784,11 +784,78 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+<<search-dsl-predicate-and,`and`>> and <<search-dsl-predicate-or,`or`>> predicates.
```
yrodiere, line 855: "and and or" feels weird :)

```suggestion
<<search-dsl-predicate-and,`and`>>/<<search-dsl-predicate-or,`or`>> predicates.
```

documentation/src/test/java/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java
```diff
@@ -168,23 +168,47 @@ public void id() {
...
 			assertThat( hits )
 					.extracting( Book::getId )
 					.containsExactlyInAnyOrder( BOOK1_ID, BOOK2_ID, BOOK3_ID );
 		} );
+	}
 
+	@Test
+	public void bool() {
 		withinSearchSession( searchSession -> {
 			// tag::bool-and[]
```
yrodiere, line 213: Maybe remove that example, as it's no longer present in the documentation?

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/SearchPredicateFactory.java
```diff
@@ -82,6 +83,55 @@ public interface SearchPredicateFactory {
 	@Deprecated
 	PredicateFinalStep bool(Consumer<? super BooleanPredicateClausesStep<?>> clauseContributor);
 
+	/**
+	 * Match documents if they match a conjunction of boolean clauses.
```
yrodiere, line 87: While you are perfectly right that "conjunction"/"disjunction" are the correct words, I suspect some people will not understand it immediately. For example, think of non-native English speakers whose native language is fundamentally different, e.g. Mandarin. Or more likely, students, even native English speakers, who haven't been studying hard enough :)

In any case, if there's a way to make this documentation simpler, let's do it?

```suggestion
	 * Match documents if they match all inner clauses.
```

Same for the other methods in this class, of course.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/AndPredicateSpecificsIT.java
```diff
@@ -0,0 +1,270 @@
...
+//		assertThatQuery( index.query()
+//				.where( f -> f.and( f.match().field( "field1" ).matching( FIELD1_VALUE1 ) )
+//						.add( f.match().field( "field2" ).matching( FIELD2_VALUE2 ) ) ) )
+//				.hasNoHits();
+//
+//		// TODO YB: Should not compile
+//		assertThatQuery( index.query()
+//				.where( f -> f.and( f.match().field( "field1" ).matching( FIELD1_VALUE1 ) )
+//						.with( b -> b.add( f.match().field( "field2" ).matching( FIELD2_VALUE2 ) ) ) ) )
+//				.hasNoHits();
```
yrodiere, line 96: Unless you intend to do something about it, you should probably remove this commented-out code.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/AndWithAddPredicateBaseIT.java
```diff
@@ -0,0 +1,98 @@
...
+		scoreIndexer.join();
+	}
+
+	@Test
+	public void takariCpSuiteWorkaround() {
+		// Workaround to get Takari-CPSuite to run this test.
+	}
+
+	@Nested
+	public static class ScoreIT extends AbstractPredicateScoreIT {
```
yrodiere, line 48: Note you could move this test to `AndPredicateBaseIT`: just add another nested test next to `ScoreIT` called `AddScoreIT`, and make sure to initialize its index in `setup()`.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/OrPredicateSpecificsIT.java
```diff
@@ -0,0 +1,219 @@
...
+				.hasDocRefHitsAnyOrder( index.typeName(), DOCUMENT_1, DOCUMENT_3 );
+	}
+
+	@Test
+	public void nested() {
+		assertThatQuery( index.query()
+				.where( f -> f.and( f.or(
+						f.match().field( "field1" ).matching( FIELD1_VALUE1 ),
+						f.match().field( "field1" ).matching( FIELD1_VALUE3 )
+				) ) ) )
```
yrodiere, line 111: This is the exact same test as in `AndPredicateSpecificsIT`; shouldn't the nesting be reversed?

yrodiere (commented):
Thanks! My comments below.

>  (Intros sound a bit verbose to me: WDYT?)

I agree; I suggested shorter alternatives. Relatedly, maybe we could simplify the javadoc; I added comments there too.

>  Some examples that only use AND/OR pattern should be simplified with the simpler syntax: OK?

Seems fair.

> There should be a section "Adding clauses dynamically with the lambda syntax" for the and predicate [...] in order to introduce the add and with syntaxes

Agreed.

> (and a reference to it in order to avoid repetitions in the or section)

I'd simply duplicate the section. The content will be minimal, with the examples making up most of it, and it's precisely the examples that will be different between `and`/`or`, so... Duplication might make things clearer in this case.

>  However, I'm wondering whether we'd need specific "overloads" of SearchQueryWhereStep.where(...) for that purpose, that clearly state if we want an and or an or here... WDYT?

I don't think that's worth the trouble, especially considering that the syntax would be even more complex than the existing `.where( (f, b) -> ... )`

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
yrodiere: > (Looks like section anchors are not much used either, but that's another discussion...)

They are useful when answering user questions and pointing to the documentation.

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
YannisBres: Perfect!

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
YannisBres: Ditto!

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
YannisBres: Arf!  TIL that an infinitive complement shall only be used when a "noun phrase object (indicating the patient)" is specified, see e.g. https://english.stackexchange.com/questions/60271/grammatical-complements-for-allow/60285#60285.
Thanks for that lesson!

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
YannisBres: :+1:

documentation/src/test/java/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java
YannisBres: Arf.  I meant to do that!  ;-)

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/SearchPredicateFactory.java
YannisBres: Yup, I was also wondering about that...
So conjunction -> all, disjunction -> any!

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/AndPredicateSpecificsIT.java
YannisBres: Oops!  I never intended to commit that!
I have stashed my secret... stash of code that should not compile!  ;-)

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/AndWithAddPredicateBaseIT.java
YannisBres: Okay!  Not a big step towards less code, but a bigger one towards less files with dumb names!  ;-)

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
YannisBres: Ah!  There's a link when we hover over titles!

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
yrodiere: I just edited my comment, because `<somebody>` was being interpreted as HTML and removed, which made my comment completely nonsensical :)

YannisBres:
DONE:
- Addressed review comments.
- In the new `and` and `or` sections, added "Adding clauses dynamically with the lambda syntax" subsections.
Note that the "cascaded `with`" syntax is only explained in the `or` section, as the point is mostly to combine two different operators (=> these two different operators must have been introduced first).
- Simplified the examples that used pure `bool()`-based AND/OR (see attached PDF for easier review):
  - Example 65. Searching with a flattened structure
  - Example 146. Declaring a named predicate
  - Skipped Example 243. Using implicit nesting as it would be confusing to switch to the simplified syntax as this example is a variation of "Example 242. Matching multiple predicates against a single nested object".
  - Example 258. Boosting on a per-predicate basis
  - Example 260. Making the score of a predicate constant
  - Example 342. Targeting a field using relative paths

Here is the generated PDF for easier review: [hibernate_search_reference.pdf](https://github.com/hibernate/hibernate-search/files/9381540/hibernate_search_reference.pdf).

And now, by reading the examples with lambdas, i.e. when the operator instantiation and the `add` operations are distant from one another, I'm wondering if such syntax isn't requiring an excessive cognitive load for casual readers to figure out what kind of operation this `add` is performing.  I'd say the cognitive load would be reduced if the methods were to be more explicitly named  `and` and `or`.  For the forthcoming `not` predicate (HSEARCH-4645), renaming `add` to something like `andNot` may also be more explicit.
Note that doing so would totally mess the current class hierarchy, which is quite operator-agnostic as it is...  ;-)

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -784,11 +784,143 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+<3> By default, a boolean predicate will match nothing if there is no clause.
+To match every document when there is no clause, add an `and` clause that matches everything.
+<4> Inside the lambda, the code is free to use any Java language constructs, such as `if` or `for`,
+to control the addition of clauses.
+In this case, we only add clauses if the relevant parameter was filled in by the user.
+====
+
+Another syntax relying on the method `with(...)` can be useful when the boolean predicate is not the root predicate,
+see <<example-and-or-dynamicParameters-with,this example>>.
+
```
yrodiere, line 836:
```suggestion
```

That's not "another" syntax, that's precisely the syntax described in the example above.

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -784,11 +784,143 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+====
+
+[[search-dsl-predicate-and-lambda]]
+=== Adding clauses dynamically with the lambda syntax
+
+It is possible to define the `and` predicate inside a lambda expression.
+This is especially useful when clauses need to be added dynamically to the `and` predicate,
+for example based on user input.
+
+.Easily adding clauses dynamically using `.where(...)` and the lambda syntax
```
yrodiere, line 816:
```suggestion
.Easily adding clauses dynamically using `with(...)` and the lambda syntax
```

Until we address https://hibernate.atlassian.net/browse/HSEARCH-4676, this example will be relying on `.with`, not `.where(BiFunction)` (which was what I meant in the original title you copy/pasted and adapted).

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -784,11 +784,143 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+====
+
+[[search-dsl-predicate-or-lambda]]
+=== Adding clauses dynamically with the lambda syntax
+
+It is possible to define the `or` predicate inside a lambda expression.
+This is especially useful when clauses need to be added dynamically to the `or` predicate,
+for example based on user input.
+
+.Easily adding clauses dynamically using `.where(...)` and the lambda syntax
```
yrodiere, line 874:
```suggestion
.Easily adding clauses dynamically using `with(...)` and the lambda syntax
```

This example will be relying on `.with`, not `.where(BiFunction)` (which was what I meant in the original title you copy/pasted and adapted).

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -784,11 +784,143 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+----
+<1> Get a custom object holding the search parameters provided by the user through a web form, for example.
+<2> Call `.where(Function)` to create the root predicate, which is not the one we're interested in here.
+<3> Call `.or().with(Consumer)` to create an inner, non-root predicate.
+The consumer, implemented by a lambda expression, will receive a collector as an argument
+and will add clauses to that collector as necessary.
+<4> Inside the lambda, the code is free to use any Java language constructs, such as `if` or `for`,
+to control the addition of clauses.
+In this case, we add one clause per author filter.
+====
```
yrodiere, line 906:
```suggestion
```

That's not "another" syntax, that's precisely the syntax described in the example above.

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -784,11 +784,143 @@ but can be <<search-dsl-predicate-common-boost,boosted>>,
...
+=== Options
+
+* The score of an `or` predicate is variable by default,
+but can be <<search-dsl-predicate-common-constantScore,made constant with `.constantScore()`>>.
+* The score of an `or` predicate can be <<search-dsl-predicate-common-boost,boosted>>
+with a call to `.boost(...)`.
+
 [[search-dsl-predicate-boolean]]
-== [[_combining_queries]] `bool`: combine predicates (or/and/...)
+== `bool`: advanced combinations of predicates (or/and/...)
```
yrodiere, line 917:
```suggestion
== [[_combining_queries]] `bool`: advanced combinations of predicates (or/and/...)
```

This is a pre-existing tag from Hibernate Search 5, we need to keep it to avoid dead links.

documentation/src/main/asciidoc/reference/search-dsl-predicate.asciidoc
```diff
@@ -825,42 +957,14 @@ and are simply used for scoring.
...
-----
-<1> The hits *should* have a `title` field matching the text `robot`,
-*or* they should match any other clause in the same boolean predicate.
-<2> The hits *should* have a `description` field matching the text `investigation`,
-*or* they should match any other clause in the same boolean predicate.
-<3> All returned hits will match *at least one* of the clauses above:
-they will have a `title` field matching the text `robot`
-*or* they will have a `description` field matching the text `investigation`.
-====
+A `bool` predicate with only `should` clauses and no `minimumShouldMatch` specification will behave as an `OR` operator.
```
yrodiere, line 960:
```suggestion
A `bool` predicate with only `should` clauses and no <<search-dsl-predicate-boolean-minimumshouldmatch,`minimumShouldMatch` specification>> will behave as an `OR` operator.
```

documentation/src/test/java/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java
```diff
@@ -168,39 +168,118 @@ public void id() {
...
+								}
+							} ) );
+						}
+					} ) )
+					.fetchHits( 20 );
+			// end::and-or-dynamicParameters-with[]
 			assertThat( hits )
 					.extracting( Book::getId )
 					.containsExactlyInAnyOrder( BOOK1_ID, BOOK2_ID, BOOK3_ID );
 		} );
```
yrodiere, line 239:
```suggestion
```

To be removed if we remove the corresponding section in the Asciidoc.

yrodiere (approved):
Here are a few suggestions, and an answer below. Since you're probably busy with something else right now, I'll take care of applying the suggestions and I'll merge.

Thanks again for all this work!

--- 

> And now, by reading the examples with lambdas, i.e. when the operator instantiation and the add operations are distant from one another, I'm wondering if such syntax isn't requiring an excessive cognitive load for casual readers to figure out what kind of operation this add is performing. I'd say the cognitive load would be reduced if the methods were to be more explicitly named and and or.

That means getting back to the situation before you de-duplicated all the interfaces, then :) That might be clearer for long chains of `.add()`, but  I find it strange to use `and`/`or` for the very first clause, and that would also feel very redundant, e.g. `f.and().and( e1 ).and( e2 )`. It would also prevent developers from writing "generic" functions that simply add multiple clauses to a boolean predicate, without knowing whether it's an `and` or an `or`; not a big deal, but worth mentioning.

> In addition to making code more self-explanatory, having specific "`add`" methods would be less error-prone when different predicates are used in different `with(...)` constructs, as in "Example 228. Easily adding clauses dynamically using with(…) and the lambda syntax".

That wouldn't make things less error-prone for `and`/`and` nesting though, e.g. `f.and().and( ... ).and( f.nested().and( ... ).and( ... ) )`.

> For the forthcoming `not` predicate (HSEARCH-4645), renaming `add` to something like `andNot` may also be more explicit.

To me that's one more reason to *not* introduce and `add` syntax for `not`: make that predicate accept a *single* inner clause and leave it at that. People can still nest an `and` predicate if they want to, and we can always try to "flatten" expressions under the hood with some tricks (which we'll probably need for the `not()` predicate anyway).

--- 

Another solution would be to alter conventions, i.e. to use `and`/`or` as the name of the lambda parameter (e.g. `f.and().with( and -> ... )`). That would reduce the cognitive load when using `with`, at least, but obviously not when using the fluent syntax `f.and().add(...).add(...)`. That being said, I'd argue the fluent syntax is more suited to short chains; I’d use `.with` for longer chains spanning many lines of code.

For now, I'll use `and`/`or` as the name of the lambda parameter in tests and examples, and I'll open another ticket to re-evaluate the problem before we release 6.2.0.Final.

yrodiere:
I rebased, applied my suggestions and force-pushed to your branch. Let's see what CI has to say...

PR #3201, opened 2022-08-17
HSEARCH-4612 Disable purgeAllOnStart by default in MassIndexer when dropAndCreateSchemaOnStart is enabled

marko-bekhta:
- https://hibernate.atlassian.net/browse/HSEARCH-4612

Hey! I thought that it would also be good to log a message if someone enables two settings at the same time, but wasn't sure about the log level... not much else to add about the changes 😃 

Also is this something that we should document in the migration guide? Or as in the end, the results are still the same we don't want to clutter the doc?

### DIFF
--- a/documentation/src/main/asciidoc/reference/indexing-massindexer.asciidoc
+++ b/documentation/src/main/asciidoc/reference/indexing-massindexer.asciidoc
@@ -217,7 +217,8 @@ you can also use a schema manager to manage schemas manually at the time of your
 <<schema-management-manager>>.
 
 |`purgeAllOnStart(boolean)`
-|`true`
+|Default value depends on <<indexing-massindexer-parameters-drop-and-create-schema,`dropAndCreateSchemaOnStart(boolean)`>>.
+Defaults to `false` if the mass indexer is configured to drop and create the schema on start, to `true` otherwise.
 |Removes all entities from the indexes before indexing.
 
 Only set this to `false` if you know the index is already empty;
--- a/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/logging/impl/Log.java
+++ b/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/logging/impl/Log.java
@@ -724,4 +724,10 @@ public interface Log extends BasicLogger {
 			value = "Exception while retrieving the Jandex index for JAR '%1$s': %2$s")
 	SearchException errorDiscoveringJandexIndex(Path jarPath, String causeMessage, @Cause Exception cause);
 
+	@LogMessage(level = Logger.Level.WARN)
+	@Message(id = ID_OFFSET + 120,
+			value = "Both \"dropAndCreateSchemaOnStart()\" and \"purgeAllOnStart()\" are enabled. " +
+					"Consider having just one setting enabled as after the index is recreated there is nothing to purge.")
+	void redundantPurgeAfterDrop();
+
 }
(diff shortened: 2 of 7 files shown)


documentation/src/main/asciidoc/reference/indexing-massindexer.asciidoc
```diff
@@ -217,7 +217,8 @@ you can also use a schema manager to manage schemas manually at the time of your
 <<schema-management-manager>>.
 
 |`purgeAllOnStart(boolean)`
-|`true`
+|Default value depends on <<indexing-massindexer-parameters-drop-and-create-schema,`dropAndCreateSchemaOnStart(boolean)`>>.
+Defaults to `false`, if the mass indexer is configured to drop and create the schema on the start, to `true` otherwise.
```
yrodiere, line 221: I think it's incorrect to put a comma here in english. Also I think it's "on start"; not "on the start".

```suggestion
|Default value depends on <<indexing-massindexer-parameters-drop-and-create-schema,`dropAndCreateSchemaOnStart(boolean)`>>.
Defaults to `false` if the mass indexer is configured to drop and create the schema on start, to `true` otherwise.
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/logging/impl/Log.java
```diff
@@ -724,4 +724,10 @@ SearchException infiniteRecursionForProjectionConstructor(
 			value = "Exception while retrieving the Jandex index for JAR '%1$s': %2$s")
 	SearchException errorDiscoveringJandexIndex(Path jarPath, String causeMessage, @Cause Exception cause);
 
+	@LogMessage(level = Logger.Level.WARN)
+	@Message(id = ID_OFFSET + 120,
+			value = "Both \"dropAndCreateSchemaOnStart()\" and \"purgeAllOnStart()\" are enabled. " +
+					"Consider having just one setting enabled as after index is recreated there is nothing to purge.")
```
yrodiere, line 730:
```suggestion
					"Consider having just one setting enabled as after the index is recreated there is nothing to purge.")
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/spi/PojoMassIndexer.java
```diff
@@ -84,7 +84,8 @@ public interface PojoMassIndexer {
 	 * Set this to false only if you know there are no
 	 * entities in the indexes: otherwise search results may be duplicated.
 	 * <p>
-	 * Defaults to {@code true}.
+	 * Default value depends on {@link #dropAndCreateSchemaOnStart(boolean)}. Defaults to {@code false} if the mass indexer
+	 * is configured to drop and create the schema on the start, to {@code true} otherwise.
```
yrodiere, line 88:
```suggestion
	 * Default value depends on {@link #dropAndCreateSchemaOnStart(boolean)}. Defaults to {@code false} if the mass indexer
	 * is configured to drop and create the schema on start, to {@code true} otherwise.
```

yrodiere (changes requested):
Thanks! I added a few comments below.

> Also is this something that we should document in the migration guide? Or as in the end, the results are still the same we don't want to clutter the doc?

We only document backwards-incompatible changes. Since existing code will work just fine after this change, there's no need to document it.

marko-bekhta:
Thanks for taking a look at this one. I think it's probably about time for me to figure out how to integrate something like "grammarly" into IDE 😆 
I've amended the requested changes and rebased on top of latest main 😃

yrodiere (approved):
LGTM, thanks!

PR #3240, opened 2022-09-12
HSEARCH-4673 Use record specific methods to get constructor parameter names when applicable

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4673

Hey Yoann! It's been a while 😑😃. I noticed this task on reflection, so I thought I'd look into it. From what I understand, it all starts inside `ProjectionConstructorProcessor#process()`, which led me to these SPI HCAnn classes... 

As these are SPI, I'm not sure where it would be better to put the utils for records and if it's ok to have two implementations for the constructor or if we should keep just one and hide the logic inside...

### DIFF
--- a/documentation/src/main/asciidoc/reference/mapping-projection.asciidoc
+++ b/documentation/src/main/asciidoc/reference/mapping-projection.asciidoc
@@ -112,13 +112,16 @@ https://hibernate.atlassian.net/browse/HSEARCH-4577[at the moment].
 
 Hibernate Search infers the path of the field to project on from the name of the corresponding constructor parameter.
 
-You should set the name of a constructor parameter (in the Java code) to the name of the field to project on,
-and make sure to compile your code with the `-parameters` compiler flag.
+You should set the name of a constructor parameter (in the Java code) to the name of the field to project on.
+For the canonical constructor of record types, you will not need to do anything else.
+However, when using `@ProjectionConstructor` on non-record types or non-canonical constructors,
+code must be compiled with the `-parameters` compiler flag.
 
 [WARNING]
 ====
-Projection types compiled without the `-parameters` compiler flag
-will lead to exceptions on startup https://hibernate.atlassian.net/browse/HSEARCH-4574[at the moment].
+Without the `-parameters` compiler flag,
+errors will be thrown on startup as soon as `@ProjectionConstructor` is applied to a non-record type or a non-canonical constructor.
+Hibernate Search may provide workarounds https://hibernate.atlassian.net/browse/HSEARCH-4574[in the future].
 ====
 
 [[mapping-projection-multiple-constructors]]
--- a/integrationtest/mapper/pojo-base/pom.xml
+++ b/integrationtest/mapper/pojo-base/pom.xml
@@ -19,6 +19,10 @@
     <name>Hibernate Search ITs - POJO Mapper Base</name>
     <description>Hibernate Search integration tests for the abstract base for POJO Mappers</description>
 
+    <properties>
+        <maven.compiler.testSources.noParameterCompilation.skip>false</maven.compiler.testSources.noParameterCompilation.skip>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>org.hibernate.search</groupId>
--- a/integrationtest/mapper/pojo-base/src/test/java17-noparameters/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/ProjectionConstructorRecordNoParametersCompilerFlagIT.java
+++ b/integrationtest/mapper/pojo-base/src/test/java17-noparameters/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/ProjectionConstructorRecordNoParametersCompilerFlagIT.java
@@ -1,0 +1,145 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.integrationtest.mapper.pojo.mapping.definition;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.lang.invoke.MethodHandles;
+import java.lang.reflect.Parameter;
+import java.util.Arrays;
+import java.util.List;
+
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.DocumentId;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.FullTextField;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.GenericField;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.Indexed;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.ProjectionConstructor;
+import org.hibernate.search.mapper.pojo.standalone.mapping.SearchMapping;
+import org.hibernate.search.mapper.pojo.standalone.session.SearchSession;
+import org.hibernate.search.util.impl.integrationtest.common.rule.BackendMock;
+import org.hibernate.search.util.impl.integrationtest.common.rule.StubSearchWorkBehavior;
+import org.hibernate.search.util.impl.integrationtest.mapper.pojo.standalone.StandalonePojoMappingSetupHelper;
+
+import org.junit.Rule;
+import org.junit.Test;
+
+public class ProjectionConstructorRecordNoParametersCompilerFlagIT {
+
+	private static final String INDEX_NAME = "index_name";
+
+	@Rule
+	public BackendMock backendMock = new BackendMock();
+
+	@Rule
+	public StandalonePojoMappingSetupHelper setupHelper = StandalonePojoMappingSetupHelper.withBackendMock( MethodHandles.lookup(), backendMock );
+
+	@Test
+	public void sourcesCompiledWithoutParametersFlag() {
+		assertThat( ConstructorWithParameters.class.getDeclaredConstructors()[0].getParameters() )
+				.extracting( Parameter::isNamePresent )
+				.containsOnly( Boolean.FALSE, Boolean.FALSE );
+	}
+
+	@Test
+	public void typeLevelAnnotation() {
+		@Indexed(index = INDEX_NAME)
+		class IndexedEntity {
+			@DocumentId
+			public Integer id;
+			@FullTextField
+			public String text;
+			@GenericField
+			public Integer integer;
+		}
+		@ProjectionConstructor
+		record MyProjection(String text, Integer integer) { }
+
+		backendMock.expectAnySchema( INDEX_NAME );
+		SearchMapping mapping = setupHelper.start()
+				.withAnnotatedTypes( MyProjection.class )
+				.setup( IndexedEntity.class );
+		testSuccessfulRootProjection(
+				mapping, IndexedEntity.class, MyProjection.class,
+				Arrays.asList(
+						Arrays.asList( "result1", 1 ),
+						Arrays.asList( "result2", 2 ),
+						Arrays.asList( "result3", 3 )
+				),
+				Arrays.asList(
+						new MyProjection( "result1", 1 ),
+						new MyProjection( "result2", 2 ),
+						new MyProjection( "result3", 3 )
+				)
+		);
+	}
+
+	@Test
+	public void constructorLevelAnnotation_canonical() {
+		@Indexed(index = INDEX_NAME)
+		class IndexedEntity {
+			@DocumentId
+			public Integer id;
+			@FullTextField
+			public String text;
+			@GenericField
+			public Integer integer;
+		}
+		record MyProjection(String text, Integer integer) {
+			@ProjectionConstructor
+			public MyProjection {
+			}
+
+			public MyProjection(String text, Integer integer, String somethingElse) {
+				this( text, integer );
+			}
+		}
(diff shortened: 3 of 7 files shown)


marko-bekhta:
No worries 😃 I’ll make the changes.

marko-bekhta:
sorry that it took me ages to go back to this PR 🙊 

I have removed the `constructorLevelAnnotation_nonCanonical` case from the no-parameter-compiler-flag test as it fails without a compiler flag. 

I also thought if we want to keep the additional sources folder as we can configure the compiler plugin with includes/excludes. 

I also noticed that the build failed for the compiler-eclipse ... so I'll try to see what I can do about it.

yrodiere:
> I also thought if we want to keep the additional sources folder as we can configure the compiler plugin with includes/excludes.

We need different configuration (compiler flags) for each source folder, so we need multiple executions... not sure what you're suggesting?

> I also noticed that the build failed for the compiler-eclipse ... so I'll try to see what I can do about it.

Looks like the main build is failing as well :)

marko-bekhta:
let's see if the build works now 🙈

I have a bit more time to actually explain what I meant 😃 hopefully, it'll make sense 😉 

>We need different configuration (compiler flags) for each source folder, so we need multiple executions... not sure what you're suggesting?

If I get it correctly... adding test sources happens before their compilation, which means once we reach the step of compiling the tests, all of them are pulled into one folder, and it doesn't really matter for the compiler plugin where they came from. 
That's why I've used a combination of `testExcludes` and `testIncludes` to filter what execution compiles which classes. 

These `testExcludes` and `testIncludes` have a small problem... if the `testIncludes` will match no classes, it then decides to pick everything and all test sources get recompiled:

```bash
[INFO] --- maven-compiler-plugin:3.10.1:testCompile (default-testCompile) @ hibernate-search-documentation ---
[INFO] Changes detected - recompiling the module!
[INFO] Compiling 372 source files to /mnt/jenkins-workdir/workspace/hibernate-search_PR-3240/documentation/target/test-classes
[INFO] 
[INFO] --- maven-compiler-plugin:3.10.1:testCompile (noparameters-testCompile) @ hibernate-search-documentation ---
[INFO] Changes detected - recompiling the module!
[INFO] Compiling 372 source files to /mnt/jenkins-workdir/workspace/hibernate-search_PR-3240/documentation/target/test-classes
```
that's what caused the failure of main build -_-

but if there's a match, everything is ok and only a single test is compiled as expected:
```bash
[INFO] --- maven-compiler-plugin:3.10.1:testCompile (default-testCompile) @ hibernate-search-integrationtest-mapper-pojo-base ---
[INFO] Changes detected - recompiling the module!
[INFO] Compiling 214 source files to /home/marko/development/projects/opensource/hibernate-search/integrationtest/mapper/pojo-base/target/test-classes
[INFO] 
[INFO] --- maven-compiler-plugin:3.10.1:testCompile (noparameters-testCompile) @ hibernate-search-integrationtest-mapper-pojo-base ---
[INFO] Changes detected - recompiling the module!
[INFO] Compiling 1 source file to /home/marko/development/projects/opensource/hibernate-search/integrationtest/mapper/pojo-base/target/test-classes
```
To make it work, I've added a configurable `skip` property to `noparameters-testCompile` execution which is set to false only when there's a test that would match a filter.

documentation/src/main/asciidoc/reference/mapping-projection.asciidoc
```diff
@@ -112,12 +112,13 @@ https://hibernate.atlassian.net/browse/HSEARCH-4577[at the moment].
 
 Hibernate Search infers the path of the field to project on from the name of the corresponding constructor parameter.
 
-You should set the name of a constructor parameter (in the Java code) to the name of the field to project on,
-and make sure to compile your code with the `-parameters` compiler flag.
+You should set the name of a constructor parameter (in the Java code) to the name of the field to project on.
+Code that contains records annotated with `@ProjectionConstructor` at the type level does not require any additional actions,
+while in other cases - it must be compiled with the `-parameters` compiler flag.
```
yrodiere, line 117:
```suggestion
For the canonical constructor of record types, you will not need to do anything else.
However, when using `@ProjectionConstructor` on non-record types or non-canonical constructors,
code must be compiled with the `-parameters` compiler flag.
```

documentation/src/main/asciidoc/reference/mapping-projection.asciidoc
```diff
@@ -112,12 +112,13 @@ https://hibernate.atlassian.net/browse/HSEARCH-4577[at the moment].
...
 
 [WARNING]
 ====
-Projection types compiled without the `-parameters` compiler flag
+Projection types, except records with only a default constructor, compiled without the `-parameters` compiler flag
 will lead to exceptions on startup https://hibernate.atlassian.net/browse/HSEARCH-4574[at the moment].
```
yrodiere, line 122: That's a bit hard to follow... maybe this is better?

```suggestion
Without the `-parameters` compiler flag,
errors will be thrown on startup as soon as `@ProjectionConstructor` is applied to a non-record type or a non-canonical constructor.
Hibernate Search may provide workarounds https://hibernate.atlassian.net/browse/HSEARCH-4574[in the future].
```

integrationtest/mapper/pojo-base/pom.xml
```diff
@@ -19,6 +19,10 @@
     <name>Hibernate Search ITs - POJO Mapper Base</name>
     <description>Hibernate Search integration tests for the abstract base for POJO Mappers</description>
 
+    <properties>
+        <maven.compiler.testSources.noParamterCompilation.skip>false</maven.compiler.testSources.noParamterCompilation.skip>
```
yrodiere, line 23:
```suggestion
        <maven.compiler.testSources.noParameterCompilation.skip>false</maven.compiler.testSources.noParameterCompilation.skip>
```

pom.xml
```diff
@@ -476,6 +476,12 @@
         <maven.compiler.testSource>${maven.compiler.testRelease}</maven.compiler.testSource>
         <maven.compiler.testTarget>${maven.compiler.testRelease}</maven.compiler.testTarget>
 
+        <!--
+            This is an explicit setting to control compiler plugin execution for the no `-parameters` flag case.
+            Needed as when `testIncludes` doesn't match any test classes, it brings all test sources and re-compiles them.
+         -->
+        <maven.compiler.testSources.noParamterCompilation.skip>true</maven.compiler.testSources.noParamterCompilation.skip>
```
yrodiere, line 483:
```suggestion
        <maven.compiler.testSources.noParameterCompilation.skip>true</maven.compiler.testSources.noParameterCompilation.skip>
```

pom.xml
```diff
@@ -1793,6 +1811,26 @@
...
+                            </configuration>
+                        </execution>
+                        <execution>
+                            <id>noparameters-testCompile</id>
+                            <phase>test-compile</phase>
+                            <goals>
+                                <goal>testCompile</goal>
+                            </goals>
+                            <configuration>
+                                <skip>${maven.compiler.testSources.noParamterCompilation.skip}</skip>
```
yrodiere, line 1826:
```suggestion
                                <skip>${maven.compiler.testSources.noParameterCompilation.skip}</skip>
```

pom.xml
```diff
@@ -1793,6 +1811,26 @@
...
+                                <fork>true</fork>
+                                <proc>none</proc>
+                                <executable>${java-version.test.compiler}</executable>
+                                <compilerArgs combine.self="override" />
+                                <testIncludes>
+                                    <testInclude>**/*NoParametersCompilerFlag*</testInclude>
+                                </testIncludes>
```
yrodiere, line 1833: Is there any way you could use the directory name instead, e.g. `src/test/java17-noparameters/**`? So that we don't need tests to be named a certain way...

pom.xml
```diff
@@ -1793,6 +1811,26 @@
                                 <compilerArgs combine.self="append">
                                     <compilerArg>-parameters</compilerArg>
                                 </compilerArgs>
+                                <testExcludes>
+                                    <testExclude>**/*NoParametersCompilerFlag*</testExclude>
```
yrodiere, line 1815: Is there any way you could use the directory name instead, e.g. `src/test/java17-noparameters/**`? So that we don't need tests to be named a certain way...

pom.xml
```diff
@@ -1793,6 +1811,26 @@
...
+                            <phase>test-compile</phase>
+                            <goals>
+                                <goal>testCompile</goal>
+                            </goals>
+                            <configuration>
+                                <skip>${maven.compiler.testSources.noParamterCompilation.skip}</skip>
+                                <fork>true</fork>
+                                <proc>none</proc>
+                                <executable>${java-version.test.compiler}</executable>
+                                <compilerArgs combine.self="override" />
```
yrodiere, line 1830: Is this really necessary?

As far as I can tell, `-parameters` is not in the parent configuration, so you don't need to override it.

And by overriding, you're removing some useful parameters, such as `-Xlint:unchecked`.

integrationtest/mapper/pojo-base/src/test/java17-noparameters/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/ProjectionConstructorRecordNoParametersCompilerFlagIT.java
```diff
@@ -0,0 +1,145 @@
...
+	public BackendMock backendMock = new BackendMock();
+
+	@Rule
+	public StandalonePojoMappingSetupHelper setupHelper = StandalonePojoMappingSetupHelper.withBackendMock( MethodHandles.lookup(), backendMock );
+
+	@Test
+	public void sourcesCompliedWithoutParametersFlag() {
+		assertThat( ConstructorWithParameters.class.getDeclaredConstructors()[0].getParameters() )
+				.extracting( Parameter::isNamePresent )
+				.containsOnly( Boolean.FALSE, Boolean.FALSE );
```
yrodiere, line 44: :+1:

yrodiere (commented):
I noticed a few typos, and added a few questions.

integrationtest/mapper/pojo-base/pom.xml
marko-bekhta: 🙈 😢 😞 🥲

pom.xml
marko-bekhta: hmm, I'll give it a try to see if all would work fine and push the change. The idea for overriding it here was to make 10000% sure that we see all parameters we are passing for this execution so that either now or in the future nothing will sneak in 😃

pom.xml
yrodiere: That's a bummer :/

> Because of how the filter works (ignores the source location), that was why I was asking if there's a point to keeping sources separately.

You're right, if we can't differentiate between source directories in the maven-compiler-plugin, then it doesn't make sense to use a different directory for tests compiled without `-parameters`.

Still, filtering by class name feels very fragile, as class renaming is not unusual is and is usually expected to have to side-effect... Maybe you could use the package name in the filter instead? Not sure that would be much better, but perhaps a little bit...

At least the test will fail if compilation used `-parameters`, so I guess there's that...

integrationtest/mapper/pojo-base/src/test/java17-noparameters/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/ProjectionConstructorRecordNoParametersCompilerFlagIT.java
```diff
@@ -0,0 +1,145 @@
...
+	private static final String INDEX_NAME = "index_name";
+
+	@Rule
+	public BackendMock backendMock = new BackendMock();
+
+	@Rule
+	public StandalonePojoMappingSetupHelper setupHelper = StandalonePojoMappingSetupHelper.withBackendMock( MethodHandles.lookup(), backendMock );
+
+	@Test
+	public void sourcesCompliedWithoutParametersFlag() {
```
yrodiere, line 41:
```suggestion
	public void sourcesCompiledWithoutParametersFlag() {
```

yrodiere (commented):
One more typo I noticed by chance. I also answered regarding the source filtering in maven-compiler-plugin.

marko-bekhta:
ok. rebased to include the latest changes 😃 

I've added a new execution for checkstyle to enforce the `PackageName` and updated the filters in the compiler plugin to use package rather than classname. I've left the `SuppressionCommentFilter` in there just in case. 

I have also tried to remove the `<compilerArgs combine.self="override" />` but with no luck - if it's not there, the sources are compiled with `-parameters`.  (could it have something to do with https://github.com/hibernate/hibernate-search/blob/main/parents/integrationtest/pom.xml#L71-L72 ?)

Let's see if the build will pass on CI 🤞🏻

yrodiere:
> I've added a new execution for checkstyle to enforce the `PackageName` and updated the filters in the compiler plugin to use package rather than classname. I've left the `SuppressionCommentFilter` in there just in case.

:+1: 

> I have also tried to remove the `<compilerArgs combine.self="override" />` but with no luck - if it's not there, the sources are compiled with `-parameters`. (could it have something to do with https://github.com/hibernate/hibernate-search/blob/main/parents/integrationtest/pom.xml#L71-L72 ?)

Right, that's probably the culprit. Changing that code you linked to configure the `default-compile` execution only should do the trick:

```
                        <execution>
                            <id>default-compile</id>
                            <configuration>
                                <compilerArgs combine.self="append">
                                    <compilerArg>-parameters</compilerArg>
                                </compilerArgs>
                            </configuration>
                        </execution>
```

yrodiere (approved):

PR #3266, opened 2022-10-07
HSEARCH-4673 Ignore noparameters tests for jdk 8/11

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4673

Update to make JDK 8/11 builds pass.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -481,7 +481,7 @@
             Needed as when `testIncludes` doesn't match any test classes, it brings all test sources and re-compiles them.
          -->
         <maven.compiler.testSources.noParameterCompilation.skip>true</maven.compiler.testSources.noParameterCompilation.skip>
-
+        <java-version.test.java17.noParameters.skip>${maven.compiler.testSources.noParameterCompilation.skip}</java-version.test.java17.noParameters.skip>
         <!--
                 Options to compile with the Eclipse compiler when building with maven (not with the IDE).
                 See profile "compiler-eclipse".
@@ -1823,7 +1823,7 @@
                                 <goal>testCompile</goal>
                             </goals>
                             <configuration>
-                                <skip>${maven.compiler.testSources.noParameterCompilation.skip}</skip>
+                                <skip>${java-version.test.java17.noParameters.skip}</skip>
                                 <fork>true</fork>
                                 <proc>none</proc>
                                 <executable>${java-version.test.compiler}</executable>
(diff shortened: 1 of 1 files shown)


integrationtest/mapper/pojo-base/pom.xml
```diff
@@ -59,5 +59,40 @@
...
+            </activation>
+            <properties>
+                <!--
+                We want to skip -noparameter compilation when testing against JDK < 17 as tests for this execution are
+                for 17+ versions only.
+                -->
+                <maven.compiler.testSources.noParameterCompilation.skip>true</maven.compiler.testSources.noParameterCompilation.skip>
+            </properties>
+        </profile>
+    </profiles>
```
yrodiere, line 96: Shouldn't this be in the root pom, like (almost) everything else related to this property (`maven.compiler.testSources.noParameterCompilation.skip`)?

yrodiere: Ah, I see, that's because you override something defined in this POM. There are ways to control this from the root POM anyway, let me check...

yrodiere: So, I *think* we could do something similar to what we do for ORM tests and DB profiles:

https://github.com/hibernate/hibernate-search/blob/684018b3b17f004ae1b3aad3091fda54ba118c8e/integrationtest/mapper/orm/pom.xml#L15-L17

https://github.com/hibernate/hibernate-search/blob/aaa193330ced3a5858daa00a940de1d4ece9f5e2/parents/integrationtest/pom.xml#L695-L698

Basically in your case, I think you would:

* revert this change
* introduce a different property, e.g. `java-version.test.java17.noParameters.skip`, to control whether the maven-compiler execution is skipped
* set `java-version.test.java17.noParameters.skip` to `${maven.compiler.testSources.noParameterCompilation.skip}` by default (in the properties of the root POM)
* force `java-version.test.java17.noParameters.skip` to `false` in the JDK8/JDK11 profiles

marko-bekhta: I'll give it a try a bit later, have to step out -_-

yrodiere (approved):

yrodiere:
Thanks. LGTM, feel free to merge once CI passes.

I re-launched CI to execute all builds (JDK8, ...).

PR #3267, opened 2022-10-10
HSEARCH-4705 Merge ES test dialects 

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4705

- replaced the `getDialect()` with `getActualVersion()` as now there's single dialect.
- there were some unused request-related methods in dialect - removed those.
- `ElasticsearchVersionUtils` has the actual version as the first parameter. I thought about hiding it initially, but as the method call doesn't look too cumbersome and makes the utils more testable, I thought I'll leave it. But if you think it would be better to hide it - let me know 😃 
- I've also thought about deleting the dialect, but there are a few tests without `TestElasticsearchClient`, so we'd need to add it if I'd push the methods into it. Also having the dialect groups the properties together - so I've left the idea 😄 

I've tested this against ES 5.6/6.8/7.7. Maybe if there'll be no changes needed we could run a build against all ES that we test on CI?


yrodiere (commented):
A few comments below.

As to running tests against all versions of Elasticsearch/OpenSearch, you can go there: https://ci.hibernate.org/job/hibernate-search/job/PR-3267/
Log in, then click "Build with parameters" to the left, then either:

* Select SUPPORTED in "ENVIRONMENT_SET" to run tests on absolutely every supported environment (all databases, all versions of Elasticsearch/OpenSearch, AWS, ...)
* OR type `elastic|opensearch` in "ENVIRONMENT_FILTER" to just run tests in environments related to Elasticsearch/OpenSearch (ignoring databases etc.).

Then click "Build".

util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/dialect/ElasticsearchVersionUtils.java
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	public static int compare(ElasticsearchVersion a, ElasticsearchVersion b) {
+		if ( !a.distribution().equals( b.distribution() ) ) {
+			throw new IllegalArgumentException( "Cannot compare different distributions" );
+		}
+
+		return Comparator.comparing( ElasticsearchVersion::major )
+				.thenComparing( version -> version.qualifier().orElse( "" ) )
+				.thenComparing( version -> version.minor().orElse( 0 ) )
```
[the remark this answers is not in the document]
marko-bekhta: > For example if you do something like isActualVersionBetweenIncluding( version, "7.0", "8.0")... wouldn't you want 8.0.1 to be included?

Hmm, I see your point... I'll see what can be done with it 😃

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/util/ElasticsearchTckBackendFeatures.java
```diff
@@ -6,31 +6,46 @@
...
+		this.actualVersion = ElasticsearchTestDialect.getActualVersion();
 	}
 
 	@Override
 	public boolean geoPointIndexNullAs() {
-		return dialect.supportsGeoPointIndexNullAs();
+		if ( isActualVersionLess( ElasticsearchTestDialect.getActualVersion(), "elastic:6.3.0" ) ) {
+			return false;
+		}
+		return true;
```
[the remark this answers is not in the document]
marko-bekhta: Yeah... I'd usually go with simply returning the expression instead of this if/return/else/return thing. If that's a code style - let's make it consistent then. I've made it like it here as we wanted to be more verbose and clear on what's happening for particular versions, and this looked more human-friendly 😄.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/util/ElasticsearchTckBackendFeatures.java
```diff
@@ -207,7 +256,22 @@ public boolean supportsFieldSortWhenFieldMissingInSomeTargetIndexes(Class<?> fie
...
+		// In 6.8.0 and below, we just can't ignore unmapped nested fields in field sorts,
+		// which means sorts will fail when the nested field is not present in all indexes.
+		// -----------------------------------------------------------------------------------------
+		// AWS apparently didn't apply this patch, which solves the problem in 6.8.1/7.1.2,
+		// to their 6.8 branch:
+		// https://github.com/elastic/elasticsearch/pull/42451
+
+		if ( isActualVersionLessOrEquals( actualVersion, "elastic:6.8.0" ) ||
+				ElasticsearchTestHostConnectionConfiguration.get().isAws() &&
+						isActualVersionEquals( actualVersion, "elastic:6.8.1" ) ) {
```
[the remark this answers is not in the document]
marko-bekhta: I've checked this part once more. I've misiterpreted the comment and was thinking that with 6.8.2+ it got fixed on AWS as well. I'm not sure abou the 7.11 part ... looking at the hierarchy of dialects it seems that
- 6.7  and earlier - `false`
- 6.8 - `true` unless on AWS (false)
- 7.0+ - `true`
is my undertanding correct here?

marko-bekhta:
I've made some changes to the comparator using the min/max int values as defaults depending on the operation we are applying. This should work for as long as the actual version doesn't have empty minor and micro parts. 

I've dropped the qualifier part entirely. My thoughts here are that if we'd have some `CR` or `beta` versions - we'd probably support them for a short period. And in such case, we'd use a full version match.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/util/ElasticsearchTckBackendFeatures.java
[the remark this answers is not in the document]
yrodiere: You are right, I misinterpreted the current dialect hierarchy. All the more reason to kill off that monstrosity :)

yrodiere (approved):
Thanks!

I added a few comments to address things I forgot to report earlier (sorry!); I expect this is just as fast as explaining what I meant, especially when it comes to renamings. Please let me know if you agree?

I just launched a CI build for all versions of ES/OpenSearch as well.

marko-bekhta:
Thanks! Looks good and makes sense 😃 I will double check on this condition though - `return !isBetween( actualVersion, "elastic:8.4.2", "elastic:8.5.0" );`  the lower bound seems a bit strange

util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/dialect/ElasticsearchTestDialect.java
```diff
@@ -6,86 +6,69 @@
...
 
-	Optional<URLEncodedString> getTypeNameForMappingAndBulkApi();
+	public boolean isEmptyMappingPossible() {
+		return isLessThan( ACTUAL_VERSION, "elastic:6.9.0" );
+	}
 
-	Boolean getIncludeTypeNameParameterForMappingApi();
+	@SuppressWarnings("deprecation")
+	public Optional<URLEncodedString> getTypeNameForMappingAndBulkApi() {
+		if ( isLessThan( ACTUAL_VERSION, "elastic:6.9.0" ) ) {
```
yrodiere, line 43: This seems to be causing errors on 6.9.

Note `Elasticsearch68TestDialect` was being used on Elasicsearch 6.9 as well...

I think this should be:

```suggestion
		if ( isLessThan( ACTUAL_VERSION, "elastic:7.0.0" ) ) {
```

util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/dialect/ElasticsearchTestDialect.java
```diff
@@ -6,86 +6,69 @@
...
-	boolean isEmptyMappingPossible();
-
-	URLEncodedString getTypeKeywordForNonMappingApi();
+	public static ElasticsearchVersion getActualVersion() {
+		return ACTUAL_VERSION;
+	}
 
-	Optional<URLEncodedString> getTypeNameForMappingAndBulkApi();
+	public boolean isEmptyMappingPossible() {
+		return isLessThan( ACTUAL_VERSION, "elastic:6.9.0" );
```
yrodiere, line 38: Note `Elasticsearch68TestDialect` was being used on Elasicsearch 6.9 as well...

I think this should be:

```suggestion
		return isLessThan( ACTUAL_VERSION, "elastic:7.0.0" );
```

util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/dialect/ElasticsearchTestDialect.java
```diff
@@ -6,86 +6,69 @@
...
+		if ( isLessThan( ACTUAL_VERSION, "elastic:6.9.0" ) ) {
+			return Optional.of( Paths.DOC );
+		}
+		return Optional.empty();
+	}
 
-	ElasticsearchRequest createTemplatePutRequest(String templateName, String pattern, int priority, JsonObject settings);
+	public Boolean getIncludeTypeNameParameterForMappingApi() {
+		if ( isBetween( ACTUAL_VERSION, "elastic:6.4.0", "elastic:6.9.0" ) ) {
+			return Boolean.TRUE;
```
yrodiere, line 51: This seems wrong? From what I can see, we were returning true for `Elasticsearch68TestDialect` only, which means for ES 6.8 and 6.9 only.

I think this should be:

```suggestion
		if ( isBetween( ACTUAL_VERSION, "elastic:6.8", "elastic:6.9" ) ) {
			return Boolean.TRUE;
```

yrodiere (changes requested):
Hey.

The build errored out, so I had a look at the version conditions. There seems to be a few problems, but I'm not sure how many. Unless I'm mistaken, the problems were not caused by my changes, but were already there in your original PR. Can you please double-check each and every condition?

To be clear, the way it works is: at version X, we will use the highest test dialect whose version is lower than or equal to X.

So for Elasticsearch 6.9, since `Elasticsearch69TestDialect` does not exist, we will use `Elasticsearch68TestDialect` (and not `Elasticsearch70TestDialect`).

So the behavior you can see in `Elasticsearch68TestDialect` should encompass ES 6.8 to 6.9. Which translates to something like `isBetween(version, "6.8", "6.9")`, I believe.

util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/dialect/ElasticsearchTestDialect.java
marko-bekhta: ooooooooooooooooook 🥲 so this was 
- `null` in 64 dialect 
- inherited `true` in 67 dialect 
- `true` in 68 dialect 
- `null` in 70-712 dialects 
but then in the pom ... 6.4 is actually a 6.6. Which ... I suppose, this converts the condition to `isBetween( ACTUAL_VERSION, "elastic:6.6", "elastic:6.9" )` 
running the tests for 6.4 6.7 and 6.8 profiles now 🥲

yrodiere: >  but then in the pom ... 6.4 is actually a 6.6. Which ... I suppose, this converts the condition to isBetween( ACTUAL_VERSION, "elastic:6.6", "elastic:6.9" )

Don't you mean `isBetween( ACTUAL_VERSION, "elastic:6.7", "elastic:6.9" )`? Since 6.6 was being tested with the 6.4 profile, which uses `null`?

util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/dialect/ElasticsearchTestDialect.java
```diff
@@ -6,86 +6,69 @@
...
+	public Optional<URLEncodedString> getTypeNameForMappingAndBulkApi() {
+		if ( isLessThan( ACTUAL_VERSION, "elastic:6.9.0" ) ) {
+			return Optional.of( Paths.DOC );
+		}
+		return Optional.empty();
+	}
 
-	ElasticsearchRequest createTemplatePutRequest(String templateName, String pattern, int priority, JsonObject settings);
+	public Boolean getIncludeTypeNameParameterForMappingApi() {
+		if ( isBetween( ACTUAL_VERSION, "elastic:6.7", "elastic:6.9.0" ) ) {
```
yrodiere, line 50: Seems suspicious that `6.9.0` is included, but not `6.9.1`?

```suggestion
		if ( isBetween( ACTUAL_VERSION, "elastic:6.7", "elastic:6.9" ) ) {
```

yrodiere (approved):
Ok, I don't know why I was worried about 6.9 since, as you pointed out, there *is* no 6.9 version of Elasticsearch... Perhaps the ranges (e.g. `[6.8,6.9)`) we used in the CI job tags got me confused.

Anyway... I see you made some changes (thanks), so I launched CI on all supported versions of Elasticsearch/OpenSearch again. Let's see how it goes.

PR #3269, opened 2022-10-14
HSEARCH-4714 add property path value validation

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4714

this would produce something like the following:
```
3:41:28,715 (main) ERROR RootFailureCollector:231 - HSEARCH000521: Hibernate Search encountered a failure during bootstrap; continuing for now to list all problems, but the process will ultimately be aborted.
Context: Standalone POJO mapping, type 'org.hibernate.search.integrationtest.mapper.pojo.mapping.IncorrectPropertyNameObjectPathIT$PhoneNumber', path '.owner', annotation '@org.hibernate.search.mapper.pojo.mapping.definition.annotation.AssociationInverseSide(extraction=@org.hibernate.search.mapper.pojo.extractor.mapping.annotation.ContainerExtraction(value={}, extract=DEFAULT), inversePath=@org.hibernate.search.mapper.pojo.mapping.definition.annotation.ObjectPath({@org.hibernate.search.mapper.pojo.mapping.definition.annotation.PropertyValue(extraction=@org.hibernate.search.mapper.pojo.extractor.mapping.annotation.ContainerExtraction(value={}, extract=DEFAULT), propertyName="broken.path.with.dots")}))'
Failure:
org.hibernate.search.util.common.SearchException: HSEARCH700121: Invalid ObjectPath encountered '@org.hibernate.search.mapper.pojo.mapping.definition.annotation.ObjectPath({@org.hibernate.search.mapper.pojo.mapping.definition.annotation.PropertyValue(extraction=@org.hibernate.search.mapper.pojo.extractor.mapping.annotation.ContainerExtraction(value={}, extract=DEFAULT), propertyName="broken.path.with.dots")})'. See cause for details.
	at org.hibernate.search.mapper.pojo.mapping.definition.annotation.processing.impl.AbstractMappingAnnotationProcessorContext.toPojoModelPathValueNode(AbstractMappingAnnotationProcessorContext.java:40) ~[classes/:?]
.....
Caused by: java.lang.IllegalArgumentException: HSEARCH900015: Property name 'broken.path.with.dots' is invalid. The value must match a property in a corresponding class.
	at org.hibernate.search.util.common.impl.Contracts.assertTrue(Contracts.java:66) ~[classes/:?]
	at org.hibernate.search.mapper.pojo.model.path.PojoModelPathPropertyNode.<init>(PojoModelPathPropertyNode.java:30) ~[classes/:?]
...
```

### DIFF
--- a/integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/IncorrectPropertyNameObjectPathIT.java
+++ b/integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/IncorrectPropertyNameObjectPathIT.java
@@ -1,0 +1,98 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.integrationtest.mapper.pojo.mapping;
+
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import java.lang.invoke.MethodHandles;
+import java.util.Collection;
+
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.AssociationInverseSide;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.DocumentId;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.GenericField;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.Indexed;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.ObjectPath;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.PropertyValue;
+import org.hibernate.search.util.common.SearchException;
+import org.hibernate.search.util.impl.integrationtest.common.rule.BackendMock;
+import org.hibernate.search.util.impl.integrationtest.mapper.pojo.standalone.StandalonePojoMappingSetupHelper;
+
+import org.junit.Rule;
+import org.junit.Test;
+
+public class IncorrectPropertyNameObjectPathIT {
+	private static final String BROKEN_PATH_WITH_DOTS = "broken.path.with.dots";
+	@Rule
+	public BackendMock defaultBackendMock = new BackendMock();
+
+	@Rule
+	public StandalonePojoMappingSetupHelper setupHelper = StandalonePojoMappingSetupHelper.withBackendMock(
+			MethodHandles.lookup(),
+			defaultBackendMock
+	);
+
+	@Test
+	public void brokenPathWithDotsThrowsException() {
+		assertThatThrownBy( () -> {
+			setupHelper.start()
+					.withAnnotatedEntityType( Person.class, Person.ENTITY_NAME )
+					.withAnnotatedEntityType( PhoneNumber.class, PhoneNumber.ENTITY_NAME )
+					.setup();
+		} ).isInstanceOf( SearchException.class )
+				.hasMessageContaining( "Property name '" + BROKEN_PATH_WITH_DOTS + "' is invalid." );
+	}
+
+
+	@Indexed(index = Person.INDEX_NAME)
+	private static class Person {
+		public static final String ENTITY_NAME = "PersonEntity";
+		public static final String INDEX_NAME = "Person";
+
+		@DocumentId
+		private Integer id;
+		@GenericField
+		private String name;
+
+		private Collection<PhoneNumber> phoneNumbers;
+
+		public Person() {
+		}
+
+		public Integer getId() {
+			return id;
+		}
+
+		public String getName() {
+			return name;
+		}
+	}
+
+	private static class PhoneNumber {
+		public static final String ENTITY_NAME = "PhoneNumber";
+
+		private Integer id;
+		@GenericField
+		private String number;
+		@AssociationInverseSide(inversePath = @ObjectPath(@PropertyValue(propertyName = BROKEN_PATH_WITH_DOTS)))
+		private Person owner;
+
+		public PhoneNumber() {
+		}
+
+		public Integer getId() {
+			return id;
+		}
+
+		public String getNumber() {
+			return number;
+		}
+
+		public Person getOwner() {
+			return owner;
+		}
+	}
+}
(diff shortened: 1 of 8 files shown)


mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/logging/impl/Log.java
```diff
@@ -739,4 +740,8 @@ SearchException infiniteRecursionForProjectionConstructor(
 					"Consider having just one setting enabled as after the index is recreated there is nothing to purge.")
 	void redundantPurgeAfterDrop();
 
+	@Message(id = ID_OFFSET + 121,
+			value = "Invalid ObjectPath encountered '%s'. See cause for details.")
+	SearchException invalidObjectPath(ObjectPath path, @Cause Exception cause);
```
yrodiere, line 745: Please include the nested exception message like we do elsewhere:

```suggestion
			value = "Invalid ObjectPath encountered '%1$s': $2$s")
	SearchException invalidObjectPath(ObjectPath path, String causeMessage, @Cause Exception cause);
```

Two reasons for doing that:

1. Hibernate Search produces failure reports that aggregate all mapping failures in a tree. That failure report only includes the "outermost" exception, without stacktrace or the message of nested exceptions. So we need the "outermost" exception message to be self-sufficient.
2. A lot of Java users will not look at nested exceptions. Some novice users won't think of it, while some more experienced users just won't like to scroll down and get irritated.

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/IncorrectPropertyNameObjectPathIT.java
```diff
@@ -0,0 +1,101 @@
...
+	public void brokenPathWithDotsThrowsException() {
+		assertThatThrownBy( () -> {
+			setupHelper.start()
+					.withAnnotatedEntityType( Person.class, Person.ENTITY_NAME )
+					.withAnnotatedEntityType( PhoneNumber.class, PhoneNumber.ENTITY_NAME )
+					.setup();
+		} ).isInstanceOf( SearchException.class )
+				.hasMessageContainingAll(
+						"propertyName=\"" + BROKEN_PATH_WITH_DOTS + "\"",
+						"Invalid ObjectPath encountered"
```
yrodiere, line 48: I would also assert that this message contains more details (in particular the message saying something like "a property name cannot contain a dot").

util/common/src/main/java/org/hibernate/search/util/common/impl/Contracts.java
```diff
@@ -60,4 +60,10 @@ public static void assertNoNullElement(Collection<?> collection, String collecti
 			throw log.collectionMustNotContainNullElement( collectionDescription );
 		}
 	}
+
+	public static void assertTrue(boolean condition, String message) {
+		if ( !condition ) {
+			throw log.conditionFailed( message );
+		}
+	}
```
yrodiere, line 68: Is that how it's done in Hibernate Validator?
I don't really see the added value compared to inlining the `if`, but if you feel strongly about it, I won't stop you...

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/model/path/PojoModelPathPropertyNode.java
```diff
@@ -25,8 +26,17 @@ public final class PojoModelPathPropertyNode extends PojoModelPath {
 	private final String propertyName;
 
 	PojoModelPathPropertyNode(PojoModelPathValueNode parent, String propertyName) {
-		this.parent = parent;
 		Contracts.assertNotNullNorEmpty( propertyName, "propertyName" );
+		Contracts.assertTrue(
+				DOT_PATTERN.split( propertyName, -1 ).length == 1,
```
yrodiere, line 31: Why split? This would work fine:

```suggestion
				!DOT_PATTERN.matcher( propertyName ).find(),
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/model/path/PojoModelPathPropertyNode.java
```diff
@@ -25,8 +26,17 @@ public final class PojoModelPathPropertyNode extends PojoModelPath {
...
+				String.format(
+						Locale.ROOT,
+						"Property name '%s' is invalid. The value must match a property in a corresponding class.",
+						propertyName
+				)
```
yrodiere, line 36: You should probably add a new `@Message` method that returns String to the appropriate `Log` class. Otherwise this message cannot be localized, and given it is rather specific to Hibernate Search, I'd rather leave the door open to localization.

Personally, I would even skip the `assertTrue` thingy and just add a method that returns an exception to the `Log` class, but it's up to you.

Also, about the message... can't we simply say `property names must not contain any dot (".")`? Since, well, the purpose of this issue was to produce clearer error messages :)
We can always elaborate as to what property names represent, like you did, but I'd only do that after explaining what's wrong exactly.

yrodiere (changes requested):
Thanks! A few comments :)

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/logging/impl/Log.java
marko-bekhta: > won't like to scroll down and get irritated.

Hah, it's funny - I'm so used to scrolling and never even check the initial log entries as it's so rarely they have anything useful 😄

util/common/src/main/java/org/hibernate/search/util/common/impl/Contracts.java
marko-bekhta: Yeah, this is how things are done in HV. I've been using a similar approach in other projects. But it's ok, you can stop me 😆

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/model/path/PojoModelPathPropertyNode.java
marko-bekhta: honestly, I'd just go with the simple `propertyName.contains(".")`, which might even work faster than patterns. But as the task was saying:
>property name doesn’t contain any dot or any metacharacter that org.hibernate.search.mapper.pojo.model.path.PojoModelPath#parse understands

I thought I'll make it as close to the parse method as possible 😃

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/model/path/PojoModelPathPropertyNode.java
marko-bekhta: >Also, about the message... 

I was torn about this part `or any metacharacter` even though the pattern seems to be using just the dot. So I wasn't sure if the set of characters will change and it won't be just about the dots or if it'll remain as is. anyway - the message is updated 😃

yrodiere (approved):
LGTM, thanks!

PR #3270, opened 2022-10-17
HSEARCH-4703 Adjust module-infos and add more tests

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4703

This was an interesting one 😄 . It also uncovered a bug in the `ClassPathAndModulePathServiceResolver`  I wasn't sure if that should go into its own ticket or if it can be a part of this PR as the change is just one line (see the last commit).

You'll also notice there's `//TODO: uncomment out and figure out what's not working`. I wanted to check with you if that part for schema validation is expected to work with Lucene or not. If it is, I'll try to find out why it behaves like that in Lucene's case.


engine/src/main/java/org/hibernate/search/engine/environment/classpath/spi/DefaultServiceResolver.java
```diff
@@ -143,7 +143,7 @@ public <S> Collection<S> loadJavaServices(Class<S> serviceContract) {
 			final Set<String> alreadyUsedProviderTypes = new LinkedHashSet<>();
 			final Set<S> services = new LinkedHashSet<>();
 			this.classLoaders.stream()
-					.map( classLpader -> ServiceLoader.load( serviceContract, aggregatedClassLoader ) )
+					.map( classLoader -> ServiceLoader.load( serviceContract, classLoader ) )
```
yrodiere, line 146: Nice catch!

integrationtest/java/modules/orm-lucene/src/main/java/module-info.java
```diff
@@ -0,0 +1,33 @@
+@SuppressWarnings( "requires-transitive-automatic" )
```
yrodiere, line 1: Looks like this warning is right, and we shouldn't be using `requires transitive` on an automatic module, because such a module never has any dependencies?

integrationtest/java/modules/orm-lucene/src/test/java/org/hibernate/search/integrationtest/java/modules/orm/lucene/service/JavaModulePathIT.java
```diff
@@ -0,0 +1,52 @@
...
+		assertEquals( 2, service.search( "foo" ).size() );
+
+		//TODO: uncomment out and figure out what's not working:
+//		assertThatThrownBy( service::triggerValidationFailure )
+//				.isInstanceOf( SearchException.class )
+//				.hasMessageContainingAll(
+//						"Hibernate Search encountered failures during Schema management",
+//						"attribute 'analyzer'",
+//						"Invalid value. Expected 'default', actual is 'myAnalyzer'"
+//				);
```
yrodiere, line 43: Lucene schema validation actually doesn't do much, and indeed it never checks analyzers: https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#mapper-orm-schema-management-concepts-lucene-schema

So yes, you can drop this and all code related to the schema validation failure (`AuthorWithSlightlyDifferentIndex`, etc.) in this module.

integrationtest/java/modules/orm-lucene/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+    <properties>
+        <test.elasticsearch.run.skip>true</test.elasticsearch.run.skip>
+        <test.database.run.skip>false</test.database.run.skip>
+
+        <maven.compiler.release>11</maven.compiler.release>
+        <!-- Override the default from the parent POM: here we really do want to use the modulepath -->
+        <failsafe.useModulePath>true</failsafe.useModulePath>
+
+        <surefire.jvm.args.module>
+            ${test.elasticsearch.connection.jvm.args}
```
yrodiere, line 28: For Lucene? I think not :)

integrationtest/java/modules/orm-lucene/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+        <relativePath>../../../pom.xml</relativePath>
+    </parent>
+    <artifactId>hibernate-search-integrationtest-java-modules-orm-lucene</artifactId>
+    <packaging>jar</packaging>
+
+    <name>Hibernate Search ITs - Java - Modules - ORM/Lucene</name>
+    <description>Hibernate Search integration tests for Java 11+ modules</description>
+
+    <properties>
+        <test.elasticsearch.run.skip>true</test.elasticsearch.run.skip>
```
yrodiere, line 20: This is the default.

integrationtest/java/modules/orm-lucene/src/main/resources/hibernate.properties
```diff
@@ -0,0 +1,17 @@
...
+hibernate.max_fetch_depth = 5
+# We can't use classes from the hibernate-testing module unless we add an explicit dependency to that module.
+#hibernate.cache.region_prefix hibernate.test
+#hibernate.cache.region.factory_class = org.hibernate.testing.cache.CachingRegionFactory
+
+# Hibernate Search properties:
+## Connection info: see POM
+hibernate.search.backend.type=lucene
+hibernate.search.automatic_indexing.synchronization.strategy = sync
+hibernate.search.backend.log.json_pretty_printing = true
```
yrodiere, line 15: This property only works with the Elasticsearch backend.

```suggestion
```

integrationtest/java/modules/pojo-standalone-elasticsearch/src/main/java/org/hibernate/search/integrationtest/java/modules/pojo/standalone/elasticsearch/service/AuthorService.java
```diff
@@ -0,0 +1,72 @@
...
+	private final Map<Integer, Author> datastore = new HashMap<>();
+	private final CloseableSearchMapping mapping;
+
+	public AuthorService() {
+		mapping = createSearchMapping();
+	}
+
+	private CloseableSearchMapping createSearchMapping() {
+		return SearchMapping.builder()
+				.property( "hibernate.search.automatic_indexing.synchronization.strategy", "sync" )
```
yrodiere, line 33: This property only works for the ORM mapper at the moment; see https://docs.jboss.org/hibernate/search/6.2/reference/en-US/html_single/#_automatic_indexing

```suggestion
```

As you noticed, you need to use `.refreshStrategy( DocumentRefreshStrategy.FORCE )` instead.

integrationtest/java/modules/pojo-standalone-lucene/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+        <relativePath>../../../pom.xml</relativePath>
+    </parent>
+    <artifactId>hibernate-search-integrationtest-java-modules-pojo-standalone-lucene</artifactId>
+    <packaging>jar</packaging>
+
+    <name>Hibernate Search ITs - Java - Modules - POJO/Lucene</name>
+    <description>Hibernate Search integration tests for Java 11+ modules</description>
+
+    <properties>
+        <test.elasticsearch.run.skip>true</test.elasticsearch.run.skip>
```
yrodiere, line 20:
```suggestion
```

integrationtest/java/modules/pojo-standalone-lucene/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+        <test.database.run.skip>false</test.database.run.skip>
+
+        <maven.compiler.release>11</maven.compiler.release>
+        <!-- Override the default from the parent POM: here we really do want to use the modulepath -->
+        <failsafe.useModulePath>true</failsafe.useModulePath>
+
+        <surefire.jvm.args.module>
+            ${test.elasticsearch.connection.jvm.args}
+        </surefire.jvm.args.module>
```
yrodiere, line 29:
```suggestion
```

integrationtest/java/modules/pojo-standalone-lucene/src/main/java/org/hibernate/search/integrationtest/java/modules/pojo/standalone/lucene/service/AuthorService.java
```diff
@@ -0,0 +1,75 @@
...
+	private final CloseableSearchMapping mapping;
+
+	public AuthorService() {
+		mapping = createSearchMapping();
+	}
+
+	private CloseableSearchMapping createSearchMapping() {
+		return SearchMapping.builder()
+				.property( "hibernate.search.automatic_indexing.synchronization.strategy", "sync" )
+				.property( "hibernate.search.backend.log.json_pretty_printing", true )
```
yrodiere, line 34:
```suggestion
```

integrationtest/java/modules/pojo-standalone-elasticsearch/src/main/java/org/hibernate/search/integrationtest/java/modules/pojo/standalone/elasticsearch/service/AuthorService.java
```diff
@@ -0,0 +1,72 @@
...
+				)
+				.build();
+	}
+
+	public void add(String name) {
+		try ( SearchSession session = mapping.createSessionWithOptions()
+				.refreshStrategy( DocumentRefreshStrategy.FORCE )
+				.build() ) {
+
+			Author author = new Author( new Random().nextInt(), name );
```
yrodiere, line 51: Better hardcode the ID (e.g. `1`); I wouldn't be surprised if some databases failed for negative IDs, or very large IDs.

integrationtest/java/modules/pojo-standalone-lucene/src/main/java/org/hibernate/search/integrationtest/java/modules/pojo/standalone/lucene/service/AuthorService.java
```diff
@@ -0,0 +1,75 @@
...
+				)
+				.build();
+	}
+
+	public void add(String name) {
+		try ( SearchSession session = mapping.createSessionWithOptions()
+				.refreshStrategy( DocumentRefreshStrategy.FORCE )
+				.build() ) {
+
+			Author author = new Author( new Random().nextInt(), name );
```
yrodiere, line 54: Better hardcode the ID (e.g. `1`); I wouldn't be surprised if some databases failed for negative IDs, or very large IDs.

mapper/orm-coordination-outbox-polling/pom.xml
```diff
@@ -87,6 +87,32 @@
...
+                <artifactId>moditect-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>add-module-infos</id>
+                        <configuration>
+                            <module>
+                                <moduleInfo>
+                                    <exports>
+                                        org.hibernate.search.mapper.orm.coordination.outboxpolling.avro.generated.impl to org.apache.avro;
+                                        org.hibernate.search.mapper.orm.coordination.outboxpolling.logging.impl to org.jboss.logging;
```
yrodiere, line 102: Are you sure you need to export something to Avro if you already have an `opens` directive? If not, I'd remove this section entirely and rely on the configuration from the parent POM.

If you really need to add an export for Avro, can you please reuse the other exports from the parent POM (`parents/public/pom.xml`?). Let's not diverge unless absolutely necessary.

yrodiere (changes requested):
Thanks! My comments below.

integrationtest/java/modules/orm-lucene/src/main/java/module-info.java
marko-bekhta: this is so bizarre. I could've sworn that it was falling with all kinds of "missing Lucene classes" errors until I put that transitive in place even, as you are saying, it shouldn't really make a difference. I've cleaned it all now, and it works just fine 🤷🏻‍♂️ 🪄

integrationtest/java/modules/pojo-standalone-lucene/src/main/java/org/hibernate/search/integrationtest/java/modules/pojo/standalone/lucene/service/AuthorService.java
```diff
@@ -0,0 +1,74 @@
...
+import org.hibernate.search.engine.search.query.SearchQuery;
+import org.hibernate.search.integrationtest.java.modules.pojo.standalone.lucene.entity.Author;
+import org.hibernate.search.mapper.pojo.standalone.loading.SelectionLoadingStrategy;
+import org.hibernate.search.mapper.pojo.standalone.mapping.CloseableSearchMapping;
+import org.hibernate.search.mapper.pojo.standalone.mapping.SearchMapping;
+import org.hibernate.search.mapper.pojo.standalone.session.SearchSession;
+
+public class AuthorService implements AutoCloseable {
+
+	private static final AtomicInteger ID_PROVIDER = new AtomicInteger( 1 );
```
yrodiere, line 24: Good idea; while not strictly necessary in the context of these tests, it's indeed cleaner than hardcoding the ID as I was suggesting, due to how the tests are separate from the creation of entities (I overlooked that part).

yrodiere (approved):
LGTM, thanks! Let's wait for CI :)

marko-bekhta:
@yrodiere OK, will do!

PR #3272, opened 2022-10-19
HSEARCH-4703 backport to 6.1

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4703

Backport of HSEARCH-4703 to 6.1. If I got it correctly, 6.1 didn't yet have a standalone mapper - so that results in 2 standalone test modules not being backported. 

When I was running the orm6/jakarta tests locally, the build was failing with some unrelated Javadoc warning - so wanted to see if it'll work fine on CI 😃


yrodiere (approved):
> If I got it correctly, 6.1 didn't yet have a standalone mapper - so that results in 2 standalone test modules not being backported.

It kind of did, but with a different name, different package names, and not published on Maven Central. So you're right, no need to backport those modules.

> so wanted to see if it'll work fine on CI

Right, let's wait for CI and merge if it works.

marko-bekhta:
aaaaaaaand it failed too 🥲 looking into it 😃

PR #3273, opened 2022-10-19
HSEARCH-4669 add more context to constructor pojection exception

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4669

Needed to bring `PojoConstructorModel` into `PojoConstructorProjectionDefinition` so that we can print a nicely formatted constructor string with `PojoConstructorModelFormatter` 😃 

I wasn't sure if we'd want to have the entire trace of nested constructors or just the only one where the initial problem occurred - so I've added that code in the separate commit so that we can drop it if we want the entire thing.

### DIFF
--- a/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/search/definition/impl/PojoConstructorProjectionDefinition.java
+++ b/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/search/definition/impl/PojoConstructorProjectionDefinition.java
@@ -33,22 +33,28 @@ import org.hibernate.search.mapper.pojo.model.spi.PojoMethodParameterModel;
 import org.hibernate.search.mapper.pojo.model.spi.PojoRawTypeIdentifier;
 import org.hibernate.search.mapper.pojo.model.spi.PojoRawTypeModel;
 import org.hibernate.search.mapper.pojo.model.spi.PojoTypeModel;
+import org.hibernate.search.mapper.pojo.reporting.impl.PojoConstructorProjectionDefinitionMessages;
+import org.hibernate.search.util.common.SearchException;
 import org.hibernate.search.mapper.pojo.reporting.spi.PojoEventContexts;
 import org.hibernate.search.util.common.impl.ToStringTreeAppendable;
 import org.hibernate.search.util.common.impl.ToStringTreeBuilder;
+import org.hibernate.search.util.common.logging.impl.CommaSeparatedClassesFormatter;
 import org.hibernate.search.util.common.logging.impl.LoggerFactory;
 import org.hibernate.search.util.common.reflect.spi.ValueCreateHandle;
 
 public final class PojoConstructorProjectionDefinition<T>
 		implements CompositeProjectionDefinition<T>, ToStringTreeAppendable {
 	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
+	private static final PojoConstructorProjectionDefinitionMessages MESSAGES = PojoConstructorProjectionDefinitionMessages.INSTANCE;
 
+	private final ConstructorIdentifier constructor;
 	private final ValueCreateHandle<? extends T> valueCreateHandle;
 	private final List<InnerProjectionDefinition> innerDefinitions;
 
-	private PojoConstructorProjectionDefinition(ValueCreateHandle<? extends T> valueCreateHandle,
+	private PojoConstructorProjectionDefinition(PojoConstructorModel<T> constructor,
 			List<InnerProjectionDefinition> innerDefinitions) {
-		this.valueCreateHandle = valueCreateHandle;
+		this.constructor = new ConstructorIdentifier( constructor );
+		this.valueCreateHandle = constructor.handle();
 		this.innerDefinitions = innerDefinitions;
 	}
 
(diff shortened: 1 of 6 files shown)


mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/search/definition/impl/PojoConstructorProjectionDefinition.java
```diff
@@ -79,6 +79,11 @@ public CompositeProjectionValueStep<?, T> apply(SearchProjectionFactory<?, ?> pr
 			return initialStep.from( innerProjections ).asArray( valueCreateHandle );
 		}
 		catch (SearchException e) {
+			if ( e.getCause() instanceof SearchException ) {
```
yrodiere, line 82: It's a good idea to reduce verbosity.

However, this implementation is a bit dodgy, as it's theoretically possible for an exception where the context was not already added to have a cause that is a `SearchException`.

Maybe you should just use a dedicated subclass of `SearchException` when you add context, and test for that? A bit like we did in `org.hibernate.search.mapper.pojo.processing.impl.PojoIndexingProcessorPropertyNode#process`?

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/search/definition/impl/PojoConstructorProjectionDefinition.java
```diff
@@ -43,12 +44,14 @@
 		implements CompositeProjectionDefinition<T>, ToStringTreeAppendable {
 	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
 
+	private final PojoConstructorModel<T> constructor;
```
yrodiere, line 47: You couldn't possibly know it, but this is actually a big no-no, as `PojoConstructorModel` is intended to be a boot-time-only thing, and `PojoConstructorProjectionDefinition` is used at runtime.

Why, you ask? `PojoConstructorModel` may have references to objects (e.g. `AbstractPojoHCAnnBootstrapIntrospector`) that we want to be garbage-collected after boot, because:

1. They may be large (a map of all types encountered so far and all their metadata, ...)
2. They expose operations that may not work fine in a native image, so we need to make sure they are not present in the heap after static init in Quarkus, so that GraalVM will infer "that class cannot work properly, but it's okay because I don't see any instance nor any way to create a new instance when executing the native image".

I suppose we could rely on `valueCreateHandle` to print *something*, but then the message would not be very user-friendly...

Maybe you can pre-build some representation of that constructor that can be used at runtime, similarly to what we do with `PojoModelPath`/`PojoModelPathPropertyNode`. Of course I'm not asking you to build a path of all constructors used in a projection definition (though that could be a nice follow-up issue), but just having a single object representing a single constructor and its parameters at runtime could be nice.

Regarding the constraints: you should preferably use `PojoRawTypeIdentifier` to represent types (use `PojoRawTypeModel.typeIdentifier()` to get an identifier), and avoid any references (direct or indirect) to `PojoTypeModel`, `PojoRawTypeModel`, `PojoConstructorModel` et al.

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/ProjectionConstructorMappingExceptionIT.java
```diff
@@ -0,0 +1,189 @@
...
+import org.hibernate.search.mapper.pojo.standalone.session.SearchSession;
+import org.hibernate.search.util.common.SearchException;
+import org.hibernate.search.util.impl.integrationtest.common.rule.BackendMock;
+import org.hibernate.search.util.impl.integrationtest.common.rule.StubSearchWorkBehavior;
+import org.hibernate.search.util.impl.integrationtest.mapper.pojo.standalone.StandalonePojoMappingSetupHelper;
+
+import org.junit.Rule;
+import org.junit.Test;
+
+public class ProjectionConstructorMappingExceptionIT {
```
yrodiere, line 34: IMO it would be fine to put these tests in `ProjectionConstructorBaseIT`.

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/ProjectionConstructorMappingExceptionIT.java
```diff
@@ -0,0 +1,189 @@
...
+
+	@Rule
+	public BackendMock backendMock = new BackendMock();
+
+	@Rule
+	public StandalonePojoMappingSetupHelper setupHelper = StandalonePojoMappingSetupHelper.withBackendMock(
+			MethodHandles.lookup(), backendMock );
+
+	@Test
+	public void compatibleProjection() {
```
yrodiere, line 44: Doesn't this test duplicate `ProjectionConstructorBaseIT#inferredInner_object`?

yrodiere (changes requested):
Thanks! My comments (and answers to your questions, I think) below.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/search/definition/impl/PojoConstructorProjectionDefinitionMessages.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.jboss.logging.Messages;
+import org.jboss.logging.annotations.Message;
+import org.jboss.logging.annotations.MessageBundle;
+
+/**
+ * Message bundle for constructor projections in the POJO mapper.
+ */
+@MessageBundle(projectCode = MessageConstants.PROJECT_CODE)
+public interface PojoConstructorProjectionDefinitionMessages {
```
yrodiere, line 19: Better move this next to other message bundles in the `reporting.impl` package. Otherwise you'll get in trouble because of Java modules (we need such classes to be opened to the jboss-logging module, and we only do that for `reporting.impl` packages at the moment).

yrodiere (changes requested):
Thanks, LGTM except for the package of the message bundles. Now let's see if it works :)

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/search/definition/impl/PojoConstructorProjectionDefinitionMessages.java
marko-bekhta: Oh, right.. moved the messages and rebased to include the latest from the main branch 😃

yrodiere (approved):

PR #3275, opened 2022-10-20
HSEARCH-4723 Fix flakiness in AutomaticIndexingOneToManyListBaseIT caused by unpredictable collection order

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4723


marko-bekhta (approved):
looks good 😃

PR #3278, opened 2022-10-21
HSEARCH-4700  Configurable behavior on full queues for indexing operations

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4700

The last two commits are cleanups in tests and replacing usage of default value - so they are mostly just noise. 

I have looked if we can extend `ScheduledThreadPoolExecutor` to handle the `queue.add` operation differently, but it seems that to get to the behaviour we want, we'd need to copy a lot of code 😞 so you won't find that in this PR 😃 

I've looked at the possibility of adding a configuration of a default operation submitter. But I suppose the way I've updated the interfaces making those default versions deprecated it wouldn't be of much use to have such a property ... 

I haven't also touched the `PojoScopeSchemaManager` yet. I wanted to make sure that I'm going in the right direction with the current changes before updating that one 😃


engine/src/main/java/org/hibernate/search/engine/backend/work/execution/impl/OperationSubmitterType.java
```diff
@@ -0,0 +1,58 @@
...
+	BLOCKING {
+		@Override
+		public <T> void submitToQueue(Queue<T> queue, T element) {
+			if ( queue instanceof BlockingQueue ) {
+				try {
+					( (BlockingQueue<T>) queue ).put( element );
+				}
+				catch (InterruptedException e) {
+					throw log.failedToSubmitToQueue( element, e.getMessage(), e );
+				}
```
yrodiere, line 32: That's not how `InterruptedException` should be handled; you're not resetting the interrupt flag correctly.

Regardless, if you change the parameter type to `BlockingQueue`, it would make sense for the method to just let the `InterruptedException` go through and be caught by the caller if necessary.

engine/src/main/java/org/hibernate/search/engine/backend/work/execution/impl/OperationSubmitterType.java
```diff
@@ -0,0 +1,58 @@
...
+	REJECTING_EXECUTION_EXCEPTION {
+		@Override
+		public <T> void submitToQueue(Queue<T> queue, T element) {
+			try {
+				queue.add( element );
+			}
+			catch (IllegalStateException e) {
+				//means we had a blocking queue, and it's full:
+				throw log.cannotAcceptMoreWork( element, e.getMessage(), e );
+			}
```
yrodiere, line 52: Here too, if you change the parameter type to `BlockingQueue`, you'll have an easier time implementing the method.

The "right" way of trying to add an element to a blocking queue is to just call `offer()` and check the return value.

engine/src/main/java/org/hibernate/search/engine/logging/impl/Log.java
```diff
@@ -530,4 +531,11 @@ SearchException cannotCreateEntityProjection(String name, @FormatWith(ClassForma
 			value = "Unable to resolve field '%1$s': %2$s")
 	SearchException unableToResolveField(String absolutePath, String causeMessage, @Cause SearchException e,
 			@Param EventContext context);
+
+	@Message(id = ID_OFFSET + 116, value = "Failed to submit %1$s to work queue: %2$s")
+	SearchException failedToSubmitToQueue(Object work, String causeMessage, @Cause Exception e);
+
+	@Message(id = ID_OFFSET + 117, value = "Work queue is full. Cannot accept %1$s at the moment: %2$s")
+	WorkQueueFullException cannotAcceptMoreWork(Object work, String causeMessage, @Cause Exception e);
+
```
yrodiere, line 540: Please don't forget to remove these if you agree with my other suggestions.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/work/IndexWorkspacePurgeIT.java
```diff
@@ -24,27 +25,27 @@ protected void ensureOperationsFail(TckBackendAccessor accessor, String indexNam
 
 	@Override
 	protected CompletableFuture<?> executeAsync(IndexWorkspace workspace) {
-		return workspace.purge( Collections.emptySet() );
+		return workspace.purge( Collections.emptySet(), OperationSubmitterType.BLOCKING );
```
yrodiere, line 28: Shouldn't this be `REJECTED_EXECUTION_EXCEPTION`?

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/event/impl/OutboxPollingOutboxEventSendingPlan.java
```diff
@@ -54,7 +55,7 @@ public void discard() {
 
 	@Override
 	public <R> CompletableFuture<MultiEntityOperationExecutionReport<R>> sendAndReport(
-			EntityReferenceFactory<R> entityReferenceFactory) {
+			EntityReferenceFactory<R> entityReferenceFactory, OperationSubmitter ignore) {
```
yrodiere, line 58: Unless you have a good reason to think this is irrelevant, I would at least assert that the submitter is `BLOCKING` here (and through an `AssertionFailedException` if not).

SPI contracts are no joke, you can't just ignore them like that :)

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/automaticindexing/session/impl/ConfiguredAutomaticIndexingSynchronizationStrategy.java
```diff
@@ -45,7 +46,8 @@ public DocumentRefreshStrategy getDocumentRefreshStrategy() {
 
 	public void executeAndSynchronize(PojoIndexingPlan indexingPlan) {
 		CompletableFuture<SearchIndexingPlanExecutionReport> reportFuture =
-				indexingPlan.executeAndReport( entityReferenceFactory ).thenApply( SearchIndexingPlanExecutionReportImpl::new );
+				indexingPlan.executeAndReport( entityReferenceFactory, OperationSubmitterType.BLOCKING )
```
yrodiere, line 49: I believe the submitter should be configurable in `AutomaticIndexingSynchronizationConfigurationContext`, as we discussed on Zulip.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/logging/impl/Log.java
```diff
@@ -767,4 +767,5 @@ ConstructorProjectionApplicationException errorApplyingProjectionConstructor(
 			String causeMessage,
 			@Cause Exception cause,
 			@Param ProjectionConstructorPath path);
+
```
yrodiere, line 770:
```suggestion
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/work/impl/PojoIndexingQueueEventProcessingPlanImpl.java
```diff
@@ -48,7 +49,7 @@ public void append(String entityName, String serializedId, PojoIndexingQueueEven
 	@Override
 	public <R> CompletableFuture<MultiEntityOperationExecutionReport<R>> executeAndReport(
 			EntityReferenceFactory<R> entityReferenceFactory) {
-		return delegate.executeAndReport( entityReferenceFactory );
+		return delegate.executeAndReport( entityReferenceFactory, OperationSubmitterType.BLOCKING );
```
yrodiere, line 52: Shouldn't this be a parameter?

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/work/impl/PojoScopeWorkspaceImpl.java
```diff
@@ -28,23 +29,23 @@ public PojoScopeWorkspaceImpl(Set<? extends PojoWorkIndexedTypeContext<?, ?>> ta
 	}
 
 	@Override
-	public CompletableFuture<?> mergeSegments() {
-		return doOperationOnTypes( IndexWorkspace::mergeSegments );
+	public CompletableFuture<?> mergeSegments(OperationSubmitter operationSubmitter) {
+		return doOperationOnTypes( indexWorkspace -> indexWorkspace.mergeSegments( operationSubmitter ) );
```
yrodiere, line 33: I hear capturing lambdas have a performance impact, so it would probably be better to pass the submitter to `doOperationOnTypes`, and turn the `Function` to a `BiFunction`, so you can write `doOperationOnTypes( IndexWorkspace::mergeSegments, operationSubmitter )`.

engine/src/main/java/org/hibernate/search/engine/backend/work/execution/spi/IndexWorkspace.java
```diff
@@ -14,12 +14,32 @@
...
 public interface IndexWorkspace {
 
-	CompletableFuture<?> mergeSegments();
+	@Deprecated
+	default CompletableFuture<?> mergeSegments() {
+		return mergeSegments( OperationSubmitter.DEFAULT );
+	}
 
-	CompletableFuture<?> purge(Set<String> routingKeys);
+	CompletableFuture<?> mergeSegments(OperationSubmitter operationSubmitter);
```
yrodiere, line 22: I'm a bit surprised you decided to add this parameter to `IndexWorkspace` in the end, since during our last discussion we agreed it was not possible to implement support for `operationSubmitter` in this interface for Lucene. But ok, let's see where this leads us...

yrodiere (changes requested):
Thanks! A few comments below.

---

I'm a bit concerned about testing coverage. By design, this PR nearly doubles the paths code could take in the impacted files, and we're not testing all those paths. I realize we *cannot* test all those paths, that would be too much work, but maybe we should at least make an attempt?

There are two areas where testing is critical, IMO:

* `BatchingExecutor`. We should probably run all tests with both submitters (they shouldn't have any impact since queues are not full, but just in case), and if possible add a few tests regarding the behavior when the queue is full.
* `LuceneParallelOrchestrator`. That orchestrator cannot implement the "reject execution" behavior, so I suggested in my comments below to throw an exception when that submitter is passed to that orchestrator. But that's only possible if we take extra care to avoid ending up in that situation in our mappers. What worries me in particular is that we probably don't have end-to-end testing for all those operations, so it's possible that the ORM mapper (for instance) would pass an unsupported submitter, and our tests wouldn't fail because they use the stub backend.
  Alternatively, maybe this orchestrator should *not* throw an exception, and ignore the submitter as you did... But in that case we would need *some* way to inform the caller that we're not doing what they requested. Maybe a one-time warning, with a stack trace? I don't know... Let's discuss that on Zulip, maybe?

---

Also, it looks like the build is failing, could you have a look please?

```
[ERROR] Constraint: hsearch:TypesMayNotDependOnImplementationTypeFromOtherModules

[ERROR] Severity: MAJOR

[ERROR] Number of rows: 15

[ERROR]             Main (non-test) types must not depend on implementation types from other modules.

[ERROR]             SPIs must be used for such dependencies.

[ERROR]             Exceptions are allowed only when the dependency type is in a util module,

[ERROR]             or the depending type is a test type.

[ERROR]         

[ERROR]   artifact1=org.hibernate.search:hibernate-search-mapper-orm-batch-jsr352-core:jar:6.2.0-SNAPSHOT, type1=org.hibernate.search.batch.jsr352.core.massindexing.step.impl.EntityWriter, artifact2=org.hibernate.search:hibernate-search-engine:jar:6.2.0-SNAPSHOT, type2=org.hibernate.search.engine.backend.work.execution.impl.OperationSubmitterType

[ERROR]   artifact1=org.hibernate.search:hibernate-search-mapper-orm:jar:6.2.0-SNAPSHOT, type1=org.hibernate.search.mapper.orm.automaticindexing.session.impl.ConfiguredAutomaticIndexingSynchronizationStrategy, artifact2=org.hibernate.search:hibernate-search-engine:jar:6.2.0-SNAPSHOT, type2=org.hibernate.search.engine.backend.work.execution.impl.OperationSubmitterType

[ERROR]   artifact1=org.hibernate.search:hibernate-search-mapper-orm:jar:6.2.0-SNAPSHOT, type1=org.hibernate.search.mapper.orm.work.impl.SearchWorkspaceImpl, artifact2=org.hibernate.search:hibernate-search-engine:jar:6.2.0-SNAPSHOT, type2=org.hibernate.search.engine.backend.work.execution.impl.OperationSubmitterType

[ERROR]   artifact1=org.hibernate.search:hibernate-search-mapper-pojo-standalone:jar:6.2.0-SNAPSHOT, type1=org.hibernate.search.mapper.pojo.standalone.work.impl.SearchIndexerImpl, artifact2=org.hibernate.search:hibernate-search-engine:jar:6.2.0-SNAPSHOT, type2=org.hibernate.search.engine.backend.work.execution.impl.OperationSubmitterType

[ERROR]   artifact1=org.hibernate.search:hibernate-search-mapper-pojo-standalone:jar:6.2.0-SNAPSHOT, type1=org.hibernate.search.mapper.pojo.standalone.work.impl.SearchIndexingPlanImpl, artifact2=org.hibernate.search:hibernate-search-engine:jar:6.2.0-SNAPSHOT, type2=org.hibernate.search.engine.backend.work.execution.impl.OperationSubmitterType

[ERROR]   artifact1=org.hibernate.search:hibernate-search-mapper-pojo-standalone:jar:6.2.0-SNAPSHOT, type1=org.hibernate.search.mapper.pojo.standalone.work.impl.SearchWorkspaceImpl, artifact2=org.hibernate.search:hibernate-search-engine:jar:6.2.0-SNAPSHOT, type2=org.hibernate.search.engine.backend.work.execution.impl.OperationSubmitterType

[ERROR]   artifact1=org.hibernate.search:hibernate-search-mapper-pojo-base:jar:6.2.0-SNAPSHOT, type1=org.hibernate.search.mapper.pojo.work.impl.PojoIndexingQueueEventProcessingPlanImpl, artifact2=org.hibernate.search:hibernate-search-engine:jar:6.2.0-SNAPSHOT, type2=org.hibernate.search.engine.backend.work.execution.impl.OperationSubmitterType

[ERROR]   artifact1=org.hibernate.search:hibernate-search-mapper-pojo-base:jar:6.2.0-SNAPSHOT, type1=org.hibernate.search.mapper.pojo.massindexing.impl.PojoMassIndexingEntityLoadingRunnable$IndexingBatch, artifact2=org.hibernate.search:hibernate-search-engine:jar:6.2.0-SNAPSHOT, type2=org.hibernate.search.engine.backend.work.execution.impl.OperationSubmitterType

[ERROR]   artifact1=org.hibernate.search:hibernate-search-mapper-pojo-base:jar:6.2.0-SNAPSHOT, type1=org.hibernate.search.mapper.pojo.massindexing.impl.PojoMassIndexingBatchCoordinator, artifact2=org.hibernate.search:hibernate-search-engine:jar:6.2.0-SNAPSHOT, type2=org.hibernate.search.engine.backend.work.execution.impl.OperationSubmitterType

[ERROR]   artifact1=org.hibernate.search:hibernate-search-backend-elasticsearch:jar:6.2.0-SNAPSHOT, type1=org.hibernate.search.backend.elasticsearch.search.query.impl.ElasticsearchSearchScrollImpl, artifact2=org.hibernate.search:hibernate-search-engine:jar:6.2.0-SNAPSHOT, type2=org.hibernate.search.engine.backend.work.execution.impl.OperationSubmitterType

[ERROR]   artifact1=org.hibernate.search:hibernate-search-backend-lucene:jar:6.2.0-SNAPSHOT, type1=org.hibernate.search.backend.lucene.search.query.impl.LuceneSearchScrollImpl, artifact2=org.hibernate.search:hibernate-search-engine:jar:6.2.0-SNAPSHOT, type2=org.hibernate.search.engine.backend.work.execution.impl.OperationSubmitterType

[ERROR]   artifact1=org.hibernate.search:hibernate-search-backend-lucene:jar:6.2.0-SNAPSHOT, type1=org.hibernate.search.backend.lucene.schema.management.impl.LuceneIndexSchemaManager, artifact2=org.hibernate.search:hibernate-search-engine:jar:6.2.0-SNAPSHOT, type2=org.hibernate.search.engine.backend.work.execution.impl.OperationSubmitterType

[ERROR]   artifact1=org.hibernate.search:hibernate-search-backend-lucene:jar:6.2.0-SNAPSHOT, type1=org.hibernate.search.backend.lucene.search.query.impl.LuceneSearchQueryImpl, artifact2=org.hibernate.search:hibernate-search-engine:jar:6.2.0-SNAPSHOT, type2=org.hibernate.search.engine.backend.work.execution.impl.OperationSubmitterType

[ERROR]   artifact1=org.hibernate.search:hibernate-search-backend-elasticsearch:jar:6.2.0-SNAPSHOT, type1=org.hibernate.search.backend.elasticsearch.search.query.impl.ElasticsearchSearchQueryImpl, artifact2=org.hibernate.search:hibernate-search-engine:jar:6.2.0-SNAPSHOT, type2=org.hibernate.search.engine.backend.work.execution.impl.OperationSubmitterType

[ERROR]   artifact1=org.hibernate.search:hibernate-search-backend-elasticsearch:jar:6.2.0-SNAPSHOT, type1=org.hibernate.search.backend.elasticsearch.schema.management.impl.ElasticsearchSchemaAccessor, artifact2=org.hibernate.search:hibernate-search-engine:jar:6.2.0-SNAPSHOT, type2=org.hibernate.search.engine.backend.work.execution.impl.OperationSubmitterType
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/work/impl/PojoScopeWorkspaceImpl.java
marko-bekhta: Yeah. This was primarily because of the `#purge()` that was already using lambda to address the routing keys. changed 😃

engine/src/main/java/org/hibernate/search/engine/backend/work/execution/spi/OperationSubmitter.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Queue;
+
+import org.hibernate.search.engine.backend.work.execution.impl.OperationSubmitterType;
+
+
+/**
+ * Interface defining how operation should be submitted to the queue.
+ */
+public interface OperationSubmitter {
+	OperationSubmitter DEFAULT = OperationSubmitterType.BLOCKING;
```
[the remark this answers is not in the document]
marko-bekhta: I've started with this `DEFAULT` keeping the config property in mind - so that it would be easier to find where we've ended up using this default and should make use of the config property. But as the property is not happening ... Removed `DEFAULT` 😃

marko-bekhta:
I'm not sure what should be done about `AutomaticIndexingSynchronizationConfigurationContext#operationSubmitter` as it is one of the places where the build fails as this method exposes an SPI submitter type on the API level... 

The same applies to `COORDINATION_EVENT_PROCESSOR_OPERATION_SUBMITTER` config property. I've added that one so we can eventually pass it to the `AutomaticIndexingQueueEventProcessingPlan` plans `processingPlan.executeAndReport( operationSubmitter )`.

The third reported issue for `OPERATION_SUBMITTER` - I've removed that one already. Right now, I can't remember why I've added it in the first place - so running tests locally and looking for a reason I added it in the first place 😄

yrodiere:
Thanks.

> I'm not sure what should be done about AutomaticIndexingSynchronizationConfigurationContext#operationSubmitter as it is one of the places where the build fails as this method exposes an SPI submitter type on the API level...

I don't think we have a choice; we'll have to promote the submitter type to API... Please just make sure to mark it as `@Incubating`, so that we can change it in the next few versions if necessary.

> The same applies to COORDINATION_EVENT_PROCESSOR_OPERATION_SUBMITTER config property. I've added that one so we can eventually pass it to the AutomaticIndexingQueueEventProcessingPlan plans processingPlan.executeAndReport( operationSubmitter ).

But configuration properties are for users. Why should this be configured by the user?

As far as I'm concerned, it's fine to hardcode the submitter to `BLOCKING` in `org.hibernate.search.mapper.orm.coordination.outboxpolling.event.impl.OutboxEventProcessingPlan#processEvents`. After all, as you rightly pointed out, this eventually does with the database so we already block.

If this seems to you like I'm changing my mind, remember that we need to clearly separate two concerns:

* The SPI, which needs to be flexible.
* The implementation (coordination-outbox-polling), which is free to set its own expectations.

marko-bekhta:
OK 😃 makes sense. Let me make the adjustments.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/schema/management/ElasticsearchIndexSchemaManagerOperation.java
```diff
@@ -22,42 +23,42 @@ enum ElasticsearchIndexSchemaManagerOperation {
...
 		@Override
-		public CompletableFuture<?> apply(IndexSchemaManager schemaManager) {
-			return schemaManager.validate( new StubUnusedContextualFailureCollector() );
+		public CompletableFuture<?> apply(IndexSchemaManager schemaManager, OperationSubmitter operationSubmitter) {
+			return schemaManager.validate( new StubUnusedContextualFailureCollector(), operationSubmitter );
 		}
 	};
 
-	public abstract CompletableFuture<?> apply(IndexSchemaManager schemaManager);
+	public abstract CompletableFuture<?> apply(IndexSchemaManager schemaManager, OperationSubmitter operationSubmitter);
```
yrodiere, line 61: As far as I can see, you always pass `BLOCKING` to this method. And it's in test code. So... Maybe just hardcode `BLOCKING` in the implementations, and don't bother changing a million tests? :)

I mean if you want to test both submitters in some tests, then go ahead, but I'm not sure that would be a great use of your time :)

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/schema/management/ElasticsearchIndexSchemaManagerValidationOperation.java
```diff
@@ -51,7 +54,8 @@ public final CompletableFuture<?> apply(IndexSchemaManager schemaManager) {
 				} ) );
 	}
 
-	protected abstract CompletableFuture<?> apply(IndexSchemaManager schemaManager, ContextualFailureCollector failureCollector);
+	protected abstract CompletableFuture<?> apply(IndexSchemaManager schemaManager, ContextualFailureCollector failureCollector,
+			OperationSubmitter operationSubmitter);
```
yrodiere, line 58: Same here, no need to add a parameter if you're only ever be testing the blocking submitter.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/work/AbstractIndexWorkspaceSimpleOperationIT.java
```diff
@@ -104,7 +105,11 @@ protected void afterInitData(StubMappedIndex index) {
 	protected abstract void assertSuccess(StubMappedIndex index);
 
 	private void setup() {
-		setupHelper.start().withIndex( index ).setup();
+		setupHelper.start()
+				// so that lucene backend won't fail on async calls using OperationSubmitter.REJECTED_EXECUTION_EXCEPTION
+				.withProperty( EngineSpiSettings.IS_WRITE_EXECUTOR_BLOCKING, false )
```
yrodiere, line 110: That's cheating :) But as I mentioned elsewhere, we don't need this property.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/work/IndexWorkspacePurgeIT.java
```diff
@@ -24,27 +25,27 @@ protected void ensureOperationsFail(TckBackendAccessor accessor, String indexNam
...
-		return workspace.purge( Collections.emptySet() );
+		return workspace.purge( Collections.emptySet(), OperationSubmitter.REJECTED_EXECUTION_EXCEPTION );
 	}
 
 	@Override
 	protected void afterInitData(StubMappedIndex index) {
 		// Make sure to flush the index, otherwise the test won't fail as expected with Lucene,
 		// probably because the index writer optimizes purges when changes are not committed yet.
-		index.createWorkspace().flush();
+		index.createWorkspace().flush( OperationSubmitter.BLOCKING );
```
yrodiere, line 35: Not your fault, but interestingly, there's a missing `.join()` here...

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/session/AutomaticIndexingSynchronizationStrategyIT.java
```diff
@@ -45,12 +47,15 @@
...
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
 
 import org.apache.logging.log4j.Level;
 import org.assertj.core.api.InstanceOfAssertFactories;
 import org.awaitility.Awaitility;
 import org.hamcrest.CoreMatchers;
 
+@RunWith(Parameterized.class)
```
yrodiere, line 58: Wouldn't it make more sense to just duplicate the method `success_custom` to test both submitters? Otherwise you're going to run every single method twice even though only a few of them are affected by your test parameter.

I wouldn't bother duplicating the other test methods, btw.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/automaticindexing/spi/AutomaticIndexingQueueEventSendingPlan.java
```diff
@@ -45,11 +46,22 @@ public interface AutomaticIndexingQueueEventSendingPlan {
...
+
+	/**
+	 * @see #sendAndReport(EntityReferenceFactory, OperationSubmitter)
+	 * @deprecated Use {@link #sendAndReport(EntityReferenceFactory, OperationSubmitter)} instead.
+	 */
+	@Deprecated
+	default <R> CompletableFuture<MultiEntityOperationExecutionReport<R>> sendAndReport(
+			EntityReferenceFactory<R> entityReferenceFactory) {
+		return sendAndReport( entityReferenceFactory, OperationSubmitter.BLOCKING );
+	}
```
yrodiere, line 65: Same comment as elsewhere: this won't preserve backwards compatibility, as this interface is supposed to be *implemented* by service providers.

You'll have to introduce *two* default methods to preserve backwards compatibility (see my other comment about this).

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/mapping/impl/HibernateOrmMapping.java
```diff
@@ -182,22 +183,22 @@ public CompletableFuture<?> start(MappingStartContext context) {
 
 		// Schema management
 		PojoScopeSchemaManager schemaManager = scope.schemaManagerDelegate();
-		return schemaManagementListener.onStart( context, schemaManager )
+		return schemaManagementListener.onStart( context, schemaManager, operationSubmitter )
```
yrodiere, line 186: It's really okay to block if queues are full on startup/shutdown, IMO. Who would want to fail if a queue gets full on startup? I'd expect everyone to want to wait until startup finishes.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/work/impl/SearchWorkspaceImpl.java
```diff
@@ -23,12 +24,12 @@ public SearchWorkspaceImpl(PojoScopeWorkspace delegate) {
...
 	@Override
 	public void mergeSegments() {
-		Futures.unwrappedExceptionJoin( mergeSegmentsAsync() );
+		Futures.unwrappedExceptionJoin( delegate.mergeSegments( OperationSubmitter.BLOCKING ) );
 	}
 
 	@Override
 	public CompletableFuture<?> mergeSegmentsAsync() {
-		return delegate.mergeSegments();
+		return delegate.mergeSegments( OperationSubmitter.REJECTED_EXECUTION_EXCEPTION );
```
yrodiere, line 32: :+1: 

Could you add something to the "behavior" section of the migration guide, though? This is technically a breaking change. 

No need to list every single method, but something along the lines of "`*Async` methods will no longer block if internal queues of operations are full, but will throw `RejectedOperationException` instead."

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/event/impl/OutboxPollingOutboxEventSendingPlan.java
yrodiere: I think you forgot to throw an exception?

There are two possibilities:

1. either this method is always non-blocking (that's what I meant by "Unless you have a good reason to think this is irrelevant") and you don't need to assert anything (not even to log a warning): the method will always be compatible with either submitter.
2. or this method is always blocking and you need to fail hard when the assertion fails, because we're not complying with the requested submitter.

As far as I can tell we're calling `persist` at some point, so this method is most likely blocking. So I would fail hard.

engine/src/main/java/org/hibernate/search/engine/cfg/spi/EngineSpiSettings.java
```diff
@@ -49,6 +50,17 @@ private EngineSpiSettings() {
...
+	/**
+	 * The boolean flag that defines if Lucene executor service will accept nonblocking {@link OperationSubmitter#REJECTED_EXECUTION_EXCEPTION}
+	 * operation submitter. If set to {@code true} and Lucene executor service receives {@link OperationSubmitter#REJECTED_EXECUTION_EXCEPTION}
+	 * operation submitter then an exception will be thrown, otherwise a submit operation will proceed.
+	 * <p>
+	 * Expects a boolean {@code true/false}.
+	 * <p>
+	 * Defaults to {@link Defaults#IS_WRITE_EXECUTOR_BLOCKING}, {@code true}.
+	 */
+	public static final String IS_WRITE_EXECUTOR_BLOCKING = PREFIX + Radicals.IS_WRITE_EXECUTOR_BLOCKING;
```
yrodiere, line 62: Please don't forget to remove changes to this file when your remove the configuration property.

yrodiere (changes requested):
Well, that's a lot of code :)

I added many comments, but we're making progress. So, thank you for your persistence!

I think there's mainly a confusion about:

* *what* should remain backwards compatible
* *how* to make an SPI backwards compatible, in particular when an interface is supposed to be implemented by service providers, not called.

I think I addressed this in my comments, but let's chat on Zulip if I'm not clear.

engine/src/test/java/org/hibernate/search/engine/backend/orchestration/spi/BatchingExecutorTest.java
```diff
@@ -315,6 +336,42 @@ public void endBatchFailure() throws InterruptedException {
...
 	}
 
+	@Test
+	public void simple_newTasksBlocked() throws InterruptedException {
+		createAndStartExecutor( 2, true );
+
+		assumeTrue(
+				"This test only makes sense for nonblocking submitter",
+				OperationSubmitter.REJECTED_EXECUTION_EXCEPTION.equals( operationSubmitter )
+		);
```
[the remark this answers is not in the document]
marko-bekhta: I've added one more test there for a blocking case. The trouble in combining them was that if we have a blocking submitter, then `executor.submit( work3Mock, operationSubmitter );` would block and never pass as `Runnable unblockExecutorSwitch = blockExecutor();` this switch will never run.... But it's good that you've pointed to this test as I've found it was missing a mock for `processorMock.endBatch()` and as a result, it was generating an unexpected exception to be handled by the processor.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/index/impl/LuceneIndexManagerImpl.java
```diff
@@ -207,7 +208,7 @@ public long computeSizeInBytes() {
 
 	@Override
 	public CompletableFuture<Long> computeSizeInBytesAsync() {
-		return schemaManager.computeSizeInBytes();
+		return schemaManager.computeSizeInBytes( OperationSubmitter.REJECTED_EXECUTION_EXCEPTION );
 	}
```
yrodiere, line 213: That's great, thanks.

Now, I'm sorry I took all that time to realize, but it just occurred to me that we're hardcoding the submitter in a *backend*... which might be a problem for Infinispan. That's fine for APIs, though.

So... could you please add a new method `CompletableFuture<Long> computeSizeInBytesAsync(OperationSubmitter)` to `LuceneIndexManager`, so that Infinispan can pick the submitter at will? Never too careful with these things...

If you adapt `computeSizeInBytesAsync()` (no-arg) to call you new method, you probably don't need to add a test (it'll be covered by existing tests).

yrodiere (changes requested):
Thanks. Still a few remaining issues... Just a few though :)

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/spi/MappingImplementor.java
```diff
@@ -36,9 +38,10 @@
 	 * executing any operations that needs to be executed before shutdown.
 	 *
 	 * @param context The pre-stop context.
+	 * @param operationSubmitter How to handle request to submit operation when the queue is full.
 	 * @return A future that completes when pre-stop operations complete.
 	 */
-	CompletableFuture<?> preStop(MappingPreStopContext context);
+	CompletableFuture<?> preStop(MappingPreStopContext context, OperationSubmitter operationSubmitter);
```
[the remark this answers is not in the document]
marko-bekhta: yeah... I think you haven't said about it explicitly 😃 and when I was looking at this `prestop` I wondered if maybe there's a use case when someone would just want to "kill the process" no matter what. anyways - changes applied 😃

engine/src/test/java/org/hibernate/search/engine/backend/orchestration/spi/BatchingExecutorTest.java
```diff
@@ -315,6 +336,92 @@ public void endBatchFailure() throws InterruptedException {
...
+				executor.submit( work3Mock, operationSubmitter );
+			}
+			catch (InterruptedException e) {
+				Thread.currentThread().interrupt();
+			}
+			return true;
+		} );
+
+		//queue is full so future won't complete.
+		assertThat( future.isDone() ).isFalse();
```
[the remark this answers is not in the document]
marko-bekhta: There's the same assert in  `OperationSubmitterTest` - so I've added `TimeUnit.SECONDS.sleep( 2 );` in both cases.

marko-bekhta:
I was looking through the changes and noticed that there are still a few places where a deprecated default is available (e.g. `PojoIndexer`), and I was wondering - should I go through all these SPI interfaces and make sure none of them has such methods added?

yrodiere:
> I was looking through the changes and noticed that there are still a few places where a deprecated default is available (e.g. `PojoIndexer`)

Yes, and that's great: we need to preserve backwards compatibility in that SPI.

To be clear:

* Backend SPI: only we implement that, and (for the most part) only we call that. It's an SPI because there are multiple implementations (Lucene and Elasticsearch) but I don't think anyone else implements it.
* `hibernate-search-mapper-pojo-base` SPIs: we implement that in the ORM and POJO Standalone backend, and so does Infinispan. So we need to preserve backwards compatibility.

> I was wondering - should I go through all these SPI interfaces and make sure none of them has such methods added?

That might be a good idea. You should be able to use the Sonar report for that, since all deprecated methods trigger a code smell.

But really, what's important is to preserve backwards compatibility. I guess if there are leftovers that were intended to preserve backwards compatibilty, but don't work properly and/or were not necessary, it's not such a big deal.

Personally, when I'm unsure about backwards compatibility, I generally use a [custom tool of mine](https://github.com/yrodiere/shell-tools/blob/main/src) to detect all API/SPI changes (running something like `src -d -f spi main..HEAD` -- `-d` = diff, `-f spi` = filter including SPIs only). Alternatively, you might be able to use `japicmp-maven-plugin`, but I think it will require a bit more configuration in Hibernate Search.

yrodiere:
Also, in case you haven't already, this might be a good read: https://hibernate.org/community/compatibility-policy/

yrodiere (approved):
LGTM, merging, thanks!

PR #3282, opened 2022-10-24
HSEARCH-4727 + HSEARCH-4732 OutboxEvent "payload" column created as blob(255) with DB2

yrodiere:
* [HSEARCH-4727](https://hibernate.atlassian.net/browse/HSEARCH-4727) OutboxEvent "payload" column created as blob(255) with DB2
* [HSEARCH-4732](https://hibernate.atlassian.net/browse/HSEARCH-4732) Run tests against DB2 and DB2 JDBC driver version 11.5.8.0


marko-bekhta (approved):

yrodiere:
I just updated the version of DB2 and its JDBC driver in the hope that this will workaround the (apparently client-side) deadlock we've experienced in the last build...

PR #3285, opened 2022-10-25
HSEARCH-4655 Add more links to the "Further reading" section of the documentation

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4655


marko-bekhta (approved):
nice 😃

PR #3287, opened 2022-10-26
[5.11] HSEARCH-4717 Upgrade to Byteman 4.0.20

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4717


marko-bekhta (approved):
that seems like haven't caused any regressions 😃  nice!

PR #3295, opened 2022-10-31
HSEARCH-4702 Move build-related modules to the build directory

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4702


backend/elasticsearch-aws/pom.xml
```diff
@@ -12,7 +12,7 @@
         <groupId>org.hibernate.search</groupId>
         <artifactId>hibernate-search-parent-public</artifactId>
         <version>6.2.0-SNAPSHOT</version>
-        <relativePath>../../parents/public</relativePath>
+        <relativePath>../../build/parents/public</relativePath>
```
marko-bekhta, line 15: I've noticed that some of these relative paths were pointing to a folder (like this one) while others (e.g. `        <relativePath>../../../build/parents/integrationtest/pom.xml</relativePath>`) are referring to a pom. Maybe it's worth making these more consistent?

marko-bekhta (approved):
Looks good! I've added one inline question 😃

backend/elasticsearch-aws/pom.xml
yrodiere: Right. I added a commit to do just that. Let's hope that consistency is there to stay... :)

PR #3297, opened 2022-11-02
HSEARCH-4739 Upgrade to jackson-databind 2.13.4.2

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4739

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -229,11 +229,11 @@
         <version.com.fasterxml.jackson>2.13.4</version.com.fasterxml.jackson>
         <!--
              In the past, there has been jackson-databind bugfix releases without any jackson-core release,
-             so we had to use different versions, e.g. jackson-core 2.9.9 and jackson-databind 2.9.9.3.
+             so we had to use different versions, e.g. jackson-core 2.13.4 and jackson-databind 2.13.4.2.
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -229,11 +229,11 @@
         <version.com.fasterxml.jackson>2.13.4</version.com.fasterxml.jackson>
         <!--
              In the past, there has been jackson-databind bugfix releases without any jackson-core release,
-             so we had to use different versions, e.g. jackson-core 2.9.9 and jackson-databind 2.9.9.3.
+             so we had to use different versions, e.g. jackson-core 2.13.4 and jackson-databind 2.13.4.1.
```
marko-bekhta, line 232: It seems that there's a `2.13.4.2`, or is this version aligned with some other dependencies?

yrodiere: You're right. I was targeting 2.13.4.1 because of a CVE, but it seems they released another version since then. Probably because of an incomplete fix... I'll update my PR.

yrodiere: Done, thanks for the heads-up!

marko-bekhta: Ohh those CVEs 🙈

PR #3299, opened 2022-11-02
HSEARCH-4615  Move more configuration of the Standalone POJO mapper to configuration properties

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4615

I haven't found any other places except the documentation where we do some testing of standalone-specific things, so I've added a new test for configurer there.
As for the multi-tenancy property - it lives in the builder, as my understanding was that we'd want to let the user override it through the "setter". If we'd want to do it in the initiator - then we'd need to pass this property/property source to the `StandalonePojoMappingInitiator` constructor, as at the time of the call to `StandalonePojoMappingInitiator#configure`, it'll be too late...

Probably not directly related to this PR, but it seems that out of the box, `SearchMappingBuilder` won't read properties from a file. I was wondering if that's intentional or if it's something we should add.


yrodiere:
> I haven't found any other places except the documentation where we do some testing of standalone-specific things, so I've added a new test for configurer there.

If the tests contain code referenced in the documentation, that's fine. If they don't, they shouldn't be there.

Please either move the tests to `integrationtest/mapper/pojo-base`, or, if you really feel like it's necessary, create a dedicated directory `integrationtest/mapper/pojo-standalone`.

>  As for the multi-tenancy property - it lives in the builder, as my understanding was that we'd want to let the user override it through the "setter". If we'd want to do it in the initiator - then we'd need to pass this property/property source to the StandalonePojoMappingInitiator constructor, as at the time of the call to StandalonePojoMappingInitiator#configure, it'll be too late...

You can access configuration through `buildContext.configurationPropertySource()` in `org.hibernate.search.mapper.pojo.mapping.spi.AbstractPojoMappingInitiator#createMapper`?

> Probably not directly related to this PR, but it seems that out of the box, SearchMappingBuilder won't read properties from a file. I was wondering if that's intentional or if it's something we should add.

It would be nice to expose something like that in the builder, indeed. Maybe open another issue and try to come up with an API in the comments on that issue? I'm thinking `.properties(ConfigurationPropertySource)` would make sense, but I'm open to suggestions.

marko-bekhta:
> You can access configuration through buildContext.configurationPropertySource() in org.hibernate.search.mapper.pojo.mapping.spi.AbstractPojoMappingInitiator#createMapper?

Wouldn't that be too late to get that property?  In a scenario when the user sets a `MULTI_TENANCY` property and then explicitly calls the builder method - we'd override the value user set in the builder with the one from the property... which I thought wasn't what we wanted... Or should we set `AbstractPojoMappingInitiator#tenancyMode == null` by default and then check if it was set by the user, if not - check the property, if that one is empty - set the default `SINGLE_TENANCY`?

Thinking more about it ... It seems that the safest option would be to remove the multitenancy from the builder (`SearchMappingBuilder`) methods. We can leave it in the configurer and as a property and then do all the init work in the `StandalonePojoMappingInitiator#configure()`. Because we can set the `MULTI_TENANCY` property through the builder and call the builder method at the same time, and it creates confusion 😃

yrodiere:
> It seems that the safest option would be to remove the multitenancy from the builder (SearchMappingBuilder) methods

That's what I had in mind, yes. Feel free to remove all setters that get in the way; this module was not published until 6.2.0.Alpha1 so there is no backwards compatibility concern.

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/mapping/SearchMappingBuilder.java
```diff
@@ -172,19 +171,6 @@ public <E> SearchMappingBuilder addEntityType(Class<E> type, String entityName,
...
 
-	/**
-	 * Enables or disables multi-tenancy.
-	 * <p>
-	 * If multi-tenancy is enabled, every {@link SearchMapping#createSession() session} will need to be assigned a tenant identifier.
-	 *
-	 * @param multiTenancyEnabled {@code true} to enable multi-tenancy, {@code false} to disable it (the default).
-	 * @return {@code this}, for call chaining.
-	 */
-	public SearchMappingBuilder multiTenancyEnabled(boolean multiTenancyEnabled) {
```
yrodiere, line 183: Can you also remove the other setters that are now in the configuration context? No need to duplicate all that.

yrodiere (commented):
Thanks. My comments below.

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/cfg/StandalonePojoMapperSettings.java
```diff
@@ -29,12 +31,38 @@ private StandalonePojoMapperSettings() {
...
+
+
 	public static class Radicals {
 
 		private Radicals() {
 		}
 
 		public static final String SCHEMA_MANAGEMENT_STRATEGY = "schema_management.strategy";
+		public static final String MAPPING_CONFIGURER = "mapping.standalone.configurer";
+		public static final String MULTI_TENANCY = "mapping.standalone.multi_tenancy";
```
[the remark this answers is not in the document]
marko-bekhta: yeah ... my main concern was that we have this in ORM mapper - e.g. https://github.com/hibernate/hibernate-search/blob/main/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/cfg/HibernateOrmMapperSettings.java#L214 and if we use the same property key it might cause a collision at some point... or are we safe? 😃

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/impl/StandalonePojoMappingInitiator.java
```diff
@@ -39,25 +60,46 @@ public StandalonePojoMappingInitiator(StandalonePojoBootstrapIntrospector intros
...
 
 	@Override
 	protected PojoMapperDelegate<StandalonePojoMappingPartialBuildState> createMapperDelegate() {
 		return new StandalonePojoMapperDelegate( entityTypeMetadataProvider, schemaManagementListener );
 	}
+
+	@Override
+	public void multiTenancyEnabled(boolean multiTenancyEnabled) {
+		this.tenancyMode( multiTenancyEnabled ? TenancyMode.MULTI_TENANCY : TenancyMode.SINGLE_TENANCY );
+	}
```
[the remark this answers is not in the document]
marko-bekhta: if we get rid of the method from context as you are suggesting then this will go away automatically 😃

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/cfg/StandalonePojoMapperSettings.java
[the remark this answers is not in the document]
yrodiere: We're safe so far, and it might be a problem in the (distant) future. But like I said, the way to solve this would be to change `PREFIX`, not to introduce fillers here :)

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/mapping/SearchMappingBuilder.java
marko-bekhta: While doing this (in that last commit), `StandalonePojoMappingSetupHelper` had to be changed as it was based on the builder. I've added the `withConfigurer` and `appendConfigurer` methods. This affected quite a few test configurations along the way.

yrodiere: Can't you hide this? You could still have the same methods as before in the setup helper, having them populate a configurer that is added by default.
See `org.hibernate.search.documentation.testsupport.DocumentationSetupHelper#defaultMappingConfigurer` for an example of what I mean.

yrodiere: I'm being annoying here because we'll likely have the same problem in Quarkus: we'll need to expose the configurer to users but also configure stuff in Quarkus itself.

If the problem is only about adding entities, then maybe `addEntity` should remain in the builder and not in the configurer...

documentation/src/main/asciidoc/reference/mapper-pojo-standalone.asciidoc
```diff
@@ -46,6 +46,25 @@ include::{sourcedir}/org/hibernate/search/documentation/mapper/pojo/standalone/e
 <4> Build the `SearchMapping`.
 ====
 
+Alternatively a configurer class implementing `StandalonePojoMappingConfigurer` can be provided.
```
yrodiere, line 49: Shouldn't the example above disappear, if the `addEntities` method is only exposed through the configurer?

integrationtest/v5migrationhelper/engine/src/test/java/org/hibernate/search/testsupport/junit/V5MigrationHelperEngineSetupHelper.java
```diff
@@ -97,6 +101,25 @@ protected SearchMappingBuilder createBuilder() {
...
+				if ( userConfigurers instanceof Iterable ) {
+					for ( Object userConfigurer : (Iterable) userConfigurers ) {
+						configurers.add( userConfigurer );
+					}
+				}
+				else {
+					configurers.add( userConfigurers );
+				}
+
+			}
```
yrodiere, line 120: This doesn't do anything?

documentation/src/main/asciidoc/reference/mapper-pojo-standalone.asciidoc
marko-bekhta: That builder change actually affected a few more tests/examples from the documentation 🙈

yrodiere (approved):

PR #3296, opened 2022-11-02
HSEARCH-4737 + HSEARCH-4738 Upgrade to Elasticsearch client 8.5.0 and add compatibility with Elasticsearch 8.5

yrodiere:
* [HSEARCH-4738](https://hibernate.atlassian.net/browse/HSEARCH-4738): Add compatibility with Elasticsearch 8.5
* [HSEARCH-4737](https://hibernate.atlassian.net/browse/HSEARCH-4737): Upgrade to Elasticsearch client 8.5.0

PR #3300, opened 2022-11-02
HSEARCH-4740 Run tests against Elasticsearch 7.17.8

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4740

[HSEARCH-4740]: https://hibernate.atlassian.net/browse/HSEARCH-4740?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -159,7 +159,7 @@ Before submitting a pull requests, check your contribution:
   No need to run the whole test suite, the Continuous Integration will take care of that.
 
 **Note**: If you want to run specific tests of the `integrationtests/backend/tck` module from the IDE,
-you will need to rely on runner classes to to run them in the appropriate context:
+you will need to rely on runner classes to run them in the appropriate context:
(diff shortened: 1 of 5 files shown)


yrodiere:
Looks like Elasticsearch 7.17.7 is affected by https://github.com/elastic/elasticsearch/issues/84601... and also another bug that I had to work around in #3296... I'll have to think a bit and see what my options are.

yrodiere:
> Looks like Elasticsearch 7.17.7 is affected by [elastic/elasticsearch#84601](https://github.com/elastic/elasticsearch/issues/84601)... and also another bug that I had to work around in #3296... I'll have to think a bit and see what my options are.

7.17.8 is still affected.

yrodiere:
I disabled some tests on Elasticsearch 7.17.7+, since that's about all we can do at this point. Let's see how it goes.

marko-bekhta (approved):

yrodiere:
It worked, merging.

PR #3301, opened 2022-11-02
HSEARCH-4678 Use UUID ids for OutboxEvent entity

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4678

As it would complicate the code to support both UUID and Long ids interchangeably, I've gone with the UUIDs. 
Using `uuid-char` should give us support on any DBs. In case the default `CustomVersionOneStrategy` won't work for somebody there's a config property to use a different strategy. 
If this works fine and no changes will be needed I'll add migration notes. I've checked and we can generate uuids in Postgres, MySQL, MariaDB, Oracle, MSSQL, CockroachDB, so we can create scripts that migrate the column type even if there are still records present in the table. DB2... doesn't have a built-in support to generate UUIDs but I've found this [post](https://community.ibm.com/community/user/hybriddatamanagement/viewdocument/generating-universally-unique-ident) that we might point DB2 users to?

> make sure that no event is left unprocessed for a very long time

I've thought about this one and maybe we should just have a timestamp in the event table and do not rely on ids...


marko-bekhta:
The reason that the build fails for ORM6 is that the `UUIDGenerator` is deprecated. And looking at the latest version - https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/annotations/UuidGenerator.java#L29 it turns out that there's no way to provide a custom strategy anymore. So based on that, should we either not expose the property at all or give the user the 3 options AUTO/RANDOM/TIME as in the latest ORM?

yrodiere:
> give the user the 3 options AUTO/RANDOM/TIME as in the latest ORM?

This looks like the safest option.

yrodiere:
>  >  make sure that no event is left unprocessed for a very long time

I'm sure I wrote that somewhere, but strangely I can't find where...

> I've thought about this one and maybe we should just have a timestamp in the event table and do not rely on ids...

We could, though we need to make sure we don't rely on that for exact ordering: that wouldn't work, as different servers might have slight different clocks (e.g. 1 second off). It would only be useful to order events approximately.

Also, this needs to be included in the mirgation script; maybe initialize the `created` column to something like `2022-11-01T00:00:00 + <event.id interpreted as a number of seconds>`?

marko-bekhta:
> I'm sure I wrote that somewhere, but strangely I can't find where...

That's from the comments in the ticket - here https://hibernate.atlassian.net/browse/HSEARCH-4678?focusedCommentId=109863 3d bullet point 😃

marko-bekhta:
In the end, the `AUTO == RANDOM`  strategy for UUID generation is used by default. It's because of this note here -https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/annotations/UuidGenerator.java#L42-L43. The `TIME` one is used in a few tests where the order seems to have an impact.

> We could, though we need to make sure we don't rely on that for exact ordering: that wouldn't work, as different servers might have slight different clocks (e.g. 1 second off). It would only be useful to order events approximately.

I'm not sure how to reach this approximate sorting without roundtrips to DB, I thought about truncating the timestamp to seconds/minutes and then relying on ID sorting, but that won't help either, even with the time-based IDs, as those would be generated on the app side where the clocks aren't in sync...

By the way, can you build the orm6 modules locally without any issues? For some reason, it always fails for me locally...

yrodiere:
As mentioned on the chat, we don't need to *make* the ordering approximate: it already is. I was just saying we need to be extra careful in the future **not** to assume this ordering is exact/reliable :)

marko-bekhta:
I've tested the migration scripts on the DBs, and those are executed fine. I am not 100% sure about the data types (`varchar`/`timestamp`/`datetime`) for all DBs though...
For the DBs that do not have a function to generate a UUID a combination of some supposed-to-be-unique sequence and regexp replace is used.

The `org.hibernate.id.uuid.UuidGenerator` for ORM 6 seems to be unusable in XML - I've posted a question on that in a zulip topic. Using the deprecated one `org.hibernate.id.UUIDGenerator` in XML passes the builds. IIUC we should create a noarg constructor in `UuidGenerator` push the logic from the existing constructor into `configure(..)` and call that `configure(..)` from the existing constructor to make it work... 

Other than that I think it's ready for a review if the build passes on the CI. I've tested the orm6 patches locally and all seems to be fine.

Also it seems that HSEARCH-4700 changes might've introduced a flaky test - https://ci.hibernate.org/blue/rest/organizations/jenkins/pipelines/hibernate-search/branches/PR-3301/runs/9/nodes/21/steps/56/log/?start=0
```
ERROR] Tests run: 16, Failures: 1, Errors: 0, Skipped: 2, Time elapsed: 11.864 s <<< FAILURE! - in org.hibernate.search.engine.backend.orchestration.spi.BatchingExecutorTest
[ERROR] org.hibernate.search.engine.backend.orchestration.spi.BatchingExecutorTest.simple_newTasksBlockedWaitAndCompletes[operation submitter = BLOCKING]  Time elapsed: 3.936 s  <<< FAILURE!
org.mockito.exceptions.verification.NoInteractionsWanted: 

No interactions wanted here:
-> at org.hibernate.search.engine.backend.orchestration.spi.BatchingExecutorTest.verifyAsynchronouslyAndReset(BatchingExecutorTest.java:434)
But found this interaction on mock 'processorMock':
-> at org.hibernate.search.engine.backend.orchestration.spi.BatchingExecutor$BatchWorker.complete(BatchingExecutor.java:202)
***
For your reference, here is the list of all invocations ([?] - means unverified).
1. -> at org.hibernate.search.engine.backend.orchestration.spi.BatchingExecutor$BatchWorker.work(BatchingExecutor.java:178)
2. -> at org.hibernate.search.engine.backend.orchestration.spi.BatchingExecutor$BatchWorker.work(BatchingExecutor.java:190)
3. [?]-> at org.hibernate.search.engine.backend.orchestration.spi.BatchingExecutor$BatchWorker.complete(BatchingExecutor.java:202)
4. -> at org.hibernate.search.engine.backend.orchestration.spi.BatchingExecutor$BatchWorker.work(BatchingExecutor.java:178)
5. -> at org.hibernate.search.engine.backend.orchestration.spi.BatchingExecutor$BatchWorker.work(BatchingExecutor.java:190)
6. -> at org.hibernate.search.engine.backend.orchestration.spi.BatchingExecutor$BatchWorker.complete(BatchingExecutor.java:202)

	at org.hibernate.search.engine.backend.orchestration.spi.BatchingExecutorTest.verifyAsynchronouslyAndReset(BatchingExecutorTest.java:434)
	at org.hibernate.search.engine.backend.orchestration.spi.BatchingExecutorTest.simple_newTasksBlockedWaitAndCompletes(BatchingExecutorTest.java:414)
```
I'll keep an eye if that happens again.

documentation/src/main/asciidoc/migration/index.asciidoc
```diff
@@ -118,3 +118,187 @@ used to match no documents with the Lucene backend, but all documents with the E
 A boolean predicate with no clause will now consistently match no documents regardless of the backend.
 * API methods matching `*Async(..)` pattern (e.g. `SearchWorkspace#purgeAsync()`) will no longer block
 if internal queues of operations are full, but will throw `RejectedOperationException` instead.
+
+[[outboxpolling]]
+== Outbox polling system tables
+
```
yrodiere, line 124: Wouldn't that fit the "Data format and schema changes" section? As a subsection, perhaps?

documentation/src/main/asciidoc/migration/index.asciidoc
```diff
@@ -118,3 +118,187 @@ used to match no documents with the Lucene backend, but all documents with the E
...
+Due to changes to the entity that represents the outbox event, DB schema changes are required.
```
yrodiere, line 125:
```suggestion
If you use the incubating link:{hibernateSearchDocUrl}#coordination-outbox-polling[`outbox-polling` coordination strategy],
you will be impacted by changes to the entity that represents the outbox event,
requiring database schema changes.
```

documentation/src/main/asciidoc/migration/index.asciidoc
```diff
@@ -118,3 +118,187 @@ used to match no documents with the Lucene backend, but all documents with the E
...
+A new column -  `created::timestamp` is introduced, `id` column changes the type from integer to varchar.
+You can find suggested migration scripts for the supported databases below:
```
yrodiere, line 127: I know I used the term "supported", but I was taking liberties: we don't really provide support, and we don't even document which databases are supposed to work correctly... Let's not imply too much here.

```suggestion
You can find suggested migration scripts for the tested databases below:
```

documentation/src/main/asciidoc/migration/index.asciidoc
```diff
@@ -118,3 +118,187 @@ used to match no documents with the Lucene backend, but all documents with the E
...
+
+.Postgresql:
+[,sql]
+----
+-- add `created` column that will be used for sorting of events:
+alter table hsearch_outbox_event
+    add column created timestamp not null default now();
```
yrodiere, line 134: You're setting a default value here; is that part of the schema? Because it shouldn't be, as the schema generated by Hibernate Search doesn't include it.

Maybe you shouldn't use `not null`/`default` here, and just add an `alter` statement later to add the `not null` constraint?

documentation/src/main/asciidoc/migration/index.asciidoc
```diff
@@ -118,3 +118,187 @@ used to match no documents with the Lucene backend, but all documents with the E
...
+alter table hsearch_outbox_event
+    alter column id TYPE varchar(36) USING gen_random_uuid();
+----
+
+.CockroachDB:
+[,sql]
+----
+-- add `created` column that will be used for sorting of events:
+alter table hsearch_outbox_event
+    add column created timestamp not null default now();
```
yrodiere, line 151: Same comment regarding the `default` statement.

documentation/src/main/asciidoc/migration/index.asciidoc
```diff
@@ -118,3 +118,187 @@ used to match no documents with the Lucene backend, but all documents with the E
...
+alter table hsearch_outbox_event
+    rename column tmp to id;
+----
+
+.MySQL:
+[,sql]
+----
+-- add `created` column that will be used for sorting of events:
+alter table hsearch_outbox_event
+    add column created datetime not null default now();
```
yrodiere, line 175: Same comment regarding the `default` statement.

documentation/src/main/asciidoc/migration/index.asciidoc
```diff
@@ -118,3 +118,187 @@ used to match no documents with the Lucene backend, but all documents with the E
...
+set id = uuid()
+where 1 = 1;
+----
+
+.MariaDB:
+[,sql]
+----
+-- add `created` column that will be used for sorting of events:
+alter table hsearch_outbox_event
+    add column created datetime not null default now();
```
yrodiere, line 198: Same comment regarding the `default` statement.

documentation/src/main/asciidoc/migration/index.asciidoc
```diff
@@ -118,3 +118,187 @@ used to match no documents with the Lucene backend, but all documents with the E
...
+set id = uuid()
+where 1 = 1;
+----
+
+.DB2:
+[,sql]
+----
+-- add `created` column that will be used for sorting of events:
+alter table hsearch_outbox_event
+    add created timestamp not null default current_timestamp ;
```
yrodiere, line 218: Same comment regarding the `default` statement.

documentation/src/main/asciidoc/migration/index.asciidoc
```diff
@@ -118,3 +118,187 @@ used to match no documents with the Lucene backend, but all documents with the E
...
+set id = regexp_replace(concat(rawtohex(generate_unique()), 'AAAAAA'), '([A-F0-9]{8})([A-F0-9]{4})([A-F0-9]{4})([A-F0-9]{4})([A-F0-9]{12})', '\1-\2-\3-\4-\5')
+where 1 = 1;
+----
+
+.Oracle:
+[,sql]
+----
+-- add `created` column that will be used for sorting of events:
+alter table hsearch_outbox_event
+    add created timestamp default CURRENT_TIMESTAMP not null;
```
yrodiere, line 244: Same comment regarding the `default` statement.

documentation/src/main/asciidoc/migration/index.asciidoc
```diff
@@ -118,3 +118,187 @@ used to match no documents with the Lucene backend, but all documents with the E
...
+alter table hsearch_outbox_event
+    add constraint hsearch_outbox_event_pkey primary key (id);
+----
+
+.MSSQL:
+[,sql]
+----
+-- add `created` column that will be used for sorting of events:
+alter table hsearch_outbox_event
+    add created datetime not null default current_timestamp;
```
yrodiere, line 267: Same comment regarding the `default` statement.

documentation/src/main/asciidoc/migration/index.asciidoc
```diff
@@ -118,3 +118,187 @@ used to match no documents with the Lucene backend, but all documents with the E
...
+set id = newid()
+where 1 = 1;
+----
+
+.H2:
+[,sql]
+----
+-- add `created` column that will be used for sorting of events:
+alter table hsearch_outbox_event
+    add created datetime not null default current_timestamp;
```
yrodiere, line 291: Same comment regarding the `default` statement.

documentation/src/main/asciidoc/reference/coordination.asciidoc
```diff
@@ -299,7 +299,8 @@ Identity generator names used for these tables are prefixed with `HSEARCH_` and
 Sometimes there are specific naming conventions for database objects, or a need to separate the domain
 and technical tables.
 To allow some flexibility in this area, Hibernate Search provides a set of configuration properties
-to specify catalog/schema/table/identity generator names for outbox event and agent tables:
+to specify catalog/schema/table names for outbox event and agent tables,
+as well as identity generator name for agent table and custom UUID generator strategy for outbox table:
```
yrodiere, line 303:
```suggestion
as well as an identity generator name for the agent table and a custom UUID generator strategy for the outbox event table:
```

integrationtest/mapper/orm-coordination-outbox-polling/src/test/java/org/hibernate/search/integrationtest/mapper/orm/coordination/outboxpolling/FilteringOutboxEventFinder.java
```diff
@@ -99,15 +100,15 @@ public synchronized void hideAllEvents() {
 	public List<OutboxEvent> findOutboxEventsNoFilter(Session session) {
 		checkFiltering();
 		Query<OutboxEvent> query = session.createQuery(
-				"select e from " + ENTITY_NAME + " e order by e.id", OutboxEvent.class );
+				"select e from " + ENTITY_NAME + " e order by e.created, e.id", OutboxEvent.class );
```
yrodiere, line 103: Ha. I just remembered that we used to do that in the past, but we stopped because timestamp precision wasn't great in some databases (millisecond).

That's not a problem in real-world applications as we don't need the ordering to be exact, but it **is** a problem in tests where two subsequent events might end up with the same created timestamp and thus end up with an undefined relative order...

Let's run tests multiple times on each database before we merge this, because we might have unpleasant surprises (flaky tests).

integrationtest/mapper/orm-coordination-outbox-polling/src/test/java/org/hibernate/search/integrationtest/mapper/orm/coordination/outboxpolling/automaticindexing/OutboxPollingAutomaticIndexingEventSendingIT.java
```diff
@@ -78,6 +78,8 @@ public void setup(OrmSetupHelper.SetupContext setupContext, ReusableOrmSetupHold
 		);
 
 		setupContext.withProperty( "hibernate.search.coordination.outbox_event_finder.provider", outboxEventFinder.provider() )
+				// use timebase uuids to get predictable sorting order
+				.withProperty( "hibernate.search.coordination.entity.mapping.outboxevent.uuid_gen_strategy", "time" )
```
yrodiere, line 82: well well, if that isn't cheating :)

Ok, makes sense!

integrationtest/mapper/orm-coordination-outbox-polling/src/test/java/org/hibernate/search/integrationtest/mapper/orm/coordination/outboxpolling/automaticindexing/OutboxPollingAutomaticIndexingOutOfOrderIdsIT.java
```diff
@@ -370,13 +371,41 @@ private void updateOutboxTableRow(Session session, Integer oldId, Integer newId)
...
+				jdbc.getResultSetReturn().executeUpdate( ps );
+			}
+
+			try ( PreparedStatement ps = jdbc.getStatementPreparer().prepareStatement( OUTBOX_EVENT_UPDATE_ID_AND_TIME ) ) {
+				ps.setString( 1, uuids.get( row1 ) );
+				ps.setTimestamp( 2, times.get( row1 ) );
+				ps.setString( 3, temporaryUuid );
 
 				jdbc.getResultSetReturn().executeUpdate( ps );
 			}
```
yrodiere, line 411: Would you mind moving this repeated code to a method, say... `updateOutboxTableRow(Session session, UUID oldId, UUID newId, java.sql.Timestamp newCreated)`? :) It's a bit hard to understand as it is, even if I can guess what it's supposed to do.

orm6/mapper/orm-coordination-outbox-polling/ant-src-changes.patch
```diff
@@ -1,5 +1,38 @@
+diff --git a/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/cfg/HibernateOrmMapperOutboxPollingSettings.java b/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/cfg/HibernateOrmMapperOutboxPollingSettings.java
```
yrodiere, line 1: I suppose you used the scripts to update the patches?

Can you please push the corresponding commits to the `wip/main/dependency-update/orm6-in-main-code` branch?

yrodiere (commented):
Thanks! Here are a few comments.

integrationtest/mapper/orm-coordination-outbox-polling/src/test/java/org/hibernate/search/integrationtest/mapper/orm/coordination/outboxpolling/automaticindexing/OutboxPollingAutomaticIndexingEventSendingIT.java
marko-bekhta: 😊

> Let's run tests multiple times on each database before we merge this, because we might have unpleasant surprises (flaky tests).

Because we did get the flaky tests 😉

documentation/src/main/asciidoc/migration/index.asciidoc
marko-bekhta: That was where I wanted to put it ... but then I thought that that section is for index schema, so placed it as its own section 😄. Will move it!

yrodiere: You're right, it was, but only because I didn't anticipate DB schema changes. It makes sense to just change the purpose of that section :)

yrodiere (approved):

PR #3306, opened 2022-11-10
HSEARCH-4712  Convert synchronized blocks to java.util.concurrent.Locks for better Loom compatibility

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4712

There weren't actually that many sync blocks inside of the methods. Ones inside `IndexWriterDelegatorImpl` looked like good candidates for replacement as they are somewhat using sync blocks to do I/O operations. But the underlying Lucene code is using the same pattern, so I suppose it won't help much until that is changed as well:

```java 
  private long commitInternal(MergePolicy mergePolicy) throws IOException {
    // .... 
    synchronized(commitLock) {
      ensureOpen(false);
      // .... 
      finishCommit();
    }
    // .... 
    return seqNo;
  }
```

The other sync blocks were either in test appenders or where sync only happens on the init stage and only performs in-memory operations, so doesn't look like candidates for replacement.

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/writer/impl/IndexWriterDelegatorImpl.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/writer/impl/IndexWriterDelegatorImpl.java
@@ -184,7 +184,8 @@ public class IndexWriterDelegatorImpl implements IndexWriterDelegator {
 	private void doCommit() {
 		commitLock.lock();
 		try {
-			// NOTE: underlying Lucene code is using same pattern to sync on object block:
+			// NOTE: underlying Lucene code is using this pattern to sync on object block,
+			// which could be a problem with Loom:
 			// synchronized(commitLock)
 			delegate.commit();
 			updateCommitExpiration();
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/writer/impl/IndexWriterDelegatorImpl.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/writer/impl/IndexWriterDelegatorImpl.java
@@ -12,6 +12,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.hibernate.search.backend.lucene.logging.impl.Log;
 import org.hibernate.search.engine.common.timing.spi.TimingSource;
@@ -42,7 +43,7 @@ public class IndexWriterDelegatorImpl implements IndexWriterDelegator {
 	private final FailureHandler failureHandler;
 
 	private final SingletonTask delayedCommitTask;
-	private final Object commitLock = new Object();
+	private final ReentrantLock commitLock = new ReentrantLock();
 
 	private long commitExpiration;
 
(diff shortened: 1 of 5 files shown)


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/writer/impl/IndexWriterDelegatorImpl.java
```diff
@@ -173,15 +182,19 @@ void closeAfterFailure(Throwable throwable, Object failingOperation) {
 	}
 
 	private void doCommit() {
+		commitLock.lock();
 		try {
-			synchronized (commitLock) {
-				delegate.commit();
-				updateCommitExpiration();
-			}
+			// NOTE: underlying Lucene code is using same pattern to sync on object block:
```
yrodiere, line 187: Nice to add a comment, but I think it needs a bit more context :)

```suggestion
			// NOTE: underlying Lucene code is using this pattern to sync on object block,
			// which could be a problem with Loom:
```

engine/src/main/java/org/hibernate/search/engine/common/resources/impl/EngineThreads.java
```diff
@@ -33,6 +33,9 @@ public ScheduledExecutorService getTimingExecutor() {
 		if ( executor != null ) {
 			return executor;
 		}
+		// this sync block doesn't seem to require a change:
+		// - sync happens only once on init ?
+		// - no I/O and simple in-memory operations
```
yrodiere, line 38: I agree, but here too I think the comment needs some work.

```suggestion
		// this sync block shouldn't be a problem for Loom:
		// - sync happens only once on init ?
		// - no I/O and simple in-memory operations
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingLoggingMonitor.java
```diff
@@ -53,6 +53,9 @@ public PojoMassIndexingLoggingMonitor(int logAfterNumberOfDocuments) {
 	@Override
 	public void documentsAdded(long increment) {
 		if ( startTime == 0 ) {
+			// this sync block doesn't seem to require a change:
+			// - sync happens only once at start?
+			// - no I/O and simple in-memory operations
```
yrodiere, line 58: No, it doesn't happen just once at start. Regardless, I agree it shouldn't be a problem sine we don't use virtual threads in the mass indexer. Here too I think the comment needs some work.

```suggestion
			// this sync block doesn't seem to be a problem for Loom:
			// - always executed in MassIndexer threads, which are not virtual threads
			// - no I/O and simple in-memory operations
```

build/config/src/main/resources/checkstyle.xml
```diff
@@ -94,6 +94,13 @@
             <property name="ignoreComments" value="true" />
         </module>
 
+        <!-- Do not use sync blocks on objects -->
+        <module name="RegexpSinglelineJava">
+            <property name="format" value="synchronized\s+\(\s*(?!this\b)\w+" />
```
yrodiere, line 99: Let's be on the safe side...

```suggestion
            <property name="format" value="synchronized\s*\(\s*(?!this\b)\w+" />
```

engine/src/main/java/org/hibernate/search/engine/common/resources/impl/EngineThreads.java
yrodiere: Also, you might want to add the same comment to `org.hibernate.search.engine.common.timing.impl.DefaultTimingSource#ensureTimeEstimateIsInitialized`.

yrodiere (commented):
LGTM, just a few comments about... the comments :)

yrodiere (approved):

PR #3307, opened 2022-11-10
HSEARCH-4749 Switch to binary representation of UUIDs in DB

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4749

Tests against MSSQL uncovered a "deadlocking" issue when UUID is represented as a string in the DB, while binary representation is working just fine. So the idea is to make the UUID data type configurable and use the ones we've tested against as defaults. If there's a DB we aren't testing, the user can pick the type they want or pass their own "custom" one. 

The property key is similar to the one used in ORM. The changes in SPI interface should be safe as I don't think that anybody except HSEARCH itself should be using that.


mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/cfg/UuidDataType.java
```diff
@@ -0,0 +1,72 @@
...
+		}
+	},
+	/**
+	 *
+	 */
+	UUID( "uuid" ),
+
+	BINARY( "uuid-binary" ),
+
+	CHAR( "uuid-char" );
```
gsmet, line 47: Any reason not to call the `binary` and `char`? Not that I'm not that accustomed to the Search codebase so there might be a good reason.

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/cfg/UuidDataType.java
```diff
@@ -0,0 +1,72 @@
...
+			// See HSEARCH-4749 why MSSQL is treated differently
+			if ( dialect instanceof SQLServerDialect ) {
+				return BINARY.type( dialect );
+			}
+			return CHAR.type( dialect );
+		}
+	},
+	/**
+	 *
+	 */
```
gsmet, line 42: This empty comment should probably go away?

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/cfg/UuidDataType.java
```diff
@@ -0,0 +1,72 @@
...
+	/**
+	 * Is database specific. Some databases might not work correctly with a particular representation of UUID hence the
+	 * default provides the value against which particular database was tested. If the database is not on the testing list
+	 * the default would use a character representation as, characters should be overall supported by a database at least
+	 * as a type.
+	 * <p>
+	 * Note that in case of MS SQL database using {@link #CHAR} representation might lead to undesired behavior and
+	 * occurrence of deadlocks.
+	 */
+	DEFAULT( COORDINATION_ENTITY_MAPPING_OUTBOX_EVENT_UUID_DATA_TYPE ) {
```
gsmet, line 30: I think the value should be defined here. And maybe it should be called `default` except if `hsearch_default` is used elsewhere for a similar purpose (it might be as Yoann likes to be explicit).

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/cfg/HibernateOrmMapperOutboxPollingSettings.java
```diff
@@ -542,6 +543,7 @@ private Defaults() {
 		// Must not be longer than 20 characters, so that the generator does not exceed the 30 characters for Oracle11g
 		public static final String COORDINATION_ENTITY_MAPPING_OUTBOX_EVENT_TABLE = HSEARCH_PREFIX + "OUTBOX_EVENT";
 		public static final UuidGenerationStrategy COORDINATION_ENTITY_MAPPING_OUTBOX_EVENT_UUID_GEN_STRATEGY = UuidGenerationStrategy.AUTO;
+		public static final String COORDINATION_ENTITY_MAPPING_OUTBOX_EVENT_UUID_DATA_TYPE = "hsearch_default";
```
gsmet, line 546: Here I would reference the enum if it's not too impractical.

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/cfg/UuidDataType.java
```diff
@@ -0,0 +1,72 @@
...
+public enum UuidDataType {
+	/**
+	 * Is database specific. Some databases might not work correctly with a particular representation of UUID hence the
+	 * default provides the value against which particular database was tested. If the database is not on the testing list
+	 * the default would use a character representation as, characters should be overall supported by a database at least
+	 * as a type.
+	 * <p>
+	 * Note that in case of MS SQL database using {@link #CHAR} representation might lead to undesired behavior and
+	 * occurrence of deadlocks.
+	 */
```
gsmet, line 29: I suggested a few minor improvements (it needs to be reformatted as I added/removed some content):
```suggestion
	/**
	 * Database specific. Some databases might not work correctly with a particular representation of UUID hence the
	 * default provides a sensible value depending on which particular database is used. By default, a character representation is used as characters should be overall supported by databases.
	 * <p>
	 * But for some databases, a different value is used. For instance, in the case of Microsoft SQL Server, using a character representation might lead to undesired behavior in the form of deadlocks so a binary representation is used by default.
	 */
```

gsmet (changes requested):
I added a few suggestions.

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/cfg/UuidDataType.java
marko-bekhta: The idea was to match the types from ORM, so that it might look more familiar to the users. I'd probably say let's keep it like that for now and then let Yoann make a final call 😄 
Everything else makes sense - will make the adjustments, thanks!

marko-bekhta:
Changes applied. While doing so I've noticed that I've missed creating a constant at the most top-level here https://github.com/hibernate/hibernate-search/pull/3307/files#diff-3e0430d10bd8b7f6272338e6d746222f79ccb2c594d92704c3ed56afd8a82b56R403-R417 hence done so with the other changes.

gsmet (approved):
OK let's get this in to stabilize CI.
Maybe put it in a list of things Yoann should check when he's back.

marko-bekhta:
Thanks! yeah already started to make notes for him 😃 hopefully, it won't grow too much

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/cfg/UuidDataType.java
```diff
@@ -0,0 +1,63 @@
...
+		return externalRepresentation();
+	}
+
+	public static String uuidType(String value, Dialect dialect) {
+		for ( UuidDataType dataType : UuidDataType.values() ) {
+			if ( dataType.externalRepresentation().equals( value ) ) {
+				return dataType.type( dialect );
+			}
+		}
+		return value;
```
yrodiere, line 61: Can you please move this out of this class? The class is API, so we don't want to expose things that may have to change.

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/cfg/UuidDataType.java
```diff
@@ -0,0 +1,63 @@
...
+
+	private final String externalRepresentation;
+
+	UuidDataType(String externalRepresentation) {
+		this.externalRepresentation = externalRepresentation;
+	}
+
+	public String externalRepresentation() {
+		return externalRepresentation;
+	}
```
yrodiere, line 49: You will most likely need to introduce an `of()` method similar to [this](https://github.com/hibernate/hibernate-search/blob/5bfe116a9b5810e069d36c40e7ee86775dcc954b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/multitenancy/MultiTenancyStrategyName.java#L29-L37) so that we can configure this more easily in Quarkus.

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/cfg/HibernateOrmMapperOutboxPollingSettings.java
```diff
@@ -511,6 +523,7 @@ private CoordinationRadicals() {
 		public static final String ENTITY_MAPPING_OUTBOXEVENT_SCHEMA = ENTITY_MAPPING_OUTBOXEVENT_PREFIX + "schema";
 		public static final String ENTITY_MAPPING_OUTBOXEVENT_CATALOG = ENTITY_MAPPING_OUTBOXEVENT_PREFIX + "catalog";
 		public static final String ENTITY_MAPPING_OUTBOXEVENT_UUID_GEN_STRATEGY = ENTITY_MAPPING_OUTBOXEVENT_PREFIX + "uuid_gen_strategy";
+		public static final String ENTITY_MAPPING_OUTBOXEVENT_UUID_DATA_TYPE = ENTITY_MAPPING_OUTBOXEVENT_PREFIX + "preferred_uuid_jdbc_type";
```
yrodiere, line 526: As far as I can tell this is not just a preference, but a mandated type which Hibernate Search will always comply with.

Would you mind renaming `preferred_uuid_jdbc_type` to `uuid_jdbc_type`?

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/cfg/UuidDataType.java
yrodiere: Either is fine to me, but please keep the name of enum values and the external representation consistent unless there's a good reason.

I.e. either `BINARY` with external representation `binary` OR `UUID_BINARY` with external representation `uuid-binary`.

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/cfg/HibernateOrmMapperOutboxPollingSettings.java
marko-bekhta: Sure, will change; I've used the key as in ORM so that it might look more familiar to the users but can see how the "preferred" part is a bit confusing 😄

PR #3323, opened 2022-12-07
HSEARCH-4757 Adjust tests to address ORM changes introduced in HHH-15767

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4757

### DIFF
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/association/bytype/AbstractAutomaticIndexingAssociationBaseIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/association/bytype/AbstractAutomaticIndexingAssociationBaseIT.java
@@ -207,6 +207,10 @@ public abstract class AbstractAutomaticIndexingAssociationBaseIT<
 
 	protected abstract boolean isAssociationLazyOnContainingSide();
 
+	protected boolean isEmbeddedAssociationChangeCausingWork() {
+		return !isAssociationOwnedByContainedSide() && !isAssociationMultiValuedOnContainingSide();
+	}
+
 	private boolean isElementCollectionAssociationsOnContainingSide() {
 		return !isAssociationOwnedByContainedSide() && !isAssociationMultiValuedOnContainingSide();
 	}
@@ -450,6 +454,18 @@ public abstract class AbstractAutomaticIndexingAssociationBaseIT<
 
 			TContained oldContained = containingAssociation.get( entity1 );
 			containedAssociation.clear( oldContained );
+			// Clear the one-to-one association and flush it first, to avoid problems with unique key constraints. See details in HHH-15767
+			if ( !isAssociationMultiValuedOnContainingSide() ) {
+				containingAssociation.clear( entity1 );
+
+				if ( setupHolder.areEntitiesProcessedInSession() ) {
+					backendMock.expectWorks( _indexed().indexName() )
+							.addOrUpdate( "1", b -> {
+							} );
+				}
+				session.flush();
+			}
+
 			containingAssociation.set( entity1, containedEntity );
 			containedAssociation.set( containedEntity, entity1 );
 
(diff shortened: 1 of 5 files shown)


integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/association/bytype/AbstractAutomaticIndexingAssociationBaseIT.java
```diff
@@ -450,6 +464,15 @@ public void directAssociationUpdate_indexedEmbedded() {
...
 			TContained oldContained = containingAssociation.get( entity1 );
 			containedAssociation.clear( oldContained );
+			containingAssociation.clear( entity1 );
+
+			if ( !isOutboxPollingCoordinationStrategyPresent() ) {
+				backendMock.expectWorks( _indexed().indexName() )
+						.addOrUpdate( "1", b -> {
+						} );
+			}
+			session.flush();
```
yrodiere, line 474: I know the existing code is not very clear, but I have a suggestion to avoid complexifying it even more: could you group your additions together, with a comment on top explaining this is just to avoid problems with DB constraints?

You will need to do the same for every addition in other methods/files, though.

Something like this:

```suggestion

			// Clear the association first, to avoid problems with unique key constraints.
			containingAssociation.clear( entity1 );
			if ( !isOutboxPollingCoordinationStrategyPresent() ) {
				backendMock.expectWorks( _indexed().indexName() )
						.addOrUpdate( "1", b -> {
						} );
			}
			session.flush();
```

Or rather, with the condition I mentioned elsewhere:

```suggestion

			if ( !isAssociationMultiValuedOnContainingSide() ) {
				// Clear the association first, to avoid problems with unique key constraints.
				containingAssociation.clear( entity1 );
				if ( !isOutboxPollingCoordinationStrategyPresent() ) {
					backendMock.expectWorks( _indexed().indexName() )
							.addOrUpdate( "1", b -> {
							} );
				}
				session.flush();
			}
```

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/association/bytype/AbstractAutomaticIndexingAssociationBaseIT.java
```diff
@@ -450,6 +464,15 @@ public void directAssociationUpdate_indexedEmbedded() {
 
 			TContained oldContained = containingAssociation.get( entity1 );
 			containedAssociation.clear( oldContained );
+			containingAssociation.clear( entity1 );
```
yrodiere, line 467: I think this is only necessary for single-valued associations? Could you add a `if (!isAssociationMultiValuedOnContainingSide())` (or whatever condition makes sense) to only execute this code when absolutely necessary? 

The reason: by introducing this change, you're giving Hibernate Search more "clues" that something changed, so you're helping it detect changes. I'm worried this could hide flaws in the change detection code.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/association/bytype/AbstractAutomaticIndexingAssociationBaseIT.java
```diff
@@ -207,6 +207,20 @@ public AbstractAutomaticIndexingAssociationBaseIT(IndexedEntityPrimitives<TIndex
 
 	protected abstract boolean isAssociationLazyOnContainingSide();
 
+	protected boolean isFlushCausingAdditionalWork() {
```
yrodiere, line 210: The name of this method isn't great... we don't know which flush we're talking about, but you mean one specific flush on a specific embedded association. Maybe `isEmbeddedAssociationChangeCausingWork()` would be better?

Regardless, I suspect you don't need to override this method: you could just implement it right here with `!isAssociationOwnedByContainedSide() && !isAssociationMultiValuedOnContainingSide()`...  That's the same implementation as `isElementCollectionAssociationsOnContainingSide`, but I suspect that's just a coincidence.

yrodiere (changes requested):

yrodiere (commented):
Looks good, but if I try to run the tests against ORM 6.2, everything fails right now, though that seems related to a problem in the latest snapshot rather than in Hibernate Search itself.

Let's wait for the ORM team to solve their own problems before we test again and merge...

yrodiere (approved):
Tested locally against ORM 6.2 with the following command:

```
BRANCH_NAME=main ci/dependency-update/perform-update.sh orm6.2 version.org.hibernate.orm
```

It worked :)

PR #3333, opened 2022-12-13
HSEARCH-4667 Add a configuration properties reference to the reference documentation

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4667

I went with the annotation processor option as that's something a bit familiar 😃 
The processor goes through top-level classes and filters for the ones ending with `Settings`. It collects the properties it finds and then generates a JSON and asciidoc files. I had added JSON just in case we'd wanted maybe to combine all of the configs in one table or transform them in some other way. 

While working on this, I've noticed that we won't be able to be flexible enough if we don't have some sort of marker annotation. Hence you'll see `@HibernateSearchConfiguration`. 
Annotation processor has a couple of config properties described in `Configuration`. Then we can override some of these with an annotation if needed.

A bit of a hackish part is the way rendered Javadoc is retrieved. To not render it on our own, I've moved the Javadoc plugin to be executed before the annotation processor run. This allows us to access the rendered docs and get the parts we are interested in. In case the docs aren't there processor would try to get the "raw docs". 

here's a screen of how it looks when rendered:
image

I've noticed that there are some URLs in docs that are not pointing to Hibernate Search docs - hence some adjustments will be needed in the processor; I'll take a look into it tomorrow.


build/configuration-properties-collector/src/main/java/org/hibernate/search/configuration/properties/collector/ConfigurationPropertyCollector.java
```diff
@@ -0,0 +1,197 @@
...
+
+			org.jsoup.select.Elements blocks = javadoc.select( "#" + constant.getSimpleName() + " .block" );
+			if ( !blocks.isEmpty() ) {
+				org.jsoup.nodes.Element block = blocks.get( 0 );
+				for ( org.jsoup.nodes.Element link : block.getElementsByTag( "a" ) ) {
+					// only update links if they are not external:
+					if ( !link.hasClass( "external-link" ) ) {
+						link.attr( "href", javadocsBaseLink + packagePath + "/" + link.attr( "href" ) );
+					}
+				}
```
yrodiere, line 161: Not sure how this works, but some links seem to be incorrect. See the documentation for `hibernate.search.coordination.entity.mapping.outboxevent.catalog`, for example.

yrodiere (commented):
Nice! I must admit I didn't check the implementation, because if it works, well... It's good enough for me :)

But here are a few comments about Maven config, usage, and rendering.

backend/elasticsearch-aws/src/main/java/org/hibernate/search/backend/elasticsearch/aws/cfg/ElasticsearchAwsBackendSettings.java
```diff
@@ -6,12 +6,15 @@
...
 
+import org.hibernate.search.configuration.properties.collector.HibernateSearchConfiguration;
+
 /**
  * AWS-specific configuration properties for Elasticsearch backends.
  * <p>
  * Constants in this class are to be appended to a prefix to form a property key;
  * see {@link org.hibernate.search.engine.cfg.BackendSettings} for details.
  */
+@HibernateSearchConfiguration(prefix = "hibernate.search.backends.")
```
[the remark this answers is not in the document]
marko-bekhta: My parsing logic fails to find a difference between the first two bullets (both point to `*BackendSettings`) 😄 🙈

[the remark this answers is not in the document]
yrodiere: Looks like my writing logic failed; I updated my comment :)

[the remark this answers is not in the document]
yrodiere: But maybe you intended to extract this processor and use it in other places (e.g. Hibernate ORM)?
In that case it makes sense to rely on explicit attributes rather than a hardcoded logic.

It would also make sense to keep the explicit `type` attribute, since not every project follows the same package naming conventions and `AUTO` might not work everywhere. Also some packages in ORM are misnamed and I suspect some SPI properties are mixed with API properties, so they would need overrides.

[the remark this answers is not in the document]
marko-bekhta: I didn't want to super tie it to Search, so that's why there are those config properties or the annotation. Otherwise, yeah.. we could've put most of the filtering hardcoded into the processor itself 😃.
I've removed the type from the annotation, but the enum is still there. 
As for the ORM - if it could be of use there, I can try to see how we could adapt this processor to work with ORM too.
And thanks for updating the comment 👍🏻🏆

build/configuration-properties-collector/src/main/java/org/hibernate/search/configuration/properties/collector/ConfigurationPropertyCollector.java
```diff
@@ -0,0 +1,211 @@
...
+					if ( !link.hasClass( "external-link" ) ) {
+						if ( href.startsWith( "#" ) ) {
+							href = enclosingClass.getSimpleName().toString() + ".html" + href;
+						}
+						href = javadocsBaseLink + packagePath + "/" + href;
+					}
+					else if ( href.contains( "/search/build/parents/" ) && href.contains( "/apidocs" ) ) {
+						// means a link was to a class from other module and javadoc plugin generated some external link
+						// that won't work. So we replace it:
+						href = javadocsBaseLink + href.substring( href.indexOf( "/apidocs" ) + 8 );
```
yrodiere, line 172: No magic numbers pretty please :x 

```suggestion
						href = javadocsBaseLink + href.substring( href.indexOf( "/apidocs" ) + "/apidocs".length() );
```

build/configuration-properties-collector/src/main/java/org/hibernate/search/configuration/properties/collector/ConfigurationPropertyCollector.java
```diff
@@ -0,0 +1,211 @@
...
+						}
+					}
+				}
+			}
+		}
+		return null;
+	}
+
+	private static String[] nullIfBlank(String[] str) {
+		return str != null && false ? null : str;
```
yrodiere, line 208: `&& false`?

build/configuration-properties-collector/src/main/java/org/hibernate/search/configuration/properties/collector/utils/AsciiDocWriter.java
```diff
@@ -0,0 +1,80 @@
...
+			// nothing to write - return fast.
+			return;
+		}
+		try {
+			moduleName.ifPresent( name -> tryToWriteLine( "== ", name, writer ) );
+			writer.write( '\n' );
+			for ( Map.Entry<String, ConfigurationProperty> entry : entries ) {
+				writer.write( entry.getValue().key().resolvedKeys().stream()
+						.map( key -> String.format( Locale.ROOT, "`%s`", key ) )
+						.collect( Collectors.joining( "/" ) ) );
```
yrodiere, line 48: I suspect that won't be pretty as this will be rather long...

Maybe something like:

```asciidoc
`the.first.key`::
Other variants: `the.second.key`, `the.third.key`
+
The description.
```

build/configuration-properties-collector/src/main/java/org/hibernate/search/configuration/properties/collector/utils/AsciiDocWriter.java
```diff
@@ -0,0 +1,80 @@
...
+
+				String defaultValue = Objects.toString( entry.getValue().defaultValue(), "" );
+				if ( !defaultValue.isBlank() ) {
+					writer.write( " (default value: `" );
+					writer.write( defaultValue );
+					writer.write( "`)" );
+				}
```
yrodiere, line 55: Same here, to accomodate narrow screens (smartphones), it would be better to move that to the description.

E.g.:

```asciidoc
`the.first.key`::
Other variants: `the.second.key`, `the.third.key`
+
Default value: `the default`
+
The description.
```

And yes I know, this format was my idea: I didn't anticipate the crazy length of some properties :)

backend/elasticsearch-aws/src/main/java/org/hibernate/search/backend/elasticsearch/aws/cfg/ElasticsearchAwsBackendSettings.java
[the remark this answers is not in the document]
yrodiere: >  As for the ORM - if it could be of use there, I can try to see how we could adapt this processor to work with ORM too.

Better check with Steve if he finds that useful. But yeah IMO that could be nice.

yrodiere (commented):
My two cents. I'll wait for CI and have another look at the result, but this looks promising :)

backend/elasticsearch-aws/src/main/java/org/hibernate/search/backend/elasticsearch/aws/cfg/ElasticsearchAwsBackendSettings.java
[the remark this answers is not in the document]
marko-bekhta: let's get it working nicely here and then show what we've got 😃

build/configuration-properties-collector/src/main/java/org/hibernate/search/configuration/properties/collector/utils/AsciiDocWriter.java
marko-bekhta: image
that'll be something along those lines ^ 😃

yrodiere: Looking good! Maybe put the default value at the end, though? Since we apparently talk about detaults at the end of the javadoc.

marko-bekhta:
```bash
[ERROR] Constraint: hsearch:PublicTypesMayNotExposeInternalTypes
[ERROR] Severity: MAJOR
[ERROR] Number of rows: 2
[ERROR] API/SPI types must not expose internal types.
[ERROR]   ExposingSite=org.hibernate.search.util.common.impl.HibernateSearchConfiguration$Type type() (in org.hibernate.search.configuration.properties.collector.ConfigurationProperty), ExposedType=org.hibernate.search.util.common.impl.HibernateSearchConfiguration$Type
[ERROR]   ExposingSite=org.hibernate.search.configuration.properties.collector.ConfigurationProperty type(org.hibernate.search.util.common.impl.HibernateSearchConfiguration$Type) (in org.hibernate.search.configuration.properties.collector.ConfigurationProperty), ExposedType=org.hibernate.search.util.common.impl.HibernateSearchConfiguration$Type
```
sooo.. I guess either move out the annotation from `org.hibernate.search.util.common.impl` to `org.hibernate.search.util.common` or move all of the processor classes into `org.hibernate.search.configuration.properties.collector.impl` package?

yrodiere:
> move all of the processor classes into org.hibernate.search.configuration.properties.collector.impl package?

Yes, that seems to make sense. They're not API, after all.

yrodiere:
@marko-bekhta Looks like asciidoc rendering fails when the generated files are empty, which for some reason happens to be the case when compiling with the Eclipse compiler on CI: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-3333/17/pipeline

We definitely don't care about generating non-empty files with the Eclipse compiler: we just test this on CI to make sure contributors using Eclipse won't get compilation errors.

However, but could you look at making this work just to make CI happy, and maybe also in case it happens for other reasons?

build/configuration-properties-collector/src/main/java/org/hibernate/search/configuration/properties/collector/impl/ConfigurationPropertyCollector.java
```diff
@@ -0,0 +1,219 @@
...
+				org.jsoup.nodes.Element block = blocks.get( 0 );
+				for ( org.jsoup.nodes.Element link : block.getElementsByTag( "a" ) ) {
+					String href = link.attr( "href" );
+					// only update links if they are not external:
+					if ( !link.hasClass( "external-link" ) ) {
+						if ( href.startsWith( "#" ) ) {
+							href = enclosingClass.getSimpleName().toString() + ".html" + href;
+						}
+						href = javadocsBaseLink + packagePath + "/" + href;
+					}
```
yrodiere, line 171: Real question, not necessarily asking you to do it: would it be complex to replace links to other configuration property constants with internal links to the Asciidoc you're currently generating?

E.g. in the example below, 
`"client.instance"` would point to the item about `hibernate.search.backend.client.instance` in the reference documentation, instead of pointing to the javadoc.

image

marko-bekhta:
Hmm, I've seen something similar happening in the past...
Actually, I've been thinking about this one a bit more while I've worked on a few other tasks where I've moved the properties + in terms of potential reuse in ORM. I had an idea I wanted to try. Instead of having the processor running for each module separately - create a new module where all required source jars will be unpacked + the final JavaDoc jar, and then the processor would go through these. This way, we should have a cleaner build + the JavaDocs should have all the links correctly populated... It would mean, though, that we'd need to have a `title` property in that annotation and use it to group the properties correctly.

build/configuration-properties-collector/src/main/java/org/hibernate/search/configuration/properties/collector/impl/AsciiDocWriter.java
```diff
@@ -0,0 +1,100 @@
...
+	public void accept(Map<String, ConfigurationProperty> propertyMap, Writer writer) {
+		List<Map.Entry<String, ConfigurationProperty>> entries = propertyMap.entrySet().stream()
+				.filter( filter )
+				.collect( Collectors.toList() );
+		if ( entries.isEmpty() ) {
+			// nothing to write - return fast.
+			return;
+		}
+		try {
+			moduleName.ifPresent( name -> tryToWriteLine( "== ", name, writer ) );
```
yrodiere, line 41: It would be safer to also generate a stable anchor, preferably based on the artifact ID, because the name may change (it's mostly cosmetic).

If it's easier to include just the artifact ID and not the name, that's fine by me.

Just in case, this is what I mean by "anchor":

```
[[configuration-properties-aggregated-hibernate-search-engine]]
== Hibernate Search Engine
```

yrodiere (changes requested):
I went through the whole output, and it looks great! I definitely expected more problems, especially when converting Javadoc to HTML...

See below for a few suggestions.

Also, it looks like prefixes are missing in some cases:

image

That being said, while `type` should be documented here, `indexes` probably shouldn't, and `hibernate.search.backend` and `hibernate.search.backends` either, since those are essentially just prefixes (just, the constant name doesn't make that obvious, unfortunately).

yrodiere:
> Instead of having the processor running for each module separately - create a new module where all required source jars will be unpacked + the final JavaDoc jar, and then the processor would go through these. This way, we should have a cleaner build + the JavaDocs should have all the links correctly populated...

If you think that makes sense... sure. Please put that in the `build` directory though.

Note that we also have some plugin executions to download the javadoc of *external* projects: https://github.com/hibernate/hibernate-search/blob/2dbdfaad4b4b75483091f7d9485ed5e1d2c77e34/build/config/pom.xml#L57-L123
It must run before we generate the javadoc of each module though, so you probably can't move it to your new module.

> It would mean, though, that we'd need to have a `title` property in that annotation and use it to group the properties correctly.

Not necessarily, as Javadoc JARs include a list of packages included in that JAR (that's what we use in the plugin execution I mentioned above). So in theory you could match a given class against those lists and guess the corresponding module.

build/configuration-properties-collector/src/main/java/org/hibernate/search/configuration/properties/collector/impl/ConfigurationPropertyCollector.java
marko-bekhta: would need to see how anchor links are generated by asciidoctor... Right now we pass this  HTML as is to ascidoctor and it just adds it without processing. So if we can predict the link, we can replace it. Otherwise, we could delay the replacement up until asciidocwriter and then try to play with formatting there and use the ascidoctor anchor/reference formatting.
I'll see what is available to us within this HTML and will let you know.

build/configuration-properties-collector/src/main/java/org/hibernate/search/configuration/properties/collector/impl/AsciiDocWriter.java
```diff
@@ -0,0 +1,100 @@
...
+
+				String defaultValue = Objects.toString( entry.getValue().defaultValue(), "" );
+				if ( !defaultValue.isBlank() ) {
+					writer.write( "\n+\n" );
+					writer.write( "Default value: `" );
+					writer.write( defaultValue );
+					writer.write( '`' );
+				}
+
+				writer.write( '\n' );
```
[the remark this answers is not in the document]
marko-bekhta: I'll give this [collapsible section](https://docs.asciidoctor.org/asciidoc/latest/blocks/collapsible/ ) a try maybe it'll be ok...
something along the lines:
```
.Other configuration variants (Click here):
[%collapsible]
====
`a.b.c`, `d.f.g` ... 
====
```

build/configuration-properties-collector/src/main/java/org/hibernate/search/configuration/properties/collector/impl/AsciiDocWriter.java
```diff
@@ -0,0 +1,100 @@
...
+			// nothing to write - return fast.
+			return;
+		}
+		try {
+			moduleName.ifPresent( name -> tryToWriteLine( "== ", name, writer ) );
+			writer.write( '\n' );
+			for ( Map.Entry<String, ConfigurationProperty> entry : entries ) {
+				Iterator<String> keys = entry.getValue().key().resolvedKeys().iterator();
+				writer.write( '`' );
+				writer.write( keys.next() );
```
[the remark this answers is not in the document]
marko-bekhta: ok, now I see why I've didn't add anchors the first time 😄 
the way it generates for the list looks like:
```html
<dt class="hdlist1">
    <a id="configuration-properties-aggregated-hibernate.search.backend.multi_tenancy.strategy"></a> 
    <code>hibernate.search.backend.multi_tenancy.strategy</code>
</dt>
```
so the link is there and we could point to it but there's no visual for it 😃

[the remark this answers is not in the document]
yrodiere: > so the link is there and we could point to it but there's no visual for it

That's fine, we can use F12 if necessary. I mean you can add a visual if you feel it's worth the trouble, but an anchor without a visual is already enough as far as I'm concerned :)

build/configuration-properties-collector/src/main/java/org/hibernate/search/configuration/properties/collector/impl/AsciiDocWriter.java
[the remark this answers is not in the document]
marko-bekhta: it looks like this:
image
Collapsed and expended variants. Seems ok, except for the label 😄 - the "click here" part is so-so, but I'm not sure if we remove it that it would be obvious for the users to click it (even though it has a hand cursor on hover)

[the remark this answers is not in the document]
yrodiere: LGTM, but I'd use `Variants of this configuration property (Click here):`. YMMV :)

build/configuration-properties-collector-processor/src/main/java/org/hibernate/search/configuration/properties/collector/impl/AnnotationUtils.java
```diff
@@ -0,0 +1,64 @@
...
+
+		private AnnotationAttributeHolder(AnnotationMirror annotationMirror) {
+			this.annotationMirror = annotationMirror;
+		}
+
+		@SuppressWarnings("unchecked")
+		public <T> Optional<T> attribute(String name, Class<T> klass) {
+			return annotationMirror.getElementValues().entrySet().stream()
+					.filter( entry -> entry.getKey().getSimpleName().contentEquals( name ) )
+					.map( entry -> ( (T) entry.getValue().getValue() ) )
```
yrodiere, line 50: I missed that earlier. Careful with unchecked casts, they don't do anything and you might end up returning an instance of a different type than `T`.

```suggestion
		public <T> Optional<T> attribute(String name, Class<T> klass) {
			return annotationMirror.getElementValues().entrySet().stream()
					.filter( entry -> entry.getKey().getSimpleName().contentEquals( name ) )
					.map( entry -> klass.cast( entry.getValue().getValue() ) )
```

build/configuration-properties-collector-processor/src/main/java/org/hibernate/search/configuration/properties/collector/impl/AnnotationUtils.java
```diff
@@ -0,0 +1,64 @@
...
+					.findAny();
+		}
+
+		@SuppressWarnings("unchecked")
+		public <T> Optional<List<T>> multiAttribute(String name, Class<T> klass) {
+			return annotationMirror.getElementValues().entrySet().stream()
+					.filter( entry -> entry.getKey().getSimpleName().contentEquals( name ) )
+					.map( entry -> ( (List<AnnotationValue>) entry.getValue().getValue() ) )
+					.map( list -> (List<T>) list.stream().map( AnnotationValue::getValue ).collect( Collectors.toList() ) )
```
yrodiere, line 59: Same here, unchecked casts are bad.

```suggestion
		public <T> Optional<List<T>> multiAttribute(String name, Class<T> klass) {
			return annotationMirror.getElementValues().entrySet().stream()
					.filter( entry -> entry.getKey().getSimpleName().contentEquals( name ) )
					.map( entry -> ( (List<AnnotationValue>) entry.getValue().getValue() ) )
					.map( list -> list.stream().map( AnnotationValue::getValue ).map( klass::cast ).collect( Collectors.toList() ) )
```

build/configuration-properties-collector-processor/src/main/java/org/hibernate/search/configuration/properties/collector/impl/AsciiDocWriter.java
```diff
@@ -0,0 +1,127 @@
...
+				.collect( Collectors.toList() );
+		if ( entries.isEmpty() ) {
+			// nothing to write - return fast.
+			return;
+		}
+
+		Map<String, Collection<ConfigurationProperty>> groups = entries.stream()
+				.map( Map.Entry::getValue )
+				.collect( Collectors.groupingBy( ConfigurationProperty::moduleName, TreeMap::new, Collectors.toCollection(
+						TreeSet::new ) ) );
```
yrodiere, line 44: I don't think there's a reason for the intermediary list? Or were you working around some compiler limitation?

```suggestion
		List<Map.Entry<String, ConfigurationProperty>> entries = propertyMap.entrySet().stream()
				.filter( filter )
				.map( Map.Entry::getValue )
				.collect( Collectors.groupingBy( ConfigurationProperty::moduleName, TreeMap::new, Collectors.toCollection(
						TreeSet::new ) ) );
		if ( groups.isEmpty() ) {
			// nothing to write - return fast.
			return;
		}
```

yrodiere (approved):
Thanks! The output looks perfect. I added a last few comments on the code, I'll let you judge if they're relevant.

In any case, feel free to merge!

PR #3335, opened 2022-12-14
HSEARCH-4050 Remove restriction on exists() predicate

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4050

Removing the restrictions seems to work fine. Need to check with Yoann to see what additional tests we should add.

### DIFF
--- a/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/ExistsPredicateObjectsSpecificsIT.java
+++ b/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/ExistsPredicateObjectsSpecificsIT.java
@@ -51,6 +51,12 @@ public class ExistsPredicateObjectsSpecificsIT {
 	// it will never be matched by an exists predicate
 	private static final String DOCUMENT_5 = "5";
 
+	// Document not from the main index. All fields are filled out
+	private static final String DOCUMENT_6 = "6";
+
+	// Document not from the main index. Only object fields are populated
+	private static final String DOCUMENT_7 = "7";
+
 	public static final String ANY_STRING = "Any String";
 	public static final int ANY_INTEGER = 173173;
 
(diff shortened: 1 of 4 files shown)


yrodiere (approved):
We could add more tests with different hierarchies, e.g.:

* an object field that exists in both indexes, but has a child object field in one index and not the other.
* an object field that exists in both indexes, and has a child object field in both indexes, but with different child value fields (different names).
* etc.

But frankly... I'm not sure it's worth the effort. Feel free to merge as-is!

marko-bekhta:
I've added two more index bindings and removed `assumeFullMultiIndexCompatibilityCheck();` from those modified tests. They seem to work fine for both Lucene and elasticsearch, so I wasn't sure if we needed to keep the assumption or not ..

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/ExistsPredicateObjectsSpecificsIT.java
```diff
@@ -159,19 +165,19 @@ public void nested_multiIndexes_wrongStructure() {
...
+				.where( f -> f.exists().field( fieldPath ) ) )
+				.hasDocRefHitsAnyOrder( mainIndex.typeName(), DOCUMENT_3 );
+
+		assertThatQuery( mainIndex.createScope( noInnerNestedField ).query()
+				.where( f -> f.exists().field( fieldPath ) ) )
+				.hasDocRefHitsAnyOrder( mainIndex.typeName(), DOCUMENT_3 );
+
+		assertThatQuery( mainIndex.createScope( differentInnerNestedField ).query()
+				.where( f -> f.exists().field( fieldPath ) ) )
+				.hasDocRefHitsAnyOrder( mainIndex.typeName(), DOCUMENT_3 );
```
yrodiere, line 176: These are nice, but in order to test if this really works, we need at least one matching and one non-matching document in each index.

yrodiere (commented):
I spotted something weird (not related to your work), so I pushed an additional commit to fix this.

Regarding your tests, I added a comment below.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/ExistsPredicateObjectsSpecificsIT.java
marko-bekhta: That's what I thought 😃, but then looked at the previous tests and was hesitant as those were not adding any documents to "other" indices

yrodiere: Right, that's because there's no point adding documents to other indices if search will fail and the documents won't be returned anyway :)

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/assertion/SearchHitsAssert.java
```diff
@@ -103,6 +103,12 @@ public SearchHitsAssert<H> hasDocRefHitsAnyOrder(String typeName, String firstId
 		} );
 	}
 
+	public SearchHitsAssert<H> containsListHitsAnyOrder(String typeName, String firstId, String... otherIds) {
+		return containsListHitsAnyOrder( ctx -> {
+			ctx.doc( typeName, firstId, otherIds );
+		} );
+	}
+
```
yrodiere, line 111: Let's avoid these "incomplete" checks where extra, unwanted matches could go unnoticed... Could you remove these `contains*` methods you added, and use something like this in your checks instead? This allows checking for documents from multiple indexes in one assertion.

https://github.com/hibernate/hibernate-search/blob/6199847ddb9a404df03995d244bff758a18dce56/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/AbstractPredicateTypeCheckingAndConversionIT.java#L232-L235

yrodiere (approved):
Merging, thanks!

PR #3340, opened 2022-12-19
HSEARCH-4763 Replace usage of URL constructors with URI where possible

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4763

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/environment/classpath/spi/DefaultResourceResolver.java
+++ b/engine/src/main/java/org/hibernate/search/engine/environment/classpath/spi/DefaultResourceResolver.java
@@ -42,7 +42,10 @@ public final class DefaultResourceResolver implements ResourceResolver {
 
 		if ( stripped != null ) {
 			try {
-				return new URL( stripped ).openStream();
+				@SuppressWarnings("deprecation")
+				// TODO: HSEARCH-4765 address the URL -> URI constructor change once the URLClassLoader stops using the URL constructor
+				InputStream resourceStream = new URL( stripped ).openStream();
+				return resourceStream;
 			}
 			catch (Exception ignore) {
 				// Ignore
--- a/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/builtin/impl/DefaultJavaNetURLBridge.java
+++ b/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/builtin/impl/DefaultJavaNetURLBridge.java
@@ -8,6 +8,8 @@ package org.hibernate.search.mapper.pojo.bridge.builtin.impl;
 
 import java.lang.invoke.MethodHandles;
 import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
 
 import org.hibernate.search.mapper.pojo.logging.impl.Log;
@@ -24,15 +26,20 @@ public final class DefaultJavaNetURLBridge extends AbstractStringBasedDefaultBri
 
 	@Override
 	protected String toString(URL value) {
-		return value.toExternalForm();
+		try {
+			return value.toURI().toString();
+		}
+		catch (URISyntaxException e) {
+			throw log.badURISyntax( value.toString(), e );
+		}
 	}
 
 	@Override
 	protected URL fromString(String value) {
 		try {
-			return new URL( value );
+			return new URI( value ).toURL();
 		}
-		catch (MalformedURLException e) {
+		catch (MalformedURLException | URISyntaxException e) {
 			throw log.malformedURL( value, e );
 		}
 	}
--- a/util/common/src/test/java/org/hibernate/search/util/common/jar/impl/CodeSourceTest.java
+++ b/util/common/src/test/java/org/hibernate/search/util/common/jar/impl/CodeSourceTest.java
@@ -13,6 +13,7 @@ import static org.hibernate.search.util.impl.test.jar.JarTestUtils.toJar;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URI;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.nio.charset.StandardCharsets;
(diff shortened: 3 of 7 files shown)


mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/builtin/impl/DefaultJavaNetURLBridge.java
```diff
@@ -30,9 +32,9 @@ protected String toString(URL value) {
 	@Override
 	protected URL fromString(String value) {
 		try {
-			return new URL( value );
+			return new URI( value ).toURL();
```
yrodiere, line 35: I'm worried this could result in different behavior, since apparently `new URI` handles URL escapes while `new URL` does not.

On the other hand, it seems like the right thing to do, and I doubt many people index URL anyway.

Please mention this change in the migration guide?

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/builtin/impl/DefaultJavaNetURLBridge.java
```diff
@@ -30,9 +32,9 @@ protected String toString(URL value) {
 	@Override
 	protected URL fromString(String value) {
```
yrodiere, line 33: If you update `fromString`, maybe you should also update `toString` (using `toURI().toString()` instead of `toExternalForm()`). Mainly because of how URL/URI handle escapes differently.

Perhaps that's not necessary, though. In any case, could you add a few values to `JavaNetURLPropertyTypeDescriptor` to test URL escaping? I would also include an URL with an URL escape that has been created with `new URL`, just in case, because users could do that.

engine/src/main/java/org/hibernate/search/engine/environment/classpath/spi/DefaultResourceResolver.java
```diff
@@ -42,7 +42,7 @@ public InputStream locateResourceStream(String name) {
 
 		if ( stripped != null ) {
 			try {
-				return new URL( stripped ).openStream();
+				return new URI( stripped ).toURL().openStream();
```
yrodiere, line 45: Can you please check this will not change the behavior when "name" contains a URL escape such as `%20`? I know it's unlikely, but I don't want to mess with this rather... arcane code.

util/common/src/test/java/org/hibernate/search/util/common/jar/impl/CodeSourceTest.java
```diff
@@ -121,7 +122,7 @@ public void jar_jarScheme_classesInRoot() throws Exception {
 		} );
 
 		URL fileURL = jarPath.toUri().toURL();
```
yrodiere, line 124: Might as well use an URI here, and `fileURI.toString()` instead of `fileURL.toExternalForm()` below?

util/common/src/test/java/org/hibernate/search/util/common/jar/impl/CodeSourceTest.java
```diff
@@ -263,7 +268,7 @@ public void jar_jarScheme_specialCharacter() throws Exception {
 		);
 
 		URL fileURL = jarPath.toUri().toURL();
```
yrodiere, line 270: Might as well use an URI here, and `fileURI.toString()` instead of `fileURL.toExternalForm()` below?

yrodiere (changes requested):
Thanks. A few comments below. I think we need more tests with URL escapes, at the very least.

engine/src/main/java/org/hibernate/search/engine/environment/classpath/spi/DefaultResourceResolver.java
marko-bekhta: So it seems that escaped characters won't cause a problem, but not escaped, on the other hand, would end up throwing an exception rather than getting escaped as before. I have tried it with the next test:
```java
Path path = folder.getRoot().toPath();

Path path1 = Files.createFile( path.resolve( "file%201" ) );
Path path2 = Files.createFile( path.resolve( "file 1" ) );

Path path = folder.getRoot().toPath();

Path path1 = Files.createFile( path.resolve( "file%201" ) );
Path path2 = Files.createFile( path.resolve( "file 1" ) );

try ( FileWriter fileWriter = new FileWriter( path1.toFile() ) ) {
	fileWriter.write( "file 1" );
}
try ( FileWriter fileWriter = new FileWriter( path2.toFile() ) ) {
	fileWriter.write( "file 2" );
}

// name with `%20` result in a `file 1` being opened
try ( Scanner scanner = new Scanner( new URL( "file:" + path1 ).openStream() ) ) {
	String data = scanner.nextLine();
	assertThat( data ).isEqualTo( "file 2" );
}

// name with a space ` ` result in a `file 1` being opened
try ( Scanner scanner = new Scanner( new URL( "file:" + path2 ).openStream() ) ) {
	String data = scanner.nextLine();
	assertThat( data ).isEqualTo( "file 2" );
}

// name with %20 result in a `file 1` being opened
try ( Scanner scanner = new Scanner( new URI( "file:" + path1 ).toURL().openStream() ) ) {
	String data = scanner.nextLine();
	assertThat( data ).isEqualTo( "file 2" );
}

// name with a space ` ` result in an exception
assertThatThrownBy( () -> {
		try ( Scanner scanner = new Scanner( new URI( "file:" + path2 ).toURL().openStream() ) ) {
			String data = scanner.nextLine();
			assertThat( data ).isEqualTo( "file 2" );
		}
	}
).isInstanceOf( URISyntaxException.class );
```

marko-bekhta: That seems like a good idea, especially when in JDK 20:
```java
URL url;
try {
    @SuppressWarnings("deprecation")
    var _unused = url = new URL(base, ParseUtil.encodePath(name, false));
} catch (MalformedURLException e) {
    return null;
}
```
🙈 😃

marko-bekhta: Created https://hibernate.atlassian.net/browse/HSEARCH-4765 and added reference to it in TODOs for future selves' 😄

yrodiere (approved):

PR #3342, opened 2022-12-20
HSEARCH-4294 Expose implicit fields through search DSL

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4294

If I understood correctly, we don't want to make any changes to Lucene's backend. But just in case, I've placed the main interfaces under engine SPI, as that's where the other model builder-related interfaces are.
As for the implementation itself - because the fields are already present in the schema in one way or another, it seems that the only thing to do was to let HSEARCH know about their existence and adding them to static fields does the trick.

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/dsl/impl/ElasticsearchStringImplicitFieldContributor.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/dsl/impl/ElasticsearchStringImplicitFieldContributor.java
@@ -1,0 +1,48 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.elasticsearch.document.model.dsl.impl;
+
+import org.hibernate.search.engine.backend.document.model.dsl.spi.ImplicitFieldCollector;
+import org.hibernate.search.engine.backend.document.model.dsl.spi.ImplicitFieldContributor;
+import org.hibernate.search.engine.backend.types.Aggregable;
+import org.hibernate.search.engine.backend.types.Projectable;
+import org.hibernate.search.engine.backend.types.Searchable;
+import org.hibernate.search.engine.backend.types.Sortable;
+
+public class ElasticsearchStringImplicitFieldContributor implements ImplicitFieldContributor {
+
+	private final String name;
+	private final Searchable searchable;
+	private final Sortable sortable;
+	private final Projectable projectable;
+	private final Aggregable aggregable;
+
+	public ElasticsearchStringImplicitFieldContributor(String name) {
+		this( name, Searchable.YES, Sortable.YES, Projectable.YES, Aggregable.YES );
+	}
+
+	public ElasticsearchStringImplicitFieldContributor(String name, Searchable searchable, Sortable sortable,
+			Projectable projectable, Aggregable aggregable) {
+		this.name = name;
+		this.searchable = searchable;
+		this.sortable = sortable;
+		this.projectable = projectable;
+		this.aggregable = aggregable;
+	}
+
+	public void contribute(ImplicitFieldCollector collector) {
+		collector.addImplicitField(
+				name,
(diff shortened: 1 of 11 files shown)


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/dsl/impl/ElasticsearchStringImplicitFieldContributor.java
```diff
@@ -0,0 +1,83 @@
...
+		typeBuilder.projectable( true );
+		typeBuilder.queryElementFactory(
+				ProjectionTypeKeys.FIELD, new ElasticsearchFieldProjection.Factory<>( typeBuilder.codec() ) );
+
+		typeBuilder.aggregable( true );
+		typeBuilder.queryElementFactory(
+				AggregationTypeKeys.TERMS, new ElasticsearchTermsAggregation.Factory<>( typeBuilder.codec() ) );
+
+		return typeBuilder;
+	}
```
yrodiere, line 82: This looks like code duplication, as you're manually definining which predicates/projections/etc are available, while that's generally inferred from the type + flags (searchable/projectable/etc.). As types evolve, we're likely to forget to update this code.

Can'y you pass the type factory to implicit field contributors instead, so that they use the API to build a type?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/mapping/impl/TypeNameMapping.java
```diff
@@ -31,6 +32,8 @@ public interface TypeNameMapping {
 	 */
 	Optional<DocumentMetadataContributor> getDocumentMetadataContributor(String mappedTypeName);
 
+	Optional<ImplicitFieldContributor> getImplicitFieldContributor();
```
yrodiere, line 35: Maybe some javadoc, like the other methods? :)

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/ElasticsearchExtensionIT.java
```diff
@@ -958,6 +958,43 @@ public void aggregation_nativeField() {
 				);
 	}
 
+	@Test
+	public void implicit_fields_aggregation_entity_type() {
```
yrodiere, line 962: These tests don't rely on `ElasticsearchExtension`, they would probably deserve their own, separate test class.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/metamodel/IndexDescriptorIT.java
```diff
@@ -68,11 +71,28 @@ public void staticFields() {
...
+			Optional<IndexFieldDescriptor> entityTypeDescriptorOptional = indexDescriptor.field( "_entity_type" );
+			assertThat( entityTypeDescriptorOptional ).isPresent();
+
+			descriptors.add( idDescriptorOptional.get() );
+			descriptors.add( indexDescriptorOptional.get() );
+			descriptors.add( entityTypeDescriptorOptional.get() );
+		}
+
+		descriptors.add( valueFieldDescriptorOptional.get() );
+		descriptors.add( objectFieldDescriptorOptional.get() );
```
yrodiere, line 92: I would move this to an Elasticsearch-specific test, since this is obviously specific to Elaticsearch: even if another, hypothetical backend supported the "implicit fields" feature, the fields would be different.

As to the assertion below, you can simply relax it: use `contains` instead of `containsExactlyInAnyOrder`?

yrodiere (changes requested):
Well, it seems like this works beautifully :) I just have a few suggestions for improvements.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/document/model/dsl/impl/ElasticsearchStringImplicitFieldContributor.java
marko-bekhta: Ok, should be a bit cleaner now 😃

yrodiere (approved):
LGTM, but the build failed because of checkstyle violations. Can you fix that and merge? Thanks!

PR #3345, opened 2022-12-22
HSEARCH-4645 Syntactic sugar for negating predicates (NOT)

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4645

I wasn't sure about the "simplification" part:
> optimize things such as f.bool().must(f.not(something)).must(somethingElse) into f.bool().mustNot(something).must(somethingElse), if possible. Or maybe we could do it ourselves.

so I've pushed it as a separate commit. It does reduce the nested predicates, but it looks so-so 😄 

[HSEARCH-4645]: https://hibernate.atlassian.net/browse/HSEARCH-4645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchBooleanPredicate.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchBooleanPredicate.java
@@ -12,6 +12,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
+import java.util.function.Consumer;
 
 import org.hibernate.search.backend.elasticsearch.gson.impl.GsonUtils;
 import org.hibernate.search.backend.elasticsearch.gson.impl.JsonAccessor;
@@ -237,33 +238,20 @@ class ElasticsearchBooleanPredicate extends AbstractElasticsearchPredicate {
 				return new ElasticsearchMatchNonePredicate( this );
 			}
 
-			if ( mustClauses != null ) {
-				Iterator<ElasticsearchSearchPredicate> mustIterator = mustClauses.iterator();
-				while ( mustIterator.hasNext() ) {
-					ElasticsearchSearchPredicate must = mustIterator.next();
-					if ( must instanceof ElasticsearchBooleanPredicate
-							&& ( (ElasticsearchBooleanPredicate) must ).hasOnlyOneMustNotClause()
-							&& ( (ElasticsearchBooleanPredicate) must ).hasNoModifiers()
-					) {
-						mustIterator.remove();
-						mustNot( ( (ElasticsearchBooleanPredicate) must ).mustNotClauses.get( 0 ) );
-					}
-				}
-				if ( mustClauses.isEmpty() ) {
-					mustClauses = null;
-				}
-			}
+			optimizeClauseCollection(
+					mustClauses,
+					this::mustNot
+			);
+
+			optimizeClauseCollection(
+					mustNotClauses,
+					this::must
+			);
+
+			checkAndClearClauseCollections();
 
 			if ( hasNoModifiers() ) {
-				if ( hasOnlyOneMustNotClause() ) {
-					ElasticsearchSearchPredicate mustNot = mustNotClauses.get( 0 );
-					if ( mustNot instanceof ElasticsearchBooleanPredicate
-							&& ( (ElasticsearchBooleanPredicate) mustNot ).hasOnlyOneMustNotClause()
-							&& ( (ElasticsearchBooleanPredicate) mustNot ).hasNoModifiers() ) {
-						return ( (ElasticsearchBooleanPredicate) mustNot ).mustNotClauses.get( 0 );
-					}
-				}
-				else if ( hasOnlyOneMustClause() ) {
+				if ( hasOnlyOneMustClause() ) {
 					return mustClauses.get( 0 );
 				}
 				else if ( hasOnlyOneShouldClause() ) {
@@ -279,15 +267,34 @@ class ElasticsearchBooleanPredicate extends AbstractElasticsearchPredicate {
 			return new ElasticsearchBooleanPredicate( this );
 		}
 
-		private boolean hasAtLeastOneMustOrFilterPredicate() {
-			return mustClauses != null || filterClauses != null;
+		private void optimizeClauseCollection(List<ElasticsearchSearchPredicate> collection,
+				Consumer<ElasticsearchSearchPredicate> newCollection) {
+			if ( collection != null ) {
+				Iterator<ElasticsearchSearchPredicate> iterator = collection.iterator();
+				while ( iterator.hasNext() ) {
+					ElasticsearchSearchPredicate clause = iterator.next();
+					if ( clause instanceof ElasticsearchBooleanPredicate
+							&& ( (ElasticsearchBooleanPredicate) clause ).hasOnlyOneMustNotClause()
+							&& ( (ElasticsearchBooleanPredicate) clause ).hasNoModifiers()
+					) {
+						iterator.remove();
+						newCollection.accept( ( (ElasticsearchBooleanPredicate) clause ).mustNotClauses.get( 0 ) );
+					}
+				}
+			}
 		}
 
-		private boolean hasOnlyOneMustNotClause() {
-			return mustNotClauses != null && mustNotClauses.size() == 1
-					&& ( mustClauses == null || mustClauses.isEmpty() )
-					&& ( shouldClauses == null || shouldClauses.isEmpty() )
-					&& ( filterClauses == null || filterClauses.isEmpty() );
+		private void checkAndClearClauseCollections() {
+			if ( mustClauses != null && mustClauses.isEmpty() ) {
+				mustClauses = null;
+			}
+			if ( mustNotClauses != null && mustNotClauses.isEmpty() ) {
+				mustNotClauses = null;
+			}
+		}
+
+		private boolean hasAtLeastOneMustOrFilterPredicate() {
+			return mustClauses != null || filterClauses != null;
 		}
 
 		private boolean hasOnlyOneMustClause() {
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchBooleanPredicate.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchBooleanPredicate.java
@@ -136,6 +136,23 @@ class ElasticsearchBooleanPredicate extends AbstractElasticsearchPredicate {
 		return builder.toString();
 	}
 
+	private boolean isOnlyMustNot() {
+		return mustNotClauses != null && !mustNotClauses.isEmpty()
+				&& ( mustClauses == null || mustClauses.isEmpty() )
+				&& ( shouldClauses == null || shouldClauses.isEmpty() )
+				&& ( filterClauses == null || filterClauses.isEmpty() );
+	}
+
+	private boolean hasOnlyOneMustNotClause() {
+		return isOnlyMustNot() && mustNotClauses.size() == 1;
+	}
+
+	@Override
+	protected boolean hasNoModifiers() {
+		return minimumShouldMatchConstraints == null
+				&& super.hasNoModifiers();
+	}
+
 	static class Builder extends AbstractElasticsearchPredicate.AbstractBuilder implements BooleanPredicateBuilder {
 		private List<ElasticsearchSearchPredicate> mustClauses;
 		private List<ElasticsearchSearchPredicate> mustNotClauses;
@@ -215,9 +232,43 @@ class ElasticsearchBooleanPredicate extends AbstractElasticsearchPredicate {
 
 		@Override
 		public SearchPredicate build() {
-			if ( mustClauses == null && shouldClauses == null && mustNotClauses == null && filterClauses == null ) {
+			if ( !hasClause() ) {
 				// HSEARCH-4619: a boolean predicate without any clause must not match anything.
 				return new ElasticsearchMatchNonePredicate( this );
+			}
+
+			if ( mustClauses != null ) {
+				Iterator<ElasticsearchSearchPredicate> mustIterator = mustClauses.iterator();
+				while ( mustIterator.hasNext() ) {
+					ElasticsearchSearchPredicate must = mustIterator.next();
+					if ( must instanceof ElasticsearchBooleanPredicate
+							&& ( (ElasticsearchBooleanPredicate) must ).hasOnlyOneMustNotClause()
+							&& ( (ElasticsearchBooleanPredicate) must ).hasNoModifiers()
+					) {
+						mustIterator.remove();
+						mustNot( ( (ElasticsearchBooleanPredicate) must ).mustNotClauses.get( 0 ) );
+					}
+				}
+				if ( mustClauses.isEmpty() ) {
+					mustClauses = null;
+				}
+			}
+
+			if ( hasNoModifiers() ) {
+				if ( hasOnlyOneMustNotClause() ) {
+					ElasticsearchSearchPredicate mustNot = mustNotClauses.get( 0 );
+					if ( mustNot instanceof ElasticsearchBooleanPredicate
+							&& ( (ElasticsearchBooleanPredicate) mustNot ).hasOnlyOneMustNotClause()
+							&& ( (ElasticsearchBooleanPredicate) mustNot ).hasNoModifiers() ) {
+						return ( (ElasticsearchBooleanPredicate) mustNot ).mustNotClauses.get( 0 );
+					}
+				}
+				else if ( hasOnlyOneMustClause() ) {
+					return mustClauses.get( 0 );
+				}
+				else if ( hasOnlyOneShouldClause() ) {
+					return shouldClauses.get( 0 );
+				}
 			}
 
 			// Forcing to Lucene's defaults. See HSEARCH-3534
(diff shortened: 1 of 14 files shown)


engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/impl/NotPredicateFinalStepImpl.java
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	public NotPredicateFinalStepImpl(SearchPredicateDslContext<?> dslContext, PredicateFinalStep searchPredicate) {
+		this( dslContext, searchPredicate.toPredicate() );
+	}
+
+	@Override
+	protected SearchPredicate build() {
+		BooleanPredicateBuilder builder = dslContext.scope().predicateBuilders().bool();
+		builder.mustNot( searchPredicate );
```
yrodiere, line 34: Better move that to the constructor, so that failures in `mustNot` (e.g. passing `null`) are reported as early as possible. You'll just store the builder instead of `searchPredicate`.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/NotPredicateSpecificsIT.java
```diff
@@ -0,0 +1,152 @@
...
+				.hasNoHits();
+	}
+
+	@Test
+	public void matchAll() {
+		assertThatQuery( index.query()
+				.where( f -> f.not( f.matchNone() ) ) )
+				.hasTotalHitCount( 3 )
+				.hasDocRefHitsAnyOrder( index.typeName(), DOCUMENT_1, DOCUMENT_2, DOCUMENT_3 );
+	}
```
yrodiere, line 78: Did you reverse the method names on purpose? `notMatchNone`/`notMatchAll`, while ugly, would be less confusing :)

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/NotPredicateSpecificsIT.java
```diff
@@ -0,0 +1,152 @@
...
+	@Test
+	public void mustWithNotInside() {
+		assertThatQuery( index.query()
+				.where( f -> f.bool().must( f.not( f.match().field( "field2" ).matching( FIELD2_VALUE1 ) ) ) ) )
+				.hasTotalHitCount( 2 )
+				.hasDocRefHitsAnyOrder( index.typeName(), DOCUMENT_2, DOCUMENT_3 );
+	}
+
+	@Test
+	public void tooMuchNestedNot() {
```
yrodiere, line 97: Technically that's not too much, since it works.

```suggestion
	public void manyNestedNot() {
```

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/NotPredicateSpecificsIT.java
```diff
@@ -0,0 +1,152 @@
...
+import org.hibernate.search.engine.search.predicate.dsl.SearchPredicateFactory;
+import org.hibernate.search.integrationtest.backend.tck.testsupport.util.rule.SearchSetupHelper;
+import org.hibernate.search.util.impl.integrationtest.mapper.stub.BulkIndexer;
+import org.hibernate.search.util.impl.integrationtest.mapper.stub.SimpleMappedIndex;
+
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+public class NotPredicateSpecificsIT {
```
yrodiere, line 23: That's for the `not`-specific ITs, but you should also add something to cover base features. In this case I think there's only the score to worry about. See `OrPredicateBaseIT` for example (though you'll only need one nested test for the `not` predicate).

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/NotPredicateSpecificsIT.java
```diff
@@ -0,0 +1,152 @@
...
+		BooleanPredicateClausesStep<?> step = f.bool().must( f.not( f.not( f.not( f.not( f.not( f.not( f.not( f.match().field( "field2" ).matching( FIELD2_VALUE1 ) ) ) ) ) ) ) ) );
+		assertThatQuery( index.query()
+				.where( step.toPredicate() ) )
+				.hasTotalHitCount( 2 )
+				.hasDocRefHitsAnyOrder( index.typeName(), DOCUMENT_2, DOCUMENT_3 );
+
+		assertThatQuery( index.query()
+				.where( f.not( ( step ) ).toPredicate() ) )
+				.hasTotalHitCount( 1 )
+				.hasDocRefHitsAnyOrder( index.typeName(), DOCUMENT_1 );
```
yrodiere, line 108: Could you also add some tests that check optimizations? Of course these would be backend-specific. You can check the content of a `SearchQuery` with `SearchQuery.queryString()`.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/NotPredicateSpecificsIT.java
marko-bekhta: haha yeah, applied the "bool logic" to the names 😃. will update it

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/NotPredicateSpecificsIT.java
marko-bekhta: Got it, will do 😃

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchBooleanPredicate.java
```diff
@@ -225,6 +225,43 @@ public SearchPredicate build() {
...
+
+					if ( lucenePredicate.filterClauses == null && lucenePredicate.mustNotClauses == null && lucenePredicate.shouldClauses == null && lucenePredicate.mustClauses != null && lucenePredicate.mustClauses.size() == 1 ) {
+						ElasticsearchSearchPredicate innerPredicate = lucenePredicate.mustClauses.iterator().next();
+						Builder builder = new Builder( scope );
+						builder.mustNot( innerPredicate );
+						return new ElasticsearchBooleanPredicate( builder );
+					}
+				}
+			}
+
```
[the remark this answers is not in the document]
yrodiere: @marko-bekhta I updated my code example, which contained some nonsensical leftover from a previous iteration of my comment :)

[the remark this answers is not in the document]
yrodiere: By the way, this is related: https://hibernate.atlassian.net/browse/HSEARCH-4679

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/NotPredicateSpecificsIT.java
marko-bekhta: right ... I didn't have a `NotPredicateBaseIT` as I made this not predicate to only have a final step, so there's no way to set a boost or anything. At that point, I was looking at this `not` as something only to negate a predicate. But I suppose, based on this, and your other comment from above, `not` should allow setting the boost/score too? Or I'm misinterpreting 😟 😃

yrodiere: > I made this not predicate to only have a final step, so there's no way to set a boost or anything

Ah, I didn't notice that.

I think that a `not` predicate will always have a score of `1` when it matches, because then the inner predicate *did not* match and thus the score of the inner predicate is irrelevant. So, `boost`, at least, could make sense, to change that score of `1` to something else. `constantScore` isn't very useful here, but it's just simpler to also expose it.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/SearchPredicateFactory.java
```diff
@@ -132,6 +132,28 @@ SimpleBooleanPredicateOptionsStep<?> or(SearchPredicate firstSearchPredicate,
...
+	 * @return The initial and final step of a DSL where the "not" predicate can be defined.
+	 * @see NotPredicateFinalStep
+	 */
+	NotPredicateFinalStep not(SearchPredicate searchPredicate);
+
+	/**
+	 * Match documents that do not satisfy the passed in predicate.
+	 * <p>
+	 * Can be used to negate a predicate. Short version of bool query with a single {@code mustNot} clause:
+	 * {@code factory -> factory.bool().mustNot( predicate )}.
```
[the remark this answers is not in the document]
yrodiere: I think you ignored this comment, was that on purpose?

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/NotPredicateSpecificsIT.java
```diff
@@ -0,0 +1,179 @@
...
+				.hasTotalHitCount( 2 )
+				.hasDocRefHitsAnyOrder( index.typeName(), DOCUMENT_2, DOCUMENT_3 );
+	}
+
+	@Test
+	public void manyNestedNot() {
+		SearchPredicateFactory f = index.createScope().predicate();
+		BooleanPredicateClausesStep<?> step = f.bool().must( f.not( f.not( f.not( f.not( f.not( f.not( f.not( f.match().field( "field2" ).matching( FIELD2_VALUE1 ) ) ) ) ) ) ) ) );
+
+		// as query strings are backend specific let's check that generated string is among expected.
```
yrodiere, line 107: Hmpf. Are you sure about this? I was thinking of creating dedicated test classes in `integrationtest/backend/lucene` and `integrationtest/backend/elasticsearch`, like we did for other backend-specific assertions...

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/NotPredicateSpecificsIT.java
```diff
@@ -0,0 +1,179 @@
...
+		assertThat(
+				Arrays.asList(
+						"+field2:[3 TO 3]", // Lucene query
+						"{\"query\":{\"match\":{\"field2\":{\"query\":3}}},\"_source\":false}" // Elasticsearch query
+				)
+		).contains( query.toQuery().queryString() );
+
+		assertThatQuery( query )
+				.hasTotalHitCount( 1 )
+				.hasDocRefHitsAnyOrder( index.typeName(), DOCUMENT_1 );
```
yrodiere, line 135: One optimization that you didn't test is the one mentioned in the ticket description:

> optimize things such as `f.bool().must(f.not(something)).must(somethingElse)` into `f.bool().mustNot(something).must(somethingElse)`

That's what implemented by the first loop turning `must` clauses into `mustNot` clauses in the predicate builders.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/SearchPredicateFactory.java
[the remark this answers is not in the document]
marko-bekhta: 😞 ohh there's another `not` that takes predicate rather than predicate step, removed from one and the other left unchanged

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/NotPredicateBaseIT.java
```diff
@@ -0,0 +1,116 @@
...
+		private static final StubMappedIndex index = StubMappedIndex.withoutFields().name( "score" );
+
+		public ScoreIT() {
+			super( index, dataSet );
+		}
+
+
+		@Override
+		public void predicateLevelBoost() {
+			// ignore
```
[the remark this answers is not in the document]
marko-bekhta: good question 😃. with ES this test ultimately generates a query (I've removed routing etc..):
```json
{
  "query": {
    "bool": {
      "must": {
        "bool": {
          "should": [
            {
              "bool": {
                "must_not": {
                  "ids": {
                    "values": [
                      "singleRoutingKey_doc_1",
                      "singleRoutingKey_doc_2"
                    ]
                  }
                }
              }
            },
            {
              "bool": {
                "boost": 7.0,
                "must_not": {
                  "ids": {
                    "values": [
                      "singleRoutingKey_doc_0",
                      "singleRoutingKey_doc_2"
                    ]
                  }
                }
              }
            }
          ]
        }
      }
    }
  }
}
```
but the results have 0 score:
```json
{
  "total": {
    "value": 2,
    "relation": "eq"
  },
  "max_score": 0.0,
  "hits": [
    {
      "_index": "score-000001",
      "_id": "singleRoutingKey_doc_1",
      "_score": 0.0,
      "_routing": "singleRoutingKey",
      "_source": {
        "_entity_type": "scoreType"
      }
    },
    {
      "_index": "score-000001",
      "_id": "singleRoutingKey_doc_0",
      "_score": 0.0,
      "_routing": "singleRoutingKey",
      "_source": {
        "_entity_type": "scoreType"
      }
    }
  ]
}
```
So I'm not sure how we could make use of a combo of `not` and `boost` for this test to pass ...  Same is with Lucene.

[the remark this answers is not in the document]
yrodiere: Sooo... are you telling me that the score is always 0, thus `constantScore` and `boost` are pointless? If that's the case, that changes everything: better remove those features from the `not` predicate altogether.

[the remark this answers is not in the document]
marko-bekhta: I guess we could add a `must` so that the scoring is enabled. Then the query would become something like this:
```json
{
  "query": {
    "bool": {
      "must": {
        "bool": {
          "should": [
            {
              "bool": {
                "must": {
                  "match_all": {}
                },
                "must_not": {
                  "ids": {
                    "values": [
                      "singleRoutingKey_doc_1",
                      "singleRoutingKey_doc_2"
                    ]
                  }
                }
              }
            },
            {
              "bool": {
                "boost": 7.0,
                "must": {
                  "match_all": {}
                },
                "must_not": {
                  "ids": {
                    "values": [
                      "singleRoutingKey_doc_0",
                      "singleRoutingKey_doc_2"
                    ]
                  }
                }
              }
            }
          ]
        }
      }
    }
  }
}
```
but then the "optimizations" would become a bit more cumbersome 😃. I guess ... if this `not` is intended for those who aren't familiar with `bool` would it be safe to assume that they wouldn't need the boost anyway? (I'll give it a try with the `match_all` in the meantime).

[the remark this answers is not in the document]
yrodiere: > We can add the implicit `match_all` _after_ the optimizations, so that would be simpler.

By that, I meant much after, when we generate JSON.

marko-bekhta:
Ok, I've made some more changes... The added Lucene's implicit filter clause still won't produce the score, so I've made it conditional if there's no boost - then we don't care about getting the score, and it remains a filter, but if there are modifiers, then we add this match all as must rather than a filter. 
I've also updated the optimization, as it wasn't including the case of a mix of must/mustNot that can be simplified in the same way 😃

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/NotPredicateBaseIT.java
```diff
@@ -0,0 +1,116 @@
...
+
+		public ScoreIT() {
+			super( index, dataSet );
+		}
+
+
+//		@Override
+//		public void predicateLevelBoost() {
+//			// ignore
+//		}
```
yrodiere, line 66:
```suggestion
```

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchBooleanPredicate.java
```diff
@@ -215,11 +250,36 @@ private void addMinimumShouldMatchConstraint(int ignoreConstraintCeiling,
...
-			if ( mustClauses == null && shouldClauses == null && mustNotClauses == null && filterClauses == null ) {
+			if ( !hasClause() ) {
 				// HSEARCH-4619: a boolean predicate without any clause must not match anything.
 				return new ElasticsearchMatchNonePredicate( this );
 			}
 
+			optimizeClauseCollection(
+					mustClauses,
+					ElasticsearchBooleanPredicate::hasOnlyOneMustNotClause,
+					clause -> clause.mustNotClauses.get( 0 ),
```
yrodiere, line 261: Looks like this argument is always the same and thus can be inlined?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/predicate/impl/LuceneBooleanPredicate.java
```diff
@@ -182,9 +196,80 @@ public boolean hasClause() {
 
 		@Override
 		public SearchPredicate build() {
+			optimizeClauseCollection(
+					mustClauses,
+					LuceneBooleanPredicate::hasOnlyOneMustNotClause,
+					clause -> clause.mustNotClauses.get( 0 ),
```
yrodiere, line 202: Same here, this argument is always the same.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/search/query/ElasticsearchBoolSearchPredicateIT.java
```diff
@@ -0,0 +1,211 @@
...
+
+		assertThat(
+				new Gson().fromJson(
+						index.query()
+								.where( f.bool().must( f.not( f.not( f.not( f.not( f.not( f.not( f.not( f.match().field( "fieldName" ).matching( "test" ) ) ) ) ) ) ) ) ).toPredicate() )
+								.toQuery()
+								.queryString(),
+						JsonObject.class
+				)
+		).isEqualTo( new Gson().fromJson(
```
yrodiere, line 48: Just so you know, there are utils to compare JSON in a lenient way, e.g.:

https://github.com/hibernate/hibernate-search/blob/fe3b77f29423afcdd6d103b51f1db6316dfd13c9/integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/schema/management/ElasticsearchIndexSchemaManagerCreationOrPreservationIT.java#L113-L118

yrodiere (commented):
Looking good :) Feel free to merge once you've seen/addressed my remaining comments. Thanks!

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/search/query/ElasticsearchBoolSearchPredicateIT.java
marko-bekhta: nice ! 😃

yrodiere (approved):
Works for me, but not for CI: looks like there's a checkstyle violation. Feel free to merge once that's fixed :)

PR #3352, opened 2023-01-02
HSEARCH-4721 Re-enable tests disabled because of HHH-15606

yrodiere:


yrodiere:
Looks like even after https://hibernate.atlassian.net/browse/HHH-15606 is fixed, there’s still a bug in our way: https://hibernate.atlassian.net/browse/HHH-15967. So we can’t re-enable tests at the moment.

PR #3355, opened 2023-01-03
HSEARCH-4753 Run tests against OpenSearch 2.3 AWS

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4753


yrodiere:
I'll need to run the Jenkins job with specific parameters in order to test that, but Jenkins is currently overburdened. Will do later.

yrodiere:
It worked, merging

PR #3356, opened 2023-01-04
HSEARCH-4773 Change how agents are stopped in massindexer

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4773

So this is a place where it starts to cause problems with the Oracle build... I've tried to keep the  `applyToAllContexts(..)` but that becomes ugly pretty fast:
```java
applyToAllContexts(
	context -> {
		try {
			Object result = Futures.unwrappedExceptionGet( context.agent().preStop() );
			context.agent().stop();
			return CompletableFuture.completedFuture( result );
		}
		catch (InterruptedException e) {
			Thread.currentThread().interrupt();
			throw new RuntimeException( e );
		}
	}
);
```

[HSEARCH-4773]: https://hibernate.atlassian.net/browse/HSEARCH-4773?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingBatchCoordinator.java
+++ b/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingBatchCoordinator.java
@@ -190,8 +190,12 @@ public class PojoMassIndexingBatchCoordinator extends PojoMassIndexingFailureHan
 		flushAndRefresh();
 		applyToAllContexts(
 				context -> context.agent().preStop()
-						.thenRun( () -> context.agent().stop() )
 		);
+		// NOTE: HSEARCH-4773 this loop was added here on purpose, as composing this stop() operation to the above future
+		// was causing an issue when running against Oracle DB. Doing it like this seems to allow a graceful stopping of the agents.
+		for ( SessionContext context : sessionContexts ) {
+			context.agent().stop();
+		}
 		sessionContexts.clear();
 	}
 


mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingBatchCoordinator.java
```diff
@@ -188,10 +188,13 @@ private void afterBatch() throws InterruptedException {
 			);
 		}
 		flushAndRefresh();
-		applyToAllContexts(
-				context -> context.agent().preStop()
-						.thenRun( () -> context.agent().stop() )
-		);
+		// Not using the `applyToAllContexts(..)` here as we want to make sure that stop() is executed right after preStop()
+		// completes and before going to another agent.
+		// Otherwise, it might cause unexpected failures when running with certain DBs (Oracle), see HSEARCH-4773
```
yrodiere, line 193: Is the problem related to locks, i.e. we're concurrently changing the same table from multiple connections and Oracle returns a "concurrent change" error?

If so, maybe we should rather try to avoid problematic "read then update" patterns such as this one:

https://github.com/hibernate/hibernate-search/blob/75d0dcdbeff51067c600837202b2b01a3de9d228/mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/cluster/impl/AgentPersister.java#L55-L65


A simple delete-by-ID (without a "find") might do the trick here, and maybe it would solve your problem?

I'm suggesting this because I suspect we could have the same problem in other situations (e.g. when shutting down Hibernate Search and removing outbox event processor agents)

PR #3361, opened 2023-01-05
HSEARCH-4679 Automatically simplify and/or predicates with a single clause

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4679

I think that...  we've covered the optimization part in the previous PR. So I've added a few more tests to include the `nested` predicate. So this PR is to try to take a stab on `<no other option set>` 😃


yrodiere (commented):
I'm not getting why you moved stuff to a new `*PredicateModifiers` class... does that achieve something?

Even if it does, I'm a bit reluctant to move features from different predicates to a single class, knowing those features (e.g. `minimumShouldMatch`) only make sense for some predicates...

marko-bekhta:
right ... so you were saying that we can "forget" about  some modifiers in the future (at least that was my understanding of the below quote 😄):

> One thing that worries me, though, is if we introduce new options in the boolean predicate and forget to test for them in <no other option set> in the condition above. Then we’d end up ignoring these options in some cases. I wonder if there’s some way to prevent that from ever happening, just by implementing the build in some smart way…

soooo I was thinking how we could make the impl in a way to make it harder to forget to include something new in those checks. "no modifiers" ==> "empty object" ==> "empty map". It is a single class, but because of the map it would really contain only what is required by a specific predicate. I guess we could cut out the `minimumShouldMatch` from the common class back to the bool predicate and extend these modifiers as a private class there so that specific features are more hidden. 

But I guess the question for me is - is it really worth using the map and so on to prevent the potential miss of a new modifier in the future or would a good comment in a few places where we declare and use the modifiers be enough? 😃

marko-bekhta:
Maps? what were we talking about? 🤔 
haha 😄, dropped that commit and added comments to both predicate and builder properties

yrodiere (approved):

PR #3362, opened 2023-01-06
HSEARCH-4775 Improve source code compatibility of Hibernate Search with Hibernate ORM 6.2

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4775


yrodiere:
No problem here, but on second thought I'll bundle this with other, related changes.

PR #3365, opened 2023-01-09
HSEARCH-4756 Use more timezone IDs in TCK tests

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4756


yrodiere:
CI timed out... I suspect all timezone IDs is a lot of timezone IDs?

marko-bekhta:
hmmmm, it's only `602` for me locally 😃 I've tried a few tests locally and it didn't take that much longer than before. Running a full tck build to see what'll happen and if there's some test where it gets stuck.

marko-bekhta:
Seem those loops were for something like query single doc by id or something? can't see why we want to keep them though 😃

yrodiere (approved):

marko-bekhta:
hmm so the failing timezones are:
```
307=1900-01-01T00:00+07:21:20[Asia/Kuching]
507=1900-01-01T00:00+07:39:40[Asia/Brunei]
```
and `Europe/Amsterdam` is also on the list
```
450=1900-01-01T00:00+00:19:32[Europe/Amsterdam]
```
I was somewhat expecting that those "broken" timezones won't be on the list

yrodiere:
Alright. Maybe let's hardcode a few exotic ones and leave it at that? Since the failures are not in Hibernate Search itself (unless it's a different problem?)

yrodiere (approved):

PR #3370, opened 2023-01-12
HSEARCH-4779 Reference the Hibernate Search 6.0 migration guide from the reference documentation

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4779

### DIFF
--- a/documentation/src/main/asciidoc/migration/index.asciidoc
+++ b/documentation/src/main/asciidoc/migration/index.asciidoc
@@ -22,6 +22,19 @@ If you're looking to migrate from an earlier version,
 you should migrate step-by-step, from one minor version to the next,
 following the migration guide of link:https://hibernate.org/search/documentation/[each version].
 
+[WARNING]
+====
+**To Hibernate Search 5 users**
+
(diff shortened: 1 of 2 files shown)


documentation/src/main/asciidoc/migration/index.asciidoc
```diff
@@ -22,6 +22,19 @@ If you're looking to migrate from an earlier version,
 you should migrate step-by-step, from one minor version to the next,
 following the migration guide of link:https://hibernate.org/search/documentation/[each version].
 
+[WARNING]
+====
+**To Hibernate Search 5 users**
+
+Be aware that a lot of APIs changed since Hibernate Search, some only because of a package change,
```
marko-bekhta, line 29:
```suggestion
Be aware that a lot of APIs have changed since Hibernate Search 5, some only because of a package change,
```
looks like version # is missing, or was that on purpose?

marko-bekhta (approved):

documentation/src/main/asciidoc/migration/index.asciidoc
yrodiere: No you're right, thanks

PR #3351, opened 2023-01-02
HSEARCH-4642 Bump dependencies / HSEARCH-4764 Bump AWS SDK to 2.19.8

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4642
https://hibernate.atlassian.net/browse/HSEARCH-4764

For HSEARCH-4642, folllows up on https://github.com/hibernate/hibernate-search/pull/3339, https://github.com/hibernate/hibernate-search/pull/3322, https://github.com/hibernate/hibernate-search/pull/3313, https://github.com/hibernate/hibernate-search/pull/3294, https://github.com/hibernate/hibernate-search/pull/3146, https://github.com/hibernate/hibernate-search/pull/3148, https://github.com/hibernate/hibernate-search/pull/3157, https://github.com/hibernate/hibernate-search/pull/3176, https://github.com/hibernate/hibernate-search/pull/3189, https://github.com/hibernate/hibernate-search/pull/3194, https://github.com/hibernate/hibernate-search/pull/3208, https://github.com/hibernate/hibernate-search/pull/3219, https://github.com/hibernate/hibernate-search/pull/3234, https://github.com/hibernate/hibernate-search/pull/3239, https://github.com/hibernate/hibernate-search/pull/3249, https://github.com/hibernate/hibernate-search/pull/3255, https://github.com/hibernate/hibernate-search/pull/3262, https://github.com/hibernate/hibernate-search/pull/3281

PR #3383, opened 2023-01-16
HSEARCH-4642 Upgrade build dependencies to the latest version

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4642

For HSEARCH-4642, folllows up on https://github.com/hibernate/hibernate-search/pull/3364, https://github.com/hibernate/hibernate-search/pull/3351, https://github.com/hibernate/hibernate-search/pull/3339, https://github.com/hibernate/hibernate-search/pull/3322, https://github.com/hibernate/hibernate-search/pull/3313, https://github.com/hibernate/hibernate-search/pull/3294, https://github.com/hibernate/hibernate-search/pull/3146, https://github.com/hibernate/hibernate-search/pull/3148, https://github.com/hibernate/hibernate-search/pull/3157, https://github.com/hibernate/hibernate-search/pull/3176, https://github.com/hibernate/hibernate-search/pull/3189, https://github.com/hibernate/hibernate-search/pull/3194, https://github.com/hibernate/hibernate-search/pull/3208, https://github.com/hibernate/hibernate-search/pull/3219, https://github.com/hibernate/hibernate-search/pull/3234, https://github.com/hibernate/hibernate-search/pull/3239, https://github.com/hibernate/hibernate-search/pull/3249, https://github.com/hibernate/hibernate-search/pull/3255, https://github.com/hibernate/hibernate-search/pull/3262, https://github.com/hibernate/hibernate-search/pull/3281

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -352,7 +352,7 @@
         <version.clean.plugin>3.2.0</version.clean.plugin>
         <!-- Needing 3.6.2 at least because of MCOMPILER-294 -->
         <version.compiler.plugin>3.10.1</version.compiler.plugin>
-        <version.dependency.plugin>3.4.0</version.dependency.plugin>
+        <version.dependency.plugin>3.5.0</version.dependency.plugin>
         <!-- Check dependencies for security vulnerabilities -->
         <version.dependency-check.plugin>7.2.1</version.dependency-check.plugin>
         <version.deploy.plugin>3.0.0</version.deploy.plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -370,7 +370,7 @@
         <version.jdeps.plugin>0.5.1</version.jdeps.plugin>
         <version.nexus-staging.plugin>1.6.13</version.nexus-staging.plugin>
         <version.processor.plugin>4.5-jdk8</version.processor.plugin>
-        <version.project-info.plugin>3.4.1</version.project-info.plugin>
+        <version.project-info.plugin>3.4.2</version.project-info.plugin>
         <version.release.plugin>2.5.3</version.release.plugin>
         <version.resources.plugin>3.3.0</version.resources.plugin>
         <version.shade.plugin>3.4.1</version.shade.plugin>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -379,7 +379,7 @@
         <!-- Surefire versions are a minefield of bugs: be careful with upgrades -->
         <version.surefire.plugin>3.0.0-M7</version.surefire.plugin>
         <version.surefire.plugin.java-version.asm>9.4</version.surefire.plugin.java-version.asm>
-        <version.failsafe.plugin>${version.surefire.plugin}</version.failsafe.plugin>
+        <version.failsafe.plugin>3.0.0-M8</version.failsafe.plugin>
```
yrodiere, line 382: Looks like dependabot made a mess here, would you mind fixing it?

yrodiere: Perhaps it would be better to remove the property `version.failsafe.plugin` altogether and use `version.surefire.plugin` everywhere, so that dependabot doesn't make the same mistake next time.

yrodiere (changes requested):
Thanks!

The Mockito upgrade seems to require a bit more work. Could you have a look?

But it's probably a good idea to do it ASAP, since we're going to move to JUnit 5 as well (I'd rather not stick to legacy technology in one area and not another when they interact with each other).

There also seems to be a problem with the failsafe upgrade, see below.

marko-bekhta:
hey hey 😃 ,

> The Mockito upgrade seems to require a bit more work. Could you have a look?

 yes sure, I was just looking into it. Adding 
```xml
<dependency>
    <groupId>net.bytebuddy</groupId>
    <artifactId>byte-buddy-agent</artifactId>
    <!-- DO NOT MANAGE this dependency. See the version property declaration. -->
    <version>${version.net.bytebuddy}</version>
</dependency>
```
to `test common` was an easy part, but there's also something going on with the stubs - so looking into it 😃

yrodiere (approved):

PR #3384, opened 2023-01-16
HSEARCH-4620 Simplify the configuration of indexing plan synchronization with the Standalone POJO Mapper

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4620

I wasn't sure about this one. Looking at the sync strategies in the ORM mapper, it seemed that those would be the same in the standalone case except for the entity reference type. But we never really use anything from the reference and it ends up being an `Object`. So the idea was to move out the strategies to the mapper base and reuse them in both ORM and standalone mappers.

Then in the standalone mapper, it doesn't make much sense to keep both low-level APIs for the document commit and sync strategies as well as a bit higher-level strategies. Otherwise, it's a bit confusing what should take precedence if both APIs are used. That's why I've dropped those low-level ones from the builder. 

Now, this is a few breaking changes being introduced 😃 hence I wanted to check if that's the way to go about this task before I figure out some of the failing tests.


yrodiere:
So... I get where you're coming from, but we promised not to break API in minors (see [the compatibility policy](https://hibernate.org/community/compatibility-policy/#api)), and while we can relax the rules from time to time, right now we can't: due to the lifecycle of some integrations, we must ensure 6.2 is fully backwards-compatible with 6.1.

I'm +1 to put in common most of this code if that makes sense, but we will have to find a way that doesn't break existing implementations of `AutomaticIndexingSynchronizationStrategy`.

That could mean:

* Introducing supertypes for `AutomaticIndexingSynchronizationConfigurationContext`, `EntityReference`, etc. in the `mapper-pojo-base` module, and deprecating the subtypes (though I'm not sure about `EntityReference`: it has far-reaching implications).
* Using adapters where relevant (e.g. a wrapper implementation to convert `AutomaticIndexingSynchronizationStrategy` in `mapper-orm` to your new equivalent)
* ...

But one thing is for sure: we don't break existing APIs in 6.2, at least not to introduce new features (bugs/security issues would be another matter).

By the way, you might want to take this opportunity to rename `AutomaticIndexingSynchronizationStrategy` to something like `IndexingPlanSynchronizationStrategy`, since, well... indexing is not quite "automatic" with the standalone POJO mapper.

marko-bekhta:
I've drafted a different approach 😄  copy-pasting the strategies to the standalone mapper would've generated less noise 😄 

This time there's a set of base/generic classes in the base-pojo that are later extended/implemented within the pojo-standalone and orm mappers. I thought that it would be best to have dedicated strategy interfaces and names for strategies per mapper. 
If that's an OK way to go there are still docs to be updated and probably some tests in the docs module.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/plan/synchronization/IndexingPlanSynchronizationStrategy.java
```diff
@@ -0,0 +1,23 @@
...
+ * where indexed entities were modified.
+ *
+ * {@code SearchSession#indexingPlanSynchronizationStrategy(IndexingPlanSynchronizationStrategy)}
+ */
+@Incubating
+public interface IndexingPlanSynchronizationStrategy<E, R extends SearchIndexingPlanExecutionReport<E>, C extends IndexingPlanSynchronizationStrategyConfigurationContext<E, R>> {
+
+	void apply(C context);
+
+}
```
yrodiere, line 23: IMO this is either not enough, or too much.

If we're going to have a sub-interface for each mapper, then it's too much and there's little value in keeping this one, since it's so simple. We could just remove it, declare the method directly in each mapper's interface, and use static methods to share the implementation of built-in strategies across mappers.

But with some changes, it would be possible to use this interface in mappers directly, and then it's not enough because we should also use the same EntityReference/report classes in mappers. Maybe we should do that? Have a look at the other comments and let's discuss it on Zulip.

Whatever happens, please rename the interface exposed to end users to `SearchIndexingPlanSynchronizationStrategy`, which is more consistent with the naming of `SearchIndexingPlan`. Yes, that means two interfaces with the same name, but in two mutually exclusive modules.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/plan/synchronization/IndexingPlanSynchronizationStrategyConfigurationContext.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.hibernate.search.engine.backend.work.execution.DocumentCommitStrategy;
+import org.hibernate.search.engine.backend.work.execution.DocumentRefreshStrategy;
+import org.hibernate.search.engine.backend.work.execution.OperationSubmitter;
+import org.hibernate.search.engine.reporting.FailureHandler;
+import org.hibernate.search.mapper.pojo.work.SearchIndexingPlanExecutionReport;
+import org.hibernate.search.util.common.annotation.Incubating;
+
+@Incubating
+public interface IndexingPlanSynchronizationStrategyConfigurationContext<E, R extends SearchIndexingPlanExecutionReport<E>> {
```
yrodiere, line 20: As far as I can see, `R` is not necessary here. You could remove it and use `SearchIndexingPlanExecutionReport<E>` everywhere you currently use `R`. The only place where it could be a problem is `org.hibernate.search.mapper.orm.work.SearchIndexingPlanExecutionReport`, and you already use adapters for that one anyway.

Thinking about it, we might be able to remove `E` too: the class `EntityReference` in the standalone and ORM mapper are identical, so we could just merge them. That would however require most extensive changes, so maybe this should be done as a follow-up, in a separate PR/issue.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/plan/synchronization/impl/ConfiguredAutomaticIndexingSynchronizationStrategy.java
```diff
@@ -4,38 +4,37 @@
...
-import org.hibernate.search.mapper.orm.common.EntityReference;
-import org.hibernate.search.mapper.orm.work.SearchIndexingPlanExecutionReport;
-import org.hibernate.search.mapper.orm.work.impl.SearchIndexingPlanExecutionReportImpl;
+import org.hibernate.search.mapper.pojo.plan.synchronization.IndexingPlanSynchronizationStrategyConfigurationContext;
+import org.hibernate.search.mapper.pojo.work.SearchIndexingPlanExecutionReport;
 import org.hibernate.search.mapper.pojo.work.spi.PojoIndexingPlan;
 import org.hibernate.search.util.common.impl.Contracts;
 
-public final class ConfiguredAutomaticIndexingSynchronizationStrategy {
+public abstract class ConfiguredAutomaticIndexingSynchronizationStrategy<E, R extends SearchIndexingPlanExecutionReport<E>> {
```
yrodiere, line 24: Is there any point in making this abstract? As far as I can see, all implementations are the same.

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/cfg/StandalonePojoMapperSettings.java
```diff
@@ -63,6 +80,9 @@ private Radicals() {
 		public static final String SCHEMA_MANAGEMENT_STRATEGY = "schema_management.strategy";
 		public static final String MAPPING_CONFIGURER = "mapping.configurer";
 		public static final String MULTI_TENANCY_ENABLED = "mapping.multi_tenancy.enabled";
+		public static final String INDEXING_PLAN_SYNCHRONIZATION_PREFIX = "standalone.indexing.plan.synchronization.";
```
yrodiere, line 83: Same as last time we discussed it: please don't add prefixes such as `standalone.` to configuration properties. We don't do that anywhere else :)

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/plan/synchronization/PojoStandaloneIndexingPlanSynchronizationStrategyNames.java
```diff
@@ -0,0 +1,49 @@
...
+	 */
+	public static final String READ_SYNC = "pojo-read-sync";
+
+	/**
+	 * A strategy that waits for index changes to be queued and applied, forces a commit and a refresh,
+	 * and waits for the commit and refresh to complete.
+	 * <p>
+	 * See the reference documentation for details.
+	 */
+	public static final String SYNC = "pojo-sync";
```
yrodiere, line 47: Why the `pojo-` prefixes?

documentation/src/main/asciidoc/reference/indexing-plan.asciidoc
```diff
@@ -106,8 +106,6 @@ See <<mapper-orm-indexing-manual-indexingplan-process-execute>> for more informa
 [[indexing-plan-synchronization]]
 == [[mapper-orm-indexing-automatic-synchronization]] Synchronization with the indexes
 
-include::components/mapper-orm-only-note.asciidoc[]
-
```
yrodiere, line 110: Good thinking, but then there's a lot more to be rewritten in this section to make it applicable to both the Standalone mapper and the ORM mapper.

I'd have a hard time listing everything, so it would probably be easier if I give it a try. I'll leave this comment here, unresolved, so that I remember to have a look before we merge.

yrodiere (commented):
There are good ideas, but as you noticed it's a lot of noise, so we should probably try to improve on this, or maybe find another solution.

Please have a look at my comments, and let's discuss this further on Zulip when you have time.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/plan/synchronization/impl/ConfiguredAutomaticIndexingSynchronizationStrategy.java
marko-bekhta: IIRC this abstract was for the:
```
-						.thenApply( SearchIndexingPlanExecutionReportImpl::new );
+						.thenApply( this::createReport );
```
as reports were mapper specific. (at least originally 😄)

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/cfg/StandalonePojoMapperSettings.java
marko-bekhta: yeah, I remember it 😃 but ... 
This as well as the  `orm-` or `pojo-` prefixes in the predefined bean names and in the new property for the ORM mapper was done because I've run into a problem that there are multiple beans with the same name defined. I had the strategies for both ORM and Standalone mappers added using the interface from the pojo-base.  
In the end, both mappers have their specific strategies, so that shouldn't be a problem anymore (I hope 🙈), but I've left the prefixes for the time being until we have a "final approach decided" 😃

documentation/src/main/asciidoc/reference/indexing-plan.asciidoc
marko-bekhta: > there are still docs to be updated and probably some tests in the docs module.

yeah.. once I've realized that methods would be called differently and that there might be even different names for the beans, I've decided to leave the docs unchanged for now and only mark an intent to modify the section so that it's "reusable" for both mappers 🙈

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/plan/synchronization/IndexingPlanSynchronizationStrategyConfigurationContext.java
```diff
@@ -0,0 +1,67 @@
...
+	 *
+	 * @param handler A handler that will be passed a future representing the progress of indexing.
+	 * Defaults to a no-op handler.
+	 * The future will be completed with an execution report once all document changes are applied.
+	 * If any document change or the commit/refresh required by{@link #documentCommitStrategy(DocumentCommitStrategy)}
+	 * and {@link #documentRefreshStrategy(DocumentRefreshStrategy)} failed,
+	 * the report will {@link SearchIndexingPlanExecutionReport#throwable() contain a throwable}
+	 * and (if applicable) {@link SearchIndexingPlanExecutionReport#failingEntities() a list of failing entities}.
+	 */
+	void indexingFutureHandler(Consumer<CompletableFuture<SearchIndexingPlanExecutionReport>> handler);
```
yrodiere, line 48: Could you use this signature, so that we easily can add generics to `SearchIndexingPlanExecutionReport` in the future?

Yes it's complex to the point of being obscure, but this is a rather advanced API anyway, so that's no big deal.

```suggestion
	void indexingFutureHandler(Consumer<? super CompletableFuture<? extends SearchIndexingPlanExecutionReport>> handler);
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/plan/synchronization/IndexingPlanSynchronizationStrategyNames.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.mapper.pojo.plan.synchronization;
+
+import org.hibernate.search.util.common.annotation.Incubating;
+
+/**
+ * The names of built-in automatic indexing synchronization strategies.
```
yrodiere, line 12:
```suggestion
 * The names of built-in indexing plan synchronization strategies.
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/work/SearchIndexingPlanExecutionReport.java
```diff
@@ -0,0 +1,26 @@
...
+	 * @return The {@link Exception} or {@link Error} thrown when indexing failed,
+	 * or {@link Optional#empty()} if indexing succeeded.
+	 */
+	Optional<Throwable> throwable();
+
+	/**
+	 * @return A list of references to entities that may not be indexed correctly as a result of the failure.
+	 * Never {@code null}, but may be empty.
+	 */
+	List<Object> failingEntities();
```
yrodiere, line 24: Please use a wildcard here so that we can introduce generics in `SearchIndexingPlanExecutionReport` later if necessary.

```suggestion
	List<?> failingEntities();
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/work/impl/DelegatingSearchIndexingPlanExecutionReport.java
```diff
@@ -0,0 +1,36 @@
...
+	@Override
+	public Optional<Throwable> throwable() {
+		return delegate.throwable();
+	}
+
+	@Override
+	public List<Object> failingEntities() {
+		return delegate.failingEntityReferences().stream()
+				.map( Object.class::cast )
+				.collect( Collectors.toList() );
```
yrodiere, line 33:
```suggestion
	public List<?> failingEntities() {
		return delegate.failingEntityReferences();
```

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/cfg/StandalonePojoMapperSettings.java
```diff
@@ -60,6 +63,20 @@ private StandalonePojoMapperSettings() {
 	public static final String MULTI_TENANCY_ENABLED = PREFIX + Radicals.MULTI_TENANCY_ENABLED;
 
 
+	/**
+	 * The synchronization strategy to use when indexing automatically.
```
yrodiere, line 67:
```suggestion
	 * The synchronization strategy to use when indexing through the
	 * {@link org.hibernate.search.mapper.pojo.standalone.session.SearchSession SearchSession}'s
	 * {@link org.hibernate.search.mapper.pojo.standalone.session.SearchSession#indexingPlan indexing plan}.
```

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/session/impl/StandalonePojoSearchSession.java
```diff
@@ -237,7 +287,55 @@ public SearchSessionBuilder loading(Consumer<SelectionLoadingOptionsStep> loadin
...
+			}
+			else if ( synchronizationStrategy == null && lowLevelConfig ) {
+				synchronizationStrategy = strategy(
+						Optional.ofNullable( commitStrategy ).orElse( DocumentCommitStrategy.FORCE ),
+						Optional.ofNullable( refreshStrategy ).orElse( DocumentRefreshStrategy.NONE )
+				);
+			}
+			else if ( synchronizationStrategy == null ) {
+				synchronizationStrategy = IndexingPlanSynchronizationStrategy.writeSync();
+			}
```
yrodiere, line 302: Don't bother, the standalone Pojo mapper is a new module which was only introduced in 6.2.0.Alpha1. We can break APIs as we please. In fact I wouldn't even document the API break.

Just remove support for setting `refreshStrategy`/`commitStrategy` directly?

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/mapping/impl/StandalonePojoMapping.java
```diff
@@ -62,9 +61,12 @@ public class StandalonePojoMapping extends AbstractPojoMappingImplementor<Standa
 
 	private final SchemaManagementListener schemaManagementListener;
 
-	StandalonePojoMapping(PojoMappingDelegate mappingDelegate, StandalonePojoTypeContextContainer typeContextContainer,
+	StandalonePojoMapping(ConfigurationPropertySource configurationPropertySource, BeanResolver beanResolver,
+			PojoMappingDelegate mappingDelegate, StandalonePojoTypeContextContainer typeContextContainer,
 			SchemaManagementListener schemaManagementListener) {
 		super( mappingDelegate );
+		this.configurationPropertySource = configurationPropertySource;
```
yrodiere, line 68: This is not a good idea, as this means the configuration (and all the bean instances it may contain) will never get garbage collected.

Can you try extracting the default strategy (as a `ConfiguredIndexingPlanSynchronizationStrategy`) in `#start`?

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/mapping/SearchMappingBuilder.java
```diff
@@ -116,9 +116,14 @@ public CloseableSearchMapping build() {
...
 					integrationPartialBuildState.finalizer( propertySource, propertyChecker );
+
+			SearchIntegrationEnvironment finalEnvironment = environment;
 			mapping = finalizer.finalizeMapping(
 					mappingKey,
-					(context, partialMapping) -> partialMapping.finalizeMapping()
+					(context, partialMapping) -> partialMapping.finalizeMapping(
+							finalEnvironment.propertySource(),
+							finalEnvironment.beanResolver()
+					)
```
yrodiere, line 126: Please revert this, and get the configuration/bean resolver from the context in `StandalonePojoMapping#start`. See my other comment in `StandalonePojoMapping`.

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/logging/impl/Log.java
```diff
@@ -91,4 +91,7 @@ SearchException unknownTypeIdentifierForIndexedEntityType(PojoRawTypeIdentifier<
 	SearchException unknownClassForMappedEntityType(@FormatWith(ClassFormatter.class) Class<?> invalidClass,
 			@FormatWith(CommaSeparatedClassesFormatter.class) Collection<Class<?>> validClasses);
 
+	@Message(id = ID_OFFSET + 19, value = "Index plan synchronization strategy and at least one of commit/refresh strategies are set at the same time."
+			+ " Use either the index plan synchronization strategy or a combination of commit/refresh strategies.")
+	SearchException conflictingIndexPlanSynchronizationConfiguration();
```
yrodiere, line 96: You can remove this, see my other comment.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/session/impl/HibernateOrmIndexingPlanSynchronizationStrategyAdapter.java
```diff
@@ -0,0 +1,82 @@
...
+@SuppressWarnings("deprecation")
+public class HibernateOrmIndexingPlanSynchronizationStrategyAdapter implements IndexingPlanSynchronizationStrategy {
+
+	private final org.hibernate.search.mapper.orm.automaticindexing.session.AutomaticIndexingSynchronizationStrategy delegate;
+
+	public HibernateOrmIndexingPlanSynchronizationStrategyAdapter(
+			org.hibernate.search.mapper.orm.automaticindexing.session.AutomaticIndexingSynchronizationStrategy delegate) {
+		this.delegate = delegate;
+	}
+
```
yrodiere, line 32:
```suggestion

	@Override
	public String toString() {
		return HibernateOrmIndexingPlanSynchronizationStrategyAdapter.class.getSimpleName() + "(" + delegate.toString() + ")";
	}

```

yrodiere (commented):
Thanks! It's coming along nicely. I have a few more comments, see below.

After that round, you can probably mark the PR as "Ready for review" since, well... I've been reviewing it anyway :)

yrodiere:
Ah, and you should probably add something to the migration guide.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/plan/synchronization/spi/ReadSyncIndexingPlanSynchronizationStrategy.java
```diff
@@ -4,36 +4,36 @@
  * License: GNU Lesser General Public License (LGPL), version 2.1 or later
  * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
  */
-package org.hibernate.search.mapper.orm.automaticindexing.session.impl;
+package org.hibernate.search.mapper.pojo.plan.synchronization.spi;
```
yrodiere, line 7: This (and the other strategy implementations) should be in an `impl` package, unless there's a real reason to expose them through an SPI (I don't think there is?).

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/session/impl/ConfiguredAutomaticIndexingStrategy.java
```diff
@@ -120,6 +114,72 @@ public void start(HibernateOrmSearchSessionMappingContext mappingContext,
...
+			}
+			catch (Exception e) {
+				@SuppressWarnings("deprecation")
+				String deprecatedProperty = HibernateOrmMapperSettings.AUTOMATIC_INDEXING_SYNCHRONIZATION_STRATEGY;
+				throw log.unableToConvertConfigurationProperty(
+						deprecatedProperty,
+						e.getMessage(),
+						e
+				);
+			}
```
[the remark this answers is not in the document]
marko-bekhta: 👍🏻 right ... so let's add `IndexingPlanSynchronizationStrategyIT` that would test the new property and keep the existing test doing what it did before 😃

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/mapping/impl/StandalonePojoMapping.java
```diff
@@ -73,6 +80,15 @@ public class StandalonePojoMapping extends AbstractPojoMappingImplementor<Standa
 	@Override
 	public CompletableFuture<?> start(MappingStartContext context) {
 		integrationHandle = context.integrationHandle();
+		// check if indexing plan sync strategy was configured and cache it for further usage when creating sessions:
+		try ( BeanHolder<? extends IndexingPlanSynchronizationStrategy> beanHolder = INDEXING_PLAN_SYNCHRONIZATION_STRATEGY.getAndTransform(
+				context.configurationPropertySource(),
+				referenceOptional -> context.beanResolver().resolve( referenceOptional.orElse(
+						StandalonePojoMapperSettings.Defaults.INDEXING_PLAN_SYNCHRONIZATION_STRATEGY ) )
+		) ) {
+			this.configuredIndexingPlanSynchronizationStrategy = beanHolder.get();
```
yrodiere, line 89: This is not an instance of `ConfiguredIndexingPlanSynchronizationStrategy`, though... ? I was suggesting to create a `ConfiguredIndexingPlanSynchronizationStrategy` and cache *that*, not the `IndexingPlanSynchronizationStrategy`.

Relatedly, note that extracting a bean from a `BeanHolder` and using it after the `BeanHolder` is closed is a big no-no, as CDI (for example) may execute some pre-destruct methods when `BeanHolder.close()` is called, rendering the bean unusable.

yrodiere (commented):
Thanks. I have a few comments; I pushed two commits to address some of it, but I think there's still something that needs to be done around the "configured" indexing plan strategy that you tried to cache.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/plan/synchronization/spi/impl/AsyncIndexingPlanSynchronizationStrategy.java
```diff
@@ -0,0 +1,62 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.mapper.pojo.plan.synchronization.spi.impl;
```
yrodiere, line 7: I think there's a mistake here, the package should be `.impl`, not `.spi.impl`...

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/session/AutomaticIndexingSynchronizationStrategyIT.java
```diff
@@ -478,7 +479,7 @@ public void invalidReference() {
 				.hasMessageContainingAll(
 						"Invalid value for configuration property '"
 								+ HibernateOrmMapperSettings.AUTOMATIC_INDEXING_SYNCHRONIZATION_STRATEGY
-								+ "': 'invalidName'",
+								+ "'",
```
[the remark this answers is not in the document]
marko-bekhta: I was thinking that maybe it's not that important to log the incorrect value compared to getting the property twice. But I like your change, thanks 😃

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/mapping/impl/StandalonePojoMapping.java
marko-bekhta: Right... I've ended up caching the `IndexingPlanSynchronizationStrategy` as I thought - because the user can override the strategy at the session level it would somewhat duplicate that logic of creating the `ConfiguredIndexingPlanSynchronizationStrategy`. But I'll give it another shot. 
oh ok.. so we'd need to hold to the bean holder and push it to the closer in `StandalonePojoMapping#close` 👍🏻

yrodiere: You don't necessarily need to duplicate the logic, or at least not a lot of it. IIRC I did it for the ORM mapper.

yrodiere (approved):
Thanks. I pushed one last commit to move the new `IndexingPlan*` classes you introduced to the `mapper.pojo.work` package, next to other indexing-plan-related classes. This also makes sense because the API classes `SearchIndexingPlan` are in the `org.hibernate.search.mapper.orm.work` package and the `org.hibernate.search.mapper.pojo.standalone.work` package in their respective module. 

Let me know what you think. Feel free squash commits as you wish (you can squash my commits into yours and lose authorship info, I don't mind if it's cleaner), but in any case: LGTM, feel free to merge :)

marko-bekhta:
fine by me 😄 👍🏻 thanks
there's also this test - https://github.com/hibernate/hibernate-search/blob/4b14294981d15f7b4ba669d97fd28fa7c4c0a508/integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/plan/synchronization/IndexingPlanSynchronizationStrategyIT.java 
maybe we should move it as well then ?

yrodiere:
> there's also this test - https://github.com/hibernate/hibernate-search/blob/4b14294981d15f7b4ba669d97fd28fa7c4c0a508/integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/plan/synchronization/IndexingPlanSynchronizationStrategyIT.java maybe we should move it as well then ?

Right, good catch. Done.

documentation/src/main/asciidoc/reference/indexing-plan.asciidoc
marko-bekhta: Just pinging on this one to see if you'd want to add anything in the docs or leave it for some other RP?

yrodiere: I think what you did is probably enough for now, we'll have another look when we address https://hibernate.atlassian.net/browse/HSEARCH-4616

marko-bekhta:
the build passed after rebase; merging. Thanks for the reviews!

PR #3385, opened 2023-01-17
HSEARCH-4781 Mention @ScaledNumberField in error messages when the decimal scale cannot be found

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4781

As it seems that the hint would be independent of the backend - there are no backend-specific overrides for `BackendMappingHints`

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/ElasticsearchIndexFieldTypeFactoryImpl.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/ElasticsearchIndexFieldTypeFactoryImpl.java
@@ -24,6 +24,8 @@ import org.hibernate.search.backend.elasticsearch.logging.impl.Log;
 import org.hibernate.search.backend.elasticsearch.types.dsl.ElasticsearchIndexFieldTypeFactory;
 import org.hibernate.search.backend.elasticsearch.types.dsl.ElasticsearchNativeIndexFieldTypeMappingStep;
 import org.hibernate.search.backend.elasticsearch.types.format.impl.ElasticsearchDefaultFieldFormatProvider;
+import org.hibernate.search.engine.backend.mapping.spi.BackendMapperContext;
+import org.hibernate.search.engine.backend.reporting.spi.BackendMappingHints;
 import org.hibernate.search.engine.backend.types.dsl.ScaledNumberIndexFieldTypeOptionsStep;
 import org.hibernate.search.engine.backend.types.dsl.StandardIndexFieldTypeOptionsStep;
 import org.hibernate.search.engine.backend.types.dsl.StringIndexFieldTypeOptionsStep;
@@ -42,14 +44,17 @@ public class ElasticsearchIndexFieldTypeFactoryImpl
 	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
 
 	private final EventContext eventContext;
+	private final BackendMapperContext backendMapperContext;
 	private final Gson userFacingGson;
 	private final ElasticsearchDefaultFieldFormatProvider defaultFieldFormatProvider;
 	private final IndexFieldTypeDefaultsProvider typeDefaultsProvider;
 
-	public ElasticsearchIndexFieldTypeFactoryImpl(EventContext eventContext, Gson userFacingGson,
+	public ElasticsearchIndexFieldTypeFactoryImpl(EventContext eventContext, BackendMapperContext backendMapperContext,
+			Gson userFacingGson,
 			ElasticsearchDefaultFieldFormatProvider defaultFieldFormatProvider,
 			IndexFieldTypeDefaultsProvider typeDefaultsProvider) {
 		this.eventContext = eventContext;
+		this.backendMapperContext = backendMapperContext;
 		this.userFacingGson = userFacingGson;
 		this.defaultFieldFormatProvider = defaultFieldFormatProvider;
 		this.typeDefaultsProvider = typeDefaultsProvider;
@@ -249,5 +254,10 @@ public class ElasticsearchIndexFieldTypeFactoryImpl
 	@Override
 	public ElasticsearchDefaultFieldFormatProvider getDefaultFieldFormatProvider() {
 		return defaultFieldFormatProvider;
+	}
+
+	@Override
+	public BackendMappingHints hints() {
+		return backendMapperContext.hints();
 	}
 }
--- a/engine/src/main/java/org/hibernate/search/engine/backend/reporting/spi/BackendMappingHints.java
+++ b/engine/src/main/java/org/hibernate/search/engine/backend/reporting/spi/BackendMappingHints.java
@@ -20,4 +20,7 @@ public interface BackendMappingHints {
 	@Message(value = "")
 	String noEntityProjectionAvailable();
 
+	@Message( "" )
+	String missingDecimalScale();
+
 }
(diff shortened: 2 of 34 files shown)


engine/src/main/java/org/hibernate/search/engine/backend/reporting/spi/BackendMappingHints.java
```diff
@@ -20,4 +20,7 @@ public interface BackendMappingHints {
 	@Message(value = "")
 	String noEntityProjectionAvailable();
 
+	@Message( "Add the @ScaledNumberField annotation to this property and configure the `decimalScale` attribute as necessary." )
```
yrodiere, line 23: That's cheating... `@ScaledNumberField` doens't exist in the engine module either.

This should just be `@Message("")`. The real implementation must be in `StandalonePojoMappingHints`/`HibernateOrmMappingHints`. At best, you can introduce an intermediary interface `PojoBackendMappingHints` to avoid duplicating the same message, but that's about all you can do.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/ElasticsearchIndexFieldTypeFactoryImpl.java
```diff
@@ -250,4 +251,9 @@ public Gson getUserFacingGson() {
 	public ElasticsearchDefaultFieldFormatProvider getDefaultFieldFormatProvider() {
 		return defaultFieldFormatProvider;
 	}
+
+	@Override
+	public BackendMappingHints hints() {
+		return BackendMappingHints.NONE;
```
yrodiere, line 257: This is very weird. The actual hints should be retrieved from the mapping, like in `org.hibernate.search.engine.backend.mapping.spi.BackendMappingContext#hints`. You'll need to create a new `hints` methods though, because this is bootstrap and the `BackendMappingContext` does not even exist yet.

Maybe some new `BackendMapperContext` that you would implement in `PojoMapper` and whose instance would be passed by `PojoMapper` to a new argument of `org.hibernate.search.engine.common.impl.MappedIndexManagerFactoryImpl#createMappedIndexManager` and eventually `org.hibernate.search.backend.elasticsearch.impl.ElasticsearchBackendImpl#createIndexManagerBuilder`, to be stored in `org.hibernate.search.backend.elasticsearch.document.model.dsl.impl.ElasticsearchIndexRootBuilder` and eventually used in the type factory when it's created in `org.hibernate.search.backend.elasticsearch.document.model.dsl.impl.ElasticsearchIndexRootBuilder#createTypeFactory`?

yrodiere (changes requested):
Thanks. That's a start, but I think you missed the point of the `BackendMappingHints` interface; see below.

engine/src/main/java/org/hibernate/search/engine/common/impl/IndexManagerBuildingStateHolder.java
```diff
@@ -164,7 +165,7 @@ private BackendInitialBuildState(EventContext eventContext,
 		}
 
 		IndexManagerInitialBuildState createIndexManagerBuildingState(
-				String indexName, String mappedTypeName) {
+				BackendMapperContext backendMapperContext,	String indexName, String mappedTypeName) {
```
yrodiere, line 168:
```suggestion
				BackendMapperContext backendMapperContext, String indexName, String mappedTypeName) {
```

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/MappedIndexManagerBuilderImpl.java
```diff
@@ -17,7 +18,7 @@ public class MappedIndexManagerBuilderImpl implements MappedIndexManagerBuilder
 	private final IndexManagerBuildingState indexManagerBuildingState;
 
 	public MappedIndexManagerBuilderImpl(IndexedEntityBindingMapperContext mapperContext,
-			IndexManagerBuildingState indexManagerBuildingState) {
+			BackendMapperContext backendMapperContext, IndexManagerBuildingState indexManagerBuildingState) {
```
yrodiere, line 21: This doesn't seems to be used?

engine/src/test/java/org/hibernate/search/engine/common/impl/IndexManagerBuildingStateHolderTest.java
```diff
@@ -334,4 +332,23 @@ public static BackendsInfo namedSingleTenancy(String name) {
...
+					public String noEntityProjectionAvailable() {
+						return "noEntityProjectionAvailable";
+					}
+
+					@Override
+					public String missingDecimalScale() {
+						return "missingDecimalScale";
+					}
+				};
+			}
```
yrodiere, line 351: Feel free to use `BackendMappingHints.NONE`.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/reporting/impl/HibernateOrmMappingHints.java
```diff
@@ -22,4 +22,7 @@ public interface HibernateOrmMappingHints extends BackendMappingHints {
 	@Message(value = "This is likely a bug, as Hibernate ORM entities should always be loadable from the database.")
 	String noEntityProjectionAvailable();
 
+	@Override
+	@Message( "Add the @ScaledNumberField annotation to this property and configure the `decimalScale` attribute as necessary." )
```
yrodiere, line 26: I think the exception can also arise when using `TypeBinder`/`PropertyBinder`, so it's entirely possible no `@*Field` annotation is involved... Maybe some less definitive message would be better:

```suggestion
	@Message( "If you used a @*Field annotation here, make sure to use @ScaledNumberField and configure the `decimalScale` attribute as necessary." )
```

Same for the standalone POJO Mapper.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/building/impl/PojoMapper.java
```diff
@@ -112,7 +115,8 @@ public PojoMapper(MappingBuildContext buildContext,
 			BeanReference<? extends IdentifierBridge<Object>> providedIdentifierBridge,
 			IdentityMappingMode containedEntityIdentityMappingMode,
 			TenancyMode tenancyMode, ReindexOnUpdate defaultReindexOnUpdate,
-			PojoMapperDelegate<MPBS> delegate) {
+			PojoMapperDelegate<MPBS> delegate,
+			BackendMappingHints hints) {
```
yrodiere, line 119: I guess you could also... delegate... to `PojoMapperDelegate` when you need to retrieve the hints?
Or even have `PojoMapperDelegate` extend `BackendMapperContext` directly, if that makes sense.

yrodiere (commented):
Thanks, getting closer. My two cents :)

Feel free to yell at me if it's getting frustrating :x

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/building/impl/PojoMapper.java
marko-bekhta: Hmm, I wanted to do so initially but `PojoMapperDelegate` seemed to have unrelated methods so I didn't want to put in the hints there, BUT if you are ok with it I'll do so 😄 .

yrodiere (approved):
LGTM! I just added a commit to introduce a default method in `PojoMapperDelegate` so that we don't break backwards compatibility (at least not in this PR).

Will merge, thanks!

marko-bekhta:
thanks!

PR #3386, opened 2023-01-18
HSEARCH-4780 Test Hibernate Search against JDK 21

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4780

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -177,6 +177,9 @@ stage('Configure') {
 							condition: TestCondition.AFTER_MERGE),
 					new JdkBuildEnvironment(version: '20', testCompilerTool: 'OpenJDK 20 Latest',
 							testLauncherArgs: '--enable-preview',
+							condition: TestCondition.AFTER_MERGE),
+					new JdkBuildEnvironment(version: '21', testCompilerTool: 'OpenJDK 21 Latest',
+							testLauncherArgs: '--enable-preview',
 							condition: TestCondition.AFTER_MERGE)
 			],
 			compiler: [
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -167,15 +167,16 @@ stage('Configure') {
 					new JdkBuildEnvironment(version: '17', testCompilerTool: 'OpenJDK 17 Latest',
 							condition: TestCondition.BEFORE_MERGE,
 							isDefault: true),
-					new JdkBuildEnvironment(version: '18', testCompilerTool: 'OpenJDK 18 Latest',
-							condition: TestCondition.AFTER_MERGE),
 					// We want to enable preview features when testing early-access builds of OpenJDK:
 					// even if we don't use these features, just enabling them can cause side effects
 					// and it's useful to test that.
 					new JdkBuildEnvironment(version: '19', testCompilerTool: 'OpenJDK 19 Latest',
 							testLauncherArgs: '--enable-preview',
```
marko-bekhta, line 174: Just wondering if we should remove the preview flag from 19 based on the comment above... (19 not being EA build anymore)

marko-bekhta (approved):

Jenkinsfile
yrodiere: The preview flag is about preview features; those can still be preview in GA releases. The goal is for the JDK team to be able to release features without committing to backwards compatibility.

So, in short, the flag is still useful :)

yrodiere: I should probably change the comment though.

yrodiere: Done, thanks for the heads-up

yrodiere:
JDK 8 tests are failing for unrelated reasons. JDK 21 tests are failing because Mockito doesn't seem to support JDK 21 yet; we'll probably need to pass an option to have bytebuddy process JDK 21 bytecode.

PR #3392, opened 2023-01-23
HSEARCH-4789 [HSearch 5] Indexing entities with a shared id across tenants results in removal from those tenants

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4789

Note this is Hibernate Search 5, so the code is completely different from 6.

This particular part of the code is... let's say interesting. I cleaned up what was reasonable, but the main goal here is really to just use a common implementation in the executors, one that handles tenant IDs correctly.

### DIFF
--- a/engine/src/main/java/org/hibernate/search/backend/impl/lucene/works/ByTermDeleteWorkExecutor.java
+++ b/engine/src/main/java/org/hibernate/search/backend/impl/lucene/works/ByTermDeleteWorkExecutor.java
@@ -27,14 +27,17 @@ import java.lang.invoke.MethodHandles;
 import org.hibernate.search.util.logging.impl.Log;
 
 /**
- * Extension of <code>DeleteLuceneWork</code> that will always perform the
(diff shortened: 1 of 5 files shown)


marko-bekhta (approved):
I think I got where the problem was, and the changes look good 😃. The only idea I had - maybe we can move the `DeleteWorkExecutor#isIdNumeric(..)` to the builder itself (`DocumentBuilderIndexedEntity`)?
That way, we wouldn't need to drag that parameter through all the methods?

yrodiere:
> maybe we can move the DeleteWorkExecutor#isIdNumeric(..) to the builder itself (DocumentBuilderIndexedEntity)?

I agree, but the fact that the call to `isIdNumeric` is "cached" is pretty much the only reason for `*ExtExecutor` classes to exist. I very much doubt these classes are useful or this provides a significant performance gain, but this is Hibernate Search 5 and I don't want to take any more risk than necessary. I'm already stretching the limits by removing code duplication, so I'll stop there :)

yrodiere:
Merging, thanks for the review.

PR #3393, opened 2023-01-24
[6.0] HSEARCH-4786 Fix non-compiling code in migration guide

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4786


marko-bekhta (approved):
Ohh, I would've thought that these code snippets come from the sources 🙈 but then .. this is a migration guide 😃 Looks good 👍🏻 ( I have to admit - I haven't tried each code snippet to make sure it compiles 😃🙈)

yrodiere:
Yeah, I don't think it's worth the trouble of setting up test cases just for the migration guide.

yrodiere:
Merged, thanks

PR #3394, opened 2023-01-26
HSEARCH-4790 Add compatibility with OpenSearch 2.5.0

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4790


integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/util/ElasticsearchTckBackendFeatures.java
```diff
@@ -52,8 +51,9 @@ public boolean normalizesStringMissingValues() {
 	public boolean normalizesStringArgumentToWildcardPredicateForAnalyzedStringField() {
 		// In ES 7.7 through 7.11, wildcard predicates on analyzed fields get their pattern normalized,
 		// but that was deemed a bug and fixed in 7.12.2+: https://github.com/elastic/elasticsearch/pull/53127
+		// Apparently the "fix" was also introduced in OpenSearch 2.5.0
 		return isBetween( actualVersion, "elastic:7.7", "elastic:7.12.1" ) ||
-				isOpenSearch( actualVersion );
+				!isLessThan( actualVersion, "opensearch:2.5.0" );
```
yrodiere, line 56: Judging from the code, `isLessThan( actualVersion, "opensearch:2.5.0" )` would return `false` for any version of Elasticsearch (see `tryOrFalse`), so your change will make `normalizesStringArgumentToWildcardPredicateForAnalyzedStringField` return true for every version of Elasticsearch... not good?

yrodiere (changes requested):

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/util/ElasticsearchTckBackendFeatures.java
marko-bekhta: hmm I've looked at the other conditions that we have, and it makes me wonder... Maybe we should be more explicit:
```
return condition(
		actualVersion,
		esVersion -> !isLessThan( esVersion, "elastic:6.4.0" ),   // check for ES
		osVersion -> true                                                                 // check for OS
);
```
Something along those lines. WDYT?

yrodiere: Ok, but please make it so we don't need the `elastic:` prefix anymore, then.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/bootstrap/ElasticsearchBootstrapIT.java
```diff
@@ -127,7 +127,11 @@ public void noVersionCheck_incompleteVersion() {
 						" use the same model dialect." +
 						" It is not the case on ES 5 in particular, since 5.6 has a dialect" +
 						" but 5.0, 5.1, etc. don't have one.",
-				isAtMost( ElasticsearchTestDialect.getActualVersion(), "elastic:5.6" )
+				condition(
+						ElasticsearchTestDialect.getActualVersion(),
+						esVersion -> esVersion.isAtMost( "5.6" ),
+						osVersion -> osVersion.none()
```
yrodiere, line 133: That's... unnecessarily complex? Personally I think a simple `false` here would make much more sense.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/bootstrap/ElasticsearchBootstrapIT.java
```diff
@@ -127,7 +127,11 @@ public void noVersionCheck_incompleteVersion() {
 						" use the same model dialect." +
 						" It is not the case on ES 5 in particular, since 5.6 has a dialect" +
 						" but 5.0, 5.1, etc. don't have one.",
-				isAtMost( ElasticsearchTestDialect.getActualVersion(), "elastic:5.6" )
+				condition(
+						ElasticsearchTestDialect.getActualVersion(),
```
yrodiere, line 131: If you're always using the version from `ElasticsearchTestDialect`, wouldn't it be simpler to expose a `condition` method in `ElasticsearchTestDialect` that implicitly uses that version?

I wonder if some renaming wouldn't also make the thing feel more smooth, e.g.:

```java
ElasticsearchTestDialect.isVersion(
		elasticsearch -> elasticsearch.isAtMost( "5.6" ),
		opensearch -> false
)
```

WDYT?

util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/dialect/ElasticsearchVersionUtils.java
```diff
@@ -31,57 +31,69 @@ private static int compare(ElasticsearchVersion a, ElasticsearchVersion b, int d
...
 	}
 
-	public static boolean isMatching(ElasticsearchVersion actual, String version) {
-		ElasticsearchVersion v = ElasticsearchVersion.of( version );
+	public static class ElasticsearchVersionCondition {
+		private final ElasticsearchVersion actual;
 
-		return v.matches( actual );
-	}
+		public ElasticsearchVersionCondition(ElasticsearchVersion actual) {
```
yrodiere, line 54:
```suggestion
		private ElasticsearchVersionCondition(ElasticsearchVersion actual) {
```

util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/dialect/ElasticsearchVersionUtils.java
```diff
@@ -31,57 +31,69 @@ private static int compare(ElasticsearchVersion a, ElasticsearchVersion b, int d
...
+			case OPENSEARCH:
+				return opensearchPredicate.test( condition );
+			default:
+				throw new IllegalStateException( "Unknown distribution" );
+		}
 	}
 
-	public static boolean isMatching(ElasticsearchVersion actual, String version) {
-		ElasticsearchVersion v = ElasticsearchVersion.of( version );
+	public static class ElasticsearchVersionCondition {
```
yrodiere, line 51: Nice idea, that way we are forced to always go through `condition` to test a version.

I'd go a step further: you could easily get rid of `ElasticsearchVersionUtils` by moving all its methods into `ElasticsearchVersionCondition` and making that class top-level.

yrodiere (commented):
Nice job. I have a few suggestions to polish the new utils, so I'll let you be the judge of what makes sense and what doesn't :) Feel free to merge once you're done.

yrodiere (approved):

marko-bekhta:
Going to merge this one as the failed test looks to be the same one that fails in the main build.

PR #3397, opened 2023-01-30
HSEARCH-1809 Introduce the context for the massindexer batch thread work

marko-bekhta:

I was cleaning up some local branches and noticed this one lying around. IIRC this was asked by someone on discourse for `5->6 ` migration. I've rebased and cleaned it up a bit.

You'll notice `BatchingThreadContext` being passed all the way to `PojoMassIndexingFailureHandledRunnable`. While it is not necessarily needed for all the classes extending this runnable, it seemed simpler to have it stored there.

Also, I wasn't sure if we really need to have this context for a workspace (`PojoMassIndexingBatchIndexingWorkspace`). It seems to make more sense for identifier and entity loading. Which ... begs the question - would such generic context be ok, or should we be more granular and allow a user to tell us what to do when loading identifiers and what to do when loading entities? In such case, I'd probably convert the `BatchingThreadContext` to more of a builder thing asking a user to provide consumers for all the exposed hook points.

### DIFF
--- a/documentation/src/main/asciidoc/reference/indexing-massindexer.asciidoc
+++ b/documentation/src/main/asciidoc/reference/indexing-massindexer.asciidoc
@@ -335,7 +335,10 @@ Implementations of `MassIndexingFailureHandler` must be thread-safe.
 
 |`environment(MassIndexingEnvironment)`
 |An empty environment (no threadlocals, ...).
-|
+|*This feature is _incubating_: it is still under active development.*
+The contract of incubating elements (e.g. types, methods, configuration properties, etc.)
+may be altered in a backward-incompatible way -- or even removed -- in subsequent releases.
+
 The component responsible for setting up an environment (threadlocals, ...) on mass indexing threads before mass indexing
 starts, and tearing down that environment after mass indexing.
 
--- a/documentation/src/main/asciidoc/reference/indexing-massindexer.asciidoc
+++ b/documentation/src/main/asciidoc/reference/indexing-massindexer.asciidoc
@@ -333,6 +333,15 @@ or for more advanced use cases, such as cancelling mass indexing on the first fa
 
 Implementations of `MassIndexingFailureHandler` must be thread-safe.
 
+|`environment(MassIndexingEnvironment)`
+|An empty environment (no threadlocals, ...).
+|
+The component responsible for setting up an environment (threadlocals, ...) on mass indexing threads before mass indexing
+starts, and tearing down that environment after mass indexing.
+
+Implementations should handle their exceptions unless it is an unrecoverable situation in which further mass indexing
+does not make sense: any exception thrown by the `MassIndexingEnvironment` will abort mass indexing.
+
 |===
 
 [[indexing-massindexer-tuning]]
--- a/integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/massindexing/MassIndexingEnvironmentIT.java
+++ b/integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/massindexing/MassIndexingEnvironmentIT.java
@@ -1,0 +1,184 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.integrationtest.mapper.pojo.massindexing;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import java.lang.invoke.MethodHandles;
+import java.util.Queue;
+import java.util.concurrent.ArrayBlockingQueue;
+
+import org.hibernate.search.engine.backend.work.execution.DocumentCommitStrategy;
+import org.hibernate.search.engine.backend.work.execution.DocumentRefreshStrategy;
+import org.hibernate.search.mapper.pojo.massindexing.MassIndexingEnvironment;
+import org.hibernate.search.integrationtest.mapper.pojo.testsupport.loading.PersistenceTypeKey;
+import org.hibernate.search.integrationtest.mapper.pojo.testsupport.loading.StubLoadingContext;
+import org.hibernate.search.integrationtest.mapper.pojo.testsupport.loading.StubMassLoadingStrategy;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.DocumentId;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.Indexed;
+import org.hibernate.search.mapper.pojo.standalone.mapping.SearchMapping;
+import org.hibernate.search.mapper.pojo.standalone.massindexing.MassIndexer;
+import org.hibernate.search.mapper.pojo.standalone.session.SearchSession;
+import org.hibernate.search.util.common.SearchException;
+import org.hibernate.search.util.common.impl.Futures;
+import org.hibernate.search.util.impl.integrationtest.common.rule.BackendMock;
+import org.hibernate.search.util.impl.integrationtest.mapper.pojo.standalone.StandalonePojoMappingSetupHelper;
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class MassIndexingEnvironmentIT {
+
+	@Rule
+	public final BackendMock backendMock = new BackendMock();
+
+	@Rule
+	public final StandalonePojoMappingSetupHelper setupHelper
+			= StandalonePojoMappingSetupHelper.withBackendMock( MethodHandles.lookup(), backendMock );
+
+	private SearchMapping mapping;
+
+	private final StubLoadingContext loadingContext = new StubLoadingContext();
+
+	@Before
+	public void setup() {
+		backendMock.expectAnySchema( Entity.INDEX );
+
+		mapping = setupHelper.start()
+				.withConfiguration( b -> {
+					b.addEntityType( Entity.class, c -> c
+							.massLoadingStrategy( new StubMassLoadingStrategy<>( Entity.PERSISTENCE_KEY ) ) );
+				} )
+				.setup( Entity.class );
+
+		backendMock.verifyExpectationsMet();
+
+		initData();
+	}
+
+	@Test
+	public void success() throws InterruptedException {
+		try ( SearchSession searchSession = mapping.createSession() ) {
+			Queue<String> before = new ArrayBlockingQueue<>( 10 );
+			Queue<String> after = new ArrayBlockingQueue<>( 10 );
+
+			MassIndexer indexer = searchSession.massIndexer()
+					// Simulate passing information to connect to a DB, ...
+					.context( StubLoadingContext.class, loadingContext )
+					.mergeSegmentsOnFinish( true )
+					.typesToIndexInParallel( 1 )
+					.threadsToLoadObjects( 1 )
+					.environment( new MassIndexingEnvironment() {
+						@Override
+						public void beforeExecution(Context context) {
+							before.add( Thread.currentThread().getName() );
+						}
+
+						@Override
+						public void afterExecution(Context context) {
+							after.add( Thread.currentThread().getName() );
+						}
+					} );
+
+			backendMock.expectWorks(
+							Entity.INDEX, DocumentCommitStrategy.NONE, DocumentRefreshStrategy.NONE
+					)
+					.add( "1", b -> {
+					} );
+
+			// purgeAtStart and mergeSegmentsAfterPurge are enabled by default,
+			// so we expect 1 purge, 1 mergeSegments and 1 flush calls in this order:
+			backendMock.expectIndexScaleWorks( Entity.INDEX, searchSession.tenantIdentifier() )
+					.purge()
+					.mergeSegments()
+					.mergeSegments()
+					.flush()
+					.refresh();
+
+			Futures.unwrappedExceptionGet( indexer.start().toCompletableFuture() );
+
+			assertThat( before ).containsOnly(
+					"Hibernate Search - Mass indexing - Entity - Entity loading - 0",
+					"Hibernate Search - Mass indexing - Entity - ID loading - 0"
+			);
+			assertThat( after ).containsOnly(
+					"Hibernate Search - Mass indexing - Entity - Entity loading - 0",
+					"Hibernate Search - Mass indexing - Entity - ID loading - 0"
+			);
+		}
+
+		backendMock.verifyExpectationsMet();
+	}
+
+	@Test
+	public void failingBeforeHook() throws InterruptedException {
+		try ( SearchSession searchSession = mapping.createSession() ) {
+			Queue<String> after = new ArrayBlockingQueue<>( 10 );
+
+			MassIndexer indexer = searchSession.massIndexer()
+					// Simulate passing information to connect to a DB, ...
+					.context( StubLoadingContext.class, loadingContext )
+					.mergeSegmentsOnFinish( true )
+					.typesToIndexInParallel( 1 )
+					.threadsToLoadObjects( 1 )
+					.environment( new MassIndexingEnvironment() {
+						@Override
+						public void beforeExecution(Context context) {
+							throw new UnsupportedOperationException( "don't call me." );
+						}
+
+						@Override
+						public void afterExecution(Context context) {
+							after.add( Thread.currentThread().getName() );
+						}
+					} );
+
+			// purgeAtStart and mergeSegmentsAfterPurge are enabled by default,
+			// so we expect 1 purge, 1 mergeSegments after that we'll get exceptions so no flush or refresh:
+			backendMock.expectIndexScaleWorks( Entity.INDEX, searchSession.tenantIdentifier() )
+					.purge()
+					.mergeSegments();
+
+			assertThatThrownBy( () -> Futures.unwrappedExceptionGet( indexer.start().toCompletableFuture() ) )
+					.isInstanceOf( SearchException.class )
+					.hasMessageContainingAll(
+							"2 failure(s) occurred during mass indexing",
+							"First failure: don't call me."
+					);
+
+			assertThat( after ).isEmpty();
+		}
+	}
+
+	private void initData() {
+		persist( new Entity( 1 ) );
+	}
+
+	private void persist(Entity entity) {
+		loadingContext.persistenceMap( Entity.PERSISTENCE_KEY ).put( entity.id, entity );
+	}
+
+	@Indexed(index = Entity.INDEX)
+	public static class Entity {
+
+		public static final String INDEX = "Entity";
+		public static final PersistenceTypeKey<Entity, Integer> PERSISTENCE_KEY =
+				new PersistenceTypeKey<>( Entity.class, int.class );
+
+		@DocumentId
+		private int id;
+
+		public Entity() {
+		}
+
+		public Entity(int id) {
+			this.id = id;
+		}
+	}
+}
--- a/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/MassIndexer.java
+++ b/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/MassIndexer.java
@@ -12,6 +12,7 @@ import java.util.concurrent.CompletionStage;
 import org.hibernate.CacheMode;
 import org.hibernate.search.mapper.pojo.massindexing.MassIndexingFailureHandler;
 import org.hibernate.search.mapper.pojo.massindexing.MassIndexingMonitor;
+import org.hibernate.search.mapper.pojo.massindexing.MassIndexingEnvironment;
 import org.hibernate.search.util.common.annotation.Incubating;
 
 /**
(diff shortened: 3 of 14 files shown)


integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/massindexing/MassIndexingBatchingThreadContextIT.java
```diff
@@ -0,0 +1,147 @@
...
+					.flush()
+					.refresh();
+
+			Futures.unwrappedExceptionGet( indexer.start().toCompletableFuture() );
+
+			assertThat( before ).hasValue(
+					0 // Hibernate Search - Mass indexing - Coordinator - 0
+							+ 1 // Hibernate Search - Mass indexing - Workspace - 0
+							+ 1 // Hibernate Search - Mass indexing - Entity - Entity loading - 0
+							+ 1 // Hibernate Search - Mass indexing - Entity - ID loading - 0
```
yrodiere, line 109: Maybe push the thread names to a (threasafe) `List` , so that you can check that explicitly and don't have to guess the threads involved here?

engine/src/main/java/org/hibernate/search/engine/environment/thread/spi/ThreadPoolProvider.java
```diff
@@ -42,9 +61,12 @@ public interface ThreadPoolProvider {
 	 * @param threads the number of threads
 	 * @param threadNamePrefix a label to identify the threads; useful for profiling.
 	 * @param queueSize the size of the queue to store Runnables when all threads are busy
+	 * @param batchingThreadContext a thread context that can be used to manipulate the execution environment, for example,
+	 * to initialize {@link ThreadLocal ThreadLocals}.
 	 * @return the new ExecutorService
 	 */
-	ThreadPoolExecutor newFixedThreadPool(int threads, String threadNamePrefix, int queueSize);
+	@Incubating
+	ThreadPoolExecutor newFixedThreadPool(int threads, String threadNamePrefix, int queueSize, BatchingThreadContext batchingThreadContext);
```
yrodiere, line 69: I think you got this wrong; *this* method should have a default implementation. Otherwise you will break existing implementations of this SPI.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoDefaultMassIndexer.java
```diff
@@ -122,8 +124,11 @@ public PojoDefaultMassIndexer monitor(MassIndexingMonitor monitor) {
 	public CompletableFuture<?> start() {
 		PojoMassIndexingBatchCoordinator coordinator = createCoordinator();
 		ExecutorService executor = mappingContext.threadPoolProvider()
-				.newFixedThreadPool( 1,
-						PojoMassIndexingBatchIndexingWorkspace.THREAD_NAME_PREFIX + "Coordinator" );
+				.newFixedThreadPool(
+						1,
+						PojoMassIndexingBatchIndexingWorkspace.THREAD_NAME_PREFIX + "Coordinator"
+						// todo: double check if we need any context around the coordinator. seems like no, but .. :)
```
yrodiere, line 130: I don't think so. IMO the context is only useful for threads where we list identifiers or load entity data.

engine/src/main/java/org/hibernate/search/engine/environment/thread/spi/BatchingThreadContext.java
```diff
@@ -0,0 +1,43 @@
...
+		@Override
+		public void afterExecution(Runnable r, Throwable t) {
+			// Do nothing.
+		}
+	};
+
+	/**
+	 * Method is going to be invoked prior to executing the {@link Runnable} in the given thread.
+	 */
+	void beforeExecution(Thread t, Runnable r);
```
yrodiere, line 36: It's probably not that useful to pass the `Thread` because it can be accessed through `Thread.currentThread()` (and I'm not sure why one would do that).

I doubt it's useful to pass the `Runnable`, and that could actually be harmful (what if the user calls `r.run()`?).

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingEntityIdentifierLoadingRunnable.java
```diff
@@ -30,10 +31,10 @@
 	private final String tenantId;
 
 	public PojoMassIndexingEntityIdentifierLoadingRunnable(PojoMassIndexingNotifier notifier,
-			PojoMassIndexingIndexedTypeGroup<E> typeGroup,
+			BatchingThreadContext treadContext, PojoMassIndexingIndexedTypeGroup<E> typeGroup,
```
yrodiere, line 34: Typo.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingEntityLoadingRunnable.java
```diff
@@ -39,11 +40,11 @@
 	private final String tenantId;
 
 	protected PojoMassIndexingEntityLoadingRunnable(PojoMassIndexingNotifier notifier,
-			PojoMassIndexingIndexedTypeGroup<E> typeGroup,
+			BatchingThreadContext treadContext, PojoMassIndexingIndexedTypeGroup<E> typeGroup,
```
yrodiere, line 43: Typo.

yrodiere (commented):
Thanks for working on this!

It looks like it could work, but I'm a bit perplexed about doing this at such a low level. There are plenty of threads involved in mass indexing (e.g. the backend threads) where the context will be ignored, so it's not like we need to go to such a low level. And the change involves breaking SPIs (or at least changing them and deprecating stuff), so we need a good reason.

On the other hand, you could simply execute the before/after in the runnables (essentially `org.hibernate.search.mapper.pojo.massindexing.impl.PojoMassIndexingEntityLoadingRunnable#PojoMassIndexingEntityLoadingRunnable` + `org.hibernate.search.mapper.pojo.massindexing.impl.PojoMassIndexingEntityIdentifierLoadingRunnable`, or maybe in the parent class `PojoMassIndexingFailureHandledRunnable`) and be done with it. This might perform slightly worse, but not a lot, and we could then evolve the feature to pass relevant information to the user-provided objects (what type are we going to index, ...).

yrodiere:
> should we be more granular and allow a user to tell us what to do when loading identifiers and what to do when loading entities

I *personally* don't think that's useful, but what do I know. I guess you could pass a `*Context` object as an argument to the `before`/`after` methods and we can always add such information there, eventually? `BatchingThreadContext` might have to be renamed though, otherwise we'd be passing context to a context, and that becomes slightly confusing :)

engine/src/main/java/org/hibernate/search/engine/environment/thread/spi/BatchingThreadContext.java
marko-bekhta: right ... that's just what the `ThreadPoolExecutor` uses for the arguments ...

> what if the user calls r.run()?

well ... that would be a shame 😄

marko-bekhta:
> you could simply execute the before/after in the runnables

I think I had that idea at some point, but then the discussion on the ticket led me to believe that the intention was to do it on the `Executor` level 😃.
Going to make some changes then 😃

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/MassIndexerThreadLifecycleHooksConfigurer.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.hibernate.search.util.common.annotation.Incubating;
+
+/**
+ * Configurer that exposes thread lifecycle hooks that can be used to manipulate the execution environment, for example,
+ * to initialize {@link ThreadLocal ThreadLocals}. See interfaces extending {@link Context} to learn which threads will
+ * attempt to execute configured hooks.
+ */
+@Incubating
+public interface MassIndexerThreadLifecycleHooksConfigurer {
```
yrodiere, line 17: That's a mouthful. Also, other configurers typically have a `configure` method, and other classes related to the `MassIndexer` typically use the `MassIndexing` prefix.

A few suggestions:

* `MassIndexingThreadScope` (inspired by JDK 20's "scoped values": https://openjdk.org/jeps/429)
* Simply `MassIndexingThreadEnvironment` or even `MassIndexingEnvironment`
* Something else?

I'd favor `MassIndexingEnvironment` since it's short and simple, WDYT?

marko-bekhta: yup, that would be a better name 👍🏻 
I haven't changed the `supportsThreadLifecycleHooks()` as it seems more appropriate in the context of `PojoMassIndexingFailureHandledRunnable`. 
Added a row to the ref doc table now that the name should be more or less stable 😃 
Also, do not trust IDE to do refactoring correctly all the time. I've ended up with `MassIndexingMassIndexingMassIndexingEnvironmentIT` 😄

documentation/src/main/asciidoc/reference/indexing-massindexer.asciidoc
```diff
@@ -333,6 +333,14 @@ or for more advanced use cases, such as cancelling mass indexing on the first fa
 
 Implementations of `MassIndexingFailureHandler` must be thread-safe.
 
+|`massIndexingEnvironment(MassIndexingEnvironment)`
```
yrodiere, line 336: If we want to be consistent with other pluggable components (monitor, failure handler) we should not use the prefix for this method, just for the interface.

```suggestion
|`environment(MassIndexingEnvironment)`
```

documentation/src/main/asciidoc/reference/indexing-massindexer.asciidoc
```diff
@@ -333,6 +333,14 @@ or for more advanced use cases, such as cancelling mass indexing on the first fa
...
+|A mass indexing environment that has no additional actions configured.
```
yrodiere, line 337:
```suggestion
|An empty environment (no threadlocals, ...).
```

documentation/src/main/asciidoc/reference/indexing-massindexer.asciidoc
```diff
@@ -333,6 +333,14 @@ or for more advanced use cases, such as cancelling mass indexing on the first fa
...
+|
+The component responsible for setting before and after actions for the threads doing mass indexing.
```
yrodiere, line 339:
```suggestion
The component responsible for setting up an environment (threadlocals, ...) on mass indexing threads before mass indexing starts, and tearing down that environment after mass indexing.
```

documentation/src/main/asciidoc/reference/indexing-massindexer.asciidoc
```diff
@@ -333,6 +333,14 @@ or for more advanced use cases, such as cancelling mass indexing on the first fa
...
+
+Implementations should handle their exceptions unless it is an unrecoverable situation in which further mass indexing
+does not make sense.
```
yrodiere, line 342:
```suggestion
Implementations should handle their exceptions unless it is an unrecoverable situation in which further mass indexing
does not make sense: any exception thrown by the `MassIndexingEnvironment` will abort mass indexing.
```

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/massindexing/MassIndexingEnvironmentIT.java
```diff
@@ -0,0 +1,184 @@
...
+			Queue<String> before = new ArrayBlockingQueue<>( 10 );
+			Queue<String> after = new ArrayBlockingQueue<>( 10 );
+
+			MassIndexer indexer = searchSession.massIndexer()
+					// Simulate passing information to connect to a DB, ...
+					.context( StubLoadingContext.class, loadingContext )
+					.mergeSegmentsOnFinish( true )
+					.typesToIndexInParallel( 1 )
+					.threadsToLoadObjects( 1 )
+					.massIndexingEnvironment( new MassIndexingEnvironment() {
```
yrodiere, line 77:
```suggestion
					.environment( new MassIndexingEnvironment() {
```

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/massindexing/MassIndexingEnvironmentIT.java
```diff
@@ -0,0 +1,184 @@
...
+		try ( SearchSession searchSession = mapping.createSession() ) {
+			Queue<String> after = new ArrayBlockingQueue<>( 10 );
+
+			MassIndexer indexer = searchSession.massIndexer()
+					// Simulate passing information to connect to a DB, ...
+					.context( StubLoadingContext.class, loadingContext )
+					.mergeSegmentsOnFinish( true )
+					.typesToIndexInParallel( 1 )
+					.threadsToLoadObjects( 1 )
+					.massIndexingEnvironment( new MassIndexingEnvironment() {
```
yrodiere, line 130:
```suggestion
					.environment( new MassIndexingEnvironment() {
```

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/MassIndexer.java
```diff
@@ -194,4 +195,16 @@ public interface MassIndexer {
...
+	 * Sets the {@link MassIndexingEnvironment} which allows manipulating the thread execution lifecycle.
+	 *
+	 * @param massIndexingEnvironment a mass indexing environment that allows to manipulate
+	 * the thread execution, for example, to initialize {@link ThreadLocal ThreadLocals}.
+	 * @return {@code this} for method chaining
+	 *
+	 * @see MassIndexingEnvironment
+	 */
+	@Incubating
+	MassIndexer massIndexingEnvironment(MassIndexingEnvironment massIndexingEnvironment);
```
yrodiere, line 209:
```suggestion
	MassIndexer environment(MassIndexingEnvironment massIndexingEnvironment);
```

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/MassIndexer.java
```diff
@@ -194,4 +195,16 @@ public interface MassIndexer {
 	 * @return {@code this} for method chaining
 	 */
 	MassIndexer failureHandler(MassIndexingFailureHandler failureHandler);
+
+	/**
+	 * Sets the {@link MassIndexingEnvironment} which allows manipulating the thread execution lifecycle.
+	 *
+	 * @param massIndexingEnvironment a mass indexing environment that allows to manipulate
+	 * the thread execution, for example, to initialize {@link ThreadLocal ThreadLocals}.
+	 * @return {@code this} for method chaining
```
yrodiere, line 204:
```suggestion
	 * Sets the {@link MassIndexingEnvironment}, which can set up an environment (thread locals, ...) in mass indexing threads.
	 *
	 * @param massIndexingEnvironment a component that gets a chance to
	 * set up e.g. {@link ThreadLocal ThreadLocals} in mass indexing threads before
	 * mass indexing starts, and to remove them after mass indexing stops.
	 * @return {@code this} for method chaining
```

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/HibernateOrmMassIndexer.java
```diff
@@ -122,4 +123,10 @@ public MassIndexer failureHandler(MassIndexingFailureHandler failureHandler) {
 	ConditionalExpression reindexOnly(Class<?> type, String conditionalExpression) {
 		return context.reindexOnly( type, conditionalExpression );
 	}
+
+	@Override
+	public MassIndexer massIndexingEnvironment(MassIndexingEnvironment massIndexingEnvironment) {
+		delegate.massIndexingEnvironment( massIndexingEnvironment );
```
yrodiere, line 129:
```suggestion
	public MassIndexer environment(MassIndexingEnvironment massIndexingEnvironment) {
		delegate.environment( massIndexingEnvironment );
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/MassIndexingEnvironment.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.mapper.pojo.massindexing;
+
+import org.hibernate.search.util.common.annotation.Incubating;
+
+/**
+ * Massindexing environment interface that exposes thread lifecycle hooks that can be used to manipulate the execution environment,
```
yrodiere, line 12:
```suggestion
 * An interface for pluggable components that set up and tear down the environment of mass indexing threads,
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoDefaultMassIndexer.java
```diff
@@ -186,6 +200,12 @@ public PojoDefaultMassIndexer failureHandler(MassIndexingFailureHandler failureH
 		return this;
 	}
 
+	@Override
+	public PojoMassIndexer massIndexingEnvironment(MassIndexingEnvironment massIndexingEnvironment) {
```
yrodiere, line 204:
```suggestion
	public PojoMassIndexer environment(MassIndexingEnvironment massIndexingEnvironment) {
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/spi/PojoMassIndexer.java
```diff
@@ -130,4 +133,16 @@ public interface PojoMassIndexer {
 	 * @return {@code this} for method chaining
 	 */
 	PojoMassIndexer failureHandler(MassIndexingFailureHandler failureHandler);
+
+	/**
+	 * Sets the {@link MassIndexingEnvironment} which allows manipulating the thread execution lifecycle.
+	 *
+	 * @param massIndexingEnvironment a mass indexing environment that allows to manipulate
+	 * the thread execution, for example, to initialize {@link ThreadLocal ThreadLocals}.
```
yrodiere, line 141:
```suggestion
	 * Sets the {@link MassIndexingEnvironment}, which can set up an environment (thread locals, ...) in mass indexing threads.
	 *
	 * @param massIndexingEnvironment a component that gets a chance to
	 * set up e.g. {@link ThreadLocal ThreadLocals} in mass indexing threads before
	 * mass indexing starts, and to remove them after mass indexing stops.
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/spi/PojoMassIndexer.java
```diff
@@ -130,4 +133,16 @@ public interface PojoMassIndexer {
...
+	 * @return {@code this} for method chaining
+	 *
+	 * @see MassIndexingEnvironment
+	 */
+	@Incubating
+	PojoMassIndexer massIndexingEnvironment(MassIndexingEnvironment massIndexingEnvironment);
```
yrodiere, line 147:
```suggestion
	PojoMassIndexer environment(MassIndexingEnvironment massIndexingEnvironment);
```

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/massindexing/MassIndexer.java
```diff
@@ -157,4 +158,15 @@ public interface MassIndexer {
 	 */
 	<T> MassIndexer context(Class<T> contextType, T context);
 
+	/**
+	 * Sets the {@link MassIndexingEnvironment} which allows manipulating the thread execution lifecycle.
+	 *
+	 * @param massIndexingEnvironment a mass indexing environment that allows to manipulate
+	 * the thread execution, for example, to initialize {@link ThreadLocal ThreadLocals}.
```
yrodiere, line 165:
```suggestion
	 * Sets the {@link MassIndexingEnvironment}, which can set up an environment (thread locals, ...) in mass indexing threads.
	 *
	 * @param massIndexingEnvironment a component that gets a chance to
	 * set up e.g. {@link ThreadLocal ThreadLocals} in mass indexing threads before
	 * mass indexing starts, and to remove them after mass indexing stops.
```

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/massindexing/impl/StandalonePojoMassIndexer.java
```diff
@@ -94,4 +95,9 @@ public <T> MassIndexer context(Class<T> contextType, T context) {
 		return this;
 	}
 
+	@Override
+	public MassIndexer massIndexingEnvironment(MassIndexingEnvironment massIndexingEnvironment) {
+		delegate.massIndexingEnvironment( massIndexingEnvironment );
```
yrodiere, line 100:
```suggestion
	public MassIndexer environment(MassIndexingEnvironment massIndexingEnvironment) {
		delegate.environment( massIndexingEnvironment );
```

yrodiere (commented):
Just a few suggestions for the documentation, which seems to have a lot of leftovers from the previous approach which was named "thread lifecycle hooks".

Also I think the setter on `MassIndexer` shouldn't be prefixed with `massIndexing*`; the context is already obvious in that case, which is why we don't use the `massIndexing*` prefix for other methods.

See my suggestions below.

documentation/src/main/asciidoc/reference/indexing-massindexer.asciidoc
marko-bekhta: yeah, I had that intent initially, but `environment` seems to be too generic and sounds "too powerful", but if it's just me then I'm happy to make a change 😄

yrodiere (approved):

yrodiere:
I added something to mention `.environment()` is incubating in the reference documentation.

Merging now, thanks!

marko-bekhta:
thanks!

PR #3398, opened 2023-01-31
HSEARCH-4772 Add failsafe to prevent log flooding for mass indexing with lots of non-critical failures

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4772

The initial idea was to allow users to set the limit on the `MassIndexer` interface level as that could be something that can change from mass indexer to mass indexer. And then, the work would happen in the notifier. (you'll find this in the first commit). But I didn't like that, as the notifier doesn't look like a good place to decide if something should or shouldn't be logged or handled otherwise. So that's when I've pushed this logic into our own `MassIndexingFailureHandler` (second commit). I liked this approach a tad better, but then if users don't like our default max number - then they would need to come up with their own handler 😃

### DIFF
--- a/documentation/src/main/asciidoc/reference/indexing-massindexer.asciidoc
+++ b/documentation/src/main/asciidoc/reference/indexing-massindexer.asciidoc
@@ -345,6 +345,17 @@ starts, and tearing down that environment after mass indexing.
 Implementations should handle their exceptions unless it is an unrecoverable situation in which further mass indexing
 does not make sense: any exception thrown by the `MassIndexingEnvironment` will abort mass indexing.
 
+|`failureFloodingThreshold(long)`
+| `100` with the default failure handler (see description)
+|*This feature is _incubating_: it is still under active development.*
+The maximum number of failures to be handled per indexed type.
+Any failures exceeding this number will be ignored and not sent for processing by `MassIndexingFailureHandler`.
+Can be set to `Long.MAX_VALUE` if none of the failures should be ignored.
+
+Defaults to a threshold defined by the failure handler in use; see `MassIndexingFailureHandler#failureFloodingThreshold`,
+`FailureHandler#failureFloodingThreshold`.
+For the default log-based failure handler, the default threshold is 100.
+
 |===
 
 [[indexing-massindexer-tuning]]
--- a/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingDelegatingFailureHandler.java
+++ b/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingDelegatingFailureHandler.java
@@ -39,4 +39,9 @@ public class PojoMassIndexingDelegatingFailureHandler implements MassIndexingFai
 		}
 		delegate.handle( builder.build() );
 	}
+
+	@Override
+	public long failureFloodingThreshold() {
+		return delegate.failureFloodingThreshold();
+	}
 }
--- a/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingNotifier.java
+++ b/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingNotifier.java
@@ -7,7 +7,12 @@
 package org.hibernate.search.mapper.pojo.massindexing.impl;
 
 import java.lang.invoke.MethodHandles;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.atomic.LongAdder;
 
@@ -16,8 +21,8 @@ import org.hibernate.search.mapper.pojo.massindexing.MassIndexingEntityFailureCo
 import org.hibernate.search.mapper.pojo.massindexing.MassIndexingFailureContext;
 import org.hibernate.search.mapper.pojo.massindexing.MassIndexingFailureHandler;
 import org.hibernate.search.mapper.pojo.massindexing.MassIndexingMonitor;
-import org.hibernate.search.util.common.logging.impl.LoggerFactory;
 import org.hibernate.search.mapper.pojo.massindexing.spi.PojoMassIndexingSessionContext;
+import org.hibernate.search.util.common.logging.impl.LoggerFactory;
 
 /**
  * A central object to which various are reported,
@@ -33,11 +38,15 @@ public class PojoMassIndexingNotifier {
 	private final AtomicReference<RecordedFailure> firstFailure =
 			new AtomicReference<>( null );
 	private final LongAdder failureCount = new LongAdder();
+	private final Map<String, AtomicLong> failureCounts = Collections.synchronizedMap( new HashMap<>() );
+	private final long failureFloodingThreshold;
 
 	public PojoMassIndexingNotifier(
-			MassIndexingFailureHandler failureHandler, MassIndexingMonitor monitor) {
+			MassIndexingFailureHandler failureHandler, MassIndexingMonitor monitor, Long failureFloodingThreshold) {
 		this.failureHandler = failureHandler;
 		this.monitor = monitor;
+		this.failureFloodingThreshold = Optional.ofNullable( failureFloodingThreshold )
+				.orElseGet( failureHandler::failureFloodingThreshold );
 	}
 
 	void reportAddedTotalCount(long totalCount) {
@@ -85,13 +94,19 @@ public class PojoMassIndexingNotifier {
 
 	void reportEntityIndexingFailure(PojoMassIndexingIndexedTypeGroup<?> typeGroup,
 			PojoMassIndexingSessionContext sessionContext, Object entity, Exception exception) {
+		String failingOperation = log.massIndexerIndexingInstance( typeGroup.notifiedGroupName() );
+
 		// Don't record these failures as suppressed beyond the first one, because there may be hundreds of them.
 		RecordedFailure recordedFailure = recordFailure( exception, false );
+		// We want to check this after recording failure above to make sure that total failure count is increased.
+		if ( shouldNotBeReported( failingOperation ) ) {
+			return;
+		}
 
 		MassIndexingEntityFailureContext.Builder contextBuilder = MassIndexingEntityFailureContext.builder();
 		contextBuilder.throwable( exception );
 		// Add minimal information here, but information we're sure we can get
-		contextBuilder.failingOperation( log.massIndexerIndexingInstance( typeGroup.notifiedGroupName() ) );
+		contextBuilder.failingOperation( failingOperation );
 		// Add more information here, but information that may not be available if the session completely broke down
 		// (we're being extra careful here because we don't want to throw an exception while handling and exception)
 		Object entityReference = extractReferenceOrSuppress( typeGroup, sessionContext, entity, exception );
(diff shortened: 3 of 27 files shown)


mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/spi/PojoMassIndexer.java
```diff
@@ -130,4 +131,20 @@ public interface PojoMassIndexer {
...
+	 * By default, this number is set to {@code 1000}.
+	 *
+	 * @param maximumFailuresHandledPerOperation The number representing the maximum failures to be handled per operation.
+	 * @return {@code this} for method chaining
+	 */
+	@Incubating
+	default PojoMassIndexer maximumFailuresHandledPerOperation(long maximumFailuresHandledPerOperation) {
+		throw new UnsupportedOperationException(
+				this.getClass() + " does not support setting the `maximumFailuresHandledPerOperation`. Implement it first." );
+	}
```
yrodiere: No need for a default method here, this SPI is expected to be *called* by mappers, not implemented. 

I see how the name `PojoDefaultMassIndexer` makes that unclear, it should probably be called `PojoMassIndexerImpl`. But, well, that's not our concern right.

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/massindexing/MassIndexer.java
```diff
@@ -143,6 +143,19 @@ public interface MassIndexer {
 	 */
 	MassIndexer failureHandler(MassIndexingFailureHandler failureHandler);
 
+	/**
+	 * Sets the maximum number of failures that will be reported and sent to {@link MassIndexingFailureHandler} per indexing operation.
+	 * Any failures exceeding this number will be ignored. A count of such ignored failures together with the operation they belong to
+	 * will be logged upon the completion of indexing process.
```
yrodiere: It's not very clear what you mean by "indexing operation". Maybe use "indexed type" instead? That's not strictly true, but at least the meaning is clearer.

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/massindexing/MassIndexer.java
```diff
@@ -143,6 +143,19 @@ public interface MassIndexer {
...
+	 * <p>
+	 * By default, this number is set to {@code 1000}.
+	 *
+	 * @param maximumFailuresHandledPerOperation The number representing the maximum failures to be handled per operation.
+	 * @return {@code this} for method chaining
+	 */
+	@Incubating
+	MassIndexer maximumFailuresHandledPerOperation(long maximumFailuresHandledPerOperation);
```
yrodiere: I would stay vague in the method name, because it's one of those cases where adding more info is just confusing. Maybe `failureLimit`? `failureFloodingThreshold`?

documentation/src/main/asciidoc/migration/index.asciidoc
```diff
@@ -415,4 +415,8 @@ query that returns the same results.
 Possible changes can include: some clauses can be rearranged, nested bool queries might be flattened, a bool query might be
 replaced with its clause.
 
-A bool query with a single `mustNot` clause and applied boost would implicitly add a `must` with `match_all` clause.
\ No newline at end of file
+A bool query with a single `mustNot` clause and applied boost would implicitly add a `must` with `match_all` clause.
+
+Due to changes in default failure handler used for mass indexing (`MassIndexingFailureHandler`)
+only first 1000 failures per operation will be reported. Any remaining failures will not be propagated to the
+background failure handler. To achieve previous behavior - pass in your own `MassIndexingFailureHandler`.
```
yrodiere, line 422:
```suggestion
Due to changes in the default failure handler used for mass indexing (`MassIndexingFailureHandler`)
only the first 1000 failures per operation will be reported. Any remaining failures will not be propagated to the
failure handler. To restore the previous behavior, pass in your own `MassIndexingFailureHandler`.
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingDelegatingFailureHandler.java
```diff
@@ -39,4 +56,44 @@ public void handle(MassIndexingEntityFailureContext context) {
...
+								unreported,
+								failure.firstThrowable.getMessage(),
+								failure.firstThrowable
+						)
+				);
+				builder.failingOperation( entry.getKey() );
+				delegate.handle( builder.build() );
+			}
+		}
+	}
```
yrodiere, line 80: I don't know how I feel about this.

On the one hand, you're right that people using a custom mass indexing failure handler probably shouldn't get that flooding protection by default, because they might rely on getting *all* errors.

On the other hand, they might want to opt-in, and with this solution, they can't. Conversely, they might want to opt out and keep using the default logging failure handler (the `delegate` in this class), and with this solution, they can't.

Another problem that bothers me: people setting a background failure handler in their configuration properties will get this flooding protection, but they might not want to.

To solve that, I think we need the following behavior:

* Move back the flooding protection logic to the notifier.
* Determine the actual limit based on a default provided by failure handlers, allowing users to override that through the `MassIndexer`:
  * The `FailureHandler` defines a `maxFailureCountPerOperation` through a getter.
  * The `MassIndexingFailureHandler` defines a `maxFailureCountPerOperation` through a getter.  `PojoMassIndexingDelegatingFailureHandler` just delegates to `FailureHandler`.
  * Those getters have a default implementation that returns `Long.MAX_VALUE`.
  * The getter of `LogFailureHandler` returns, say, 100.
  * By default we use the maximum provided by the failure handler, except when the user overrides it explicitly on the `MassIndexer`.

As to the signature of the getter... something like this?

```
/**
 * @param threshold The number of failures during one mass indexing beyond which the
 * failure handler will no longer be notified. This threshold is reached separately for each indexed type.
 * May be overridden by mass indexer parameters
 * (see {@code failureFloodingThreshold(long)} in the {@code MassIndexer} interface).
 */
default long failureFloodingThreshold() {
    return Long.MAX_VALUE;
}
```

And then in `MassIndexer`:

```
/**
 * @param threshold The number of failures during one mass indexing beyond which the
 * failure handler will no longer be notified. This threshold is reached separately for each indexed type.
 * Overrides the {@link MassIndexingFailureHandler#failureFloodingThreshold threshold defined by the failure handler itself}.
 * Defaults to {@code 100} with the default failure handler.
 * @return {@code this}, for method chaining.
 */
MassIndexer failureFloodingThreshold(long threshold);
```

WDYT?

yrodiere (commented):
Thanks.

> But I didn't like that, as the notifier doesn't look like a good place to decide if something should or shouldn't be logged or handled otherwise.

That's a good point. I agree, but we can probably find a better middle ground. See my comments below.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingDelegatingFailureHandler.java
```diff
@@ -39,4 +40,9 @@ public void handle(MassIndexingEntityFailureContext context) {
 		}
 		delegate.handle( builder.build() );
 	}
+
+	@Override
+	public long failureFloodingThreshold() {
+		return FAILURE_FLOODING_THRESHOLD;
```
yrodiere, line 46: I think it would be better to call `delegate.failureFloodingThreshold()` here, and to implement the default of `100` in `LogFailureHandler`. I think I mentioned this in my comment?

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingNotifier.java
```diff
@@ -130,6 +162,25 @@ void reportIndexingCompleted() {
 			return;
 		}
 
+		// report other failures that went over the limits:
```
yrodiere, line 165: That's a good idea, but reporting those failures is exactly what we *don't* want :)

I think you meant this?

```suggestion
		// report that some failures went unreported
```

documentation/src/main/asciidoc/reference/indexing-massindexer.asciidoc
```diff
@@ -345,6 +345,12 @@ starts, and tearing down that environment after mass indexing.
 Implementations should handle their exceptions unless it is an unrecoverable situation in which further mass indexing
 does not make sense: any exception thrown by the `MassIndexingEnvironment` will abort mass indexing.
 
+|`failureFloodingThreshold(long)`
+| -
```
yrodiere, line 349:
```suggestion
| `100` with the default failure handler (see description)
```

documentation/src/main/asciidoc/reference/indexing-massindexer.asciidoc
```diff
@@ -345,6 +345,12 @@ starts, and tearing down that environment after mass indexing.
...
+|The number representing the maximum failures to be handled per indexed type.
+Any failures exceeding this number will be ignored and not sent for processing by `MassIndexingFailureHandler`.
+Can be set to `Long.MAX_VALUE` if none of the failures should be ignored.
+
```
yrodiere, line 353:
```suggestion
+
Defaults to a threshold defined by the failure handler in use; see `MassIndexingFailureHandler#failureFloodingThreshold`, `FailureHandler#failureFloodingThreshold`.
For the default log-based failure handler, the default threshold is 100.

```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingDelegatingFailureHandler.java
marko-bekhta: hmm ... that would mean that we need to push the `failureFloodingThreshold()` to `org.hibernate.search.engine.reporting.FailureHandler`. But that is a more global handler and not mass indexer specific ... I'll keep it as is, and we can return to this on Monday 😃

yrodiere: Yes you would need to *also* have that method on `org.hibernate.search.engine.reporting.FailureHandler`.
That makes sense because `FailureHandler` is the default failure handler in mass indexing.

marko-bekhta: Ohh ok. What concerned me was that this `org.hibernate.search.engine.reporting.FailureHandler` is also used for those `IndexingPlanSynchronizationStrategy`ies, but the threshold is only applicable to mass-indexing (at this moment). Maybe for `org.hibernate.search.engine.reporting.FailureHandler`, we call it `massIndexingFailureFloodingThreshold()` or something along those lines 😃

yrodiere: If you want to, but I'll be fine with just `failureFloodingThreshold()`. We can always expand the principle of failure flooding protection to other areas later if it makes sense. For `IndexingPlanSynchronizationStrategy`, it doesn't make sense, really, since there's only ever one failure being reported in the end.

marko-bekhta: I've also added `@Incubating` to `FailureHandler` new methods and mentioned that `failureFloodingThreshold(long)` is incubating in the docs.

yrodiere (approved):
Thanks, LGTM

yrodiere:
Tests seem to be failing :)

marko-bekhta:
Well, of course, they are 😢 😭 if someone was running the tests locally as:
```
mvn clean install -am -pl :hibernate-search-integrationtest-mapper-orm,:hibernate-search-mapper-pojo-base,:hibernate-search-integrationtest-mapper-pojo-standalone-realbackend,:hibernate-search-integrationtest-mapper-orm-realbackend
```

PR #3399, opened 2023-02-01
HSEARCH-1809 Follow-up: Only make the failure within one operation for predictable results

marko-bekhta:

We have this in the coordinator:
```java
Futures.unwrappedExceptionGet( Futures.firstFailureOrAllOf( allFutures ) );
```
And from what it looks on CI:
```
Expecting throwable message:
  "HSEARCH700042: 1 failure(s) occurred during mass indexing. See the logs for details. First failure: don't call me."
```
It managed only to throw one exception before the coordinator thread took over and stopped everything. 
Hence I've made a slight change to only fail in one runnable to have a predictable result. An alternative could've been to remove the `2` from the message check and just look for
```
 failure(s) occurred during mass indexing",
```


yrodiere (approved):

marko-bekhta:
Thanks for reviewing, going to merge😃

PR #3400, opened 2023-02-01
HSEARCH-4736 Add offloading operation submitter

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4736

I've tried to keep `OperationSubmitter` closed for the users, so they are forced to use whatever is provided with Search. 
Also, with this new submitter type, it is somewhat possible to offload work in `LuceneSyncWorkOrchestratorImpl` but I wasn't sure if that's something worth supporting, so I left it more on the "idea level".

I've kept the same "hacks" for the BatchExecutor test as in the blocking case as it seems we might stumble upon the same issue with offloading.

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/ElasticsearchBatchingWorkOrchestrator.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/ElasticsearchBatchingWorkOrchestrator.java
@@ -57,7 +57,7 @@ public class ElasticsearchBatchingWorkOrchestrator
 	private final BackendThreads threads;
 	private final FailureHandler failureHandler;
 
-	private HashTable<BatchingExecutor<ElasticsearchBatchedWorkProcessor>> executors;
+	private HashTable<BatchingExecutor<ElasticsearchBatchedWorkProcessor, ElasticsearchBatchedWork<?>>> executors;
 
 	/**
 	 * @param name The name of the orchestrator thread (and of this orchestrator when reporting errors)
@@ -102,7 +102,7 @@ public class ElasticsearchBatchingWorkOrchestrator
 			) );
 		}
 
-		for ( BatchingExecutor<?> executor : executors ) {
+		for ( BatchingExecutor<?, ?> executor : executors ) {
 			executor.start( threads.getWorkExecutor() );
 		}
 	}
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/ElasticsearchBatchingWorkOrchestrator.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/ElasticsearchBatchingWorkOrchestrator.java
@@ -7,7 +7,6 @@
 package org.hibernate.search.backend.elasticsearch.orchestration.impl;
 
 import java.util.concurrent.CompletableFuture;
-import java.util.function.Function;
 
 import org.hibernate.search.backend.elasticsearch.cfg.ElasticsearchIndexSettings;
 import org.hibernate.search.backend.elasticsearch.link.impl.ElasticsearchLink;
@@ -98,7 +97,8 @@ public class ElasticsearchBatchingWorkOrchestrator
 					processor,
 					queueSize,
 					true,
-					failureHandler
+					failureHandler,
+					blockingRetryProducer
 			) );
 		}
 
@@ -108,9 +108,8 @@ public class ElasticsearchBatchingWorkOrchestrator
 	}
 
 	@Override
-	protected void doSubmit(ElasticsearchBatchedWork<?> work, OperationSubmitter operationSubmitter,
-			Function<ElasticsearchBatchedWork<?>, Runnable> blockingRetryProducer) throws InterruptedException {
-		executors.get( work.getQueuingKey() ).submit( work, operationSubmitter, blockingRetryProducer );
+	protected void doSubmit(ElasticsearchBatchedWork<?> work, OperationSubmitter operationSubmitter) throws InterruptedException {
+		executors.get( work.getQueuingKey() ).submit( work, operationSubmitter );
 	}
 
 	@Override
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/ElasticsearchBatchingWorkOrchestrator.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/ElasticsearchBatchingWorkOrchestrator.java
@@ -7,6 +7,7 @@
 package org.hibernate.search.backend.elasticsearch.orchestration.impl;
 
 import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
 
 import org.hibernate.search.backend.elasticsearch.cfg.ElasticsearchIndexSettings;
 import org.hibernate.search.backend.elasticsearch.link.impl.ElasticsearchLink;
@@ -107,8 +108,13 @@ public class ElasticsearchBatchingWorkOrchestrator
 	}
 
 	@Override
-	protected void doSubmit(ElasticsearchBatchedWork<?> work, OperationSubmitter operationSubmitter) throws InterruptedException {
-		executors.get( work.getQueuingKey() ).submit( work, operationSubmitter );
+	protected void doSubmit(ElasticsearchBatchedWork<?> work, OperationSubmitter operationSubmitter,
+			Function<ElasticsearchBatchedWork<?>, Runnable> blockingRetryProducer) throws InterruptedException {
+		executors.get( work.getQueuingKey() ).submit(
+				work,
+				operationSubmitter,
+				w -> blockingRetryProducer.apply( (ElasticsearchBatchedWork<?>) w )
+		);
 	}
 
 	@Override
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/ElasticsearchBatchingWorkOrchestrator.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/ElasticsearchBatchingWorkOrchestrator.java
@@ -110,11 +110,7 @@ public class ElasticsearchBatchingWorkOrchestrator
 	@Override
 	protected void doSubmit(ElasticsearchBatchedWork<?> work, OperationSubmitter operationSubmitter,
 			Function<ElasticsearchBatchedWork<?>, Runnable> blockingRetryProducer) throws InterruptedException {
-		executors.get( work.getQueuingKey() ).submit(
-				work,
-				operationSubmitter,
-				w -> blockingRetryProducer.apply( (ElasticsearchBatchedWork<?>) w )
-		);
+		executors.get( work.getQueuingKey() ).submit( work, operationSubmitter, blockingRetryProducer );
 	}
 
 	@Override
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/orchestration/impl/LuceneParallelWorkOrchestratorImpl.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/orchestration/impl/LuceneParallelWorkOrchestratorImpl.java
@@ -8,7 +8,6 @@ package org.hibernate.search.backend.lucene.orchestration.impl;
 
 import java.lang.invoke.MethodHandles;
 import java.util.concurrent.CompletableFuture;
-import java.util.function.Function;
 
 import org.hibernate.search.backend.lucene.logging.impl.Log;
 import org.hibernate.search.backend.lucene.lowlevel.index.impl.IndexAccessor;
@@ -65,8 +64,7 @@ public class LuceneParallelWorkOrchestratorImpl
 	}
 
 	@Override
-	protected void doSubmit(WorkExecution<?> workExecution, OperationSubmitter operationSubmitter,
-			Function<WorkExecution<?>, Runnable> blockingRetryProducer) {
+	protected void doSubmit(WorkExecution<?> workExecution, OperationSubmitter operationSubmitter) {
 		if ( !OperationSubmitter.blocking().equals( operationSubmitter ) && threads.isWriteExecutorBlocking() ) {
 			throw log.nonblockingOperationSubmitterNotSupported();
 		}
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/orchestration/impl/LuceneParallelWorkOrchestratorImpl.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/orchestration/impl/LuceneParallelWorkOrchestratorImpl.java
@@ -67,7 +67,7 @@ public class LuceneParallelWorkOrchestratorImpl
 	@Override
 	protected void doSubmit(WorkExecution<?> workExecution, OperationSubmitter operationSubmitter,
 			Function<WorkExecution<?>, Runnable> blockingRetryProducer) {
-		if ( !OperationSubmitter.BLOCKING.equals( operationSubmitter ) && threads.isWriteExecutorBlocking() ) {
+		if ( !OperationSubmitter.blocking().equals( operationSubmitter ) && threads.isWriteExecutorBlocking() ) {
 			throw log.nonblockingOperationSubmitterNotSupported();
 		}
 		executor.submit( workExecution );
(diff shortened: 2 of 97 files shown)


engine/src/main/java/org/hibernate/search/engine/backend/work/execution/OperationSubmitter.java
```diff
@@ -16,44 +18,95 @@
  * Interface defining how operation should be submitted to the queue.
  */
 @Incubating
-public enum OperationSubmitter {
+public abstract class OperationSubmitter {
```
yrodiere, line 21: I pushed a commit to remove the delegation and the `Operation` interface. As far as I can tell, this works just as well. Please have a look at the commit and let's discuss it here if you disagree.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/orchestration/impl/ElasticsearchBatchingWorkOrchestrator.java
```diff
@@ -107,8 +108,9 @@ protected void doStart(ConfigurationPropertySource propertySource) {
 	}
 
 	@Override
-	protected void doSubmit(ElasticsearchBatchedWork<?> work, OperationSubmitter operationSubmitter) throws InterruptedException {
-		executors.get( work.getQueuingKey() ).submit( work, operationSubmitter );
+	protected void doSubmit(ElasticsearchBatchedWork<?> work, OperationSubmitter operationSubmitter,
+			Function<ElasticsearchBatchedWork<?>, Runnable> blockingRetryProducer) throws InterruptedException {
+		executors.get( work.getQueuingKey() ).submit( work, operationSubmitter, blockingRetryProducer );
```
yrodiere, line 113: I pushed a commit to add generics, so that we could remove the unnecessary lambda whose only purpose was to cast a work.
Please have a look at that commit and let's discuss it here if you disagree.

yrodiere (commented):
Thanks. I pushed two commits and explained why in the comments. I also have a few other comments, see below.

engine/src/main/java/org/hibernate/search/engine/backend/work/execution/OperationSubmitter.java
marko-bekhta: Yup, this works too 😃 thanks
A sealed interface/class would've been nice, but well ... let's leave it for some later Search version 😃

engine/src/main/java/org/hibernate/search/engine/backend/work/execution/OperationSubmitter.java
```diff
@@ -16,44 +18,95 @@
...
 	 * queue is a {@link java.util.concurrent.BlockingQueue} and it is at its maximum capacity, until some elements
 	 * are removed from the queue
 	 */
-	BLOCKING {
-		@Override
-		public <T> void submitToQueue(BlockingQueue<T> queue, T element) throws InterruptedException {
-			queue.put( element );
-		}
-	},
+	public static final OperationSubmitter BLOCKING = new BlockingOperationSubmitter();
```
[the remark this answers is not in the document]
marko-bekhta: yeah you are right. I didn't want to break the usage of it ( that's if Fabio had already used the submitter) but then hey we haven't released it yet and it's incubating! 😄 will clean this up!

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/orchestration/impl/LuceneParallelWorkOrchestratorImpl.java
```diff
@@ -64,7 +65,8 @@ protected void doStart(ConfigurationPropertySource propertySource) {
 	}
 
 	@Override
-	protected void doSubmit(WorkExecution<?> workExecution, OperationSubmitter operationSubmitter) {
+	protected void doSubmit(WorkExecution<?> workExecution, OperationSubmitter operationSubmitter,
+			Function<WorkExecution<?>, Runnable> blockingRetryProducer) {
 		if ( !OperationSubmitter.BLOCKING.equals( operationSubmitter ) && threads.isWriteExecutorBlocking() ) {
```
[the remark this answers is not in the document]
marko-bekhta: So you mean to do something similar to what I've tried in the other Lucene orchestrator? : 
```java
if ( OperationSubmitter.rejectedExecutionException().equals( operationSubmitter ) &&
		threads.isWriteExecutorBlocking() ) {
	throw log.nonblockingOperationSubmitterNotSupported();
}
else if ( OperationSubmitter.blocking().equals( operationSubmitter ) || 
		(OperationSubmitter.rejectedExecutionException().equals( operationSubmitter ) && !threads.isWriteExecutorBlocking()) ) {
	executor.submit( workExecution );
}
else {
	// otherwise, assume that it's offloading submitter and we want to offload right away:
	operationSubmitter.submitToQueue(
			ALWAYS_BLOCKED_QUEUE,
			workExecution,
			work -> submit( work, OperationSubmitter.blocking() )
	);
}
```
Just calling the `blockingRetryProducer` won't change much as the executor we are offloading to lives within the submitter... Or I've missed your point 😃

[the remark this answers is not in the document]
yrodiere: Yes, you're right.

However, I wouldn't use a hack based on an always blocked queue... Let's get through this.

---

So, first, I just noticed there's something wrong with `BackendThreads.isWriteExecutorBlocking()`: now that we don't necessarily rely on `ThreadPoolProvider#newScheduledExecutor` to create our write executor, it does not make sense to use `ThreadPoolProvider#isScheduledExecutorBlocking` to determine whether the executor is blocking. Maybe add a `isBlocking()` method to `SimpleScheduledExecutor` instead? For `DelegatingSimpleScheduledExecutor`, you'll probably have to pass a boolean to the constructor to be able to answer.

---

Now that we got this out of the way... instead of using an always-blocked queue, we might want to introduce a new method in `OperationSubmitter`?

E.g.:

```
	public abstract <T extends Runnable> void submitToExecutor(SimpleScheduledExecutor executor, T element,
			Function<? super T, Runnable> blockingRetryProducer) throws InterruptedException;
```

`.blocking()` would just run `executor.submit(element)`, `.rejectedExcutionException()` would throw a (meaningful) exception if `executor.isSubmitBlocking()`, and the offloading submitter would just offload all the time.

---

You may run into problems with JQAssistant because `OperationSubmitter` is API, so its methods cannot refer to an SPI type such as `SimpleScheduledExecutor`. While `OperationSubmitter` is API, we don't expect its methods to be called by users, a bit like `ElasticsearchExtension`: 

https://github.com/hibernate/hibernate-search/blob/618594c74ea2cca16ce64a17210325fe886648ab/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/ElasticsearchExtension.java#L45-L47

So, feel free to add a similar exception for `OperationSubmitter` in JQAssistant rules. See here:

https://github.com/hibernate/hibernate-search/blob/618594c74ea2cca16ce64a17210325fe886648ab/build/jqassistant/rules.xml#L384-L385

---

Side note: we could improve event further by adding relevant methods in  `SimpleScheduledExecutor`: `submitWithoutBlockingOrReject`/`submitWithoutBlockingOrReturnNull` (the names are bad, but you see what I mean). That way each `OperationSubmitter` would have a way to do what it wants in `submitToExecutor`. But since we cannot implement `submitWithoutBlockingOrReject`/`submitWithoutBlockingOrReturnNull` with a `ScheduledExecutorService`, this improvement is probably not worth our time.

engine/src/main/java/org/hibernate/search/engine/backend/orchestration/spi/BatchingExecutor.java
```diff
@@ -28,16 +29,17 @@
  * Useful when works can be merged together for optimization purposes (bulking in Elasticsearch),
  * or when they should never be executed in parallel (writes to a Lucene index).
  */
-public final class BatchingExecutor<P extends BatchedWorkProcessor> {
+public final class BatchingExecutor<P extends BatchedWorkProcessor, W extends BatchedWork<? super P>> {
```
yrodiere, line 32: Did you *have* to do this? Because I don't see what this brings, other than an extra generic parameter (which, by the way, is a backwards-incompatible change).

Can you revert the commit that did this, and if there's a problem with that, let's discuss it on Zulip?

engine/src/main/java/org/hibernate/search/engine/backend/work/execution/OperationSubmitter.java
```diff
@@ -8,52 +8,128 @@
...
+	 */
+	public abstract <T> void submitToQueue(BlockingQueue<? super T> queue, T element,
+			Consumer<? super T> blockingRetryProducer) throws InterruptedException;
+
+	/**
+	 * Defines how an element will be submitted to the executor.
+	 * Depending on the implementation might throw {@link RejectedExecutionException} or offload the submit operation to an offload executor.
+	 */
+	public abstract <T extends Runnable> void submitToExecutor(SimpleScheduledExecutor executor, T element,
+			Function<? super T, Runnable> blockingRetryProducer) throws InterruptedException;
```
yrodiere, line 53: Ah, if we switch to `Consumer<? super T> blockingRetryProducer` for the other method, maybe we should here, too? :)

engine/src/main/java/org/hibernate/search/engine/backend/work/execution/OperationSubmitter.java
```diff
@@ -8,52 +8,128 @@
...
+		public <T> void submitToQueue(BlockingQueue<? super T> queue, T element, Consumer<? super T> blockingRetryProducer) {
+			if ( !queue.offer( element ) ) {
+				executor.accept( () -> blockingRetryProducer.accept( element ) );
+			}
+		}
+
+		@Override
+		public <T extends Runnable> void submitToExecutor(SimpleScheduledExecutor executor, T element,
+				Function<? super T, Runnable> blockingRetryProducer) {
+			this.executor.accept( blockingRetryProducer.apply( element ) );
```
yrodiere, line 131: I see you added an "offer" method to `SimpleScheduledExecutor`, but you're not using it here?

engine/src/main/java/org/hibernate/search/engine/common/execution/spi/SimpleScheduledExecutor.java
```diff
@@ -6,16 +6,22 @@
...
 import java.util.concurrent.TimeUnit;
 
+import org.hibernate.search.engine.logging.impl.Log;
 import org.hibernate.search.util.common.annotation.Incubating;
+import org.hibernate.search.util.common.logging.impl.LoggerFactory;
 
 @Incubating
 public interface SimpleScheduledExecutor {
 
+	Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
```
yrodiere, line 23: This public field effectively becomes SPI, and thus we're leaking `Log`, which is an internal class...

Since you're not using it anyway, please remove it?

engine/src/main/java/org/hibernate/search/engine/backend/work/execution/OperationSubmitter.java
```diff
@@ -8,52 +8,128 @@
...
+	public static OperationSubmitter blocking() {
+		return BLOCKING;
+	}
+
 	/**
 	 * When using this submitter adding a new element will cause a {@link RejectedExecutionException} when the underlying
 	 * queue is a {@link java.util.concurrent.BlockingQueue} and it is at its maximum capacity.
 	 */
-	REJECTED_EXECUTION_EXCEPTION {
+	public static OperationSubmitter rejectedExecutionException() {
```
yrodiere, line 68: Just something to consider, if you think it makes sense: now that you have `blocking()` on one side and `offloading()` on the other, naming would feel more consistent if this one was called simply `rejecting()`.

engine/src/main/java/org/hibernate/search/engine/backend/work/execution/OperationSubmitter.java
marko-bekhta: right ... that `offer()` was mostly for the rejecting submitter (if we'd call `submitToExecutor` on it). In this case, I'm 50-50 as to if we should call it or not. On one side, yeah, we are trying not to block the thread, and if we are about to do so we are offloading the submit to another thread that would block, so it makes sense that if we cannot offload because that would block the current thread we should throw an exception. On the other hand, we are trying to offload - so wouldn't a user expect that it would actually get offloaded?

yrodiere: Well I was thinking the `offloading` submitter should catch the `RejectedExecutionException` and offload?

marko-bekhta: hmm so let's see ...

  | submit to queue | submit to executor
-- | -- | --
blocking | just submits to the queue and blocks | just submits to the executor and blocks
rejecting | check if the queue is full. If the queue still can store something - accept that work, if the queue is full - throw exception | that's where the new method is used for the executor. We don't want to block so we make an attempt to offer the new work to it. In that default implementation, it relies on `isBlocking()`. if the executor is blocking, we just fail, even if the executor is empty and can accept work without blocking
offloading | similar to rejecting attempts to add work to the queue, but if the queue is full, then makes "offload" to do blocking adding to the queue in some other thread. | just offloads to the other thread right away in a blocking manner to make sure that work is submitted at some point.

With that written down ... I think it makes sense to do:
```java
public <T extends Runnable> void submitToExecutor(SimpleScheduledExecutor executor, T element,
		Function<? super T, Runnable> blockingRetryProducer) {
	try {
		executor.offer( element );
	}
	catch (Exception e) {
		this.offloadingExecutor.accept( blockingRetryProducer.apply( element ) );
	}
}
```
was that what you had in mind?

yrodiere: Well, that would be a `catch (RejectedOperationException e)`, because catching `Exception` is not what we want here. If `executor` is `null` and that leads to a `NullPointerException`, I'd rather not offload the operation, I want to propagate the `NullPointerException` :)

Actually, even catching `RejectedOperationException` is questionable (using exceptions for flow control being evil, and all that), which is why I initially suggested that the `offer` method return `null` when rejecting the operation, but I agree it's also dodgy in its own way. And, since rejecting an operation should be, well, exceptional, I guess `RejectedOperationException` makes sense. So, ok, `RejectedOperationException` it is.

But otherwise, yes, that's what I had in mind.

yrodiere: Meh, not worth the trouble. Just keep `RejectedOperationException` :)

engine/src/test/java/org/hibernate/search/engine/backend/orchestration/spi/BatchingExecutorTest.java
```diff
@@ -63,9 +66,11 @@ public class BatchingExecutorTest {
 
 	@Parameterized.Parameters(name = "operation submitter = {0}")
 	public static Object[][] params() {
-		return Arrays.stream( OperationSubmitter.values() )
-				.map( value -> new Object[] { value } )
-				.toArray( Object[][]::new );
+		return new Object[][] {
+				{ "BLOCKING", OperationSubmitter.blocking() },
+				{ "REJECTED_EXECUTION_EXCEPTION", OperationSubmitter.rejecting() },
+				{ "OFFLOADING", OperationSubmitter.offloading( r -> Executors.newFixedThreadPool( 1 ).submit( r ) ) }
```
yrodiere, line 72: Creating a thread pool and not shutting it down leaks resources. Not critical, I know, but still, this would be more correct:

```suggestion
				{ "OFFLOADING", OperationSubmitter.offloading( CompletableFuture::runAsync ) }
```

engine/src/test/java/org/hibernate/search/engine/backend/work/execution/OperationSubmitterExecutorTest.java
```diff
@@ -0,0 +1,169 @@
...
+import org.hibernate.search.engine.common.execution.spi.SimpleScheduledExecutor;
+
+import org.junit.Before;
+import org.junit.Test;
+
+public class OperationSubmitterExecutorTest {
+
+	private SimpleScheduledExecutor executor;
+	@Before
+	public void setUp() throws Exception {
```
yrodiere, line 34: This needs an `@After` method to shut down the executor.

engine/src/main/java/org/hibernate/search/engine/backend/orchestration/spi/BatchingExecutor.java
yrodiere: We ended up keeping the change; see https://hibernate.zulipchat.com/#narrow/stream/132092-hibernate-search-dev/topic/HSEARCH-4736

yrodiere (approved):
LGTM, just a few resource leak in tests. Feel free to merge after they're addressed. Thanks!

yrodiere (approved):

PR #3415, opened 2023-02-16
HSEARCH-4767 Support multiple tenants in index workspaces

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4767

### DIFF
--- a/util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/assertion/StubIndexScaleWorkAssert.java
+++ b/util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/assertion/StubIndexScaleWorkAssert.java
@@ -41,8 +41,8 @@ public class StubIndexScaleWorkAssert {
 		boolean hasAnyMismatch;
 		boolean mismatch = checkForMismatch( builder, "type", expected.getType(), actual.getType() );
 		hasAnyMismatch = mismatch;
-		mismatch = checkForMismatch( builder, "tenantIdentifier",
-				expected.getTenantIdentifier(), actual.getTenantIdentifier()
+		mismatch = checkForMismatch( builder, "tenantIdentifiers",
+				expected.getTenantIdentifiers(), actual.getTenantIdentifiers()
 		);
 		hasAnyMismatch = hasAnyMismatch || mismatch;
 		mismatch = checkForMismatch( builder, "routingKeys",
--- a/util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/rule/BackendMock.java
+++ b/util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/rule/BackendMock.java
@@ -6,6 +6,8 @@
  */
 package org.hibernate.search.util.impl.integrationtest.common.rule;
 
+import static org.hibernate.search.util.common.impl.CollectionHelper.asSetIgnoreNull;
+
 import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashSet;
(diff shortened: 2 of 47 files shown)


util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/assertion/StubIndexScaleWorkAssert.java
```diff
@@ -42,7 +42,7 @@ public StubIndexScaleWorkAssert matches(StubIndexScaleWork expected) {
 		boolean mismatch = checkForMismatch( builder, "type", expected.getType(), actual.getType() );
 		hasAnyMismatch = mismatch;
 		mismatch = checkForMismatch( builder, "tenantIdentifier",
```
yrodiere, line 44:
```suggestion
		mismatch = checkForMismatch( builder, "tenantIdentifiers",
```

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/rule/BackendMock.java
```diff
@@ -189,14 +192,11 @@ public DocumentWorkCallListContext expectWorks(String indexName, String tenantId
...
-		return expectIndexScaleWorks( indexName , null );
-	}
-
-	public IndexScaleWorkCallListContext expectIndexScaleWorks(String indexName, String tenantId) {
+	public IndexScaleWorkCallListContext expectIndexScaleWorks(String indexName, String... tenantIds) {
 		CallQueue<IndexScaleWorkCall> callQueue = backendBehavior().getIndexScaleWorkCalls( indexName );
 		return new IndexScaleWorkCallListContext(
-				indexName, tenantId,
+				indexName,
+				Arrays.asList( tenantIds ).stream().filter( Objects::nonNull ).collect( Collectors.toList() ),
```
yrodiere, line 199: Better use a `Set` here, so that order is irrelevant. That would require using a `Set` in the mappers as well, though (as I suggested elsewhere).

util/internal/integrationtest/mapper/stub/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/stub/StubMappedIndex.java
```diff
@@ -169,7 +170,10 @@ public IndexWorkspace createWorkspace(StubSession sessionContext) {
 	}
 
 	public IndexWorkspace createWorkspace(String tenantId) {
-		return delegate().createWorkspace( mapping, tenantId );
+		return delegate().createWorkspace(
+				mapping,
+				tenantId == null ? Collections.emptySet() : Collections.singleton( tenantId )
```
yrodiere, line 175: You're doing this a lot, all over the place. Maybe consider adding a method to `org.hibernate.search.util.common.impl.CollectionHelper`?

yrodiere (commented):
Thanks. Looks good overall, I just added a few minor comments.

util/internal/integrationtest/mapper/stub/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/stub/StubMappedIndex.java
marko-bekhta: Yeah, I've thought about some tenant utils, but that didn't make much sense, good point about the collection helper, will do!

yrodiere (approved):

marko-bekhta:
Every time I trust IDE to do the refactoring:
```
* @deprecated Use {@link #createWorkspace(BackendMappingContext, Set) <String>)} instead.
```
what's that `<String>` doing there 😢 😞 😄

yrodiere:
Test failures are unrelated. Merging, thanks!

PR #3416, opened 2023-02-16
HSEARCH-4766 Use single executor for mass indexing agents

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4766

I've tried to see if we can use a single mass indexing agent (that would contain multiple `TenantDelegate` in case of multitenancy). The problem with such an approach is that eventually, we are creating `SessionHelper`  that requires a single tenant to start a transaction... 
Hence instead I've just passed the executor to the `OutboxPollingMassIndexerAgent#start ` through a context.

### DIFF
--- a/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingBatchCoordinator.java
+++ b/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingBatchCoordinator.java
@@ -13,12 +13,14 @@ import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.function.Function;
 
 import org.hibernate.search.engine.backend.work.execution.OperationSubmitter;
 import org.hibernate.search.engine.reporting.spi.RootFailureCollector;
 import org.hibernate.search.mapper.pojo.massindexing.MassIndexingEnvironment;
 import org.hibernate.search.mapper.pojo.massindexing.spi.PojoMassIndexerAgent;
+import org.hibernate.search.mapper.pojo.massindexing.spi.PojoMassIndexerAgentStartContext;
 import org.hibernate.search.mapper.pojo.massindexing.spi.PojoMassIndexingMappingContext;
 import org.hibernate.search.mapper.pojo.reporting.impl.PojoEventContextMessages;
 import org.hibernate.search.mapper.pojo.schema.management.spi.PojoScopeSchemaManager;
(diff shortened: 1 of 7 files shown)


mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingBatchCoordinator.java
```diff
@@ -76,6 +79,15 @@ public PojoMassIndexingBatchCoordinator(PojoMassIndexingMappingContext mappingCo
 		this.dropAndCreateSchemaOnStart = dropAndCreateSchemaOnStart;
 		this.purgeAtStart = purgeAtStart;
 		this.mergeSegmentsAfterPurge = mergeSegmentsAfterPurge;
+
+		this.agentStartContext = new PojoMassIndexerAgentStartContextImpl(
+				mappingContext.threadPoolProvider()
+						.newScheduledExecutor(
+								1,
+								PojoMassIndexingBatchIndexingWorkspace.THREAD_NAME_PREFIX + "Mass indexer agent"
+						),
```
yrodiere, line 88: We don't always need this executor, though... only when using outbox-polling coordination.

Can you create the executor lazily, on the first call to `PojoMassIndexerAgentStartContextImpl#scheduledExecutor`? That way we don't use resources unnecessarily.

yrodiere (commented):
LGTM, but we should only create the executor when necessary, see below.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingBatchCoordinator.java
marko-bekhta: I went with adding `clear()` to the start context, as otherwise, we'd call `agentStartContext.scheduledExecutor().shutdownNow()` and start/close the executor which is probably not what we'd want.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/spi/PojoMassIndexerAgentStartContext.java
```diff
@@ -15,4 +15,9 @@ public interface PojoMassIndexerAgentStartContext {
 	ScheduledExecutorService scheduledExecutor();
 
 	FailureHandler failureHandler();
+
+	/**
+	 * Clears up any used resources. For example attempts to shut down the executor if one was initialized.
+	 */
+	void clear();
```
yrodiere, line 22: This doesn't need to be exposed in the SPI, it's strictly an internal thing.

yrodiere (changes requested):

yrodiere (approved):

PR #3424, opened 2023-02-21
HSEARCH-4804 Fix flaky test OutboxPollingAutomaticIndexingLifecycleIT with postgresql 

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4804

I'm going blind here since I couldn't reproduce the problem locally, but I think there's a decent chance of at least reducing, if not eliminating, flakiness.

See the commit messages for details.


yrodiere:
At least this build didn't fail... I'll merge and see how it goes.

PR #3425, opened 2023-02-22
Revert HSEARCH-4721, re-disable tests because of HHH-16219 

yrodiere:

See https://hibernate.atlassian.net/browse/HSEARCH-4721?focusedCommentId=110953S
See https://hibernate.atlassian.net/browse/HHH-16219


marko-bekhta (approved):
ohh re-re-disable 😃 🙈

PR #3436, opened 2023-03-07
HSEARCH-4814 Use failure reporter in offloading operation submitter

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4814

I might've missed the point, but I didn't find a use for the `T` in a `BiConsumer<T, Throwable>`, so I just went with a simple `Consumer<Throwable>`. 
I also went ahead and wrapped the call to the `executor.accept(..)` with a try/catch even though it should happen in the same thread, but since we already have a reporter in place... 
And replaced the offloading lambdas with a class.
BTW did this happen anywhere, or it's more of a precaution?

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/orchestration/impl/LuceneParallelWorkOrchestratorImpl.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/orchestration/impl/LuceneParallelWorkOrchestratorImpl.java
@@ -7,6 +7,7 @@
 package org.hibernate.search.backend.lucene.orchestration.impl;
 
 import java.util.concurrent.CompletableFuture;
+import java.util.function.BiConsumer;
 
 import org.hibernate.search.backend.lucene.lowlevel.index.impl.IndexAccessor;
 import org.hibernate.search.backend.lucene.resources.impl.BackendThreads;
@@ -21,6 +22,8 @@ import org.hibernate.search.util.common.reporting.EventContext;
 public class LuceneParallelWorkOrchestratorImpl
 		extends AbstractWorkOrchestrator<LuceneParallelWorkOrchestratorImpl.WorkExecution<?>>
 		implements LuceneParallelWorkOrchestrator {
+
+	private static final BiConsumer<WorkExecution<?>, Throwable> ASYNC_FAILURE_REPORTER = WorkExecution::markAsFailed;
 
 	private final IndexAccessor indexAccessor;
 	private final IndexAccessorWorkExecutionContext context;
@@ -60,7 +63,7 @@ public class LuceneParallelWorkOrchestratorImpl
 
 	@Override
 	protected void doSubmit(WorkExecution<?> workExecution, OperationSubmitter operationSubmitter) throws InterruptedException {
-		operationSubmitter.submitToExecutor( executor, workExecution, blockingRetryProducer );
+		operationSubmitter.submitToExecutor( executor, workExecution, blockingRetryProducer, ASYNC_FAILURE_REPORTER );
 	}
 
 	@Override
@@ -95,8 +98,12 @@ public class LuceneParallelWorkOrchestratorImpl
 			}
 			catch (Throwable e) {
 				context.getIndexAccessor().cleanUpAfterFailure( e, work.getInfo() );
-				result.completeExceptionally( e );
+				markAsFailed( e );
 			}
+		}
+
+		public void markAsFailed(Throwable throwable) {
+			result.completeExceptionally( throwable );
 		}
 	}
 
--- a/engine/src/main/java/org/hibernate/search/engine/backend/orchestration/spi/BatchingExecutor.java
+++ b/engine/src/main/java/org/hibernate/search/engine/backend/orchestration/spi/BatchingExecutor.java
@@ -13,6 +13,7 @@ import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Future;
+import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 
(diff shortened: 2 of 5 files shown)


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/orchestration/impl/LuceneParallelWorkOrchestratorImpl.java
```diff
@@ -60,7 +60,7 @@ protected void doStart(ConfigurationPropertySource propertySource) {
 
 	@Override
 	protected void doSubmit(WorkExecution<?> workExecution, OperationSubmitter operationSubmitter) throws InterruptedException {
-		operationSubmitter.submitToExecutor( executor, workExecution, blockingRetryProducer );
+		operationSubmitter.submitToExecutor( executor, workExecution, blockingRetryProducer, workExecution::markAsFailed );
```
yrodiere, line 63: Use a BiConsumer to avoid capturing lambdas, which have a performance impact:

```suggestion
		operationSubmitter.submitToExecutor( executor, workExecution, blockingRetryProducer, WorkExecution::markAsFailed );
```

Also, you could put `WorkExecution::markAsFailed` into a `static final` constant.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/orchestration/impl/LuceneParallelWorkOrchestratorImpl.java
```diff
@@ -94,10 +94,14 @@ public void run() {
...
 			}
 			catch (Throwable e) {
-				context.getIndexAccessor().cleanUpAfterFailure( e, work.getInfo() );
-				result.completeExceptionally( e );
+				markAsFailed( e );
 			}
 		}
+
+		public void markAsFailed(Throwable throwable) {
+			context.getIndexAccessor().cleanUpAfterFailure( throwable, work.getInfo() );
```
yrodiere, line 102: This doesn't seem necessary if we just failed to *submit* the work. If that happens we didn't even touch the index accessor.

engine/src/main/java/org/hibernate/search/engine/backend/orchestration/spi/BatchingExecutor.java
```diff
@@ -127,7 +127,7 @@ public void submit(W work, OperationSubmitter operationSubmitter) throws Interru
 					"Attempt to submit a work to executor '" + name + "', which is stopped."
 			);
 		}
-		operationSubmitter.submitToQueue( workQueue, work, blockingRetryProducer );
+		operationSubmitter.submitToQueue( workQueue, work, blockingRetryProducer, work::markAsFailed );
```
yrodiere, line 130: Use `BatchedWork::markAsFailed` to avoid capturing lambdas, which have a performance impact.
Also, you could put `BatchedWork::markAsFailed` into a `static final` constant.

engine/src/main/java/org/hibernate/search/engine/backend/work/execution/OperationSubmitter.java
```diff
@@ -118,20 +118,53 @@ private OffloadingExecutorOperationSubmitter(Consumer<Runnable> executor) {
...
+		public <T> void submitToQueue(BlockingQueue<? super T> queue, T element, Consumer<? super T> blockingRetryProducer,
+				Consumer<Throwable> asyncFailureReporter) {
 			if ( !queue.offer( element ) ) {
-				executor.accept( () -> blockingRetryProducer.accept( element ) );
+				try {
+					this.executor.accept( new RetryAction<>( element, blockingRetryProducer, asyncFailureReporter ) );
+				}
+				catch (Throwable throwable) {
+					asyncFailureReporter.accept( throwable );
+				}
```
yrodiere, line 129: I think you can leave this out. Callers will want to catch synchronous failures.

engine/src/main/java/org/hibernate/search/engine/backend/work/execution/OperationSubmitter.java
```diff
@@ -118,20 +118,53 @@ private OffloadingExecutorOperationSubmitter(Consumer<Runnable> executor) {
...
 				executor.offer( element );
 			}
 			catch (RejectedExecutionException e) {
-				this.executor.accept( () -> blockingRetryProducer.accept( element ) );
+				try {
+					this.executor.accept( new RetryAction<>( element, blockingRetryProducer, asyncFailureReporter ) );
+				}
+				catch (Throwable throwable) {
+					asyncFailureReporter.accept( throwable );
+				}
```
yrodiere, line 145: I think you can leave this out. Callers will want to catch synchronous failures as usual.

yrodiere (changes requested):
Thanks!

> I might've missed the point, but I didn't find a use for the T in a BiConsumer<T, Throwable>, so I just went with a simple Consumer<Throwable>.

See below.

>  I also went ahead and wrapped the call to the executor.accept(..) with a try/catch even though it should happen in the same thread, but since we already have a reporter in place...

See below.

>  And replaced the offloading lambdas with a class.

Fine by me.

>  BTW did this happen anywhere, or it's more of a precaution?

It's a precaution, but I'm sure it would happen eventually and it would be hell to debug :x

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/orchestration/impl/LuceneParallelWorkOrchestratorImpl.java
marko-bekhta: > Use a BiConsumer to avoid capturing lambdas, which have a performance impact:

Thanks, I knew there was something behind it 😃

yrodiere (approved):

yrodiere:
Test failure is unrelated, merging. Thanks!

PR #3437, opened 2023-03-08
HSEARCH-2366 Add an Elasticsearch mapping export tool

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-2366

Should we add some options parameter into the export method even if it'll be empty for now or contain some setting to export mappings (what it's doing now) so that in the future we won't break the API or ... as it's incubating - don't bother with it now and later we'll just break the API? 😄 

Then there also was a mention of a CLI ... is that what we need or that was just a thought?


integrationtest/mapper/orm-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/orm/realbackend/schema/management/ElasticsearchSchemaManagerExporterIT.java
```diff
@@ -0,0 +1,151 @@
...
+				.withProperty( "hibernate.search.backends." + Article.BACKEND_NAME + ".version", "8.6" )
+				.setup( Book.class, Article.class );
+
+		Path directory = temporaryFolder.newFolder().toPath();
+		Search.session( entityManagerFactory.createEntityManager() ).schemaManager().exportSchema( directory );
+
+		String bookIndex = Files.readString(
+				directory.resolve( "backend" ) // as we are using the default backend
+						.resolve( "indexes" )
+						.resolve( Book.class.getName() ) // we use FQN as who knows maybe someone will decide to have same class names in different packages
```
yrodiere, line 53: That's weird, we should be using the entity name, which is unique within a persistence unit. The entity name should be referred to as "type name" at the backend level, and it's the default for the index name.

Which brings me to this: you're using the index name in the export, right?

integrationtest/mapper/orm-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/orm/realbackend/schema/management/LuceneSchemaManagerExporterIT.java
```diff
@@ -0,0 +1,61 @@
...
+				.withProperty( "hibernate.search.backends." + Article.BACKEND_NAME + ".type", "lucene" )
+				.setup( Book.class, Article.class );
+
+		Path directory = temporaryFolder.newFolder().toPath();
+		Search.session( entityManagerFactory.createEntityManager() ).schemaManager().exportSchema( directory );
+
+		String bookIndex = Files.readString(
+				directory.resolve( "backend" ) // as we are using the default backend
+						.resolve( "indexes" )
+						.resolve( Book.class.getName() ) // we use FQN as who knows maybe someone will decide to have same class names in different packages
```
yrodiere, line 49: Same, this should be the entity name, not the FQCN.

integrationtest/mapper/pojo-standalone-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/standalone/realbackend/schema/management/ElasticsearchSchemaManagerExporterIT.java
```diff
@@ -0,0 +1,207 @@
...
+
+		Path directory = temporaryFolder.newFolder().toPath();
+		try ( SearchSession session = mapping.createSession() ) {
+			session.schemaManager().exportSchema( directory );
+		}
+
+		String bookIndex = Files.readString(
+				directory.resolve( "backend" ) // as we are using the default backend
+						.resolve( "indexes" )
+						.resolve( Book.class.getName() ) // we use FQN as who knows maybe someone will decide to have same class names in different packages
```
yrodiere, line 57: Technically this should be the index name, which defaults to the entity name, which defaults to the simple name of the class, and can be configured this in the mapper. See: https://github.com/hibernate/hibernate-search/blob/37121f38cfb368d5654534c7e28ee46656895b75/integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/IndexedBaseIT.java#L67-L126

We shouldn't be using the FQCN here, unless that's what the user selected explicitly.

integrationtest/mapper/pojo-standalone-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/standalone/realbackend/schema/management/LuceneSchemaManagerExporterIT.java
```diff
@@ -0,0 +1,117 @@
...
+
+		Path directory = temporaryFolder.newFolder().toPath();
+		try ( SearchSession session = mapping.createSession() ) {
+			session.schemaManager().exportSchema( directory );
+		}
+
+		String bookIndex = Files.readString(
+				directory.resolve( "backend" ) // as we are using the default backend
+						.resolve( "indexes" )
+						.resolve( Book.class.getName() ) // we use FQN as who knows maybe someone will decide to have same class names in different packages
```
yrodiere, line 53: Same here, this should not be using the FQCN.

documentation/src/main/asciidoc/reference/schema.asciidoc
```diff
@@ -145,6 +145,13 @@ because <<schema-management-concepts-lucene-schema,local Lucene indexes don't ha
 |[[mapper-orm-schema-management-manager-drop-and-create]]`dropAndCreate()`
 |Drops existing indexes and re-creates them and their schema.
 |<<schema-management-concepts-drop-loses-data,All indexed data will be lost>>.
+|[[mapper-orm-schema-management-manager-export-schema]]`exportSchema(Path)`
+|Exports the schema into a provided target directory. The output is backend specific.
+In case of the <<backend-elasticsearch,Elasticsearch backend>> - json files representing index mapping
+and some additional information on the index will be generated in a directory tree grouped by backends and indexed types.
+While for the <<backend-lucene,Lucene backend>>, only a file tree will be generated representing the backends and indexed types,
+as this backend does not support the schema exporting.
+|This API is still incubating and might change in the future.
```
yrodiere, line 154: I'd move this to a dedicated section, be it only to add an example?

yrodiere (commented):
Thanks, my comments below.

> Should we add some options parameter into the export method even if it'll be empty for now

No, we can always add another `exportSchema` method with different parameters later. And we can do that in a backwards-compatible way.

> Then there also was a mention of a CLI ... is that what we need or that was just a thought?

It's obsolete, let's not bother with that. We'll assume that people wanting to export their schema will need to start their application in offline mode; though we might need to document that explicitly.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/index/impl/IndexManagerBackendContext.java
```diff
@@ -191,7 +191,7 @@ IOStrategy createIOStrategy(ConfigurationPropertySource propertySource) {
 	}
 
 	LuceneIndexSchemaManager createSchemaManager(SchemaManagementIndexManagerContext context) {
-		return new LuceneIndexSchemaManager( workFactory, context );
+		return new LuceneIndexSchemaManager( backendAPI, workFactory, context );
```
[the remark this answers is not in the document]
marko-bekhta: Yeah, just wanted to make it more similar between backends 😃. Updated in the latest changes.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/schema/management/impl/ElasticsearchIndexSchemaManager.java
```diff
@@ -118,4 +123,9 @@ public CompletableFuture<?> validate(ContextualFailureCollector failureCollector
 						: schemaAccessor.waitForIndexStatus( indexNames, executionOptions, operationSubmitter )
 				);
 	}
+
+	@Override
+	public void exportSchema(Path targetDirectory, String name) {
```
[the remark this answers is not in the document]
marko-bekhta: I went with a slightly different API, but it should still be flexible enough (I hope) to do the Quarkus work later. Let me know what you think 😄

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/schema/management/impl/ElasticsearchSchemaExporter.java
```diff
@@ -0,0 +1,98 @@
...
+		targetDirectory = targetDirectory
+				.resolve( backendName.map( bn -> FileSystems.getDefault().getPath( "backends", bn ) ).orElse( DEFAULT_BACKEND_PATH ) )
+				.resolve( "indexes" )
+				.resolve( name );
+		checkOrCreateTargetDirectory( targetDirectory );
+
+		try {
+			writeString(
+					targetDirectory.resolve( "index.json" ),
+					indexMetadata.toString()
```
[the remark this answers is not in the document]
marko-bekhta: Done.

integrationtest/mapper/orm-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/orm/realbackend/schema/management/ElasticsearchSchemaManagerExporterIT.java
marko-bekhta: I think what confused me was that example from the task:
```
backend/indexes/MyEntity/index.json
backends/someNamedBackend/indexes/Foo/index.json
```
seeing that I thought that you wanted to see class names rather than index names 😄. But if it was index name all the time ... well there was a simpler way to get it compared to what I've done initially. It is an index name now.

documentation/src/main/asciidoc/reference/schema.asciidoc
marko-bekhta: I'll move it once you take a look at the new API and we finalize it 👍🏻😄

engine/src/main/java/org/hibernate/search/engine/mapper/schema/management/SchemaExport.java
```diff
@@ -0,0 +1,34 @@
...
+	/**
+	 * A simple exporter that writes the schema out to the provided target directory. Renders a file tree containing
+	 * all backends that this export represents and their corresponding indexes.
+	 *
+	 * @param targetDirectory The target directory to generate the output into.
+	 * @return The exporter that should be provided to {@link SchemaExport#applyElementExporter(Consumer)}
+	 */
+	static Consumer<SchemaElementContext> fileExporter(Path targetDirectory) {
+		return new FileSchemaElementExporter( targetDirectory, SchemaElementContext.OutputFormat.SIMPLE );
+	}
```
yrodiere, line 33: Or simply expose a `toFiles` method?

```suggestion
	/**
	 * Writes the schema out to the provided target directory. Renders a file tree containing the schema of
	 * all backends that this export represents and their corresponding indexes.
	 *
	 * @param targetDirectory The target directory to generate the output into.
	 * @return The exporter that should be provided to {@link SchemaExport#applyElementExporter(Consumer)}
	 */
	default void toFile(Path targetDirectory) {
		applyElementExporter( new FileSchemaElementExporter( targetDirectory, SchemaElementContext.OutputFormat.SIMPLE ) );
	}
```

engine/src/main/java/org/hibernate/search/engine/mapper/schema/management/SchemaElementContext.java
```diff
@@ -0,0 +1,49 @@
...
+	 * @param stream The stream to which the content should be written.
+	 * @throws IOException If any {@link IOException I/O problems} occur during the process.
+	 * @see OutputFormat
+	 */
+	void write(OutputFormat format, OutputStream stream) throws IOException;
+
+	/**
+	 * Unwraps this context to a more specific (backend) version.
+	 */
+	<T extends SchemaElementContext> T unwrap(Class<T> klass);
```
yrodiere, line 48: Would you mind using `ElasticsearchExtension` instead? I've been kind of encouraged to expose `unwrap` methods in the past, but I definitely don't like it...

The nice thing about `ElasticsearchExtension` is that:

1. It's discoverable; you don't need to know the name of the Elasticsearch-specific class, you just pass `ElasticsearchExtension.get()` and automatically get the correct type as a result.
2. It works with parameterized types, though here that's not important.

See for example `org.hibernate.search.backend.elasticsearch.search.query.impl.ElasticsearchSearchQueryImpl#extension`.

engine/src/main/java/org/hibernate/search/engine/mapper/schema/management/SchemaElementContext.java
```diff
@@ -0,0 +1,49 @@
...
+import java.io.OutputStream;
+import java.util.Optional;
+
+import org.hibernate.search.util.common.annotation.Incubating;
+
+@Incubating
+public interface SchemaElementContext {
+	enum OutputFormat {
+		/**
+		 * Generate a JSON representation of this context.
```
yrodiere, line 19: That's backend-specific though...  and it's not even useful at the moment since there's only one format.

I'd remove this completely. If this becomes necessary case we can expose additional `write` methods in `ElasticsearchSchemaElementContext`.

engine/src/main/java/org/hibernate/search/engine/mapper/schema/management/SchemaElementContext.java
```diff
@@ -0,0 +1,49 @@
...
+	/**
+	 * @return The name of the backend to which the index this context represents belongs to.
+	 * If it is a part of the default backend - {@link Optional#empty() an empty optional} will be returned.
+	 */
+	Optional<String> backendName();
+
+	/**
+	 * @return The name of the index this context represents.
+	 */
+	String indexName();
```
yrodiere, line 33: The downside of this compared to my visitor is that it doesn't work well with exports that are not strictly related to the *index* schema. Imagine we export the scripts to initialize a Kafka stream for Debezium, for example; what would `indexName` return?

That's why I think instead of a `Consumer<SchemaElementContext>`, you should expose some kind of visitor interface such as the `SearchSchemaCollector`/`IndexSchemaCollector` in my initial proposal.

The good news is, it's totally compatible with your concept of "unwrappable"/"extendable" `SchemaElementContext`.

yrodiere (commented):
Added comments on the API; didn't review the rest though. Please ping me when you're completely done for a full review.

engine/src/main/java/org/hibernate/search/engine/mapper/schema/management/SchemaElementContext.java
marko-bekhta: Right... I think that I've missed the point that this collector would be used to walk through different "schemas". That's why it took me a different path 😃 
Now, if I got you correctly - we want to have this general-purpose schema walker that we'll extend in the future by adding more visit methods. This would apply to `SearchSchemaManager`/`PojoScopeSchemaManager` while the currently underlying `IndexSchemaManager`s are fine with just providing the "element" that we'll feed into this collector since the index schema manager is .. well for index schema 😃 

If that's what we are looking for - I'll clean things up before you do the proper review 🙈😃

yrodiere: Yes that's the idea :100:

yrodiere (commented):
Thanks. Here's another round of comments :)

documentation/src/test/java/org/hibernate/search/documentation/mapper/orm/schema/management/HibernateOrmSchemaManagerIT.java
```diff
@@ -84,6 +96,57 @@ public void selectType() {
...
+			// tag::schema-export[]
+			SearchSchemaManager schemaManager = searchSession.schemaManager(); // <1>
+			schemaManager.exportSchema( SearchSchemaCollectors.indexSchemaToFile( targetDirectory ) ); // <2>
+			// end::schema-export[]
+
+			String bookIndex = Files.readString(
+					targetDirectory.resolve( "backend" )
+							.resolve( "indexes" )
+							.resolve( Book.class.getSimpleName() )
+							.resolve( "index.json" ) );
```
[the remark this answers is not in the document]
marko-bekhta: Yeah 😃 I have mixed feelings about that one 😄 
I do like the idea of having multiple files or files specific to the backend.. But on the other hand, I had a "problem" with that `toDirectory` method being sort of out of place... what if we'd needed to export to some other place, maybe through HTTP or something? Would we be adding more `toSmth()` methods?
I'd prefer to remove the `write` or `toDirectory` methods from this export interface and have the actual exporting logic live in the collector implementation. But that would mean that the collector from the engine would need to delegate to the collectors from backends, which leads to probably otherwise unnecessary complexity 😄 
Anywaaay, if you are ok with `toDirectory(..)` I'll make the changes 😃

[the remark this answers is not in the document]
yrodiere: > what if we'd needed to export to some other place, maybe through HTTP or something

We're free to reject features that we don't think make sense, and in this case I would.

But if you really don't want to rely on the filesystem at this level, I'd say you could use something like `.toFiles(BiConsumer<String, OutputStream>)`.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/schema/management/impl/ElasticsearchIndexSchemaExportImpl.java
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	@Override
+	public List<JsonObject> bodyParts() {
+		return Collections.unmodifiableList( request.bodyParts() );
+	}
+
+	private class SimpleExporter implements Consumer<OutputStream> {
+		@Override
+		public void accept(OutputStream stream) {
```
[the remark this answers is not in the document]
marko-bekhta: yeah, I thought I'll keep a class since we are planning to have other exporters in the future, but I guess .. when we'll have them we'll create a class 😃

integrationtest/mapper/orm-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/orm/realbackend/util/Article.java
```diff
@@ -0,0 +1,42 @@
...
+package org.hibernate.search.integrationtest.mapper.orm.realbackend.util;
+
+import javax.persistence.Entity;
+import javax.persistence.Id;
+
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.FullTextField;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.Indexed;
+
+@Entity(name = Article.NAME)
+@Indexed(index = Article.NAME, backend = Article.BACKEND_NAME )
```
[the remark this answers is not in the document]
marko-bekhta: I suppose then I'll also remove the index name from the book class next to this one?
```java
@Entity(name = Book.NAME)
@Indexed(index = Book.NAME)
public class Book 
```

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/schema/management/impl/ElasticsearchIndexSchemaExportImpl.java
```diff
@@ -0,0 +1,82 @@
...
+		for ( Map.Entry<String, String> entry : request.parameters().entrySet() ) {
+			queryParams.addProperty( entry.getKey(), entry.getValue() );
+		}
+		List<JsonObject> parts = request.bodyParts();
+		JsonElement body = parts.size() == 1 ?
+				parts.get( 0 ) :
+				parts.stream().collect( JsonArray::new, JsonArray::add, JsonArray::addAll );
+
+		write( targetDirectory, "create-index.json", body );
+		write( targetDirectory, "query-params.json", queryParams );
```
yrodiere, line 57: Ah, you moved the query params out of `create-index.json`? Then maybe call that `create-index-query-params.json`? Otherwise we'll be in trouble if we add other requests (e.g. `drop-index.json`) in the future.

```suggestion
		write( targetDirectory, "create-index.json", body );
		write( targetDirectory, "create-index-query-params.json", queryParams );
```

documentation/src/main/asciidoc/reference/schema.asciidoc
```diff
@@ -258,3 +258,82 @@ on an AWS Elasticsearch cluster.
...
+[WARNING]
+====
+The schema export is backend specific.
+Only the <<backend-elasticsearch,Elasticsearch backend>> provides the information required to create an index.
+The <<backend-lucene,Lucene backend>> does not provide any actual schema information beyond the list of indexes.
+====
+
+Hibernate Search ships with a built-in schema collector that writes the content of a schema export to a disk.
+All indexes represented by a `SearchSchemaManager` will be written to their own directory as an `index.json` file containing
+information on the index. The output is backend specific.
```
yrodiere, line 278: The file name is obsolete... But this needs to be reworded anyway as we didn't explain schema collectors at this point, and the user actually doesn't need to know we use a schema collector under the hood:

```suggestion
The <<mapper-orm-schema-management-manager,schema manager>> provides a way to export schemas to the filsystem. The output is backend-specific.

[NOTE]
====
Schema exports are constructed based on the mapping information and configurations (e.g. such as the backend version).
Resulting exports are not compared to or validated against the actual backend schema.
====
```

documentation/src/main/asciidoc/reference/schema.asciidoc
```diff
@@ -258,3 +258,82 @@ on an AWS Elasticsearch cluster.
...
+
+The file tree structure of an export is shown below:
+[source]
+----
+# For a default backend the index schema will be written into:
+.../backend/indexes/<index name>/index.json
+# For additional named backends:
+.../backends/<name of a particular backend>/indexes/<index name>/index.json
+----
```
yrodiere, line 287: This is out of date.

```suggestion
For Elasticsearch, the files provide the necessary information to create indexes (along with their settings and mapping). The file tree structure of an export is shown below:
[source]
----
# For the default backend the index schema will be written to:
.../backend/indexes/<index name>/create-index.json
.../backend/indexes/<index name>/create-index-query-params.json
# For additional named backends:
.../backends/<name of a particular backend>/indexes/<index name>/create-index.json
.../backends/<name of a particular backend>/indexes/<index name>/create-index-query-params.json
----
```

documentation/src/main/asciidoc/reference/schema.asciidoc
```diff
@@ -258,3 +258,82 @@ on an AWS Elasticsearch cluster.
...
+<1> Retrieve the `SearchSchemaManager` from a `SearchSession`.
+<2> Export the schema to a target directory.
+====
+
+[[schema-management-export-custom]]
+=== Exporting to a custom collector
+
+<<mapper-orm-schema-management-manager,Search schema managers>> allow to walk through the schema exports based on the data such managers contains.
+To do so a `SearchSchemaCollector` must be implemented and passed to the schema manager's `exportSchema(..)` method.
```
yrodiere, line 303:
```suggestion
To do so a `SearchSchemaCollector` must be implemented and passed to the schema manager's `exportExpectedSchema(..)` method.
```

documentation/src/main/asciidoc/reference/schema.asciidoc
```diff
@@ -258,3 +258,82 @@ on an AWS Elasticsearch cluster.
...
+
+Any schema exports are constructed based on the mapping information and configurations (e.g. such as the backend version).
+Resulting exports are not compared to or validated against the actual backend schema.
```
yrodiere, line 306:
```suggestion
[NOTE]
====
Schema exports are constructed based on the mapping information and configurations (e.g. such as the backend version).
Resulting exports are not compared to or validated against the actual backend schema.
====
```

documentation/src/main/asciidoc/reference/schema.asciidoc
```diff
@@ -258,3 +258,82 @@ on an AWS Elasticsearch cluster.
...
+include::{sourcedir}/org/hibernate/search/documentation/mapper/orm/schema/management/HibernateOrmSchemaManagerIT.java[tags=walking-the-schema]
+----
+<1> Retrieve the `SearchSchemaManager` from a `SearchSession`.
+<2> Instantiate and pass the `SearchSchemaCollector` to walk a schema.
+<3> Create a name from an index and backend names.
+====
+
+[TIP]
+====
+To access additional, backend specific, functionality an extension to `IndexSchemaExport` can be applied:
```
yrodiere, line 322:
```suggestion
To access backend-specific functionality, an extension to `SchemaExport` can be applied:
```

documentation/src/main/asciidoc/reference/schema.asciidoc
```diff
@@ -258,3 +258,82 @@ on an AWS Elasticsearch cluster.
...
+=====
+[source, JAVA, indent=0]
+----
+include::{sourcedir}/org/hibernate/search/documentation/mapper/orm/schema/management/ElasticsearchHibernateOrmSchemaManagerIT.java[tags=walking-the-schema-extension]
+----
+<1> Extend the `IndexSchemaExport` with the Elasticsearch extension.
```
yrodiere, line 328:
```suggestion
<1> Extend the `SchemaExport` with the Elasticsearch extension.
```

engine/src/main/java/org/hibernate/search/engine/mapper/schema/management/SchemaExport.java
```diff
@@ -0,0 +1,36 @@
...
+import java.nio.file.Path;
+
+import org.hibernate.search.util.common.SearchException;
+import org.hibernate.search.util.common.annotation.Incubating;
+
+@Incubating
+public interface SchemaExport {
+
+	/**
+	 * Writes the content of this export to a provided output stream.
```
yrodiere, line 18:
```suggestion
	 * Writes the content of this export to a directory on the filesystem.
```

engine/src/main/java/org/hibernate/search/engine/mapper/schema/management/SchemaExport.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.engine.mapper.schema.management;
```
yrodiere, line 7: Just noticed the package... IMO this would fit better in `engine.common.schema.management`, since it's not strictly about the mapper. Same for `SchemaExportExtension`.

engine/src/main/java/org/hibernate/search/engine/mapper/schema/management/SearchSchemaCollector.java
```diff
@@ -0,0 +1,28 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.engine.mapper.schema.management;
```
yrodiere, line 7: Not sure this is the right place... Either you consider this is about the backend/index and then it should be in `engine.backend.*`, or you consider it's about the mapper and then it should be in the `pojo-base` module so that we can add methods specific to the POJO mappers.

What about the [original plan](https://github.com/hibernate/hibernate-search/pull/3437#discussion_r1130658256) of having `IndexSchemaCollector` live in the engine in a `backend.schema.management` package, and `SearchSchemaCollector` live in `pojo-base` (and extending `IndexSchemaCollector`)?

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/schema/management/impl/IndexSchemaExportFileCollector.java
```diff
@@ -0,0 +1,70 @@
...
+
+	public IndexSchemaExportFileCollector(Path targetDirectory) {
+		this.targetDirectory = targetDirectory;
+	}
+
+	@Override
+	public void indexSchema(Optional<String> backendName, String indexName, SchemaExport export) {
+		Path target = targetDirectory
+				.resolve( backendName
+						.map( bn -> FileSystems.getDefault().getPath( "backends", bn ) )
```
yrodiere:
```suggestion
						.map( bn -> Paths.get( "backends", bn ) )
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/schema/management/impl/PojoSchemaManagementIndexedTypeContext.java
```diff
@@ -15,4 +16,5 @@ public interface PojoSchemaManagementIndexedTypeContext {
 
 	IndexSchemaManager schemaManager();
 
+	IndexManager toAPI();
```
yrodiere, line 19: This is never used.

```suggestion
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/schema/management/impl/PojoSchemaManagementIndexedTypeContext.java
```diff
@@ -6,6 +6,7 @@
  */
 package org.hibernate.search.mapper.pojo.schema.management.impl;
 
+import org.hibernate.search.engine.backend.index.IndexManager;
```
yrodiere, line 9:
```suggestion
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/impl/PojoIndexedTypeManager.java
```diff
@@ -109,6 +110,11 @@ public IndexSchemaManager schemaManager() {
 		return indexManager.schemaManager();
 	}
 
+	@Override
+	public IndexManager toAPI() {
+		return indexManager.toAPI();
+	}
+
```
yrodiere, line 117: This is never used.

```suggestion
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/impl/PojoIndexedTypeManager.java
```diff
@@ -10,6 +10,7 @@
 import java.util.Set;
 import java.util.function.Supplier;
 
+import org.hibernate.search.engine.backend.index.IndexManager;
```
yrodiere, line 13:
```suggestion
```

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/schema/management/ElasticsearchIndexSchemaManagerExporterIT.java
```diff
@@ -0,0 +1,134 @@
...
+						"        \"custom-pattern-replace\": {" +
+						"          \"type\": \"pattern_replace\"," +
+						"          \"pattern\": \"[^0-9]\"," +
+						"          \"replacement\": \"0\"," +
+						"          \"tags\": \"CASE_INSENSITIVE|COMMENTS\"" +
+						"        }" +
+						"      }" +
+						"    }" +
+						"  }" +
+						"}",
```
yrodiere, line 125: We'll need to run those tests against all Elasticsearch/OpenSearch versions. I suspect you won't get the same JSON in all case and thus get failures...

It might be a good idea to be less strict in your assertions, i.e. use `assertJsonEqualsIgnoringUnknownFields` and only check for a few fields. Also, I'd disable this test for ES 6 and below, because it's going to fail on older versions for sure (because of the concept of named types).

integrationtest/mapper/pojo-standalone-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/standalone/realbackend/schema/management/ElasticsearchSchemaManagerExporterIT.java
```diff
@@ -0,0 +1,246 @@
...
+						"      \"_entity_type\": {" +
+						"        \"type\": \"keyword\"," +
+						"        \"index\": false," +
+						"        \"doc_values\": true" +
+						"      }" +
+						"    }," +
+						"    \"dynamic\": \"strict\"" +
+						"  }," +
+						"  \"settings\": {}" +
+						"}",
```
yrodiere, line 82: Same as above:

> We'll need to run those tests against all Elasticsearch/OpenSearch versions. I suspect you won't get the same JSON in all case and thus get failures...
> 
> It might be a good idea to be less strict in your assertions, i.e. use `assertJsonEqualsIgnoringUnknownFields` and only check for a few fields. Also, I'd disable this test for ES 6 and below, because it's going to fail on older versions for sure (because of the concept of named types).

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/schema/management/impl/FileSearchSchemaCollector.java
```diff
@@ -0,0 +1,70 @@
...
+
+	public FileSearchSchemaCollector(Path targetDirectory) {
+		this.targetDirectory = targetDirectory;
+	}
+
+	@Override
+	public void indexSchema(Optional<String> backendName, String indexName, SchemaExport export) {
+		Path target = targetDirectory
+				.resolve( backendName
+						.map( bn -> FileSystems.getDefault().getPath( "backends", bn ) )
```
yrodiere, line 39:
```suggestion
						.map( bn -> Paths.get( "backends", bn ) )
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/schema/management/impl/FileSearchSchemaCollector.java
```diff
@@ -0,0 +1,70 @@
...
+import org.hibernate.search.engine.mapper.schema.management.SearchSchemaCollector;
+import org.hibernate.search.engine.reporting.spi.EventContexts;
+import org.hibernate.search.mapper.pojo.logging.impl.Log;
+import org.hibernate.search.util.common.logging.impl.LoggerFactory;
+
+public class FileSearchSchemaCollector implements SearchSchemaCollector {
+
+	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
+
+	private static final Path DEFAULT_BACKEND_PATH = FileSystems.getDefault().getPath( "backend" );
```
yrodiere, line 27:
```suggestion
	private static final Path DEFAULT_BACKEND_PATH = Paths.get( "backend" );
```

yrodiere (commented):
Thanks! Looking good overall, I mostly just have a few comments about packages and tests.

integrationtest/mapper/pojo-standalone-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/standalone/realbackend/schema/management/ElasticsearchSchemaManagerExporterIT.java
marko-bekhta: hmm right, I thought that that might be the case ... but then this particular test has `withProperty( "hibernate.search.backend.version", "8.6" )` so I'd hopped that that would force the version and keep the json stable? Or you are saying do not force a version but get it form the backend and then pass it in as if there's no ES running ? 😃

yrodiere: Ah right, that would work. I missed that.

Though if we only test one version, I'd rather we make sure we test the latest one... I'm absolutely convinced we'll never update this `"8.6"` :)

If you want to test this offline, maybe use `org.hibernate.search.util.impl.integrationtest.backend.elasticsearch.dialect.ElasticsearchTestDialect#getActualVersion`, and disable the test for older versions of Elasticsearch?

marko-bekhta: ok will do 😃

marko-bekhta:
```
[ERROR] Constraint: hsearch:TypesExtendingTypeFromAnotherModuleMustHaveModuleSpecificKeywordInName
[ERROR] Severity: MAJOR
[ERROR] Number of rows: 2
[ERROR]             Main (non-test) types extending/implementing a Hibernate Search type from another module must have a module-specific keyword in their name,
[ERROR]             either at the very start or just after "Abstract".
[ERROR]             This allows to more easily understand which module a given type comes from.
[ERROR]             Exceptions are allowed when:
[ERROR]             - the misnamed type is an anonymous type or a generated type;
[ERROR]             - or the misnamed type is deprecated;
[ERROR]             - or the implemented type is deprecated;
[ERROR]             - or the implementing type is an inner/nested type;
[ERROR]             and its surrounding type does use the module-specific keyword;
[ERROR]             - or the implemented type is in a util module,
[ERROR]             in which case the implemented interface may just be a detail.
[ERROR]         
[ERROR]   misnamedTypeArtifact.moduleSpecificKeyword=Pojo, misnamedType=org.hibernate.search.mapper.pojo.schema.management.impl.FileSearchSchemaCollector, externalParentType=org.hibernate.search.engine.backend.schema.management.IndexSchemaCollector
[ERROR]   misnamedTypeArtifact.moduleSpecificKeyword=Pojo, misnamedType=org.hibernate.search.mapper.pojo.schema.management.SearchSchemaCollector, externalParentType=org.hibernate.search.engine.backend.schema.management.IndexSchemaCollector
[ERROR] -------------------------------------------------------------------
[ERROR] 
```
I cannot say that I like the idea of naming these `PojoSearchSchemaCollector` and `PojoFileSearchSchemaCollector` / `PojoSearchSchemaCollectorToFiles` ... 🥲

yrodiere:
> I cannot say that I like the idea of naming these `PojoSearchSchemaCollector` and `PojoFileSearchSchemaCollector` / `PojoSearchSchemaCollectorToFiles` ...

The one about `FileSearchSchemaCollector` will disappear once we solve the problem for `SearchSchemaCollector`, I think.

Two solutions:

1. Add an exception to the JQAssistant rule...
2. Use composition instead of inheritance. I.e. make `IndexSchemaCollector` SPI, don't extend it in `SearchSchemaCollector`, declare an `exportExpectedSchema` method in `SearchSchemaCollector`, and in `org.hibernate.search.mapper.pojo.schema.management.impl.PojoScopeSchemaManagerImpl#exportExpectedSchema(org.hibernate.search.mapper.pojo.schema.management.SearchSchemaCollector)` create an instance of `IndexSchemaCollector` that delegates to the `SearchSchemaCollector`.

Your choice!

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/schema/management/SearchSchemaCollector.java
```diff
@@ -0,0 +1,31 @@
...
+	 * Called when an {@link SchemaExport index schema export} is encountered.
+	 *
+	 * @param backendName The name of the index's backend, or {@link Optional#empty()} for the default backend.
+	 * @param indexName The name of the index.
+	 * @param export The index schema export from a schema manager this collector is passed to.
+	 */
+	default void indexSchema(Optional<String> backendName, String indexName, SchemaExport export) {
+		// The default implementation that does nothing.
+		// Done so on purpose so when any other visiting methods are added to this interface users would still be
+		// able to only override/implement just those visitors that they are interested in.
```
yrodiere, line 28: You can add the default implementation then? It's a backwards-compatible change.

documentation/src/main/asciidoc/reference/schema.asciidoc
```diff
@@ -258,3 +258,85 @@ on an AWS Elasticsearch cluster.
...
+include::{sourcedir}/org/hibernate/search/documentation/mapper/orm/schema/management/HibernateOrmSchemaManagerIT.java[tags=schema-export]
+----
+<1> Retrieve the `SearchSchemaManager` from a `SearchSession`.
+<2> Export the schema to a target directory.
+====
+
+[[schema-management-export-custom]]
+=== Exporting to a custom collector
+
+<<mapper-orm-schema-management-manager,Search schema managers>> allow to walk through the schema exports based on the data such managers contains.
```
yrodiere, line 303: Just a grammar mistake I didn't spot earlier.

```suggestion
<<mapper-orm-schema-management-manager,Search schema managers>> allow walking through the schema exports based on the data such managers contains.
```

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/schema/management/impl/SearchSchemaManagerImpl.java
```diff
@@ -56,6 +58,16 @@ public void dropAndCreate() {
...
 
+	@Override
+	public void exportExpectedSchema(SearchSchemaCollector collector) {
+		delegate.exportExpectedSchema( collector );
+	}
+
+	@Override
+	public void exportExpectedSchema(Path targetDirectory) {
+		delegate.exportExpectedSchema( targetDirectory );
+	}
```
yrodiere, line 69: It's odd, Sonar reports this lines as not covered in tests: https://sonarcloud.io/component_measures?metric=new_coverage&selected=org.hibernate.search%3Ahibernate-search-parent%3Amapper%2Fpojo-standalone%2Fsrc%2Fmain%2Fjava%2Forg%2Fhibernate%2Fsearch%2Fmapper%2Fpojo%2Fstandalone%2Fschema%2Fmanagement%2Fimpl%2FSearchSchemaManagerImpl.java&view=list&pullRequest=3437&id=org.hibernate.search%3Ahibernate-search-parent

but from what I can see they should be, because of `integrationtest/mapper/pojo-standalone-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/standalone/realbackend/schema/management/ElasticsearchSchemaManagerExporterIT.java` in particular.

Could you check those tests are actually getting executed?

Maybe it's just a problem of this integration test module (which is relatively new, I believe) not being included in the aggregated test coverage report... no dependency for that module in `build/reports/pom.xml`?

yrodiere (approved):
Thanks. Just a few odd things, feel free to merge once that's solved!

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/schema/management/impl/SearchSchemaManagerImpl.java
marko-bekhta: tests are being executed:
```
[INFO] --- maven-failsafe-plugin:3.0.0-M8:integration-test (it-elasticsearch) @ hibernate-search-integrationtest-mapper-pojo-standalone-realbackend ---
[INFO] Using configured provider org.apache.maven.surefire.junitcore.JUnitCoreProvider
[INFO] 
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.hibernate.search.integrationtest.mapper.pojo.standalone.realbackend.EntityAsGraphSmokeIT
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.672 s - in org.hibernate.search.integrationtest.mapper.pojo.standalone.realbackend.EntityAsGraphSmokeIT
[INFO] Running org.hibernate.search.integrationtest.mapper.pojo.standalone.realbackend.EntityAsTreeSmokeIT
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.537 s - in org.hibernate.search.integrationtest.mapper.pojo.standalone.realbackend.EntityAsTreeSmokeIT
[INFO] Running org.hibernate.search.integrationtest.mapper.pojo.standalone.realbackend.mapping.DecimalScaleMappingIT
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.027 s - in org.hibernate.search.integrationtest.mapper.pojo.standalone.realbackend.mapping.DecimalScaleMappingIT
[INFO] Running org.hibernate.search.integrationtest.mapper.pojo.standalone.realbackend.schema.management.ElasticsearchSchemaManagerExporterIT
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.057 s - in org.hibernate.search.integrationtest.mapper.pojo.standalone.realbackend.schema.management.ElasticsearchSchemaManagerExporterIT
```
so hopefully the problem is indeed in the missing dependency in the reporting module

yrodiere: Looks like this fixed it. Thanks!

yrodiere (approved):

PR #3438, opened 2023-03-08
HSEARCH-4808 Add keep alive configuration property

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4808

If you'd have an idea on how we can test a positive scenario for this property - that would be great 😃

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/cfg/ElasticsearchBackendSettings.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/cfg/ElasticsearchBackendSettings.java
@@ -342,6 +342,21 @@ public final class ElasticsearchBackendSettings {
 	public static final String SCROLL_TIMEOUT = "scroll_timeout";
 
 	/**
+	 * How long connections to the Elasticsearch cluster can be kept idle.
+	 * <p>
+	 * Expects a positive Long value of milliseconds, such as 60000,
+	 * or a String that can be parsed into such Integer value.
+	 * <p>
+	 * If the response from an Elasticsearch cluster contains a {@code Keep-Alive} header,
+	 * then the effective max idle time will be whichever is lower:
+	 * the duration from the {@code Keep-Alive} header or the value of this property (if set).
+	 * <p>
+	 * If this property is not set, only the {@code Keep-Alive} header is considered,
+	 * and if it's absent, idle connections will be kept forever.
+	 */
+	public static final String MAX_KEEP_ALIVE = "max_keep_alive";
+
+	/**
 	 * Default values for the different settings if no values are given.
 	 */
 	public static final class Defaults {
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/ElasticsearchClientFactoryImpl.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/ElasticsearchClientFactoryImpl.java
@@ -129,6 +129,11 @@ public class ElasticsearchClientFactoryImpl implements ElasticsearchClientFactor
 			.asBeanReference( ElasticsearchHttpClientConfigurer.class )
 			.build();
 
+	private static final OptionalConfigurationProperty<Long> MAX_KEEP_ALIVE =
+			ConfigurationProperty.forKey( ElasticsearchBackendSettings.MAX_KEEP_ALIVE )
+					.asLongStrictlyPositive()
+					.build();
+
 	@Override
 	public ElasticsearchClientImplementor create(BeanResolver beanResolver, ConfigurationPropertySource propertySource,
 			ThreadProvider threadProvider, String threadNamePrefix,
@@ -258,6 +263,11 @@ public class ElasticsearchClientFactoryImpl implements ElasticsearchClientFactor
 			);
 
 			builder.setDefaultCredentialsProvider( credentialsProvider );
+		}
+
+		Optional<Long> maxKeepAlive = MAX_KEEP_ALIVE.get( propertySource );
+		if ( maxKeepAlive.isPresent() ) {
+			builder.setKeepAliveStrategy( new CustomConnectionKeepAliveStrategy( maxKeepAlive.get() ) );
 		}
 
 		ElasticsearchHttpClientConfigurationContextImpl clientConfigurationContext =
--- a/documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
+++ b/documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
@@ -291,6 +291,23 @@ The default for this property is `10`.
 +
 These properties expect a positive <<configuration-property-types,Integer value>>, such as `20`.
 
+Keep Alive::
++
+[source]
+----
+hibernate.search.backend.max_keep_alive = 10000
+----
+* `max_keep_alive` defines how long connections to the Elasticsearch cluster can be kept idle.
++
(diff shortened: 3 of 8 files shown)


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/cfg/ElasticsearchBackendSettings.java
```diff
@@ -341,6 +342,22 @@ private ElasticsearchBackendSettings() {
...
+	 * <p>
+	 * Expects a positive Integer value of milliseconds, such as 60000,
+	 * or a String that can be parsed into such Integer value.
+	 * <p>
+	 * If received response from an Elasticsearch cluster contains a {@code Keep-Alive} header - then its timeout value
+	 * will be honored if it's less than the one configured with this property.
+	 * <p>
+	 * If the value is not set then the default {@link ConnectionKeepAliveStrategy keep alive strategy} will be used.
+	 * <p>
+	 * Value is not set by default.
```
yrodiere, line 357: This is a bit confusing, especially the default since it requires knowning what `ConnectionKeepAliveStrategy` does exactly. Let's sum it up?

```suggestion
	 * If the response from an Elasticsearch cluster contains a {@code Keep-Alive} header,
	 * then the effective max idle time will be whichever is lower:
	 * the duration from the {@code Keep-Alive} header or the value of this property (if set).
	 * <p>
	 * If this property is not set, only the {@code Keep-Alive} header is considered,
	 * and if it's absent, idle connections will be kept forever.
```

documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
```diff
@@ -291,6 +291,14 @@ The default for this property is `10`.
 +
 These properties expect a positive <<configuration-property-types,Integer value>>, such as `20`.
 
+Keep Alive::
++
+[source]
+----
+hibernate.search.backend.max_keep_alive = 10000
+----
+* `max_keep_alive` the duration of time in milliseconds which the connection can be safely kept idle.
```
yrodiere, line 300:
```suggestion
* `max_keep_alive` defines how long connections to the Elasticsearch cluster can be kept idle.
+
If the response from an Elasticsearch cluster contains a `Keep-Alive` header,
then the effective max idle time will be whichever is lower:
the duration from the `Keep-Alive` header or the value of this property (if set).
+
If this property is not set, only the `Keep-Alive` header is considered,
and if it's absent, idle connections will be kept forever.
```

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/ElasticsearchClientFactoryImpl.java
```diff
@@ -129,6 +129,11 @@ public class ElasticsearchClientFactoryImpl implements ElasticsearchClientFactor
 			.asBeanReference( ElasticsearchHttpClientConfigurer.class )
 			.build();
 
+	private static final OptionalConfigurationProperty<Long> MAX_KEEP_ALIVE =
+			ConfigurationProperty.forKey( ElasticsearchBackendSettings.MAX_KEEP_ALIVE )
+					.asLongPositiveOrZero()
```
yrodiere, line 134: Note that `0` is apparently interpreted as "keep alive forever"... That should probably be documented?

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/client/ElasticsearchClientFactoryImplIT.java
```diff
@@ -1008,6 +1008,19 @@ public void clientInstance() throws IOException {
 		}
 	}
 
+	@Test
+	public void maxKeepAliveNegativeValue() {
```
yrodiere, line 1012: To test "positive" cases, you may be able to leverage `ElasticsearchHttpClientConfigurer` to insert some sort of spy on connection opening/closing. That should allow you to test that connections are kept alive as desired.

yrodiere (commented):
Thanks, comments below.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/client/impl/ElasticsearchClientFactoryImpl.java
marko-bekhta: hmm, from the docs:
> A value of 0 or less may be returned to indicate that there is no suitable suggestion.

that made me think that that would mean that it'll be up to a `ConnectionReuseStrategy` to do whatever it wants... 

Maybe it's better to just make our property strictly positive and not allow zero?

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/client/ElasticsearchClientFactoryImplIT.java
marko-bekhta: I haven't found any method other than "proxying" the connection pool manager to check which connections are released.

documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
```diff
@@ -291,6 +291,21 @@ The default for this property is `10`.
...
 These properties expect a positive <<configuration-property-types,Integer value>>, such as `20`.
 
+Keep Alive::
++
+[source]
+----
+hibernate.search.backend.max_keep_alive = 10000
+----
+* `max_keep_alive` defines how long connections to the Elasticsearch cluster can be kept idle.
++
```
yrodiere, line 301: Oops, forgot about the property type.

```suggestion
* `max_keep_alive` defines how long connections to the Elasticsearch cluster can be kept idle.
+
Expects a positive <<configuration-property-types,Long value>> in milliseconds, such as `60000`.
+
```

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/client/ElasticsearchClientFactoryImplIT.java
yrodiere: Works for me, thanks

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/cfg/ElasticsearchBackendSettings.java
```diff
@@ -341,6 +341,21 @@ private ElasticsearchBackendSettings() {
 	 */
 	public static final String SCROLL_TIMEOUT = "scroll_timeout";
 
+	/**
+	 * How long connections to the Elasticsearch cluster can be kept idle.
+	 * <p>
+	 * Expects a positive Integer value of milliseconds, such as 60000,
```
yrodiere, line 347:
```suggestion
	 * Expects a positive Long value of milliseconds, such as 60000,
```

yrodiere (commented):
Thanks. Just a few things in the documentation (my fault...). Once fixed, feel free to merge.

documentation/src/main/asciidoc/reference/backend-elasticsearch.asciidoc
marko-bekhta: 🙈😃

yrodiere (approved):

yrodiere:
Merged, thanks!

PR #3442, opened 2023-03-14
HSEARCH-4810 Fix flaky test with MSSQL

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4810

It was easier this time, as I was able to simply avoid hitting the database concurrently, as the concurrent access was an assertion in tests.


yrodiere:
It passed on all DBs, so I'll merge and hope for the best :)

PR #3447, opened 2023-03-20
HSEARCH-4819 Use the appropriate ORM 6 Dialects in v5 migration helper tests

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4819

Searched for `org\.hibernate\.dialect\..+Dialect`, and only these two places that failed in the builds had the dialects for replacement... 
I thought we might get away y simply updating the dialects in the main code ... but the classes are on different sides of the hierarchy, so that check would fail 😞.
Changes are here https://github.com/hibernate/hibernate-search/commit/199b5f166aa9c866d3604cc9e30c0fc69965729c, and this PR is to update the patches. 

Should I backport this patch to 6.1 or just cherrypick the commit into `wip/6.1/dependency-update/orm6.2`


yrodiere (approved):
Thanks, LGTM.

> Should I backport this patch to 6.1 or just cherrypick the commit into wip/6.1/dependency-update/orm6.2

Please backport if possible, that'll be one fewer quirk in the `wip/*` branches.

PR #3458, opened 2023-03-30
HSEARCH-4831 Fix Hibernate Search tests against latest ORM 6.2/6.3

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4831

Tested locally against the latest 6.2, and this fixes the problem.

Backport to 6.1: #3459


marko-bekhta (commented):
Looks nice 😃 
Quick question - since `SecurityManager` is for removal and `StackWalker` is unavailable in 8, I wondered if something like `new RuntimeException().getStackTrace()` wouldn't do the trick?

yrodiere:
> Quick question - since `SecurityManager` is for removal and `StackWalker` is unavailable in 8, I wondered if something like `new RuntimeException().getStackTrace()` wouldn't do the trick?

In short: yes it could have worked, but I'm overengineering stuff so that we get the best solution as soon as we move to a JDK 11 baseline...

`Thread.getStackTrace()` and similar are not recommended it's slow, and you don't get all info, for example you don't have actual `Class` objects, just the class names. But I agree that here, it wouldn't have mattered...

FWIW I didn't start from scratch, most of this code is very similar to what you'll find in `org.hibernate.bytecode.internal.bytebuddy.HibernateMethodLookupDispatcher`.

PR #3462, opened 2023-03-31
HSEARCH-1383 Ability to disable automatic indexing programmatically at the session level

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-1383

The filter is applied in the process method for indexing the plan for this first iteration, and then there's also usage of it in the event listener to reduce the amount of work.


marko-bekhta:
So the idea of filtering inside the ORM event listener somewhat failed 🥲. Since we attach the listener to the session factory, it behaves more like an application filter rather than a session filter.
I suppose we can add checks to various add/update/delete methods of `AbstractPojoTypeIndexingPlan` to prevent from building up the state....

yrodiere:
> So the idea of filtering inside the ORM event listener somewhat failed

It can work, but you need the listener to delegate those checks to something that's session-bound. 

I gave it some thought, and I think the best approach would be for you to pass the type context to `HibernateSearchEventListener#getCurrentIndexingPlan`/`#getCurrentIndexingPlanIfExisting` and have those methods return some type-specific indexing plan.

I created a branch to show what I mean, feel free to take those commits in your own branch and adapt them as necessary: https://github.com/hibernate/hibernate-search/compare/main...yrodiere:HSEARCH-1383

marko-bekhta:
Hopefully, it is a tad bit better now 🙈 
by the way, I haven't used your suggestion on how to check if a type is an entity, as it didn't work as a check for a mapped super type, and I thought we wanted to give the user an option to pass these supertypes to this filter (see `HibernateOrmTypeContextContainer.java#isManagedType()` I'm not sure if there's a better way to check it, but if not maybe that's something to add to ORM?).

yrodiere:
> I thought we wanted to give the user an option to pass these supertypes to this filter

I don't think we do. I believe mapped superclasses can be interfaces, and that opens a whole can of worms due to multiple inheritance: what if a class extends two mapped "superinterfaces", one being included and the other being excluded?

Better keep it simple, IMO.

marko-bekhta:
ohhh OK 🙈, then some changes are in order 😄 give me a moment 😃

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/automaticindexing/filter/PojoAutomaticIndexingTypeFilterContext.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+package org.hibernate.search.mapper.pojo.automaticindexing.filter;
+
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.Indexed;
+import org.hibernate.search.util.common.annotation.Incubating;
+
+/**
+ * A context that helps with the indexing filter configuration.
+ * <p>
+ * Note, that only indexed types can be passed to this context's methods. Passing any other types will lead to an
```
yrodiere, line 15: This is a curiously German syntax :) "Beachten Sie, dass [...]"
I don't think there's a comma here in English.

```suggestion
 * Note that only indexed types can be passed to this context's methods. Passing any other types will lead to an
```

documentation/src/test/java/org/hibernate/search/documentation/mapper/orm/automaticindexing/HibernateOrmAutomaticIndexingFilterIT.java
```diff
@@ -0,0 +1,166 @@
...
+		this.entityManagerFactory = setupHelper.start()
+				.setup( EntityA.class, EntityExtendsA1.class, EntityExtendsA2.class );
+	}
+
+	@Before
+	public void clearFilter() throws Exception {
+		Search.automaticIndexingFilter(
+				entityManagerFactory,
+				ctx -> { /*clear out any settings from tests*/ }
+		);
```
yrodiere, line 48: I don't believe this should be necessary since a new entity factory will be created for each test method.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/logging/impl/Log.java
```diff
@@ -321,4 +321,16 @@ void failedToResolveStateRepresentation(String path, @FormatWith(EventContextFor
...
+	@Message(id = ID_OFFSET + 124,
+			value = "'%1$s' cannot be included and excluded at the same time within one filter. " +
+					"Already included types: '%2$s'. " +
+					"Already excluded types: '%3$s'.")
+	SearchException automaticIndexingFilterCannotIncludeExcludeSameType(PojoRawTypeIdentifier<?> typeIdentifier,
+			Set<PojoRawTypeIdentifier<?>> includes, Set<PojoRawTypeIdentifier<?>> excludes);
+
+	@Message(id = ID_OFFSET + 125, value = "Use a different coordination strategy or an application-level automatic indexing filter. " +
+			"The outbox polling coordination strategy allows only disabling the indexing of all types at the session level. " +
+			"Applying a session-level automatic indexing filter that does not disable all types is prohibited.")
```
yrodiere, line 334:
```suggestion
	@Message(id = ID_OFFSET + 125, value = "Unable to apply the given filter at the session level with the outbox polling coordination strategy. " +
			"With this coordination strategy, applying a session-level automatic indexing filter is only allowed if it excludes all types.")
```

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/automaticindexing/filter/impl/HibernateOrmAutomaticIndexingTypeFilterContext.java
```diff
@@ -0,0 +1,158 @@
...
+				Collections.unmodifiableSet( allIncludes ),
+				Collections.unmodifiableSet( allExcludes ),
+				allTypesProcessed
+		);
+	}
+
+	private Optional<Class<?>> findClosestClass(Class<?> current, Set<PojoRawTypeIdentifier<?>> collection) {
+		Class<?> closest = Object.class;
+		for ( PojoRawTypeIdentifier<?> identifier : collection ) {
+			if ( identifier.javaClass().isAssignableFrom( current ) ) {
```
yrodiere, line 138: You can't do that kind of things with named types (see `org.hibernate.search.integrationtest.mapper.orm.dynamicmap.DynamicMapBaseIT`).

Really, you need information that's only available on startup in order to handle that kind of things.

Please reconsider relying on a map `type => subtypes` built on startup like `org.hibernate.search.mapper.pojo.mapping.impl.PojoTypeManagerContainer#indexedBySuperType`, as I suggested some time ago :)

By the way, you might be able to implement this class in the `pojo-base` module, which would make all this available to the pojo-standalone module. I don't know if it's worth the trouble though?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/automaticindexing/filter/impl/HibernateOrmAutomaticIndexingTypeFilterContext.java
```diff
@@ -0,0 +1,158 @@
...
+				}
+				else if ( closestInclude.isPresent() ) {
+					allIncludes.add( typedIdentifier );
+				}
+				else {
+					// if we don't find either include or exclude then we will defer the decision to either app filter or to a default (include)
+					// but that will happen in the filter.
+					allTypesProcessed = false;
+				}
+			}
```
yrodiere, line 124: A slightly simpler (I think) algorithm would just build a list of all types referenced in includes/excludes, sort them (top classes first), then go through that list in order, applying filters as it goes:

* A is included => add all subclasses of A to `allIncluded`, remove them from `allExcluded`
* B is excluded => remove all subclasses of B from `allIncluded`, add them to `allExcluded`

Feel free to keep this algorithm though, I'm just mentioning this in case you have a hard time adapting it to my suggestion below.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/mapping/impl/HibernateOrmTypeContextContainer.java
```diff
@@ -99,6 +107,12 @@ private HibernateOrmTypeContextContainer(Builder builder, SessionFactoryImplemen
 			byJpaEntityNameContent.put( typeContext.jpaEntityName(), typeContext );
 
 			byHibernateOrmEntityNameContent.put( typeContext.hibernateOrmEntityName(), typeContext );
+
+			for ( PojoRawTypeIdentifier<?> superType : typeContext.ascendingSuperTypes() ) {
+				if ( isManagedType( sessionFactory, superType.javaClass() ) ) {
+					indexedWithSuperTypesByExactClassContent.put( superType.javaClass(), superType );
```
yrodiere, line 113: So... not necessarily indexed then? This adds contained types.

Maybe rename that stuff to `allIncludingSuperTypesByExactClassContent`, or something like that?

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/AbstractAutomaticIndexingFilterIT.java
```diff
@@ -0,0 +1,436 @@
...
+import org.hibernate.search.util.impl.integrationtest.common.rule.BackendMock;
+import org.hibernate.search.util.impl.integrationtest.mapper.orm.OrmSetupHelper;
+import org.hibernate.search.util.impl.integrationtest.mapper.orm.ReusableOrmSetupHolder;
+
+import org.junit.Before;
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.rules.MethodRule;
+
+public class AbstractAutomaticIndexingFilterIT {
```
yrodiere, line 34:
```suggestion
public abstract class AbstractAutomaticIndexingFilterIT {
```

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/automaticindexing/filter/impl/HibernateOrmApplicationAutomaticIndexingTypeFilter.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.hibernate.search.mapper.orm.mapping.impl.HibernateOrmMapping;
+import org.hibernate.search.mapper.pojo.automaticindexing.filter.PojoAutomaticIndexingTypeFilterConfigurer;
+import org.hibernate.search.mapper.pojo.automaticindexing.filter.spi.PojoAutomaticIndexingTypeFilterHolder;
+
+public final class HibernateOrmApplicationAutomaticIndexingTypeFilter {
+
+	private static final PojoAutomaticIndexingTypeFilterHolder INDEXING_TYPE_FILTER_HOLDER = new PojoAutomaticIndexingTypeFilterHolder(
+			HibernateOrmAutomaticIndexingTypeFilter.IncludeAll.INSTANCE
+	);
```
yrodiere, line 17: Why isn't this just held in `SearchMapping`?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/mapping/impl/HibernateOrmTypeContextContainer.java
```diff
@@ -109,6 +123,17 @@ private HibernateOrmTypeContextContainer(Builder builder, SessionFactoryImplemen
...
+	}
+
+	private boolean isManagedType(SessionFactoryImplementor sessionFactory, Class<?> javaClass) {
+		try {
+			sessionFactory.getMetamodel().managedType( javaClass );
+		}
+		catch (Exception e) {
+			return false;
+		}
+		return true;
```
yrodiere, line 136: FYI in case you need it: this could be implemented in the `pojo-base` module, just using different metadata (`org.hibernate.search.mapper.pojo.model.additionalmetadata.impl.PojoTypeAdditionalMetadata#isEntity`, which does cover all Hibernate ORM entities as well).

yrodiere (changes requested):
Thanks! Here are a few comments, finally. Sorry for the delay.

documentation/src/test/java/org/hibernate/search/documentation/mapper/orm/automaticindexing/HibernateOrmAutomaticIndexingFilterIT.java
marko-bekhta: Right... it will go away when I move the app filter into search mapping. With what's in the current changes - we'd store a filter in a static variable, and that would not clean up itself (so I suppose one more reason not to do it the way it's done 😃 🙈)

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/mapping/impl/HibernateOrmTypeContextContainer.java
marko-bekhta: Thanks. I think I've tried this method at some point, but it would return false for a mapped supertype. I'll see how that change for the types goes 😃

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/ApplicationAutomaticIndexingFilterIT.java
```diff
@@ -0,0 +1,273 @@
...
+								.exclude( EntityA.INDEX )
+				)
+		).isInstanceOf( SearchException.class )
+				.hasMessageContainingAll(
+						EntityA.class.getName(),
+						"cannot be included and excluded at the same time within one filter",
+						"Already included types:",
+						"Already excluded types: '[]'"
+				);
+	}
```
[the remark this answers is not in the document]
marko-bekhta: Good point! Most of these cases are tested in the session filter test. These two here are just to make sure that app filter throws exceptions too. Since app/session filters are the same impl I've only kept these two tests here. I'll add a test for a string-that-is-not-a-name, for class -- there're a few already.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/impl/PojoMappingDelegateImpl.java
```diff
@@ -142,4 +145,19 @@ public PojoIndexer createIndexer(PojoWorkSessionContext context) {
...
+		return context.createFilter();
+	}
+
+	@Override
+	public ConfiguredSearchIndexingPlanFilter configuredSearchIndexingPlanFilter(SearchIndexingPlanFilter filter,
+			ConfiguredSearchIndexingPlanFilter fallback) {
+		SearchIndexingPlanFilterContextImpl context = new SearchIndexingPlanFilterContextImpl( typeManagers );
+		filter.apply( context );
+		return context.createFilter( fallback );
+	}
```
yrodiere, line 162: If you just expose the second method and pass `null` where relevant, you can get rid of the `*Creator` interface and expose the method directly on `HibernateOrmSearchSessionMappingContext`.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/work/SearchIndexingPlanFilterContext.java
```diff
@@ -0,0 +1,75 @@
...
+package org.hibernate.search.mapper.pojo.work;
+
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.Indexed;
+import org.hibernate.search.util.common.annotation.Incubating;
+
+/**
+ * A context that helps with the indexing filter configuration.
+ * <p>
+ * Note that only indexed types can be passed to this context's methods. Passing any other types will lead to an
+ * exception being thrown. Methods only accept classes and not interfaces.
```
yrodiere, line 16: And contained types... and supertypes of indexed or contained types...

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/work/SearchIndexingPlanFilterContext.java
```diff
@@ -0,0 +1,75 @@
...
+ *     <li>By default, types are included unless any of previous include/exclude rules apply</li>
+ * </ul>
+ */
+@Incubating
+public interface SearchIndexingPlanFilterContext {
+
+	/**
+	 * Specify an indexed type to include, along with (unless specified otherwise) all its subtypes.
+	 *
+	 * @param name The name of an indexed type.
```
yrodiere, line 42: It's not exactly just about indexed types though? Please review the rest of the javadoc, I suspect there's the same mistake elsewhere (including in other classes).

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/work/spi/ConfiguredSearchIndexingPlanFilter.java
```diff
@@ -0,0 +1,81 @@
...
+		public boolean supportsEventQueue() {
+			// cannot work with outbox polling since this filter would allow for all events to be persisted,
+			// but then an application level filter might think otherwise when the events will be processed.
+			return false;
+		}
+	}
+
+	class ExcludeAll implements ConfiguredSearchIndexingPlanFilter {
+
+		static final ExcludeAll INSTANCE = new ExcludeAll();
```
yrodiere, line 53: Singletons need a private constructor.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/work/spi/ConfiguredSearchIndexingPlanFilter.java
```diff
@@ -0,0 +1,81 @@
...
+		}
+		if ( excludes.isEmpty() ) {
+			return IncludeAll.INSTANCE;
+		}
+		return new Filter( includes );
+	}
+
+	class IncludeAll implements ConfiguredSearchIndexingPlanFilter {
+
+		public static final IncludeAll INSTANCE = new IncludeAll();
```
yrodiere, line 36: Singletons need a private constructor.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/work/spi/ConfiguredSearchIndexingPlanFilter.java
```diff
@@ -0,0 +1,81 @@
...
+
+		@Override
+		public boolean isIncluded(PojoRawTypeIdentifier<?> typeIdentifier) {
+			return true;
+		}
+
+		@Override
+		public boolean supportsEventQueue() {
```
yrodiere, line 44: `supportsAsyncProcessing` would better reflect the intent, I think.

marko-bekhta: 👍🏻 makes sense 😃. I wanted it to be aligned with `ConfiguredAutomaticIndexingStrategy#usesEventQueue()`

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/impl/PojoTypeManagerContainer.java
```diff
@@ -130,14 +159,23 @@ public <E> void addIndexed(PojoRawTypeModel<E> typeModel, PojoIndexedTypeManager
...
+						}
+					} );
 		}
 
-		public <E> void addContained(PojoContainedTypeManager<?, E> typeManager) {
+		public <E> void addContained(PojoRawTypeModel<E> typeModel, PojoContainedTypeManager<?, E> typeManager) {
 			contained.add( typeManager );
+			typeModel.descendingSuperTypes()
+					.map( PojoRawTypeModel::typeIdentifier )
+					.filter( clazz -> !clazz.javaClass().isInterface() )
```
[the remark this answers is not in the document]
marko-bekhta: Yeah... I've decided to leave it as this until I add a proper test for those dynamic types.
I suspect that the dynamic type would be a named type, so I thought we might get away by checking if a type is not named and is an interface or something along those lines 😃. But anyways, I'll add a test case and make the adjustments 🙈

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/work/spi/ConfiguredSearchIndexingPlanFilter.java
marko-bekhta: Ok, will do  😃 👍🏻

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/impl/PojoTypeManagerContainer.java
```diff
@@ -38,16 +38,19 @@ public static Builder builder() {
 
 	private final KeyValueProvider<PojoRawTypeIdentifier<?>, AbstractPojoTypeManager<?, ?>> byExactType;
 	private final KeyValueProvider<PojoRawTypeIdentifier<?>, PojoIndexedTypeManager<?, ?>> indexedByExactType;
+	private final KeyValueProvider<Class<?>, Set<PojoRawTypeIdentifier<?>>> allByClass;
```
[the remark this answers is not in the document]
marko-bekhta: I didn't think we should keep the entire type manager, as we only needed an identifier. will change it to `AbstractPojoTypeManager` or whatever common class would make sense 😃.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/impl/PojoTypeManagerContainer.java
```diff
@@ -89,6 +100,11 @@ public <E> AbstractPojoTypeManager<?, E> forExactType(PojoRawTypeIdentifier<E> t
 		return (AbstractPojoTypeManager<?, E>) byExactType.getOrFail( typeIdentifier );
 	}
 
+	@Override
+	public Set<PojoRawTypeIdentifier<?>> byClass(Class<?> klass) {
+		return allByClass.getOrFail( klass );
```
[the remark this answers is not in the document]
marko-bekhta: I've got rid of this one 🥲

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/logging/impl/Log.java
```diff
@@ -788,4 +788,23 @@ ConstructorProjectionApplicationException errorApplyingProjectionConstructor(
...
+	SearchException indexingPlanFilterCannotIncludeExcludeSameType(PojoRawTypeIdentifier<?> typeIdentifier,
+			Set<PojoRawTypeIdentifier<?>> includes, Set<PojoRawTypeIdentifier<?>> excludes);
+
+	@Message(id = ID_OFFSET + 130,
+			value = "No matching entity type for class '%1$s'."
+					+ " This class is neither an entity type mapped in Hibernate Search nor a superclass of such entity type."
+					+ " Note interfaces are not considered superclasses and are not permitted here."
+					+ " Valid classes are: %2$s")
+	SearchException unknownClassForNonInterfaceSuperType(Class<?> invalidClass,
+			Collection<Class<?>> validClasses);
```
yrodiere, line 809: Just noticed this wasn't using formatters. You might want to use:

* `@FormatWith(ClassFormatter.class)`
* `@FormatWith(CommaSeparatedClassesFormatter.class)`

yrodiere (commented):
LGTM apart from two minor comments. Feel free to merge once they're addressed.
Thanks!

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/work/impl/SearchIndexingPlanFilterContextImpl.java
```diff
@@ -0,0 +1,172 @@
...
+	public SearchIndexingPlanFilterContextImpl(PojoScopeTypeContextProvider typeManager) {
+		this.contextProvider = typeManager;
+
+		Map<Class<?>, PojoRawTypeIdentifier<?>> typeIdentifierByClassContent = new LinkedHashMap<>();
+		for ( PojoRawTypeIdentifier<?> identifier : typeManager.allNonInterfaceSuperTypes() ) {
+			Class<?> klass = identifier.javaClass();
+			// at this point we only care about classes since we'll be using this map to find identifiers by a class
+			// in include(Class)/exclude(Class) methods.
+			// for any dynamic types we are expecting users to use the include(String)/exclude(String) instead.
+			if ( !klass.isInterface() ) {
```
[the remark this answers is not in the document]
marko-bekhta: mmmm right.... me coming up with confusing behaviour and naming again 🙈 😄 sorry 
that should've been something like - `all-supertypes-of-indexed-and-contained-types-excluding-interfaces-unless-those-interfaces-are-named-interfaces-coming-from-dynamic-type-definitions`, but that name might be a bit too long even for Java 🙈

Overall the idea is that when a user calls include/exclude -- we want to convert a string/class to a type identifier. We cannot get the entire hierarchy at that point since if we do so and store it in the include/exclude collections, we can end up throwing an exception when the user starts to add some other sub/super types to the filter as those would already be in the collections. 
So basically, we need a converter from string/class to type identifier. And this converter should allow for `Object` and all other supertypes we want to be included.
Then once we are done collecting info from the user, we need a type-identifier -> subtypes so that we can build an actual filter. 

And here, this `KeyValueProvider` is only for those class methods, so I didn't want it to have a `Map` in there too... Alternatively, we could say that we would allow Map here and that if the user passes it then we include/exclude all those dynamic types... but I thought that what if a user decides to create their own entity and implement a map on it ... so I didn't want to go that route 😃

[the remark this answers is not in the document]
yrodiere: I'm having trouble understanding.

Let's try it another way: if you remove this check, what breaks?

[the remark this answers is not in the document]
marko-bekhta: If we'd remove this, then <Map.class, "OneOfDynamicTypes"> will be in the `KeyValueProvider`.
no current tests will fail. But we'd be able to make a call like:
```
context.exclude(Map.class)
```
and it will not fail. But since this `KeyValueProvider` only would point to one of the dynamic types we won't be able to say which type exactly will be excluded ....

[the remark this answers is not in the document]
yrodiere: Ok I see. Maybe you need such a test?

Anyway, why don't you restore the  `KeyValueProvider<Class, Set<Type context>>` that you had in previous iterations, and make sure to only put non-named types in there ? After all, there is no reason to build this map every single time you instantiate a context, since you can build it on startup and make it a key value provider.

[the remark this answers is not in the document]
marko-bekhta: +1 added a test. 
As for the `KeyValueProvider<Class, Set<Type context>> ` ... 
I could do something like:
```java
private final Map<Class<?>, PojoRawTypeModel<?>> indexedContainedSuperType = new LinkedHashMap<>();

public <E> void addIndexed(PojoRawTypeModel<E> typeModel, PojoIndexedTypeManager<?, E> typeManager) {
	indexed.add( typeManager );
        // .....
	typeModel.descendingSuperTypes()
			.filter( type -> !type.typeIdentifier().javaClass().isInterface() )
			.forEach( type -> {
				indexedContainedSuperType.put( type.typeIdentifier().javaClass(), type );
			} );
}
```
But that wouldn't be a `AbstractPojoTypeManager` or other similar `Type context`s. Or I can do a similar to what we have here - `KeyValueProvider<Class, Set<TypeIdentifier>>`, but then as you've mentioned earlier it wouldn't really be a `typeManagerContainer`.
This `KeyValueProvider` must contain a pair `<Object.class, "anything that we can get a TypeIdentifier<Object> from">`, since its only purpose is to convert `Class<?>` passed into `SearchIndexingPlanFilterContext#inlcude(Class)`/`SearchIndexingPlanFilterContext#exclude(Class)` to a `PojoRawTypeIdentifier`. We never have a `PojoIndexedTypeManager` for the `Object.class` right?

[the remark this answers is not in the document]
yrodiere: Ok, I just had a closer look at why you even need this map and... you can just use `PojoRawTypeIdentifier.of(clazz)` xD

It's not great, but that's what we do elsewhere already (`org.hibernate.search.mapper.pojo.standalone.mapping.impl.StandalonePojoMapping#createScope`).

As to failing when the class is an interface... I think it all boils down to failing when the type identifier doesn't match anything in `contextProvider.allNonInterfaceSuperTypes()`. And if there's a named type using an interface, its type id will never match `PojoRawTypeIdentifier.of(clazz)` (since that results in an unnamed type ID).

By the way I also noticed that `org.hibernate.search.mapper.pojo.scope.impl.PojoScopeTypeContextProvider#allByNonInterfaceSuperType` is implemented to rely on a `Map` (`allByNonInterfaceSuperType`) and return an empty optional when there's no match, but I don't think this should ever happen? We definitely expect everything passed to `include`/`exclude` to have an effect.

To sum up I'd suggest the following:

* don't build a map every single time you create a `SearchIndexingPlanFilterContextImpl`
* use `PojoRawTypeIdentifier.of(clazz)` to turn a class into a type id
* add a check to `include(Class)`/`exclude(Class)` that the type identifier is present in `allNonInterfaceSuperTypes()`. If it's not, you can throw `log.unknownClassForNonInterfaceSuperType()`. If necessary introduce a `allNonInterfaceSuperClasses()` that returns a set of `Class` instead of type identifiers, but that's mostly for the error message.
* add a test with `.include(Integer.class)`/`.exclude(...)` and check that it fails with a meaningful and appropriate message.
* [probably already done] add a test with `.include(someinterface)`/`.exclude(...)` and check that it fails with a meaningful and appropriate message
* [probably already done] add a test with `.include("someNameThatDoesNotMatchAnEntity")`/`.exclude(...)` and check that it fails with a meaningful and appropriate message

Also, unrelated, but I just thought about it:

* Add a test with `.include("validEntityName")`/`.exclude(...)` but with the entity name pointing to an entity that has absolutely no Hibernate Search annotations and is not `@IndexedEmbedded` anywhere either... but it has indexed subclasses. This is expected to work, I think? But I don't think it will.

Does all that make sense?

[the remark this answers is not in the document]
marko-bekhta: > you can just use PojoRawTypeIdentifier.of(clazz) xD

heyy !! that's a cheatcode 😆 ahah
I've seen that method before, and I thought that it wouldn't work if we had an indexed entity with a name specified, (I was thinking that in that case, the name won't be null and hence equals wouldn't work, and we wouldn't find anything in the keyvalue providers ...) ohhhhh I should've just tried that to make sure that it wasn't actually so 🙈 🤦🏻‍♂️ 😆 
and yes all makes sense, thanks 😃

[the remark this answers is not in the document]
marko-bekhta: > Add a test with .include("validEntityName")/.exclude(...) but with the entity name pointing to an entity that has absolutely no Hibernate Search annotations and is not @IndexedEmbedded anywhere either... but it has indexed subclasses. This is expected to work, I think? But I don't think it will.

Right, so this didn't work as you've been suspecting 😃 🎉.

There are two commits with new changes. The first one addresses your suggestions and adds a new key-value provider for the entity name since existing ones do not cover that last test case... I did put it in `PojoTypeManagerContainer` even though it is storing the `PojoRawTypeModel`, but I guess we won't have anything else for entity-supertypes", right?
The last one adds more tests for dynamic type cases when we try to pass names of a supertype of an indexed subtype or just a name of an entity that is not mapped in HSEARCH.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/impl/PojoTypeManagerContainer.java
```diff
@@ -39,16 +40,23 @@ public static Builder builder() {
...
+	private final Set<PojoRawTypeIdentifier<?>> allIndexedAndContainedTypes;
+	private final Set<PojoRawTypeIdentifier<?>> allNonInterfaceSuperTypes;
+	private final Set<Class<?>> allNonInterfaceSuperTypesClasses;
 
 	private PojoTypeManagerContainer(Builder builder) {
 		// Use a LinkedHashMap for deterministic iteration in the "all" set
 		Map<PojoRawTypeIdentifier<?>, AbstractPojoTypeManager<?, ?>> byExactTypeContent = new LinkedHashMap<>();
 		Map<PojoRawTypeIdentifier<?>, PojoIndexedTypeManager<?, ?>> indexedByExactTypeContent = new LinkedHashMap<>();
 		Map<String, AbstractPojoTypeManager<?, ?>> byEntityNameContent = new LinkedHashMap<>();
+		Map<String, PojoRawTypeModel<?>> typeModelByEntityNameContent = new LinkedHashMap<>();
```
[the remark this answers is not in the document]
marko-bekhta: yes sir 😃 👍🏻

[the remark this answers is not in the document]
yrodiere: Sorry :x But this time it's not just me nitpicking, it has real implications with native images...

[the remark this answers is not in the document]
marko-bekhta: hey, no worries 😃 I was hesitant to go with the type identifier since, at some point earlier, you mentioned that this thing should provide type managers, so that's why I've called it out in the comment so that you suggest a better thing to save in that map 😉

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/work/impl/SearchIndexingPlanFilterContextImpl.java
```diff
@@ -0,0 +1,157 @@
...
+		);
+	}
+
+	private void exclude(Set<PojoRawTypeIdentifier<?>> allIndexedAndContainedTypes,
+			Set<PojoRawTypeIdentifier<?>> allIncludes, Set<PojoRawTypeIdentifier<?>> allExcludes,
+			Set<PojoRawTypeIdentifier<?>> processed, PojoRawTypeIdentifier<?> typeIdentifier) {
+		contextProvider.allByNonInterfaceSuperType( typeIdentifier ).forEach( typeContext -> {
+			PojoRawTypeIdentifier<?> identifier = typeContext.typeIdentifier();
+			// we only care about indexed or contained types. Other types shouldn't be passed to the filter.
+			if ( allIndexedAndContainedTypes.contains( identifier ) ) {
```
yrodiere, line 128: As far as I can tell, all type contexts refer to either an indexed or a contained type, so this check will always pass and can be removed.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/work/impl/SearchIndexingPlanFilterContextImpl.java
```diff
@@ -0,0 +1,157 @@
...
+		} );
+	}
+
+	private void include(Set<PojoRawTypeIdentifier<?>> allIndexedAndContainedTypes,
+			Set<PojoRawTypeIdentifier<?>> allIncludes, Set<PojoRawTypeIdentifier<?>> allExcludes,
+			Set<PojoRawTypeIdentifier<?>> processed, PojoRawTypeIdentifier<?> typeIdentifier) {
+		contextProvider.allByNonInterfaceSuperType( typeIdentifier ).forEach( typeContext -> {
+			PojoRawTypeIdentifier<?> identifier = typeContext.typeIdentifier();
+			// we only care about indexed or contained types. Other types shouldn't be passed to the filter.
+			if ( allIndexedAndContainedTypes.contains( identifier ) ) {
```
yrodiere, line 142: As far as I can tell, all type contexts refer to either an indexed or a contained type, so this check will always pass and can be removed.

yrodiere (commented):
Thanks, LGTM, though I think I spotted an unnecessary check. Please have a look and then feel free to merge!

yrodiere (approved):

yrodiere:
Merged, thanks!

yrodiere:
Actually, not merged, GitHub is having problems...

PR #3468, opened 2023-04-06
HSEARCH-4811 Automated code formatting

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4811

There's a `format-sources` profile:
```
mvn -Pformat-sources
```
This will reformat the code and rearrange the imports. 
I've also added a style XML to our build module since https://github.com/hibernate/hibernate-ide-codestyles isn't published and downloading it at a build time didn't seem like a good idea 😉 

I haven't done the reformatting as it changes almost all the files ...


build/config/src/main/resources/hibernate_noorm_style.xml
```diff
@@ -0,0 +1,380 @@
+<?xml version="1.0" encoding="utf-8"?>
+<profiles version="21">
+	<profile kind="CodeFormatterProfile" name="hibernate-noorm-style" version="21">
```
yrodiere, line 3: This is a good basis for our configuration, but let's not limit ourselves to it: it's only used in OGM (discontinued) and Validator (unlikely to switch to automated formatting anytime soon).

I think we can adapt the codestyle to limit the amount of changes it will lead to, i.e. to match the effective codestyle in Hibernate Search.

For example, I'd rather keep this:

```
       private static final ConfigurationProperty<Boolean> SIGNING_ENABLED =
                       ConfigurationProperty.forKey( ElasticsearchAwsBackendSettings.SIGNING_ENABLED )
                                       .asBoolean()
                                       .withDefault( ElasticsearchAwsBackendSettings.Defaults.SIGNING_ENABLED )
                                       .build();
```

Rather than formatting it like this:

```
       private static final ConfigurationProperty<Boolean> SIGNING_ENABLED = ConfigurationProperty.forKey(
                       ElasticsearchAwsBackendSettings.SIGNING_ENABLED
       )
                       .asBoolean()
                       .withDefault( ElasticsearchAwsBackendSettings.Defaults.SIGNING_ENABLED )
                       .build();
```

AFAICS, most of the questionable formatting comes from line breaking rules... Could you look into making the codestyle match the current state of the code more closely? Note you might need some changes in the plugin config as well... In particular I wonder if there's an option to prevent the plugin from removing existing line breaks... there's usually one in IDEs.

pom.xml
```diff
@@ -3052,6 +3125,16 @@
...
 
+        <profile>
+            <id>format-sources</id>
+            <properties>
+                <impsort-maven-plugin.goal>sort</impsort-maven-plugin.goal>
+            </properties>
+            <build>
+                <defaultGoal>impsort:sort formatter:format</defaultGoal>
+            </build>
+        </profile>
```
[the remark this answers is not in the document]
yrodiere: Also, you'll probably be able to remove some checkstyle rules that are strictly just about code style.

yrodiere (commented):
Thanks. I gave it a first look, here are a few comments.

build/config/src/main/resources/hibernate_noorm_style.xml
marko-bekhta: Yeah... I'll double-check the plugin as I haven't found anything that would help us with that. I was hoping that something like:
```xml
<setting id="org.eclipse.jdt.core.formatter.join_wrapped_lines" value="false"/>
```
would help us... but that didn't work out either 😢. But I'll try to find something that works 😄 as otherwise we'll disappoint Emmanuel and/or Hardy - https://github.com/hibernate/hibernate-search/blob/58c62aea92e5c511ecf131ba9a4ca8f19fe85821/integrationtest/v5migrationhelper/engine/src/test/java/org/hibernate/search/test/dsl/SpatialDSLTest.java#L24-L26

pom.xml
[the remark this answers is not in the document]
yrodiere: Looks like checkstyle is still enabled?

I'm getting this:

> [ERROR] src/test/java/org/hibernate/search/engine/backend/work/execution/OperationSubmitterQueueTest.java:[72,43] (coding) OneStatementPerLine: Only one statement per line allowed.

build/config/src/main/resources/hibernate_noorm_style.xml
yrodiere: At first sight it looks better now, but there's still some very questionable changes... I don't know how configurable this thing is, but I'll mention what I spotted just in case you can do something about it :)

From what I can see we'd get even fewer changes if:

* line breaks were preferably inserted _before_ `implements`/`extends`/`&&`/`||`/`+`/`throws`.
* line breaks were preferably inserted outside of generic arguments (e.g. in `engine/src/main/java/org/hibernate/search/engine/backend/work/execution/OperationSubmitter.java`, `OperationSubmitter#submitToQueue`)
* `extends`/`implements` clauses split over multiple lines got automatically indented (e.g. in `engine/src/main/java/org/hibernate/search/engine/backend/document/model/spi/IndexObjectField.java`).
* generics split over multiple lines got automatically indented (e.g. in `engine/src/main/java/org/hibernate/search/engine/backend/document/model/spi/IndexObjectField.java`).
* ternary operators were preferably formatted with `?` / `:` on their own line when split over multiple lines (e.g. in `BeanResolverImpl#toSources`)
* (maybe) formatting for javadoc was disabled? I saw there's an option about that. I must admit it seems to do more harm than good, e.g. in `EngineSettings#BACKENDS`...
* Maybe more, feel free to discuss oddities on Zulip.

pom.xml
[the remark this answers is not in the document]
marko-bekhta: > Looks like checkstyle is still enabled?

Yes, I've only removed those rules about spaces, braces and other "cosmetic" stuff. 
This one-statement-per-line looked to me as something we might want to keep, as there's no strict equivalent for it in formatter settings. This error comes from a broken lambda line where the formatter did a poor job of cutting it in the middle... 
Once we will want to merge this in I plan to run the automatic formatting and then do some manual cleanup after (in some cases like this one the formatter won't complain and change things back after a manual update)

build/config/src/main/resources/hibernate_noorm_style.xml
marko-bekhta: > line breaks were preferably inserted before implements/extends/&&/||/+/throws.
 
this seems possible

> line breaks were preferably inserted outside of generic arguments (e.g. in engine/src/main/java/org/hibernate/search/engine/backend/work/execution/OperationSubmitter.java, OperationSubmitter#submitToQueue)

with this one it seems that it's either make this work (so that generic arguments in methods are not wrapped) or wrap things at class declaration level:
```java
public final class ElasticsearchSearchIndexScopeImpl
		extends AbstractSearchIndexScope<
				ElasticsearchSearchIndexScopeImpl,
				ElasticsearchIndexModel,
				ElasticsearchSearchIndexNodeContext,
				ElasticsearchSearchIndexCompositeNodeContext>
		implements ElasticsearchSearchIndexScope<ElasticsearchSearchIndexScopeImpl>,
		ElasticsearchSearchQueryIndexScope<ElasticsearchSearchIndexScopeImpl> {
```

> extends/implements clauses split over multiple lines got automatically indented (e.g. in engine/src/main/java/org/hibernate/search/engine/backend/document/model/spi/IndexObjectField.java).

this seems possible

> generics split over multiple lines got automatically indented (e.g. in engine/src/main/java/org/hibernate/search/engine/backend/document/model/spi/IndexObjectField.java).

Indention is controlled by one property and if we make this case work all other things get indented, so it seems it's better to leave this un-indented.

> ternary operators were preferably formatted with ? / : on their own line when split over multiple lines (e.g. in BeanResolverImpl#toSources)

this seems possible

> (maybe) formatting for javadoc was disabled? I saw there's an option about that. I must admit it seems to do more harm than good, e.g. in EngineSettings#BACKENDS...

this seems possible

yrodiere:
@marko-bekhta Gentle reminder that you told me you were going to rebase/polish this... :)

PR #3469, opened 2023-04-07
HSEARCH-4149 Add lowest/highest missing() sort options

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4149

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/sort/comparatorsource/impl/LuceneTextFieldComparatorSource.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/sort/comparatorsource/impl/LuceneTextFieldComparatorSource.java
@@ -32,16 +32,30 @@ public class LuceneTextFieldComparatorSource extends LuceneFieldComparatorSource
 
 	@Override
 	public FieldComparator<?> newComparator(String fieldname, int numHits, int sortPos, boolean reversed) {
-		final boolean sortMissingLast = missingLast() ^ reversed;
+		final boolean considerMissingHighest;
+		if ( SortMissingValue.MISSING_LOWEST.equals( missingValue ) ) {
+			considerMissingHighest = false;
+		}
+		else if ( SortMissingValue.MISSING_HIGHEST.equals( missingValue ) ) {
+			considerMissingHighest = true;
+		}
+		else if ( SortMissingValue.MISSING_LAST.equals( missingValue ) ) {
+			// To appear last, missing values must be considered highest, or lowest if the order is reversed.
+			considerMissingHighest = !reversed;
+		}
+		else { // SortMissingValue.MISSING_FIRST, the default
+			// To appear first, missing values must be considered lowest, or highest if the order is reversed.
+			considerMissingHighest = reversed;
+		}
 		TextMultiValuesToSingleValuesSource source =
 				TextMultiValuesToSingleValuesSource.fromField( fieldname, multiValueMode, nestedDocsProvider );
 
-		return new FieldComparator.TermOrdValComparator( numHits, fieldname, sortMissingLast ) {
+		return new FieldComparator.TermOrdValComparator( numHits, fieldname, considerMissingHighest ) {
 			@Override
 			protected SortedDocValues getSortedDocValues(LeafReaderContext context, String field) throws IOException {
 				SortedDocValues sortedDocValues = source.getValues( context );
 
-				if ( missingValue == null || missingFirst() || missingLast() ) {
+				if ( missingValue == null || isOneOfSortMissingValues() ) {
 					return sortedDocValues;
 				}
 
@@ -50,11 +64,7 @@ public class LuceneTextFieldComparatorSource extends LuceneFieldComparatorSource
 		};
 	}
 
-	private boolean missingFirst() {
-		return SortMissingValue.MISSING_FIRST.equals( missingValue );
-	}
-
-	private boolean missingLast() {
-		return SortMissingValue.MISSING_LAST.equals( missingValue );
+	private boolean isOneOfSortMissingValues() {
+		return missingValue instanceof SortMissingValue;
 	}
 }
--- a/documentation/src/main/asciidoc/reference/search-dsl-sort.asciidoc
+++ b/documentation/src/main/asciidoc/reference/search-dsl-sort.asciidoc
@@ -376,6 +376,11 @@ The behavior for missing values can be controlled explicitly through the `.missi
 
 * `.missing().first()` puts documents with no value in first position (regardless of the sort order).
 * `.missing().last()` puts documents with no value in last position (regardless of the sort order).
+* `.missing().lowest()` interprets missing values as the lowest value:
+it puts documents with no value in the first position when using ascending order or in the last position when using descending order.
+* `.missing().highest()` interprets missing values as the highest value:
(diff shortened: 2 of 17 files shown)


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/sort/comparatorsource/impl/LuceneTextFieldComparatorSource.java
```diff
@@ -32,7 +32,7 @@ public LuceneTextFieldComparatorSource(String nestedDocumentPath, Object missing
 
 	@Override
 	public FieldComparator<?> newComparator(String fieldname, int numHits, int sortPos, boolean reversed) {
-		final boolean sortMissingLast = missingLast() ^ reversed;
+		final boolean sortMissingLast = missingLast() ^ reversed && !missingLowest() || missingHighest();
```
yrodiere, line 35: I don't know about you, but I don't read this and immediately know what's the precedence of `^` vs `&&`... 

Anyway, this code was already unnecessarily complex before, maybe it's time for a change.

First, we should rename the variable to `considerMissingHighest`, because that's what it actually means; see how its value is interpreted in `org.apache.lucene.search.FieldComparator.TermOrdValComparator#TermOrdValComparator(int, java.lang.String, boolean)`.

Second, maybe these if/else blocks would be clearer?

```suggestion
		final boolean considerMissingHighest;
		if ( SortMissingValue.MISSING_LOWEST.equals( missingValue ) ) {
			considerMissingHighest = false;
		}
		else if ( SortMissingValue.MISSING_HIGHEST.equals( missingValue ) ) {
			considerMissingHighest = true;
		}
		else if ( SortMissingValue.MISSING_LAST.equals( missingValue ) ) {
			// To appear last, missing values must be considered highest, or lowest if the order is reversed.
			considerMissingHighest = !reversed;
		}
		else { // SortMissingValue.MISSING_FIRST, the default
			// To appear first, missing values must be considered lowest, or highest if the order is reversed.
			considerMissingHighest = reversed;
		}
```

documentation/src/main/asciidoc/reference/search-dsl-sort.asciidoc
```diff
@@ -376,6 +376,10 @@ The behavior for missing values can be controlled explicitly through the `.missi
 
 * `.missing().first()` puts documents with no value in first position (regardless of the sort order).
 * `.missing().last()` puts documents with no value in last position (regardless of the sort order).
+* `.missing().lowest()` depends on the sort order. In the case of ascending order, the documents with no value will go first.
+The documents with no value will go last if descending order is used.
+* `.missing().highest()` depends on the sort order. In the case of ascending order, the documents with no value will go last.
+The documents with no value will go first if descending order is used.
```
yrodiere, line 382: Maybe this would be more straightforward?

```suggestion
* `.missing().lowest()` interprets missing values as the lowest value: its puts documents with no value in first position when using ascending order, or in last position when using descending order.
* `.missing().highest()` interprets missing values as the highest value: its puts documents with no value in last position when using ascending order, or in first position when using descending order.
```

engine/src/main/java/org/hibernate/search/engine/search/sort/dsl/DistanceSortMissingValueBehaviorStep.java
```diff
@@ -37,6 +37,30 @@
 	 */
 	N first();
 
+	/**
+	 * Give documents with missing values the highest value when sorting.
+	 * <p>
+	 * This instruction depends on whether the sort order is being ascending or descending.
+	 * <p>
+	 * The documents with missing values will be at the top in case of the descending order.
+	 * Otherwise, in case of the ascending order - documents will be at the bottom.
```
yrodiere, line 46: Maybe this is more straightforward?

```suggestion
	 * This puts documents with missing values last when using ascending order,
	 * or first when using descending order.
```

engine/src/main/java/org/hibernate/search/engine/search/sort/dsl/DistanceSortMissingValueBehaviorStep.java
```diff
@@ -37,6 +37,30 @@
...
+	 */
+	N highest();
+
+	/**
+	 * Give documents with missing values the lowest value when sorting.
+	 * <p>
+	 * This instruction depends on whether the sort order is being ascending or descending.
+	 * <p>
+	 * The documents with missing values will be at the bottom in case of the descending order.
+	 * Otherwise, in case of the ascending order - documents will be at the top.
```
yrodiere, line 58: Maybe this is more straightforward?

```suggestion
	 * This puts documents with missing values first when using ascending order,
	 * or last when using descending order.
```

engine/src/main/java/org/hibernate/search/engine/search/sort/dsl/FieldSortMissingValueBehaviorStep.java
```diff
@@ -38,6 +38,30 @@
 	 */
 	N first();
 
+	/**
+	 * Give documents with missing values the highest value when sorting.
+	 * <p>
+	 * This instruction depends on whether the sort order is being ascending or descending.
+	 * <p>
+	 * The documents with missing values will be at the top in case of the descending order.
+	 * Otherwise, in case of the ascending order - documents will be at the bottom.
```
yrodiere, line 47: Maybe this is more straightforward?

```suggestion
	 * This puts documents with missing values last when using ascending order,
	 * or first when using descending order.
```

engine/src/main/java/org/hibernate/search/engine/search/sort/dsl/FieldSortMissingValueBehaviorStep.java
```diff
@@ -38,6 +38,30 @@
...
+	 */
+	N highest();
+
+	/**
+	 * Give documents with missing values the lowest value when sorting.
+	 * <p>
+	 * This instruction depends on whether the sort order is being ascending or descending.
+	 * <p>
+	 * The documents with missing values will be at the bottom in case of the descending order.
+	 * Otherwise, in case of the ascending order - documents will be at the top.
```
yrodiere, line 59: Maybe this is more straightforward?

```suggestion
	 * This puts documents with missing values first when using ascending order,
	 * or last when using descending order.
```

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/sort/SearchSortIT.java
```diff
@@ -397,6 +397,46 @@ public void toAbsolutePath_withRoot_null() {
...
+
+
+		assertThatQuery(
+				scope.query()
+						.where( f -> f.matchAll() )
+						.sort( s -> s.field( "string" ).desc().missing().lowest() )
+						.toQuery()
+		).hasDocRefHitsExactOrder( mainIndex.typeName(), THIRD_ID, SECOND_ID, FIRST_ID, EMPTY_ID );
+	}
+
```
yrodiere, line 439: I don't get why these tests are necessary... this is already tested elsewhere.

I don't know why other, pre-existing methods in this class rely on `.missing().last()`, but I doubt they're actually intended to test this feature. More likely these tests were introduced back when there was no consistent default behavior for missing values across backends.

yrodiere (commented):
Thanks! Comments below.

PR #3476, opened 2023-04-20
HSEARCH-4837 Conditionally advance to a next document within an id projection values

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4837

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/collector/impl/IdentifierValues.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/collector/impl/IdentifierValues.java
@@ -25,7 +25,13 @@ public final class IdentifierValues implements Values<String> {
 
 	@Override
 	public String get(int doc) throws IOException {
-		currentLeafIdDocValues.advance( doc );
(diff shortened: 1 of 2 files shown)


yrodiere (commented):
The solution makes sense, but if we do things that way we will have to adapt every other `Values` implementation. For example `org.hibernate.search.backend.lucene.lowlevel.collector.impl.DocumentReferenceValues#get` would have the same problem.

I'm not sure that's the right approach, though. Generally, most `Values#get` implementations assume they will only get called once per document, and I think it should stay that way as it's more in line with how Lucene works.

If we're going to remove the ability to use the identifier projection within object projections (https://hibernate.atlassian.net/browse/HSEARCH-4827), we may not want to fix this at all, since the problem won't manifest itself anymore. And if we're going to add a `.fromRoot()` projection like I suggested on  https://hibernate.atlassian.net/browse/HSEARCH-4827 , then we can make it the responsibility of `.fromRoot()` to only call `Values#get` once per (root) document, by caching the last retrieved value and last requested root document.

marko-bekhta:
closing as discussed in the chat https://hibernate.zulipchat.com/#narrow/stream/132092-hibernate-search-dev/topic/HSEARCH-4837/near/351286855

PR #3477, opened 2023-04-20
HSEARCH-4827 Don't allow some projections within a nested context

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4827

sooooo 😄  the idea here 

- use the same check `context.absoluteCurrentFieldPath() != null` as in highlights
- put this check in an abstract projection and provide a name/hint in impls
- let users know in the migration guide about the change

then a few things:
- not sure how to go about entity projection. Since it not necessarily reaches the projection, but might instead use a composite projection... 
https://github.com/hibernate/hibernate-search/blob/58c62aea92e5c511ecf131ba9a4ca8f19fe85821/engine/src/main/java/org/hibernate/search/engine/search/projection/dsl/impl/EntityProjectionOptionsStepImpl.java#L65-L75
- not sure about the need for `EventContext`. I've added one ti the exception, but more like a placeholder... Should we just create some new context method in `EventContexts`, or I've missed something, and we already have access to some meaningful context already?


documentation/src/main/asciidoc/migration/index.asciidoc
```diff
@@ -389,4 +389,9 @@ query that returns the same results.
...
 replaced with its clause.
 
-A bool query with a single `mustNot` clause and applied boost would implicitly add a `must` with `match_all` clause.
\ No newline at end of file
+A bool query with a single `mustNot` clause and applied boost would implicitly add a `must` with `match_all` clause.
+
+Any of `id`/`entity`/`entityReference`/`documentReference`/`score` as well as Lucene-specific `document`/`explanation`
+and Elasticsearch-specific `jsonHit`/`source`/`explanation` projections are now not allowed within a context of a nested projection (`object`).
+These projections are targeting the root object and placing them inside a nested projection might lead to confusing results.
+Using these within such `object` projection will now cause an exception.
```
yrodiere: > let users know in the migration guide about the change

This is not necessary as the `object` projection was introduced in 6.2.0.Alpha. We don't document breaking changes between alphas in the migration guide.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/AbstractNotNestableElasticsearchProjection.java
```diff
@@ -0,0 +1,56 @@
...
+			throw log.cannotUseProjectionInNestedContext(
+					projectionName(),
+					unsupportedNestingHint(),
+					EventContexts.indexSchemaRoot()
+			);
+		}
+		return doRequest( requestBody, context );
+	}
+
+	protected abstract String projectionName();
```
yrodiere, line 37: For consistency, and since it doesn't make sense to translate this anyway, could you rather use projection keys that you'd add to `org.hibernate.search.engine.search.projection.spi.ProjectionTypeKeys`? We already use such keys in some exception messages.

You can add backend-specific variants if necessary, like we did with e.g. `org.hibernate.search.backend.lucene.search.predicate.impl.LucenePredicateTypeKeys`.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/AbstractNotNestableElasticsearchProjection.java
```diff
@@ -0,0 +1,56 @@
...
+	protected final Set<String> indexNames;
+
+	protected AbstractNotNestableElasticsearchProjection(ElasticsearchSearchIndexScope<?> scope) {
+		indexNames = scope.hibernateSearchIndexNames();
+	}
+
+	@Override
+	public final Extractor<?, P> request(JsonObject requestBody, ProjectionRequestContext context) {
+		if ( context.absoluteCurrentFieldPath() != null ) {
+			throw log.cannotUseProjectionInNestedContext(
```
yrodiere, line 28: FWIW you don't need a whole abstract class just for that, you could just add a method `checkNotNested(SearchQueryElementTypeKey projectionKey, String hint)` to `ProjectionRequestContext`, and call it from the relevant projections.

We already have something similar in `org.hibernate.search.backend.elasticsearch.search.projection.impl.ProjectionRequestContext#checkValidField`.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/AbstractNotNestableElasticsearchProjection.java
```diff
@@ -0,0 +1,56 @@
...
+					projectionName(),
+					unsupportedNestingHint(),
+					EventContexts.indexSchemaRoot()
```
yrodiere, line 31: `indexSchemaRoot` definitely isn't correct since this exception happens only when we're *not* at the schema root. I'd remove the context (you don't *have* to provide one) or change it to `EventContexts.fromIndexFieldAbsolutePath(...)`.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/ElasticsearchSearchHints.java
```diff
@@ -0,0 +1,57 @@
...
+
+import org.hibernate.search.engine.backend.reporting.spi.BackendSearchHints;
+import org.hibernate.search.util.common.logging.impl.MessageConstants;
+
+import org.jboss.logging.Messages;
+import org.jboss.logging.annotations.Message;
+import org.jboss.logging.annotations.MessageBundle;
+
+@MessageBundle(projectCode = MessageConstants.PROJECT_CODE)
+public interface ElasticsearchSearchHints extends BackendSearchHints {
```
yrodiere, line 17: Nice idea, TBH I wouldn't have bothered to explain *why* it's forbidden, I'd just have told users they can't do it.

BTW there's nothing backend-specific about most hints below... you could just move the backend-independent `@Message` annotations to `BackendSearchHints`?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -734,4 +734,8 @@ SearchException invalidSingleValuedProjectionOnValueFieldInMultiValuedObjectFiel
 
 	@Message(id = ID_OFFSET + 157, value = "Unable to export the schema for '%1$s' index: %2$s" )
 	SearchException unableToExportSchema(String indexName, String message, @Cause IOException e);
+
+	@Message(id = ID_OFFSET + 158, value = "'%1$s' projections are only applicable within the root projection context. "
```
yrodiere, line 738: I'm not sure "root projection context" is meaningful for all users... Maybe rather:

```suggestion
	@Message(id = ID_OFFSET + 158, value = "'%1$s' projections cannot be nested in an object projection. "
```

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/projection/UnsupportedNestingProjectionBaseIT.java
```diff
@@ -0,0 +1,158 @@
...
+		).isInstanceOf( SearchException.class )
+				.hasMessageContainingAll(
+						"'ID' projections are only applicable within the root projection context",
+						"An ID projection represents the document ID and adding it as a part of the nested object projection might produce misleading results since it is always a root document ID and not a nested object ID."
+
+				);
+	}
+
+	@Test
+	@Ignore("entity projection might not load an entity but do a composite projection from the root context instead...")
```
yrodiere, line 58: > not sure how to go about entity projection. Since it not necessarily reaches the projection, but might instead use a composite projection...

Maybe it would be a good idea to introduce an "EntityCompositeProjection" to model the fact that such a composite projection is supposed to return the root entity?

Basically it would just be a wrapper projection that would delegate to the composite projection, with an additional check in `request` to prevent it from being used in nested contexts... We wouldn't expose it in APIs, just in a new `SearchProjection<T> entityComposite(SearchProjection<T> delegate)`  method in `org.hibernate.search.engine.search.projection.spi.SearchProjectionBuilderFactory`. We'd call it from `org.hibernate.search.engine.search.projection.dsl.impl.EntityProjectionOptionsStepImpl#toProjection(org.hibernate.search.engine.search.projection.spi.ProjectionMappedTypeContext)` when building a non-loading entity projection.

WDYT?

By the way, if you do this, you can probably remove `org.hibernate.search.engine.search.projection.spi.SearchProjectionBuilderFactory#rootContext` and all related code, as it won't be useful anymore. Maybe in a separate commit so it can be restored later if we introduce `fromRoot`.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/projection/IdentifierProjectionBaseIT.java
```diff
@@ -196,10 +199,17 @@ private void initData() {
...
 
 		IndexBinding(IndexSchemaElement root) {
 			// this field is irrelevant for the test
 			name = root.field( "name", f -> f.asString() ).toReference();
+			// these fields are also irrelevant for the test, we only need them so that there's a nested context we can
+			// create a projection in:
+			IndexSchemaObjectField nested = root.objectField( "nested", ObjectStructure.NESTED ).multiValued();
+			this.nestedString = nested.field( "string", f -> f.asString() ).toReference();
+			this.nested = nested.toReference();
 		}
```
yrodiere, line 213: Am I missing something or are these changes unnecessary?

yrodiere (commented):
Thanks, below a few comments (and answer to your questions, hopefully)

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -742,4 +742,8 @@ SearchException invalidSingleValuedProjectionOnValueFieldInMultiValuedObjectFiel
 	@Message(id = ID_OFFSET + 159, value = "Invalid use of 'missing().lowest()' for a descending distance sort. " +
 			"Elasticsearch always assumes missing values have a distance of '+Infinity', and this behavior cannot be customized. ")
 	SearchException missingLowestOnDescSortNotSupported(@Param EventContext context);
+
+	@Message(id = ID_OFFSET + 160, value = "'%1$s' projections cannot be nested in an object projection. "
```
yrodiere, line 746: The word "projection" already appears in the projection key.

```suggestion
	@Message(id = ID_OFFSET + 160, value = "'%1$s' cannot be nested in an object projection. "
```

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/FieldProjectionRequestContext.java
```diff
@@ -57,6 +59,17 @@ public void checkValidField(String absoluteFieldPath) {
 		}
 	}
 
+	@Override
+	public void checkNotNested(SearchQueryElementTypeKey<?> projectionKey, String hint) {
+		if ( absoluteCurrentFieldPath() != null ) {
+			throw log.cannotUseProjectionInNestedContext(
+					projectionKey.toString(),
+					hint,
+					EventContexts.indexSchemaRoot()
```
yrodiere, line 68: As said in my earlier review:

> `indexSchemaRoot` definitely isn't correct since this exception happens only when we're not at the schema root. I'd remove the context (you don't have to provide one) or change it to EventContexts.fromIndexFieldAbsolutePath(...).

IMO `EventContexts.fromIndexFieldAbsolutePath(absoluteCurrentFieldPath())` would be fine here.

engine/src/main/java/org/hibernate/search/engine/search/projection/spi/ProjectionTypeKeys.java
```diff
@@ -20,5 +20,10 @@ public static <T> SearchQueryElementTypeKey<T> key(String name) {
 	public static final SearchQueryElementTypeKey<FieldProjectionBuilder.TypeSelector> FIELD = key( "field" );
 	public static final SearchQueryElementTypeKey<DistanceToFieldProjectionBuilder> DISTANCE = key( "distance" );
 	public static final SearchQueryElementTypeKey<CompositeProjectionBuilder> OBJECT = key( "object" );
-
+	public static final SearchQueryElementTypeKey<CompositeProjectionBuilder> ID = key( "id" );
+	public static final SearchQueryElementTypeKey<CompositeProjectionBuilder> DOCUMENT_REFERENCE = key( "document reference" );
+	public static final SearchQueryElementTypeKey<CompositeProjectionBuilder> ENTITY = key( "entity" );
+	public static final SearchQueryElementTypeKey<CompositeProjectionBuilder> ENTITY_REFERENCE = key( "entity reference" );
```
yrodiere, line 26: I'd avoid spaces in keys, and since you don't use the builder you might as well hide its type.

```suggestion
	public static final SearchQueryElementTypeKey<?> ID = key( "id" );
	public static final SearchQueryElementTypeKey<?> DOCUMENT_REFERENCE = key( "document-reference" );
	public static final SearchQueryElementTypeKey<?> ENTITY = key( "entity" );
	public static final SearchQueryElementTypeKey<?> ENTITY_REFERENCE = key( "entity-reference" );
```

engine/src/main/java/org/hibernate/search/engine/search/projection/spi/ProjectionTypeKeys.java
```diff
@@ -20,5 +20,10 @@ public static <T> SearchQueryElementTypeKey<T> key(String name) {
...
+	public static final SearchQueryElementTypeKey<CompositeProjectionBuilder> EXPLANATION = key( "explanation" );
```
yrodiere, line 27: I know both backends provide this projection, but technically it's still a an extension in both cases, since they behave differently (return type is wildly different). Could you move this to backend-specific classes?

engine/src/main/java/org/hibernate/search/engine/search/projection/spi/ProjectionTypeKeys.java
```diff
@@ -20,5 +20,10 @@ public static <T> SearchQueryElementTypeKey<T> key(String name) {
...
+	public static final SearchQueryElementTypeKey<CompositeProjectionBuilder> SCORE = key( "score" );
```
yrodiere, line 28:
```suggestion
	public static final SearchQueryElementTypeKey<?> SCORE = key( "score" );
```

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchProjectionTypeKeys.java
```diff
@@ -0,0 +1,22 @@
...
+import org.hibernate.search.engine.search.common.spi.SearchQueryElementTypeKey;
+import org.hibernate.search.engine.search.projection.spi.CompositeProjectionBuilder;
+
+public final class ElasticsearchProjectionTypeKeys {
+
+	private ElasticsearchProjectionTypeKeys() {
+	}
+
+	public static final SearchQueryElementTypeKey<CompositeProjectionBuilder> JSON_HIT = key( "JSON hit" );
+	public static final SearchQueryElementTypeKey<CompositeProjectionBuilder> SOURCE = key( "source" );
```
yrodiere, line 20: Since it's a key, it'd stick to a simpler character set, say lowercase alphanumerics + hyphens like we do elsewhere?

Also since you don't use the builder, you might as well hide its type (`CompositeProjectionBuilder` is incorrect anyway):

```suggestion
	public static final SearchQueryElementTypeKey<?> JSON_HIT = key( "json-hit" );
	public static final SearchQueryElementTypeKey<?> SOURCE = key( "source" );
```

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/projection/impl/LuceneProjectionTypeKeys.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.hibernate.search.engine.search.common.spi.SearchQueryElementTypeKey;
+import org.hibernate.search.engine.search.projection.spi.CompositeProjectionBuilder;
+
+public final class LuceneProjectionTypeKeys {
+
+	private LuceneProjectionTypeKeys() {
+	}
+
+	public static final SearchQueryElementTypeKey<CompositeProjectionBuilder> DOCUMENT = key( "document" );
```
yrodiere, line 19: Since you don't use the builder, you might as well hide its type (`CompositeProjectionBuilder` is incorrect anyway):

```suggestion
	public static final SearchQueryElementTypeKey<?> DOCUMENT = key( "document" );
```

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/projection/impl/LuceneProjectionTypeKeys.java
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.lucene.search.projection.impl;
+
+import static org.hibernate.search.engine.search.projection.spi.ProjectionTypeKeys.key;
+
+import org.hibernate.search.engine.search.common.spi.SearchQueryElementTypeKey;
+import org.hibernate.search.engine.search.projection.spi.CompositeProjectionBuilder;
```
yrodiere, line 12:
```suggestion
```

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/ElasticsearchProjectionTypeKeys.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.elasticsearch.search.projection.impl;
+
+import static org.hibernate.search.engine.search.projection.spi.ProjectionTypeKeys.key;
+
+import org.hibernate.search.engine.search.common.spi.SearchQueryElementTypeKey;
+import org.hibernate.search.engine.search.projection.spi.CompositeProjectionBuilder;
```
yrodiere, line 12:
```suggestion
```

yrodiere (commented):
Thanks; just a few comments, feel free to merge once they're addressed.

yrodiere (approved):
LGTM, but this seems to conflict with your PRs that I just merged :) Would you please rebase?

yrodiere:
Merged, thanks!

PR #3485, opened 2023-04-25
HSEARCH-4839 [DO NOT MERGE] Test Hibernate Search regularly against the latest version of Lucene

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4839

This is more to see what changes we need for the upgrade. I'll push it to the upstream repo once you take a look at it 😃 
The summary of changes:

Should we address these in the current version?
- adding the hppc lib explicitly to the dependency list (since we have the Lucene backend rely on it, should we list it as an explicit dependency now for the 8.11?)
- there are some schema checks during the writes - when a nested object is written, its ID doesn't match the schema as we were doing this - `document.add( MetadataFields.searchableMetadataField( MetadataFields.idFieldName(), rootId ) );`. But the ID of a root document has doc-values set to `binary`. Doing `document.add( MetadataFields.searchableRetrievableMetadataField( MetadataFields.idFieldName(), rootId ) );` helps with that.

Other changes that are more for the Lucenes 9.5 version:
- there's a change in comparators. Now they have a bool flag `boolean enableSkipping` instead of `int sortPos`. Looks like this flag needs to be forced to `false` for text and geopoint fields. (enableSkipping – true if the comparator can skip documents via LeafFieldComparator.competitiveIterator()). In 9.5, skipping is determined by `i == 0` where `i` is an index from `SortField[] fields`.
- `FileSystemAccessStrategy.SIMPLE` must now be removed as it's no longer in Lucene
- some package changes in imports
- `lucene-analyzers-common` is now `lucene-analysis-common`
- `currentLeafReader.document( doc, visitor )` is now deprecated, but there's an alternative.
- `SortedSetDocValues.NO_MORE_ORDS` is now deprecated in favour of using `values.docValueCount()`. It is OK for the places where we iterate over things. Not so OK in `TextMultiValues`. Looking inside Lucene - there's still some usage of this constant in internals. In some other places it got just replaced with `-1`. 
- there's no more `LongIntScatterMap` in newer version of hppc ... replaced with `LongIntHashMap`
- `Facets` now has new `getAllChildren(..)` method - in our impl it is delegating to `getTopChildren(..)`
- `ConjunctionDISI` is an internal class - now we can replace it with `ConjunctionUtils`
- `Query` now has a `visit(QueryVisitor visitor)` visitor method, but it doesn't look like we have to bother about it.
- `NormsFieldExistsQuery` and `DocValuesFieldExistsQuery` are deprecated now and are simply extending the `FieldExistsQuery`


backend/lucene/pom.xml
```diff
@@ -44,6 +44,10 @@
             <groupId>org.apache.lucene</groupId>
             <artifactId>lucene-facet</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.carrotsearch</groupId>
+            <artifactId>hppc</artifactId>
+        </dependency>
```
yrodiere, line 50: Please don't use dependency management, see my main comment above.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/document/impl/LuceneNestedObjectFieldBuilder.java
```diff
@@ -30,7 +30,9 @@ void contribute(MultiTenancyStrategy multiTenancyStrategy, String tenantId, Stri
 		// We own the document content, so we finalize it ourselves.
 		Document document = documentContent.finalizeDocument( multiTenancyStrategy, tenantId, routingKey );
 		document.add( MetadataFields.searchableMetadataField( MetadataFields.typeFieldName(), MetadataFields.TYPE_CHILD_DOCUMENT ) );
-		document.add( MetadataFields.searchableMetadataField( MetadataFields.idFieldName(), rootId ) );
+		// IMPORTANT: Since there is some schema validation now in Lucene the id field of a nested object must match the configuration
+		// of a root document id field. Hence, `searchableRetrievableMetadataField` is used so that the doc_values are set to BINARY as for a root id.
+		document.add( MetadataFields.searchableRetrievableMetadataField( MetadataFields.idFieldName(), rootId ) );
```
yrodiere, line 35: > there are some schema checks during the writes - when a nested object is written, its ID doesn't match the schema as we were doing this - document.add( MetadataFields.searchableMetadataField( MetadataFields.idFieldName(), rootId ) );. But the ID of a root document has doc-values set to binary. Doing document.add( MetadataFields.searchableRetrievableMetadataField( MetadataFields.idFieldName(), rootId ) ); helps with that.

You mean Lucene 9 has more checks and fails because of this discrepancy?

The thing is, we really don't need docvalues for this field in nested documents, and I wonder what the impact of adding it would be...

Also, I suspect changing this may not fix the problem existing indexes, depending how the thing gets validated. Is it validated globally when Lucene opens the index or only locally for newly indexed documents?

I wonder if we shouldn't just use a different field name for the ID docvalues... We may be able to introduce that without breaking backwards compatibility (try to retrieve the ID from `__HSEARCH_id_docvalue` first, then fall back to `__HSEARCH_id`). Testing this would be annoying, but doable :)

Maybe it would be a good idea to have a look how Elasticsearch does it. If they just enable docvalues for IDs in nested fields, then so be it.

Can you please create a ticket for this in 6.2? Better address this soon.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/directory/FileSystemAccessStrategyName.java
```diff
@@ -15,11 +15,6 @@
 public enum FileSystemAccessStrategyName {
 
 	AUTO( "auto" ),
-	/**
-	 * @deprecated This strategy will be removed in a future version of Lucene.
-	 */
-	@Deprecated
-	SIMPLE( "simple" ),
```
yrodiere, line 22: Can you please also remove the corresponding entry from the reference documentation, if any?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/facet/impl/LongMultiValueFacetCounts.java
```diff
@@ -76,6 +77,11 @@ private void increment(long value) {
 		}
 	}
 
+	@Override
+	public FacetResult getAllChildren(String dim, String... path) throws IOException {
+		return getTopChildren( Integer.MAX_VALUE, dim, path );
+	}
```
yrodiere, line 83: Maybe just throw an exception? I don't think we would use this method anyway. Same for other implementations.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/query/impl/MappedTypeNameQuery.java
```diff
@@ -72,4 +73,9 @@ public boolean isCacheable(LeafReaderContext ctx) {
 			}
 		};
 	}
+
+	@Override
+	public void visit(QueryVisitor visitor) {
+		// do nothing
```
yrodiere, line 79:
```suggestion
		visitor.visitLeaf( this );
```

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/query/impl/ExplicitDocIdsQuery.java
```diff
@@ -76,4 +77,9 @@ public boolean isCacheable(LeafReaderContext ctx) {
 		};
 	}
 
+	@Override
+	public void visit(QueryVisitor visitor) {
+		// do nothing
```
yrodiere, line 82:
```suggestion
		visitor.visitLeaf( this );
```

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/predicate/impl/LuceneExistsPredicate.java
```diff
@@ -58,7 +57,7 @@ private NormsBasedBuilder(LuceneSearchIndexScope<?> scope, LuceneSearchIndexValu
 
 		@Override
 		protected Query buildQuery() {
-			return new NormsFieldExistsQuery( absoluteFieldPath );
+			return new FieldExistsQuery( absoluteFieldPath );
```
yrodiere, line 60: If this `FieldExistsQuery` always works regardless of the field type, maybe you can merge all the different builders/factories in this class?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/sort/comparatorsource/impl/LuceneTextFieldComparatorSource.java
```diff
@@ -50,7 +51,8 @@ else if ( SortMissingValue.MISSING_LAST.equals( missingValue ) ) {
 		TextMultiValuesToSingleValuesSource source =
 				TextMultiValuesToSingleValuesSource.fromField( fieldname, multiValueMode, nestedDocsProvider );
 
-		return new FieldComparator.TermOrdValComparator( numHits, fieldname, considerMissingHighest ) {
+		// forcing to not skipping documents
+		return new TermOrdValComparator( numHits, fieldname, considerMissingHighest, reversed, false ) {
```
yrodiere, line 55: I'm confused, why do we need this? Do we get a wrong value in `enableSkipping`? If so, shouldn't we rather investigate why we're getting a wrong value?

util/internal/integrationtest/backend/lucene/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/lucene/query/SlowQuery.java
```diff
@@ -68,6 +69,11 @@ public boolean isCacheable(LeafReaderContext ctx) {
 		};
 	}
 
+	@Override
+	public void visit(QueryVisitor visitor) {
+		// do nothing
```
yrodiere, line 74:
```suggestion
		visitor.visitLeaf( this );
```

yrodiere (commented):
Thanks!

Here are a few answers.

> adding the hppc lib explicitly to the dependency list (since we have the Lucene backend rely on it, should we list it as an explicit dependency now for the 8.11?)

If we do use it ourselves in the Lucene backend, then sure. But please don't use dependency management, use the same hacks than for Hibernate ORM dependencies: we want to align on whatever version Hibernate ORM/Lucene depends on, and get maven-enforcer-plugin to fail the build whenever we diverge. See the dependency to hibernate-commons-annotations for an example.

> there's a change in comparators. Now they have a bool flag boolean enableSkipping instead of int sortPos. Looks like this flag needs to be forced to false for text and geopoint fields. (enableSkipping – true if the comparator can skip documents via LeafFieldComparator.competitiveIterator()). In 9.5, skipping is determined by i == 0 where i is an index from SortField[] fields.

Ok... Is that a problem?

> FileSystemAccessStrategy.SIMPLE must now be removed as it's no longer in Lucene

Seems fine, there was a deprecation warning when people used it anyway.

> some package changes in imports

> lucene-analyzers-common is now lucene-analysis-common

Not a problem, we explicitly document that Lucene classes may change without warning and that we don't promise anything on that front.

> currentLeafReader.document( doc, visitor ) is now deprecated, but there's an alternative.

> SortedSetDocValues.NO_MORE_ORDS is now deprecated in favour of using values.docValueCount(). It is OK for the places where we iterate over things. Not so OK in TextMultiValues. Looking inside Lucene - there's still some usage of this constant in internals. In some other places it got just replaced with -1.

This is surprising, especially since there are probably places where we can't return a reliable `docValueCount()`. Maybe the method is allowed to return an upper bound?

Anyway, that's probably worth a dedicated ticket targeting Search 7.0.

> there's no more LongIntScatterMap in newer version of hppc ... replaced with LongIntHashMap
> Facets now has new getAllChildren(..) method - in our impl it is delegating to getTopChildren(..)
> ConjunctionDISI is an internal class - now we can replace it with ConjunctionUtils
> Query now has a visit(QueryVisitor visitor) visitor method, but it doesn't look like we have to bother about it.
> NormsFieldExistsQuery and DocValuesFieldExistsQuery are deprecated now and are simply extending the FieldExistsQuery

Seems fine?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/document/impl/LuceneNestedObjectFieldBuilder.java
marko-bekhta: > You mean Lucene 9 has more checks and fails because of this discrepancy?

Yes, that is what it looks like. It happens when a document is indexed. 

I'll take a look into Elasticsearch and then will create a ticket 👍🏻 😃

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/facet/impl/LongMultiValueFacetCounts.java
marko-bekhta: hah 😄 that was the second option I was considering 😄 will do

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/predicate/impl/LuceneExistsPredicate.java
marko-bekhta: Yup, this makes sense - Lucene now have a huge if-else-switch covering these scenarios 🙈 will merge

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/sort/comparatorsource/impl/LuceneTextFieldComparatorSource.java
marko-bekhta: `enableSkipping` originally comes from the `FieldValueHitQueue`:
```java
for (int i = 0; i < numComparators; ++i) {
      SortField field = fields[i];
     // ... 
     // enableSkipping = i == 0 
      comparators[i] = field.getComparator(size, i == 0);
    }
```
So there's not much we can do about this, unless we pass a dummy field into sorting, which I don't think we want to do...
Then the value can be set to false afterwards by calling one of `SortField#setOptimizeSortWithIndexedData()` `SortField#setOptimizeSortWithPoints()` but both are deprecated for removal in Lucene 10.

Text sort works ok for as long as we don't do a multi-index search where one of the indexes is missing a field... 
I guess we could check if it's a multiindex search in the `AbstractLuceneDocumentValueSort` constructor and if so set `sortField.setOptimizeSortWithIndexedData( false );` but that method is already deprecated 🙈 

I've also looked a bit closer at Geopoint sorting... numeric comparators are more or less the same between two versions. Problem is that we are getting a `new IllegalArgumentException( "Field " + field + " is indexed with multiple dimensions, sorting is not supported");` 
From what it looks in 8.11:
```java
  // For numeric sort fields, if true, indicates the DocValues and Point fields
  // with the same name have the exactly same data indexed.
  // This allows to use them to optimize sort and skip non-competitive documents.
  private boolean canUsePoints = false;
```
This flag that enables optimization is off. And it is never flipped (we never call `SortField#setCanUsePoints()`).
But in 9.5 similar flag:
```java
  // Indicates if sort should be optimized with indexed data. Set to true by default.
  @Deprecated private boolean optimizeSortWithIndexedData = true;
```
is enabled by default. Which leads to:
```java
if (info.getPointDimensionCount() > 1) {
          throw new IllegalArgumentException(
              "Field " + field + " is indexed with multiple dimensions, sorting is not supported");
        }
```

And since you've mentioned Elasticsearch in the other place -- I've looked at what they are doing for their comparators and ... ```return new LongComparator(numHits, null, null, reversed, enableSkiping: false) {``` 🥲

yrodiere: Ok, ok, I'll trust you. I guess this boolean is just a hint. Thanks for looking into this.

Someday, we might want to investigate whether we can make skipping actually work in at least some cases, but that will probably require support for index-time sorting (where documents are stored in the index already sorted), so it's definitely low priority.

backend/lucene/pom.xml
marko-bekhta: hmmmmmmmm. So with Lucene 9.5 this becomes a `<scope>runtime</scope>` dependency in the `lucene-facet` and enforcer plugin doesn't catch the version mismatch...

yrodiere: What if we use the runtime scope ourselves? It would be fine as we definitely won't expose this dependency through our API.

marko-bekhta: Mmm, I guess we cannot since we use types from it in our own facets - the compilation of those would fail, right?

yrodiere: Right. Well, I don't know. Good luck :D

yrodiere:
I pushed your branch to https://github.com/hibernate/hibernate-search/tree/wip/main/dependency-update/lucene9 , and merged your other PR that adds regular testing.

Thanks for working on this, and feel free to open a PR to https://github.com/hibernate/hibernate-search/tree/wip/main/dependency-update/lucene9 if you need more changes to fix the tests.

Closing.

PR #3487, opened 2023-04-26
HSEARCH-4843 Introduce a common EntityReference interface and deprecate mapper-specific ones

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4843

This turned out to be quite a big cleanup... @marko-bekhta feel free to give the internal parts only a cursory look.

 The most important thing here is to make sure I didn't break anything, and hopefully tests will detect that. Maybe you could check my changes to the tests make sense?

I checked that no backwards-incompatible API change occurred, but I'd appreciate a second opinion there, too.

Thanks!


yrodiere:
Ah, and if you need an explanation of what's going on, the commit messages should help :)

marko-bekhta:
It's just 246 changed files 😃😆 give me a second 😄

yrodiere:
I just pushed additional commits to address some build failures. I'll squash them when you're done reviewing :)

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/work/SearchIndexingPlanExecutionReport.java
```diff
@@ -21,6 +23,6 @@ public interface SearchIndexingPlanExecutionReport {
 	 * @return A list of references to entities that may not be indexed correctly as a result of the failure.
 	 * Never {@code null}, but may be empty.
 	 */
-	List<?> failingEntities();
+	List<? extends EntityReference> failingEntities();
```
marko-bekhta, line 26: This changes the API, but it's a new interface anyway. The only question I wonder is if we should mark it as `@Incubating`?

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/spi/PojoMassIndexingMappingContext.java
```diff
@@ -6,21 +6,16 @@
...
 
 /**
  * Contextual information about a search mapping.
  */
 public interface PojoMassIndexingMappingContext extends BackendMappingContext {
 
-	/**
-	 * @return A {@link EntityReferenceFactory} that relies on the object's class to return entity types.
-	 */
-	EntityReferenceFactory<?> entityReferenceFactory();
```
marko-bekhta, line 22: Since this is implemented by mappers, should we convert this to a default method throwing an exception and deprecate it? (assuming someone created their own mapper 😄) Or because we are the callers of this method -- we can just drop it.

marko-bekhta (commented):
Looks nice!
I've added a few inline questions/comments 😃 
BTW should we maybe also put an `@incubating` annotation on the new `EntityReference`/`SimpleEntityReference` just to be safe, even though it probably wouldn't need many changes 😃

util/internal/integrationtest/mapper/stub/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/stub/StubEntityReference.java
```diff
@@ -36,18 +42,18 @@ public boolean equals(Object obj) {
 		if ( !( obj instanceof EntityReference ) ) {
 			return false;
 		}
-		StubEntityReference other = (StubEntityReference) obj;
-		return typeName.equals( other.typeName ) && Objects.equals( id, other.id );
+		EntityReference other = (EntityReference) obj;
+		return type.equals( other.type() ) && name.equals( other.name() ) && Objects.equals( id, other.id() );
```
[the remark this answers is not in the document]
yrodiere: Right, I suppose we shouldn't. Will change that, thanks.

documentation/src/main/asciidoc/reference/configuration.asciidoc
```diff
@@ -425,7 +425,9 @@ include::{sourcedir}/org/hibernate/search/documentation/reporting/failurehandler
 <5> `handle(EntityIndexingFailureContext)` is called for failures occurring when indexing entities.
 <6> On top of the failing operation and throwable,
 the context also lists references to entities that could not be indexed correctly because of the failure.
-<7> Use the context-provided information to report the failure in any relevant way.
+<7> Entity references expose the entity type, the entity name and entity identifier.
```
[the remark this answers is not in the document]
yrodiere: Will do, thanks

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/work/SearchIndexingPlanExecutionReport.java
yrodiere: Technically this is a backwards-compatible change :) That's why I insisted on using wildcards here.

True, the `@Incubating`  annotation was missing, I'll add it.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/spi/PojoMassIndexingMappingContext.java
yrodiere: This method was just moved to the superinterface `BackendMappingContext`, it's not a breaking change.

That being said there are probably plenty of breaking changes in mapper SPIs. I gave up on ensuring backwards compatibility and just documented the change (since we're allowed to break SPIs in minors).

marko-bekhta (approved):

PR #3490, opened 2023-05-02
HSEARCH-2192 Follow-up: Remove the `max analyzed offset` option

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-2192

follows up on https://github.com/hibernate/hibernate-search/pull/3418

I've updated the log IDs since this wasn't released yet.

`checkApplicability()` is gone, as it was only needed to be able to check/log message for the unsupported case for this option.

`LimitTokenOffsetAnalyzer` is gone, as it was only to support this option.


yrodiere:
Test failures are unrelated, merging. Thanks!

PR #3496, opened 2023-05-05
HSEARCH-4828 Lucene's unified highlighter should produce multiple highlighted snippets for multi-valued fields

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4828


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/highlighter/impl/LuceneUnifiedSearchHighlighter.java
```diff
@@ -139,37 +143,110 @@ else if ( BoundaryScannerType.SENTENCE.equals( LuceneUnifiedSearchHighlighter.th
...
 		}
 	}
 
 	// `DefaultPassageFormatter` uses a string builder to create a resulting highlighted string.
-	// Elasticsearch uses their own formatter that creates an array of snippets instead.
-	// Hence, the results won't match if more than one snipped is needed to provide the results.
-	private static class PassageFormatterWithEncoder extends DefaultPassageFormatter {
+	// This is a copy of a `DefaultPassageFormatter` that instead of doing string concatenation is putting each passage
+	// as a new element:
+	private static class PassageFormatterWithEncoder extends PassageFormatter {
```
yrodiere, line 179: Looks like [coverage for this class is not perfect](https://sonarcloud.io/code?id=org.hibernate.search%3Ahibernate-search-parent&pullRequest=3496&selected=org.hibernate.search%3Ahibernate-search-parent%3Abackend%2Flucene%2Fsrc%2Fmain%2Fjava%2Forg%2Fhibernate%2Fsearch%2Fbackend%2Flucene%2Fsearch%2Fhighlighter%2Fimpl%2FLuceneUnifiedSearchHighlighter.java&line=179)... Unless I'm wrong it's a rather critical piece of code, so do you think you could add a few unit tests? Or alternatively an integration tests that involves overlapping terms, whatever you prefer.

yrodiere (commented):
Thanks! A few comments below.

documentation/src/main/asciidoc/reference/search-dsl-highlighting.asciidoc
```diff
@@ -304,8 +304,9 @@ include::{sourcedir}/org/hibernate/search/documentation/search/highlighting/High
 
 [NOTE]
 ====
-These options are supported by all highlighter types on the Elasticsearch backend,
-but only by plain and fast-vector highlighters with the Lucene backend.
+These options are supported by all highlighter types on the Elasticsearch backend.
+As for the Lucene backend -- the number of fragments is also supported by all highlighter types,
+while only plain and fast-vector highlighters support fragment size.
```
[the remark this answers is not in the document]
marko-bekhta: I have been thinking about doing so. Sadly those `BreakIterator`s won't help. We may try to use them for some Lucene-specific highlighter options. As for this fragment size option, we'd need something like this - https://github.com/elastic/elasticsearch/blob/main/server/src/main/java/org/elasticsearch/lucene/search/uhighlight/BoundedBreakIteratorScanner.java#L36 . And since that iterator is a bit lengthy I've kept it out 😃🙈

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/highlighter/impl/LuceneUnifiedSearchHighlighter.java
marko-bekhta: yeah 🥲😃  will try with IT first to keep this class private.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/highlighter/impl/LuceneUnifiedSearchHighlighter.java
```diff
@@ -139,37 +143,110 @@ else if ( BoundaryScannerType.SENTENCE.equals( LuceneUnifiedSearchHighlighter.th
...
-			for ( String highlight : highlights.get( fieldsIn[0] ) ) {
-				if ( highlight != null ) {
-					result.add( highlight );
+			List<TextFragment> highlights = highlighter.highlightField( fieldsIn, query, leafReaderContext.docBase + doc, maxPassagesIn );
+			if ( highlights == null ) {
+				return Collections.emptyList();
+			}
+			else {
+				if ( Boolean.TRUE.equals( LuceneUnifiedSearchHighlighter.this.orderByScore ) ) {
+					highlights.sort( Comparator.comparingDouble( TextFragment::score ).reversed() );
```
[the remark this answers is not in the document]
marko-bekhta: I've also created a `CustomScoreOrderFragmentsBuilder` for a FVH since the original one creates a comparator all the time

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/highlighter/impl/LuceneUnifiedSearchHighlighter.java
marko-bekhta: Hmmm, so I've tried to come up with a case where there'll be overlaps. From what I can see, it probably won't happen since a unified highlighter cannot "highlight phrases"... So now I'm not sure if we really need to keep that loop ... I've also looked at the Lucene test for this passage formatter and there's no case for overlapping 🙈. I suppose we can create a test for a formatter and build passages ourselves so that they have overlaps, but I'm not sure if we should if that's not going to happen... 
I've added a line to the docs mentioning that phrase highlighting is only possible with FVH.

yrodiere: Are you sure it's not possible to get overlaps with non-phrase highlighting?

I know there are analyzers that will turn `wi-fi` into three terms: `wi`, `fi` and `wifi`. With such an analyzer, a query such as `OR('wi', 'wifi')` would potentially match both the whole original term (`wi-fi`) and part of it (`wi`), so there would be overlap... ?

Anyway, I'd indeed recommend just a unit test, since it looks rather complicated to set up in an integration test.

PR #3506, opened 2023-05-16
HSEARCH-4839 Follow-up: Make sure to rebuild everything when running a Lucene upgrade job

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4839

Just having the Lucene backend rebuilt is not enough since there are also changes in some test utils as well as changes in the batch module caused by the driver upgrade... It looks like it would be just safer to rebuild the entire project like we do for the orm6-in-main-code


marko-bekhta:
Merging this one to see if it'll help with the dependency upgrade build over the weekend

PR #3518, opened 2023-05-31
DRAFT: HSEARCH-1182  experiments

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-1182

This is just to get some initial feedback. `ExcludePathsIT` is the test I've been experimenting with. 
The general idea is that a user can apply either include or exclude paths. The filtering part happens in `IndexSchemaFilter#isPathIncludedInternal` by just adding a check if the path wasn't explicitly excluded. Everything else in here is just "noise" to get the info from an annotation to the filter 😃. Is this going in the right direction?

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
+++ b/engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
@@ -119,8 +119,9 @@ class IndexSchemaFilter {
 	 */
 	private boolean isPathIncludedInternal(int relativeDepth, String relativePath,
 			boolean markAsEncountered, boolean includedByChild) {
-		boolean includedByThis = depthFilter.isEveryPathIncludedAtDepth( relativeDepth )
-				|| pathFilter.isExplicitlyIncluded( relativePath );
+		boolean includedByThis = ( depthFilter.isEveryPathIncludedAtDepth( relativeDepth )
+				|| pathFilter.isExplicitlyIncluded( relativePath ) )
+				&& !pathFilter.isExplicitlyExcluded( relativePath );
 
 		boolean includedByParent = true;
 		/*
@@ -128,7 +129,7 @@ class IndexSchemaFilter {
 		 * by reducing the includeDepth in particular,
 		 * but it cannot include paths that are filtered out by a child.
 		 */
-		if ( parent != null ) {
+		if ( includedByThis && parent != null ) {
 			includedByParent = parent.isPathIncludedInternal(
 					relativeDepth + 1,
 					definition.relativePrefix() + relativePath,
@@ -227,7 +228,7 @@ class IndexSchemaFilter {
 		DepthFilter newDepthFilter = DepthFilter.of( definition.includeDepth() );
 
 		// The new path filter according to the given includedPaths
-		PathFilter newPathFilter = PathFilter.of( definition.includePaths() );
+		PathFilter newPathFilter = PathFilter.of( definition.includePaths(), definition.excludePaths() );
 
 		return new IndexSchemaFilter(
 				this, definition, pathTracker,
--- a/engine/src/test/java/org/hibernate/search/engine/mapper/mapping/building/impl/ConfiguredIndexSchemaManagerNestingContextTest.java
+++ b/engine/src/test/java/org/hibernate/search/engine/mapper/mapping/building/impl/ConfiguredIndexSchemaManagerNestingContextTest.java
@@ -99,7 +99,7 @@ public class ConfiguredIndexSchemaManagerNestingContextTest {
 
 		ConfiguredIndexSchemaNestingContext level1Context = checkSimpleIndexedEmbeddedIncluded(
 				"level1", rootContext, typeModel1Mock, "level1.prefix1_",
-				null, null
+				null, null, null
 		);
 		checkFooBarIncluded( "prefix1_", level1Context );
 
@@ -113,7 +113,7 @@ public class ConfiguredIndexSchemaManagerNestingContextTest {
 
 		ConfiguredIndexSchemaNestingContext level2Context = checkSimpleIndexedEmbeddedIncluded(
 				"prefix1_level2", level1Context, typeModel2Mock, "level2.prefix2_",
-				null, null
+				null, null, null
 		);
 		checkFooBarIncluded( "prefix2_", level2Context );
 	}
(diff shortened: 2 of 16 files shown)


engine/src/test/java/org/hibernate/search/engine/mapper/mapping/building/impl/ConfiguredIndexSchemaManagerNestingContextTest.java
```diff
@@ -921,10 +921,96 @@ public void indexedEmbedded_includePaths_embedding_depth2AndIncludePaths() {
...
+		checkCompositeIncluded( "prefix2_level3", level2Context, "prefix2_level3" );
+		checkLeafExcluded( "level3.foo", level2Context, "level3.foo" );
+		checkCompositeExcluded( "level3.foo", level2Context, "level3.foo" );
+
+		// Check IndexedEmbedded composition with a prefix
+
+		level2Context = checkSimpleIndexedEmbeddedIncluded(
+				"level2", level1Context, typeModel2Mock, "level2.prefix2_",
+				null, null, null
+		);
```
yrodiere, line 984: I think the rest of the test are split into one test method per scenario... perhaps it would be good to do the same here? It's already complex enough without having to figure out which parts of the code above in the same method are relevant and which are not xD

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/smoke/ExcludePathsIT.java
```diff
@@ -0,0 +1,291 @@
...
+import org.hibernate.search.util.common.impl.CollectionHelper;
+import org.hibernate.search.util.impl.integrationtest.common.rule.BackendMock;
+import org.hibernate.search.util.impl.integrationtest.mapper.pojo.standalone.StandalonePojoMappingSetupHelper;
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+
+public class ExcludePathsIT {
```
yrodiere, line 36: I'd avoid putting too much in the "smoke" package, better organize tests by feature...

Any reason you didn't just put this next to `org.hibernate.search.integrationtest.mapper.pojo.mapping.definition.IndexedEmbeddedBaseIT#includePaths`, reusing `doTestEmbeddedRuntime` if possible?

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/smoke/ExcludePathsIT.java
```diff
@@ -0,0 +1,291 @@
...
+							)
+					)
+					.add( "6", b -> b
+							.field( "myLocalDateField", entity6.getLocalDate() )
+							.field( "myTextField", entity6.getText() )
+							.objectField( "customBridgeOnClass", b2 -> b2
+									.field( "text", entity6.getText() )
+									.field( "date", entity6.getLocalDate() )
+							)
+					);
```
yrodiere, line 214: I hope you'll simplify all this, there seems to be a fair amount of unnecessary complexity :)

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/IndexedEmbedded.java
```diff
@@ -135,6 +136,9 @@
 	 */
 	String[] includePaths() default {};
 
+	@Incubating
+	String[] excludePaths() default {};
```
yrodiere, line 140: Needs javadoc. I know, it's just a draft, but more importantly: keep in mind that the documentation (javadoc and reference) for `includeDepths` might need adjustments as well.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/programmatic/PropertyMappingIndexedEmbeddedStep.java
```diff
@@ -53,6 +53,8 @@ default PropertyMappingIndexedEmbeddedStep includePaths(String ... paths) {
 	 */
 	PropertyMappingIndexedEmbeddedStep includePaths(Collection<String> paths);
 
+	PropertyMappingIndexedEmbeddedStep excludePaths(Collection<String> paths);
```
yrodiere, line 56: Needs javadoc.

yrodiere (commented):
Yes, this seems to be going in the right direction. I added a few comments below.

The challenge will be in testing this exhaustively, I believe. You've started that, though I didn't check in details what was covered.

I'd encourage you to try to think of every possible corner case and interaction with existing features:

* A parent excluding fields explicitly included by a child
* A parent excluding fields implicitly included by a child through includeDepth
* A parent excluding fields implicitly included by a child that just includes everything

Don't forget non-functional areas, e.g. errors for includes/excludes on invalid paths (that's what the path trackers are for):

* If a parent tries to exclude a field that doesn't exist, we'd want an error
* If a parent tries to exclude a field that's already excluded by a child, we'd want an error (it can be the same error as above, no need to be specific)

Also, please close the PR and reopen it later if you intend to iterate some more, so that I don't get flooded with notifications xD

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
```diff
@@ -119,16 +119,17 @@ public boolean isPathIncluded(String relativePath) {
...
+				&& !pathFilter.isExplicitlyExcluded( relativePath );
 
 		boolean includedByParent = true;
 		/*
 		 * The parent can filter out paths that are considered as included by a child,
 		 * by reducing the includeDepth in particular,
 		 * but it cannot include paths that are filtered out by a child.
 		 */
-		if ( parent != null ) {
+		if ( includedByThis && parent != null ) {
```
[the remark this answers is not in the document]
marko-bekhta: yup, I also wasn't sure, but I think the tests passed just with that change (I'll double-check it while working on this patch)

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/smoke/ExcludePathsIT.java
marko-bekhta: +100 😄 
I just liked the model in `AnnotationMappingSmokeIT` so I've put it here while I was looking into how things work internally with all these paths. I'll probably just drop this test in the end. Thanks for pointing me to `IndexedEmbeddedBaseIT` 😃

marko-bekhta:
Hey, thanks for taking a look at this and for suggestions 🎉 I'll close the PR and will come back with a new one once I'm ready 😄

PR #3521, opened 2023-06-02
Add .mailmap to the root of the repository

yrodiere:
So that:

1. `git log` and similar commands stop displaying "Jenkins user <some random email>" for release commits from back when Jenkins was incorrectly configured, and just display "Hibernate-CI <ci@hibernate.org>" instead.
2. `git shortlog` correctly groups together commits of people who committed with multiple names (e.g. Sanne Grinovero vs. Sanne.Grinovero).

See https://git-scm.com/docs/gitmailmap


yrodiere:
Note this won't work in IDEA though:

* https://youtrack.jetbrains.com/issue/IDEA-160677
* https://youtrack.jetbrains.com/issue/IDEA-121066

PR #3522, opened 2023-06-02
HSEARCH-4149 Follow-up: Remove duplicate sentence in documentation

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4149

This was probably caused by some mixup while addressing review comments.


marko-bekhta (approved):
Thanks!

PR #3525, opened 2023-06-06
HSEARCH-1182 Add `excludePaths` to IndexedEmbedded

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-1182

Not going into recursion in `isPathIncludedInternal` when `!includedByThis` is messing up the path tracking, so I removed the change and added a note there so someone wouldn't try it in the future 😄

### DIFF
--- a/documentation/src/main/asciidoc/reference/mapping-indexedembedded.asciidoc
+++ b/documentation/src/main/asciidoc/reference/mapping-indexedembedded.asciidoc
@@ -257,7 +257,7 @@ which Hibernate Search will detect and reject with an exception.
 
 To address these problems, it is possible to filter the fields to embed,
 to only include those that are actually useful.
-Two filtering attributes are available on `@IndexedEmbedded` and may be combined:
+Available filtering attributes on `@IndexedEmbedded` are:
 
 `includePaths`::
 The paths of index fields from the indexed-embedded element that should be embedded.
@@ -267,24 +267,39 @@ i.e. they must not include its <<mapping-indexedembedded-name,name>>
 or <<mapping-indexedembedded-prefix,prefix>>.
 +
 This takes precedence over `includeDepth` (see below).
++
+Cannot be used in combination with `excludePaths` in the same `@IndexedEmbedded`.
+`excludePaths`::
+The paths of index fields from the indexed-embedded element that must *not* be embedded.
++
+Provided paths must be relative to the indexed-embedded element,
+i.e. they must not include its <<mapping-indexedembedded-name,name>>
+or <<mapping-indexedembedded-prefix,prefix>>.
++
+This takes precedence over `includeDepth` (see below).
++
+Cannot be used in combination with `includePaths` in the same `@IndexedEmbedded`.
 `includeDepth`::
 The number of levels of indexed-embedded that will have all their fields included by default.
 +
 `includeDepth` is the number of `@IndexedEmbedded` that will be traversed
 and for which all fields of the indexed-embedded element will be included,
-even if these fields are not included explicitly through `includePaths`:
+even if these fields are not included explicitly through `includePaths`,
+unless these fields are excluded explicitly through `excludePaths`:
 +
  * `includeDepth=0` means that fields of the indexed-embedded element are *not* included,
 nor is any field of nested indexed-embedded elements,
 unless these fields are included explicitly through `includePaths`.
  * `includeDepth=1` means that fields of the indexed-embedded element *are* included,
+unless these fields are excluded explicitly through `excludePaths`,
 but *not* fields of nested indexed-embedded elements,
 unless these fields are included explicitly through `includePaths`.
  * And so on.
 +
 The default value depends on the value of the `includePaths` attribute:
-if `includePaths` is empty, the default is `Integer.MAX_VALUE` (include all fields at every level)
-if `includePaths` is *not* empty, the default is `0`
++
+* if `includePaths` is empty, the default is `Integer.MAX_VALUE` (include all fields at every level)
+* if `includePaths` is *not* empty, the default is `0`
 (only include fields included explicitly).
 
 [NOTE]
@@ -298,8 +313,16 @@ constraints only need to match the nearest static parent of a dynamic field
 in order for that field to be included.
 ====
 
-Below are two examples: one leveraging `includePaths` only,
-and one leveraging `includePaths` and `includeDepth`.
+[NOTE]
+.Mixing `includePaths` and `excludePaths` at different nesting levels
+====
+In general, it is possible to use `includePaths` and `excludePaths` at different levels of nested `@IndexedEmbedded`.
+When doing so, keep in mind that the filter at each level can only reference reachable paths,
+i.e. a filter cannot reference a path that was excluded by a nested `@IndexedEmbedded` (implicitly or explicitly).
+====
+
+Below are three examples: one leveraging `includePaths` only,
+one leveraging `excludePaths`, and one leveraging `includePaths` and `includeDepth`
 
 [[indexedembedded-includePath]]
 .Filtering indexed-embedded fields with `includePaths`
@@ -322,6 +345,29 @@ and *not* explicitly included either because `includePaths` on `parents` does no
 [source, JAVA, indent=0, subs="+callouts"]
 ----
 include::{sourcedir}/org/hibernate/search/documentation/mapper/orm/indexedembedded/includepaths/Human.java[tags=include;!getters-setters]
+----
+====
+
+[[indexedembedded-excludePath]]
+.Filtering indexed-embedded fields with `excludePaths`
+====
+This mapping will result in the same schema as in the <<indexedembedded-includePath>> example, but through using the `excludePaths` instead.
+Following fields in the `Human` index will be declared:
+
+* `name`
+* `nickname`
+* `parents.name`: implicitly included because `includeDepth` on `parents` defaults to `Integer.MAX_VALUE`.
+* `parents.nickname`: implicitly included because `includeDepth` on `parents` defaults to `Integer.MAX_VALUE`.
+* `parents.parents.name`: implicitly included because `includeDepth` on `parents` defaults to `Integer.MAX_VALUE`.
+
+The following fields in particular are excluded:
+
+* `parents.parents.nickname`: *not* included because `excludePaths` explicitly excludes `parents.nickname`.
+* `parents.parents.parents`/`parents.parents.parents.<any-field>`: *not* included because `excludePaths` explicitly excludes `parents.parents` stopping any further traversing.
+
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/mapper/orm/indexedembedded/excludepaths/Human.java[tags=include;!getters-setters]
 ----
 ====
 
--- a/engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/ConfiguredIndexSchemaNestingContext.java
+++ b/engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/ConfiguredIndexSchemaNestingContext.java
@@ -87,7 +87,7 @@ class ConfiguredIndexSchemaNestingContext implements IndexSchemaNestingContext {
 			int nextDotIndex = prefixToParse.indexOf( '.', afterPreviousDotIndex );
 			while ( nextDotIndex >= 0 ) {
 				// Make sure to mark the paths as encountered in the filter
-				String objectNameRelativeToFilter = prefixToParse.substring( 0, nextDotIndex );
+				String objectNameRelativeToFilter = prefixToParse.substring( unconsumedPrefix.length(), nextDotIndex );
 
 				// we don't want to proceed if a subpath is already excluded:
 				if ( !filter.isPathIncluded( objectNameRelativeToFilter ) ) {
--- a/engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/ConfiguredIndexSchemaNestingContext.java
+++ b/engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/ConfiguredIndexSchemaNestingContext.java
@@ -80,18 +80,22 @@ class ConfiguredIndexSchemaNestingContext implements IndexSchemaNestingContext {
 			IndexedEmbeddedPathTracker pathTracker,
 			NestedContextBuilder<T> contextBuilder) {
 		IndexSchemaFilter composedFilter = filter.compose( definition, pathTracker );
+
 		if ( !composedFilter.isEveryPathExcluded() ) {
 			String prefixToParse = unconsumedPrefix + definition.relativePrefix();
 			int afterPreviousDotIndex = 0;
 			int nextDotIndex = prefixToParse.indexOf( '.', afterPreviousDotIndex );
 			while ( nextDotIndex >= 0 ) {
-				String objectName = prefixToParse.substring( afterPreviousDotIndex, nextDotIndex );
-				contextBuilder.appendObject( objectName );
-
 				// Make sure to mark the paths as encountered in the filter
 				String objectNameRelativeToFilter = prefixToParse.substring( 0, nextDotIndex );
-				// We only use isPathIncluded for its side effect: it marks the path as encountered
-				filter.isPathIncluded( objectNameRelativeToFilter );
+
+				// we don't want to proceed if a subpath is already excluded:
+				if ( !filter.isPathIncluded( objectNameRelativeToFilter ) ) {
+					return Optional.empty();
+				}
+
+				String objectName = prefixToParse.substring( afterPreviousDotIndex, nextDotIndex );
+				contextBuilder.appendObject( objectName );
 
 				afterPreviousDotIndex = nextDotIndex + 1;
 				nextDotIndex = prefixToParse.indexOf( '.', afterPreviousDotIndex );
--- a/engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
+++ b/engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
@@ -108,7 +108,7 @@ class IndexSchemaFilter {
 	/**
 	 * @param relativeDepth The number of indexed-embeddeds between this filter and the end of the path.
 	 * @param relativePath The path relative to this filter.
-	 * @param markAsEncountered Whether the path should be marked as encountered
+	 * @param markIncludedAsEncountered Whether the included path should be marked as encountered
 	 * ({@code true}), or not ({@code false}).
 	 * Useful when processing "theoretical" paths, such as the includePaths of a child filter, in particular.
 	 * @param includedByChild Whether a child filter included this path.
@@ -118,7 +118,7 @@ class IndexSchemaFilter {
 	 * @return {@code true} if the path is included, {@code false} otherwise.
 	 */
 	private boolean isPathIncludedInternal(int relativeDepth, String relativePath,
-			boolean markAsEncountered, boolean includedByChild) {
+			boolean markIncludedAsEncountered, boolean includedByChild) {
 		boolean includedByThis = ( depthFilter.isEveryPathIncludedAtDepth( relativeDepth )
 				|| pathFilter.isExplicitlyIncluded( relativePath ) )
 				&& !pathFilter.isExplicitlyExcluded( relativePath );
@@ -136,7 +136,7 @@ class IndexSchemaFilter {
 			includedByParent = parent.isPathIncludedInternal(
 					relativeDepth + 1,
 					definition.relativePrefix() + relativePath,
-					markAsEncountered,
+					markIncludedAsEncountered,
 					// need to include a results from the previous step as it may be that we have an include at the current level
 					// but it was excluded at some ancestor level, and we want to pass it up to the parent.
 					includedByThis && includedByChild
@@ -145,7 +145,12 @@ class IndexSchemaFilter {
 
 		boolean included = includedByParent && includedByThis;
 
-		if ( markAsEncountered && pathTracker != null ) {
+		// when we are checking if any path is included we want to make sure we record the paths we've checked, but they are excluded:
+		// - if we end up excluding the entire object because *none* of the "includePaths" are actually included
+		// we want to record the path so that the exclude filter will know it wasn't useless,
+		// - if we end up including the object because some of the "includePaths" are actually included we'll mark those
+		// paths later when we'll be considering adding them.
+		if ( ( markIncludedAsEncountered || !included ) && pathTracker != null ) {
 			pathTracker.markAsEncountered(
 					relativePath, includedByThis, includedByChild
 			);
--- a/engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
+++ b/engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
@@ -119,20 +119,27 @@ class IndexSchemaFilter {
 	 */
 	private boolean isPathIncludedInternal(int relativeDepth, String relativePath,
 			boolean markAsEncountered, boolean includedByChild) {
-		boolean includedByThis = depthFilter.isEveryPathIncludedAtDepth( relativeDepth )
-				|| pathFilter.isExplicitlyIncluded( relativePath );
+		boolean includedByThis = ( depthFilter.isEveryPathIncludedAtDepth( relativeDepth )
+				|| pathFilter.isExplicitlyIncluded( relativePath ) )
+				&& !pathFilter.isExplicitlyExcluded( relativePath );
 
 		boolean includedByParent = true;
 		/*
 		 * The parent can filter out paths that are considered as included by a child,
 		 * by reducing the includeDepth in particular,
 		 * but it cannot include paths that are filtered out by a child.
+		 *
+		 * Note: If we only go into recursion when `includedByThis == true` path tracking of parent nodes will not record
+		 * the route.
 		 */
-		if ( parent != null ) {
+		if ( includedByThis && parent != null ) {
 			includedByParent = parent.isPathIncludedInternal(
 					relativeDepth + 1,
 					definition.relativePrefix() + relativePath,
-					markAsEncountered, includedByThis
+					markAsEncountered,
+					// need to include a results from the previous step as it may be that we have an include at the current level
+					// but it was excluded at some ancestor level, and we want to pass it up to the parent.
+					includedByThis && includedByChild
 			);
 		}
 
@@ -140,7 +147,7 @@ class IndexSchemaFilter {
 
 		if ( markAsEncountered && pathTracker != null ) {
 			pathTracker.markAsEncountered(
-					relativePath, includedByThis && includedByChild
+					relativePath, includedByThis, includedByChild
 			);
 		}
 
@@ -227,7 +234,7 @@ class IndexSchemaFilter {
 		DepthFilter newDepthFilter = DepthFilter.of( definition.includeDepth() );
 
 		// The new path filter according to the given includedPaths
-		PathFilter newPathFilter = PathFilter.of( definition.includePaths() );
+		PathFilter newPathFilter = PathFilter.of( definition.includePaths(), definition.excludePaths() );
 
 		return new IndexSchemaFilter(
 				this, definition, pathTracker,
--- a/engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
+++ b/engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
@@ -102,7 +102,7 @@ class IndexSchemaFilter {
 	}
 
 	public boolean isPathIncluded(String relativePath) {
-		return isPathIncludedInternal( 0, relativePath, true, true );
+		return isPathIncludedInternal( 0, relativePath, true );
 	}
 
 	/**
@@ -111,14 +111,10 @@ class IndexSchemaFilter {
 	 * @param markIncludedAsEncountered Whether the included path should be marked as encountered
 	 * ({@code true}), or not ({@code false}).
 	 * Useful when processing "theoretical" paths, such as the includePaths of a child filter, in particular.
-	 * @param includedByChild Whether a child filter included this path.
-	 * Useful when marking a path as encountered: if it is included by this filter,
-	 * but not by any children, the corresponding includePaths is not effective,
-	 * and thus we should consider the path as excluded for the purpose of detecting ineffective includePaths.
 	 * @return {@code true} if the path is included, {@code false} otherwise.
 	 */
 	private boolean isPathIncludedInternal(int relativeDepth, String relativePath,
-			boolean markIncludedAsEncountered, boolean includedByChild) {
+			boolean markIncludedAsEncountered) {
 		boolean includedByThis = ( depthFilter.isEveryPathIncludedAtDepth( relativeDepth )
 				|| pathFilter.isExplicitlyIncluded( relativePath ) )
 				&& !pathFilter.isExplicitlyExcluded( relativePath );
@@ -128,18 +124,12 @@ class IndexSchemaFilter {
 		 * The parent can filter out paths that are considered as included by a child,
 		 * by reducing the includeDepth in particular,
 		 * but it cannot include paths that are filtered out by a child.
-		 *
-		 * Note: If we only go into recursion when `includedByThis == true` path tracking of parent nodes will not record
-		 * the route.
 		 */
 		if ( includedByThis && parent != null ) {
 			includedByParent = parent.isPathIncludedInternal(
 					relativeDepth + 1,
 					definition.relativePrefix() + relativePath,
-					markIncludedAsEncountered,
-					// need to include a results from the previous step as it may be that we have an include at the current level
-					// but it was excluded at some ancestor level, and we want to pass it up to the parent.
-					includedByThis && includedByChild
+					markIncludedAsEncountered
 			);
 		}
 
@@ -151,9 +141,7 @@ class IndexSchemaFilter {
 		// - if we end up including the object because some of the "includePaths" are actually included we'll mark those
 		// paths later when we'll be considering adding them.
 		if ( ( markIncludedAsEncountered || !included ) && pathTracker != null ) {
-			pathTracker.markAsEncountered(
-					relativePath, includedByThis, includedByChild
-			);
+			pathTracker.markAsEncountered( relativePath, includedByThis );
 		}
 
 		return included;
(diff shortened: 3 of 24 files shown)


engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
```diff
@@ -222,21 +233,69 @@ public IndexSchemaFilter compose(IndexedEmbeddedDefinition definition, IndexedEm
...
 				this, definition, pathTracker,
 				newDepthFilter, newPathFilter
 		);
 	}
 
+	private Set<String> findConflictingIncludeExcludePaths(IndexedEmbeddedDefinition definition) {
+		Set<String> conflictingPaths = new LinkedHashSet<>();
+		for ( String path : definition.excludePaths() ) {
+			conflictingPathExplicitlyIncluded( definition.relativePrefix() + path )
+					.ifPresent( conflictingPaths::add );
```
yrodiere, line 258: I wonder what's the point of this...

We're testing a path that's explicitly excluded. Since a given filter cannot use both `includePaths` and `excludePaths`, it's safe to say that paths won't be explicitly included by the same filter.

As to parent filters... I would expect path tracking to be enough to detect such conflicts, i.e. the parent filter will have an `excludePaths` filter for a path that is simply never encountered, so you'll get an error automatically?

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/IndexedEmbeddedBaseIT.java
```diff
@@ -1617,7 +1822,480 @@ class EntityC {
...
+				.setup() )
+				.isInstanceOf( SearchException.class )
+				.satisfies( FailureReportUtils.hasFailureReport()
+						.typeContext( Model.IndexedEntity.class.getName() )
+						.pathContext( ".included<no value extractors>.subIncluded<no value extractors>.subSubIncluded" )
+						.failure(
+								"Found conflicting include/exclude paths configuration.",
+								"Included paths are: '[]', excluded paths are: '[subSubIncludedString]'.",
+								"To resolve the conflict either change the configuration of the current property or look at `includePaths` or `excludePaths` of the first node of paths in '[included.subIncluded.subSubIncluded.subSubIncludedString]'."
+						)
```
yrodiere, line 1895: I'd expect this test to just raise the same error as in `error_excludePaths_nonMatched`, because `subIncluded.subSubIncluded.subSubIncludedString` simply doesn't exist... Why do we need a different check?

If you need more arguments:

1. This message says there's a conflict between `includePaths` and `excludePaths`, but also says included paths are empty... ?
2. This message recommends to fix the problem in the child, while the problem really is in the parent which tries to reference a path that doesn't match anything.

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/IndexedEmbeddedBaseIT.java
```diff
@@ -1617,7 +1822,480 @@ class EntityC {
...
+								"Found conflicting include/exclude paths configuration.",
+								"Included paths are: '[subSubIncludedString]', excluded paths are: '[]'.",
+								"To resolve the conflict either change the configuration of the current property or look at `includePaths` or `excludePaths` of the first node of paths in '[included.subIncluded.subSubIncluded.subSubIncludedString]'."
+						)
+				);
+	}
+
+	/*
+	 * Since parent adds a path to be excluded and child uses an include filter that explicitly includes a different field -- there's no conflict.
+	 * TODO: filter may look as redundant, do we want to fail in such case?
```
yrodiere, line 1971: Yes we do want to fail. `excludePaths = "subIncluded.subSubIncluded.subSubIncludedString"` is effectively what the "useless exclude paths" detection is about: this filter should be reported as useless.

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/IndexedEmbeddedBaseIT.java
```diff
@@ -1617,7 +1822,480 @@ class EntityC {
...
+		assertThatThrownBy( () -> setupHelper.start()
+				.withAnnotatedEntityTypes( Model.IndexedEntity.class )
+				.setup() )
+				.isInstanceOf( SearchException.class )
+				.satisfies( FailureReportUtils.hasFailureReport()
+						.typeContext( Model.Included.class.getName() )
+						.failure(
+								"An @IndexedEmbedded defines excludePaths filters that were already excluded by the parent filter",
+								"Already excluded paths: [subIncludedStringA].",
+								"Check the filters for typos, or remove them if they are already excluded by a parent and hence are not useful."
```
yrodiere, line 2297: Here too, there's no need for a specific filter. The parent `excludePaths` should be detected as useless.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/IndexedEmbedded.java
```diff
@@ -126,34 +127,58 @@
...
+	String[] excludePaths() default {};
+
 	/**
 	 * The number of levels of indexed-embedded that will have all their fields included by default.
 	 * <p>
 	 * {@code includeDepth} is the number of `@IndexedEmbedded` that will be traversed
 	 * and for which all fields of the indexed-embedded element will be included,
-	 * even if these fields are not included explicitly through {@code includePaths}:
+	 * even if these fields are not included explicitly through {@code includePaths},
+	 * unless a field is excluded explicitly through {@code excludePaths}:
```
yrodiere, line 164:
```suggestion
	 * unless these fields are excluded explicitly through {@code excludePaths}:
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/IndexedEmbedded.java
```diff
@@ -126,34 +127,58 @@
...
 	 * <ul>
 	 * <li>{@code includeDepth=0} means fields of the indexed-embedded element are <strong>not</strong> included,
 	 * nor is any field of nested indexed-embedded elements,
 	 * unless these fields are included explicitly through {@link #includePaths()}.
 	 * <li>{@code includeDepth=1} means fields of the indexed-embedded element <strong>are</strong> included,
+	 * unless a field is explicitly excluded through {@code excludePaths},
```
yrodiere, line 170:
```suggestion
	 * unless these fields are explicitly excluded through {@code excludePaths},
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/IndexedEmbedded.java
```diff
@@ -126,34 +127,58 @@
...
-	 * if {@link #includePaths()} is empty, the default is {@code Integer.MAX_VALUE} (include all fields at every level)
-	 * if {@link #includePaths()} is <strong>not</strong> empty, the default is {@code 0}
-	 * (only include fields included explicitly).
+	 * The default value depends on the value of {@link #includePaths()}/{@link #excludePaths()} attributes:
+	 * <ul>
+	 * <li> if both {@link #includePaths()} and {@link #excludePaths()} are empty, the default is {@code Integer.MAX_VALUE} (include all fields at every level) </li>
+	 * <li> if {@link #includePaths()} is <strong>not</strong> empty and {@link #excludePaths()} is empty, the default is {@code 0} (only include fields included explicitly)</li>
+	 * <li> if {@link #includePaths()} is empty and {@link #excludePaths()} is <strong>not</strong> empty, the default is {@code Integer.MAX_VALUE} (include all fields except those that are excluded explicitly)</li>
+	 * <li> having both {@link #includePaths()} and {@link #excludePaths()} <strong>not</strong> empty is not allowed and will result in an exception.</li>
+	 * </ul>
```
yrodiere, line 181: Same thing I already mentioned elsewhere: IMO this change is not necessary.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/programmatic/PropertyMappingIndexedEmbeddedStep.java
```diff
@@ -53,6 +54,26 @@ default PropertyMappingIndexedEmbeddedStep includePaths(String ... paths) {
 	 */
 	PropertyMappingIndexedEmbeddedStep includePaths(Collection<String> paths);
 
+	/**
+	 * @param paths The paths of index fields to exclude.
+	 * @return {@code this}, for method chaining.
+	 *
+	 * @see IndexedEmbedded#excludePaths() ()
```
yrodiere, line 61:
```suggestion
	 * @see IndexedEmbedded#excludePaths()
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/programmatic/PropertyMappingIndexedEmbeddedStep.java
```diff
@@ -53,6 +54,26 @@ default PropertyMappingIndexedEmbeddedStep includePaths(String ... paths) {
...
+	@Incubating
+	default PropertyMappingIndexedEmbeddedStep excludePaths(String... paths) {
+		return excludePaths( Arrays.asList( paths ) );
+	}
+
+	/**
+	 * @param paths The paths of index fields to exclude.
+	 * @return {@code this}, for method chaining.
+	 *
+	 * @see IndexedEmbedded#excludePaths() ()
```
yrodiere, line 72:
```suggestion
	 * @see IndexedEmbedded#excludePaths()
```

yrodiere (changes requested):
Thanks. I need to have a closer look at `ConfiguredIndexSchemaManagerNestingContextTest`, but I think there's something to be done on the error reporting. See below, but in short I think the "useless paths" detection should already cover what you tried to address yourself (as I warned you on Zulip... :) ); if it doesn't, then there's probably a problem with path tracking.

documentation/src/main/asciidoc/reference/mapping-indexedembedded.asciidoc
```diff
@@ -267,25 +267,41 @@ i.e. they must not include its <<mapping-indexedembedded-name,name>>
...
-The default value depends on the value of the `includePaths` attribute:
-if `includePaths` is empty, the default is `Integer.MAX_VALUE` (include all fields at every level)
-if `includePaths` is *not* empty, the default is `0`
-(only include fields included explicitly).
+The default value depends on the value of `includePaths`/`excludePaths` attributes:
++
+* if both `includePaths` and `excludePaths` are empty, the default is `Integer.MAX_VALUE` (include all fields at every level)
+* if `includePaths` is *not* empty and `excludePaths` is empty, the default is `0` (only include fields included explicitly)
+* if `includePaths` is empty and `excludePaths` is *not* empty, the default is `Integer.MAX_VALUE` (include all fields except those that are excluded explicitly)
+* having both `includePaths` and `excludePaths` *not* empty is *not* allowed and will result in an exception.
```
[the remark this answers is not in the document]
marko-bekhta: mmm 
> * if `includePaths` is empty and `excludePaths` is *not* empty, the default is `Integer.MAX_VALUE` (include all fields except those that are excluded explicitly)

if we have some exclude paths -- depth is max int... now I wonder if others would also miss it and if we should rephrase this 🙈

documentation/src/main/asciidoc/reference/mapping-indexedembedded.asciidoc
```diff
@@ -298,8 +314,16 @@ constraints only need to match the nearest static parent of a dynamic field
...
 ====
 
-Below are two examples: one leveraging `includePaths` only,
-and one leveraging `includePaths` and `includeDepth`.
+[NOTE]
+.Mixing `includePaths` and `excludePaths` at different nesting levels
+====
+In general, it is possible to use `includePaths` and `excludePaths` at different levels of nested `@IndexedEmbedded`.
+When doing so, keep in mind that the resulting filters cannot be conflicting, i.e. including a property at one level
+and excluding it at the other will result in an exception being thrown.
```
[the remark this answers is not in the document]
marko-bekhta: Ouch 😆  but that was the whole point of adding that `findConflictingIncludeExcludePaths` stuff ... 
What I was trying to address there was something like:
```java
class IndexedEntity {
	@IndexedEmbedded(excludePaths = "subIncluded.subIncludedString")
	Included included;
}

class Included {
	@KeywordField
	String includedString;
	// but we've excluded it explicitly already at the very top with exclude filter
	// should we include? exclude? no idea let's fail instead :)
	@IndexedEmbedded(includePaths = "subIncludedString") 
	SubIncluded subIncluded;
}

class SubIncluded {
	@KeywordField
	String subIncludedString;
}
```
and what I was thinking that we cannot simply leave it to path tracking or what we had before, since for example we could have something included/excluded because of a depth, but if it is included by depth and excluded by a excludePaths entry - it's ok we should allow it...

documentation/src/main/asciidoc/reference/mapping-indexedembedded.asciidoc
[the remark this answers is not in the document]
yrodiere: But you also have:

> * if both `includePaths` and `excludePaths` are empty, the default is `Integer.MAX_VALUE` (include all fields at every level)

So as soon as `includePaths` is empty, the default is `Integer.MAX_VALUE`, whether `excludePaths` is empty or not. `excludePaths` does not matter.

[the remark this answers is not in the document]
marko-bekhta: well now 😄 if you put it like that...  it makes `excludePaths` feel... excluded 😆🤣

documentation/src/main/asciidoc/reference/mapping-indexedembedded.asciidoc
```diff
@@ -552,4 +599,4 @@ This mapping will declare the following fields in the `Book` index:
 ----
 include::{sourcedir}/org/hibernate/search/documentation/mapper/orm/indexedembedded/onelevel/IndexedEmbeddedOneLevelIT.java[tags=programmatic]
 ----
-====
\ No newline at end of file
+====
```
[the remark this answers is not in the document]
marko-bekhta: yeah, I think it comes from IDE (Settings -> Editor -> General -> Ensure line feed at file end on Save) and I might have it in git as well (an old habit I guess 🙈🥲)
I may or may not be wrong, but I think that the Eclipse formatter does this too, (or maybe I've added that in the config for it) and my idea was to have these blank lines added while we are modifying the files anyway so that when we do the mass formatting there'll be fewer changes...

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
```diff
@@ -119,14 +122,19 @@ public boolean isPathIncluded(String relativePath) {
 	 */
 	private boolean isPathIncludedInternal(int relativeDepth, String relativePath,
 			boolean markAsEncountered, boolean includedByChild) {
-		boolean includedByThis = depthFilter.isEveryPathIncludedAtDepth( relativeDepth )
-				|| pathFilter.isExplicitlyIncluded( relativePath );
+		boolean explicitlyExcludedByThis = pathFilter.isExplicitlyExcluded( relativePath );
```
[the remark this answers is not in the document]
marko-bekhta: This was for those already-excluded paths ... now it's back to being one line 😃

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
```diff
@@ -142,6 +150,9 @@ private boolean isPathIncludedInternal(int relativeDepth, String relativePath,
 			pathTracker.markAsEncountered(
 					relativePath, includedByThis && includedByChild
 			);
+			if ( explicitlyExcludedByThis && !includedByParent ) {
+				pathTracker.markAsAlreadyExcluded( relativePath );
+			}
```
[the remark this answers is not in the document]
marko-bekhta: That would be a `IndexedEmbeddedBaseIT#parentIncludeChildExcludeForDifferentFieldsIsFine`

documentation/src/main/asciidoc/reference/mapping-indexedembedded.asciidoc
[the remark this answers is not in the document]
yrodiere: > my idea was to have these blank lines added while we are modifying the files anyway so that when we do the mass formatting there'll be fewer changes...

Good point

marko-bekhta:
Note that there're a few `//TODO: to Yoann:` 😉 those are in places that you'd probably want to double-check 😃 
It seemed easier to note things down in the code than trying to point to them from here 🙈

documentation/src/main/asciidoc/reference/mapping-indexedembedded.asciidoc
```diff
@@ -267,24 +267,39 @@ i.e. they must not include its <<mapping-indexedembedded-name,name>>
...
++
+Cannot be used in combination with `includePaths` in the same `@IndexedEmbedded`.
 `includeDepth`::
 The number of levels of indexed-embedded that will have all their fields included by default.
 +
 `includeDepth` is the number of `@IndexedEmbedded` that will be traversed
 and for which all fields of the indexed-embedded element will be included,
-even if these fields are not included explicitly through `includePaths`:
+even if these fields are not included explicitly through `includePaths`,
+unless a field is excluded explicitly through `excludePaths`:
```
yrodiere, line 288:
```suggestion
unless these fields are excluded explicitly through `excludePaths`:
```

documentation/src/main/asciidoc/reference/mapping-indexedembedded.asciidoc
```diff
@@ -298,8 +313,17 @@ constraints only need to match the nearest static parent of a dynamic field
...
-Below are two examples: one leveraging `includePaths` only,
-and one leveraging `includePaths` and `includeDepth`.
+[NOTE]
+.Mixing `includePaths` and `excludePaths` at different nesting levels
+====
+In general, it is possible to use `includePaths` and `excludePaths` at different levels of nested `@IndexedEmbedded`.
+When doing so, keep in mind that the filter at each level can only reference reachable paths,
+i.e. filter cannot reference a path that, in general is there but, was made unreachable by a filter at a lower level that
+excluded at least one of this path's nodes (implicitly or explicitly).
+====
```
yrodiere, line 323:
```suggestion
i.e. a filter cannot reference a path that was excluded by a nested `@IndexedEmbedded` (implicitly or explicitly).
====
```

documentation/src/main/asciidoc/reference/mapping-indexedembedded.asciidoc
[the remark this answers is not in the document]
yrodiere: Discussed here: https://hibernate.zulipchat.com/#narrow/stream/132092-hibernate-search-dev/topic/HSEARCH-1182.20IndexEmbedded.20exclude.20paths

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
```diff
@@ -119,28 +123,36 @@ public boolean isPathIncluded(String relativePath) {
...
+				&& !pathFilter.isExplicitlyExcluded( relativePath );
 
 		boolean includedByParent = true;
 		/*
 		 * The parent can filter out paths that are considered as included by a child,
 		 * by reducing the includeDepth in particular,
 		 * but it cannot include paths that are filtered out by a child.
+		 *
+		 * Note: If we only go into recursion when `includedByThis == true` path tracking of parent nodes will not record
+		 * the route.
```
yrodiere, line 137: Is this still relevant? I'd expect path tracking in parent nodes to not make the difference between a non-existing path and a path excluded in children, so I don't see how this is important.

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
```diff
@@ -119,28 +123,36 @@ public boolean isPathIncluded(String relativePath) {
...
 		 */
 		if ( parent != null ) {
 			includedByParent = parent.isPathIncludedInternal(
 					relativeDepth + 1,
 					definition.relativePrefix() + relativePath,
-					markAsEncountered, includedByThis
+					markAsEncountered,
+					// need to include a results from the previous step as it may be that we have an include at the current level
+					// but it was excluded at some child level above, and we want to pass it down to the parent.
```
yrodiere, line 145: "child level above" ? Is that the right term?

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/spi/IndexedEmbeddedPathTracker.java
```diff
@@ -43,13 +45,35 @@ public Set<String> uselessIncludePaths() {
...
 
+	public Set<String> uselessExcludePaths() {
+		Set<String> uselessExcludePaths = new LinkedHashSet<>();
+		for ( String path : definition.excludePaths() ) {
+			if (
+				// since it's exclude we should always get a FALSE value from this map for such key.
+				// Hence, we only want to check for fields that we never encountered i.e. some typos in property names or just some "pointless" text :)
+					!encounteredFieldPaths.containsKey( path )
+							// OR the field was also "explicitly" excluded by a child, which would be redundant...
+							// If it wasn't encountered than it is already covered by a containsKey check above.
```
yrodiere, line 56:
```suggestion
							// If it wasn't encountered then it is already covered by a containsKey check above.
```

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/IndexedEmbeddedBaseIT.java
```diff
@@ -1617,10 +1822,682 @@ class EntityC {
...
+								"Non-matching includePaths filters: [subIncluded.subSubIncluded.subSubIncludedString].",
+								"Encountered field paths: [includedString, subIncluded, subIncluded.subIncludedString]. Check the filters for typos, or remove them if they are not useful."
+						)
+				);
+	}
+
+	/*
+	 * Child has a property (that we included explicitly), so if we are trying to exclude it at a parent level it should
+	 * get excluded without complains. *BUT*, because it's a single field in that embedded we end up dropping the entire embedded
+	 * resulting in an error.
```
yrodiere, line 1905: Okay so this is confusing... I'd expect no error in this case, or at least not something that tells me `subIncluded.subSubIncluded.subSubIncludedString` doesn't exist, since it actually does.

> because it's a single field in that embedded we end up dropping the entire embedded

This is fine, but since the dropping was triggered by the exclusion of `subIncluded.subSubIncluded.subSubIncludedString`, I'd expect that exclusion to have marked the path as encountered, and then we would not report the filter as useless... ? Isn't something wrong here?

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/IndexedEmbeddedBaseIT.java
```diff
@@ -1617,10 +1822,682 @@ class EntityC {
...
+										.objectField( "subSubIncluded", b4 -> b4
+												.field( "subSubIncludedString", "d" )
+										)
+								)
+						)
+		);
+	}
+
+	/*
+	 * Since parent adds a path to be excluded and child uses an include filter that explicitly includes a different field -- parent won't find a filed it tries to exclude
```
yrodiere, line 2084:
```suggestion
	 * Since parent adds a path to be excluded and child uses an include filter that explicitly includes a different field -- parent won't find a field it tries to exclude
```

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/IndexedEmbeddedBaseIT.java
```diff
@@ -1617,10 +1822,682 @@ class EntityC {
...
+		assertThatThrownBy( () -> setupHelper.start()
+				.withAnnotatedEntityTypes( Model.IndexedEntity.class )
+				.setup() )
+				.isInstanceOf( SearchException.class )
+				.satisfies( FailureReportUtils.hasFailureReport()
+						.typeContext( Model.IndexedEntity.class.getName() )
+						.failure(
+								"An @IndexedEmbedded defines excludePaths filters that do not match anything",
+								"Non-matching excludePaths filters: [subIncluded.subIncludedStringA].",
+								"Encountered field paths: [includedString, subIncluded, subIncluded.subIncludedStringA, subIncluded.subIncludedStringB, subIncluded.subIncludedStringC].",
```
yrodiere, line 2407: `subIncluded.subIncludedStringA` is non-matching but still encountered?

There's something wrong with the message at the very least, and possibly with the algorithm.
I'd expect paths that are excluded in children to simply not be reported to parents, leading to those paths not being marked as encountered in parents... ?

yrodiere (commented):
Getting there... Though this is getting tricky. My comments below.

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/ConfiguredIndexSchemaNestingContext.java
```diff
@@ -107,11 +104,44 @@ public <T> Optional<T> addIndexedEmbeddedIfIncluded(
...
+		int afterPreviousDotIndex = 0;
+		int nextDotIndex = prefixToParse.indexOf( '.', afterPreviousDotIndex );
+		boolean shouldInclude = true;
+		while ( nextDotIndex >= 0 ) {
+			// Make sure to mark the paths as encountered in the filter
+			String objectNameRelativeToFilter = prefixToParse.substring( 0, nextDotIndex );
+
+			//TODO: to Yoann: I am not convinced if this is a correct way to address it or not, but based on the tests
+			// we have for prefixes I've guessed that if we have a prefixed embedded then `unconsumedPrefix` will have that prefix in it
+			// and we not necessarily need to "include" a path-with-prefix in the filter if we want to have a property to be included ( DeepPathWithLeadingPrefixCase )
```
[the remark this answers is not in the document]
marko-bekhta: These 3 would fail:

TestInvalidPaths.testDepthMatchesPathMarkedAsEncountered:
```
org.hibernate.search.util.common.SearchException: HSEARCH000520: Hibernate Search encountered failures during bootstrap. Failures:

    Hibernate ORM mapping: 
        type 'org.hibernate.search.test.embedded.path.validation.A': 
            failures: 
              - HSEARCH000216: An @IndexedEmbedded defines includePaths filters that do not match anything. Non-matching includePaths filters: [c.indexed]. Encountered field paths: [prefixedc]. Check the filters for typos, or remove them if they are not useful.
```

TestInvalidPaths.testDepthExceedsPathMarkedAsEncountered:
```
org.hibernate.search.util.common.SearchException: HSEARCH000520: Hibernate Search encountered failures during bootstrap. Failures:

    Hibernate ORM mapping: 
        type 'org.hibernate.search.test.embedded.path.validation.A': 
            failures: 
              - HSEARCH000216: An @IndexedEmbedded defines includePaths filters that do not match anything. Non-matching includePaths filters: [c.indexed]. Encountered field paths: [prefixedc]. Check the filters for typos, or remove them if they are not useful.
```

TestInvalidPaths.testInvalidDeepSimplePathWithLeadingPrefix:
```
org.hibernate.search.util.common.SearchException: HSEARCH000520: Hibernate Search encountered failures during bootstrap. Failures:

    Hibernate ORM mapping: 
        type 'org.hibernate.search.test.embedded.path.validation.A': 
            failures: 
              - HSEARCH000216: An @IndexedEmbedded defines includePaths filters that do not match anything. Non-matching includePaths filters: [c.indexed]. Encountered field paths: [prefixedc]. Check the filters for typos, or remove them if they are not useful.
        type 'org.hibernate.search.test.embedded.path.validation.DeepPathWithLeadingPrefixCase': 
            failures: 
              - HSEARCH000216: An @IndexedEmbedded defines includePaths filters that do not match anything. Non-matching includePaths filters: [b.c.indexed, b.c.dne]. Encountered field paths: [notJustAb]. Check the filters for typos, or remove them if they are not useful.
```

Before any changes - we'd check that some paths are included and be happy about it. It wouldn't matter that the path that we get at that point is "not included" when the filter is called inside the while loop. And now that we have to make this additional check, we'd need to have a path that matches something 🥲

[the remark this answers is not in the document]
yrodiere: Can you push the code that leads to these errors somewhere, so that I have a closer look?

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
yrodiere: Not sure I understand what you mean exactly... the "previous child level" is just... the parent?

How about:

```suggestion
					// but it was excluded at some ancestor level above, and we want to pass it up to the parent.
```

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
marko-bekhta: Adding that check actually helps to make the messages cleaner ([in this test for example](https://github.com/hibernate/hibernate-search/pull/3525#discussion_r1221406802))
But it is breaking a few preexisting tests, so I didn't want to make any changes there.
For example:
`org.hibernate.search.test.embedded.path.validation.TestInvalidPaths#testInvalidDeepSimplePathWithLeadingPrefix`:
expects to get `"Encountered field paths: [notJustAb, b.c, b.c.indexed, b.c.notIndexed]"`
but will only get: `Encountered field paths: [notJustAb, b.c, b.c.indexed].`

Or the path tracking test in wants to see a path that "shouldn't" be there - https://github.com/hibernate/hibernate-search/blob/78c77ae0d0347d8250afa028297f8555283f1d5b/engine/src/test/java/org/hibernate/search/engine/mapper/mapping/building/impl/ConfiguredIndexSchemaManagerNestingContextTest.java#L491

yrodiere: Feel free to add the check if it fixes tests with no downside.

>  For example:
> `org.hibernate.search.test.embedded.path.validation.TestInvalidPaths#testInvalidDeepSimplePathWithLeadingPrefix`:
> expects to get `"Encountered field paths: [notJustAb, b.c, b.c.indexed, b.c.notIndexed]"`
> but will only get: `Encountered field paths: [notJustAb, b.c, b.c.indexed].` 

The new behavior seems more correct.

> Or the path tracking test in wants to see a path that "shouldn't" be there - 

Same, the new behavior seems more correct. If the path is excluded at level 2, I don't see why it should be visible in the path tracker for level 1.

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/ConfiguredIndexSchemaNestingContext.java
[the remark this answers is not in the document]
marko-bekhta: yup, sure, 🙈  pushed it here - https://github.com/marko-bekhta/hibernate-search/blob/feat/HSEARCH-1182-exclude-paths-indexed-embedded-error-reproducer and added a [comment with failing tests](https://github.com/marko-bekhta/hibernate-search/blob/e9e2ee4009c547798d1e0e854b0d750c45288cb2/engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/ConfiguredIndexSchemaNestingContext.java#L94C1-L103)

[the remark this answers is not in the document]
yrodiere: You actually discovered a bug, and it was hidden in plain sight in one of our tests...

I sent a PR to your branch: https://github.com/marko-bekhta/hibernate-search/pull/135

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/IndexedEmbeddedBaseIT.java
marko-bekhta: +1 On it being not very clear. One could try to make an argument that `subIncluded.subSubIncluded.subSubIncludedString` is a useless filter as in such case `subIncluded.subSubIncluded` would do the same. 
The particular problem comes from the fact that we eliminate the entire embedded by `IndexSchemaFilter#isEveryPathExcluded()` which explicitly asks not to track paths... Let me think if we can catch that in some clean way ...

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
```diff
@@ -118,29 +118,41 @@ public boolean isPathIncluded(String relativePath) {
...
 		/*
 		 * The parent can filter out paths that are considered as included by a child,
 		 * by reducing the includeDepth in particular,
 		 * but it cannot include paths that are filtered out by a child.
+		 *
+		 * Note: If we only go into recursion when `includedByThis == true` path tracking of parent nodes will not record
+		 * the route.
 		 */
-		if ( parent != null ) {
+		if ( includedByThis && parent != null ) {
```
yrodiere, line 135: Shouldn't this be:

```suggestion
		if ( includedByThis && includedByChild && parent != null ) {
```

... ?

Though now that I'm thinking of this, I wonder if the `includedByChild` parameter is even necessary anymore. We only ever call this method with `includedByChild == true`, don't we?

yrodiere (commented):
LGTM, just one remaining comment; I suspect you might be able to simplify things a bit (maybe even remove the map about "includedByChild" you added to the path tracker xD)

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
marko-bekhta: Hmm, maybe ... I'll double-check it (and if the other map is still necessary)  😃 making changes to these paths is like playing Minesweeper 💣 😆

marko-bekhta: sooooooooo it looks like, map is not needed, `includedByChild` is not needed 😃 
Thanks for looking into this 🥇😃

yrodiere (approved):

yrodiere:
Great job, thanks! Let's merge when CI agrees.

PR #3526, opened 2023-06-06
HSEARCH-4863 Update Highlight DSL docs

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4863

- add a few highlight result examples. I've looked at highlighter options and it might be nice to have example results for some, but then I suppose we'd need to add them for all options -_-  and I'm not sure that we should. Also, I've tried to add the results with `include` but the end result is not looking nice so it's just a "hardcoded" text 😃 
- change the section title to match others
- split the type examples into their own blocks ( didn't like how it looked originally) 
- use the same Book for all examples (so I just dropped that book-with-highlights and rearranged the properties a bit so that we wouldn't need to change the tests/examples)


documentation/src/main/asciidoc/reference/search-dsl-highlighting.asciidoc
```diff
@@ -54,6 +54,20 @@ If there's no match within the field we want to highlight, be it because the doc
...
+    ["The Boscombe Valley <em>Mystery</em>"],
+    # Hit #1:
+    ["A Caribbean <em>Mystery</em>"],
+    # Hit #2:
+    ["A <em>Mystery</em> of <em>Mysteries</em>: The Death and Life of Edgar Allan Poe"]
+    # Other hits
+    ...
+]
+----
+Note that the result of applying a highlight projection is always a list of `String`.
```
yrodiere, line 70: If you add a blank line before this, it will move the snippet out of the callout `<5>` and then you might be able to use callouts instead of comments:

```suggestion
Note that the result of applying a highlight projection is always a list of `String`.

As an example, the result may look like:

[source, JSON]
----
[
    ["The Boscombe Valley <em>Mystery</em>"], // <1>
    ["A Caribbean <em>Mystery</em>"], // <2>
    ["A <em>Mystery</em> of <em>Mysteries</em>: The Death and Life of Edgar Allan Poe"]  // <3>
]
----
<1> First hit.
<2> Second hit.
<3> Third hit.
```

(I used the JSON syntax highlighting here since we already did that [elsewhere](https://github.com/hibernate/hibernate-search/blob/26f1da6bef433ca33650d22b010ace005c745e91/documentation/src/main/asciidoc/reference/binding-valuebridge.asciidoc?plain=1#L73-L82))

documentation/src/main/asciidoc/reference/search-dsl-projection.asciidoc
```diff
@@ -1201,6 +1201,17 @@ a single or a multivalued one:
...
+----
+[
+    # Hit #0:
+    ["The Automatic <em>Detective</em>"],
+    # Hit #1:
+    ["Dirk Gently's Holistic <em>Detective</em> Agency"],
+    # Other hits
+    ...
+]
+----
```
yrodiere, line 1214: Same as above, maybe use callouts + a hit with two elements?

yrodiere (approved):
Thanks!

LGTM, I just added a suggestion to improve the example, but feel free to merge as is if you disagree.

documentation/src/main/asciidoc/reference/search-dsl-highlighting.asciidoc
marko-bekhta: 😃 
also, I've tried and from what it looks it's not necessary to have a  `//` in json callouts so I've used just <#>

yrodiere (approved):
LGTM, thanks!

PR #3528, opened 2023-06-08
HSEARCH-4847 Extract getting started guides

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4847

It would probably be easier to look at this per commit 🙈 since there are only just a few changes 🙈 🙈 🙈 

Links in HTML are generated as relative 
```
<a href="../getting-started/orm/index.html#mapper-orm-getting-started">Getting started with the Hibernate ORM integration</a>
```
So that works nicely as long as we don't change the structure Asciidoctor generated.  As for the pdf .... not so nice as it seems to just try and add an absolute file path 😃.


documentation/src/main/assembly/dist.xml
```diff
@@ -15,25 +15,68 @@
...
+        <!-- so that the links between the files would still work fine. -->
         <fileSet>
             <directory>${asciidoctor.aggregated-resources-dir}/css/</directory>
-            <outputDirectory>reference/en-US/html_single/css/</outputDirectory>
+            <outputDirectory>public/en-US/html_single/reference/css/</outputDirectory>
         </fileSet>
         <fileSet>
             <directory>${asciidoctor.aggregated-resources-dir}/script/</directory>
-            <outputDirectory>reference/en-US/html_single/script/</outputDirectory>
+            <outputDirectory>public/en-US/html_single/reference/script/</outputDirectory>
```
yrodiere, line 28: If you're going to change the document structure, you will need to update some harcoded links.

Look for `https://docs.jboss.org` thoughought the codebase.

yrodiere (commented):
Thanks. I have a few comments, see below.

documentation/pom.xml
```diff
@@ -392,9 +393,11 @@
...
                                 <configuration>
                                     <backend>pdf</backend>
-                                    <sourceDocumentName>reference/index.asciidoc</sourceDocumentName>
-                                    <outputDirectory>${asciidoctor.base-output-dir}/reference/pdf</outputDirectory>
-                                    <outputFile>hibernate_search_reference.pdf</outputFile>
+                                    <sourceDirectory>src/main/asciidoc/public</sourceDirectory>
+                                    <outputDirectory>${asciidoctor.base-output-dir}/public/pdf</outputDirectory>
+                                    <preserveDirectories>true</preserveDirectories>
+                                    <!-- since the file name will be used for all the generated PDFs, let's keep it more generic: -->
+                                    <outputFile>hibernate_search.pdf</outputFile>
```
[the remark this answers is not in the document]
marko-bekhta: I've added a `<files>` section to dist.xml so that we can rename files. we weren't copying many files with those `*/*.smth` patterns anyway 😃

documentation/src/main/assembly/dist.xml
marko-bekhta: I think the structure remained, with these latest changes, so no changes to doc URLs 🙈

documentation/src/main/asciidoc/public/reference/index.adoc
```diff
@@ -0,0 +1,70 @@
...
+:revdate: {docdate}
+:sectanchors:
+:anchor:
+:toc: left
+:toclevels: 4
+:docinfodir: {docinfodir}
+:docinfo: shared,private
+:title-logo-image: image:hibernate_logo_a.png[align=left,pdfwidth=33%]
+
+:relfileprefix: ../../../just-to-be-removed/
```
yrodiere, line 12: Those "just to be removed" things are confusing. I'd suggest adding a comment each time you do that, explaining why?

documentation/src/main/assembly/dist.xml
```diff
@@ -72,13 +95,61 @@
...
+        <file>
+            <source>${asciidoctor.base-output-dir}/pdf/migration/index.pdf</source>
+            <outputDirectory>migration/pdf/</outputDirectory>
+            <destName>hibernate-search-migration.pdf</destName>
+        </file>
+        <file>
+            <source>${asciidoctor.base-output-dir}/pdf/internals/index.pdf</source>
+            <outputDirectory>internals/pdf/</outputDirectory>
+            <destName>hibernate-search-internals.pdf</destName>
+        </file>
```
yrodiere, line 153: This needs to be conditional, or to ignore missing files, because PDFs are not generated by default (it's too slow).

I'm getting this when I don't generate PDFs:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-assembly-plugin:3.4.2:single (make-dist) on project hibernate-search-documentation: Failed to create assembly: Error adding file to archive: /home/yrodiere/workspaces/main/hibernate-search/documentation/target/asciidoctor/pdf/public/reference/index.pdf -> [Help 1]
```

yrodiere (commented):
Thanks! This looks really good.

Last round of comments, feel free to merge after they are addressed :)

documentation/src/main/assembly/dist.xml
marko-bekhta: ohh right, good point 😃 will do something about it 👍🏻

marko-bekhta: I didn't see anything useful in maven-assembly-plugin to make pdfs optional, so I've moved the copying of pdfs to a different descriptor file, which is included in a pdf profile.

documentation/src/main/asciidoc/public/reference/index.adoc
marko-bekhta: added a comment with an example + turned out it wasn't needed in all the places -- so I've simplified some of these URLs.

yrodiere (approved):

yrodiere:
Thanks! Let's merge once CI agrees.

yrodiere:
Jenkins, retest this please.

PR #3531, opened 2023-06-12
HSEARCH-4870 Add compatibility with OpenSearch 2.8 / HSEARCH-4871 Upgrade to Elasticsearch client 8.8.1

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4870
https://hibernate.atlassian.net/browse/HSEARCH-4871

I've run the tests locally -- there's still a bug with BigDecimal in 8.8.1 😞😃


yrodiere (commented):
Thanks.

I believe the second commit message is wrong, it should mention 2.8 instead of 2.7.

Other than that, LGTM

yrodiere:
Actually the fourth commit message seems wrong as well :grin:

yrodiere (approved):

PR #3533, opened 2023-06-12
HSEARCH-4869 Drop compatibility with Hibernate ORM 6.1 and 6.0 in -orm6 artifacts, get rid of deprecated/internal class references

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4869

### DIFF
--- a/orm6/mapper/orm/ant-src-changes.patch
+++ b/orm6/mapper/orm/ant-src-changes.patch
@@ -498,6 +498,32 @@ index 21503d08be..af34072fdf 100644
  						.collect( Collectors.toList() )
  		);
  	}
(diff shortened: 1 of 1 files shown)


yrodiere:
Actually the diff might be more readable if you have a look at the corresponding commits [here](https://github.com/hibernate/hibernate-search/commits/wip/main/dependency-update/orm6-in-main-code)

marko-bekhta (approved):
Looks nice!
I cannot see changes from the last commit (changes to`HibernateOrmMassEntityLoader`) in the orm6 branch. Is it GH misbehaving or do you just need to push them 🙈 😃

yrodiere:
Thanks.

> I cannot see changes from the last commit (changes to`HibernateOrmMassEntityLoader`) in the orm6 branch. Is it GH misbehaving or do you just need to push them 🙈 😃 

I fixed up an earlier commit that was removing unnecessary casts to `Serializable`, that's why you don't see it in later commits.

PR #3534, opened 2023-06-13
HSEARCH-4872 Rebuild only some of the modules to speedup the build

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4872

### DIFF
--- a/ci/dependency-update/Jenkinsfile
+++ b/ci/dependency-update/Jenkinsfile
@@ -38,10 +38,16 @@ Map settings() {
 			]
 		case 'lucene9':
 			return [
-					updateProperties: ['version.org.apache.lucene']
-					// In this case we'll rebuild and test everything. Since we have some JDK 11 changes as well as
-					// some driver updates. These changes spans across multiple modules and we would either need to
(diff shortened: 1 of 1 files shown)


yrodiere:
Thanks.

We'll need to check that `hibernate-search-integrationtests-backend-lucene` actually gets executed even when you add those options; can you please check that?

Merging.

marko-bekhta:
I've tried to run it locally:
```bash
./ci/list-dependent-integration-tests.sh hibernate-search-backend-lucene
```
has the `hibernate-search-integrationtest-backend-lucene`  in the list, and then I've just added `-pl` with the output of this script + another `-pl` for the list from `additionalMavenArgs` -- it even passed the tests 😃 but we probably would want to double-check it when it runs on CI. So I'll look at the results once we have a next run

marko-bekhta:
sooo it seems that it worked 🎉 😄 
```
[INFO] Hibernate Search ITs - Backend - Lucene ............ SUCCESS [01:37 min]
```
https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search-dependency-update/detail/main/110/pipeline/444

PR #3535, opened 2023-06-13
HSEARCH-4873 Test Hibernate Search against JDK 22

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4873


marko-bekhta (approved):
🎉 😃

yrodiere:
The build failed because of a deprecation warning, I just pushed a commit to main to fix that. Let's try again.

PR #3544, opened 2023-06-19
HSEARCH-4876 StackOverflow error when processing cyclic indexed embedded with excludePaths

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4876

see also https://github.com/yrodiere/hibernate-search/pull/68

We might be able to get rid of some of the "string operations", but let's see first if the algorithm is ok, as it makes it easier to "debug" 😃


engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
```diff
@@ -196,24 +199,113 @@ private String getPathFromSameIndexedEmbeddedSinceNoCompositionLimits(IndexedEmb
...
+				longestCycle.substring( longestCycleNode.definition.relativePrefix().length() ),
+				longestCycleNode
+		) ) {
+			return null;
+		}
+
+		// 2. maybe there's a node following the one causing the cycle that would break it at some point.
+		// If so we want to go one node at a time, and only check the "new nodes" i.e. if we encounter a same node for the second time
+		// it means even if that node at the current step may break the cycle - we've already checked it before and it didn't
+		// in the previous steps, meaning that if that node breaks a cycle, it wouldn't be the one we are currently in:
```
yrodiere, line 266: This is... hard to follow. At least I am not sure I understand :/ Sorry...

Maybe let's agree on some vocabulary:

1. "node" => "filter"
2. "same node" => "same definition" or "another filter with the same definition" (at least I think that's what you mean? Not sure...)
3. Maybe not "the node causing the cycle" but "the first filter with the given definition" or "the filter at the start of the cycle".
4. "the node we're currently in" => "the current filter (`this`)"?
5. Feel free to use "the filter being added" somewhere if it makes sense, since cycle detection is done exclusively when adding a new filter.

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
```diff
@@ -196,24 +199,113 @@ private String getPathFromSameIndexedEmbeddedSinceNoCompositionLimits(IndexedEmb
...
+		return ( cycle.substring( path.length() ) + cycle.substring( 0, path.length() ) ) // <-- making a permutation
+				.substring( currentNodeRelativePath.length() ); // <-- dropping the current node
+	}
+
+	private String removeDot(String string) {
+		return string != null && string.endsWith( "." ) ? string.substring( 0, string.length() - 1 ) : string;
+	}
+
+	private boolean isPotentiallyExcludedPath(String path, IndexSchemaFilter node) {
+		return node.definition != null && (
```
yrodiere, line 302:
```suggestion
		return node.definition != null // The root node doesn't exclude any path.
				&& (
```

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
```diff
@@ -196,24 +199,113 @@ private String getPathFromSameIndexedEmbeddedSinceNoCompositionLimits(IndexedEmb
...
+		) ) {
+			return null;
+		}
+
+		// 2. maybe there's a node following the one causing the cycle that would break it at some point.
+		// If so we want to go one node at a time, and only check the "new nodes" i.e. if we encounter a same node for the second time
+		// it means even if that node at the current step may break the cycle - we've already checked it before and it didn't
+		// in the previous steps, meaning that if that node breaks a cycle, it wouldn't be the one we are currently in:
+		Set<IndexedEmbeddedDefinition> encounteredNodesSoFar = new HashSet<>();
+		// we don't want to start from the beginning but from the next node from the one at which we've found the longest cycle:
```
yrodiere, line 268: Wouldn't checking the shortest cycle be enough here? Or maybe just the first occurrence of the cycle?

I can't imagine how e.g. cycle number 3 would change things and close what was considered an infinite cycle in cycle 2.

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
```diff
@@ -196,24 +199,111 @@ private String getPathFromSameIndexedEmbeddedSinceNoCompositionLimits(IndexedEmb
...
+		for ( int index = paths.indexOf( longestCycle ); index < nodes.size(); index++ ) {
+			IndexSchemaFilter node = nodes.get( index );
+			// so we know that we are in the cycle, and we know what the cycle looks like, if the current node can break a cycle
+			// it would mean that if we make a cycle permutation that starts with the "current node ^" and this same node can
+			// potentially break the cycle -- then we are ok to break it:
```
yrodiere, line 273: I must admit this seems more complex than what I imagined, but maybe I'm missing some context.
Maybe let's discuss this over video chat.

marko-bekhta:
Prefixes caused a few more edge cases 😃. I've tried to explain it all in the comments where new code was added so that we know why things are as they are 😄

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
```diff
@@ -196,24 +206,91 @@ private String getPathFromSameIndexedEmbeddedSinceNoCompositionLimits(IndexedEmb
...
-			/*
-			 * No recursion limits, no parent: this is the root.
-			 * I we reach this point, it means there was no recursion limit at all,
-			 * but we did not encounter the IndexedEmbedded we were looking for.
-			 */
+
+		// If we reached this point -- we've found a cycle, and we need to check if any exclude paths can break it:
+		if ( isPotentiallyExcludedPath(
+				// when we check the node we don't need the node's relative path at the beginning of the path we are checking:
+				previousCycle.substring( previousCycleFilter.definition.relativePrefix().length() ),
```
yrodiere, line 273: Looks to me like you should rather adjust your loop so that `previousCycle` matches what you need.

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
```diff
@@ -196,24 +206,91 @@ private String getPathFromSameIndexedEmbeddedSinceNoCompositionLimits(IndexedEmb
...
+	 * and through its parents will cover the definitions from the cycle itself as well as any definitions out of the cycle that are in the path leading to the cycle.
+	 *
+	 * In case we don't have a repeating cycle yet - we'll be starting from the "current filter"
+	 * (previousCycleFilter == this, previousCycle == this.definition.relativePrefix() )
+	 */
+	private String findCycle(IndexedEmbeddedDefinition definitionToBeAdded) {
+		String shortestCycle = null;
+		String currentCycle = null;
+		String previousCycle = this.definition.relativePrefix();
+		IndexSchemaFilter localParent = this;
```
yrodiere, line 239: Maybe `currentFilter`? You don't even initialize this with a parent :)

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
```diff
@@ -196,24 +206,91 @@ private String getPathFromSameIndexedEmbeddedSinceNoCompositionLimits(IndexedEmb
...
 			return null;
 		}
+
+		// We haven't found an exclude that would break the cycle - so we return the shortest cycle
+		return shortestCycle;
+	}
+
+	private String removeDot(String string) {
+		return string != null && string.endsWith( "." ) ? string.substring( 0, string.length() - 1 ) : string;
+	}
```
yrodiere, line 285: Didn't we discuss just ignoring that kind of things? It didn't work?

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/ConfiguredIndexSchemaNestingContext.java
```diff
@@ -85,6 +85,55 @@ public <T> Optional<T> addIndexedEmbeddedIfIncluded(
...
+				//     B b;
+				// }
+				// class B {
+				//     @IndexedEmbedded(prefix="obj.prefix_")
+				//     C c;
+				// }
+				// class C { ... }
+				//
+				// When we'll be adding node `C c` checking of `obj.` will happen in the while loop, but not `obj.prefix_`
+				// and that is why we are checking it with this additional explicit exclude check.
```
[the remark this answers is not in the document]
marko-bekhta: discussed here https://hibernate.zulipchat.com/#narrow/stream/132092-hibernate-search-dev/topic/HSEARCH-1182.20IndexEmbedded.20exclude.20paths/near/367837397
(incorrect expectations in a prefix-no-dots case, hence this will go away)

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/ConfiguredIndexSchemaNestingContext.java
```diff
@@ -85,6 +85,55 @@ public <T> Optional<T> addIndexedEmbeddedIfIncluded(
...
+				//     B b;
+				// }
+				// class B {
+				//     @IndexedEmbedded(prefix="obj.prefix_")
+				//     C c;
+				// }
+				// class C { ... }
+				//
+				// when we'd be adding `C c` we will end up checking `obj.prefix_` if it is included. BUT because we have
+				// an include path -- this sub-path will result in a "false-positive" exclude from `filter.isPathIncluded()`
```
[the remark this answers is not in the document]
marko-bekhta: discussed here https://hibernate.zulipchat.com/#narrow/stream/132092-hibernate-search-dev/topic/HSEARCH-1182.20IndexEmbedded.20exclude.20paths/near/367837397
(incorrect expectations in a prefix-no-dots case, hence this will go away)

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
marko-bekhta: mmm, I think we've talked about ignoring this in the `PathFilter#isPotentiallyExcluded()` where we wouldn't care if we may cut some property into two parts and would just use a "starts with"... as eventually the cycle will grow, and the path will not be "potentially excluded" anymore.
Here though, we need to pass something to the `IndexSchemaFilter`, and that one expects that there are no dots in the end... So I guess we need this one, but calling it inside of the recursion makes no sense -- I'll move the call before the recursion happens.

yrodiere: Ok. In that case... if you end up with a trailing dot, you're not manipulating a path... there's some trailing prefix that you probably should not even have included in the first place. 

Maybe instead of removing a trailing dot, you should be removing everything after (and including) the last dot at some point? I think we even discussed that...

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
```diff
@@ -196,24 +195,91 @@ private String getPathFromSameIndexedEmbeddedSinceNoCompositionLimits(IndexedEmb
...
+	 * The idea is to find the longest cycle so far possible.
+	 * There is a chance that the cycle may repeat a few times if there are exclude paths that may break the cycle at some point.
+	 * Let's assume we have a cycle where we are trying to add next `a.` definition same as the already added ones before:
+	 * a.b.c.a.b.c.a.b.c.a.b.c. + a.
+	 * 1   ^ 2     3     4        5
+	 *
+	 * - we'll be looking for a definition `a.` with index `1` and that would define our "longest cycle".
+	 * - we want to know the filter representing the definition `c.` (marked with `^`) that is right before the `a.` with index `2` as well as the
+	 * cycle-path leading to it from `5` -- "a.b.c.a.b.c.a.b.c."
+	 * that will be the filter and the path we'll pass to the "isPotentiallyExcludedPath" check. These will be our `previousCycleFilter` and `previousCycle`
```
yrodiere, line 216:
```suggestion
	 * that will be the filter and the path we'll pass to the "isPotentiallyExcludedPath" check. These will be our `previousCycleFilter` and `pathFromPreviousCycleFilter`
```

You probably should have another look at the rest of this comment to be sure it still matches the code.

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
yrodiere: Bump, just in case you missed this :)


> Ok. In that case... if you end up with a trailing dot, you're not manipulating a path... there's some trailing prefix that you probably should not even have included in the first place.
> 
> Maybe instead of removing a trailing dot, you should be removing everything after (and including) the last dot at some point? I think we even discussed that...

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
```diff
@@ -196,24 +195,91 @@ private String getPathFromSameIndexedEmbeddedSinceNoCompositionLimits(IndexedEmb
...
+		// If we reached this point -- we've found a cycle, and we need to check if any exclude paths can break it:
+		if ( isPotentiallyExcludedPath(
+				// when we check the path against the filter it expects that the path doesn't end with a dot:
+				removeDot( pathFromPreviousCycleFilter ),
+				previousCycleFilter
+		) ) {
 			return null;
 		}
+
+		// We haven't found an exclude that would break the cycle - so we return the shortest cycle
```
yrodiere, line 268:
```suggestion
		// We haven't found an exclude that would break the cycle - so we return the closest cycle
```

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
```diff
@@ -196,24 +195,91 @@ private String getPathFromSameIndexedEmbeddedSinceNoCompositionLimits(IndexedEmb
...
+		if ( this.definition == null ) {
+			// we are at root
+			return null;
+		}
+
+		return findCycle( definition );
+	}
+
+	/*
+	 * The idea is to find the longest cycle so far possible.
```
yrodiere, line 207:
```suggestion
	 * The idea is to find the farthest cycle, i.e. the cycle closest to the root.
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/IndexedEmbedded.java
```diff
@@ -135,10 +135,11 @@
 	 * @return The paths of index fields to include explicitly.
 	 * Provided paths must be relative to the indexed-embedded element,
 	 * i.e. they must not include the {@link #name()}.
+	 * Paths <b>must</b> lead to a property and <b>cannot</b> end with some {@link #prefix() prefix used to construct a property name}.
```
yrodiere, line 138:
```suggestion
	 * Paths <b>must</b> lead to a field and <b>cannot</b> end with some {@link #prefix() prefix used to construct a field name}.
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/IndexedEmbedded.java
```diff
@@ -152,7 +153,9 @@
 	 * @return The paths of index fields to exclude explicitly.
 	 * Provided paths must be relative to the indexed-embedded element,
 	 * i.e. they must not include the {@link #name()}.
+	 * Paths <b>must</b> lead to a property and <b>cannot</b> end with some {@link #prefix() prefix used to construct a property name}.
```
yrodiere, line 156:
```suggestion
	 * Paths <b>must</b> lead to a field and <b>cannot</b> end with some {@link #prefix() prefix used to construct a field name}.
```

yrodiere (commented):
Thanks, here are a few last comments.

engine/src/main/java/org/hibernate/search/engine/mapper/mapping/building/impl/IndexSchemaFilter.java
yrodiere: > Let's take a simple case where we don't have any prefixes messing up things for us. And we have a cycle:
> [...]
> Our exclude paths in the filters do not have . at the end of the path so passing pathFromPreviousCycleFilter as is won't work. And in this case, removing the dot at the end of the string is what I'd expect to happen.

Agreed.

> now a case with a prefix-without-a-dot. Let's say we have :
> [...]
> So at this point, we can either send c.a.prefix_c.a.prefix_c.a.prefix_ (path as is) or c.a.prefix_c.a.prefix_c.a (path without an "unterminated prefix" in the end that's if I understand you correctly is what you are suggesting) and both would work fine with our path filters

Agreed.

> now if we have a case without any dots in the prefixes....
> [...]
> then we will only have an infinite sequence of prefix_a_prefix_b_+, and we won't be able to pass anything to our path filter if we'd only wanted to pass whatever comes prior to the last dot. so that would lead to an overflow and cycle not being detected ...

In this particular case, we know there's an infinite cycle. `excludePaths` don't matter as they cannot break the cycle (since they will never match any path in the cycle). So IMO it's fine if either:

* `isPotentiallyExcluded()` returns `false` for an empty path
* or if you just add an `if` in `findCycle` that doesn't even call `isPotentiallyExcluded()` if "pathFromPreviousFilterAfterRemovingTrailingPrefix" is empty.

engine/src/main/java/org/hibernate/search/engine/common/tree/impl/TreeFilter.java
```diff
@@ -280,17 +287,37 @@ private String findCycle(MappingElement mappingElement, String relativePrefix, T
...
-				) ) );
+						|| (
+						filterNode.parent != null
+								&& isPotentiallyExcludedPath(
+								// if we have an empty path here it means that we are at the first iteration of cycle detection,
+								// and we've found our first cycle, but that cycle is built from some objects that are using prefixes without dots,
+								// so we are looking for a first occurrence of a dot after which we can start composing a path of proper-object-names.
+								path.isEmpty()
+										? pathOfFullObjectNames( filterNode.relativePrefix )
+										: filterNode.relativePrefix + path,
```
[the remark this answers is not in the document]
marko-bekhta: +1 I think it makes sense. And it also allows us to detect some cycles sooner so that's also a plus.
 THANKS! 🏆 🎖️ 🥇 😄

yrodiere (approved):
LGTM, thanks a lot!

PR #3546, opened 2023-06-21
HSEARCH-4725 Follow-up: Make tests compatible with JDK 8/11

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4725

I've noticed that jdk 8/11 build didn't like static classes 😞 : 
https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/main/697/pipeline/106


yrodiere (approved):
Right, nice catch. Thanks.

PR #3549, opened 2023-06-21
HSEARCH-4868 Drop compatibility with Elasticsearch 5.x/6.x 

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4868

I think we don't want this one in until we have a 6.2 branch 😃


yrodiere (approved):
LGTM, thanks.

Indeed, we should wait until 6.2 gets branched out.

yrodiere:
@marko-bekhta Looks like we'll need a rebase for this one :)

yrodiere (approved):

PR #3553, opened 2023-06-22
HSEARCH-4688 Upgrade to Jandex 3.0

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4688

I was looking to see if anything obvious is needed for this upgrade. Seems like there's only one deprecated method usage. 
But I did notice that there were some issues with dependency convergence, so I've made some changes to poms.


pom.xml
```diff
@@ -278,7 +278,7 @@
              and to our own explicit dependencies in the relevant artifacts.
           -->
         <version.io.smallrye.jandex>3.0.5</version.io.smallrye.jandex>
-        <version.org.hibernate.commons.annotations.orm6>6.0.5.Final</version.org.hibernate.commons.annotations.orm6>
+        <version.org.hibernate.commons.annotations.orm6>6.0.6.Final</version.org.hibernate.commons.annotations.orm6>
```
yrodiere, line 281: Not okay to do this. We need to be aligned on ORM.

yrodiere (changes requested):
Okay... I think we'd better do this at the same time as the ORM 6 upgrade. See the wip branch for the ORM 6 upgrade, I think there's already some (limited) Jandex 3 upgrade there? At the very least we have a dedicated wip branch for the Jandex 3 upgrade.

marko-bekhta:
> Okay... I think we'd better do this at the same time as the ORM 6 upgrade

+100000000000 yes 😃, this PR is towards our `/wip/main/dependency-update/orm6-in-main-code`

I was just checking to see if we will need anything else once we are ready to merge that branch to the main. And that's where I've noticed this.

pom.xml
marko-bekhta: +1 that is actually realigning it back (https://github.com/hibernate/hibernate-orm/blob/e8f96761378720e7fcc2e84113ccca720c89f2a9/settings.gradle#L61) 
we have the correct version in our main branch and in `wip/main/dependency-update/orm6-in-main-code`  we have several commits upgrading orm and annotation versions, so it ends up being incorrect 🙈

yrodiere:
>  this PR is towards our `/wip/main/dependency-update/orm6-in-main-code`

Ah, okay, I didn't notice this. Then feel free to merge, assuming you tested locally, because CI won't work on PRs like this.
But please don't mark the issue as fixed until it's been merged into main.

yrodiere:
Superseded by #3562

PR #3543, opened 2023-06-19
HSEARCH-4865 Update build dependencies

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4865

follows up on https://github.com/hibernate/hibernate-search/pull/3532, https://github.com/hibernate/hibernate-search/pull/3524

PR #3557, opened 2023-06-26
HSEARCH-4865 Update build dependencies

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4865

follows up on https://github.com/hibernate/hibernate-search/pull/3543, https://github.com/hibernate/hibernate-search/pull/3532, https://github.com/hibernate/hibernate-search/pull/3524

Not sure if we are going to include this one or not. If not -- I'll move it to another JIRA 😃.


yrodiere:
> Not sure if we are going to include this one or not. If not -- I'll move it to another JIRA smiley.

Please do, I've already closed this JIRA issue and I am releasing.

In the future, I'm thinking maybe we should stop using JIRAs for simple build dependency updates... I started that, and in retrospect it's not bringing much. Maybe **for build dependencies only**, we should just approve and merge dependabot's PR directly once CI is green?

marko-bekhta:
> stop using JIRAs for simple build dependency updates

+1. It makes sense to have a dedicated JIRA for something like ES client upgrade as that may somewhat affect users, but for build dependencies ... 
How would we let CI not complain about message format ? 😃🙈

yrodiere:
>  How would we let CI not complain about message format?

The bot, you mean? For now I'd say we just ignore it when we know the PR is only about build dependencies.

If you want something better, we'll have to find some way of distinguishing between PRs that just update build dependencies and PRs that update code or other dependencies... I don't see an obvious way to do that...

Maybe we could add support for excluding PRs matching a specific regexps in https://github.com/hibernate/hibernate-github-bot , i.e. tell it to ignore PRs from dependabot whose title contains '.*\bmaven\b.*\bplugin\b.*'? We wouldn't catch *everything*, but already a lot, and we could always add more regexps in the future. Feel free to have a look if you're interested in adding stuff to the bot :)

PR #3561, opened 2023-06-27
HSEARCH-4884   Configuration property to have a dedicated docvalue field for the root document id in the Lucene backend's document

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4884

That seems doable 😃, I've cherrypicked a first commit that originally introduced the change. And in the second commit, I've added a way we could pick which way to go. Do you have any suggestions for the property name? and how does this look overall?

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/document/impl/LuceneLegacyIdReaderWriter.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/document/impl/LuceneLegacyIdReaderWriter.java
@@ -1,0 +1,34 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.lucene.document.impl;
+
+import java.io.IOException;
+
+import org.hibernate.search.backend.lucene.lowlevel.common.impl.MetadataFields;
+
+import org.apache.lucene.document.Document;
+import org.apache.lucene.index.BinaryDocValues;
+import org.apache.lucene.index.DocValues;
+import org.apache.lucene.index.LeafReader;
+
+public final class LuceneLegacyIdReaderWriter implements LuceneIdReaderWriter {
+
+	public static final LuceneIdReaderWriter INSTANCE = new LuceneLegacyIdReaderWriter();
+
+	private LuceneLegacyIdReaderWriter() {
+	}
+
+	@Override
+	public BinaryDocValues idDocValues(LeafReader reader) throws IOException {
+		return DocValues.getBinary( reader, MetadataFields.idFieldName() );
+	}
+
+	@Override
+	public void write(String id, Document document) {
+		document.add( MetadataFields.searchableRetrievableMetadataField( MetadataFields.idFieldName(), id ) );
+	}
+}
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/impl/LuceneBackendFactory.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/impl/LuceneBackendFactory.java
@@ -21,12 +21,16 @@ import org.hibernate.search.backend.lucene.cache.QueryCachingConfigurationContex
 import org.hibernate.search.backend.lucene.cache.QueryCachingConfigurer;
 import org.hibernate.search.backend.lucene.cache.impl.LuceneQueryCachingContext;
 import org.hibernate.search.backend.lucene.cfg.LuceneBackendSettings;
+import org.hibernate.search.backend.lucene.document.impl.LuceneIdReaderWriter;
+import org.hibernate.search.backend.lucene.document.impl.LuceneLegacyIdReaderWriter;
+import org.hibernate.search.backend.lucene.document.impl.LuceneModernIdReaderWriter;
 import org.hibernate.search.backend.lucene.logging.impl.Log;
 import org.hibernate.search.backend.lucene.multitenancy.MultiTenancyStrategyName;
 import org.hibernate.search.backend.lucene.multitenancy.impl.DiscriminatorMultiTenancyStrategy;
 import org.hibernate.search.backend.lucene.multitenancy.impl.MultiTenancyStrategy;
 import org.hibernate.search.backend.lucene.multitenancy.impl.NoMultiTenancyStrategy;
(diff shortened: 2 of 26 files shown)


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/document/impl/LuceneLegacyIdReaderWriter.java
```diff
@@ -0,0 +1,29 @@
...
+public class LuceneLegacyIdReaderWriter implements LuceneIdReaderWriter {
+
+	@Override
+	public BinaryDocValues idDocValues(LeafReader reader) throws IOException {
+		return DocValues.getBinary( reader, MetadataFields.idFieldName() );
+	}
+
+	@Override
+	public void write(String id, Document document) {
+		document.add( MetadataFields.retrievableMetadataField( MetadataFields.idFieldName(), id ) );
```
yrodiere, line 27: This would need to be `searchableRetrievableMetadataField`, wouldn't it?

documentation/src/main/asciidoc/migration/index.adoc
```diff
@@ -43,9 +43,13 @@ Hibernate Search's requirements did not change in version {hibernateSearchVersio
...
 == Data format and schema changes
 
-Indexes created with Hibernate Search {hibernateSearchPreviousStableVersionShort}
+Elasticsearch indexes created with Hibernate Search {hibernateSearchPreviousStableVersionShort}
 can be read from and written to with Hibernate Search {hibernateSearchVersion}.
 
+Reading and writing to Lucene indexes created with Hibernate Search {hibernateSearchPreviousStableVersionShort}
+using Hibernate Search {hibernateSearchVersion} may lead to exceptions, since there were incompatible changes applied to internal fields.
+You must recreate your Lucene indexes and reindex your database. The easiest way to do so is to use link:{hibernateSearchDocUrl}#indexing-massindexer[the `MassIndexer`] with link:{hibernateSearchDocUrl}#indexing-massindexer-parameters-drop-and-create-schema[`dropAndCreateSchemaOnStart(true)`].
+
```
yrodiere, line 52: This should not be added. We're not introducing a backwards-incompatible change.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/impl/LuceneBackendFactory.java
```diff
@@ -165,6 +171,11 @@ private MultiTenancyStrategy getMultiTenancyStrategy(ConfigurationPropertySource
 		}
 	}
 
+	private LuceneIdReaderWriter getIdReaderWriter(ConfigurationPropertySource propertySource) {
+		// TODO: read a property
```
yrodiere, line 175: Say, `hibernate.search.backend.schema.id.strategy`, with two possible values: `lucene8` and `lucene9`?

I'd document this in a new subsection under https://docs.jboss.org/hibernate/search/6.2/reference/en-US/html_single/#backend-lucene-schema . Make sure to mention this can only be set at the backend level...

yrodiere (commented):
Thanks, looks like a plan. I added comments below.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/document/impl/LuceneLegacyIdReaderWriter.java
marko-bekhta: Probably yes 😃 I haven't really compared it to what we had yet, I just wanted to see how many changes we'd need to get it configured 😃.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/impl/LuceneBackendFactory.java
marko-bekhta: ok, thanks 😃 I'll start cleaning up then 😃

marko-bekhta:
OK, I've ended up dropping that first commit. Also, there's a warning printed when users use the lucene8 strategy, suggesting changing to a new one.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/impl/LuceneBackendFactory.java
```diff
@@ -165,6 +175,18 @@ private MultiTenancyStrategy getMultiTenancyStrategy(ConfigurationPropertySource
...
 	}
 
+	@SuppressWarnings("deprecation")
+	private LuceneIdReaderWriter getIdReaderWriter(ConfigurationPropertySource propertySource) {
+		LuceneIdReaderWriter.Strategy strategy = SCHEMA_ID_STRATEGY.get( propertySource );
+
+		if ( LuceneIdReaderWriter.Strategy.LUCENE_8.equals( strategy ) ) {
+			log.usingDeprecatedSchemaIdStrategy( SCHEMA_ID_STRATEGY.resolveOrRaw( propertySource ),
+					LuceneIdReaderWriter.Strategy.LUCENE_9.externalRepresentation() );
+		}
```
yrodiere, line 185: We can't exactly print a warning when using `lucene8` because:

1. It's the default.
2. At the moment it's perfectly reasonable to keep using this strategy. I expect only WildFly will ever make use of `lucene9` TBH.

documentation/src/main/asciidoc/public/reference/_backend-lucene.adoc
```diff
@@ -559,6 +559,40 @@ When indexing, a discriminator field holding the tenant ID is populated transpar
 When searching, a filter targeting the tenant ID field is added transparently to the search query
 to only return search hits for the current tenant.
 
+[[backend-lucene-schema-id-strategy]]
+=== ID strategy
+
+With the upcoming Lucene 9 upgrade, Hibernate Search will require backward incompatible schema changes.
+In particular, the way it works with IDs will change.
+That means that once the change happens, all Lucene indexes must be recreated, and data needs to be repopulated.
```
yrodiere, line 567: Repetition of "that" :)

```suggestion
This means that once the change happens, all Lucene indexes must be recreated, and data needs to be repopulated.
```

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/impl/LuceneBackendFactory.java
marko-bekhta: mmm OK 😃. I'll drop the warning.  I suppose there's no SUGGESTION logger level we could use 😄. Yeah ... I was just hoping that if we log it, someone might notice it and will decide to prepare for the upgrade before actually upgrading to 7 😃 And  I thought we could get away with a warning since that is an old behaviour we are about to change...

yrodiere (approved):

PR #3562, opened 2023-06-27
Upgrade to ORM 6 in main code

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-3277
https://hibernate.atlassian.net/browse/HSEARCH-4394
https://hibernate.atlassian.net/browse/HSEARCH-4395
https://hibernate.atlassian.net/browse/HSEARCH-4465
https://hibernate.atlassian.net/browse/HSEARCH-4482
https://hibernate.atlassian.net/browse/HSEARCH-4509
https://hibernate.atlassian.net/browse/HSEARCH-4523
https://hibernate.atlassian.net/browse/HSEARCH-4569
https://hibernate.atlassian.net/browse/HSEARCH-4596
https://hibernate.atlassian.net/browse/HSEARCH-4606
https://hibernate.atlassian.net/browse/HSEARCH-4627
https://hibernate.atlassian.net/browse/HSEARCH-4635
https://hibernate.atlassian.net/browse/HSEARCH-4664
https://hibernate.atlassian.net/browse/HSEARCH-4672
https://hibernate.atlassian.net/browse/HSEARCH-4688
https://hibernate.atlassian.net/browse/HSEARCH-4708
https://hibernate.atlassian.net/browse/HSEARCH-4758
https://hibernate.atlassian.net/browse/HSEARCH-4768
https://hibernate.atlassian.net/browse/HSEARCH-4775
https://hibernate.atlassian.net/browse/HSEARCH-4783
https://hibernate.atlassian.net/browse/HSEARCH-4791
https://hibernate.atlassian.net/browse/HSEARCH-4805
https://hibernate.atlassian.net/browse/HSEARCH-4809
https://hibernate.atlassian.net/browse/HSEARCH-4819
https://hibernate.atlassian.net/browse/HSEARCH-4829
https://hibernate.atlassian.net/browse/HSEARCH-4833
https://hibernate.atlassian.net/browse/HSEARCH-4842
https://hibernate.atlassian.net/browse/HSEARCH-4846
https://hibernate.atlassian.net/browse/HSEARCH-4858
https://hibernate.atlassian.net/browse/HSEARCH-4860
https://hibernate.atlassian.net/browse/HSEARCH-4869
https://hibernate.atlassian.net/browse/HSEARCH-4875
https://hibernate.atlassian.net/browse/HSEARCH-4755
https://hibernate.atlassian.net/browse/HSEARCH-4888

😃 

I've amended some commits to address changes from https://github.com/hibernate/hibernate-search/pull/3553
Also:

- there were some formatting / import issues -- I've amended the changes into commits introducing the import/format issue
- Removed the following commit and used HSEARCH-4482 for the one forbidding EM:
Revert "TODOTICKET Forbid Uses of Hibernate ORM 6's org.hibernate.tuple.entity.EntityMetamodel"

- updated the following commits with HSEARCH-4394:
TODOTICKET Disable -orm6 modules
TODOTICKET Remove support for Java EE
TODOFIXME Disable Spring ITs because Spring does not support Jakarta EE yet

Let's see if CI will want to build this 😃 🙈


integrationtest/mapper/orm-spring/pom.xml
```diff
@@ -60,6 +60,11 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-data-jpa</artifactId>
             <exclusions>
+                <!-- Using a newer version of Hibernate ORM, with a different groupID -->
+                <exclusion>
+                    <groupId>org.hibernate</groupId>
+                    <artifactId>hibernate-core</artifactId>
+                </exclusion>
```
yrodiere, line 67: This seems unnecessary since we disabled this module and will only re-enable it when switching to Spring Boot 3.

integrationtest/showcase/library/pom.xml
```diff
@@ -73,6 +77,11 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-data-jpa</artifactId>
             <exclusions>
+                <!-- Using a newer version of Hibernate ORM, with a different groupID -->
+                <exclusion>
+                    <groupId>org.hibernate</groupId>
+                    <artifactId>hibernate-core</artifactId>
+                </exclusion>
```
yrodiere, line 84: This seems unnecessary as we've disabled this module and will only re-enable it when upgrading to Spring Boot 3, which uses Hibernate ORM 6.

mapper/orm-batch-jsr352/core/pom.xml
```diff
@@ -49,6 +49,33 @@
...
+               Execution add-module-infos of goal org.moditect:moditect-maven-plugin:1.0.0.RC2:add-module-info failed: Module jakarta.interceptor not found, required by jakarta.transaction
+         -->
+        <dependency>
+            <groupId>jakarta.enterprise</groupId>
+            <artifactId>jakarta.enterprise.cdi-api</artifactId>
+            <!-- DO NOT REMOVE and DO NOT MANAGE the version of this dependency. See the version property declaration. -->
+            <version>${version.jakarta.enterprise}</version>
+            <scope>provided</scope>
+        </dependency>
+
```
yrodiere: Did you check whether this workaround is still necessary? We've upgraded these dependencies multiple times since the workaround was first introduced.

mapper/orm-coordination-outbox-polling/pom.xml
```diff
@@ -35,6 +35,34 @@
...
+             and that doesn't play well with Moditect, which ends up throwing an exception:
+               Execution add-module-infos of goal org.moditect:moditect-maven-plugin:1.0.0.RC2:add-module-info failed: Module jakarta.interceptor not found, required by jakarta.transaction
+         -->
+        <dependency>
+            <groupId>jakarta.enterprise</groupId>
+            <artifactId>jakarta.enterprise.cdi-api</artifactId>
+            <!-- DO NOT REMOVE and DO NOT MANAGE the version of this dependency. See the version property declaration. -->
+            <version>${version.jakarta.enterprise}</version>
+            <scope>provided</scope>
+        </dependency>
```
yrodiere, line 64: Did you check whether this workaround is still necessary? We've upgraded these dependencies multiple times since the workaround was first introduced.

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/event/impl/OutboxEventLoader.java
```diff
@@ -50,7 +50,7 @@ final class OutboxEventLoader implements ToStringTreeAppendable {
 		// so we can afford to just skip events that are already locked,
 		// and process them later when they are no longer locked.
 		if ( dialect.supportsSkipLocked() ) {
-			lockOptions = new LockOptions( LockMode.PESSIMISTIC_WRITE ).setTimeOut( LockOptions.SKIP_LOCKED );
+			lockOptions = new LockOptions( LockMode.UPGRADE_SKIPLOCKED );
```
yrodiere, line 53: It seems we (I?) forgot to update the comment above:

```
 		// Thus, our last option is to actually enforce locks ahead of time (LockModeType.PESSIMISTIC_WRITE),
		// and to avoid conflicts by simply never working on events that are already locked (LockOptions.SKIP_LOCKED).
```

Should be:

```
		// Thus, our last option is to actually enforce locks ahead of time (UPGRADE),
		// and to avoid conflicts by simply never working on events that are already locked (SKIP_LOCKED):
		// that's what LockMode.UPGRADE_SKIPLOCKED is for.
```

mapper/orm/pom.xml
```diff
@@ -60,6 +60,26 @@
...
+            <groupId>jakarta.xml.bind</groupId>
+            <artifactId>jakarta.xml.bind-api</artifactId>
+            <!-- DO NOT REMOVE and DO NOT MANAGE the version of this dependency. See the version property declaration. -->
+            <version>${version.jakarta.xml.bind}</version>
+            <exclusions>
+                <!-- Also imported by org.glassfish.jaxb:jaxb-core and its transitive dependency org.eclipse.angus:angus-activation -->
+                <exclusion>
+                    <groupId>jakarta.activation</groupId>
+                    <artifactId>jakarta.activation-api</artifactId>
+                </exclusion>
```
yrodiere, line 80: Shouldn't we rather exclude the dependency from jaxb-core?
That's what we did for Jakarta artifacts in Search 6: https://github.com/hibernate/hibernate-search/blob/8a75143253ace18e18e55ac31cddd1c428446983/jakarta/build/parents/public/pom.xml#L49-L79

BTW we can use dependency management in the root POM for this. We just can't use a version in the dependency management entry.

orm6/build/parents/integrationtest/pom.xml
```diff
@@ -82,7 +82,7 @@
                 <exclusions>
                     <!-- We use Jandex 3 -->
                     <exclusion>
-                        <groupId>org.jboss</groupId>
+                        <groupId>io.smallrye</groupId>
                         <artifactId>jandex</artifactId>
```
yrodiere, line 86: This seems wrong, but probably won't matter if you delete the `orm6` directory.

orm6/build/parents/internal/pom.xml
```diff
@@ -82,7 +82,7 @@
                 <exclusions>
                     <!-- We use Jandex 3 -->
                     <exclusion>
-                        <groupId>org.jboss</groupId>
+                        <groupId>io.smallrye</groupId>
                         <artifactId>jandex</artifactId>
```
yrodiere, line 86: This seems wrong, but probably won't matter if you delete the `orm6` directory.

orm6/build/parents/public/pom.xml
```diff
@@ -83,7 +83,7 @@
                 <exclusions>
                     <!-- We use Jandex 3 -->
                     <exclusion>
-                        <groupId>org.jboss</groupId>
+                        <groupId>io.smallrye</groupId>
                         <artifactId>jandex</artifactId>
```
yrodiere, line 87: This seems wrong, but probably won't matter if you delete the `orm6` directory.

pom.xml
```diff
@@ -1306,37 +1238,6 @@
...
-                <artifactId>jaxb-runtime</artifactId>
-                <!-- DO NOT MANAGE the version of this dependency. -->
-                <exclusions>
-                    <!-- Also imported by jakarta.xml.bind-api -->
-                    <exclusion>
-                        <groupId>com.sun.activation</groupId>
-                        <artifactId>jakarta.activation</artifactId>
-                    </exclusion>
-                </exclusions>
-            </dependency>
```
yrodiere: Ah, see. Here. That's where we should solve dependency convergence issues related to jakarta.activation.

pom.xml
```diff
@@ -1676,6 +1577,12 @@
                                     <version>${maven.min.version}</version>
                                 </requireMavenVersion>
                                 <banDuplicatePomDependencyVersions />
+                                <bannedDependencies>
+                                    <excludes>
+                                        <!-- We use Jandex 3 instead: io.smallrye:jandex, io.smallrye:jandex-maven-plugin -->
+                                        <exclude>org.jboss:jandex*</exclude>
+                                    </excludes>
+                                </bannedDependencies>
```
yrodiere, line 1585: I wonder whether we should also add rules for `javax` dependencies... at the very least the most important ones, such as JPA, CDI, transactions, etc. WDYT?

util/internal/integrationtest/mapper/orm/src/main/resources/hibernate.properties
```diff
@@ -25,3 +25,10 @@ hibernate.cache.region.factory_class = org.hibernate.testing.cache.CachingRegion
 
 hibernate.implicit_naming_strategy = component-path
 
+# Temporary workaround for:
+# - https://hibernate.atlassian.net/browse/HHH-15054
+#   (affects org/hibernate/search/test/proxy/ProxyTest)
+# - https://hibernate.atlassian.net/browse/HHH-15058
+#   (affects tests in org/hibernate/search/integrationtest/mapper/orm/automaticindexing/association/bytype/manytomany,
+#   in particular)
+hibernate.mapping.default_list_semantics=bag
```
yrodiere, line 34: I think this is no longer necessary since this is the default since 6.0.0.CR2: https://github.com/hibernate/hibernate-orm/commit/7aaeebe3af8344d9f49718047b0624b6fa0b8ef9

yrodiere (commented):
Thanks! My comments below.

I think you should probably also have a look at `ci/dependency-update/Jenkinsfile` and remove `ci/dependency-update/rules-orm6-in-main-code.xml`, since this won't be needed anymore?

pom.xml
marko-bekhta: Yes, it was there in the root pom. But for some reason, maven was complaining about missing versions on these dependencies, and since there was a comment about not adding a version, I've tried to move it to the ORM mapper module since that's where we'd be adding ORM. It may be that the reason maven complained was that one of these got listed as an explicit dependency in some module (I think it was the property collector one...) I'll try moving it back to the root pom and will let you know how it goes 😃

yrodiere: Ok, thanks.

> I've tried to move it to the ORM mapper module since that's where we'd be adding ORM

FYI there are several other places where we add the ORM dependency independently. Mainly utils modules, I believe, but still, if the dependency management option doesn't work out, you might want to be careful about those.

mapper/orm-coordination-outbox-polling/pom.xml
marko-bekhta: `jakarta.enterprise.cdi-api` is still needed, but `jakarta.interceptor-api` can be removed same as in the other case, but  `jakarta.xml.bind-api` is additionally required. I think it was complaining about this one when we had an xml-bind dependency in orm mapper (the one to address exclusions)

mapper/orm/pom.xml
marko-bekhta: so we are ending up with this:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.3.0:enforce (enforce-dependency-convergence) on project hibernate-search-mapper-orm: 
[ERROR] Rule 0: org.apache.maven.enforcer.rules.dependency.DependencyConvergence failed with message:
[ERROR] Failed while enforcing releasability.
[ERROR] 
[ERROR] Dependency convergence error for jakarta.activation:jakarta.activation-api:jar:2.1.0 paths to dependency are:
[ERROR] +-org.hibernate.search:hibernate-search-mapper-orm:jar:7.0.0-SNAPSHOT
[ERROR]   +-org.hibernate.orm:hibernate-core:jar:6.2.5.Final:compile
[ERROR]     +-jakarta.xml.bind:jakarta.xml.bind-api:jar:4.0.0:runtime
[ERROR]       +-jakarta.activation:jakarta.activation-api:jar:2.1.0:runtime
[ERROR] and
[ERROR] +-org.hibernate.search:hibernate-search-mapper-orm:jar:7.0.0-SNAPSHOT
[ERROR]   +-org.hibernate.orm:hibernate-core:jar:6.2.5.Final:compile
[ERROR]     +-org.glassfish.jaxb:jaxb-runtime:jar:4.0.2:runtime
[ERROR]       +-org.glassfish.jaxb:jaxb-core:jar:4.0.2:runtime
[ERROR]         +-jakarta.activation:jakarta.activation-api:jar:2.1.1:runtime
[ERROR] and
[ERROR] +-org.hibernate.search:hibernate-search-mapper-orm:jar:7.0.0-SNAPSHOT
[ERROR]   +-org.hibernate.orm:hibernate-core:jar:6.2.5.Final:compile
[ERROR]     +-org.glassfish.jaxb:jaxb-runtime:jar:4.0.2:runtime
[ERROR]       +-org.glassfish.jaxb:jaxb-core:jar:4.0.2:runtime
[ERROR]         +-org.eclipse.angus:angus-activation:jar:2.0.0:runtime
[ERROR]           +-jakarta.activation:jakarta.activation-api:jar:2.1.1:runtime
[ERROR] 
```
I thought I'd exclude it from API one since there's a 2.1.0, while the core one is using 2.1.1

yrodiere: I don't know... it seems wrong to me that jaxb-core would override the version of Jakarta dependencies... I'll let you decide, but please do that in the root POM if possible.

BTW, when we will just import BOMs, this whole mess wouldn't be ours to solve. Hopefully we'll be able to do that soon (in another PR, though). There's https://hibernate.atlassian.net/browse/HSEARCH-4699 to use Hibernate ORM's [`hibernate-platform`](https://central.sonatype.com/artifact/org.hibernate.orm/hibernate-platform/6.2.5.Final/overview). For Jakarta there's [this one](https://mvnrepository.com/artifact/jakarta.platform/jakarta.jakartaee-bom) and maybe others, but I'd rather not use it before Hibernate ORM does.

yrodiere:
@marko-bekhta Looks like you used https://hibernate.atlassian.net/browse/HSEARCH-4394 for some commits where you should have used https://hibernate.atlassian.net/browse/HSEARCH-4395 instead

documentation/src/main/asciidoc/migration/index.adoc
```diff
@@ -44,6 +44,13 @@ Hibernate Search {hibernateSearchVersion} now:
 - is based on Lucene 9 for its Lucene backend.
 - requires using Hibernate ORM versions from the 6.2.x family.
 
+When migrating from previous versions of Hibernate Search
+- If you were using Hibernate ORM 6.2 and your corresponding Hibernate Search dependencies included references
+to artifacts ending with -orm6 (e.g. hibernate-search-mapper-orm-orm6) just remove the -orm6
+(e.g. use hibernate-search-mapper-orm instead) when you upgrade to this version of Hibernate Search.
```
yrodiere, line 50:
```suggestion
to artifacts ending with `-orm6` (e.g. `hibernate-search-mapper-orm-orm6`) just remove the `-orm6` suffix
(e.g. use `hibernate-search-mapper-orm` instead) when you upgrade to this version of Hibernate Search.
```

yrodiere (approved):

.jqassistant.yml
yrodiere, line 1: Is there a way to move this file to `build/jqassistant`?

yrodiere:
The PR build passed :tada: 
I just launched a full build just in case

marko-bekhta:
Thanks!!! 😄 🎉 

I found a few more upgrades when I was searching for ticket references in the code -- CDI/Weld/JBeret (and related dependabot updates) but I'll leave them for a follow-up PR

marko-bekhta:
OK, so jdk11 build failed as spring boot wants at min jdk17, I've updated how we include modules in integration pom +  since spring now works on jdk20,21 I've enabled those tests too.
As for the cockroachdb -- it had a DB in dialect class, and dialect is simply CockroachDialect so changed that too

yrodiere (approved):

marko-bekhta:
hmm so it looks like CockroachDB doesn't feel good in this latest build same as it was in https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-3562/22/pipeline/111 .... 

I'll try to run all these tests locally to see if I get the same result

yrodiere:
Looks like it was a transient failure...  I say let's merge because other PRs depending on this one are starting to accumulate, and let's see if we get more failures on main branch builds later.

Merging, thanks a lot!

marko-bekhta:
Thanks for merging 🎉 😃 

So the tests finished locally aaand... 

That's from a local run it is a bit faster than on the CI:
```
INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Hibernate Search ITs - ORM 7.0.0-SNAPSHOT:
[INFO] 
[INFO] Hibernate Search ITs - ORM ......................... SUCCESS [35:50 min]
[INFO] Hibernate Search ITs - ORM - CDI ................... SUCCESS [ 18.646 s]
[INFO] Hibernate Search ITs - ORM - Envers ................ SUCCESS [ 21.769 s]
[INFO] Hibernate Search ITs - ORM - Real backend .......... SUCCESS [01:19 min]
[INFO] Hibernate Search ITs - ORM - Coordination - Outbox Polling SUCCESS [58:32 min]
[INFO] Hibernate Search ITs - Showcase - Library .......... SUCCESS [01:47 min]
[INFO] Hibernate Search Distribution ...................... SUCCESS [ 10.038 s]
[INFO] Hibernate Search Reports ........................... SUCCESS [  0.767 s]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:38 h
```

while from running these on 6.2:

```
[INFO] Reactor Summary for Hibernate Search ITs - ORM 6.2.0-SNAPSHOT:
[INFO] 
[INFO] Hibernate Search ITs - ORM ......................... SUCCESS [23:28 min]
[INFO] Hibernate Search ITs - ORM - CDI ................... SUCCESS [ 22.858 s]
[INFO] Hibernate Search ITs - ORM - Envers ................ SUCCESS [ 24.046 s]
[INFO] Hibernate Search ITs - ORM - Real backend .......... SUCCESS [01:20 min]
[INFO] Hibernate Search ITs - ORM - Coordination - Outbox Polling SUCCESS [39:28 min]
[INFO] Hibernate Search ITs - Showcase - Library .......... SUCCESS [01:49 min]
[INFO] Hibernate Search Distribution ...................... SUCCESS [ 21.891 s]
[INFO] Hibernate Search Reports ........................... SUCCESS [  1.008 s]
[INFO] BUILD SUCCESS
[INFO] Total time:  01:07 h
```

for 6.2 we have those tests for orm6 artefacts living behind the `if ( !buildEnv.slow ) {` check .... Could it be that ORM 6 performs slower than ORM 5 did in this case 😨😃?
The most time increase comes from tests from `org.hibernate.search.integrationtest.mapper.orm.automaticindexing.association.bytype` (e.g. https://ci.hibernate.org/job/hibernate-search/job/PR-3562/26/testReport/org.hibernate.search.integrationtest.mapper.orm.automaticindexing.association.bytype.onetomany/) . Wasn't there something you had to do for those to limit some fetching (or something) not to get too many joins in Mysql?

yrodiere:
> Could it be that ORM 6 performs slower than ORM 5 did in this case?

It could be a performance regression, though I wouldn't assume that just now. You did upgrade to a newer version of CockroachDB in Search 7.0, didn't you?

We might not be using the latest version of CockroachDB, or not be configuring CockroachDB correctly, or not be configuring ORM right. Maybe there are adjustments to be made for tests? ORM's own tests would be a good source of information).

BTW CockroachDB tests in ORM are rather slow as well.

>  The most time increase comes from tests from org.hibernate.search.integrationtest.mapper.orm.automaticindexing.association.bytype (e.g. https://ci.hibernate.org/job/hibernate-search/job/PR-3562/26/testReport/org.hibernate.search.integrationtest.mapper.orm.automaticindexing.association.bytype.onetomany/)

These tests are probably the most resource-intensive (lots of test methods), so that doesn't seem particularly worrying.

FWIW, in the past most of the slowness of CockroachDB came from creating and dropping schemas. There are all sorts of constraints that make that slow in CockroachDB, which is not a big deal in production but really is in tests. I think there were some talks to provide some options to make those faster using some sort of "testing mode" in CockroachDB, but I'm not sure where those talks went.

> Wasn't there something you had to do for those to limit some fetching (or something) not to get too many joins in Mysql?

I did, but those were changes in the mapping (so they affect all DBs) and they are merged now. Proof is, mysql/mariadb tests do pass.

PR #3564, opened 2023-06-29
Update XML schema of JQAssistant rules

yrodiere:
Only impacts the build, no Jira issue.


marko-bekhta (approved):
Nice 😃 
There's also a new major version out now it seems: https://repo1.maven.org/maven2/com/buschmais/jqassistant/jqassistant-maven-plugin/2.0.3/

maybe we'll update some time in the future 😃

PR #3572, opened 2023-07-03
HSEARCH-4890 Adapt calls to RootGraph#appliesTo() to Hibernate ORM 6.3

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4890

~~Bassed on #3562 , which must be merged first.~~ => Done

### DIFF
--- a/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/loading/impl/MutableEntityLoadingOptions.java
+++ b/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/loading/impl/MutableEntityLoadingOptions.java
@@ -10,6 +10,7 @@ import java.util.ArrayList;
 import java.util.List;
 
 import org.hibernate.metamodel.mapping.EntityMappingType;
+import org.hibernate.metamodel.model.domain.EntityDomainType;
 import org.hibernate.search.util.common.impl.Contracts;
 
 public class MutableEntityLoadingOptions {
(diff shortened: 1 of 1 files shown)


mapper/orm/src/main/java/org/hibernate/search/mapper/orm/loading/impl/MutableEntityLoadingOptions.java
```diff
@@ -31,14 +32,19 @@ public void fetchSize(int fetchSize) {
...
-		String hibernateOrmEntityName = entityMappingType.getEntityName();
-		for ( EntityGraphHint<?> entityGraphHint : entityGraphHints ) {
-			if ( entityGraphHint.graph.appliesTo( hibernateOrmEntityName ) ) {
-				return entityGraphHint;
+		EntityMappingType testedType = entityMappingType;
+		while ( testedType != null ) {
+			for ( EntityGraphHint<?> entityGraphHint : entityGraphHints ) {
+				// This cast is fine because a RootGraph always applies to an entity type
+				EntityDomainType<?> graphedType = (EntityDomainType<?>) entityGraphHint.graph.getGraphedType();
+				if ( graphedType.getHibernateEntityName().equals( testedType.getEntityName() ) ) {
```
marko-bekhta, line 43: seems like `graphedType.getTypeName()` might have the same behaviour

marko-bekhta (approved):
🎉 😃 
All seems good!
just had a question on that `getHibernateEntityName` method

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/loading/impl/MutableEntityLoadingOptions.java
yrodiere: It seems so, currently... but then the javadoc of `ManagedDomainType#getTypeName` is wrong, since one may customize the (Hibernate ORM) entity name to something completely different than the FQCN, even for non-dynamic-map entities...
I'd rather not rely on that until we clarify.

PR #3578, opened 2023-07-03
Add Hibernate GitHub Bot ignore rules

marko-bekhta:
Since it should support multiple rules, I thought let's keep the regex simple 😃


yrodiere:
Thanks!

PR #3586, opened 2023-07-05
HSEARCH-5029 Simplify entity removal in ReusableOrmSetupHolder

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5029

This won't solve the problem, but since I've been looking at this part of the code and noticed that both ORM tickets are now resolved, we can get rid of the if and make things a bit faster. I've tried to run a test on a cockroachdb: without changes  ~3min40sec, with changes ~3min.


yrodiere (approved):

marko-bekhta:
yeah ... opened https://hibernate.atlassian.net/browse/HHH-16897 😢 and looking into one more category of "failures"

marko-bekhta:
The other errors were related to the cases when we were trying to delete entities that had self-references (parent-child), leading to constraint violation errors. I suppose there's not much we can do, unless we make a cleanup ourselves in the tests before falling back to the helper delete logic.
I've converted this PR to a draft to wait for the fix in ORM.

yrodiere:
@marko-bekhta Should we close this or do you want to open a Jira issue and get this merged in 7.1?

yrodiere (approved):

PR #3587, opened 2023-07-05
HSEARCH-4894 Use in-memory storage for CockroachDB

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4894

So using in-memory storage seems to help a bit with the schema operations (while before the change (for a specific test file), it was taking ~10 sec to create and drop the schema, now it is ~3sec)  See https://www.cockroachlabs.com/docs/stable/local-testing.html . Applying other options suggested under that link didn't do much so I haven't added those in this PR. 
Also, while at it, I've upgraded the DB version. It might be just me 😄 but it seemed that tests run a tod faster on this version 😆

### DIFF
--- a/build/parents/integrationtest/pom.xml
+++ b/build/parents/integrationtest/pom.xml
@@ -361,6 +361,15 @@
                                         <exec>
                                             <arg>start-single-node</arg>
                                             <arg>--insecure</arg>
+                                            <!-- Using in-memory storage to make things a bit faster -->
+                                            <arg>--store=type=mem,size=.5</arg>
+                                            <!--
+                                                See https://www.cockroachlabs.com/docs/v23.1/transaction-retry-error-reference.html#readwithinuncertaintyintervalerror.
+                                                Increasing this value will increase the time to recovery of failures as well as the frequency of uncertainty-based read restarts.
+                                                Default value is 500ms.
+                                            -->
+                                            <arg>--max-offset=100ms</arg>
+                                            <arg>--log=sinks: {stderr: {channels: [OPS]}}</arg>
                                         </exec>
                                     </cmd>
                                     <ulimits>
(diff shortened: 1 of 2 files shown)


yrodiere (approved):

marko-bekhta:
Surprisingly I cannot reproduce the `OutboxPollingAutomaticIndexingConcurrencyIT#noDeadlock` failure locally 🥲

yrodiere:
> Surprisingly I cannot reproduce the `OutboxPollingAutomaticIndexingConcurrencyIT#noDeadlock` failure locally smiling_face_with_tear

https://www.cockroachlabs.com/docs/v23.1/transaction-retry-error-reference.html#readwithinuncertaintyintervalerror

I suppose the "gateways" mentioned in this thing are the CockroachDB JDBC drivers? Since we only have one CockroachDB container, I don't see how it could be anything else.

I think we'll have to implement better client-side retries at some point, because the current one may flood logs with (expected) errors in such cases...

In the meantime, there's always this:

> If you [trust your clocks](https://www.cockroachlabs.com/docs/v23.1/operational-faqs#what-happens-when-node-clocks-are-not-properly-synchronized), you can try lowering the [--max-offset option to cockroach start](https://www.cockroachlabs.com/docs/v23.1/cockroach-start#flags), which provides an upper limit on how long a transaction can continue to restart due to uncertainty.

All clients are running on the same JVM, so that should be safe... I think?

yrodiere:
Damn, still 1 hour and 30 minutes... Well, at least it's lower than our timeout.
Merging, thanks!

marko-bekhta:
aaaand we had another build failed https://ci.hibernate.org/job/hibernate-search/job/main/753/ for the same `OutboxPollingAutomaticIndexingConcurrencyIT#noDeadlock` -_- so changing that max offset didn't help

PR #3588, opened 2023-07-06
HSEARCH-4861 Add public BOM

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4896

Since I didn't want to move the properties from the root pom as those are grouped nicely there, I've ended up creating two BOMs with root as a parent. 
I've used the quarkus plugin to generate the effective public BOM and filtered out a few things from it that got into it only for build or dependency convergence resolution reasons. 
Let's see if this builds on CI 🙈 😄

### DIFF
--- a/bom/build/pom.xml
+++ b/bom/build/pom.xml
@@ -18,6 +18,11 @@
     </description>
     <packaging>pom</packaging>
 
+    <properties>
+        <!-- Public modules must be signed. -->
+        <release.gpg.signing.skip>false</release.gpg.signing.skip>
+    </properties>
+
     <dependencyManagement>
         <dependencies>
             <dependency>
--- a/bom/build/pom.xml
+++ b/bom/build/pom.xml
@@ -1,0 +1,535 @@
+<?xml version="1.0" encoding="utf-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.hibernate.search</groupId>
+        <artifactId>hibernate-search-parent</artifactId>
+        <version>7.0.0-SNAPSHOT</version>
+        <relativePath>../..</relativePath>
+    </parent>
+
+    <artifactId>hibernate-search-build-bom</artifactId>
+    <name>Hibernate Search - Build BOM</name>
+    <description>
+        Provides dependency management for all dependencies required during the build time.
+        Imports the ones from the public BOM.
+    </description>
+    <packaging>pom</packaging>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.hibernate.search</groupId>
+                <artifactId>hibernate-search-bom</artifactId>
+                <version>${project.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+
+            <!-- Remaining Hibernate Search dependencies -->
+            <dependency>
+                <groupId>org.hibernate.search</groupId>
+                <artifactId>hibernate-search-configuration-properties-collector</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.hibernate.search</groupId>
+                <artifactId>hibernate-search-documentation</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.hibernate.search</groupId>
+                <artifactId>hibernate-search-documentation</artifactId>
+                <type>pom</type>
+                <version>${project.version}</version>
+            </dependency>
+
+            <!-- Import Hibernate ORM BOM -->
+            <dependency>
+                <groupId>org.hibernate.orm</groupId>
+                <artifactId>hibernate-platform</artifactId>
+                <version>${version.org.hibernate.orm}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+            <!-- Imported dependencies that we want to override -->
+            <dependency>
+                <groupId>org.jboss.logging</groupId>
+                <artifactId>jboss-logging</artifactId>
+                <version>${version.org.jboss.logging.jboss-logging}</version>
+            </dependency>
+
+            <!-- Other public dependencies -->
+            <!-- Elasticsearch backend -->
+            <dependency>
+                <groupId>org.elasticsearch.client</groupId>
+                <artifactId>elasticsearch-rest-client</artifactId>
+                <version>${version.org.elasticsearch.client}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.elasticsearch.client</groupId>
+                <artifactId>elasticsearch-rest-client-sniffer</artifactId>
+                <version>${version.org.elasticsearch.client}</version>
+            </dependency>
+            <dependency>
+                <groupId>com.google.code.gson</groupId>
+                <artifactId>gson</artifactId>
+                <version>${version.com.google.code.gson}</version>
+            </dependency>
+            <dependency>
+                <groupId>software.amazon.awssdk</groupId>
+                <artifactId>auth</artifactId>
+                <version>${version.software.amazon.awssdk}</version>
+            </dependency>
+
+            <!-- Lucene backend -->
+            <dependency>
+                <groupId>org.apache.lucene</groupId>
+                <artifactId>lucene-core</artifactId>
+                <version>${version.org.apache.lucene}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.lucene</groupId>
+                <artifactId>lucene-analysis-common</artifactId>
+                <version>${version.org.apache.lucene}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.lucene</groupId>
+                <artifactId>lucene-queryparser</artifactId>
+                <version>${version.org.apache.lucene}</version>
+                <exclusions>
+                    <!-- This artifact contains unstable, experimental code that we cannot reasonably rely on. -->
+                    <exclusion>
+                        <groupId>org.apache.lucene</groupId>
+                        <artifactId>lucene-sandbox</artifactId>
+                    </exclusion>
+                </exclusions>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.lucene</groupId>
+                <artifactId>lucene-join</artifactId>
+                <version>${version.org.apache.lucene}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.lucene</groupId>
+                <artifactId>lucene-facet</artifactId>
+                <version>${version.org.apache.lucene}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.lucene</groupId>
+                <artifactId>lucene-highlighter</artifactId>
+                <version>${version.org.apache.lucene}</version>
+            </dependency>
+            <dependency>
+                <groupId>com.carrotsearch</groupId>
+                <artifactId>hppc</artifactId>
+                <version>${version.com.carrotsearch.hppc}</version>
+            </dependency>
+
+            <!-- Apache Avro used to serialize/deserialize objects in outbox polling coordination -->
+            <dependency>
+                <groupId>org.apache.avro</groupId>
+                <artifactId>avro</artifactId>
+                <version>${version.org.apache.avro}</version>
+            </dependency>
+
+            <!-- JakartaEE dependencies -->
+            <dependency>
+                <groupId>jakarta.batch</groupId>
+                <artifactId>jakarta.batch-api</artifactId>
+                <version>${version.jakarta.batch}</version>
+                <scope>provided</scope>
+            </dependency>
+
+            <!-- JBeret used by batch-jsr352 jberet module -->
+            <dependency>
+                <groupId>org.jberet</groupId>
+                <artifactId>jberet-core</artifactId>
+                <version>${version.org.jberet}</version>
+                <exclusions>
+                    <!-- These dependencies are marked as "provided" but will never be there at runtime.
+                         This confuses maven-enforcer-plugin, so we'll exclude them. -->
+                    <exclusion>
+                        <groupId>org.jboss.logging</groupId>
+                        <artifactId>jboss-logging-processor</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>org.infinispan</groupId>
+                        <artifactId>infinispan-core</artifactId>
+                    </exclusion>
+                    <exclusion>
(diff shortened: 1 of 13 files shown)


bom/internal/pom.xml
```diff
@@ -0,0 +1,305 @@
...
+                <artifactId>hibernate-search-configuration-properties-collector</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.hibernate.search</groupId>
+                <artifactId>hibernate-search-documentation</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+
+            <!-- Hibernate Search Test utils and Tests -->
```
yrodiere, line 38: As far as I can tell there are only tests here.
You could probably move all the **integration** test utils here.

bom/internal/pom.xml
```diff
@@ -0,0 +1,305 @@
...
+        <groupId>org.hibernate.search</groupId>
+        <artifactId>hibernate-search-parent</artifactId>
+        <version>7.0.0-SNAPSHOT</version>
+        <relativePath>../..</relativePath>
+    </parent>
+
+    <artifactId>hibernate-search-internal-bom</artifactId>
+    <name>Hibernate Search - BOM Internal</name>
+    <packaging>pom</packaging>
+
```
yrodiere, line 16: You need to add some properties to skip publishing this BOM, I think.
We probably have to publish `public` and `public-build`, but not this one.

bom/internal/pom.xml
```diff
@@ -0,0 +1,305 @@
...
+            <dependency>
+                <groupId>com.fasterxml.jackson.core</groupId>
+                <artifactId>jackson-annotations</artifactId>
+                <version>${version.com.fasterxml.jackson}</version>
+            </dependency>
+            <dependency>
+                <groupId>com.fasterxml.jackson.core</groupId>
+                <artifactId>jackson-databind</artifactId>
+                <version>${version.com.fasterxml.jackson.databind}</version>
+            </dependency>
```
yrodiere, line 125: If it's used by the Elasticsearch REST client, we need this in the public BOM.

Maybe we should have some automation to check that all non-test dependencies of public artifacts are managed in the BOM... because they must be.

However, the only way I can think to enforce that would be to [ban transitive dependencies](https://maven.apache.org/enforcer/enforcer-rules/banTransitiveDependencies.html) in public modules, which would force us to have explicit dependencies for everything in our public modules, which would in turn force us to have a version for every one of those dependencies. But damn, that's yucky. Definitely not something I'd like to see in Hibernate Search.

I'm open to ideas... :/

bom/internal/pom.xml
```diff
@@ -0,0 +1,305 @@
...
+
+            <!-- Test -->
+            <dependency>
+                <groupId>junit</groupId>
+                <artifactId>junit</artifactId>
+                <version>${version.junit}</version>
+            </dependency>
```
yrodiere, line 132: I think JUnit is used in the build of public modules as well... You probably need to include it in the `public-build` BOM?
Same with many dependencies below. In short anything that is used in unit (surefire) tests must be in `public-build`; anything that is used only in integration tests is safe to leave here.

This does sound complicated though... I wonder if it wouldn't be simpler to just have a public BOM + a build BOM, for both public and internal modules.
Or a public BOM + the rest of dependency management in one of the parent POMs (the root POM or the public parent POM).

Before you start frowning at me, I'll point out this is _not_ the same thing as your original proposal ;D

marko-bekhta: 😏 haha
yeah, that does make sense (about junit), on the other hand I'm surprised it didn't fail without it and all other dependencies... 
I've started with keeping dependencies in the root and adding the public bom. But if we want that public bom to have a parent of root or any other parent, it will pick up all of the remaining dependencies from root ...
If we don't add a parent, then we'd need to move the version properties to this bom (which I didn't want to do to keep things grouped in the root)...
then I added a build bom (it was called internal 😃) and used that to import in all 3 parents (public, internal, integrationtest) and it worked. But then I thought that we probably don't want to import an "internal bom", that most likely will not be published, into a public parent ... so that's what led me to create that weird bom with exclusions....

Nooow going back to it -- having that build/internal bom and use it for all parents would work, but then we'd just need to publish it too?

yrodiere: I suspect we'll always need to publish whatever manages dependencies of our public artifacts, including test ones, otherwise the published POM for those artifacts won't be complete and this will lead to problems for consumers.

Unless we do like gradle and clean up the POMs before publishing them, but I don't know how to do that in Maven and don't want to make the setup too complicated...

yrodiere: BTW we currently do publish all our dependency management, including for tests and stuff: it's in our root POM, which does get published.

So yeah, I'd say either:

* Put the "public" dependencies in the public BOM and the other ones in a "build" BOM.
* OR put the "public" dependencies in the public BOM and the other ones in the `parents/public` parent POM (and don't use that parent for the BOM but the root one instead, which will require *some* duplication I suppose)

integrationtest/showcase/library/pom.xml
```diff
@@ -37,6 +37,20 @@
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+            <!--
+                NOTE: spring brings its versions of the elasticsearch client.
+                We want to test on the version we are using hence we override it here:
+             -->
```
yrodiere, line 43: Shouldn't we just override the property `elasticsearch.version`, like we do with `hibernate.version` above, and like we [recommend in the documentation](https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#compatibility-framework-spring-boot-dependency-versions)?

bom/internal/pom.xml
marko-bekhta: After some back and forth, I've added these to the public BOM, and updated the comments next to libs/version
It looks like core is used by es client sniffer, while all three are used by avro and wiremock. Since avro is also in a list of public dependencies, I suppose we cannot use the trick to add these dependencies into specific modules using avro/wiremock directly without having them in the dependency management....

> Maybe we should have some automation to check that all non-test dependencies of public artifacts are managed in the BOM... because they must be.

There's this - https://maven.apache.org/enforcer/enforcer-api/writing-a-custom-rule.html I'll take a look if we could create some rule that would check for it, but that's probably for a different PR 😃

yrodiere: > There's this - https://maven.apache.org/enforcer/enforcer-api/writing-a-custom-rule.html I'll take a look if we could create some rule that would check for it, but that's probably for a different PR smiley

Thanks! Agreed, this is out of scope.

integrationtest/showcase/library/pom.xml
marko-bekhta: Good point 😃 
It looks like they've changed the property name in the new version of their bom, so I've updated the doc and added a tip.

distribution/src/main/assembly/dist.xml
```diff
@@ -225,6 +225,13 @@
             <directory>../documentation/target/dist/</directory>
             <outputDirectory>docs/</outputDirectory>
         </fileSet>
+        <fileSet>
+            <directory>../bom/public/target/</directory>
+            <outputDirectory>dist/bom/</outputDirectory>
+            <includes>
+                <include>pom.xml</include>
+            </includes>
+        </fileSet>
```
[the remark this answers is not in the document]
marko-bekhta: 🤷🏻‍♂️ me neither, I just thought that as it's a "public artifact" -- let's include it as well, let people at least look at a pom if they don't use them 😄 
Anyway 😃 I reverted this change.

bom/build/pom.xml
```diff
@@ -0,0 +1,406 @@
...
+    <name>Hibernate Search - Build BOM</name>
+    <description>
+        Provides dependency management for all dependencies required during the build time.
+        Imports the ones from the public BOM.
+    </description>
+    <packaging>pom</packaging>
+
+    <properties>
+        <!-- Do not publish internal modules -->
+        <skipNexusStagingDeployMojo>true</skipNexusStagingDeployMojo>
```
yrodiere, line 23: Wait, this one is not internal though, it's referenced from public modules so we do need to publish it?

documentation/src/main/asciidoc/public/reference/_compatibility.adoc
```diff
@@ -128,6 +128,13 @@ there will probably be a mismatch,
 and generally overriding Spring Boot's version to match Hibernate Search's version will work fine.
 ====
 
+[TIP]
+====
+If, after setting the properties above,
+you still are getting the same version of the libraries --
+check if property names in the Spring Boot's BOM have changed, and if so -- use the new property name.
```
yrodiere, line 135:
```suggestion
you still are getting the same version of the libraries,
check if property names in the Spring Boot's BOM have changed, and if so use the new property name.
```

integrationtest/showcase/library/pom.xml
yrodiere: Looks like you didn't remove dependency management?

marko-bekhta: hmm, so I've just tested it once again, and seems like this approach doesn't work for us:
```
[INFO] --- maven-dependency-plugin:3.6.0:tree (default-cli) @ hibernate-search-integrationtest-showcase-library ---
[INFO] org.hibernate.search:hibernate-search-integrationtest-showcase-library:jar:7.0.0-SNAPSHOT
[INFO] +- org.hibernate.search:hibernate-search-mapper-orm:jar:7.0.0-SNAPSHOT:compile
[INFO] |  +- org.hibernate.search:hibernate-search-engine:jar:7.0.0-SNAPSHOT:compile
[INFO] |  +- org.hibernate.search:hibernate-search-mapper-pojo-base:jar:7.0.0-SNAPSHOT:compile
[INFO] |  +- org.hibernate.search:hibernate-search-util-common:jar:7.0.0-SNAPSHOT:compile
[INFO] |  +- org.hibernate.orm:hibernate-core:jar:6.2.5.Final:compile
[INFO] |  \- org.hibernate.common:hibernate-commons-annotations:jar:6.0.6.Final:compile
[INFO] +- org.hibernate.search:hibernate-search-backend-lucene:jar:7.0.0-SNAPSHOT:compile
[INFO] +- org.hibernate.search:hibernate-search-backend-elasticsearch:jar:7.0.0-SNAPSHOT:compile
[INFO] |  +- org.elasticsearch.client:elasticsearch-rest-client:jar:8.7.1:compile
[INFO] |  \- org.elasticsearch.client:elasticsearch-rest-client-sniffer:jar:8.7.1:compile
[INFO] +- org.hibernate.search:hibernate-search-backend-elasticsearch-aws:jar:7.0.0-SNAPSHOT:test
[INFO] \- org.hibernate.search:hibernate-search-util-internal-integrationtest-mapper-orm:jar:7.0.0-SNAPSHOT:test
[INFO]    +- org.hibernate.search:hibernate-search-util-internal-test-orm:jar:7.0.0-SNAPSHOT:test
[INFO]    |  \- org.hibernate.search:hibernate-search-util-internal-test-common:jar:7.0.0-SNAPSHOT:test
[INFO]    +- org.hibernate.search:hibernate-search-util-internal-integrationtest-common:jar:7.0.0-SNAPSHOT:test
[INFO]    \- org.hibernate.orm:hibernate-testing:jar:6.2.5.Final:test
[INFO]       \- org.hibernate.orm:hibernate-community-dialects:jar:6.2.5.Final:test
```
So I've tried to do it on a spring boot project and it worked  fine there ... Looking into it -- I suppose the reason is because we are importing the pom (so I guess the versions/properties at that point are resolved?) while in case of a boot project these dependencies/properties/versions get to the user pom through parents ...

So I'm not sure... we could leave it as is and use whatever the versions are in boot, or force it with an override like the one I haven't removed 🙈 😃

yrodiere: Ah, it's possible that overriding properties only works when using Spring Boot's parent POM, and not when using the BOM.

If true, we'd need to add dependency management for ORM as well...

Or actually, we could just import the Hibernate Search public BOM just after Spring Boot's?

Could you look into this and update the documentation accordingly?

bom/public/pom.xml
```diff
@@ -0,0 +1,254 @@
...
+            </dependency>
+
+            <!-- Import Hibernate ORM BOM -->
+            <dependency>
+                <groupId>org.hibernate.orm</groupId>
+                <artifactId>hibernate-platform</artifactId>
+                <version>${version.org.hibernate.orm}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
```
gsmet, line 98: I'm late to the party but: the best practice for a low level library BOM is to only include in the BOM the artifacts you produce and not the artifacts that you depend on.
The case of Quarkus is a bit different because we want to provide a cohesive platform.

I strongly suggest that, In the case of Hibernate Search, the BOM should only contain the Hibernate Search artifacts and that's it. Also I don't think you should use the Quarkus plugin to generate a BOM.

Last but not least, check  your BOM with `mvn help:effective-pom`  at the end to make sure it's lean and that it only contains the Hibernate Search artifacts.

I don't remember all the research we did when we worked on that but if you need further details, @aloubyansky might be able to help.

gsmet (changes requested):

bom/public/pom.xml
yrodiere: Okay... I think it makes sense. At the very least it's a more conservative approach.

So, my description on the ticket was wrong. Sorry about that @marko-bekhta .

In practice I suppose this means moving all non-Hibernate-Search artifacts (even the Hibernate ORM BOM) out of the `bom/public/pom.xml` into `bom/build/pom.xml`, and changing a bit the documentation since using the Hibernate Search BOM won't solve any of the problems mentioned there anymore.

Does it make sense @marko-bekhta?

marko-bekhta: Hey, no worries 😃 
Yes, it makes sense. As for the docs, I assume we could just "solve it" by importing two BOMs ORM/Search ones instead. I'll test if that will work 😃 

> Also I don't think you should use the Quarkus plugin to generate a BOM.

As for the plugin, we had some artifacts to exclude so that was a useful tool. With ORM BOM gone, we wouldn't need to exclude anything. From what I've seen the plugin creates a resulting pom similar to what `help:effective-pom` produces, resolving all the properties and creating a single XML. Otherwise, we'd have a BOM referencing the root Search pom. I liked the "clean" one more 😃, but if the one with a root parent and unresolved properties is OK I think we can remove the plugin entirely.

aloubyansky: > Also I don't think you should use the Quarkus plugin to generate a BOM.

Flattening is a good idea. Typically, people use https://www.mojohaus.org/flatten-maven-plugin for this.
You could also use the Quarkus platform plugin, if you like, which was specialized for Quarkus platform BOM flattening. In case it's relevant, it allows configuring a few options such as `excludeArtifactKeys` (exclude some constraints from the original BOM, possibly inherited from an external POM) and `filterInvalidConstraints`, which, if enabled, will validate each constraint in the BOM and exclude those that don't seem to exist.

yrodiere (approved):
LGTM, thanks @marko-bekhta!

yrodiere:
Ah, you might want to update the getting started guides to use the BOMs. Feel free to merge that without a review.

PR #3593, opened 2023-07-10
HSEARCH-4742 Add an option to add property source to Standalone POJO mapper builder

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4742

### DIFF
--- a/documentation/src/main/asciidoc/public/reference/_mapper-pojo-standalone.adoc
+++ b/documentation/src/main/asciidoc/public/reference/_mapper-pojo-standalone.adoc
@@ -341,6 +341,22 @@ so its behavior is roughly similar to that described in <<coordination-none>>,
 except entity data extracting happens on session closing instead of happening on Hibernate ORM session flushes,
 and indexing happens immediately after that instead of happening on transaction commit.
 
+[[mapper-pojo-standalone-configuration-from-property-file]]
+== Reading configuration properties from a file
+
+The Standalone POJO Mapper `SearchMappingBuilder` can also take properties from a `Reader` compatible with `java.util.Properties#load(Reader)`:
+
+.Loading configuration properties from a file using a `Reader`
+====
+[source, JAVA, indent=0]
+----
+include::{sourcedir}/org/hibernate/search/documentation/mapper/pojo/standalone/entrypoints/StandalonePojoReadPropertiesFromFileIT.java[tags=setup]
+----
+<1> Get a reader representing a property file with configuration properties.
+<2> Start configuring the Standalone POJO Mapper.
+<3> Pass the property file reader to the builder.
+====
+
 [[mapper-pojo-standalone-configuration-other]]
 == Other configuration
 
--- a/documentation/src/test/java/org/hibernate/search/documentation/mapper/pojo/standalone/entrypoints/StandalonePojoReadPropertiesFromFileIT.java
+++ b/documentation/src/test/java/org/hibernate/search/documentation/mapper/pojo/standalone/entrypoints/StandalonePojoReadPropertiesFromFileIT.java
@@ -1,0 +1,87 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.documentation.mapper.pojo.standalone.entrypoints;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.nio.charset.StandardCharsets;
+
+import org.hibernate.search.documentation.testsupport.BackendConfigurations;
+import org.hibernate.search.documentation.testsupport.TestConfiguration;
+import org.hibernate.search.mapper.pojo.standalone.entity.SearchIndexedEntity;
+import org.hibernate.search.mapper.pojo.standalone.mapping.CloseableSearchMapping;
+import org.hibernate.search.mapper.pojo.standalone.mapping.SearchMapping;
+import org.hibernate.search.mapper.pojo.standalone.session.SearchSession;
+import org.hibernate.search.util.impl.integrationtest.common.TestConfigurationProvider;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class StandalonePojoReadPropertiesFromFileIT {
+
+	private CloseableSearchMapping theSearchMapping;
+
+	@Rule
+	public TestConfigurationProvider configurationProvider = new TestConfigurationProvider();
+
+	@Before
+	public void setup() throws IOException {
+		// tag::setup[]
+		try (
+				Reader propertyFileReader = /* ... */ // <1>
+						// end::setup[]
+						new InputStreamReader(
+								getClass().getClassLoader().getResourceAsStream( "configuration/standalone-test.properties" ),
+								StandardCharsets.UTF_8
+						)
+		// tag::setup[]
+		) {
+			CloseableSearchMapping searchMapping = SearchMapping.builder() // <2>
+					.properties( propertyFileReader ) // <3>
+					// end::setup[]
+					.properties( TestConfiguration.standalonePojoMapperProperties(
+							configurationProvider,
+							BackendConfigurations.simple()
+					) )
+					// tag::setup[]
+					.build();
+			// end::setup[]
+			this.theSearchMapping = searchMapping;
+			// tag::setup[]
+		}
+		// end::setup[]
+	}
+
+	@After
+	public void cleanup() {
+		if ( theSearchMapping != null ) {
+			CloseableSearchMapping searchMapping = theSearchMapping;
+			searchMapping.close();
+		}
+	}
+
+	@Test
+	public void mappingContainsExpectedEntities() {
+		assertThat( theSearchMapping.allIndexedEntities() )
+				.extracting( SearchIndexedEntity::name )
+				.contains( "Book", "Associate", "Manager" );
+	}
+
+	@Test
+	public void searchSession() {
+		SearchMapping searchMapping = theSearchMapping;
+		try ( SearchSession searchSession = searchMapping.createSession() ) {
+			assertThat( searchSession ).isNotNull();
+			assertThat( searchSession.isOpen() ).isTrue();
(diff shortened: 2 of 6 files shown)


engine/src/main/java/org/hibernate/search/engine/cfg/impl/PropertiesConfigurationPropertySource.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.Properties;
+import java.util.stream.Collectors;
+
+public class PropertiesConfigurationPropertySource extends MapConfigurationPropertySource {
+
+	public PropertiesConfigurationPropertySource(InputStream propertyStream) throws IOException {
+		super( buildProperties( propertyStream ) );
+	}
```
yrodiere, line 20: If everything is going to be implemented in a static method whose result gets passed to the constructor, you don't really need a dedicated class. A static method in `MapConfigurationPropertySource` (e.g. `fromProperties`) would suffice.

documentation/src/main/asciidoc/public/reference/_mapper-pojo-standalone.adoc
```diff
@@ -341,6 +341,23 @@ so its behavior is roughly similar to that described in <<coordination-none>>,
...
+The Standalone POJO Mapper `SearchMappingBuilder` can also take properties from a `ConfigurationPropertySource`,
+in particular that source can be based on a property file:
+
+.Loading configuration properties from a file
+====
+[source, JAVA, indent=0]
+----
+include::{sourcedir}/org/hibernate/search/documentation/mapper/pojo/standalone/entrypoints/StandalonePojoReadPropertiesFromFileIT.java[tags=setup]
+----
+<1> Get an input stream representing a property file with configuration properties.
```
yrodiere, line 356: To be updated if you switch to `Reader`.

documentation/src/test/java/org/hibernate/search/documentation/mapper/pojo/standalone/entrypoints/StandalonePojoReadPropertiesFromFileIT.java
```diff
@@ -0,0 +1,81 @@
...
+public class StandalonePojoReadPropertiesFromFileIT {
+
+	private CloseableSearchMapping theSearchMapping;
+
+	@Rule
+	public TestConfigurationProvider configurationProvider = new TestConfigurationProvider();
+
+	@Before
+	public void setup() throws IOException {
+		try ( // tag::setup[]
```
yrodiere, line 37: I'd include the `try` in the example, so that indentation won't seem wrong.

Ideally you'd also move the declaration of `CloseableSearchMapping searchMapping` out of the `try` block, to show that the input stream doesn't need to remain open forever.

documentation/src/test/java/org/hibernate/search/documentation/mapper/pojo/standalone/entrypoints/StandalonePojoReadPropertiesFromFileIT.java
```diff
@@ -0,0 +1,81 @@
...
+				InputStream propertyFile = /* ... */ // <1>
+						// end::setup[]
+						getClass().getClassLoader().getResourceAsStream( "configuration/standalone-test.properties" ) ) {
+
+
+			// tag::setup[]
+			CloseableSearchMapping searchMapping = SearchMapping.builder() // <2>
+					.properties( AllAwareConfigurationPropertySource.fromPropertyFile( propertyFile ) ) // <3>
```
yrodiere, line 45: Hm. Annoying that users have to rely on SPIs.

I'd suggest either adding a `.properties()` method that accepts the inputstream/reader directly, or moving `AllAwareConfigurationPropertySource` to API and moving its static methods to `ConfigurationPropertySource` (but the methods would still return a `AllAwareConfigurationPropertySource`). The first solution seems simpler.

yrodiere (changes requested):
Thanks. A few comments below.

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/bootstrap/ConfiguringPropertiesInSearchMappingBuilderIT.java
```diff
@@ -0,0 +1,123 @@
...
+			propertyFile = AllAwareConfigurationPropertySource.fromPropertyFile( properties );
+		}
+
+		SearchMappingBuilder builder = SearchMapping.builder( MethodHandles.lookup() )
+				.property( StandalonePojoMapperSpiSettings.BEAN_PROVIDER, new ForbiddenBeanProvider() )
+				.properties( propertyFile )
+				.property(
+						StandalonePojoMapperSettings.INDEXING_PLAN_SYNCHRONIZATION_STRATEGY,
+						IndexingPlanSynchronizationStrategyNames.SYNC
+				)
```
yrodiere, line 62: No sure what's the purpose of this and the corresponding `async` in the property file.

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/mapping/SearchMappingBuilder.java
```diff
@@ -95,6 +98,44 @@ public SearchMappingBuilder properties(Map<String, ?> map) {
...
+	 *
+	 * @param propertySource A configuration property source to add as an override to previously set properties.
+	 * @return {@code this}, for call chaining.
+	 * @see ConfigurationPropertySource#withOverride(ConfigurationPropertySource)
+	 */
+	public SearchMappingBuilder properties(ConfigurationPropertySource propertySource) {
+		// "Reset" the map so the following properties added to it will not be potentially overridden
+		// by a property source we are about to add.
+		this.properties = new HashMap<>();
+		this.propertySource = this.propertySource.withOverride( propertySource )
```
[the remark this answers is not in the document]
marko-bekhta: Good point 😃 btw it seems that we weren't tracking the keys in the standalone mapper anyway since we were creating the source in the constructor before any properties were added to the map, so the checker had an empty key set...

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/bootstrap/ConfiguringPropertiesInSearchMappingBuilderIT.java
marko-bekhta: The idea was to check that properties are getting overridden, but since I've lost the index and search part of the test somewhere ... well the test looks different now anyway 😃

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/mapping/SearchMappingBuilder.java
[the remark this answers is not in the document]
yrodiere: > btw it seems that we weren't tracking the keys in the standalone mapper anyway since we were creating the source in the constructor before any properties were added to the map, so the checker had an empty key set...

Hmm, true. We might want to move the retrieval of all available property keys from `org.hibernate.search.engine.cfg.spi.ConfigurationPropertyChecker#wrap` to `org.hibernate.search.engine.cfg.spi.ConfigurationPropertyChecker#beforeBoot`. But that'll require keeping a list of all wrapped configuration property sources.

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/mapping/SearchMappingBuilder.java
```diff
@@ -95,11 +96,55 @@ public SearchMappingBuilder properties(Map<String, ?> map) {
...
+				propertySource = propertySource.withOverride( getPropertySource( properties, propertyChecker ) );
+			}
+		}
+		else {
+			// means we were only adding properties to the map, so we need to create a source before moving forward:
+			propertySource = properties.isEmpty()
+					? ConfigurationPropertySource.empty()
+					: getPropertySource( properties, propertyChecker );
+		}
+
```
yrodiere, line 147: It seems like this would get much simpler (and robust?) if upon calling `.properties()` you just put the content of the file into `this.properties`... ?

yrodiere (approved):
LGTM, though it seems like there's an opportunity to simplify this even more, now that you've gotten rid of `properties(ConfigurationPropertySource)`.

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/mapping/SearchMappingBuilder.java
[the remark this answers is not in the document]
yrodiere: Or do what you did, that works too :)

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/mapping/SearchMappingBuilder.java
marko-bekhta: IDK 😃, the only reason for keeping "multiple" property sources would be to be able to separate the properties and know where they came from. But if there's no real benefit in that then we can simplify it. Let me know how to proceed 😃

yrodiere: That would be nice, but would also require that we somehow keep the name of the source file in the `ConfigurationPropertySource` for debugging purposes. And since the input is a `Reader`... we can't do that.

I'd favor simplifying this, then?

marko-bekhta: From a new configuration property source to a single couple-lines-method 😄

yrodiere (approved):

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/mapping/SearchMappingBuilder.java
yrodiere: Yeah, I suppose that's how experimentation works: sometimes you have to put in effort to determine what you should *not* do :) Thanks for taking care of this!

PR #3595, opened 2023-07-11
HSEARCH-4889 Forbid usage of internal/incubating Hibernate ORM code (API/SPI/...)

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4889

This is using Jandex to generate text files compatible with forbiddenapis. There are two report generators - one for incubating stuff and another for internal packages. 
I think this would be more flexible for us if for example, we'd need only to include some classes from an entire package marked as incubating in ORM, then we could ignore the package itself and look for all the classes in it to print in the file.
I have added regex rules per each thing currently "violating" the usage (and I was surprised to see some of them 😃).

### DIFF
--- a/build/config/pom.xml
+++ b/build/config/pom.xml
@@ -162,18 +162,25 @@
                             <goal>java</goal>
                         </goals>
                         <configuration>
-                            <mainClass>org.hibernate.checkstyle.report.ReportGenerator</mainClass>
+                            <mainClass>org.hibernate.checkstyle.report.AnnotationReportGenerator</mainClass>
                             <arguments>
                                 <!-- The location of classes to process: -->
                                 <argument>${tmpdir.dependencies-to-parse}/hibernate-core</argument>
                                 <!--
-                                    The report output file path.
+                                    The report output directory path.
                                     We write to the build output directory so that the report is included in the jar:
                                 -->
-                                <argument>${project.build.outputDirectory}/hibernate-core-incubating.txt</argument>
+                                <argument>${project.build.outputDirectory}</argument>
+                                <!-- The report file name main part. Generator will add any suffixes and extensions it needs to it. -->
+                                <argument>hibernate-core-incubating</argument>
                                 <!-- The annotation to look for: -->
                                 <argument>org.hibernate.Incubating</argument>
                                 <!-- Ignore rules: -->
+                                <!--
+                                    Rules can start with a `-Rinternal ` or a `-Rpublic ` prefixes.
+                                    This would mean that a corresponding pattern will only be applied to filtering of internal/public rules.
+                                    If no prefix is specified then rule is considered to be both public and internal at the same time.
+                                -->
                                 <argument>^org\.hibernate\.query\.Query$</argument>
                                 <argument>^org\.hibernate\.query\.SelectionQuery$</argument>
                                 <argument>^org\.hibernate\.query\.MutationQuery$</argument>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -1856,6 +1884,20 @@
...
+                                        <type>jar</type>
+                                        <path>hibernate-core-incubating-internal.txt</path>
+                                    </signaturesArtifact>
+                                    <signaturesArtifact>
+                                        <groupId>org.hibernate.search</groupId>
+                                        <artifactId>hibernate-search-build-config</artifactId>
+                                        <version>${project.version}</version>
+                                        <type>jar</type>
+                                        <path>hibernate-core-internal-packages-internal.txt</path>
+                                    </signaturesArtifact>
```
yrodiere, line 1900: I'd say using incubating or internal Hibernate ORM stuff in tests is mostly okay.
It doesn't affect users, so we have more leeway to fix problems.
And in the case of "incubating" we might even *want* to test how Hibernate Search behaves when using incubating features of Hibernate ORM.

build/config/pom.xml
```diff
@@ -121,6 +126,125 @@
...
+                                <!-- Rules for things used in tests: -->
+                                <argument>-Rinternal ^org\.hibernate\.SessionEventListener$</argument>
+                                <argument>-Rinternal ^org\.hibernate\.tool\.schema\.spi\.SchemaCreator$</argument>
+                                <argument>-Rinternal ^org\.hibernate\.tool\.schema\.spi\.SchemaValidator$</argument>
+                                <argument>-Rinternal ^org\.hibernate\.tool\.schema\.spi\.SchemaMigrator$</argument>
+                                <argument>-Rinternal ^org\.hibernate\.tool\.schema\.spi\.SchemaDropper$</argument>
+                                <argument>-Rinternal ^org\.hibernate\.tool\.schema\.spi\.SchemaManagementTool$</argument>
+                                <argument>-Rinternal ^org\.hibernate\.tool\.schema\.spi\.ExtractionTool$</argument>
+                                <argument>-Rinternal ^org\.hibernate\.tool\.schema\.spi\.ExecutionOptions$</argument>
+                                <argument>-Rinternal ^org\.hibernate\.collection\.spi\.LazyInitializable$</argument>
```
yrodiere, line 206: I'd remove these exceptions, see below: it's mostly fine to use internal/incubating stuff in tests.

build/config/pom.xml
```diff
@@ -121,6 +126,125 @@
...
+                                <argument>^org\.hibernate\.metamodel\.mapping\.Bindable$</argument>
+                                <argument>^org\.hibernate\.metamodel\.mapping\.SelectableMapping$</argument>
+                                <argument>^org\.hibernate\.metamodel\.mapping\.SelectablePath$</argument>
+                                <argument>^org\.hibernate\.metamodel\.model\.domain\.JpaMetamodel$</argument>
+
+                                <argument>^org\.hibernate\.boot\.spi\.BootstrapContext$</argument>
+                                <argument>^org\.hibernate\.collection\.spi\.PersistentCollection$</argument>
+                                <argument>^org\.hibernate\.type\.spi\.TypeConfiguration$</argument>
+                                <argument>^org\.hibernate\.query\.spi\.ScrollableResultsImplementor$</argument>
+                                <argument>^org\.hibernate\.query\.sqm\.spi\.SqmCreationContext$</argument>
```
yrodiere, line 195: All of these make sense. Though I'd expect us to use `@SuppressForbiddenApis` to add exceptions instead, like we've done in other cases.

yrodiere (commented):
Thanks. My comments below.

>  I think this would be more flexible for us if for example, we'd need only to include some classes from an entire package marked as incubating in ORM, then we could ignore the package itself and look for all the classes in it to print in the file.

Not sure I'm getting this. Why would we need to to that, when Hibernate ORM provides a list of incubating packages/classes already?

Did you discuss with Steve the possibility for ORM to generate files in the format that ForbiddenAPIs expects?

>  I have added regex rules per each thing currently "violating" the usage (and I was surprised to see some of them smiley).

Any reason to do this instead of using `@SuppressForbiddenApis` in the offending code?

marko-bekhta:
> Any reason to do this instead of using @SuppressForbiddenApis in the offending code?

hmm, it's been a while 😃 .. I think that when we discussed this last time, we were saying that `@SuppressForbiddenApis` would suppress everything (may suppress too much) and that it would be better if we could ignore specific forbidden-api-rules. Since there's no option to do so in the plugin -- I've added the filtering to the file generation. And because of that, I've ended up with what we have here 😄. 

And that is also why I didn't go with the reports from the ORM ... I have started with those and modified their generation to match the forbidden API format, but then tried them and decided that doing it this way would be more flexible for us 😃🤷🏻‍♂️

yrodiere:
I'd argue others could benefit of Hibernate ORM publishing forbiddenapis signature, and you could just contribute the regexp filtering directly to the forbiddenapis plugin (where I agree it could be useful: sometimes you just want to ignore one signature everywhere, as opposed to ignoring its uses individually).

But, okay, I won't object: you're the one doing the work and you'll probably be the one maintaining this code anyway ;D

Can you please just address the other comments, in particular the one about test rules? Then I think we can merge.

yrodiere (approved):

PR #3598, opened 2023-07-12
HSEARCH-4395 Remove duplicated property for H2 version

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4395

I've noticed that Depedabot was complaining that the version needs to be upgraded ... 
Using the switch-to-orm6 ticket since we've done "a few changes" to poms within that task 😄 
This property is also duplicated in the 6.2 branch, but since the version match for both properties there, it gives predictable results.


yrodiere (approved):

yrodiere:
Nice catch, thanks!

PR #3599, opened 2023-07-14
HSEARCH-4897 Add Configuration Provider SPI

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4897


engine/src/main/java/org/hibernate/search/engine/cfg/impl/FallbackConfigurationPropertySource.java
```diff
@@ -40,6 +42,26 @@ public Optional<String> resolve(String key) {
 		}
 	}
 
+	@Override
+	public ScopedConfigurationPropertySource withScope(BeanResolver beanResolver, String namespace, String name) {
+		ConfigurationPropertySource scopedMain = main;
+		ConfigurationPropertySource scopedFallback = fallback;
+		if ( main instanceof ScopedConfigurationPropertySource ) {
+			scopedMain = ( (ScopedConfigurationPropertySource) main ).withScope( beanResolver, namespace, name );
+		}
```
marko-bekhta, line 51: Note to myself - don't rescope the fallback source

engine/src/main/java/org/hibernate/search/engine/cfg/impl/OverriddenConfigurationPropertySource.java
```diff
@@ -47,6 +49,26 @@ public Optional<String> resolve(String key) {
...
+			scopedOverride = ( (ScopedConfigurationPropertySource) override ).withScope( beanResolver, namespace, name );
+		}
+
+		if ( scopedMain != main || scopedOverride != override ) {
+			return new OverriddenConfigurationPropertySource(
+					scopedMain,
+					scopedOverride
+			);
+		}
+		return this;
```
yrodiere, line 69: If you're going to implement `withScope` on every `ConfigurationPropertySource` subclass and default to `return this` everywhere, you might as well get rid of `ScopedConfigurationPropertySource` and declare `withScope` directly in `ConfigurationPropertySource`, with a default implementation returning `this`. At least that way you wouldn't have to use `instanceof` everywhere... :zipper_mouth_face: 

Though this looks more complicated than I expected, I need to understand this `withScope` method before I can tell for sure...

engine/src/main/java/org/hibernate/search/engine/cfg/impl/EngineConfigurationUtils.java
```diff
@@ -9,29 +9,46 @@
...
+			String indexName) {
 		return engineSource -> {
-			ConfigurationPropertySource backendSource = extractorForBackend.extract( engineSource );
-			return backendSource.withMask( BackendSettings.INDEXES ).withMask( indexName )
+			ScopedConfigurationPropertySource backendSource = extractorForBackend.extract( engineSource );
+			return backendSource
+					// First we want to get the defaults for the default index
+					.withScope( beanResolver, ConfigurationScopeNamespace.INDEX )
+					// First we want to get the defaults for the default index
+					.withScope( beanResolver, ConfigurationScopeNamespace.INDEX, indexName )
```
yrodiere, line 50: Something looks off here.

engine/src/main/java/org/hibernate/search/engine/cfg/impl/EngineConfigurationUtils.java
```diff
@@ -9,29 +9,46 @@
...
+					.withScope( beanResolver, ConfigurationScopeNamespace.BACKEND )
+					.withMask( EngineSettings.Radicals.BACKEND );
 		}
 		else {
-			return engineSource -> engineSource.withMask( EngineSettings.Radicals.BACKENDS )
+			return engineSource -> engineSource
+					// First we want to get the defaults for the default backend
+					.withScope( beanResolver, ConfigurationScopeNamespace.BACKEND )
+					// and then on top of that put the backend specific ones (for a particular backend name)
+					.withScope( beanResolver, ConfigurationScopeNamespace.BACKEND, backendNameOptional.get() )
```
yrodiere, line 34: Are you sure the second call to `withScope` won't just override the first one, i.e. the defaults for the default backend will be ignored?

engine/src/main/java/org/hibernate/search/engine/cfg/impl/RescopableConfigurationPropertySource.java
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	@Override
+	public ScopedConfigurationPropertySource withScope(BeanResolver beanResolver, String namespace, String name) {
+		ConfigurationScope reduced = scope.reduce( namespace, name );
+
+		return new RescopableConfigurationPropertySource(
+				reduced,
+				main,
+				scopeFallback( beanResolver, reduced )
```
yrodiere, line 51: Can't you just use the `beanResolver` passed to the constructor, and remove the `beanResolver` parameter from the `withScope` method?

engine/src/main/java/org/hibernate/search/engine/cfg/impl/EngineConfigurationUtils.java
```diff
@@ -9,29 +9,46 @@
...
 	}
 
-	public static ConfigurationPropertySourceExtractor extractorForBackend(Optional<String> backendNameOptional) {
+	public static ConfigurationPropertySourceExtractor extractorForBackend(BeanResolver beanResolver,
+			Optional<String> backendNameOptional) {
 		if ( !backendNameOptional.isPresent() ) {
-			return engineSource -> engineSource.withMask( EngineSettings.Radicals.BACKEND );
+			return engineSource -> engineSource
+					.withScope( beanResolver, ConfigurationScopeNamespace.BACKEND )
+					.withMask( EngineSettings.Radicals.BACKEND );
```
yrodiere, line 27: I don't understand why you need to push the notion of scope into `ConfigurationPropertySource` and can't simply use something like this here:

```java
			return engineSource -> engineSource.withMask( EngineSettings.Radicals.BACKEND )
						.withOverride( configurationProvider.get( new ConfigurationScope(ConfigurationScopeNamespace.BACKEND, backendNameOptional ) );
```

yrodiere (commented):
The introduction of the notion of "rescoping" is rather unexpected, and I don't understand why it's necessary. See my comments below.

Maybe let's discuss this here or on Zulip before you do further changes?

marko-bekhta:
I've added another commit with an alternative on how we can change the scope. There's an `almostRealisticScenario` test that shows how this works, and the main changes are in `ScopedConfigurationPropertySource`, everything else is mostly noise. 
The idea is that we need to have an "almost clean" source from the user, and we also want to track what changes we are applying to the property sources. why almost clean -- we want to apply any prefix/mask etc but not the fallbacks that come from the scope change. Then any time a scope is changed we generate a new fallback and apply it to this almost-clean user source. When we are generating the scope fallback we are applying any mask/prefix etc changes to the sources coming from the provider depending on the scope they are for.

engine/src/main/java/org/hibernate/search/engine/cfg/impl/EngineConfigurationUtils.java
yrodiere: I still don't understand... can you please explain what's wrong with the example above?

marko-bekhta: mmm so for this part:
```java
new ConfigurationScope(ConfigurationScopeNamespace.BACKEND, null )
```
we don't want to do that just inline since we'd be losing the scope sequence and the provider would only get the "backend", but what we want them to know is `global: backend: backend (some-backend-name): ....`  i.e. how did we get to the scope we are asking for. Let's say we are at the index scope and we have multiple  backends, per each backend we'd want to set a different index override - so we'd need to know from which backend we are coming from....

now as to why we can't simply apply `withFallback`. Let's look at a different example. Assuming we have a property that we want to set at all scopes to its own value. As soon as we apply the first fallback (global one) that key will have a value attached to it. And if the property is redefined for any other deeper scope -- that value is going to be ignored:
```
user source().
    withFallback(globalScopeSource(mykey=myvalue)
    .withFallback(backendScopeSource(mykey=myothervalue)
```
will always give us `myvalue` if we ask for a `source.get("mykey")`

so what we want to happen is this instead:
user source().
    .withFallback(backendScopeSource(mykey=myothervalue)
    withFallback(globalScopeSource(mykey=myvalue)
```
but when we are applying a new scope (or trying to add a new fallback) our property source is already modified and has some fallbacks attached to it. So we need to get way back to the `user source()`... but then we also have all those prefix/mask/override etc applied here and there...

Comment (marko-bekhta) [reply engine/src/main/java/org/hibernate/search/engine/cfg/impl/EngineConfigurationUtils.java]: an alternative-alternative I've been thinking about was to add some sort of priority to the property... i.e. even if the fallback is applied at some later stage but has a higher priority the value from that fallback would "win". But that would mean that the get methods of property sources would need to go through the tree and compare these priorities... which seems to complicate things even more 😃

Comment (yrodiere) [reply engine/src/main/java/org/hibernate/search/engine/cfg/impl/EngineConfigurationUtils.java]: > we don't want to do that just inline since we'd be losing the scope sequence and the provider would only get the "backend", but what we want them to know is global: backend: backend (some-backend-name): .... i.e. how did we get to the scope we are asking for. Let's say we are at the index scope and we have multiple backends, per each backend we'd want to set a different index override - so we'd need to know from which backend we are coming from....

Right, sure, we need a sequence of scopes. No argument there, my example was incomplete.

> Let's look at a different example. Assuming we have a property that we want to set at all scopes to its own value

We don't, though... ? Each scope has a different set of available properties. In places where properties can be applied to multiple scopes, I'd expect us to just apply defaults for each scope.

E.g. index properties can also be applied at the backend scope, to set defaults. So what we have currently (for the default backend) is something like this:

```java
userSource().withMask("hibernate.search.backend.indexes.<indexName>.")
        .withFallback(userSource().withMask("hibernate.search.backend."))
```

I'd have expected you to just do this:

```java
userSource().withMask("hibernate.search.backend.indexes.<indexName>.")
        .withFallback(<defaults from ConfigurationProvider for index <indexName>>)
        .withFallback(userSource().withMask("hibernate.search.backend."))
        .withFallback(<defaults from ConfigurationProvider for the default backend>)
```

And this just... doesn't require any change to `ConfigurationSource`?

marko-bekhta: > E.g. index properties can also be applied at the backend scope, to set defaults. So what we have currently (for the default backend) is something like this:

Well, not really, we don't have a clean user source anywhere but at the very beginning. When we are extracting a property source for the index - we only have access to the backend source. At that point, the fallback is already applied to it. 
Now if we drag the unmodified user property source, we would need to apply the masks and fallbacks to it when we are about to extract properties. So that means either having the modifying operations saved somewhere or applying them as you showed in the example... but that seems error-prone to apply it like that...
In the second example, I think we need to have `.withFallback(<defaults from ConfigurationProvider for the global scope>)` at the end, too, I'd expect that the global scope can define any property (or more generally -- any scope for that matter can define a property that is "deeper"), even a backend/index specific one.
Anyway, if you think that we don't need this, a couple of questions before I make changes:
- where'd you save the current-scope-so-far?
- should we pass around and store in class fields the clean user source and apply the transformations in methods that are using the source?

yrodiere: > Well, not really, we don't have a clean user source anywhere but at the very beginning. When we are extracting a property source for the index - we only have access to the backend source. At that point, the fallback is already applied to it.

Ok, but that just means we'd have this:

```java
var userSource = ...;
var backendSource = userSource.withMask("hibernate.search.backend.")
        .withFallback(<defaults from ConfigurationProvider for the default backend>);
var indexSource = backendSource().withMask("indexes.<indexName>.")
        .withFallback(<defaults from ConfigurationProvider for index <indexName>>);
```

This seems to still match what we need? The only downside is that defaults from the backend that use the prefix `indexes.<indexName>` would take precedence over defaults from the index, but, well... That seems unlikely. Whoever implements `ConfigurationProvider` just needs to be careful about not introducing conflicts.

> but that seems error-prone to apply it like that...

I'm not seeing how exactly this is error prone, could you please explain?

>  In the second example, I think we need to have .withFallback(<defaults from ConfigurationProvider for the global scope>) at the end, too, I'd expect that the global scope can define any property (or more generally -- any scope for that matter can define a property that is "deeper"), even a backend/index specific one.

I'd expect that to be applied much earlier, e.g.:

```java
var userSource = <something>.withFallback(<defaults from ConfigurationProvider for the global scope>);
var backendSource = userSource.withMask("hibernate.search.backend.")
        .withFallback(<defaults from ConfigurationProvider for the default backend>);
var indexSource = backendSource().withMask("indexes.<indexName>.")
        .withFallback(<defaults from ConfigurationProvider for index <indexName>>);
```

> Anyway, if you think that we don't need this

Depends what you mean by "this" xD

> where'd you save the current-scope-so-far?

I think in most (all?) cases we can rebuild it from the context. E.g. in `org.hibernate.search.engine.common.impl.IndexManagerBuildingStateHolder.BackendInitialBuildState#createIndexManagerBuildingState` we have access to `backendBuildContext.backendName()`.

>    should we pass around and store in class fields the clean user source and apply the transformations in methods that are using the source?

From what I can see you don't necessarily need the "clean" user source. Deriving the backend source from the "defaulted" global one, or the index source from the "defaulted" backend one, seems fine.

yrodiere: Discussed through other channels, we agreed to go with the "dumb" solution I suggested, mainly because `ConfigurationProvider` is a (relatively obscure) SPI so we can reasonably require from implementors that they take care not to introduce conflicts between properties from e.g. the global scope and the index scope.

marko-bekhta:
superseded by https://github.com/hibernate/hibernate-search/pull/3632

PR #3602, opened 2023-07-18
HSEARCH-4861 Add properties to the build BOM

marko-bekhta:

- so that they are visible and upgradable by a maven version plugin

followup on https://github.com/hibernate/hibernate-search/pull/3588

### DIFF
--- a/bom/build/pom.xml
+++ b/bom/build/pom.xml
@@ -21,6 +21,12 @@
     <properties>
         <!-- Public modules must be signed. -->
         <release.gpg.signing.skip>false</release.gpg.signing.skip>
+        <!--
+            Since the properties live in the parent POM maven version plugin does not pick them up for an update.
+            Hence, we trick the plugin into updating the versions by introducing a property that resides in the same POM
+            as a dependency management is defined, and which is copying the value of the main POM property.
+        -->
+        <version.upgradeable.org.hibernate.orm>${version.org.hibernate.orm}</version.upgradeable.org.hibernate.orm>
     </properties>
 
     <dependencyManagement>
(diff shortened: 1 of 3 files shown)


bom/build/pom.xml
```diff
@@ -21,6 +21,12 @@
     <properties>
         <!-- Public modules must be signed. -->
         <release.gpg.signing.skip>false</release.gpg.signing.skip>
+        <!--
+            Since the properties live in the parent POM maven version plugin does not pick them up for an update.
+            Hence, we trick the plugin into updating the versions by introducing a property that resides in the same POM
+            as a dependency management is defined, and which is copying the value of the main POM property.
+        -->
+        <version.upgradeable.org.hibernate.orm>${version.org.hibernate.orm}</version.upgradeable.org.hibernate.orm>
```
yrodiere, line 29: I suspect this might create problems with dependabot in the future, if we decide to enable it for Hibernate ORM.

Also, we'll have to resort to the same workaround every single time we update another property, and I'm not fond of the idea...

Can't we just change how we invoke the maven-versions-plugin to make it work correctly? Possibly by adding the parent module to the list of modules to be updated?

marko-bekhta: yeah... +1 on that 😃. Though... I think that we are already executing the plugin on all modules: https://github.com/hibernate/hibernate-search/blob/fbd4ab868cca0a0a44d67595486aa4e544217287/ci/dependency-update/perform-update.sh#L74-L79 and back then I didn't see any useful configs in the plugin (https://www.mojohaus.org/versions/versions-maven-plugin/examples/update-properties.html). I can try to look at their sources/debug to see what's happening in this case.

yrodiere: If we're already executing this on all projects, then this looks like a bug in the plugin...

+1 to investigate what's going on with BOMs.

marko-bekhta: OK, starting to look into it 😃

PR #3605, opened 2023-07-19
HSEARCH-4900 Add custom enforcer rules

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4900

since we've changed what we include in the BOM, the original idea wasn't making sense anymore, so instead, I have added a couple of simple rules that would:
- check we've included all public artifacts in the BOM
- check the jakarta-persistence version property matches whatever the Hibernate ORM BOM tells us it should be.

Overall, it looks like these custom rules can be pretty flexible since they have access to the entire project structure and to Maven itself (to resolve dependencies from a repository, for example)

### DIFF
--- a/build/config/pom.xml
+++ b/build/config/pom.xml
@@ -68,6 +68,48 @@
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-enforcer-plugin</artifactId>
+                <executions>
+                    <!--
+                        We cannot include this execution in the root pom because it depends on a project artifact
+                        that we need to build first.
+                        It should be enough to only run this rule on this build module, since the check it performs is always
+                        the same - get the property value, and compare it to the dependency version from dependency management.
+                    -->
+                    <execution>
+                        <id>enforce-custom-rules</id>
+                        <goals>
+                            <goal>enforce</goal>
+                        </goals>
+                        <configuration>
+                            <rules>
+                                <versionAlignRule>
+                                    <propertiesToCheck>
+                                        <!--
+                                            We want to make sure that the value we store in the property matches
+                                            the version that is imported from the Hibernate ORM BOM
+                                         -->
+                                        <item>
+                                            <property>${version.jakarta.persistence}</property>
+                                            <artifact>jakarta.persistence:jakarta.persistence-api</artifact>
+                                            <failOnNotFound>true</failOnNotFound>
+                                        </item>
+                                    </propertiesToCheck>
+                                </versionAlignRule>
+                            </rules>
+                        </configuration>
+                    </execution>
+                </executions>
+                <dependencies>
+                    <dependency>
+                        <groupId>org.hibernate.search</groupId>
+                        <artifactId>hibernate-search-build-enforcer</artifactId>
+                        <version>${project.version}</version>
+                    </dependency>
+                </dependencies>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-dependency-plugin</artifactId>
                 <executions>
                     <!-- See https://maven.apache.org/plugins/maven-dependency-plugin/examples/unpacking-artifacts.html -->
--- a/build/enforcer/pom.xml
+++ b/build/enforcer/pom.xml
@@ -1,0 +1,84 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ ~ Hibernate Search, full-text search for your domain model
+ ~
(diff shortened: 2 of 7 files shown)


build/config/pom.xml
```diff
@@ -66,6 +66,48 @@
...
     <build>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-enforcer-plugin</artifactId>
+                <executions>
+                    <!--
+                        We cannot include this execution in the root pom because it depends on a project artifact
+                        that we need to build first.
+                        It should be enough to only run this rule on this build module, sine the check it performs is always
```
yrodiere, line 76:
```suggestion
                        It should be enough to only run this rule on this build module, since the check it performs is always
```

build/config/pom.xml
```diff
@@ -66,6 +66,48 @@
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-enforcer-plugin</artifactId>
+                <executions>
+                    <!--
+                        We cannot include this execution in the root pom because it depends on a project artifact
+                        that we need to build first.
```
yrodiere, line 75: Maybe this is an argument in favor of a dedicated "build" parent POM that is not the root POM, and could thus depend on `hibernate-search-build-enforcer`? Assuming `hibernate-search-build-enforcer` does not need all our dependency/plugin configuration...

build/enforcer/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+            as a result the required files to run the checks will be missing.
+         -->
+        <checkstyle.skip>true</checkstyle.skip>
+        <forbiddenapis.skip>true</forbiddenapis.skip>
+        <format.skip>true</format.skip>
+
+        <skipNexusStagingDeployMojo>true</skipNexusStagingDeployMojo>
+        <jqassistant.skip>true</jqassistant.skip>
+
+        <version.sisu-maven-plugin>0.9.0.M1</version.sisu-maven-plugin>
```
yrodiere, line 35: Maybe put this in the root POM, with plugin management just for the version? We've managed to put plugin versions in a central place so far, better stick to that.

build/enforcer/src/main/java/org/hibernate/search/build/enforcer/DependencyManagementIncludesAllPublicArtifactsRule.java
```diff
@@ -0,0 +1,64 @@
...
+import javax.inject.Inject;
+import javax.inject.Named;
+
+import org.apache.maven.enforcer.rule.api.AbstractEnforcerRule;
+import org.apache.maven.enforcer.rule.api.EnforcerRuleException;
+import org.apache.maven.execution.MavenSession;
+import org.apache.maven.model.Dependency;
+import org.apache.maven.project.MavenProject;
+
+@Named("dependencyManagementIncludesAllPublicArtifactsRule") // rule name - must start from lowercase character
```
yrodiere, line 22:
```suggestion
@Named("dependencyManagementIncludesAllPublicArtifactsRule") // rule name - must start with lowercase character
```

build/config/pom.xml
```diff
@@ -66,6 +66,48 @@
...
+                                    <propertiesToCheck>
+                                        <!--
+                                            We want to make sure that the value we store in the property matches
+                                            the version that is imported from the Hibernate ORM BOM
+                                         -->
+                                        <item>
+                                            <property>${version.jakarta.persistence}</property>
+                                            <artifact>jakarta.persistence:jakarta.persistence-api</artifact>
+                                            <failOnNotFound>true</failOnNotFound>
+                                        </item>
```
yrodiere, line 96: Nice!

yrodiere (commented):
Nice, thanks! A few suggestions below, mostly typos.

build/enforcer/pom.xml
```diff
@@ -0,0 +1,85 @@
...
+                        and additionally, it is using javax-inject so this would not match our main enforcer rules.
+                    -->
+                    <skip>true</skip>
+                </configuration>
+            </plugin>
+            <plugin>
+                <!-- generate index of project components -->
+                <groupId>org.eclipse.sisu</groupId>
+                <artifactId>sisu-maven-plugin</artifactId>
+                <version>${version.sisu-maven-plugin}</version>
```
yrodiere, line 74:
```suggestion
```

yrodiere (commented):
You forgot to remove something ;)

Apart from that, feel free to merge.

yrodiere (approved):

PR #3610, opened 2023-07-25
HSEARCH-4403 Support ORM's discriminator-based multi-tenancy in the outbox-polling module

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4403

Opening as a draft as this won't work for now while the changes are not applied to ORM


mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/cluster/impl/OutboxPollingAgentAdditionalJaxbMappingProducer.java
```diff
@@ -168,38 +132,55 @@ public Collection<MappingDocument> produceMappings(ConfigurationPropertySource p
...
+			mappings = createMappings(
+					schema.orElse( "" ),
+					catalog.orElse( "" ),
+					table.orElse( HibernateOrmMapperOutboxPollingSettings.Defaults.COORDINATION_ENTITY_MAPPING_AGENT_TABLE ),
+					resolvedUuidType, resolvedPayloadType, resolvedUuidStrategy,
+					buildingContext.getMetadataCollector().getFilterDefinition( TenantIdBinder.FILTER_NAME ) != null
+			);
+		}
+
+		log.agentGeneratedEntityMapping( marshall( mappings ) );
```
yrodiere, line 162: You should use a formatter in the definition of the logging method, so that `marshall` is only called when this is actually logged (in this case, when debug level is enabled, IIRC).

And if that's too complicated, then `mappings.toString()` may be good enough for what we're doing here (debugging).

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/cluster/impl/OutboxPollingAgentAdditionalJaxbMappingProducer.java
```diff
@@ -168,38 +132,55 @@ public Collection<MappingDocument> produceMappings(ConfigurationPropertySource p
...
+			String resolvedUuidStrategy = uuidStrategy.orElse(
+					HibernateOrmMapperOutboxPollingSettings.Defaults.COORDINATION_ENTITY_MAPPING_AGENT_UUID_GEN_STRATEGY )
+					.strategy();
+
+			mappings = createMappings(
+					schema.orElse( "" ),
+					catalog.orElse( "" ),
+					table.orElse( HibernateOrmMapperOutboxPollingSettings.Defaults.COORDINATION_ENTITY_MAPPING_AGENT_TABLE ),
+					resolvedUuidType, resolvedPayloadType, resolvedUuidStrategy,
+					buildingContext.getMetadataCollector().getFilterDefinition( TenantIdBinder.FILTER_NAME ) != null
```
yrodiere, line 158: Maybe create a `isDiscriminatorMultiTenancyEnabled()` method somewhere (e.g. `OrmUtils`), because the meaning of this isn't quite obvious xD

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/cluster/impl/OutboxPollingAgentAdditionalJaxbMappingProducer.java
```diff
@@ -168,38 +132,55 @@ public Collection<MappingDocument> produceMappings(ConfigurationPropertySource p
...
+				.enumAttribute( "type", null, false )
+				.attribute( "name", null, false )
+				.attribute( "expiration", null, false )
+				.enumAttribute( "state", null, false )
+				.attribute( "totalShardCount", null, true )
+				.attribute( "assignedShardIndex", null, true )
+				.attribute( "payload", Length.LONG32, true, resolvedPayloadType );
+		if ( tenantIdRequired ) {
+			builder.tenantId( "tenantId" );
+		}
```
yrodiere, line 183: Did you test somehow that this is effective, i.e. Hibernate ORM doesn't expect a `tenantId` column in the table when using single-tenancy or schema-based/database-based multi-tenancy?

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/event/impl/OutboxPollingOutboxEventAdditionalJaxbMappingProducer.java
yrodiere, line 1: Same comments here as on the corresponding class for Agents.

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/mapping/impl/AdditionalMappingContributorBuilder.java
```diff
@@ -0,0 +1,122 @@
...
+import org.hibernate.boot.jaxb.mapping.JaxbColumn;
+import org.hibernate.boot.jaxb.mapping.JaxbEntity;
+import org.hibernate.boot.jaxb.mapping.JaxbEntityMappings;
+import org.hibernate.boot.jaxb.mapping.JaxbId;
+import org.hibernate.boot.jaxb.mapping.JaxbIndex;
+import org.hibernate.boot.jaxb.mapping.JaxbTable;
+import org.hibernate.boot.jaxb.mapping.JaxbTenantId;
+import org.hibernate.boot.jaxb.mapping.JaxbUuidGenerator;
+
+public class AdditionalMappingContributorBuilder {
```
yrodiere, line 26: Technically this would be `AdditionalMappingBuilder`?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/bootstrap/impl/HibernateSearchCompositeMappingProducer.java
```diff
@@ -6,51 +6,52 @@
...
+					.coordinationStrategyConfiguration().mappingProducers() ) {
+				for ( Map.Entry<Class<?>, JaxbEntityMappings> entry : mappingProducer.produceMappings(
+						propertySource,
+						metadata.getDatabase().getDialect(),
+						buildingContext
+				).entrySet() ) {
+					contributions.contributeEntity( entry.getKey() );
+					contributions.contributeBinding( entry.getValue() );
+				}
+			}
```
yrodiere, line 54: And this diff right here is why one doesn't rework indentation of a whole method just because one doesn't like early returns :D

yrodiere (commented):
Some early comments while we're waiting for the necessary changes to be merged in Hibernate ORM :)

And BTW, you might want to ping someone so that it's merged before 6.3.0.Final gets released...

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/cfg/impl/UuidDataTypeUtils.java
```diff
@@ -29,14 +30,19 @@ private UuidDataTypeUtils() {
...
+				return SqlTypes.BINARY;
 			}
-			return UUID_CHAR;
+			return SqlTypes.CHAR;
+		}
+		if ( UUID_CHAR.equals( value ) ) {
+			return SqlTypes.CHAR;
+		}
+		else {
+			return SqlTypes.BINARY;
```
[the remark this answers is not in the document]
marko-bekhta: Well.... it'll be a "hello-binary-kitty" 😉😄

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/cfg/impl/UuidDataTypeUtils.java
```diff
@@ -29,14 +30,19 @@ private UuidDataTypeUtils() {
 	 * <p>
 	 * Otherwise, when {@code value != "default" } a user passed value will be used.
```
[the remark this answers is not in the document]
marko-bekhta: I was 50-50 on this one. I guess if we change the expectations and instead of a string, we say that a user can pass a SQL-type-code ... then it still can work. I'm not sure if there's a simple alternative though ... using something like `org.hibernate.annotations.Type` will still break this...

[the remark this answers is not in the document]
yrodiere: From what I can see some configuration properties in Hibernate ORM accept either the name of a `SqlTypes` constant or its value, e.g. `hibernate.type.preferred_duration_jdbc_type`. Depending how it's implemented we might do the same here, automatically resolving a name to a code?

See https://github.com/hibernate/hibernate-orm/pull/6562/files#diff-fc943226b0861c9546796194526878ba1a19474f164a99ab16bb5cfc886f656d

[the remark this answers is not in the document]
marko-bekhta: Hmmmm yeah. ORM uses this converter:
https://github.com/hibernate/hibernate-orm/blob/ae8b3f9a33f204086cffcb857a4863f20c2953ad/hibernate-core/src/main/java/org/hibernate/internal/util/config/ConfigurationHelper.java#L589-L611
while it is private and lives in an internal module the `JdbcTypeNameMapper` it is using is in a public module so I guess we can do something similar 👍🏻😃

marko-bekhta:
waiting for the fix from https://github.com/hibernate/hibernate-orm/pull/7171

yrodiere:
Looks like it's failing:

```

[ERROR] Constraint: hsearch:TypesMayNotDependOnImplementationTypeFromOtherModules

[ERROR] Severity: MAJOR

[ERROR] Number of rows: 2

[ERROR]             Main (non-test) types must not depend on implementation types from other modules.

[ERROR]             SPIs must be used for such dependencies.

[ERROR]             Exceptions are allowed only when the dependency type is in a util module,

[ERROR]             or the depending type is a test type.

[ERROR]         

[ERROR]   artifact1=org.hibernate.search:hibernate-search-mapper-orm-coordination-outbox-polling:jar:7.0.0-SNAPSHOT, type1=org.hibernate.search.mapper.orm.coordination.outboxpolling.cluster.impl.OutboxPollingAgentAdditionalJaxbMappingProducer, artifact2=org.hibernate.search:hibernate-search-mapper-orm:jar:7.0.0-SNAPSHOT, type2=org.hibernate.search.mapper.orm.common.impl.HibernateOrmUtils

[ERROR]   artifact1=org.hibernate.search:hibernate-search-mapper-orm-coordination-outbox-polling:jar:7.0.0-SNAPSHOT, type1=org.hibernate.search.mapper.orm.coordination.outboxpolling.event.impl.OutboxPollingOutboxEventAdditionalJaxbMappingProducer, artifact2=org.hibernate.search:hibernate-search-mapper-orm:jar:7.0.0-SNAPSHOT, type2=org.hibernate.search.mapper.orm.common.impl.HibernateOrmUtils

```

I guess you can have separate utils directly in the `outbox-polling` module if necessary

marko-bekhta:
will need ORM 6.3.1 for this one 🙈

yrodiere:
> will need ORM 6.3.1 for this one 🙈

That's... annoying. We may have to postpone to Search 7.1 then. We'll see.

marko-bekhta:
I've added a patch to ORM; the info on the preferred id JDBC type wasn't passed further from the XML.... 
But there's also a difference in behaviour with long32varbinary.. maybe let's talk about that today

build/parents/build/pom.xml
```diff
@@ -95,7 +95,7 @@
 
         <!-- >>> ORM 6 with Jakarta Persistence -->
         <!-- IMPORTANT: Make sure that version.jakarta.persistence aligns with the imported version! -->
-        <version.org.hibernate.orm>6.3.0.Final</version.org.hibernate.orm>
+        <version.org.hibernate.orm>6.3.1.Final</version.org.hibernate.orm>
```
yrodiere, line 98: Putting this as a reminder: we'll need a separate Jira issue for this.

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/event/impl/OutboxPollingOutboxEventAdditionalJaxbMappingProducer.java
```diff
@@ -123,32 +89,21 @@ public final class OutboxPollingOutboxEventAdditionalJaxbMappingProducer
...
-
+	public Map<Class<?>, JaxbEntityMappings> produceMappings(ConfigurationPropertySource propertySource, Dialect dialect,
+			MetadataBuildingContext buildingContext) {
 		Optional<String> mapping = OUTBOXEVENT_ENTITY_MAPPING.get( propertySource );
 		Optional<String> schema = ENTITY_MAPPING_OUTBOXEVENT_SCHEMA.get( propertySource );
 		Optional<String> catalog = ENTITY_MAPPING_OUTBOXEVENT_CATALOG.get( propertySource );
 		Optional<String> table = ENTITY_MAPPING_OUTBOXEVENT_TABLE.get( propertySource );
+
 		Optional<UuidGenerationStrategy> uuidStrategy = ENTITY_MAPPING_OUTBOXEVENT_UUID_GEN_STRATEGY.get( propertySource );
 		Optional<String> uuidType = ENTITY_MAPPING_OUTBOXEVENT_UUID_TYPE.get( propertySource );
```
yrodiere, line 102: Could you rather use `getAndTransform` here and call `UuidDataTypeUtils.uuidType` in the lambda? Then your error message can be much simpler as Hibernate Search will mention the property key and value automatically (among other things).

yrodiere: Same for the other mapping producer, obviously.

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/event/impl/OutboxPollingOutboxEventAdditionalJaxbMappingProducer.java
```diff
@@ -157,50 +112,60 @@ public Collection<MappingDocument> produceMappings(ConfigurationPropertySource p
...
+
+	private static JaxbEntityMappings createMappings(String schema, String catalog,
+			String table, int resolvedUuidType, String resolvedUuidStrategy,
+			boolean tenantIdRequired) {
+		AdditionalMappingBuilder builder = new AdditionalMappingBuilder(
+				OutboxEvent.class, ENTITY_NAME )
+				.id( resolvedUuidType, resolvedUuidStrategy )
+				.index( "entityIdHash" )
+				.index( "status" )
+				.index( "processAfter" )
```
yrodiere, line 157: Just to be sure: you did check that the resulting schema on the database side remains the same, right? I wonder about these indexes in particular, they might have a different name as they are defined slightly differently.

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/logging/impl/Log.java
```diff
@@ -198,17 +199,24 @@ void startingOutboxMassIndexerAgent(String name,
...
+	SearchException unableToProcessEntityMappings(String causeMessage, @Cause Exception cause);
+
+	@Message(id = ID_OFFSET + 36,
+			value = "Couldn't interpret a '%1$s' property value '%2$s' as JDBC type code or type code name. %3$s")
+	SearchException unableToParseJdbcTypeCode(String property, String value, String causeMessage, @Cause Exception cause);
+
+	@LogMessage(level = WARN)
+	@Message(id = ID_OFFSET + 37,
+			value = "Configuration property '%1$s' is configured with a deprecated value '%2$s'. "
+					+ "Use the '%3$s' instead.")
```
yrodiere, line 220:
```suggestion
					+ "Use '%3$s' instead.")
```

yrodiere (changes requested):
Thanks. A few comments on the latest version :)

integrationtest/mapper/orm-coordination-outbox-polling/src/test/java/org/hibernate/search/integrationtest/mapper/orm/coordination/outboxpolling/mapping/AbortedEventsGenerator.java
```diff
@@ -8,94 +8,116 @@
...
+			}
+			else {
+				currentTenant.set( tenantId );
+				return with( sessionFactory, tenantId ).applyInTransaction( this::generateThreeAbortedEvents );
+			}
 		}
-		else {
-			with( sessionFactory, tenantId ).runInTransaction( this::generateThreeAbortedEvents );
+		finally {
+			backendMock.verifyExpectationsMet();
```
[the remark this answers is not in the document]
marko-bekhta: hmm ok, yeah I wanted to make the check of expectations, but you have a point if it fails prior to it then it may be unclear why... will update this

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/mapping/impl/HibernateOrmMappingInitiator.java
```diff
@@ -74,7 +76,9 @@ public static HibernateOrmMappingInitiator create(Metadata metadata, IndexView j
 				HibernateOrmUtils.getServiceOrFail( serviceRegistry, HibernateSearchPreIntegrationService.class );
 
 		boolean multiTenancyEnabled = ( (MetadataImplementor) metadata ).getMetadataBuildingOptions()
-				.isMultiTenancyEnabled();
+				.isMultiTenancyEnabled()
+				// discriminator multi-tenancy
+				|| isDiscriminatorMultiTenancyEnabled( metadata );
```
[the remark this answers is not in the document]
marko-bekhta: Yeah, the comment stayed from the time when there was no util method, and there was just that long line looking up the filter.

util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/multitenancy/impl/MultitenancyTestHelper.java
```diff
@@ -22,6 +25,23 @@
...
+								return currentTenantSupplier.get();
+							}
+
+							@Override
+							public boolean validateExistingCurrentSessions() {
+								return false;
+							}
+						}
+				) );
+	}
```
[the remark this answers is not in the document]
marko-bekhta: I've been following the tests that are in ORM for this discriminator multitenantcy and they all have used this current tenant id resolver, so I assumed that's a "requirement". But now that you've asked, I've tested it without the resolver, and all is ok. Will clean this up to make it closer to what it was before.

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/event/impl/OutboxPollingOutboxEventAdditionalJaxbMappingProducer.java
marko-bekhta: Hmm, I'm not sure that this will work... Since we are only logging a warning rather than throwing an exception, the logged message won't get the context of the property-being-resolved... I'll keep it the way it is for now, and we can discuss this tomorrow.

yrodiere: So first, I was wrong in suggesting `getAndTransform`, it's actually `getAndMap` that you should use.

Regarding your comment, `unableToParseJdbcTypeCode` would work just fine as it's an exception, but `usingDeprecatedPropertyValue` is indeed a warning (I didn't notice that)  and would indeed still need access to the property name.

IMO you can still go with `getAndMap`, and resolve the property name if a warning really is needed. Maybe change the signature of your method to something like:

```
public static int uuidType(ConfigurationPropertySource source, OptionalConfigurationProperty property, Dialect dialect) {
```

... and work from there?

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/event/impl/OutboxPollingOutboxEventAdditionalJaxbMappingProducer.java
marko-bekhta: Yes, when I was looking at the column types I've also looked at the queries generated for creating indexes -- and those looked the same (same index names)

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/event/impl/OutboxPollingOutboxEventAdditionalJaxbMappingProducer.java
yrodiere: I meant it when I suggested this signature:

```
public static int uuidType(ConfigurationPropertySource source, OptionalConfigurationProperty property, Dialect dialect) {
```

If you also pass the value, there's not much sense in doing this. The idea was that `uuidType()` itself would call `property.get(source).getAndTransform(value -> { ... })`.

yrodiere: Ah, I see you called `getAndMap` in the callers... whatever, this works too. Ignore me :)

yrodiere (approved):

PR #3615, opened 2023-07-31
HSEARCH-4907 Fix links and formatting in the migration guide

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4907

See also #3616 (similar, but not identical)


marko-bekhta (approved):
🎉

PR #3617, opened 2023-08-01
HSEARCH-4908 Upgrade to Hibernate ORM 6.3.0.CR1

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4908


marko-bekhta:
hm, that's strange...
```
Caused by: org.hibernate.AnnotationException: Entity class 'org.hibernate.search.integrationtest.spring.jta.entity.Box' is annotated '@Cache' but it is a subclass in an entity inheritance hierarchy (only root classes may define second-level caching semantics)
```
I'd expect that the upgrade job would've also failed with these errors

PR #3618, opened 2023-08-01
Followup HSEARCH-4861 Move properties closer to dependency management

marko-bekhta:

I left the plugins as they were, and instead of importing a build BOM, there's a dependencies parent that contains properties and dependency management.


build/parents/dependencies/pom.xml
```diff
@@ -7,26 +7,166 @@
...
+        <version.com.fasterxml.jackson>2.15.0</version.com.fasterxml.jackson>
+        <!-- slf4j: used by the AWS SDK -->
+        <version.org.slf4j>1.7.36</version.org.slf4j>
+
+        <!-- >>> ORM 6 with Jakarta Persistence -->
+        <!-- IMPORTANT: Make sure that version.jakarta.persistence aligns with the imported version! -->
+        <version.org.hibernate.orm>6.2.7.Final</version.org.hibernate.orm>
+
+        <javadoc.org.hibernate.orm.url>https://docs.jboss.org/hibernate/orm/${parsed-version.org.hibernate.orm.majorVersion}.${parsed-version.org.hibernate.orm.minorVersion}/javadocs/</javadoc.org.hibernate.orm.url>
+        <documentation.org.hibernate.orm.url>https://docs.jboss.org/hibernate/orm/${parsed-version.org.hibernate.orm.majorVersion}.${parsed-version.org.hibernate.orm.minorVersion}/userguide/html_single/Hibernate_User_Guide.html</documentation.org.hibernate.orm.url>
```
yrodiere, line 98: So this property is defined in `/build/parents/dependencies/pom.xml`, but it relies on `parsed-version.org.hibernate.orm.majorVersion` which is defined in `/pom.xml`, but is generated based on `version.org.hibernate.orm` which is defined in `/build/parents/dependencies/pom.xml`.

... are you sure it's a good idea not to move the plugin config to this pom?

marko-bekhta: Was I sure about it - no 😃, did I try that it somehow was working and was surprised to see that it does - yes 😆.
yeah, my thought was since it worked, maybe it would be easier to find the plugin configs if they stayed in the same place. But I agree that it looks strange. So I've split the executions and moved the properties related to the docs and based on lib versions to the other pom.

yrodiere: You really don't want to move plugin configuration here, do you xD

I'll try something else: see the configuration of the maven-compiler-plugin in `/pom.xml`: it references the version of jboss-logging... which is defined in `/build/parents/dependencies/pom.xml`!

I really don't think it makes sense to configure plugins in a context that does not have (conceptually) access to dependency versions, since some plugins will need to refer to those versions...

If your concern is about publishing the root POM or the public BOM, then that point is moot once we'll flatten all POMs (#3603).

If your concern is about not being able to import the public BOM if we move plugin configuration here, I don't think it matters anymore, since we only expose our own artifacts in that BOM now, and you removed any chance that we will forget about one in #3605. So the BOM will always be aligned with our build.

The way I see it, considering the various ways we've had to alter our plans regarding BOMs, there are only two ways forward that make sense now:

1. Have a dedicated `build/parents/build/pom.xml` that contains all properties (versions and others), dependency management, and plugins, separated from the root POM that would basically become an "aggregator POM".
2. Move versions back to the root POM.

I can see value in 1 over 2, be it only as a way to un-clutter the root POM. Option 2 would be more in line with other projects such as Validator, but I'm not sure we even care nowadays, with some projects using Gradle and not others, and so many differences between projects. Feel free to start a thread on Zulip if you want to discuss our options.

marko-bekhta: > You really don't want to move plugin configuration here, do you xD

was it too obvious? 😆 🤣 

Ok so I've moved most of the plugins to this new pom, some of them remained in the root, and there are comments on why they are there.
As to some moving the IDE profiles -- it somewhat makes sense to have them in the root pom, and I'd rather keep them there, but then they are referencing properties e.g. IDEA is using the jboss-logging/jmh/java-test versions ... so they got moved.
I've left the reporting in the root pom since I wasn't sure about it. and also that last profile for including reporting module ... also stayed in the root.

now a slight issue I can see is that any plugin we want to call from the command line if not defined in the root would be executed with some random version before reaching the modules that have this new one as a parent (that's why I've kept the clean/install plugins in the root).

The next logical step seems to clean up the modules 😄, meaning it might make sense to have something like a couple of modules in the root (`helpers` `boms` `actual` obviously with better names). `Helpers` would have the build configs, property collector, enforcer rules etc, basically anything that we'd later be referencing during the main build.... it would mean that some plugins like compilation or maybe for tests would have a management entry in the root pom and then configured in the modules...

yrodiere: > As to some moving the IDE profiles -- it somewhat makes sense to have them in the root pom, and I'd rather keep them there, but then they are referencing properties e.g. IDEA is using the jboss-logging/jmh/java-test versions ... so they got moved.

For everything I can see in this profile (jboss-logging/jmh/java-test, dependency convergence), this makes sense, since the corresponding configuration would not be relevant for modules that do not extend the `build` parent POM  (BOM, `build/config`, `build/enforcer`, ...).

> now a slight issue I can see is that any plugin we want to call from the command line if not defined in the root would be executed with some random version before reaching the modules that have this new one as a parent (that's why I've kept the clean/install plugins in the root).

I suppose we could use plugin management to set all the plugin versions in the root POM, and in `build/parent/build/pom.xml` we'd only include plugin configuration (and only for those plugins where we need configuration)? Since only the configuration of those plugins is actually a problem to have in the root POM, not version management. I'll let you decide if this makes sense (and if you're not fed up with changing that stuff ;) ).

> Helpers would have the build configs, property collector, enforcer rules etc, basically anything that we'd later be referencing during the main build....

That's what we currently have with the `build` directory, no?

I'm not getting what needs to be cleaned up exactly.

yrodiere: > > now a slight issue I can see is that any plugin we want to call from the command line if not defined in the root would be executed with some random version before reaching the modules that have this new one as a parent (that's why I've kept the clean/install plugins in the root).
> 
> I suppose we could use plugin management to set all the plugin versions in the root POM, and in `build/parent/build/pom.xml` we'd only include plugin configuration (and only for those plugins where we need configuration)? Since only the configuration of those plugins is actually a problem to have in the root POM, not version management. I'll let you decide if this makes sense (and if you're not fed up with changing that stuff ;) ).

By the way, good point about some plugins needing to be configured at the root (in the aggregator POM, essentially).

Looking closer at what you did, there seems to be a few more plugins in that case, in particular jqassistant and sonar, maybe others. Those plugins actually take advantage of the fact they're executed in the aggregator POM.

Maybe only a few plugins actually need to be in `build/parents/build/pom.xml`... essentially anything that is specific to Hibernate Search's runtime (e.g. annotation processors, ...), as opposed to the way we want modules to be built in general.

I think I'm starting to be a nuisance rather than a help here. Do you want me to have a closer look and push a commit with something that seems to make sense to me?

yrodiere:
The build failed, it seems list-container-images.sh doesn't work with these changes

build/parents/dependencies/pom.xml
yrodiere: See https://hibernate.zulipchat.com/#narrow/stream/132092-hibernate-search-dev/topic/HSEARCH-4861 for the rest of the discussion

pom.xml
```diff
@@ -698,7 +460,9 @@
              - https://www.eclemma.org/jacoco/trunk/doc/report-aggregate-mojo.html
              - Not relevant anymore, but we used to merge *.exec files: https://www.eclemma.org/jacoco/trunk/doc/merge-mojo.html
          -->
-        <sonar.coverage.jacoco.xmlReportPaths>${rootProject.directory}/build/reports/target/site/jacoco-aggregate/jacoco.xml</sonar.coverage.jacoco.xmlReportPaths>
+        <sonar.coverage.jacoco.xmlReportPaths>
+            ${rootProject.directory}/build/reports/target/site/jacoco-aggregate/jacoco.xml
+        </sonar.coverage.jacoco.xmlReportPaths>
```
yrodiere, line 465: Let's not ask for trouble, there's always a chance a Maven plugin won't trim input properly...

```suggestion
        <sonar.coverage.jacoco.xmlReportPaths>${rootProject.directory}/build/reports/target/site/jacoco-aggregate/jacoco.xml</sonar.coverage.jacoco.xmlReportPaths>
```

pom.xml
```diff
@@ -1522,8 +770,6 @@
                         <systemPropertyVariables>
                             <org.hibernate.search.test.property-set-in-maven-config>someValue</org.hibernate.search.test.property-set-in-maven-config>
                             <com.sun.management.jmxremote>true</com.sun.management.jmxremote>
```
yrodiere, line 772:
```suggestion
````

Unrelated to your PR, but I think both of these properties can be removed. As far as I can tell they're not useful.

File: pom.xml [hunk 4]
```diff
@@ -1569,10 +815,6 @@
...
```
yrodiere, line 817:
```suggestion
````
Unrelated to your PR, but I think both of these properties can be removed. As far as I can tell they're not useful.

File: pom.xml [hunk 5]
```diff
@@ -1724,119 +966,248 @@
                 </plugin>
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
-                    <artifactId>maven-gpg-plugin</artifactId>
-                    <version>${version.gpg.plugin}</version>
+                    <artifactId>maven-antrun-plugin</artifactId>
+                    <version>${version.maven.antrun.plugin}</version>
+                </plugin>
```
yrodiere, line 971: This plugin was used for applying patches to generate the `-orm6`/`-jakarta` artifacts, it can be safely removed (along with the version property).

build/parents/build/pom.xml
```diff
@@ -0,0 +1,1522 @@
...
+
+    <artifactId>hibernate-search-parent-build</artifactId>
+    <name>Hibernate Search - Parent Build</name>
+    <description>
+        Provides dependency and plugin management required during the build time.
+    </description>
+    <packaging>pom</packaging>
+
+    <properties>
+        <!-- >>> Common -->
```
yrodiere, line 21:
```suggestion
        <!-- Main dependencies -->

        <!-- >>> Common -->
```

build/parents/build/pom.xml
```diff
@@ -0,0 +1,1522 @@
...
+        <version.org.wildfly.security.wildfly-security-manager>1.1.2.Final</version.org.wildfly.security.wildfly-security-manager>
+        <version.org.google.guava>31.1-jre</version.org.google.guava>
+
+        <!-- >>> Spring integration tests -->
+        <version.org.springframework.boot>3.1.1</version.org.springframework.boot>
+        <!-- Spring Boot 3 Atomikos starter and related libs version: -->
+        <version.com.atomikos>6.0.0</version.com.atomikos>
+
+        <!-- Public modules must be signed. -->
+        <release.gpg.signing.skip>false</release.gpg.signing.skip>
```
yrodiere, line 170:
```suggestion
        <!-- Only public modules must be signed. -->
        <release.gpg.signing.skip>false</release.gpg.signing.skip>
```

We'll probably want to move this to the root POM when we start flattening POMs and no longer publish the root POM.

build/parents/build/pom.xml
```diff
@@ -0,0 +1,1522 @@
...
+            <plugins>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-compiler-plugin</artifactId>
+                    <executions>
+                        <execution>
+                            <id>default-compile</id>
+                            <configuration>
+                                <fork>true</fork>
+                                <executable>${java-version.main.compiler}</executable>
```
yrodiere, line 799: These two should probably be in the root POM?

build/parents/build/pom.xml
```diff
@@ -0,0 +1,1522 @@
...
+                                    </path>
+                                </annotationProcessorPaths>
+                            </configuration>
+                        </execution>
+                        <execution>
+                            <id>default-testCompile</id>
+                            <configuration>
+                                <fork>true</fork>
+                                <proc>none</proc>
+                                <executable>${java-version.test.compiler}</executable>
```
yrodiere, line 819: These should probably be in the root POM?

build/parents/build/pom.xml
```diff
@@ -0,0 +1,1522 @@
...
+                 -->
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-surefire-plugin</artifactId>
+                    <configuration>
+                        <systemPropertyVariables>
+                            <org.hibernate.search.test.property-set-in-maven-config>someValue</org.hibernate.search.test.property-set-in-maven-config>
+                            <com.sun.management.jmxremote>true</com.sun.management.jmxremote>
+                            <org.hibernate.search.enable_performance_tests>${test.performance.enable}</org.hibernate.search.enable_performance_tests>
+                            <org.hibernate.search.fail_on_unreleased_service>true</org.hibernate.search.fail_on_unreleased_service>
```
yrodiere, line 1075:
```suggestion
                            <org.hibernate.search.enable_performance_tests>${test.performance.enable}</org.hibernate.search.enable_performance_tests>
```

build/parents/build/pom.xml
```diff
@@ -0,0 +1,1522 @@
...
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-failsafe-plugin</artifactId>
+                    <configuration>
+                        <systemPropertyVariables>
+                            <org.hibernate.search.test.property-set-in-maven-config>someValue</org.hibernate.search.test.property-set-in-maven-config>
+                            <com.sun.management.jmxremote>true</com.sun.management.jmxremote>
+                            <org.hibernate.search.version>${project.version}</org.hibernate.search.version>
+                            <org.hibernate.search.enable_performance_tests>${test.performance.enable}</org.hibernate.search.enable_performance_tests>
+                            <org.hibernate.search.fail_on_unreleased_service>true</org.hibernate.search.fail_on_unreleased_service>
```
yrodiere, line 1088:
```suggestion
                            <org.hibernate.search.version>${project.version}</org.hibernate.search.version>
                            <org.hibernate.search.enable_performance_tests>${test.performance.enable}</org.hibernate.search.enable_performance_tests>
```

build/parents/build/pom.xml
```diff
@@ -0,0 +1,1522 @@
...
+                    </dependencies>
+                    <configuration>
+                        <!-- store outside of target to speed up formatting when mvn clean is used -->
+                        <cachedir>.cache/formatter-maven-plugin-${version.formatter-maven-plugin}</cachedir>
+                        <configFile>hibernate_search_style.xml</configFile>
+                        <lineEnding>LF</lineEnding>
+                        <skip>${format.skip}</skip>
+                        <removeTrailingWhitespace>true</removeTrailingWhitespace>
+                    </configuration>
+                </plugin>
```
yrodiere, line 1139: This is already in the root POM, the only difference AFAICT being `<skip>${format.skip}</skip>`. Might as well move that to the root POM as well, and remove this entirely?

build/parents/build/pom.xml
```diff
@@ -0,0 +1,1522 @@
...
+                                        </pluginExecution>
+                                    </pluginExecutions>
+                                </lifecycleMappingMetadata>
+                            </configuration>
+                        </plugin>
+                    </plugins>
+                </pluginManagement>
+            </build>
+        </profile>
+    </profiles>
```
yrodiere, line 1521: These probably belong in the root POM? Only the last parts of the IDEA profile (dependencies and build) really need to be here, and you can just split the profile.

pom.xml
```diff
@@ -153,8 +153,8 @@
     <modules>
         <module>build/enforcer</module>
         <module>bom/public</module>
-        <module>bom/build</module>
         <module>build/config</module>
+        <module>build/parents/build</module>
```
yrodiere, line 157: Technically build/config needs build/parents/build at the moment...

```suggestion
        <module>build/parents/build</module>
        <module>build/config</module>
```

pom.xml
```diff
@@ -153,8 +153,8 @@
     <modules>
         <module>build/enforcer</module>
```
yrodiere, line 154: Shouldn't this one use `build/parents/build`? Given `build/config` does... IDK?

yrodiere (commented):
Well that was a rough ride, but at least there's no major concern anymore xD

Thanks for working on this. Here's a last batch of comments.

build/parents/build/pom.xml
marko-bekhta: Yeah, I wanted to keep these IDE profiles together in one place, but this should work too 😃

yrodiere (approved):
I pushed one last commit to address some redundant configuration that I spotted.

Feel free to merge if you agree with those changes.

marko-bekhta:
Thanks!

yrodiere:
Thanks to you :)

PR #3619, opened 2023-08-01
HSEARCH-4744 Try to read a nested jar on JDK13+

marko-bekhta:

I've changed the exception from IO to Search one so that we wouldn't need a throwing bi function. Since the util tests do not run on JDK11 it seems safe to just update the test to expect that the jar will get processed OK. 
Also, I've tried these changes on a spring project -- worked fine when running as a repackaged jar.

### DIFF
--- a/integrationtest/mapper/orm-spring-uberjar/test/src/test/java/org/hibernate/search/integrationtest/spring/repackaged/test/RepackagedAppIT.java
+++ b/integrationtest/mapper/orm-spring-uberjar/test/src/test/java/org/hibernate/search/integrationtest/spring/repackaged/test/RepackagedAppIT.java
@@ -1,0 +1,71 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.integrationtest.spring.repackaged.test;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.nio.file.Path;
+import java.util.jar.JarEntry;
+
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.ProjectionConstructor;
+import org.hibernate.search.util.common.jar.impl.JandexUtils;
+
+import org.junit.Test;
+
+import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.FieldInfo;
+import org.jboss.jandex.Index;
+
+import acme.org.hibernate.search.integrationtest.spring.repackaged.model.MyEntity;
+import acme.org.hibernate.search.integrationtest.spring.repackaged.model.MyProjection;
+import org.springframework.boot.loader.jar.JarFile;
+
+public class RepackagedAppIT {
+
+	/**
+	 * Test makes sure that our utils can read the classes from the nested jar inside a Spring's repackaged uberjar.
+	 * We try to create index and then see that that index has the info we need e.g. projection/entity classes etc.
+	 */
+	@Test
+	public void canReadJar() throws Exception {
+		Path target = Path.of( this.getClass().getProtectionDomain().getCodeSource().getLocation().toURI() );
+		Path jar = target.resolve( "../lib/app.jar" );
+
+		try ( JarFile outerJar = new JarFile( jar.toFile() ) ) {
+			for ( JarEntry jarEntry : outerJar ) {
+				if ( jarEntry.getName().contains( "hibernate-search-integrationtest-spring-repackaged-model" ) ) {
+					URL innerJarURL = outerJar.getNestedJarFile( jarEntry ).getUrl();
+
+					try ( URLClassLoader isolatedClassLoader = new URLClassLoader( new URL[] { innerJarURL }, null ) ) {
+						Class<?> classInIsolatedClassLoader = isolatedClassLoader.loadClass( MyEntity.class.getName() );
+						URL location = classInIsolatedClassLoader.getProtectionDomain().getCodeSource().getLocation();
+						assertThat( location.getProtocol() ).isEqualTo( "jar" );
+						Index index = JandexUtils.readOrBuildIndex( location );
+
+						ClassInfo classInfo = index.getClassByName( DotName.createSimple( MyEntity.class.getName() ) );
+						assertThat( classInfo ).isNotNull();
+						FieldInfo name = classInfo.field( "name" );
+						assertThat( name ).isNotNull();
+						assertThat( name.type().name().toString() ).isEqualTo( String.class.getName() );
+
+						classInfo = index.getClassByName( DotName.createSimple( MyProjection.class.getName() ) );
+						assertThat( classInfo ).isNotNull();
+						assertThat(
+								classInfo.annotation( DotName.createSimple( ProjectionConstructor.class.getName() ) ) )
+								.isNotNull();
+						name = classInfo.field( "name" );
+						assertThat( name ).isNotNull();
+						assertThat( name.type().name().toString() ).isEqualTo( String.class.getName() );
+					}
+				}
+			}
+		}
+	}
+}
--- a/integrationtest/pom.xml
+++ b/integrationtest/pom.xml
@@ -42,6 +42,7 @@
             <modules>
                 <!-- Spring Boot 3 requires at minimum JDK 17. Hence, we want to ignore these modules on JDK 11: -->
                 <module>mapper/orm-spring</module>
+                <module>mapper/orm-spring-uberjar</module>
                 <module>showcase/library</module>
             </modules>
         </profile>
--- a/util/common/src/main/java/org/hibernate/search/util/common/jar/impl/CodeSource.java
+++ b/util/common/src/main/java/org/hibernate/search/util/common/jar/impl/CodeSource.java
@@ -11,6 +11,8 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.invoke.MethodHandles;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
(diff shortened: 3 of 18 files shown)


util/common/src/main/java/org/hibernate/search/util/common/jar/impl/CodeSource.java
```diff
@@ -169,11 +193,22 @@ private String extractedJarNestedPath(URI jarUri) throws IOException {
...
 			int secondPathSeparatorIndex = spec.indexOf( JAR_URI_PATH_SEPARATOR, afterPathSeparatorIndex );
-			if ( 0 <= secondPathSeparatorIndex && secondPathSeparatorIndex + JAR_URI_PATH_SEPARATOR.length() < spec.length() ) {
-				// TODO HSEARCH-4744 support reading the content of nested JARs
-				throw log.cannotOpenNestedJar( jarUri );
+			while ( 0 <= secondPathSeparatorIndex ) {
+				Path nestedPathInFileSystem = classesPathInFileSystem.resolve(
+						spec.substring( afterPathSeparatorIndex, secondPathSeparatorIndex )
+				);
+				if ( Files.isRegularFile( nestedPathInFileSystem ) ) {
+					nonDefaultFileSystem = NESTED_JAR_FILESYSTEM_CREATOR.apply( nestedPathInFileSystem, jarUri );
```
yrodiere, line 201: This is a loop. If you're overwriting the previous filesystem, it probably won't ever be closed.

I'd suggest to keep a list of filesystems to close in `CodeSource`?

util/common/src/main/java/org/hibernate/search/util/common/logging/impl/Log.java
```diff
@@ -135,7 +135,9 @@ SearchException errorInvokingStaticMember(Member member, String argumentsAsStrin
 	IOException cannotInterpretCodeSourceUrl(URL url);
 
 	@Message(id = ID_OFFSET + 18,
-			value = "Cannot open a ZIP filesystem for code source at '%1$s', because the URI points to content inside a nested JAR.")
-	IOException cannotOpenNestedJar(URI uri);
+			value = "Cannot open a ZIP filesystem for code source at '%1$s', because the URI points to content inside a nested JAR. "
+					+ "Run your application on JDK13+ to get nested JAR support, "
+					+ "or disable JAR scanning by setting a mapping configurer that calls .discoverAnnotatedTypesFromRootMappingAnnotations(false)")
```
yrodiere, line 140:
```suggestion
					+ "or disable JAR scanning by setting a mapping configurer that calls .discoverAnnotatedTypesFromRootMappingAnnotations(false)."
					+ " See the reference documentation for information about mapping configurers.")
```

yrodiere (commented):
> Since the util tests do not run on JDK11

Well that's a bummer.

>  it seems safe to just update the test to expect that the jar will get processed OK

I'd rather make sure JDK12 and below users will get a useful exception message...

>  Also, I've tried these changes on a spring project -- worked fine when running as a repackaged jar.

Do you think there's an easy way to add an integration test for that? Ideally one that runs on JDK11 as well, so that we can test the exception message in that case.

Surely Spring must have a way to run an integration test based on the actual nested JAR...

Jenkinsfile
```diff
@@ -447,7 +447,7 @@ stage('Non-default environments') {
 				withMavenWorkspace {
 					// Re-run integration tests against the JARs produced by the default build,
 					// but using a different JDK to build and run the tests.
-					mavenNonDefaultBuild buildEnv, "-f integrationtest"
+					mavenNonDefaultBuild buildEnv, "-pl " + sh(script: "./ci/list-all-integration-tests.sh", returnStdout: true).trim() + ":hibernate-search-util-common"
```
yrodiere, line 450: Or just this and remove the copy-pasted script:

```suggestion
					mavenNonDefaultBuild buildEnv, "-pl " + sh(script: "./ci/list-dependent-integration-tests.sh hibernate-search-engine", returnStdout: true).trim() + " -pl :hibernate-search-util-common"
```

util/internal/test/common/src/main/java/org/hibernate/search/util/impl/test/jar/JarTestUtils.java
```diff
@@ -53,6 +53,9 @@ public static Path toJar(TemporaryFolder temporaryFolder, Path jarOrDirectoryPat
 			URI jarUri = new URI( "jar:file", null, jarPath.toUri().getPath(), null );
 			Map<String, String> zipFsEnv = new HashMap<>();
 			zipFsEnv.put( "create", "true" );
+			// tests that run on JDK11 and try to use Spring Boot 2.7 cannot work with a compressed jar,
+			// so we try to create it with no compression:
+			zipFsEnv.put( "noCompression", "true" );
```
yrodiere, line 58: Seems odd, but ok. Maybe just do this on JDK 11 then?

yrodiere (commented):
Hmm okay, but executing the build on hibernate-search-util-common is cheating a bit, as the main code will be recompiled with JDK 17. Who knows how this could change the behavior.

> > Do you think there's an easy way to add an integration test for that? Ideally one that runs on JDK11 as well, so that we can test the exception message in that case.
> 
> I didn't see anything that would run the tests against the repackaged jar ... so I suppose what we can try is to add a module that creates that jar, and then in the test just execute the jar 😄 should I try that?

I'm a bit surprised this is not an option Spring Boot offers... Let's discuss this on Zulip

integrationtest/pom.xml
```diff
@@ -43,6 +43,7 @@
                 <!-- Spring Boot 3 requires at minimum JDK 17. Hence, we want to ignore these modules on JDK 11: -->
                 <module>mapper/orm-spring</module>
                 <module>showcase/library</module>
+                <module>spring</module>
```
yrodiere, line 46: Considering that we already have `mapper/orm-spring`, and that this new module uses Hibernate ORM as well... maybe it would be clearer if this was in `mapper/orm-spring-uberjar` or something like this?

integrationtest/spring/application/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+    <artifactId>hibernate-search-integrationtest-spring-repackaged-application</artifactId>
+
+    <name>Hibernate Search ITs - Spring Repackaged JAR Application</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.hibernate.search</groupId>
+            <artifactId>hibernate-search-integrationtest-spring-repackaged-model</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
yrodiere, line 19: Do you need dependencies to test this? I would have expected the uberjar to put the application classes in a nested jar as well... ?

If you do need dependencies, that's actually good news as this means far fewer applications are affected than I originally thought.

integrationtest/spring/application/src/main/resources/application.yaml
```diff
@@ -0,0 +1,16 @@
+spring.jpa:
+  properties:
+    hibernate:
+      dialect: org.hibernate.dialect.H2Dialect
+    hibernate.search:
+      backend:
+        type: lucene
```
yrodiere, line 7: This shouldn't be necessary as you only have the Lucene backend in your dependencies: it'll be picked by default.

integrationtest/spring/model/src/main/java/acme/org/hibernate/search/integrationtest/spring/repackaged/model/MyOtherEntity.java
```diff
@@ -0,0 +1,24 @@
...
+
+import jakarta.persistence.Entity;
+import jakarta.persistence.Id;
+
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.FullTextField;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.Indexed;
+
+@Entity
+@Indexed
+public class MyOtherEntity {
```
yrodiere, line 17: Maybe also include a projection class somewhere? After all, that's the only point of classpath scanning.

integrationtest/spring/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+
+    <modules>
+        <module>model</module>
+        <module>application</module>
+        <module>test</module>
+    </modules>
+
+    <properties>
+        <test.elasticsearch.run.skip>${test.elasticsearch.connection.uris.defined}</test.elasticsearch.run.skip>
+        <test.database.run.skip>false</test.database.run.skip>
```
yrodiere, line 24: This is supposed to enable starting DB containers using docker-maven-plugin. But from what I can see you hardcoded H2 as the database for this project, so you don't need this.

IMO you should either adapt the project configuration to work with non-H2 databases (probably not worth it), or use `true` here with a comment explaining that we just test H2.

integrationtest/spring/application/src/main/java/org/hibernate/search/integrationtest/spring/repackaged/application/Config.java
```diff
@@ -0,0 +1,52 @@
...
+				Search.session( entityManager )
+						.search( MyEntity.class )
+						.select( f -> f.id() )
+						.where( f -> f.matchAll() )
+						.fetchTotalHitCount();
+				transaction.commit();
+				System.out.println( "Hibernate Search read the nested JAR." );
+			}
+		}
+	}
```
yrodiere, line 51: Is this actually executed in the build?

integrationtest/spring/application/pom.xml
marko-bekhta: Yup. That's a +. Spring has that boot-inf folder in the jar, and then there is a classes folder with classes from the "current module" and a lib folder with all the jars, so yeah, it should only affect those who put their entities etc. in some other jar.

integrationtest/spring/pom.xml
marko-bekhta: ohh good point! I thought I've removed these. Yeah no need to have this for various backends or DBs as the test isn't really dependant on them.

integrationtest/spring/application/src/main/java/org/hibernate/search/integrationtest/spring/repackaged/application/Config.java
marko-bekhta: OK, an updated version 😃 
I've added a maven exec that runs this jar -- since the way it's written, it would start the context and shutdown; we can execute it just from the maven build itself... I also added a simple empty spring test to this app module that would just make sure that the context is correctly loaded.

integrationtest/mapper/orm-spring-uberjar/test/src/test/java/org/hibernate/search/integrationtest/spring/repackaged/test/RepackagedAppIT.java
```diff
@@ -0,0 +1,71 @@
...
+import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.FieldInfo;
+import org.jboss.jandex.Index;
+
+import acme.org.hibernate.search.integrationtest.spring.repackaged.model.MyEntity;
+import acme.org.hibernate.search.integrationtest.spring.repackaged.model.MyProjection;
+import org.springframework.boot.loader.jar.JarFile;
+
+public class RepackagedAppIT {
```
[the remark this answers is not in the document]
yrodiere: Actually scratch that, since the end-to-end integration test just executes an app, so it might not detect regressions such as "we open the nested JAR but it's empty"

yrodiere (approved):

PR #3620, opened 2023-08-03
[6.2] HSEARCH-4907 Follow-up: Clarify that hibernate.search.automatic_indexing.enable_dirty_check still works for now

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4907

The previous wording was apparently confusing and made it seem like the property is already ignored; see https://github.com/quarkusio/quarkus-updates/pull/47#discussion_r1280427545


marko-bekhta (approved):
+1 this gives a better explanation 😃

marko-bekhta (approved):

PR #3626, opened 2023-08-07
Follow-up on HSEARCH-4744 for branch 6.2 - fix JDK 8 compilation error in tests

yrodiere:

Follows up on #3622

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -161,7 +161,7 @@ stage('Configure') {
 					// See http://www.oracle.com/technetwork/java/javase/eol-135779.html
 					// We only run the tests with JDK8, but we compile them with JDK17 (with --release 8).
 					new JdkBuildEnvironment(version: '8', testLauncherTool: 'OpenJDK 8 Latest',
-							condition: TestCondition.AFTER_MERGE),
+							condition: TestCondition.BEFORE_MERGE),
 					new JdkBuildEnvironment(version: '11', testCompilerTool: 'OpenJDK 11 Latest',
 							condition: TestCondition.AFTER_MERGE),
 					new JdkBuildEnvironment(version: '17', testCompilerTool: 'OpenJDK 17 Latest',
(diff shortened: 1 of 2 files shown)


marko-bekhta (approved):
thanks!

yrodiere:
So, it turns out the shell execution from Maven to run the repackaged Spring app isn't great, as:

1. It isn't skipped when using skipITs
2. It isn't considered as a test by e.g. Jenkins, which fails the build when the execution fails, but doesn't mark tests as failed: Screenshot from 2023-08-08 08-15-41

I'll try to send a follow-up PR to fix that on `main` and `6.2`.

marko-bekhta:
hmm so for 6.2 the code is only set for jdk11: https://github.com/hibernate/hibernate-search/commit/98481f8471663259bb9bc7bcc2d9b8984777f3b6#diff-4cb9c3ffe73349d56259137c76c673e09359aa546c712d955711792dfb9ebec0R116-R132 
I forgot about the jdk8 entirely 🙈

yrodiere:
Yes there's that, and also we were using the ORM 6 BOM in these tests, which is no good in Search 6.2 :x

integrationtest/mapper/orm-spring-uberjar/pom.xml
```diff
@@ -31,31 +31,24 @@
...
-                <scope>import</scope>
-            </dependency>
             <dependency>
                 <groupId>org.springframework.boot</groupId>
                 <artifactId>spring-boot-dependencies</artifactId>
                 <version>${version.org.springframework.boot}</version>
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+            <!-- The Spring BOM uses a version of bytebuddy that's too old for Mockito to work correctly -->
```
marko-bekhta, line 41: So we'll leave the ORM version to whatever spring has, and since this uberjar is not using mockito do we need to override the byte buddy?

marko-bekhta:
> Yes there's that, and also we were using the ORM 6 BOM in these tests, which is no good in Search 6.2 :x

ohh so 5.6 didn't yet have a BOM? 🙈

yrodiere:
> > Yes there's that, and also we were using the ORM 6 BOM in these tests, which is no good in Search 6.2 :x
> 
> ohh so 5.6 didn't yet have a BOM? 🙈

Not that I know, but in any case we were using the Maven property that holds the ORM 6 version.

integrationtest/mapper/orm-spring-uberjar/pom.xml
yrodiere: You're right on both counts, but I just copy-pasted from other Spring tests. If you want to fix the setup feel free to send another PR.

PR #3627, opened 2023-08-07
Followup HSEARCH-4861 Make tests work on nondefault JDKs

marko-bekhta:

follows up on https://github.com/hibernate/hibernate-search/pull/3618

### DIFF
--- a/integrationtest/mapper/orm-spring-uberjar/test/pom.xml
+++ b/integrationtest/mapper/orm-spring-uberjar/test/pom.xml
@@ -87,14 +87,14 @@
                     <execution>
                         <phase>verify</phase>
                         <goals>
-                            <goal>java</goal>
+                            <goal>exec</goal>
                         </goals>
                         <configuration>
                             <!--
                                 We don't want to execute the spring jar if we are not running the tests.
                              -->
                             <skip>${failsafe.spring.skip}</skip>
-                            <mainClass>org.springframework.boot.loader.JarLauncher</mainClass>
+                            <executable>${java-version.test.launcher}</executable>
                             <arguments>
                                 <argument>-jar</argument>
                                 <argument>${project.build.directory}/lib/app.jar</argument>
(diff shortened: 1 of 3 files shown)


integrationtest/mapper/orm-spring-uberjar/test/pom.xml
```diff
@@ -87,14 +87,14 @@
...
-                            <goal>java</goal>
+                            <goal>exec</goal>
                         </goals>
                         <configuration>
                             <!--
                                 We don't want to execute the spring jar if we are not running the tests.
                              -->
                             <skip>${failsafe.spring.skip}</skip>
-                            <mainClass>org.springframework.boot.loader.JarLauncher</mainClass>
+                            <executable>${java-version.test.launcher}</executable>
```
yrodiere, line 97: We might want this on 6.2 as well, be it only for consistency? Just so we don't get surprised if we end up enabling Spring Boot 3 + JDK 21 tests on that branch one day.

pom.xml
```diff
@@ -338,6 +338,21 @@
...
+        <java-version.test.java11.add-test-source-phase>generate-test-sources</java-version.test.java11.add-test-source-phase>
+        <java-version.test.java17.add-test-source-phase>generate-test-sources</java-version.test.java17.add-test-source-phase>
+
+        <!--
+            This is an explicit setting to control compiler plugin execution for the no `-parameters` flag case.
+            Needed as when `testIncludes` doesn't match any test classes, it brings all test sources and re-compiles them.
+         -->
+        <maven.compiler.testSources.noParameterCompilation.skip>true</maven.compiler.testSources.noParameterCompilation.skip>
+        <java-version.test.java17.noParameters.skip>${maven.compiler.testSources.noParameterCompilation.skip}</java-version.test.java17.noParameters.skip>
+
```
yrodiere, line 355: I'd say we should move the corresponding plugin executions/configuration up as well, be it only for clarity?

PR #3630, opened 2023-08-08
HSEARCH-4744 follow-up: Run all Spring repackaged application tests within failsafe 

yrodiere:
So that test results are reported like any other test.

Follows up on #3619 , #3626 

Sorry I had to move code around, but the tests are essentially the same, just organized differently.

I *think* I made one improvement to the no-repackaged tests, because I *think* they did not actually run Hibernate Search before.

Backported to 6.2 in #3631


marko-bekhta (approved):
👍🏻

yrodiere:
Jenkins, retest this please

PR #3633, opened 2023-08-09
HSEARCH-4893 Switch to io.github.dmlloyd.module-info:module-info plugin

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4893

hey Yoann, I've started looking if we can use the plugin. Here is a converted util module. I have added a few things to the plugin, I'll add inline comments on that.


pom.xml
```diff
@@ -1048,6 +1049,23 @@
...
+                    <version>${version.module-info}</version>
+                    <executions>
+                        <execution>
+                            <id>module-info</id>
+                            <phase>prepare-package</phase>
+                            <goals>
+                                <goal>generate</goal>
+                            </goals>
+                            <configuration>
+                                <exportExcludes>^.*\b(impl|internal|_private|private_)\b.*$</exportExcludes>
```
marko-bekhta, line 1064: This is a new property. The plugin as it is now would not exclude `impl` packages, so I've added this regexp to make it configurable. An alternative would be to disable the automatic listing of exports and add them manually.

util/common/src/main/java/module-info.yml
```diff
@@ -0,0 +1,21 @@
+name: org.hibernate.search.util.common
+
+requires:
+  - module: org.jboss.jandex
+  - module: org.jboss.logging
+
+exports:
+  - package: org\.hibernate\.search\.util\.common\b(\..+)*\.impl
+    pattern: true
```
marko-bekhta, line 9: This also is a new option; without it we'd need to list all those `impl` packages and add the same `to`s to each

marko-bekhta:
```
[INFO] Reactor Summary for Hibernate Search ITs - Java - Modules - ORM/Elasticsearch 7.0.0-SNAPSHOT:
[INFO] 
[INFO] Hibernate Search ITs - Java - Modules - ORM/Elasticsearch SUCCESS [ 14.636 s]
[INFO] Hibernate Search ITs - Java - Modules - ORM/Lucene . SUCCESS [  2.976 s]
[INFO] Hibernate Search ITs - Java - Modules - ORM/Elasticsearch outbox polling SUCCESS [ 14.705 s]
[INFO] Hibernate Search ITs - Java - Modules - POJO/Elasticsearch SUCCESS [ 12.500 s]
[INFO] Hibernate Search ITs - Java - Modules - POJO/Lucene  SUCCESS [  2.091 s]
[INFO] BUILD SUCCESS
[INFO] Total time:  47.118 s
```
Module tests pass.

BTW jberet module can now have a module-info since they've fixed the automatic module name.

I've also inspected a few of the generated module-infos and they are mostly looking the same.

backend/elasticsearch-aws/src/main/java/module-info.yml
```diff
@@ -0,0 +1,32 @@
...
+  - module: org.apache.httpcomponents.httpclient
+  - module: software.amazon.awssdk.regions
+
+  - module: org.apache.httpcomponents.httpasyncclient
+  - module: org.apache.httpcomponents.httpclient
+  - module: org.apache.httpcomponents.httpcore
+  - module: software.amazon.awssdk.regions
+  - module: software.amazon.awssdk.auth
+  - module: software.amazon.awssdk.http
+  - module: org.jboss.logging
```
yrodiere, line 14: Wait, I though this new plugin could auto-generate that kind of things, like moditect does... did I misunderstand?

Having to maintain these lists manually would be a big step back from what we had with moditect... Ideally we'd want the list automatically generated, with the option to amend it for false-positives/false-negatives.

engine/src/main/java/module-info.yml
```diff
@@ -0,0 +1,21 @@
...
+exports:
+  - package: org\.hibernate\.search(\..+)*\.(logging|reporting)\.impl
+    pattern: true
+    to:
+      - org.jboss.logging
+      - org.hibernate.search.engine
+  - package: org\.hibernate\.search(\..+)*\.impl
+    pattern: true
+    to:
+      - org.hibernate.search.engine
```
yrodiere, line 21: Do we have to repeat these for every module? Or is there some concept of includes that we could leverage?

yrodiere (commented):
Thanks for experimenting with this. A few comments below...

backend/elasticsearch-aws/src/main/java/module-info.yml
marko-bekhta: >  // todo: detect requires

^ on autodetecting the requires 😃. So I suppose we could try to add that in the plugin ...

engine/src/main/java/module-info.yml
marko-bekhta: I haven't seen anything in the docs directly. But it is using `snakeyaml` so if that one can understand some sort of includes we can give that a try

backend/elasticsearch-aws/src/main/java/module-info.yml
yrodiere: I suppose we could, but if we end up working a lot to get this new plugin up to speed with moditect, then we might need to question whether it's really worth it... let's discuss this on Zulip, you are best-placed to judge of that.

PR #3636, opened 2023-08-10
HSEARCH-4913 Generate module-info for `hibernate-search-mapper-orm-batch-jsr352-jberet`

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4913

### DIFF
--- a/build/parents/public/pom.xml
+++ b/build/parents/public/pom.xml
@@ -198,7 +198,6 @@
                             <goal>add-module-info</goal>
                         </goals>
                         <configuration>
-                            <jvmVersion>9</jvmVersion>
                             <module>
                                 <moduleInfo>
                                     <name>${java.module.name}</name>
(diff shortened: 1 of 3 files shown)


build/parents/public/pom.xml
```diff
@@ -231,7 +230,12 @@
                                 </moduleInfo>
                             </module>
                             <jdepsExtraArgs>
-                                <arg>--multi-release=9</arg>
+                                <!--
+                                    Running jdpes tool will fail without this option, if there are any multi-release jars
```
yrodiere, line 234:
```suggestion
                                    Running the jdeps tool will fail without this option, if there are any multi-release jars
```

build/parents/public/pom.xml
```diff
@@ -231,7 +230,12 @@
...
+                                    on the path. And since some modules might have such jars, e.g. util-common has Jandex,
+                                    it's safer to add this additional option:
```
yrodiere, line 236: Okay. Does this mean we generate a multi-release JAR, or just that we pass an extra argument to `jdeps` during the build, with no impact on the produced JAR?

yrodiere (approved):
Works for me, thanks, though there's a typo :)

build/parents/public/pom.xml
marko-bekhta: From what it looks: only an extra argument. I've looked inside the resulting jar - module-info is just in the root, and there's no additional versions folder for multi-release

PR #3651, opened 2023-08-14
Update build dependencies

marko-bekhta:
since there are a lot of updates today, let's see if they will build OK

### DIFF
--- a/.github/hibernate-github-bot.yml
+++ b/.github/hibernate-github-bot.yml
@@ -4,7 +4,7 @@ jira:
   ignore:
       # Allow build dependency upgrades i.e. maven plugin version upgrades to not follow the PR formatting rules:
     - user: dependabot[bot]
-      titlePattern: ".*\\bmaven\\b.*\\bplugin\\b.*"
+      titlePattern: ".*\\b(maven|surefire)\\b.*\\bplugin\\b.*"
       # Allow Groovy (that is only used for scripting maven plugin) version upgrades to not follow the PR formatting rules:
     - user: dependabot[bot]
       titlePattern: ".*\\bgroovy-jsr223\\b.*"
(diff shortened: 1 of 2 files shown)


yrodiere (approved):
Thanks for taking care of this!

marko-bekhta:
mmm, asciidoc generation was failing... 
I've removed the dependency on jruby -- since that was the one that was causing the issue (misaligned version vs the one needed by asciidoctor). I suppose since we are not using jruby for anything but asciidoctor -- maybe it's ok to let it bring any version it needs? Also seems like their pdf lib came a long way since our last update 😃 I've tried to bump it to the latest and it seems to build PDFs without issues so let's give that a try ...

yrodiere:
> I've removed the dependency on jruby -- since that was the one that was causing the issue (misaligned version vs the one needed by asciidoctor). I suppose since we are not using jruby for anything but asciidoctor -- maybe it's ok to let it bring any version it needs? Also seems like their pdf lib came a long way since our last update smiley I've tried to bump it to the latest and it seems to build PDFs without issues so let's give that a try ...

Feel free to update that as you wish but if you make such significant changes:

* Please make sure our javascript (in the left panel TOC in particular) continues working
* Please make sure the output in both HTML and PDF stays the same
* Please consider updating dependencies similarly in https://github.com/hibernate/hibernate-asciidoctor-extensions and https://github.com/hibernate/hibernate-asciidoctor-theme, and perhaps triggering a release (on Jenkins CI) so that we upgrade them in Hibernate Search

yrodiere (approved):

marko-bekhta:
BTW Yoann, do we want to backport any of these build-stuff-upgrades to 6.2? or we'll only do the ES upgrades?

yrodiere:
> or we'll only do the ES upgrades?

That. No need to backport everything to 6.2, we'll only do it for important bugfixes or security issues, and very exceptionally for features (like Elasticsearch 8.9 compatibility, but not sure we'll do ES 9.0).

PR #3654, opened 2023-08-15
HSEARCH-4910 Update Elasticsearch image config for tests

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4910

Looking more at those templates being loaded, there's also `stack.templates.enabled` (https://www.elastic.co/guide/en/elasticsearch/reference/current/index-management-settings.html), but it seems to have no effect if passed as env variable (I guess it should be a part of elasticsearch.yml config? or maybe it just not related to the templates being loaded 😃)... 
for my local start with the changes:
```
[INFO] DOCKER> [elastic/elasticsearch:8.9.0] "elasticsearch": Waited on url http://localhost:9200 5609 ms
```
without the changes:
```
[INFO] DOCKER> [elastic/elasticsearch:8.9.0] "elasticsearch": Waited on url http://localhost:9200 6147 ms
```
soooo hopefully will be a tod better for CI builds too...

### DIFF
--- a/build/parents/integrationtest/pom.xml
+++ b/build/parents/integrationtest/pom.xml
@@ -93,16 +93,6 @@
                                              it's not practical for testing, so we disable that.
                                          -->
                                         <xpack.security.enabled>false</xpack.security.enabled>
-                                        <!-- Disable some features that are not needed in our tests and just slow down startup -->
-                                        <xpack.profiling.enabled>false</xpack.profiling.enabled>
-                                        <xpack.monitoring.templates.enabled>false</xpack.monitoring.templates.enabled>
-                                        <xpack.ml.enabled>false</xpack.ml.enabled>
-                                        <xpack.watcher.enabled>false</xpack.watcher.enabled>
-                                        <xpack.ent_search.enabled>false</xpack.ent_search.enabled>
-                                        <stack.templates.enabled>false</stack.templates.enabled>
-                                        <cluster.deprecation_indexing.enabled>false</cluster.deprecation_indexing.enabled>
-                                        <indices.lifecycle.history_index_enabled>false</indices.lifecycle.history_index_enabled>
-                                        <slm.history_index_enabled>false</slm.history_index_enabled>
                                         <!-- Limit the RAM usage.
                                              Recent versions of ES limit themselves to 50% of the total available RAM,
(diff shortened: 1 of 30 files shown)


yrodiere:
Thanks. I'll add a commit to raise the timeout and merge.

I'll contribute the same changes (including raising the timeout) to Quarkus, where [we're having the same problem](https://quarkusio.zulipchat.com/#narrow/stream/191168-core-team/topic/Main.20broken.3F/near/385351136).

yrodiere:
> Looking more at those templates being loaded, there's also `stack.templates.enabled` (https://www.elastic.co/guide/en/elasticsearch/reference/current/index-management-settings.html), but it seems to have no effect if passed as env variable (I guess it should be a part of elasticsearch.yml config? or maybe it just not related to the templates being loaded smiley)...

It did have an effect for me, so I'll add it to your commit

marko-bekhta:
thanks 😃

yrodiere:
As I feared:

```
"error.type":"java.lang.IllegalArgumentException","error.message":"unknown setting [xpack.profiling.enabled] please check that any required plugins are installed, or check the breaking changes documentation for removed settings
```

I'll disable these options on Elasticsearch 8.7 and earlier.

yrodiere:
@marko-bekhta I went with version-specific settings in the end, see the last commit... Not great, but not terrible I guess.

build/parents/integrationtest/pom.xml
```diff
@@ -493,6 +484,20 @@
             </properties>
         </profile>
 
+        <profile>
+            <id>elasticsearch-container-disable-useless-features</id>
```
marko-bekhta, line 488: Mmmm, so this one is a bit confusing... 
from what it looks, we'd always pass the env file as long as we start an ES container.

marko-bekhta:
> I went with version-specific settings in the end, see the last commit... Not great, but not terrible I guess.

well.. unless there's a `<start.under.one.sec>true</start.under.one.sec>` property it's probably the best we can do 😄 so +1 😃

build/parents/integrationtest/pom.xml
yrodiere: Ah right, that was an earlier attempt, I should remove it.

PR #3667, opened 2023-08-21
HSEARCH-4917 Group build dependency updates in the Dependabot's config

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4917

I've tried to use `|` in the patterns, thinking it would allow me to combine some, but as you can see ... it didn't work, so patterns are listed in a more dependency-explicit way 😃 
I've also removed the jruby rule as now we have a config that does not need jruby dependency. And as we've upgraded our spring tests we can now do some upgrades of it from time to time, hence that version limit for it is gone. While testing it on my repo `org.bsc.maven:maven-processor-plugin` kept coming up with a jre8 version... so there's an ignore rule for it now too 😈😃

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -53,17 +53,12 @@ updates:
       # AWS SDK releases way too often (every week?); ignore all patch updates
       - dependency-name: "software.amazon.awssdk:*"
         update-types: ["version-update:semver-patch"]
-      # JRuby releases way too often (every two weeks?) and is only used during the build; ignore all patch updates
-      - dependency-name: "org.jruby:jruby-complete"
-        update-types: ["version-update:semver-patch"]
       # Groovy is only used during the build; ignore all patch updates
       - dependency-name: "org.apache.groovy:groovy-jsr223"
         update-types: [ "version-update:semver-patch" ]
       # We don't care that much about being on the very latest version of some integration test dependencies
       - dependency-name: "org.springframework.boot:*"
         update-types: [ "version-update:semver-patch" ]
-        # ignore spring boot 3+ for now as some required dependencies are not compatible with this version yet:
-        versions: ["[3.0.0,)"]
       # We strictly align these dependencies on the version used in Hibernate ORM.
       - dependency-name: "io.smallrye:jandex"
       - dependency-name: "jakarta.persistence:jakarta.persistence-api"
@@ -86,3 +81,7 @@ updates:
         update-types: ["version-update:semver-major"]
       # We only define a maven's minimum version, so we don't need it to be updated to the latest:
       - dependency-name: "org.apache.maven:maven-core"
+      # This dependency uses classifiers (-jdk8, ...) in its version, and dependabot can't decide which versions are relevant.
+      # See https://github.com/dependabot/dependabot-core/issues/4028
+      # And, additionally, it is only for performance testing.
+      - dependency-name: "org.bsc.maven:maven-processor-plugin"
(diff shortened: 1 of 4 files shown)


.github/dependabot.yml
```diff
@@ -13,24 +13,52 @@ updates:
...
         update-types: ["version-update:semver-patch"]
-      # JRuby releases way too often (every two weeks?) and is only used during the build; ignore all patch updates
-      - dependency-name: "org.jruby:jruby-complete"
-        update-types: ["version-update:semver-patch"]
       # Groovy is only used during the build; ignore all patch updates
       - dependency-name: "org.apache.groovy:groovy-jsr223"
         update-types: [ "version-update:semver-patch" ]
       # We don't care that much about being on the very latest version of some integration test dependencies
       - dependency-name: "org.springframework.boot:*"
         update-types: [ "version-update:semver-patch" ]
```
yrodiere, line 61: Shouldn't this move to the `build-dependencies` group as well? Spring really is a build dependency, we don't actually publish any artifact that depends on it.

.github/dependabot.yml
```diff
@@ -13,24 +13,52 @@ updates:
...
+      # This group combines all build-only dependencies. Published artifacts do not depend on them.
+      # Grouping such dependencies will make Dependabot create PRs with a branch name
+      # following the pattern (`dependabot/maven/build-dependencies-.*`)
+      # and with a title like `Bump the build-dependencies group with 8 updates` that we can easily
+      # use for Hibernate Bot rules.
+      build-dependencies:
+        patterns:
+          # Maven plugin patterns:
+          - "*plugin*"
+          - "*maven*"
```
yrodiere, line 26: I think the first one is a bit dangerous, as it's not unthinkable that a dependency of ours would have plugin in its name. I could imagine a `jackson-something-plugin` dependency in the Elasticsearch Sniffer, for example.

Maybe this would be more reasonable?

```suggestion
          - "*maven*plugin*"
```

yrodiere (commented):
Just a few comments, otherwise LGTM :)

.github/dependabot.yml
marko-bekhta: mmm... probably yes 😃. We had this rule before:
```
".*\\b(maven|surefire)\\b.*\\bplugin\\b.*"
```
but as we have a line for `surefire`, it is probably safe to combine it as you are suggesting. And then we'll wait and see what Dependabot will come up with that doesn't fit the patterns 😈😃

.github/dependabot.yml
marko-bekhta: hmm, aaaaaactually .... https://github.com/marko-bekhta/hibernate-search/pull/157
it seems like Dependabot is trying to upgrade the spring version where we don't want it to be upgraded ... maybe let's ignore spring upgrades in general?

yrodiere: That's annoying. Would using a separate property help? I.e. define `                <version.org.springframework.boot.v2>2.7.14</version.org.springframework.boot.v2>` in the build-parent POM and use ` <version.org.springframework.boot>${version.org.springframework.boot.v2}</version.org.springframework.boot>` in the profile... maybe that'd confuse dependabot in just the right way?

Otherwise, sure, +1 to ignore spring upgrades if there's no workaround.

marko-bekhta: or I just drop that profile 😄 
this was coming from the profile in common utils, but since we've ended up running tests in a spring IT module we weren't actually rebuilding common-utils sooooo, that should work too 😃 🤞🏻

yrodiere: Good point!

yrodiere (approved):

PR #3668, opened 2023-08-22
HSEARCH-4909 Add more info on indexing plan synchronization and coordination

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4909

I've scrolled through other log messages looking for conflicting configs and it seems mostly old/new property set at the same time, which I think we shouldn't document additionally 😃; indexing plan filter limitation is already mentioned in the reference; nothing else caught my eye 😃

### DIFF
--- a/documentation/src/main/asciidoc/public/reference/_indexing-plan.adoc
+++ b/documentation/src/main/asciidoc/public/reference/_indexing-plan.adoc
@@ -113,10 +113,12 @@ include::../components/_writing-reading-intro-note.adoc[]
 
 [WARNING]
 ====
-Indexing synchronization is only relevant when <<coordination-none,coordination is disabled>>.
-
-With coordination strategies such as <<coordination-outbox-polling,`outbox-polling`>>,
-indexing happens in background threads and is always asynchronous.
+When using the <<coordination-outbox-polling,`outbox-polling` coordination strategy>>,
+the actual indexing plan performing the index changes
+is created asynchronously in a background thread.
+Because of that, with that coordination strategy it does not make sense
+to set a non-default indexing plan synchronization strategy,
+and doing so will lead to an exception on startup.
 ====
 
 When a transaction is committed (<<mapper-orm,Hibernate ORM integration>>)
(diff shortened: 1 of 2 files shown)


documentation/src/main/asciidoc/public/reference/_indexing-plan.adoc
```diff
@@ -113,10 +113,17 @@ include::../components/_writing-reading-intro-note.adoc[]
...
+is used, when it comes to indexing, Hibernate Search relies on indexing plans.
+Depending on the coordination strategy a plan resulting in index changes may be constructed within the same
+session, where the entities have been modified (when <<coordination-none,coordination is disabled>>),
+or with an entirely different session in some background thread (with <<coordination-outbox-polling,`outbox-polling`>> enabled).
+
+In the first scenario it makes sense to control how the indexing plan synchronization happens,
+since it has a direct impact on the application thread behavior.
+While in the second one, indexing happens in background threads and is always asynchronous.
+Because of that, setting a non-default indexing plan synchronization strategy, with the
+<<coordination-outbox-polling,`outbox-polling`>> coordination being enabled, will lead to an exception at the startup.
```
yrodiere, line 126: This is probably a bit too long for a warning: if we want people to read it, we need to keep it short and to the point :D 

```suggestion
When using the <<coordination-outbox-polling,`outbox-polling` coordination strategy>>,
the actual indexing plan performing the index changes
is created asynchronously in a background thread.

Because of that, with that coordination strategy it does not make sense
to set a non-default indexing plan synchronization strategy,
and doing so will lead to an exception on startup.
```

documentation/src/main/asciidoc/public/reference/_indexing-plan.adoc
```diff
@@ -113,10 +113,17 @@ include::../components/_writing-reading-intro-note.adoc[]
 
 [WARNING]
 ====
-Indexing synchronization is only relevant when <<coordination-none,coordination is disabled>>.
-
-With coordination strategies such as <<coordination-outbox-polling,`outbox-polling`>>,
-indexing happens in background threads and is always asynchronous.
+No matter if the <<coordination-none,coordination is disabled>> or a coordination strategy such as <<coordination-outbox-polling,`outbox-polling`>>
+is used, when it comes to indexing, Hibernate Search relies on indexing plans.
```
yrodiere, line 117: Not true for mass indexing though.

Also, it seems to me that the confusion in [the Discourse topic](https://discourse.hibernate.org/t/configuration-property-automatic-indexing-synchronisation-strategy-causes-exceptions-when-used-with-outbox-polling-coordination/7995) was to think that because listener-triggered indexing is disabled, indexing plan configuration does not matter. Which is wrong because indexing plans can still be created explicitly.

So I'm not sure what this sentence brings to the table... ? I'd just remove it TBH.

```suggestion
```

yrodiere (commented):
I added a few comments... TBH I'm not convinced this is making things clearer :)

yrodiere (approved):

PR #3669, opened 2023-08-23
HSEARCH-4784 Add a script to produce a list of IT modules based on changes

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4784

Hey Yoann 😃 I've started looking at this CI-related task yesterday. I thought that before I try modifying the job I'll look into  this bit first:

> Ideally we would add some code to detect when changes in a given PR have no impact on a given environment and thus skip some environments. E.g. if there are no changes in hibernate-search-backend-elasticsearch or any of its dependencies (engine, …),  then there’s no need to run tests against all flavors of OpenSearch. That would allow us to expand a bit the subset of environments we test against PRs. But maybe that should be a separate Jira issue?

I've added a script that would generate a list of modules to be tested based on the git diff. With this list, we could check if `hibernate-search-integrationtest-backend-elasticsearch` is present, and if so run those ES/OS-specific jobs... 

Opening this as a draft as it's a script only ...

### DIFF
--- a/ci/list-integration-tests-for-modified-modules.sh
+++ b/ci/list-integration-tests-for-modified-modules.sh
@@ -1,0 +1,50 @@
+#!/bin/bash -e
+# Lists test modules based on git diff check.
+
+if (( $# < 2 ))
+then
+	echo 1>&2 "Usage: $0 <main branch, which we'll compare to to identify changes> <branch with changes we want to identify>"
+	echo 1>&2 "Aborting."
+	exit 1
+fi
+
+# `exec` will echo a based dir and the artifact id.
+#   Basedir is going to be used to figure out to which module folder the modified files belong to
+#   Artifact ID is going to be used to match the basedir to an id that can be later passed as a mvn build parameter
+#
+# The output format is `artifact-id-and-basedir-output :artifactId=basedir`.
+# The prefix (`artifact-id-and-basedir-output`) is used for easier identification of the lines we are interested in from maven's output
+readarray -t lines < <(./mvnw exec:exec@print-artifact-id-and-basedir | grep artifact-id-and-basedir-output | awk '{print $2}')
+
+declare -a artifacts
+declare -a paths
+
+# We split the `:artifactId=basedir` and put them in their own arrays.
+# Basedir will be used as patterns to determine which module the change belongs to.
+for index in "${!lines[@]}"
+do
+	artifacts[index]=${lines[$index]%=*}
+	paths[index]=${lines[$index]#*=}
+	paths[index]="${paths[index]//${paths[0]}"/"/""}"
+done
+
+# get the diffs, which will produce relative paths of modified files:
+readarray -t lines < <(git --no-pager diff --name-only $1..$2)
+
+declare -A modifiedArtifactIds
+
+# using an associated array as a set to have a unique collection of modified artifact IDs
+for modifiedFile in "${lines[@]}"
+do
+  for index in "${!paths[@]}"
+    do
+      if [[ $modifiedFile =~ ${paths[$index]} ]]; then
+          modifiedArtifactIds[${artifacts[$index]}]="modified"
+      fi
+    done
+done
+
+if [ "${#modifiedArtifactIds[@]}" -ne 0 ]; then
+   $(dirname "$(realpath "$0")")/list-dependent-integration-tests.sh "${!modifiedArtifactIds[@]}"
+fi
+
--- a/pom.xml
+++ b/pom.xml
@@ -1173,6 +1173,23 @@
                             <inheritIo>true</inheritIo>
                         </configuration>
                     </execution>
+                    <execution>
+                        <id>print-artifact-id-and-basedir</id>
+                        <!--
+                            We don't want this to be executed on a regular basis.
+                            This output is used in the implementation of `list-integration-tests-for-modified-modules.sh`.
+                            See the script itself for any details.
+                         -->
+                        <phase>none</phase>
+                        <goals>
+                            <goal>exec</goal>
+                        </goals>
+                        <configuration>
+                            <executable>echo</executable>
+                            <commandlineArgs>artifact-id-and-basedir-output ${project.artifactId}=${project.basedir}</commandlineArgs>
+                            <inheritIo>true</inheritIo>
+                        </configuration>
+                    </execution>
                 </executions>
             </plugin>
             <plugin>


ci/list-integration-tests-for-modified-modules.sh
```diff
@@ -0,0 +1,50 @@
...
+  for index in "${!paths[@]}"
+    do
+      if [[ $modifiedFile =~ ${paths[$index]} ]]; then
+          modifiedArtifactIds[${artifacts[$index]}]="modified"
+      fi
+    done
+done
+
+if [ "${#modifiedArtifactIds[@]}" -ne 0 ]; then
+   $(dirname "$(realpath "$0")")/list-dependent-integration-tests.sh "${!modifiedArtifactIds[@]}"
```
yrodiere, line 48: That would be good for the main build, but not exactly what we want for HSEARCH-4784, I think?

We'd have e.g. an "Elasticsearch 7.17" job that only executes integration tests for anything that depends (even indirectly) on `hibernate-search-backend-elasticsearch`. So we'd actually want to test whether the list of modified artifacts contains `hibernate-search-backend-elasticsearch` or a dependency of `hibernate-search-backend-elasticsearch`... 

Another way of putting it: we'd want to execute integration tests that are listed in **both** `list-dependent-integration-tests.sh "${!modifiedArtifactIds[@]}"` and `list-dependent-integration-tests.sh `:hibernate-search-backend-elasticsearch`. But maybe that's what you intended to do, as a post-processing step on the result of this script?

yrodiere (commented):
Thanks for working on this!

The script looks good, though I didn't test it, I'll trust you on that one :)

Two comments though:

1. We might want to ignore some changes, like to the README.
2. More generally, you might want to be careful of changes to nested directories. What if I change a file in `build/parents/public`, is it considered a change in the root project as well?

Also, I believe there's something similar in Quarkus that relies on off-the-shelf components to determine which modules need to be rebuilt. You might want to have a look; see https://github.com/quarkusio/quarkus/blob/main/.github/workflows/ci-actions-incremental.yml

And, of course, the question now becomes how to integrate the solution into a Jenkinsfile :)

marko-bekhta:
> We might want to ignore some changes, like to the README.
More generally, you might want to be careful of changes to nested directories. What if I change a file in build/parents/public, is it considered a change in the root project as well?

The root gets ignored with this script entirely... since the basedir produces an absolute path without the  `/` at the end, it is not replaced, and as git diff generates relative paths -- there's never a match on the root project.. 
But as to the nested folders ... it's a good point! 😃 I think if we search for the longest matching pattern in that loop it'll solve the issue...

I'm going to take a look at what Quarkus is doing, thanks for the link 😃

ci/list-integration-tests-for-modified-modules.sh
yrodiere: > if we want the list of tests ... then we probably need to `mvn ... -amd -pl :list of modified modules`

I think `list-dependent-integration-tests.sh "${!modifiedArtifactIds[@]}"` like you did is probably fine for our purposes. The only difference is that we wouldn't list non-IT modules, but who cares :)

Looks like this would do what we want:

```
grep -xFf <(ci/list-integration-tests-for-modified-modules.sh $CHANGE_TARGET HEAD) <(list-dependent-integration-tests.sh :hibernate-search-backend-elasticsearch)
```

That's a rather complex command though, maybe it would be better to do this directly in your new script and rename it to something like `list-dependent-integration-tests-for-modified-modules.sh <base> <target> <module> [<other modules> ...]` or something.

yrodiere:
> The root gets ignored with this script entirely... since the basedir produces an absolute path without the / at the end, it is not replaced, and as git diff generates relative paths -- there's never a match on the root project..

But what if we change the root POM? We'll probably want to test everything then, so I wouldn't flat out ignore changes on the root project...

marko-bekhta:
> Also, I believe there's something similar in Quarkus that relies on off-the-shelf components to determine which modules need to be rebuilt. You might want to have a look; see https://github.com/quarkusio/quarkus/blob/main/.github/workflows/ci-actions-incremental.yml

I've tried to use this `gitflow-incremental-builder` in Search 😃 .... aaaaaaaaand it failed successfully 😆 
it gets very confused from us importing a `hibernate-search-bom` in `hibernate-search-parent-build` and throws stackoverflows .... But with that out of the way ... When branches are identical it doesn't want to build anything and when I've made changes to the elasticsearch-backend:
```
[INFO] Changed artifactIds (1):
[INFO] 
[INFO] - hibernate-search-backend-elasticsearch
[INFO] 
[INFO] gitflow-incremental-builder exiting...
[INFO] Reactor Build Order:
[INFO] 
[INFO] Hibernate Search Backend - Elasticsearch                           [jar]
[INFO] Hibernate Search Backend - Elasticsearch - AWS integration         [jar]
[INFO] Hibernate Search Utils - Internal - ITs - Elasticsearch            [jar]
[INFO] Hibernate Search ITs - Backend - Elasticsearch                     [jar]
[INFO] Hibernate Search ITs - POJO Standalone - Real backend              [jar]
[INFO] Hibernate Search ITs - ORM - Real backend                          [jar]
[INFO] Hibernate Search ITs - ORM - Batch JSR-352                         [jar]
[INFO] Hibernate Search ITs - Performance - Backend - Elasticsearch       [jar]
[INFO] Hibernate Search ITs - Java - Modules - ORM/Elasticsearch          [jar]
[INFO] Hibernate Search ITs - Java - Modules - ORM/Elasticsearch outbox polling [jar]
[INFO] Hibernate Search ITs - Java - Modules - POJO/Elasticsearch         [jar]
[INFO] Hibernate Search Documentation                                     [jar]
[INFO] Hibernate Search Reports                                           [pom]
```

If I change  the engine module it want's to rebuild most of the stuff. But can be limited to es-backend stuff only with:
```bash
mvn test -Dincremental -amd -pl :hibernate-search-backend-elasticsearch
```
```
[INFO] Changed artifactIds (1):
[INFO] 
[INFO] - hibernate-search-engine (but deselected)
[INFO] 
[INFO] gitflow-incremental-builder exiting...
[INFO] Reactor Build Order:
[INFO] 
[INFO] Hibernate Search Backend - Elasticsearch                           [jar]
[INFO] Hibernate Search Backend - Elasticsearch - AWS integration         [jar]
[INFO] Hibernate Search Utils - Internal - ITs - Elasticsearch            [jar]
[INFO] Hibernate Search ITs - Backend - Elasticsearch                     [jar]
[INFO] Hibernate Search ITs - POJO Standalone - Real backend              [jar]
[INFO] Hibernate Search ITs - ORM - Real backend                          [jar]
[INFO] Hibernate Search ITs - ORM - Batch JSR-352                         [jar]
[INFO] Hibernate Search ITs - Performance - Backend - Elasticsearch       [jar]
[INFO] Hibernate Search ITs - Java - Modules - ORM/Elasticsearch          [jar]
[INFO] Hibernate Search ITs - Java - Modules - ORM/Elasticsearch outbox polling [jar]
[INFO] Hibernate Search ITs - Java - Modules - POJO/Elasticsearch         [jar]
[INFO] Hibernate Search Documentation                                     [jar]
[INFO] Hibernate Search Reports                                           [pom]
[INFO] 
```
If I do the change to the lucene backend only and try the above^
```bash
mvn test -Dincremental -amd -pl :hibernate-search-backend-elasticsearch
```
it would still run all the same tests as in the previous example. But if `gib.disableSelectedProjectsHandling=true` then the result is:
```
[INFO] Changed artifactIds (1):
[INFO] 
[INFO] - hibernate-search-backend-lucene (but deselected)
[INFO] 
[INFO] gitflow-incremental-builder exiting...
[INFO] Reactor Build Order:
[INFO] 
[INFO] Hibernate Search ITs - POJO Standalone - Real backend              [jar]
[INFO] Hibernate Search ITs - ORM - Real backend                          [jar]
[INFO] Hibernate Search ITs - ORM - Batch JSR-352                         [jar]
[INFO] Hibernate Search Documentation                                     [jar]
[INFO] Hibernate Search Reports                                           [pom]
```

yrodiere:
> it gets very confused from us importing a hibernate-search-bom in hibernate-search-parent-build and throws stackoverflows

I guess we could stop doing that and just duplicate the dependency management, especially given it's only about our own modules and given all the checks you put in place.

> If I do the change to the lucene backend only and try the above^
> 
> ```shell
> mvn test -Dincremental -amd -pl :hibernate-search-backend-elasticsearch
> ```
> it would still run all the same tests as in the previous example. But if `gib.disableSelectedProjectsHandling=true` then the result is:

So... it doesn't work, right? We don't have a single command that would run nothing if only Lucene modules are impacted, while running Elasticsearch ITs if there is a change in the engine?

> EDIT: there are also some gib properties that can limit what's going to be rebuilt, so overall, it seems that we can try this in Search to 😃

You mean if we use some plugin-specific properties instead of `-pl`, we can make it work?

marko-bekhta:
> So... it doesn't work, right? We don't have a single command that would run nothing if only Lucene modules are impacted, while running Elasticsearch ITs if there is a change in the engine?

well yes, on the other hand these three:

```
[INFO] Hibernate Search ITs - POJO Standalone - Real backend              [jar]
[INFO] Hibernate Search ITs - ORM - Real backend                          [jar]
[INFO] Hibernate Search ITs - ORM - Batch JSR-352                         [jar]
```
Since they run tests for both backends in their executions ... we wouldn't be able to do much with our custom script, unless we "special-case" them somehow... 
What if instead we split those test modules ^ into 3 (actual tests, tests+lucene, tests+es) something like backend tck + backend specific IT modules ?

yrodiere:
> Since they run tests for both backends in their executions ... we wouldn't be able to do much with our custom script, unless we "special-case" them somehow...
> What if instead we split those test modules ^ into 3 (actual tests, tests+lucene, tests+es) something like backend tck + backend specific IT modules ?

Not worth the effort IMO. It's fine if those run a few extra tests that we don't need; it won't change much in the grand scheme of things.

The only real reason I could think of for doing this is cleanup.

But there's at least one test in there that requires both backends (in the same test execution), so this looks like a headach. Really, I wouldn't bother.

> there's a multiplebackends execution too... so that complicates it a bit more 😃

Yep that's the one I meant.


> > You mean if we use some plugin-specific properties instead of -pl, we can make it work?
> 
> maybe not `-pl` but those `-amd` `-am` etc (https://github.com/gitflow-incremental-builder/gitflow-incremental-builder#mvn--pl) there are some [`buildDownstream`](https://github.com/gitflow-incremental-builder/gitflow-incremental-builder#gibbuilddownstream) and [`buildUpstream`](https://github.com/gitflow-incremental-builder/gitflow-incremental-builder#gibbuildUpstream) that could change the behaviour...

Ok, I'll trust you to find the right one :) That's fine by me.

PR #3670, opened 2023-08-23
HSEARCH-4924 Configure a simple gitflow incremental build for non-default envs

marko-bekhta:

This config seems to make sense to me 😃, we'd build the entire project. Then, for any non-default env, we enable the incremental profile where this new plugin lives, and depending on the type of the env or what parameters we are passing to maven, we'd tweak the plugin properties. I've added some inline comments explaining the choice of params 😃

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -255,7 +255,7 @@ stage('Configure') {
 					new OpenSearchLocalBuildEnvironment(version: '2.6.0', condition: TestCondition.ON_DEMAND),
 					new OpenSearchLocalBuildEnvironment(version: '2.7.0', condition: TestCondition.ON_DEMAND),
 					new OpenSearchLocalBuildEnvironment(version: '2.8.0', condition: TestCondition.ON_DEMAND),
-					new OpenSearchLocalBuildEnvironment(version: '2.9.0', condition: TestCondition.AFTER_MERGE)
+					new OpenSearchLocalBuildEnvironment(version: '2.9.0', condition: TestCondition.BEFORE_MERGE)
 					// See https://opensearch.org/lines/2x.html for a list of all 2.x versions
 			],
 			esAws: [
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -149,6 +149,7 @@ import org.hibernate.jenkins.pipeline.helpers.alternative.AlternativeMultiMap
 @Field boolean enableDefaultBuild = false
 @Field boolean enableDefaultBuildIT = false
 @Field boolean deploySnapshot = false
+@Field boolean incrementalBuild = false
 
 this.helper = new JobHelper(this)
 
@@ -356,6 +357,10 @@ Some useful filters: 'default', 'jdk', 'jdk-10', 'eclipse', 'postgresql', 'elast
 			environments.content.any { key, envSet -> envSet.enabled.any { buildEnv -> buildEnv.requiresDefaultBuildArtifacts() } } ||
 			deploySnapshot
 
+	if (helper.scmSource.pullRequest) {
+		incrementalBuild = true
+	}
+
 	echo """Branch: ${helper.scmSource.branch.name}
 PR: ${helper.scmSource.pullRequest?.id}
 params.ENVIRONMENT_SET: ${params.ENVIRONMENT_SET}
@@ -366,7 +371,36 @@ Resulting execution plan:
     enableDefaultBuildIT=$enableDefaultBuildIT
     environments=${environments.enabledAsString}
     deploySnapshot=$deploySnapshot
+    incrementalBuild=$incrementalBuild
 """
+}
+
+stage('Pre-build sources') {
+	// we want this stage to only be executed when we are planning to use incremental build
+	if (!incrementalBuild) {
+		echo 'Skipping pre-building sources for non-pull request builds.'
+		helper.markStageSkipped()
+		return
+	}
+	if (!enableDefaultBuild) {
+		echo 'Skipping default build and integration tests in the default environment'
+		helper.markStageSkipped()
+		return
+	}
+	runBuildOnNode( NODE_PATTERN_BASE ) {
+		withMavenWorkspace {
+			sh """ \
+					mvn clean install \
+					--fail-at-end \
+					-Pdist -Pjqassistant -Pci-sources-check \
+					-DskipITs -DskipTests \
+			"""
+
+			dir(helper.configuration.maven.localRepositoryPath) {
+				stash name:'default-build-result', includes:"org/hibernate/search/**"
+			}
+		}
+	}
 }
 
 stage('Default build') {
@@ -377,9 +411,16 @@ stage('Default build') {
 	}
 	runBuildOnNode( NODE_PATTERN_BASE, [time: 2, unit: 'HOURS'] ) {
 		withMavenWorkspace(mavenSettingsConfig: deploySnapshot ? helper.configuration.file.deployment.maven.settingsId : null) {
+			if ( incrementalBuild ) {
+				dir(helper.configuration.maven.localRepositoryPath) {
+					unstash name:'default-build-result'
+				}
+			}
+			// If we are in the pull request we've already run JQAssistant and source formatting checks
 			String mavenArgs = """ \
 					--fail-at-end \
-					-Pdist -Pcoverage -Pjqassistant -Pci-sources-check \
+					-Pdist -Pcoverage \
+					${incrementalBuild ? ('-Dincremental -Dgib.referenceBranch=refs/remotes/origin/'+helper.scmSource.pullRequest.target.name) : '-Pjqassistant -Pci-sources-check'} \
 					${enableDefaultBuildIT ? '' : '-DskipITs'} \
 					${toTestJdkArg(environments.content.jdk.default)} \
 			"""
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -883,6 +898,7 @@ void mavenNonDefaultBuild(BuildEnvironment buildEnv, String args, List<String> a
 					${toTestJdkArg(buildEnv)} \
 					--fail-at-end \
 					$argsWithProjectSelection \
+					-Dincremental \
```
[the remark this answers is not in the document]
yrodiere: Also, maybe you should add this to `argsWithProjectSelection` instead (see code just above), so that we avoid unnecessarily pulling container images?

pom.xml
```diff
@@ -1822,6 +1828,37 @@
...
+                        <groupId>io.github.gitflow-incremental-builder</groupId>
+                        <artifactId>gitflow-incremental-builder</artifactId>
+                        <extensions>true</extensions>
+                        <!-- https://github.com/gitflow-incremental-builder/gitflow-incremental-builder#configuration
+                             General recap: Anything that is directly set in <configuration> cannot be redefined via '-D...'!
+                             See also: https://github.com/gitflow-incremental-builder/gitflow-incremental-builder/issues/213 -->
+                        <configuration>
+                            <!-- pointless to attempt incremental build if something like mvnw was changed
+                                 (and also potentially wrong, given that independent-projects might not be built) -->
+                            <skipIfPathMatches>\.github[/\\].*|\.mvn[/\\].*|mvnw.*|.*\.adoc|.*\.md</skipIfPathMatches>
```
yrodiere, line 1855: Does this mean we'll rebuild everything as soon as there's a change in asciidoc or markdown files? :scream:

pom.xml
```diff
@@ -1822,6 +1828,37 @@
...
+                        </configuration>
```
yrodiere, line 1856: We might want to use https://github.com/gitflow-incremental-builder/gitflow-incremental-builder#gibexcludepathsmatching in order to skip rebuilds when changing READMEs etc.?

Feel free to open another issue about that though.

yrodiere (commented):
Thanks, I added a few comments below.

I have a few more general comments:

1. I don't think we should use incremental builds for non-PR jobs, because that would reduces the confidence that a green light on `main` means everything is good and we can release...
2. This does not technically address [HSEARCH-4784](https://hibernate.atlassian.net/browse/HSEARCH-4784) Move testing of non-default environments to a separate job, though I agree it does improve on the problems that led me to open that issue.
However, if we don't use incremental builds for non-PR jobs (see comments), so these items from the issue still need to be addressed, and I believe this requires a separate job:
> We’d forbid concurrent builds, similarly to other jobs, but probably wouldn’t cancel running builds.
> When running against main branches (not PRs), it would have to be rate-limited (throttled), similarly to other ci/dependency-update/Jenkinsfile.
> It would probably need some form of parameters, to be able to run tests against a particular platform easily. I’d argue that using only some regexp condition would be fine. We can definitely simplify what we’re currently doing in the main Jenkins job.
> It would run against branches, but also against PRs. At the very least it should test PRs against a subset of all environments (currently I think we do Eclipse compiler, Oracle, PostgreSQL and that’s it; maybe we’d want to add the latest early-access JDK).
Maybe you should just open a different issue and change the issue key on this PR, so that you can merge and think about HSEARCH-4784 again when you have time?
2. Your changes actually reduce the workload of building PRs, so we might want to take advantage of that to switch a few strategic environements from `AFTER_MERGE` to `BEFORE_MERGE`? The latest OpenSearch env comes to mind in particular. Maybe some database as well, since those will only get built when we actually change something to the ORM integration.
3. I believe you should be able to go a step further and apply the incremental build to the "Default build" stage as well... Either you find a way to make sure that build will also build all the dependencies of modules it executes, or (probably safer) you just split the stage into two, the first stage building without executing integration test (`-DskipTests`) and storing built artifacts for use in other stages, and the second one executing all integration tests but with `-Dincremental -Dgib.disableSelectedProjectsHandling=true`.

marko-bekhta:
Thanks for looking at this 😃 

> I don't think we should use incremental builds for non-PR jobs, because that would reduces the confidence that a green light on main means everything is good and we can release...

+1 since we are comparing with the main branch then when we'd run the build against the main itself I was expecting it to run a full build (from docs: By default, no changed file results in GIB being skipped)... but that would only cover main and not 6.2 or others ... There is a `gib.disableIfBranchMatches` where we could limit it to PRs (that's if the Jenkins helper wouldn't work for some reason)

and another note on this matter.. we'd actually need to read the branch in the job, if it is a PR then we'd want to know if we opened it against main or maybe it's a backport then we'd need to compare against a different branch....

> This does not technically address [HSEARCH-4784](https://hibernate.atlassian.net/browse/HSEARCH-4784) Move testing of non-default environments to a separate job, though I agree it does improve on the problems that led me to open that issue.
However, if we don't use incremental builds for non-PR jobs (see comments), so these items from the issue still need to be addressed, and I believe this requires a separate job:

+1000000 😃 I was thinking of building on top of this, taking smaller steps, I actually already had a local branch where I've been experimenting with splitting the job, but it's toooooo soon to be pushed 😃.

> Maybe you should just open a different issue and change the issue key on this PR, so that you can merge and think about HSEARCH-4784 again when you have time?

+1 that makes sense, will do so, that way we can test these changes on real builds sooner.

> Your changes actually reduce the workload of building PRs, so we might want to take advantage of that to switch a few strategic environements from AFTER_MERGE to BEFORE_MERGE? The latest OpenSearch env comes to mind in particular. Maybe some database as well, since those will only get built when we actually change something to the ORM integration.

I've looked through the list of our envs ... for DBs the one after_merge are mostly those long-running ones (db2, cockroach...)... so I'm not sure if we want them on the list, while on the other hand those are ones causing failures 😞  
As for the OpenSearch -- maybe run the older one (1.3), I'm thinking that in most cases latest OpenSearch behaves more or less similar to Elasticsearch we are testing against, and those older versions sometimes have differences, like the issue we've seen this week with a test failing because of a different default shard count... I'm just thinking out loud here 😄 

> I believe you should be able to go a step further and apply the incremental build to the "Default build" stage as well... Either you find a way to make sure that build will also build all the dependencies of modules it executes, or (probably safer) you just split the stage into two, the first stage building without executing integration test (-DskipTests) and storing built artifacts for use in other stages, and the second one executing all integration tests but with -Dincremental -Dgib.disableSelectedProjectsHandling=true.

Nice idea, I'll give that a try.

marko-bekhta:
^ looks like I've confused the Hibernate Bot 😃 🙈

marko-bekhta:
> Nope, your commits do mention HSEARCH-4784 instead of HSEARCH-4924, you should probably change that :)

hah the comment got updated now, initially, it was saying that the PR looks ok, while I've changed the link but not the commits 😃

Jenkinsfile
```diff
@@ -377,15 +411,22 @@ stage('Default build') {
...
+				dir(helper.configuration.maven.localRepositoryPath) {
+					unstash name:'default-build-result'
+				}
+			}
+			// If we are in the pull request we've already run JQAssistant and source formatting checks
 			String mavenArgs = """ \
 					--fail-at-end \
-					-Pdist -Pcoverage -Pjqassistant -Pci-sources-check \
+					-Pdist -Pcoverage \
+					${incrementalBuild ? ('-Dincremental -Dgib.referenceBranch=refs/remotes/origin/'+helper.scmSource.pullRequest.target.name) : '-Pjqassistant -Pci-sources-check'} \
```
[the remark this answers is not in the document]
yrodiere: Oh, but that would require running unit tests in the first stage and not here, and then they wouldn't be included in Sonar test coverage reports... That would be annoying.

I wonder what happens if we just execute Sonar in every single stage... Would Sonar aggregate the results? :thinking:
If so that would be really great as we would also be able to execute sonar in (BEFORE_MERGE) non-default environements and would get test coverage numbers that more closely reflect reality.

That being said, I also wonder if Sonar test coverage numbers will still be relevant with incremental builds... Let's hope that Sonar only considers test coverage on changed code.

Jenkinsfile
```diff
@@ -402,27 +443,27 @@ stage('Default build') {
...
+									-Dsonar.pullrequest.key=${helper.scmSource.pullRequest.id} \
+									-Dsonar.pullrequest.base=${helper.scmSource.pullRequest.target.name} \
+									${helper.scmSource.gitHubRepoId ? """ \
+											-Dsonar.pullrequest.provider=GitHub \
+											-Dsonar.pullrequest.github.repository=${helper.scmSource.gitHubRepoId} \
+									""" : ''} \
+							""" : """ \
+									-Dsonar.branch.name=${helper.scmSource.branch.name} \
+							"""} \
+					"""
```
yrodiere, line 466: Having fun with automatic formatting aren't you? :)

I must say this one doesn't seem particularly good...

Jenkinsfile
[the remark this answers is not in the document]
yrodiere: Okay that's complex... it may be a separate problem to handle in a separate issue, e.g. "Build public modules only once in PR builds on CI" and maybe also "Enable Sonar test coverage on non-default environment builds on CI". WDYT, should I open those?

yrodiere (commented):
Thanks! LGTM apart from `excludePathsMatching`, I think my other comments do not necessarily need to be addressed now (see below).

Jenkinsfile
marko-bekhta: 🙈👀:wasntme:😃

yrodiere (approved):

PR #3671, opened 2023-08-23
HSEARCH-4925 Improve documentation of Elasticsearch/OpenSearch compatibility

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4925

PR #3672, opened 2023-08-25
HSEARCH-4925 Follow-up: Improve documentation (and tests) of Elasticsearch/OpenSearch compatibility

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4925

Follows up on #3671.

Looks like I made a very dumb mistake resulting in relevant tests never being executed.

I suppose we don't currently test on any platform where `supportsIndexClosingAndOpening()` would return `false`, and that's why the inverted assumption didn't cause any failure...

But let's leave the assumption anyway, because we'll need it to test Amazon OpenSearch Serverless.


yrodiere:
Started a build to run tests in all ES/OpenSearch environments: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-3672/3/pipeline

marko-bekhta (approved):
🙈

yrodiere:
The build is failing due to a problem with the Sonar plugin. Converting to draft, I'll try a fix...

yrodiere:
The failures are just for environements that we don't have a running service for (e.g. older versions of Elasticsearch). So, it works. Merging.

PR #3676, opened 2023-08-25
HSEARCH-4926 Improve casts in HashTable

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4926

This doesn't change the behavior in practice, but is slightly more
correct, and will avoid Sonar reporting a bug.


marko-bekhta (approved):

yrodiere:
It's just sonar failing, which should be addressed in #3678. Merging.

PR #3678, opened 2023-08-25
HSEARCH-4924 Run pre-build and build stages on the same node

marko-bekhta:

I was looking at this: https://www.jenkins.io/blog/2018/07/02/whats-new-declarative-piepline-13x-sequential-stages/ (Running Multiple Stages with the Same agent, or environment, or options) Let's see if this builds 😃 🙈

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -375,19 +375,19 @@ Resulting execution plan:
 """
 }
 
-stage('Pre-build sources') {
-	// we want this stage to only be executed when we are planning to use incremental build
-	if (!incrementalBuild) {
-		echo 'Skipping pre-building sources for non-pull request builds.'
-		helper.markStageSkipped()
-		return
-	}
-	if (!enableDefaultBuild) {
-		echo 'Skipping default build and integration tests in the default environment'
-		helper.markStageSkipped()
-		return
-	}
-	runBuildOnNode( NODE_PATTERN_BASE ) {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -402,14 +402,13 @@ stage('Pre-build sources') {
...
 }
-
-stage('Default build') {
-	if (!enableDefaultBuild) {
-		echo 'Skipping default build and integration tests in the default environment'
-		helper.markStageSkipped()
-		return
-	}
-	runBuildOnNode( NODE_PATTERN_BASE, [time: 2, unit: 'HOURS'] ) {
+runBuildOnNode( NODE_PATTERN_BASE, [time: 2, unit: 'HOURS'] ) {
```
yrodiere, line 405: Each `runBuildOnNode` instruction will potentially pick another node with the same label. You need to put both `stage` blocks inside the same `runBuildOnNode` block.

yrodiere (changes requested):
> I was looking at this: https://www.jenkins.io/blog/2018/07/02/whats-new-declarative-piepline-13x-sequential-stages/ (Running Multiple Stages with the Same agent, or environment, or options) Let's see if this builds 😃 🙈

That's the declarative syntax, which unfortunately is not relevant here.

This file uses the scripted syntax:  https://www.jenkins.io/doc/book/pipeline/syntax/#scripted-pipeline

yrodiere (approved):

PR #3679, opened 2023-08-25
HSEARCH-4867 Experimental compatibility with Amazon OpenSearch Serverless

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4867

~~Based on #3672 , which must be merged first.~~ => Done

In short, this overcomes all problems reported by RiVogel in the Jira issue, but the tests just use a local OpenSearch instance because of problems setting up an Amazon OpenSearch Serverless cluster (see https://hibernate.atlassian.net/browse/HSEARCH-4919).

So, we really can't consider this working now, but at least it's there for anyone to try and we will eventually be able to get to testing and perfecting it (see https://hibernate.atlassian.net/browse/HSEARCH-4919).

There are a few limitations, but I think they are acceptable.

@marko-bekhta I'd recommend reviewing commit by commits, and skipping the commits that are strictly about tests, because they are uninteresting and not even critical (since we don't test against Amazon OpenSearch Serverless at the moment).

### DIFF
--- a/backend/elasticsearch-aws/src/main/java/org/hibernate/search/backend/elasticsearch/aws/impl/AwsSigningRequestInterceptor.java
+++ b/backend/elasticsearch-aws/src/main/java/org/hibernate/search/backend/elasticsearch/aws/impl/AwsSigningRequestInterceptor.java
@@ -40,15 +40,15 @@ class AwsSigningRequestInterceptor implements HttpRequestInterceptor {
 
 	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
 
-	private static final String ELASTICSEARCH_SERVICE_NAME = "es";
-
 	private final Aws4Signer signer;
 	private final Region region;
+	private final String service;
 	private final AwsCredentialsProvider credentialsProvider;
 
-	AwsSigningRequestInterceptor(Region region, AwsCredentialsProvider credentialsProvider) {
+	AwsSigningRequestInterceptor(Region region, String service, AwsCredentialsProvider credentialsProvider) {
 		this.signer = Aws4Signer.create();
 		this.region = region;
+		this.service = service;
 		this.credentialsProvider = credentialsProvider;
 	}
 
@@ -73,7 +73,7 @@ class AwsSigningRequestInterceptor implements HttpRequestInterceptor {
 		Aws4SignerParams signerParams = Aws4SignerParams.builder()
 				.awsCredentials( credentials )
 				.signingRegion( region )
-				.signingName( ELASTICSEARCH_SERVICE_NAME )
+				.signingName( service )
 				.build();
 
 		awsRequest = signer.sign( awsRequest, signerParams );
@@ -125,6 +125,11 @@ class AwsSigningRequestInterceptor implements HttpRequestInterceptor {
 		else {
 			path = pathAndQuery;
 			queryParameters = Collections.emptyList();
+		}
+
+		// For some reason this is needed on Amazon OpenSearch Serverless
+		if ( "aoss".equals( service ) ) {
+			awsRequestBuilder.appendHeader( "x-amz-content-sha256", "required" );
 		}
 
 		awsRequestBuilder.encodedPath( path );
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
@@ -14,6 +14,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
 
+import org.hibernate.search.backend.elasticsearch.ElasticsearchDistributionName;
 import org.hibernate.search.backend.elasticsearch.ElasticsearchVersion;
 import org.hibernate.search.backend.elasticsearch.client.spi.ElasticsearchRequest;
 import org.hibernate.search.backend.elasticsearch.client.spi.ElasticsearchResponse;
@@ -176,7 +177,7 @@ public interface Log extends BasicLogger {
 	SearchException failedToDetectElasticsearchVersion(String causeMessage, @Cause Exception e);
 
 	@Message(id = ID_OFFSET_LEGACY_ES + 81,
-			value = "Incompatible Elasticsearch version running on the cluster: '%s'."
+			value = "Incompatible Elasticsearch version: '%s'."
 					+ " Refer to the documentation to know which versions of Elasticsearch"
 					+ " are compatible with Hibernate Search.")
 	SearchException unsupportedElasticsearchVersion(ElasticsearchVersion version);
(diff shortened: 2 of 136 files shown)


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/impl/DeleteByQueryWork.java
```diff
@@ -60,7 +64,7 @@ public Builder(URLEncodedString indexName, JsonObject payload, ElasticsearchWork
 			super( ElasticsearchRequestSuccessAssessor.DEFAULT_INSTANCE );
 			this.indexName = indexName;
 			this.payload = payload;
-			this.refreshWorkBuilder = workFactory.refresh().index( indexName );
+			this.refreshWorkBuilder = workFactory.isRefreshSupported() ? workFactory.refresh().index( indexName ) : null;
```
marko-bekhta, line 67: I wonder if we could have some "constant/stub" builder so we would not need to make those other null checks in this class

documentation/src/main/asciidoc/public/reference/_backend-elasticsearch.adoc
```diff
@@ -73,6 +76,38 @@ and as a result <<mapper-orm-schema-management-manager-create-or-update,automati
...
+link:{amazonOpenSearchServerlessUrl}[Amazon OpenSearch Serverless] compatibility is implemented and incubating;
+feel free to provide feedback on https://hibernate.atlassian.net/browse/HSEARCH-4867[HSEARCH-4867].
+
+However, be aware that:
+
+* Hibernate Search does not currently get tested against Amazon OpenSearch Serverless;
+see https://hibernate.atlassian.net/browse/HSEARCH-4919[HSEARCH-4919].
+* Connecting to an Amazon OpenSearch Service cluster
+requires <<backend-elasticsearch-configuration-aws,proprietary authentication>> that involves extra configuration.
+* Compatibility with Amazon OpenSearch Service must be enabled
```
marko-bekhta, line 93:
```suggestion
* Compatibility with Amazon OpenSearch Serverless must be enabled
```
Maybe say that it's about serverless specifically (and in the previous line item too)

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/util/ElasticsearchTckBackendFeatures.java
```diff
@@ -203,9 +203,53 @@ public static boolean supportsIndexClosingAndOpening() {
 		return isActualVersion(
 				// See https://docs.aws.amazon.com/opensearch-service/latest/developerguide/supported-operations.html#version_7_1
 				// See https://docs.aws.amazon.com/opensearch-service/latest/developerguide/supported-operations.html#version_7_4
-				esVersion -> !( esVersion.isAws() && esVersion.isLessThan( "7.4" ) ),
+				es -> !( es.isAws() && es.isLessThan( "7.4" ) ),
 				// See https://docs.aws.amazon.com/opensearch-service/latest/developerguide/supported-operations.html#version_opensearch_1.0
-				osVersion -> true
+				os -> true,
+				aoos -> false
```
marko-bekhta, line 209:
```suggestion
				aoss -> false
```
this one caught my eye 😃

marko-bekhta (approved):
looks nice 👍🏻😃

util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/rule/TestElasticsearchClient.java
```diff
@@ -98,6 +98,11 @@ public IndexClient(URLEncodedString primaryIndexName, URLEncodedString writeAlia
 			this.primaryIndexName = primaryIndexName;
 			this.writeAlias = writeAlias;
 			this.readAlias = readAlias;
+
+			// Always delete indexes manipulated through test client after the test;
+			// leaving indexes on the cluster could cause problems
+			// with Amazon OpenSearch Serverless in particular.
+			TestElasticsearchClient.this.registerIndexForCleanup( primaryIndexName );
```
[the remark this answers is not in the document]
yrodiere: I don't think that's worth it:

* Last time I checked dropping an index was relatively fast (a few milliseconds).
* This will only impact a few tests that actually rely on `TestElasticsearchClient`.
* This is a more of a safety net, as Hibernate Search should generally delete these indexes when it stops anyway.

backend/elasticsearch-aws/src/main/java/org/hibernate/search/backend/elasticsearch/aws/impl/AwsSigningRequestInterceptor.java
```diff
@@ -127,6 +127,11 @@ private SdkHttpFullRequest toAwsRequest(HttpRequest request, HttpContext context
 			queryParameters = Collections.emptyList();
 		}
 
+		// For some reason this is needed on Amazon OpenSearch Serverless
+		if ( "aoss".equals( service ) ) {
+			awsRequestBuilder.appendHeader( "x-amz-content-sha256", "required" );
```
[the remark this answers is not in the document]
yrodiere: They can already do that using the same API that the AWS module relies on; see https://docs.jboss.org/hibernate/search/6.2/reference/en-US/html_single/#backend-elasticsearch-configuration-http-client

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -810,4 +814,30 @@ SearchException cannotHighlightInNestedContext(String currentNestingField,
...
+	@Message(id = ID_OFFSET + 175,
+			value = "Unexpected Amazon OpenSearch Serverless version: '%1$s'."
+					+ " Amazon OpenSearch Serverless doesn't use version numbers."
+					+ " Set the version to simply '%2$s'.")
+	SearchException unexpectedAwsOpenSearchServerlessVersion(ElasticsearchVersion configuredVersion,
+			ElasticsearchVersion expectedAWSOpenSearchServerlessVersion);
+
+	@Message(id = ID_OFFSET + 176, value = "Cannot execute '%s' because Amazon OpenSearch Serverless does not support this operation."
+			+ " Either avoid this operation or switch to another Elasticsearch/OpenSearch distribution.")
+	SearchException cannotExecuteOperationOnAmazonOpenSearchServerless(String flush);
```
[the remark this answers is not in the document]
yrodiere: I tried to limit the impact of those changes, because I wouldn't want to change e.g. the mapper just because of this incubating feature.

Anyway, in the case of index open/close, the operations are executed conditionally, only if there are settings changes, so we can't really fail early.

Thanks for the suggestion, will apply.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/impl/DeleteByQueryWork.java
yrodiere: We could, but it seems to me that it would be worse because now we'd have a builder class to maintain, even though its only goal is to produce nothing... ?

Anyway, this code is covered by tests, so I'm not particularly worried about the smelly `null` usage.

documentation/src/main/asciidoc/public/reference/_backend-elasticsearch.adoc
yrodiere: Good catch, will do

yrodiere:
Thanks for the review; I think I addressed all comments. Let's wait for the other PR to be merged and I'll rebase this one.

yrodiere:
I started an extensive build there: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-3679/8/pipeline

yrodiere:
I had to fix a few tests that were failing on the default build as a result of my changes... hopefully the default build should pass now, at least.
Let's see about the others: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-3679/16/pipeline

yrodiere:
(I'll also address the test failures)

yrodiere:
There was just one test failing, and it was because of a missing assumption. I fixed that, let's wait for CI and merge: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-3679/22/pipeline

PR #3682, opened 2023-08-28
Add more test/build dependencies to the `build-dependencies` group

marko-bekhta:
😃

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -46,6 +46,14 @@ updates:
           - "org.jboss.marshalling:jboss-marshalling" # JBeret IT dependency
           - "org.wildfly.security:wildfly-security-manager" # JBeret IT dependency
           - "org.springframework.boot:*" # Spring is only for ITs
+          - "org.mockito:*"
+          - "org.hamcrest:*"
+          - "org.apache.logging.log4j:*"
+          - "org.assertj:*"
+          - "org.awaitility:*"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -46,6 +46,15 @@ updates:
           - "org.jboss.marshalling:jboss-marshalling" # JBeret IT dependency
           - "org.wildfly.security:wildfly-security-manager" # JBeret IT dependency
           - "org.springframework.boot:*" # Spring is only for ITs
+          - "org.mockito:*"
+          - "org.hamcrest:*"
+          - "org.apache.logging.log4j:*"
+          - "org.slf4j:*"
```
yrodiere, line 52: This is not a build-only dependency, unfortunately.

https://github.com/hibernate/hibernate-search/blob/d0df46828e80fe360d904c5715d0dedeab4df6d0/build/parents/build/pom.xml#L93-L94

```suggestion
```

yrodiere (changes requested):

.github/dependabot.yml
marko-bekhta: right.... but from the comment on it it looks we only use it to align the version so enforcer do not complain. aaand that means we'd update it once we update SDK... so I suppose we don't need to update it at all 😄

yrodiere (approved):

PR #3686, opened 2023-08-28
HSEARCH-4924 Follow-up: minor improvements to incremental builds

yrodiere:

Some minor syntax changes + we'll mark stages as skipped if the incremental builds skips all relevant modules.

I only tested the various options/commands locally, I'm hoping to test it for real in another PR.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -899,10 +899,12 @@ void mavenNonDefaultBuild(BuildEnvironment buildEnv, String args, List<String> a
 	// We want to run relevant integration test modules only (see array of module names)
 	// and in PRs we want to run only those affected by changes
 	// (see gib.disableSelectedProjectsHandling=true).
+	String incrementalProjectsListFile = 'target/.gib-impacted'
 	String argsWithProjectSelection = """ \
 		${incrementalBuild ? """ \
 				-Dincremental -Dgib.disableSelectedProjectsHandling=true \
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -921,6 +929,11 @@ void mavenNonDefaultBuild(BuildEnvironment buildEnv, String args, List<String> a
 					--fail-at-end \
 					$argsWithProjectSelection \
 	"""
+
+	if (incrementalBuild && 0 == sh(script: "test ! -s '${incrementalProjectsListFile}'", returnStatus: true)) {
```
marko-bekhta, line 933: maybe we could do that before running the tests ^ . Or we' want the output of this exact build to determine if the stage needs to be marked as skipped?

yrodiere: Well the file is created by gib during the Maven build, and we're only interested in it if it's empty, i.e. if the Maven build did close to nothing... so I don't think it's worth a separate Maven execution?

marko-bekhta: thanks 😃 
yeah it seemed confusing, running the build to mark it skipped, but yeah it makes sense now 😃, we basically want  to see that nothing was tested on CI side without looking into the logs and so on, right ?

yrodiere: Right:

```
	// In incremental builds, the Maven execution above
	// created a file listing projects relevant to the incremental build.
	// If it is empty, it means the incremental build didn't actually do anything,
	// so make sure to mark the stage as skipped.
```

PR #3689, opened 2023-08-29
HSEARCH-3654 Upgrade Junit where possible

marko-bekhta:

just a couple changes 😉😄 

while the first commit is all junit changes squashed, in the second one I've tried to create an extension that we might use for those remaining reusable-orm-helper kind of tests... 
Executing that test will produce a sequence:
```
[root]
env(String, int, boolean)
Env #0: Initializing
Env #0: test1
Env #0: test2
Env #1: Initializing
Env #1: test1
Env #1: test2
Env #2: Initializing
Env #2: test1
Env #2: test2
```
`Env #{x}: Initializing` -- where the init method is called for `x` set of arguments
`Env #{x}: test{y}` -- execution of test y for `x` set of arguments


yrodiere (commented):
Thanks! I (obviously) didn't review all files, but I had a look at a few tests and the main utils. Here are a few comments; we'll probably need to iterate :)

backend/elasticsearch/src/test/java/org/hibernate/search/backend/elasticsearch/gson/spi/GsonParsingTest.java
```diff
@@ -8,46 +8,43 @@
...
-	@Parameterized.Parameter(0)
-	public String jsonToParse;
-	@Parameterized.Parameter(1)
-	public Class<?> targetType;
-
-	public GsonParsingTest() {
-		gson = GsonProvider.create( GsonBuilder::new, true ).getGsonNoSerializeNulls();
+	public Gson init() {
+		return GsonProvider.create( GsonBuilder::new, true ).getGsonNoSerializeNulls();
 	}
```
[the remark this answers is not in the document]
marko-bekhta: This one is actually good; it probably got caught in all the other changes... we can have a no-arg constructor, but not the constructor with parameters when we are having `@ParameterizedTest`

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/util/TestedFieldStructure.java
```diff
@@ -25,7 +23,7 @@
  * You need set up your test to create one field per structure,
  * and then query the "fieldStructure" object in your test instance to target a different field
  * based on that.
- * See for example {@link FieldSortBaseIT}.
+ * See for example {@code org.hibernate.search.integrationtest.backend.tck.search.sort.FieldSortBaseIT}.
```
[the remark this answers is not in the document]
marko-bekhta: Since I've removed most of the `public` modifiers... the class is in a different package, so it's not "visible" to this Javadoc anymore ...

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/util/extension/SearchSetupHelper.java
```diff
@@ -37,42 +37,101 @@
...
+import org.junit.jupiter.api.extension.TestExecutionExceptionHandler;
+
+public class SearchSetupHelper
+		implements AfterAllCallback, AfterEachCallback, AfterTestExecutionCallback,
+		BeforeAllCallback, BeforeEachCallback, BeforeTestExecutionCallback, TestExecutionExceptionHandler {
+
+	private enum Type {
+		CLASS,
+		METHOD;
+	}
```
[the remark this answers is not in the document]
marko-bekhta: hmm, I'm not sure that `@TestInstance` can work here... this `Type` ^ is to decide which callback pair will be called `*Each` or `*All` ... 
Or do you mean that we should read the lifecycle from the extension context and then use it instead of the custom enum? (not sure if that'll be simpler though... )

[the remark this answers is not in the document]
yrodiere: > hmm, I'm not sure that `@TestInstance` can work here... this `Type` ^ is to decide which callback pair will be called `*Each` or `*All` ... 

Ah, right, this doesn't look related.

> Or do you mean that we should read the lifecycle from the extension context and then use it instead of the custom enum? (not sure if that'll be simpler though... )

> (not sure if that'll be simpler though... )

In the helpers maybe not. If it's simpler in tests themselves (i.e. no explicit `.create()`/`createGlobal()`, I'd still consider this a win though.

Perhaps you could just detect whether the `start` method gets called from a `@BeforeAll` method or from a `@BeforeEach`/`@Test` method?

* If it's called from a `@BeforeAll` method, you know any mapping created in that method will have to be cleaned up in `@AfterAll`.
* If it's called from a `@BeforeEach` or `@Test` method, you know any mapping created in that method will have to be cleaned up in `@AfterEach`.

Assuming the extension's own `beforeEachCallback` gets called before any `@BeforeEach`/`@Test` method, this could easily be achieved by updating some state in the helper when `beforeEachCallback` gets called.

mapper/orm/src/test/java/org/hibernate/search/mapper/orm/model/impl/HibernateOrmBootstrapIntrospectorAccessTypeTest.java
```diff
@@ -26,117 +26,138 @@
...
 
-public class HibernateOrmBootstrapIntrospectorAccessTypeTest
+class HibernateOrmBootstrapIntrospectorAccessTypeTest
 		extends AbstractHibernateOrmBootstrapIntrospectorPerReflectionStrategyTest {
 
-	@Test
-	public void entity_defaultFieldAccess() {
-		HibernateOrmBootstrapIntrospector introspector = createIntrospector( EntityWithDefaultFieldAccess.class );
+	@ParameterizedTest(name = "Reflection strategy = {0}")
+	@MethodSource("params")
```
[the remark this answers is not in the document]
marko-bekhta: yeah 😔,
before this, we were using something similar but on a smaller scale:
```java
@Parameterized.Parameters(name = "{0} - {1}")
Obecjt[][] params() {
```
I suppose I can remove the names 😃... on the other hand, with that in place, the test names should better match with what we had, if we'd need to compare the results...
Also in some cases, parameters are lambdas so that wouldn't be a much useful output in the tests too...
soooooooo IDK 🤷🏻‍♂️

util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/extension/TestElasticsearchClient.java
```diff
@@ -48,26 +48,34 @@
...
 
 	private final List<URLEncodedString> createdIndicesNames = new ArrayList<>();
 
+	private TestElasticsearchClient() {
+	}
+
+	public static TestElasticsearchClient create() {
+		return new TestElasticsearchClient();
+	}
+
```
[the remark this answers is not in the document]
marko-bekhta: 😋 haha, yeah.... since this is "an extension" (it has some junit callbacks implemented) and as most of our extensions are using some static method to init itself in the test class, I wanted to make it more consistent and updated this one to follow that approach as well

mapper/orm/src/test/java/org/hibernate/search/mapper/orm/model/impl/HibernateOrmBootstrapIntrospectorAccessTypeTest.java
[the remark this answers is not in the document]
yrodiere: > on the other hand, with that in place, the test names should better match with what we had, if we'd need to compare the results...

Hmm ok, we can do this change last then. Or in a separate commit. Or not at all, we'll see.

>  Also in some cases, parameters are lambdas so that wouldn't be a much useful output in the tests too...

Not useful, but hopefully won't hurt either.

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/extension/ActualBackendSetupStrategy.java
```diff
@@ -43,19 +42,22 @@ public String toString() {
...
 				continue;
 			}
-			if ( ruleChain == null ) {
-				ruleChain = RuleChain.emptyRuleChain();
+			if ( finalExtension == null ) {
+				finalExtension = extension.get();
+			}
+			else {
+				// TODO: fix me
+				finalExtension = new ComposedExtension( extension.get(), finalExtension );
```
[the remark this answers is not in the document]
marko-bekhta: I'm soooo glad that I've left soooo much context in this comment letting everyone know what needs to be fixed here ... 😆 😔

util/internal/test/common/src/test/java/org/hibernate/search/util/impl/test/reflect/ExperimentsTest.java
```diff
@@ -0,0 +1,58 @@
...
+		return Arrays.asList(
+				Arguments.of( "string1", 1, true ),
+				Arguments.of( "string2", 2, true ),
+				Arguments.of( "string3", 3, true )
+		);
+	}
+
+	@MultiRunExtension.EnvironmentTest(init = "init")
+	@MethodSource("params")
+	void env(String string, int number, boolean bool) {
```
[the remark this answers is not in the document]
marko-bekhta: hmm, this `TestTemplateInvocationContextProvider` that I've tried to used is working with the test method, in this particular case it's any method annotated with `@EnvironmentTest`, and all it does is basically "providing a different context to this method to execute"... We can use it as a `@BeforeEach` . The execution of this `@MultiRunExtension.EnvironmentTest(init = "init") void env(String string, int number, boolean bool)` is performed by JUnit, and I just append a call to either init or to `@MultiRunExtension.TestToExecute` actual test method in the after-callback...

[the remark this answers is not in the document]
yrodiere: And ideally we'd use `@Test` instead of `@MultiRunExtension.TestToExecute`, but I'm not sure how that could work exactly. Maybe the extension could prevent JUnit's default execution of `@Test` methods somehow? Not sure side-effects from one method to another are allowed, but it's worth having a look, at least.

util/internal/test/common/src/test/java/org/hibernate/search/util/impl/test/reflect/Experiments2Test.java
```diff
@@ -0,0 +1,295 @@
...
+import org.junit.jupiter.api.extension.DynamicTestInvocationContext;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.InvocationInterceptor;
+import org.junit.jupiter.api.extension.RegisterExtension;
+import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.platform.commons.util.AnnotationUtils;
+import org.junit.platform.commons.util.ReflectionUtils;
+
+class Experiments2Test {
```
marko-bekhta, line 41: This class contains a different approach to this "problem". Here, dynamic tests are used through `@TestFactory`.  With it we can produce a better tree structure for the executed tests:
```
[root]
	testsFromStaticClass()
		Configuration group: [string1] [1] [true]
			test1()
			test2()
		Configuration group: [string2] [2] [true]
			test1()
			test2()
		Configuration group: [string3] [3] [true]
			test1()
			test2()
		Configuration group: [string3] [3] [false]
			test1()
			test2()
	testsInnerClass()
		Configuration group: [string1] [1] [true]
			test1()
			test2()
			test3()
		Configuration group: [string2] [2] [true]
			test1()
			test2()
			test3()
		Configuration group: [string3] [3] [true]
			test1()
			test2()
			test3()
		Configuration group: [string3] [3] [false]
			test1()
			test2()
			test3()
```
The downside is that we can't really add extensions within the "test inner classes", and if we need any extensions, they must live at the root class level.
BeforeEach/AfterEach methods in the root will be treated as BeforeAll/AfterAll for those inner test classes.
If we'd need to execute something around the inner test methods from the root extension, we can try to do so through a `InvocationInterceptor#interceptDynamicTest` and do something before/after executing ` invocation.proceed();`

It's probably easier to run this test in the IDE see how the integration works, and all the output to get better picture 😃 (Note in the IDE you wouldn't be able to jump to the test method from the test panel, since the test.. are dynamic 😃)

util/internal/test/common/src/test/java/org/hibernate/search/util/impl/test/reflect/ExperimentsTest.java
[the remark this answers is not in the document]
yrodiere: > I see you can add whatever extension you want in `TestTemplateInvocationContext#getAdditionalExtensions`, so maybe you could use `org.junit.jupiter.api.extension.ExecutionCondition` or `org.junit.jupiter.api.extension.InvocationInterceptor` to:
> 
>     1. Skip some executions of the `setup` method.
> 
>     2. Replace some executions of the `setup` method with the actual tests.

From your other experiment, I take it this didn't work out?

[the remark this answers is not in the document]
marko-bekhta: 😔 nope... it didn't 🥲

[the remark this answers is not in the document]
yrodiere: The additional extensions didn't work, or... ?

util/internal/test/common/src/test/java/org/hibernate/search/util/impl/test/reflect/Experiments2Test.java
marko-bekhta: ohh and we'll need to update jqassistant rules:
```
[ERROR] Constraint: hsearch:NestedTestTypesShouldBeAnnotatedWithNested
[ERROR] Severity: MAJOR
[ERROR] Number of rows: 2
[ERROR]             Nested test types should be annotated with @Nested in order to be actually executed
[ERROR]             when their surrounding type is executed.
[ERROR]             Exceptions are only allowed for:
[ERROR]             - Abstract test classes.
[ERROR]         
[ERROR]   Test=org.hibernate.search.util.impl.test.reflect.Experiments2Test$TheseAreActualTestsButTheyNeedParametersTest
[ERROR]   Test=org.hibernate.search.util.impl.test.reflect.Experiments2Test$TheseAreActualTestsButTheyNeedParametersAndClassIsStaticTest
[ERROR] -------------------------------------------------------------------
```
Maybe having some "special name pattern" for these types of inner tests so that we could exclude them from queries ....

util/internal/test/common/src/test/java/org/hibernate/search/util/impl/test/reflect/ExperimentsTest.java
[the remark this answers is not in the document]
marko-bekhta: See the new variant 😄 after our discussion 😉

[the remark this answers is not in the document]
yrodiere: Nice improvement, thanks!

I didn't look at the code yet, but I'll make a few comments on the API.

About the cleanup, I'd suggest in particular:

* Renaming `@MultiRunExtension.EnvironmentTest` to `@ParameterizedSetup`
* Using `@Test` instead of `@MultiRunExtension.TestToExecute` if possible (see below), or failing that renaming it to`@ParameterizedSetup.Test`.

About the behavior, two questions:

1. The "env" (setup) method will likely need to set fields that will have to be read by test methods, which I expect will require using `@TestInstance(Lifecycle.PER_CLASS)`. Could you check this, and if it is indeed required, could you either:
   a. Find a way to force the PER_CLASS lifecycle as soon as `MultiRunExtension` is used
   b. OR (if the above is not possible) check the test lifecycle in `MultiRunExtension` and throw an exception if it's `PER_METHOD` (e.g. "this extension requires test instances to be shared across test methods; apply `@TEstInstance(Lifecycle.PER_CLASS)` to your test").
2. Did you look into using simply `@Test`/`@ParameterizedTest` for the test methods? If that's not possible from the extension on the `env` method, you could try moving the extension annotation to the class level, so that it applies extensions to all test methods, effectively removing all test methods from the testing plan, and restoring them as part of the setup?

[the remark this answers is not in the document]
marko-bekhta: ooook, I think you can now look at the `ParameterizedClassExtensionTest` to see how it can look like, and even work 😆.
There is a marker annotation `@ParameterizedClass` beside the fact that it has a `@TestInstance(Lifecycle.PER_CLASS)` to be able to correctly access context and prevent some weird behavior it also serves as a marker for a filter `ParameterizedClassPostDiscoveryFilter`. This filter is there to tell JUnit to ignore `@Test`/`@ParameterizedTest` if we are working with a parameterized class.
Then there is a `@ParameterizedSetup` that we put on the setup method along side some of the argument source annotations.
Then you can have `@Test`/`@ParameterizedTest`  in the class and the extension will discover and execute them ...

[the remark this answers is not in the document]
yrodiere: Well API-wise, this looks perfect. I'll add comments about various additional checks that could make sense and will try to finish my review.
Thanks a lot!

documentation/src/test/java/org/hibernate/search/documentation/mapper/orm/alternative/alternativebinder/AlternativeBinderIT.java
```diff
@@ -44,14 +45,12 @@ public static List<? extends Arguments> params() {
...
-			BackendConfigurations.simple() );
+	public DocumentationSetupHelper setupHelper = DocumentationSetupHelper.withSingleBackend( BackendConfigurations.simple() );
 	private EntityManagerFactory entityManagerFactory;
 
 	public void init(Boolean annotationProcessingEnabled, HibernateOrmSearchMappingConfigurer mappingContributor) {
-		setupHelper.withAnnotationProcessingEnabled( annotationProcessingEnabled )
-				.withMappingConfigurer( mappingContributor );
-		entityManagerFactory = setupHelper.start().setup( BlogEntry.class );
+		entityManagerFactory = setupHelper.start( variant( annotationProcessingEnabled, mappingContributor ) )
+				.setup( BlogEntry.class );
```
yrodiere, line 53: Interestingly, this feels very close to `@ParameterizedSetup`, except... the setup needs to be re-executed for every test method (at least in this test, because we need to clean up data).

Maybe you could use `@ParameterizedSetup` more broadly if there was a way to tell it when to perform setup? E.g. `@ParameterizedSetup(PER_CLASS)` vs. `@ParameterizedSetup(PER_METHOD)`?

documentation/src/test/java/org/hibernate/search/documentation/mapper/orm/alternative/alternativebinder/AlternativeBinderIT.java
```diff
@@ -44,14 +45,12 @@ public static List<? extends Arguments> params() {
 	}
 
 	@RegisterExtension
-	public DocumentationSetupHelper setupHelper = DocumentationSetupHelper.withSingleBackend(
-			BackendConfigurations.simple() );
+	public DocumentationSetupHelper setupHelper = DocumentationSetupHelper.withSingleBackend( BackendConfigurations.simple() );
 	private EntityManagerFactory entityManagerFactory;
 
 	public void init(Boolean annotationProcessingEnabled, HibernateOrmSearchMappingConfigurer mappingContributor) {
```
yrodiere, line 51: The idea was to use the Variant in the `Arguments` and to pass the variant around... otherwise it's not bringing much :)

util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/OrmSetupHelper.java
```diff
@@ -60,53 +65,57 @@ public final class OrmSetupHelper
...
 	public static OrmSetupHelper withMultipleBackends(BackendConfiguration defaultBackendConfiguration,
 			Map<String, BackendConfiguration> namedBackendConfigurations) {
-		return withMultipleBackends( Type.METHOD, defaultBackendConfiguration, namedBackendConfigurations );
+		return withMultipleBackends( Type.METHOD, defaultBackendConfiguration, namedBackendConfigurations,
+				JUPITER_ASSUMPTION_CHECK );
 	}
 
 	public static OrmSetupHelper withBackendMockGlobal(BackendMock backendMock) {
-		return withBackendMock( Type.CLASS, backendMock );
+		return withBackendMock( Type.CLASS, backendMock, VINTAGE_ASSUMPTION_CHECK );
```
yrodiere, line 87: If this is about JUnit 4 vs. JUnit 5, you might want to use a different suffix for this method than "Global"... ?

util/internal/test/common/src/main/java/org/hibernate/search/util/impl/test/extension/parameterized/ParameterizedClass.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+import org.junit.jupiter.api.TestInstance;
+
+/**
+ * A marker annotation to let JUnit engine know how to treat the test class annotated with it.
+ * <p>
+ * Tests that are expected to run as parameterized classes must have a setup method annotated with {@link ParameterizedSetup}.
```
yrodiere, line 17: > A marker annotation to let JUnit engine know how to treat the test class annotated with it.

This applies to pretty much every JUnit annotation :)

```suggestion
 * Marks test classes that carry class-level parameters, requiring a 
 * <p>
 * Such test classes must have a setup method annotated with {@link ParameterizedSetup}.
```

util/internal/test/common/src/main/java/org/hibernate/search/util/impl/test/extension/parameterized/ParameterizedClass.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public @interface ParameterizedClass {
```
yrodiere, line 21: Could you somehow add a check for classes annotated with this, so that an error is thrown if there is no `ParameterizedSetup`-annotated method?

Because that would be a pretty dangerous configurations: tests would be filtered and nothing would replace them...

I'm guessing this could be done through another extension added by this annotation.

util/internal/test/common/src/main/java/org/hibernate/search/util/impl/test/extension/parameterized/ParameterizedClassExtension.java
```diff
@@ -0,0 +1,188 @@
...
+		return envArguments.get( envIndex )[parameterContext.getIndex()];
+	}
+
+	@Override
+	public boolean supportsTestTemplate(ExtensionContext context) {
+		return isAnnotated( context.getTestMethod(), ParameterizedSetup.class );
+	}
+
+	@Override
+	public Stream<TestTemplateInvocationContext> provideTestTemplateInvocationContexts(ExtensionContext context) {
```
yrodiere, line 57: Should this throw an exception if the test class is not annotated with `@ParameterizedClass`?

util/internal/test/common/src/main/java/org/hibernate/search/util/impl/test/extension/parameterized/ParameterizedClassTestMethodDiscoverer.java
```diff
@@ -0,0 +1,102 @@
...
+
+	private ParameterizedClassTestMethodDiscoverer() {
+	}
+
+	public static List<ParameterizedTestMethodInvoker> discover(Class<?> testClass, ExtensionContext extensionContext) {
+		List<ParameterizedTestMethodInvoker> testMethods = new ArrayList<>();
+		for ( Method method : testClass.getDeclaredMethods() ) {
+			if ( ParameterizedClassUtils.isActualTestMethodToExecute( method ) ) {
+				testMethods.addAll( toTestMethods( method, extensionContext ) );
+			}
```
yrodiere, line 33: I think there's room for some checks to avoid dumb mistakes here (I guarantee I'll make one someday).

If a method in a `@ParameterizedClass` test class is annotated with any JUnit annotation beside those recognized by `ParameterizedClassExtension` (i.e. `@Test`, `@ParameterizedTest`, `@ParameterizedSetup`, and the various `@MethodSource`/`@CsvSource`/etc.) we should probably throw an exception because nobody will handle that annotation.

Also I think `@Nested` classes probably won't work, so we should throw an exception if there is any?

util/internal/test/common/src/main/java/org/hibernate/search/util/impl/test/extension/parameterized/ParameterizedClassExtension.java
```diff
@@ -0,0 +1,188 @@
...
+							return new ParameterizedClassTestTemplateInvocationContext(
+									testMethod.getName(),
+									MethodInterceptor.TEST_EXECUTION_INTERCEPTOR
+							);
+						}
+						else {
+							return new ParameterizedClassTestTemplateInvocationContext(
+									"Setup Parameters: " + Arrays.toString( envArguments.get( envIndex ) ),
+									MethodInterceptor.SETUP_EXECUTION_INTERCEPTOR
+							);
```
yrodiere, line 106: Could this be turned into a "BeforeEachCallback" extension added (manually) to every test instead?

That way:

1. The setup won't appear as a test in the IDE or surefire reports
2. A failure to run the setup will automatically mark the corresponding tests as failed (currently they get executed anyway... potentially using the previous setup).

yrodiere (commented):
Some more comments on `@ParameterizedSetup` et. al.

I'll probably wait for this to stabilize before I have a closer look at a random sample of tests...

util/internal/test/common/src/main/java/org/hibernate/search/util/impl/test/extension/parameterized/ParameterizedClass.java
marko-bekhta: In some cases, an Extension might be too late to detect problems in the test config, I've already had some checks in the `ParameterizedClassPostDiscoveryFilter` so I'll expend those and make them more "visible" 😃

util/internal/test/common/src/main/java/org/hibernate/search/util/impl/test/extension/parameterized/ParameterizedClassExtension.java
marko-bekhta: yeah, I'll have a check for it in `ParameterizedClassPostDiscoveryFilter`

util/internal/test/common/src/main/java/org/hibernate/search/util/impl/test/extension/parameterized/ParameterizedClassExtension.java
marko-bekhta: I thought that we might want to see the setup as a line, but if not... 😈 I'll control the setup with the method interceptor then and we'll just have a single  `ParameterizedClassTestTemplateInvocationContext` without the if/else part

documentation/src/test/java/org/hibernate/search/documentation/mapper/orm/alternative/alternativebinder/AlternativeBinderIT.java
marko-bekhta: see `AnalysisBuiltinIT` as an example 😃

util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/OrmSetupHelper.java
marko-bekhta: Once we replace the `ReusableOrmSetupHolder` all of this will go away, for now I've renamed it to "legacy" 😃

util/internal/test/common/src/main/java/org/hibernate/search/util/impl/test/extension/parameterized/ParameterizedClass.java
marko-bekhta: BTW... if we do the checks in the filter and throw an exception, the entire test suite will fail at the very beginning, so we wouldn't need to wait until the end to learn that we have a misconfigured test.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/util/extension/SearchSetupHelper.java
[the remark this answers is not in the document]
marko-bekhta: yeah, I was waiting for the parameterized class to be more or less finished before looking into this. I think that the "autodetection" that I've added in the new commit covers our needs and there's less manual work required 😃

marko-bekhta:
I've converted a couple of tests to use `@ParameterizedClass`  to see how CI will report them. If there will be a mismatch in test counts for those particular tests ... then we better check the counts and confirm all is ok before updating all remaining tests....

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/analysis/AnalysisBuiltinIT.java
```diff
@@ -18,51 +19,50 @@
...
 	private final SimpleMappedIndex<IndexBinding> index = SimpleMappedIndex.of( IndexBinding::new );
 
-	public AnalysisBuiltinIT(SearchSetupHelper setupHelper) {
-		this.setupHelper = setupHelper;
-	}
-
-	@Before
-	public void setup() {
-		setupHelper.start().withIndex( index ).setup();
+	@ParameterizedSetup(ParameterizedSetup.Lifecycle.PER_METHOD)
```
yrodiere, line 58: So, the API is looking great, thanks for that.

One last concern I can raise as I read through the code: there's little reason to combine `ParameterizedSetup.Lifecycle.PER_METHOD` with the `@TestInstance(PER_CLASS)` introduced by `@ParameterizedClass`... And this combination exposes us to bugs in our tests if the `setup` method does not always initialize all fields (leaves them to their default value): then method executions might be affected by the remnants of a previous setup.

Unfortunately, it seems [we can't set the test-instance lifecycle programmatically](https://github.com/junit-team/junit5/blob/71798ca59488362164123053e51f9c28abdcaf16/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestInstanceLifecycleUtils.java#L35-L44), so we can't just adapt it automatically based on this "parameterized setup lifecycle".

I wonder if it wouldn't be better to remove the parameter from here, and instead parameterize this using the class-level annotation:

* For "class" setup we'd annotate classes with `@ParameterizedPerClass` (meta-annotated with @TestInstance(PER_CLASS)`.
* For "method" setup we'd annotate classes with `@ParameterizedPerMethod` (not meta-annotated).
* The checks in the filter would verify that classes are annotated with either annotation.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/util/extension/SearchSetupHelper.java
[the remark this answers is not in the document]
yrodiere: Yep, what you did for `SearchSetupHelper` looks good, thanks :+1:

util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/ReusableOrmSetupHolder.java
```diff
@@ -170,12 +180,12 @@ public static ReusableOrmSetupHolder withBackendMocks(BackendMock defaultBackend
 			allBackendMocks.add( defaultBackendMock );
 		}
 		allBackendMocks.addAll( namedBackendMocks.values() );
-		return new ReusableOrmSetupHolder( OrmSetupHelper.withBackendMocks( defaultBackendMock, namedBackendMocks ),
+		return new ReusableOrmSetupHolder( OrmSetupHelper.withBackendMocksLegacy( defaultBackendMock, namedBackendMocks ),
```
yrodiere, line 183: Reminder this class is using legacy stuff and should still be converted.

By the way, I think you can take advantage of some JUnit 5 features, e.g. the ParameterResolver thing for setup methods expecting parameters of type `OrmSetupHelper.SetupContext` or `ReusableOrmSetupHolder.DataClearConfig`. The `@ReusableOrmSetupHolder.Setup` annotation might become completely unnecessary in that case.

Though of course, tests requiring bytecode enhancement might get in the way...

yrodiere (commented):
A few comments on the latest changes, thanks :) Feel free to continue with `ReusableOrmSetupHolder` and remaining tests once you've read through my comments.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/analysis/AnalysisBuiltinIT.java
marko-bekhta: hey, I've just tested it and it seems like we can do this:

- leave the `@ParameterizedClass` as is with `@TestInstance(TestInstance.Lifecycle.PER_CLASS)`
- if we need to do `@ParameterizedSetup(ParameterizedSetup.Lifecycle.PER_METHOD)` we add `@TestInstance( TestInstance.Lifecycle.PER_METHOD ) @ParameterizedClass` to the test class. 

The way JUnit first looks for a direct annotation and then checks for a meta-annotations, so we can override the value from the meta `@ParameterizedClass` in specific tests where we need the context to be recreated for each test

yrodiere: That's another possibility, the main downside compared to my proposal being that the lifecycle is not explicit for `@ParameterizedClass` and actually defaults to something different than in plain JUnit (PER_CLASS instead of PER_METHOD).

marko-bekhta: Ok, I think this version does it all 😄. There's a `@ParameterizedClass` that doesn't have a `@TestInstance` and which we use to annotate `ParameterizedPerClass` and `ParameterizedPerMethod` (both having `TestInstance`) this way we can still use the `ParameterizedClass` for the lookups where we need to identify tests we are interested in, and also have "shortcuts" for `@ParameterizedClass + @TestInstance` 

I've also moved the test execution into interceptor and dropped the `AfterTestExecutionCallback`

yrodiere: These annotations LGTM, feel free to move on to the ReusableOrmSetupHelper stuff :) Or create a separate issue if you intended to work on it separately (maybe you'll have to, if the tests that rely on it use bytecode enhancement?)

I'll mark the PR as draft, please mark it as ready for review when it *is* ready for review :)

yrodiere: Well it's already marked as draft, so:

> ~~I'll mark the PR as draft,~~ please mark it as ready for review when it *is* ready for review :)

marko-bekhta: From what I see... the remaining things are:

- [ ] Finish with bytecode enhancement "extension"
- [ ] Replace tests with ReusableOrmSetupHolder
  - [ ] Remove the remaining `org.junit` assumptions/assertions etc
  - [ ] remove extension methods with `*Legacy` suffix
- [ ] Update some of the parameterized tests to use `@ParameterizedClass` instead.
- [ ] Figure out what needs to be fixed in `ActualBackendSetupStrategy#getTestRule()`
- [ ] Check executed test count

If we won't be updating the `main` too much, I can continue adding more to this branch.... otherwise it'll be easier to get what we have in and then continue in a new branch... 🙈

yrodiere: > If we won't be updating the `main` too much, I can continue adding more to this branch.... otherwise it'll be easier to get what we have in and then continue in a new branch... 🙈

Personally I'll just merge my pending PRs and then, if there are no new bug reports, I'll probably hold off for a few weeks.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/search/SearchQueryBaseIT.java
```diff
@@ -77,27 +80,31 @@ public class SearchQueryBaseIT {
...
+	@RegisterExtension
+	public static OrmSetupHelper ormSetupHelper = OrmSetupHelper.withBackendMock( backendMock );
+	private SessionFactory sessionFactory;
 
-	@Rule
-	public MethodRule setupHolderMethodRule = setupHolder.methodRule();
+	@DataClearConfigConfigurer
+	public void tearDown(DataClearConfigConfigurer.DataClearConfig config) throws Exception {
+		config.clearOrder( Book.class, Author.class, NotIndexed.class );
+	}
```
marko-bekhta, line 93: Hey Yoann,

The idea here is to move the cleaning-up logic into the `OrmSetupHelper` itself. If there's a `@DataClearConfigConfigurer` annotated method in the test class using this helper then the helper will clean the data after the test execution. It doesn't look like we need to care about tracking the configuration parameters anymore. Since we have a `@ParameterizedClass` it should take care of setting up the session factory, while the only thing remaining is the data cleanup.

From what it looks, we don't really need to have the `ReusableOrmSetupHolder` in all the cases... (see previous commit where I've updated a few tests previously relying on this holder). Or we don't need to cleanup the data between tests in some examples as well (wich may help us speed up the execution a bit).
Also, we don't need the `@ParameterizedClass` in all cases since we can easily convert the test to use `@TestInstance(TestInstance.Lifecycle.PER_CLASS)` + `@BeforeAll` and simple `OrmSetupHelper` + adding the `@DataClearConfigConfigurer` when we need to clean the data.

What do you think of this approach?

integrationtest/mapper/orm-jakarta-batch/src/test/java/org/hibernate/search/integrationtest/jakarta/batch/massindexing/MassIndexingJobWithCompositeIdIT.java
```diff
@@ -54,30 +52,25 @@
...
-	public void setup(OrmSetupHelper.SetupContext setupContext) {
-		setupContext.withAnnotatedTypes( EntityWithIdClass.class, EntityWithEmbeddedId.class )
-				.withProperty( HibernateOrmMapperSettings.INDEXING_LISTENERS_ENABLED, false );
-	}
-
-	@Before
-	public void initData() throws Exception {
-		emf = setupHolder.entityManagerFactory();
+	@BeforeAll
+	void setup() {
```
yrodiere, line 70: I read something about non-static `@BeforeAll` methods only working with JDK 16+... you might want to try running this on JDK 11 before you do this everywhere.

integrationtest/mapper/orm-jakarta-batch/src/test/java/org/hibernate/search/integrationtest/jakarta/batch/massindexing/MassIndexingJobWithCompositeIdIT.java
```diff
@@ -54,30 +52,25 @@
...
+		emf = ormSetupHelper.start().withAnnotatedTypes( EntityWithIdClass.class, EntityWithEmbeddedId.class )
+				.withProperty( HibernateOrmMapperSettings.INDEXING_LISTENERS_ENABLED, false )
+				.setup();
```
yrodiere, line 73: This is nice, but this test assumes that indexes are empty when test methods get first executed... You broke that assumptions, and we may now have some tests that would succeed even if the mass indexing job did nothing at all: we check whether entities are indexed, and they are! But only because we didn't clear indexes between tests...

If you're going this route, I think it would be safer for the `OrmSetupHelper` to automatically clear data/indexes between tests when used in a `@BeforeAll`... and provide an option to disable data clearing for those few tests that need persistent data across tests.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/search/SearchQueryBaseIT.java
yrodiere: Hey,

Good points.

I don't think we need an annotations or a dedicated method in the test, though, do we?

You could just allow something like that:

```
setupHelper.start(...)
    .dataClearing(config -> { ... })
    .setup();
```

The call to `clearAfterTest` would store some runnable in the extension or test context, and the runnable would get executed after each test.

Also... I think we probably want data clearing to get executed by default (opt-out), at least when the session factory is set up in `@BeforeAll`. See my other comments for the "why".

util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/OrmSetupHelper.java
```diff
@@ -208,6 +217,34 @@ protected void close(SessionFactory toClose) {
...
+		this.ormSetupHelperCleaner.cleanupData( sessionFactory );
+	}
+
+	@Override
+	public void interceptTestMethod(Invocation<Void> invocation, ReflectiveInvocationContext<Method> invocationContext,
+			ExtensionContext extensionContext)
+			throws Throwable {
+		invocation.proceed();
+		this.ormSetupHelperCleaner.cleanupData( sessionFactory );
+	}
```
yrodiere, line 246: Isn't this equivalent to an `AfterEachCallback`?

marko-bekhta: yes, that's mostly the same. I've been using this invocation interceptor as initially I was looking for the "configuration parameters" that were used in the junit4 rule to detect if the changes require data cleanup or recreating the session and only `invocationContext` had that info accessible without resorting to reflection. But with that gone (comparing parameters) I'll move these to afterEach.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/search/SearchQueryBaseIT.java
marko-bekhta: That was a nice idea, I went with the annotations as that was how we were doing it before, but defining it in the helper makes more sense 😃 👍🏻. I think I've managed to make it delete the DB data if the helper is static, as for the index -- from what it seems, it's better to opt-in for that since a lot of our tests are based on the mock and if we drop-create the index -- it'll generate an additional "event" that we need to add to the expectations. + this wasn't happening in the past so that seems like a safe default ...

integrationtest/mapper/orm-jakarta-batch/src/test/java/org/hibernate/search/integrationtest/jakarta/batch/massindexing/MassIndexingJobWithCompositeIdIT.java
marko-bekhta: Hmm, I've tested this by running tests with jdk11 locally and it seems fine...
I went with this approach mostly to get build failures if a `@TestInstance(PER_CLASS)`  is missing on a test with `@BeforeAll` as we'd almost always need to have it if we do the before-all initialization. But let's run the jdk11 build on CI once we are done with everything else, and if there are any problems I'll update it to a static method.

marko-bekhta:
I've changed it to ready for review. You can now take a look at the bytecode-enhanced test utils and the data cleaning 😃

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/extension/ActualBackendSetupStrategy.java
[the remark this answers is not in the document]
marko-bekhta: this is gone now 😃 sooooo nothing to fix anymore 🙈

build/parents/build/pom.xml
```diff
@@ -631,11 +633,27 @@
             </dependency>
 
             <!-- Test -->
+            <!-- JUnit 4 is only used for legacy v5 tests that we won't be changing-->
+            <!-- Also used on some tests that we cannot migrate for now -->
             <dependency>
                 <groupId>junit</groupId>
                 <artifactId>junit</artifactId>
                 <version>${version.junit}</version>
             </dependency>
```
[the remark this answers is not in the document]
marko-bekhta: After the remaining tests are updated to junit5, this is only used for v5 migration helper tests. And looks as Surefire executes them fine within their own module.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/search/SearchQueryBaseIT.java
yrodiere: > this wasn't happening in the past

I beg to differ:

https://github.com/hibernate/hibernate-search/blob/5e8810801f9c920c5d59fd7ff201f91371673053/util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/ReusableOrmSetupHolder.java#L394-L402

> as for the index -- from what it seems, it's better to opt-in for that since a lot of our tests are based on the mock and if we drop-create the index

You could also just adjust the behavior based on whether the backend is a mock or not; that's precisely what we used to do in `ReusableOrmSetupHolder`:

https://github.com/hibernate/hibernate-search/blob/5e8810801f9c920c5d59fd7ff201f91371673053/util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/ReusableOrmSetupHolder.java#L155-L172

You could change how it works, e.g. add something to `BackendSetupStrategy`, if that's easier. But my point is you can avoid the index clearing when using a backend mock :)

marko-bekhta: riiiiiiight 😃 
When I was saying `this wasn't happening in the past` I meant in the regular `OrmSetupHelper`. I've took the bits from the reusable one so I've seen the part you are pointing to. It was done with `ReusableOrmSetupHolder#withSingleBackend` so Ive added an explicit config to clear the index in these cases... I am just wondering if it is worth to clear the index if it wasn't cleared before (in terms of time-to-execute-the-tests). If you think it won't impact the time that much and it is worth I'll figure something 😃

yrodiere: > am just wondering if it is worth to clear the index if it wasn't cleared before (in terms of time-to-execute-the-tests).

IMO if we clear the database we must clear the indexes as well (except for mock backends, obviously). Otherwise we'll run into problems.

Is clearing the indexes a big cost? Not that much, especially considering that the vast majority of ORM tests use a `BackendMock`, where we won't (shouldn't) clear the indexes anyway. For other tests, we might lose 50ms per clear; we'll survive.

Now, is it worth clearing the indexes all the time? I'd say no, it's only worth it if we clear the database, and that should only happen when the `SessionFactory` was created in a `@BeforeAll`... and I think you already implemented that optimization, so it should be fine :)

marko-bekhta:
Found one more "missing" test and updated it so it gets executed. I've opened https://hibernate.atlassian.net/browse/HSEARCH-4989 to address the jqassistant additional rules + to make sure that it is executed correctly. I'll check the counts after this run to make sure that that missing test is now executed, but other than that, I don't plan to make more updates to this PR 😃

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/search/loading/SearchQueryEntityLoadingGraphIT.java
```diff
@@ -70,25 +63,34 @@ protected BackendMock backendMock() {
...
+	@MethodSource("params")
+	void setup(SingleTypeLoadingModel<T> model, SingleTypeLoadingMapping mapping) {
+		this.model = model;
+		this.mapping = mapping;
+		backendMock.expectAnySchema( model.getIndexName() );
+		sessionFactory = ormSetupHelper.start().withConfiguration( c -> mapping.configure( c, model ) )
+				.dataClearing( true, config -> config.clearIndexData( false )
+						.preClear( model.getIndexedClass(), model::clearContainedEager )
+						.clearOrder( model.getContainedClass(), model.getIndexedClass() ) )
+				.setup();
```
yrodiere, line 89: I just tried to run this test, and it appears that session factories all get closed at the very end of the test run, instead of being closed when moving from one setup to the next. This could cause serious problems. Could you please have a look?

To reproduce, put a breakpoint on this line as well as in `org.hibernate.internal.SessionFactoryImpl#close`, then run the test in debug mode.

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/extension/MappingSetupHelper.java
```diff
@@ -89,44 +48,48 @@ public String toString() {
...
 		return backendSetupStrategy.start( setupContext, configurationProvider, setupContext.backendMappingHandlePromise );
 	}
 
 	@Override
 	public void afterAll(ExtensionContext context) throws Exception {
-		delegate.afterAll( context );
+		configurationProvider.afterAll( context );
+		if ( callOncePerClass ) {
+			cleanUp();
+		}
```
yrodiere, line 67: I don't understand how this is supposed to work, but anyway: considering there may be multiple session factories created, potentially each at a different time (beforeall, beforeeach, test, ...), wouldn't it make more sense to attach a "scope" to every created resource, and decide whether to clean it up based on that?

E.g. you could have two `toClose` lists, `toCloseAfterTest` and `toCloseAfterAll`. In `setup` you'd put the closeable in the relevant list based on the current value of `callOncePerClass` (`currentScope` would be a better name BTW). And then you execute the corresponding cleanup in the corresponding callbacks (afterEach/afterAll).

This might be related to the problem I noticed in `SearchQueryEntityLoadingGraphIT`, BTW.

util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/OrmSetupHelper.java
```diff
@@ -117,43 +110,48 @@ private static OrmSetupHelper withBackendMocks(Type type, BackendMock defaultBac
...
-	private CoordinationStrategyExpectations coordinationStrategyExpectations =
-			DEFAULT_COORDINATION_STRATEGY_EXPECTATIONS;
-
-	private OrmSetupHelper(BackendSetupStrategy backendSetupStrategy, Collection<BackendMock> backendMocks,
-			SchemaManagementStrategyName schemaManagementStrategyName, Type type) {
-		super( backendSetupStrategy, type );
+	private final BiConsumer<Boolean, String> assumptionTrueCheck;
+	private OrmSetupHelperCleaner ormSetupHelperCleaner;
+	private CoordinationStrategyExpectations coordinationStrategyExpectations = DEFAULT_COORDINATION_STRATEGY_EXPECTATIONS;
+	private SessionFactoryImplementor sessionFactory;
```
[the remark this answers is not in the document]
yrodiere: Right, it'll be a bit harder, but see my suggestion in another comment.

util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/OrmSetupHelper.java
```diff
@@ -189,17 +192,39 @@ protected void close(SessionFactory toClose) {
 		}
 	}
 
+	@Override
+	public void afterTestExecution(ExtensionContext context) {
+		// if test was aborted then we don't want to clean the data since the test wasn't executed.
+		if ( !context.getExecutionException().map( Object::getClass )
+				.map( org.opentest4j.TestAbortedException.class::equals ).orElse( Boolean.FALSE ) ) {
+			this.ormSetupHelperCleaner.cleanupData( sessionFactory );
+		}
```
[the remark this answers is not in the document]
marko-bekhta: mm right... I've changed the order somewhat on purpose 😃. assuming we have a test that configures this helper (and builds a session factory) in a before method we can't do the cleanup earlier then this before is executed... then the helper will destroy the session factory in the corresponding after (all/each). so that's why it cleans the data after the execution and uses the same session factory that was used to run the tests....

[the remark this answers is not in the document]
yrodiere: > assuming we have a test that configures this helper (and builds a session factory) in a before method we can't do the cleanup earlier then this before is executed

If it's a `@BeforeEach` method, then there's no point clearing data anyway, the session factory should be closed.

If it's a `@BeforeAll` method... I don't really understand the problem. You can (and probably should) run your stuff in a beforeEachCallback, no?

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/search/loading/SearchQueryEntityLoadingGraphIT.java
```diff
@@ -70,25 +63,34 @@ protected BackendMock backendMock() {
...
 	}
 
-	@Before
-	public void initData() {
+	@ParameterizedSetupBeforeTest
```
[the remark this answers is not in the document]
marko-bekhta: Those before/after callbacks are a bit more hidden, and I didn't see a clean way to get to them... 
The only problem is really that `@BeforeEach` is executed before the setup method... but `@AfterEach` is executed as "expected" after each test method + `*All` are also both fine. So I'll add a check only for a `@BeforeEach`.

util/internal/test/common/src/main/java/org/hibernate/search/util/impl/test/extension/parameterized/ParameterizedClassPostDiscoveryFilter.java
```diff
@@ -28,6 +31,19 @@
...
+			// Not really related to parameterized tests but to not create another filter just for it ...
+			// We want to double-check that we don't have any static classes with @Nested annotation:
+			for ( Class<?> inner : testClass.getDeclaredClasses() ) {
+				if ( Modifier.isStatic( inner.getModifiers() ) && isAnnotated( inner, Nested.class ) ) {
+					throw new IllegalStateException(
+							"Nested test classes annotated with @Nested cannot be static, as they won't be executed. "
+									+ "Make " + inner + " non-static or remove the annotation (@Nested)." );
+				}
+			}
+		}
```
yrodiere, line 45: Since this isn't related to `@ParameterizedSetup`, it would probably be best to move it to a separate filter?

pom.xml
```diff
@@ -810,15 +810,6 @@
...
                     <dependencies>
-                        <!--
-                             Force tests to execute with JUnit 4
-                             See http://maven.apache.org/surefire/maven-failsafe-plugin/examples/junit.html
-                         -->
-                        <dependency>
-                            <groupId>org.apache.maven.surefire</groupId>
-                            <artifactId>surefire-junit47</artifactId>
-                            <version>${version.surefire.plugin}</version>
-                        </dependency>
```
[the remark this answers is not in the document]
yrodiere: > Before merging, we will need to check that the number of tests executed before this PR is consistent with the number of tests executed after.


Can you confirm everything was fine when you checked with the latest version of this PR? For posterity :)

yrodiere (approved):
This looks bulletproof now, and the bytecode enhancement stuff seems to work flawlessly! Great!

I must admit I didn't go through many tests, but I checked a few and everything looks good.

I added a few comments, but feel free to merge as soon as you're satisfied. If this happens soon, we could have this in 7.0... :)

pom.xml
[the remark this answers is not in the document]
marko-bekhta: 😄 yeah I was looking into it, actually. Seems like some tests got reported into a different group, so trying to figure out what happened (more like, what can we modify in the plugin config related to reporting, or should we even modify it)

util/internal/test/common/src/main/java/org/hibernate/search/util/impl/test/extension/parameterized/ParameterizedClassExtension.java
```diff
@@ -199,7 +216,10 @@ public void interceptTestTemplateMethod(Invocation<Void> invocation,
 					read( extensionContext, StoreKey.TEST_TO_RUN, ParameterizedTestMethodInvoker.class );
 			testMethod.invoke( extensionContext.getRequiredTestInstance() );
 			envInitialized = !envInitializationFailed;
+
+			if ( !hasMoreTests ) {
+				cleanUp.accept( ExtensionScope.PARAMETERIZED_CLASS_SETUP );
+			}
```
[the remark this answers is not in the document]
marko-bekhta: I've tried a few things but because of this:

> That means that, given two extensions Extension1 and Extension2 with Extension1 registered before Extension2, any "before" callbacks implemented by Extension1 are guaranteed to execute before any "before" callbacks implemented by Extension2. Similarly, given the two same two extensions registered in the same order, any "after" callbacks implemented by Extension1 are guaranteed to execute after any "after" callbacks implemented by Extension2. Extension1 is therefore said to wrap Extension2.

the only way I figured that worked is to register an extension with the `@ParameterizedClass` annotation (see the `ParameterizedClassCleanupExtension`). It means that we need to store the boolean in the extension store, and read it in the other extension, but ... it works 😃

pom.xml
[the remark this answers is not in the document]
marko-bekhta: ok, so the tests are all executed. Cannot see why `ObjectProjectionBaseIT` and `CompositeProjectionBaseIT` get reported into root tests though...

yrodiere (approved):

pom.xml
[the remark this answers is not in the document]
yrodiere: No idea, but while looking at those tests I saw that `SearchSetupHelper` was still using this `callOncePerClass` stuff that you removed from `MappingSetupHelper`. Though I'd mention it so that you can adjust this if necessary.

yrodiere (approved):

pom.xml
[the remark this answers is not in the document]
yrodiere: @marko-bekhta Can we merge now? :)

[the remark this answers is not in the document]
marko-bekhta: yes 😃 🎉

marko-bekhta:
Thank you @yrodiere for all the suggestions! 😃🎉

PR #3690, opened 2023-08-29
HSEARCH-4924 Follow-up: ensure Jacoco coverage numbers are correct in default build + speed up some Maven executions on CI

yrodiere:

Test coverage is currently at 0 on Maven because the modules that contain classes that we need coverage numbers for *must* be included in the "default build - second round" that runs ITs. Otherwise the Jacoco execution will not report coverage for the corresponding classes.

This fixes that, and a bit more.


yrodiere:
FWIW this finally fixed Sonar reports: https://sonarcloud.io/summary/overall?id=org.hibernate.search%3Ahibernate-search-parent \o/

PR #3691, opened 2023-08-29
Remove a few build environments that are not actually available on CI

yrodiere:
Because this gets in the way when we run all tests: we always get failures for those few environments that we know cannot work.


marko-bekhta (approved):
👍🏻 😃

PR #3693, opened 2023-08-30
HSEARCH-4914 Handle transaction deadlocks more gracefully when retrieving outbox events (especially on CockroachDB)

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4914

Essentially I'm giving up on avoiding deadlocks on CockroachDB. We'll retry instead, and will just avoid reporting errors.

### DIFF
--- a/mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/event/impl/OutboxPollingEventProcessor.java
+++ b/mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/event/impl/OutboxPollingEventProcessor.java
@@ -18,6 +18,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
 import jakarta.persistence.OptimisticLockException;
+import jakarta.persistence.PessimisticLockException;
 
 import org.hibernate.engine.spi.SessionImplementor;
 import org.hibernate.search.engine.backend.orchestration.spi.SingletonTask;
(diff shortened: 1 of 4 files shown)


mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/event/impl/OutboxPollingEventProcessor.java
```diff
@@ -275,9 +277,21 @@ public CompletableFuture<?> work() {
...
+					List<OutboxEvent> events;
+					try {
+						events = eventFinder.get().findOutboxEvents( session, batchSize );
+						if ( events.isEmpty() ) {
+							// Nothing to do, try again later (complete() will be called, re-scheduling the polling for later)
+							return;
+						}
+					}
+					catch (OptimisticLockException lockException) {
+						// Don't be fooled by the exception type, this is actually a *pessimistic* lock failure.
```
marko-bekhta, line 290: That's ... quite intuitive 😄 
Is it CockroachDB that returns false info, or it's just the way ORM behaves?
BTW, should we catch the pessimistic one as well?

marko-bekhta (approved):
👍🏻 😃

mapper/orm-coordination-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/coordination/outboxpolling/event/impl/OutboxPollingEventProcessor.java
yrodiere: >  Is it CockroachDB that returns false info, or it's just the way ORM behaves?

Just the way ORM behaves.

>  BTW, should we catch the pessimistic one as well?

I'll do that, but I've never seen ORM throw one.

PR #3694, opened 2023-08-30
HSEARCH-4920 Add CI jobs to test against the latest version of Lucene/Elasticsearch

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4920

I haven't found if there are snapshots for ES/OS images, so I haven't added anything for them...
As for the Lucene upgrade, I thought we could run the upgrade tests for the current 9 series as well as for the upcoming 10.

### DIFF
--- a/integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/util/ElasticsearchTckBackendFeatures.java
+++ b/integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/util/ElasticsearchTckBackendFeatures.java
@@ -200,6 +200,14 @@ public class ElasticsearchTckBackendFeatures extends TckBackendFeatures {
 		);
 	}
 
+	@Override
+	public boolean supportsHighlighterUnifiedPhraseMatching() {
+		return isActualVersion(
+				esVersion -> !esVersion.isLessThan( "8.10" ),
+				osVersion -> false
+		);
+	}
+
 	public static boolean supportsIndexClosingAndOpening() {
 		return isActualVersion(
 				// See https://docs.aws.amazon.com/opensearch-service/latest/developerguide/supported-operations.html#version_7_1
--- a/util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/dialect/ElasticsearchTestDialect.java
+++ b/util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/dialect/ElasticsearchTestDialect.java
@@ -6,12 +6,16 @@
  */
 package org.hibernate.search.util.impl.integrationtest.backend.elasticsearch.dialect;
 
+import java.io.IOException;
 import java.util.Comparator;
+import java.util.Optional;
 import java.util.function.Predicate;
 
 import org.hibernate.search.backend.elasticsearch.ElasticsearchDistributionName;
 import org.hibernate.search.backend.elasticsearch.ElasticsearchVersion;
 import org.hibernate.search.util.impl.integrationtest.backend.elasticsearch.ElasticsearchTestHostConnectionConfiguration;
+import org.hibernate.search.util.impl.integrationtest.backend.elasticsearch.rule.TestElasticsearchClient;
+import org.hibernate.search.util.impl.integrationtest.common.TestConfigurationProvider;
 
 public class ElasticsearchTestDialect {
 
@@ -29,8 +33,28 @@ public class ElasticsearchTestDialect {
 			ElasticsearchDistributionName distribution =
 					ElasticsearchDistributionName.of( System.getProperty(
 							"org.hibernate.search.integrationtest.backend.elasticsearch.distribution" ) );
-			String versionString = System.getProperty(
-					"org.hibernate.search.integrationtest.backend.elasticsearch.version" );
+			String versionString = "";
+
+			if ( distribution != ElasticsearchDistributionName.AMAZON_OPENSEARCH_SERVERLESS ) {
+				try ( TestElasticsearchClient client = new TestElasticsearchClient() ) {
+					client.open( new TestConfigurationProvider(), Optional.empty() );
+					versionString = client.getActualVersion();
+					if ( versionString != null ) {
+						// If we got a snapshot version back from the actual cluster we want to drop that qualifier part of version.
+						// Qualifiers are not allowed when comparing versions with a comparator from this class
+						// (see this#compare(ElasticsearchVersion a, ElasticsearchVersion b, int defaultInt)).
+						int dashIndex = versionString.indexOf( '-' );
+						if ( dashIndex > -1 ) {
+							versionString = versionString.substring( 0, dashIndex );
+						}
+					}
+				}
+				catch (IOException e) {
+					throw new IllegalStateException(
+							"Wasn't able to detect the actual version of " + distribution.toString() + "cluster", e );
+				}
+			}
+
 			actualVersion = ElasticsearchVersion.of( distribution, versionString.isBlank() ? null : versionString );
 		}
 		return actualVersion;
--- a/build/parents/build/pom.xml
+++ b/build/parents/build/pom.xml
@@ -1344,6 +1344,22 @@
                         <updatePolicy>always</updatePolicy>
                     </snapshots>
                 </repository>
+                <repository>
+                    <id>apache-snapshots</id>
+                    <url>https://repository.apache.org/content/groups/snapshots</url>
+                    <layout>default</layout>
+                    <releases>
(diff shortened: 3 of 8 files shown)


yrodiere (commented):
LGTM for the Lucene part. Did you test those builds locally, just to be sure it works? See https://github.com/hibernate/hibernate-search/blob/main/MAINTAINERS.md#dependency-update-pipeline

As for OpenSearch/Elasticsearch, I can see [mentions of an snapshots repo in the OpenSearch source code](https://github.com/opensearch-project/OpenSearch/blob/79e5aeed7adc5a02d3d974378d1b7f29df3ded3e/build.gradle#L352C12-L352C55), [OpenSearch (rarely) releases betas/rcs that we could take advantage of](https://hub.docker.com/r/opensearchproject/opensearch/tags?page=1&ordering=last_updated&name=rc), [I can see SNAPSHOTs of Elasticsearch containers in their own docker registry](https://www.docker.elastic.co/r/elasticsearch?show_snapshots=true), and [I found references to snapshot repositories in other Elastic projects](https://github.com/elastic/elasticsearch-hadoop/blob/b0ad5c085a0876eb46ff18a3bcfa922447417f7f/qa/build.gradle#L42-L43C35).
So, could you please give it another try? Be it just to explain why none of the above would work.
Note it might be more complex than for Lucene, I'm not denying that. See the Jira issue for some thoughts.

marko-bekhta:
Hey, thanks for the links! I was looking at the ES docs and didn't find a snapshot repository (especially since there's a concept of snapshots in ES so 😄)
I have tested the Lucene 9 upgrade, but since it was failing if only a test module was built (I assumed that it might have to do with pulling snapshots) I wanted to see if it'll be ok on CI (Running a full build with the upgrade did work).
But now since you've asked 😃 I've tested the Lucene 10 upgrade as well and turned out there's a missing constant `org.apache.lucene.util.automaton.RegExp#COMPLEMENT` in the main branch... but it is not deprecated or anything like that in the current 9x series ... so I've removed the testing of Lucene 10 snapshots for now....
As for the Elasticsearch client -- the last published snapshot that I've found was 8.5.4 ... so since it's a bit dated 😃 I've tried running the tests against the latest container snapshot but keeping the client at whatever version it is in the pom.

yrodiere:
>  But now since you've asked 😃 I've tested the Lucene 10 upgrade as well and turned out there's a missing constant org.apache.lucene.util.automaton.RegExp#COMPLEMENT in the main branch... but it is not deprecated or anything like that in the current 9x series ... so I've removed the testing of Lucene 10 snapshots for now....

Please keep it, we need to address that. Maybe in a `wip/` branch? And we need to deprecate this flag in Hibernate Search 7.0... perhaps even 6.2.2?

>  As for the Elasticsearch client -- the last published snapshot that I've found was 8.5.4 ... so since it's a bit dated 😃 I've tried running the tests against the latest container snapshot but keeping the client at whatever version it is in the pom.

Yes, not upgrading the client is fine.

Oh, so they have a float tag for the latest version, that's great.

yrodiere:
Any luck on the OpenSearch side of things?

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/util/ElasticsearchTckBackendFeatures.java
```diff
@@ -200,6 +201,19 @@ public boolean supportsHighlighterPlainOrderByScoreMultivaluedField() {
...
+				esVersion -> !esVersion.isLessThan( "8.10" ) || isLatestSnapshotContainer(),
+				osVersion -> false
+		);
+	}
+
+	public static boolean isLatestSnapshotContainer() {
+		return System.getProperty( "org.hibernate.search.integrationtest.backend.elasticsearch.image.tag", "" )
+				.toLowerCase( Locale.ROOT ).contains( "snapshot" );
+	}
+
```
yrodiere, line 216: This should probably be in `ElasticsearchTestDialect`?

And TBH in this case I'm wondering if we shouldn't rather try to guess the version in `ElasticsearchTestDialect` by sending a request to the container in `getActualVersion`... I.e.:


Unset the version in Maven properties:

```
-Dtest.elasticsearch.run.version=
```

Add this in `ElasticsearchTestDialect#getActualVersion`:

```
if ( distribution != AMAZON_OPENSEARCH_SERVERLESS && (version == null || version.isBlank()) ) {
    try ( TestElasticsearchClient client = new TestElasticsearchClient() ) {
        client.open(...); 
        // This will probably use org.hibernate.search.backend.elasticsearch.client.impl.ElasticsearchClientUtils#tryGetElasticsearchVersion
        version = client.callSomeMethodToGetTheVersion();
    }
}
```

marko-bekhta: mmmm, ok I'll give that a try 😃

util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/dialect/ElasticsearchTestDialect.java
```diff
@@ -29,8 +33,26 @@ public static ElasticsearchVersion getActualVersion() {
...
+					if ( versionString != null ) {
+						// If we got a snapshot version back from the actual cluster we want to drop that qualifier part of version.
+						// Qualifiers are not allowed when comparing versions with a comparator from this class
+						// (see this#compare(ElasticsearchVersion a, ElasticsearchVersion b, int defaultInt)).
+						versionString = versionString.substring( 0, versionString.indexOf( '-' ) );
+					}
+				}
+				catch (IOException e) {
+					versionString =
+							System.getProperty( "org.hibernate.search.integrationtest.backend.elasticsearch.version", "" );
```
yrodiere, line 52: So if this property is set, it's ignored by default in favor of automatic detection? Not a great UX xD

Either:

* you're sure this automatic detecting will always work and feel free to just remove `org.hibernate.search.integrationtest.backend.elasticsearch.version` everywhere
* or you're not and we'd better take `org.hibernate.search.integrationtest.backend.elasticsearch.version` into account when it's set.

In any case, you should probably update the documentation (CONTRIBUTING.md or MAINTAINERS.md) accordingly :)

util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/dialect/ElasticsearchTestDialect.java
```diff
@@ -29,8 +33,26 @@ public static ElasticsearchVersion getActualVersion() {
...
 					ElasticsearchDistributionName.of( System.getProperty(
 							"org.hibernate.search.integrationtest.backend.elasticsearch.distribution" ) );
-			String versionString = System.getProperty(
-					"org.hibernate.search.integrationtest.backend.elasticsearch.version" );
+			String versionString = "";
+
+			if ( distribution != ElasticsearchDistributionName.AMAZON_OPENSEARCH_SERVERLESS ) {
+				try ( TestElasticsearchClient client = new TestElasticsearchClient() ) {
+					client.open( new TestConfigurationProvider(), Optional.empty() );
+					// This will probably use org.hibernate.search.backend.elasticsearch.client.impl.ElasticsearchClientUtils#tryGetElasticsearchVersion
```
yrodiere, line 41: That was for you, not to be kept in the codebase...

```suggestion
```

yrodiere (changes requested):
Thanks. A few comments.

util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/dialect/ElasticsearchTestDialect.java
yrodiere: Thanks.

> In any case, you should probably update the documentation (CONTRIBUTING.md or MAINTAINERS.md) accordingly :)

Ok so I see `test.elasticsearch.version` is still needed regardless, so let's ignore that suggestion.

yrodiere (approved):

yrodiere:
Merging, thanks

util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/dialect/ElasticsearchTestDialect.java
marko-bekhta: Yeah... unless we want to pass the version of es container to start in some other way....

PR #3695, opened 2023-08-31
HSEARCH-4924 Follow-up: Fix markStageSkipped() marking all parallel executions as skipped

yrodiere:

This is to fix the fact that all non-default environments are marked as skipped as soon as one of them is, like here: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-3679/22/pipeline

Let's see if it works...


marko-bekhta (approved):
🤞🏻 😃

PR #3699, opened 2023-09-01
HSEARCH-4933 Follow-up: Fix incorrect bytebuddy changes

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4933

I used the wrong version of bytebuddy and forgot that we also need to test against ORM 5, where we still need `net.bytebuddy.experimental=true` on Java 21...

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2710,6 +2710,11 @@
                 </property>
             </activation>
             <properties>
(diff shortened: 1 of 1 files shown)


yrodiere:
Now why my using the wrong version of bytebuddy didn't fail the build... I'd very much like to know that. I will have a look.

marko-bekhta (approved):

PR #3701, opened 2023-09-01
[WIP] Lucene 10 SNAPSHOT upgrade

marko-bekhta:
Apparently, deprecating Lucene's constant `NO_MORE_ORDS` wasn't just for fun on their side 😄.
In Lucene 10 stored set doc values wouldn't return -1 when they are "finished"... For example `SingletonSortedSetDocValues` would just always return the ord it has leading to a never-ending loop, others can end up throwing some exceptions...

I've tested the patched  `DocValuesTextMultiValues` and `JoiningTextMultiValuesSource` on main with Lucene 9 and it passed the TCK tests. I think it probably would be better to make these in the main, and let it be a part of Search 7.0.

Then as for the missing `RegexpQueryFlag.COMPLEMENT` counterpart in Lucene ... I've removed it from Lucene backend, but left it as is in the tests and Search enum until we decide how we want to treat that option in general...

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/docvalues/impl/TextMultiValues.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/docvalues/impl/TextMultiValues.java
@@ -22,8 +22,6 @@ import org.apache.lucene.index.SortedSetDocValues;
  */
 public abstract class TextMultiValues {
 
-	protected static final long NO_MORE_ORDS = -1L;
-
 	/**
 	 * Sole constructor. (For invocation by subclass
 	 * constructors, typically implicit.)
@@ -91,6 +89,8 @@ public abstract class TextMultiValues {
 
 		protected final SortedSetDocValues values;
 		protected long nextOrd;
+		protected int docValueCount = 0;
+		protected int currentDocIndex = 0;
 
 		DocValuesTextMultiValues(SortedSetDocValues values) {
 			this.values = values;
@@ -99,20 +99,25 @@ public abstract class TextMultiValues {
 		@Override
 		public boolean advanceExact(int doc) throws IOException {
 			boolean found = values.advanceExact( doc );
-			nextOrd = found ? values.nextOrd() : NO_MORE_ORDS;
+			if ( found ) {
+				nextOrd = values.nextOrd();
+				docValueCount = values.docValueCount();
+				currentDocIndex = 0;
+			}
 			return found;
 		}
 
 		@Override
 		public boolean hasNextValue() throws IOException {
-			return nextOrd != NO_MORE_ORDS;
+			return currentDocIndex < docValueCount;
 		}
 
 		@Override
 		public long nextOrd() throws IOException {
-			long result = nextOrd;
+			currentDocIndex++;
+			long previousOrd = nextOrd;
 			nextOrd = values.nextOrd();
-			return result;
+			return previousOrd;
 		}
 
 		@Override
--- a/documentation/src/test/java/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java
+++ b/documentation/src/test/java/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java
@@ -894,7 +894,7 @@ public class PredicateDslIT {
 			hits = searchSession.search( Book.class )
 					.where( f -> f.regexp().field( "description" )
 							.matching( "r@t" )
-							.flags( RegexpQueryFlag.ANY_STRING, RegexpQueryFlag.COMPLEMENT )
+							.flags( RegexpQueryFlag.ANY_STRING )
 					)
 					.fetchHits( 20 );
 			// end::regexp-flags[]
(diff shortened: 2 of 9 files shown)


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/docvalues/impl/TextMultiValues.java
```diff
@@ -99,20 +99,25 @@ protected static class DocValuesTextMultiValues extends TextMultiValues {
 		@Override
 		public boolean advanceExact(int doc) throws IOException {
 			boolean found = values.advanceExact( doc );
-			nextOrd = found ? values.nextOrd() : NO_MORE_ORDS;
+			if ( found ) {
+				nextOrd = values.nextOrd();
+				docValueCount = values.docValueCount();
+				currentDocIndex = 0;
+			}
```
yrodiere, line 106: Shouldn't there be an `else` here to reset these variables to `0`, just to be safe?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/docvalues/impl/TextMultiValues.java
```diff
@@ -99,20 +99,25 @@ protected static class DocValuesTextMultiValues extends TextMultiValues {
 		@Override
 		public boolean advanceExact(int doc) throws IOException {
 			boolean found = values.advanceExact( doc );
-			nextOrd = found ? values.nextOrd() : NO_MORE_ORDS;
+			if ( found ) {
+				nextOrd = values.nextOrd();
+				docValueCount = values.docValueCount();
+				currentDocIndex = 0;
```
yrodiere, line 105: If I understand correctly what this variable stands for, I think a better name would be `nextOrdIndex`?

documentation/src/test/java/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java
```diff
@@ -894,7 +894,7 @@ public void regexp() {
 			hits = searchSession.search( Book.class )
 					.where( f -> f.regexp().field( "description" )
 							.matching( "r@t" )
-							.flags( RegexpQueryFlag.ANY_STRING, RegexpQueryFlag.COMPLEMENT )
+							.flags( RegexpQueryFlag.ANY_STRING )
```
yrodiere, line 897: We will need to update the documentation as well.

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/testsupport/util/LuceneTckBackendFeatures.java
```diff
@@ -58,4 +58,9 @@ public boolean supportsHighlighterUnifiedTypeFragmentSize() {
 	public boolean supportsHighlighterUnifiedPhraseMatching() {
 		return true;
 	}
+
+	@Override
+	public boolean supportsComplementRegexFlag() {
+		return false;
+	}
```
yrodiere, line 65: I don't think you need to bother with this. Let's remove it for all backends. Elasticsearch will likely remove support when it upgrades to Lucene 10 anyway.

yrodiere (commented):
Thanks.

> I've tested the patched `DocValuesTextMultiValues` and `JoiningTextMultiValuesSource` on main with Lucene 9 and it passed the TCK tests. I think it probably would be better to make these in the main, and let it be a part of Search 7.0.

Agreed. I created https://hibernate.atlassian.net/browse/HSEARCH-4939. Please send a PR for that?

> Then as for the missing `RegexpQueryFlag.COMPLEMENT` counterpart in Lucene ... I've removed it from Lucene backend, but left it as is in the tests and Search enum until we decide how we want to treat that option in general...

Let's deprecate it in 6.2 and remove it in 7.0.
I opened https://hibernate.atlassian.net/browse/HSEARCH-4937 and https://hibernate.atlassian.net/browse/HSEARCH-4938 .

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/testsupport/util/LuceneTckBackendFeatures.java
marko-bekhta: mm, ok 😃. I was only wondering if that'll match whatever the "compatible" versions of Elasticsearch/OpenSearch would be at the time we switch to Lucene 10.

yrodiere: I meant that we'll remove the feature from all backends, so we'll remove the corresponding tests as well, so we won't need this.

Some Elasticsearch/OpenSearch versions may support the operator, but Hibernate Search won't, so we don't really need to bother with this.

At least that's how I see it.

marko-bekhta:
> Agreed. I created https://hibernate.atlassian.net/browse/HSEARCH-4939. Please send a PR for that?

Done: https://github.com/hibernate/hibernate-search/pull/3703 that's the PR 😃

marko-bekhta:
with https://github.com/hibernate/hibernate-search/pull/3710 we wouldn't need an upgrade branch for lucene 10 (at least for now). closing.

PR #3708, opened 2023-09-05
Check that the branch name follows a simple pattern without any unexpected symbols that might need escaping

marko-bekhta:
I've initially thought about wrapping the line
```
-Dsonar.pullrequest.branch=${helper.scmSource.branch.name} \
```
with some double quotes, but then thought that maybe it'll be better to just not allow doing "crazy" branch names in the first place 😃


yrodiere (approved):
Fair enough

PR #3713, opened 2023-09-07
HSEARCH-4943 Add notes on differences between ORM 5 and ORM 6 outbox tables updates

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4943


yrodiere (approved):
Thanks.
I'm not sure "In case of using" is proper English... Maybe "When using" instead?
Anyway, feel free to merge.

PR #3719, opened 2023-09-14
HSEARCH-4947 - fixing loading jar files on windows

liptga:
HSEARCH-4947


yrodiere:
Thanks for the PR, but this doesn't seem like the right way to fix this as other tests are now failing.

I'll let @marko-bekhta have a look when he's back on a Windows-enabled workstation, and I'll personally _try_ to set up a Windows CI.

yrodiere:
Superseded by #3730. Thanks again!

PR #3722, opened 2023-09-15
HSEARCH-4934 Drop compatibility with older versions of Elasticsearch/OpenSearch

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4934


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/dialect/impl/ElasticsearchDialectFactory.java
```diff
@@ -78,6 +78,7 @@ public ElasticsearchModelDialect createModelDialect(ElasticsearchVersion version
 
 	private ElasticsearchModelDialect createModelDialectElastic(ElasticsearchVersion version) {
 		OptionalInt majorOptional = version.majorOptional();
+		OptionalInt minorOptional = version.minor();
```
yrodiere, line 81: I think you forgot about protocol dialects?

Also, aren't there some methods in `TckBackendFeatures` that you could remove because they now always return the same value for all supported versions?

yrodiere (approved):

yrodiere:
Merged, thanks!

PR #3724, opened 2023-09-15
HSEARCH-4944 Use correct project id to recompile for ORM 6.3 upgrade tests

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4944


yrodiere (commented):
I think you need the -orm6 suffix on this branch, no ? Fwiw there are instructions in maintainers.md on how to run this stuff locally do you can check if it works.

marko-bekhta:
> I think you need the -orm6 suffix on this branch, no

True! I've run the tests for `hibernate-search-integrationtest-mapper-orm`, the same I've done for main while testing the patch... but the main in 6.2 isn't ORM6 🤦🏻‍♂️ 😃 
thanks for noticing. 
Now I've tested the correct artifact locally, so it should be better now...

PR #3729, opened 2023-09-18
Restore GitHub Actions workflow and build on Windows

yrodiere:
Draft, since @marko-bekhta is working on making the build pass on Windows.

PR #3730, opened 2023-09-19
HSEARCH-4947 Make sure that nested jars can be read on Windows

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4947

The idea here is that since we are trying to construct a URI then we should construct it from the URI 😃 
Since in the case of Linux `Path` path would have the same `/` that is needed for a URI, but then in Windows, it'll be `\`, and URI construction won't work...
And the same is about the test updates; the changes are to make the asserts compare the URIs rather than Path vs URI

I've reverted the ORM version to 6.2 series to test the spring boot uberjar module -- the tests passed with these changes.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,6 +64,10 @@ jobs:
           # We can't start Linux containers on GitHub Actions' Windows VMs,
           # so we can't run Elasticsearch tests.
           # See https://github.com/actions/runner-images/issues/1143#issuecomment-972929995
+          # For some reason docker-maven-plugin will error out with "All pipe instances are busy"
+          # on GitHub Actions' Windows VMs (when using Bash?),
+          # so we also explicitly disable docker-maven-plugin.
+          # See https://github.com/fabric8io/docker-maven-plugin/issues/548#issuecomment-255477600
           - {
             name: "Windows JDK 17",
             runs-on: 'windows-latest',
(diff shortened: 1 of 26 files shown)


yrodiere (approved):
LGTM, but let's merge #3729 first and rebase this one, so we can run the relevant tests on the Windows CI.

yrodiere:
@marko-bekhta Can you please open a backport PR for 6.2?

marko-bekhta:
ok will do it once we run this with Win tests on CI 😃

yrodiere:
@marko-bekhta You can rebase now :)

yrodiere:
... and powershell doesn't support escaping newlines, nice. I'll push a fix to your branch.

yrodiere:
I disabled Elasticsearch tests on Windows, because of https://github.com/actions/runner-images/issues/1143#issuecomment-972929995

Let's see if tests using the stub and Lucene backends pass, at least.

yrodiere (approved):

yrodiere:
> I've rebased this PR and removed `ElasticsearchSchemaManagerExporterIT` and `LuceneSchemaManagerExporterIT` from the `it-multiplebackends` run since they require the backend to be set explicitly. We are testing the backend-specific schema export in corresponding backend runs ....

Right, that's what I meant to do to begin with... sorry :/

yrodiere:
FYI the outbox-polling tests now get executed, but some of them on Windows for some reason... and it's hard to know what's going on because we don't get a build report because of https://github.com/actions/upload-artifact/issues/240 :roll_eyes: 
I tried to find a workaround to actually upload a build report, and I'll try to see what's going on.

yrodiere:
Ok, I disabled the outbox-polling tests on Windows for now and created https://hibernate.atlassian.net/browse/HSEARCH-4965 to investigate the failures later.
If the build passes now, let's merge.
Sorry for the delays.

PR #3731, opened 2023-09-19
HSEARCH-4957 Upgrade to Jackson 2.15.2

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4957

### DIFF
--- a/backend/elasticsearch/pom.xml
+++ b/backend/elasticsearch/pom.xml
@@ -51,6 +51,14 @@
             <artifactId>gson</artifactId>
         </dependency>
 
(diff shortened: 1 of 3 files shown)


marko-bekhta:
Hmm ok 😃 I thought we wanted to keep it aligned with Elasticsearch rather than use the latest 🤷🏻‍♂️ 🙈 😃

yrodiere:
Yes that's what we wanted, but then: https://github.com/wildfly/wildfly/pull/17191

marko-bekhta (approved):

yrodiere:
If we test it, we might as well upgrade :shrug:

PR #3734, opened 2023-09-19
HSEARCH-4959 Upgrade to ORM 6.3.1 / HSEARCH-4935 Re-enable Spring integration tests after next 6.3 ORM upgrade

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4959
https://hibernate.atlassian.net/browse/HSEARCH-4935

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -628,7 +628,8 @@ stage('Non-default environments') {
 								mavenNonDefaultBuild buildEnv, """ \
 										--fail-fast \
 										-pl ${[
-											'org.hibernate.search:hibernate-search-integrationtest-backend-elasticsearch'
+											'org.hibernate.search:hibernate-search-integrationtest-backend-elasticsearch',
+											'org.hibernate.search:hibernate-search-integrationtest-showcase-library'
 											 ].join(',')} \
 										-Dtest.elasticsearch.distribution=$buildEnv.distribution \
 										-Dtest.elasticsearch.version=$buildEnv.version \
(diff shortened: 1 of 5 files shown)


yrodiere (approved):

yrodiere:
The build is failing... MultiTenancyHelper strikes again, it seems.

marko-bekhta:
yeeeeeeeah 😄 
and additional assert was added in this commit refactoring the connection provider https://github.com/hibernate/hibernate-orm/blob/d4accd023317cb429964cb109c0b4162d1ee2675/hibernate-testing/src/main/java/org/hibernate/testing/env/ConnectionProviderBuilder.java 
And we cannot reach the `assumeTrue( dialect instanceof H2Dialect );` to skip the tests since ORM tries to open a connection to create a dialect... I've added a property `hibernate.temp.use_jdbc_metadata_defaults` to the helper that should prevent ORM from going to DB ... alternatively... we could wrap `ConnectionProvider connectionProvider = ConnectionProviderBuilder.buildConnectionProvider( tenantId );` into try-catch and if we catch an assert error we can "assume that test is only applicable for h2 DBs and we are running on non h2 one" ...

yrodiere:
> I've added a property hibernate.temp.use_jdbc_metadata_defaults to the helper that should prevent ORM from going to DB ...

Makes sense. It will probably break someday, but we can improvise when that happens.

> we could wrap ConnectionProvider connectionProvider = ConnectionProviderBuilder.buildConnectionProvider( tenantId ); into try-catch and if we catch an assert error we can "assume that test is only applicable for h2 DBs and we are running on non h2 one" ...

No thanks :D 

I guess another alternative would be to have a look at how such assumptions are implemented in the SkipForDialect/RequiresDialect annotations in hibernate testing. But if your current solution work, let's stick to that.

marko-bekhta:
> I guess another alternative would be to have a look at how such assumptions are implemented in the SkipForDialect/RequiresDialect annotations in hibernate testing

That was a nice suggestion 😃; since there's that utility `DialectContext.getDialect()`, this approach seems cleaner since we wouldn't even try to go through builder methods in this case.

yrodiere (approved):

PR #3736, opened 2023-09-19
HSEARCH-4920 Follow-up: Don't pre-pull images from elastic docker registry

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4920

I'd expect something along these lines to work: we iterate over various docker registries, and if the image cannot be pulled, it is added for an attempt to be pulled from the next registry and so on...
I've added the exact change with a different registry order to the main Jenkins file so we can test it in the PR..

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -989,7 +989,8 @@
                                     if ( isFalseString( '${test.database.run.cockroachdb.skip}' ) ) {
                                         images += '${test.database.run.cockroachdb.image.name}:${test.database.run.cockroachdb.image.tag}'
                                     }
-                                    new File('${containerImagesListFile}').append( images.join('\n') + '\n' )
+                                    // Exclude images from non-dockerhub repositories
+                                    new File('${containerImagesListFile}').append( images.findAll{ref -> !(ref ==~ /^(?!((.*\.)?docker\.io))[\w\d\.]+\/[\w\d\.]+\/[\w\d\.]+(:.+)?$/)}.join('\n') + '\n' )
                                 ]]></script>
                             </configuration>
                         </execution>


Jenkinsfile
```diff
@@ -910,13 +910,30 @@ def pullContainerImages(String mavenArgs) {
...
-		for (int i = 0; i < containerImageRefs.length; i++) {
-			containerImageRef = containerImageRefs[i]
-			docker.image( containerImageRef ).pull()
+	String[] dockerRegistries = ['https://docker.elastic.co/v1/', 'https://index.docker.io/v1/']
+	for (i in 0..<dockerRegistries.length) {
+		if (containerImageRefs.length == 0) {
+			break
 		}
+		List<String> failedImageRefs = []
+		docker.withRegistry(dockerRegistries[i], 'hibernateci.hub.docker.com') {
```
yrodiere, line 919: `hibernateci.hub.docker.com` points to credentials that will only work on dockerhub, so I'm not sure what you're going for here... ?

Essentially you only need `withRegistry()` when you need authentication, which we don't need for docker.elastic.co.

FWIW, when you don't need authentication, you also don't really need this pre-build pulling. We could simply filter out all images that obviously are not on dockerhub (because they start with `docker.elastic.co/`, ...) and have them pulled during the build.

Jenkinsfile
```diff
@@ -910,13 +910,30 @@ def pullContainerImages(String mavenArgs) {
 	if ( containerImageRefs.length == 0 ) {
 		return
 	}
-	docker.withRegistry('https://index.docker.io/v1/', 'hibernateci.hub.docker.com') {
-		// Cannot use a foreach loop because then Jenkins wants to serialize the iterator,
-		// and obviously the iterator is not serializable.
```
yrodiere, line 915: You might want to keep this comment at the top level, as the index-based loops are likely to raise an eyebrow or two.

ci/dependency-update/Jenkinsfile
```diff
@@ -42,7 +42,7 @@ Map settings() {
 					updateProperties: [],
 					onlyRunTestDependingOn: ['hibernate-search-backend-elasticsearch'],
 					// We want to use the snapshot version of an image from the ES registry since that's where they are publishing their snapshots.
-					additionalMavenArgs: '-Dtest.elasticsearch.run.elastic.image.name=docker.elastic.co/elasticsearch/elasticsearch -Dtest.elasticsearch.run.elastic.image.tag=master-SNAPSHOT',
+					additionalMavenArgs: '-Dtest.elasticsearch.run.elastic.image.tag=master-SNAPSHOT',
```
yrodiere, line 45: I'm not sure I like this change, as future me will most likely end up reading this code and wonder "but, but, if I run `docker pull elastic/elasticsearch:master-SNAPSHOT` it doesn't work!?!".

Any reason you didn't want to follow my suggestion of just... not trying to pull images that are not from dockerhub before the build?

marko-bekhta: well ... I wanted to try and have something more generic that we can potentially reuse if we start pulling images from other registries (e.g. maybe at some point from amazon for OpenSearch etc...) and ` if (containerImageRef.startsWith('docker.elastic.co/') ) { continue }` didn't look as something very scalable in those terms 😃

Jenkinsfile
```diff
@@ -914,6 +914,10 @@ def pullContainerImages(String mavenArgs) {
 		// Cannot use a foreach loop because then Jenkins wants to serialize the iterator,
 		// and obviously the iterator is not serializable.
 		for (int i = 0; i < containerImageRefs.length; i++) {
+			if (containerImageRef.startsWith('docker.elastic.co')) {
```
yrodiere, line 917: +1 would work right now. If you want to be generic, you could do something like:

```
if ( containerImageRef has two slashes && does not start with `([^/]*\.)?docker\.io` )
```

Also just so you know (since I know you don't like duplicating code between Jenkinsfiles), it would be perfectly acceptable to do the filtering [in the groovy code](https://github.com/hibernate/hibernate-search/blob/e78126f71f9cdd4973789be29602a7e7e4ae4285/pom.xml#L953-L993).

pom.xml
```diff
@@ -376,8 +376,8 @@
         <test.elasticsearch.version>${version.org.elasticsearch.latest}</test.elasticsearch.version>
 
         <test.elasticsearch.run.elastic.skip>true</test.elasticsearch.run.elastic.skip>
-        <test.elasticsearch.run.elastic.image.name>elastic/elasticsearch</test.elasticsearch.run.elastic.image.name>
-        <test.elasticsearch.run.elastic.image.tag>${test.elasticsearch.version}</test.elasticsearch.run.elastic.image.tag>
+        <test.elasticsearch.run.elastic.image.name>docker.elastic.co/elasticsearch/elasticsearch</test.elasticsearch.run.elastic.image.name>
+        <test.elasticsearch.run.elastic.image.tag>master-SNAPSHOT</test.elasticsearch.run.elastic.image.tag>
```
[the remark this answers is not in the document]
marko-bekhta: yes 😃 🙈  (`[HSEARCH-4920 TEST DON'T MERGE]`)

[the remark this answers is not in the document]
marko-bekhta: ok so that seems to work:
```
[INFO] DOCKER> Pulling from elasticsearch/elasticsearch
[INFO] DOCKER> Digest: sha256:f201f483c2d97aea6f588990118ac32d4cbfbc0423b3a0a3885efa98d4e3f64c
[INFO] DOCKER> Status: Downloaded newer image for docker.elastic.co/elasticsearch/elasticsearch:master-SNAPSHOT
[INFO] DOCKER> Pulled docker.elastic.co/elasticsearch/elasticsearch:master-SNAPSHOT in 23 seconds 
[INFO] DOCKER> [docker.elastic.co/elasticsearch/elasticsearch:master-SNAPSHOT] "elasticsearch": Start container 9e9ffae99d94
[INFO] DOCKER> [docker.elastic.co/elasticsearch/elasticsearch:master-SNAPSHOT] "elasticsearch": Waiting on url http://localhost:9200 with method GET for status 200.
```
I'm going to revert that test commit now 😃

pom.xml
```diff
@@ -989,7 +989,7 @@
                                     if ( isFalseString( '${test.database.run.cockroachdb.skip}' ) ) {
                                         images += '${test.database.run.cockroachdb.image.name}:${test.database.run.cockroachdb.image.tag}'
                                     }
-                                    new File('${containerImagesListFile}').append( images.join('\n') + '\n' )
+                                    new File('${containerImagesListFile}').append( images.findAll{ref -> !(ref ==~ /^(?!((.*\.)?docker\.io))[\w\d\.]+\/[\w\d\.]+\/[\w\d\.]+(:.+)?$/)}.join('\n') + '\n' )
```
yrodiere, line 992:
```suggestion
                                    // Exclude images from non-dockerhub repositories
                                    new File('${containerImagesListFile}').append( images.findAll{ref -> !(ref ==~ /^(?!((.*\.)?docker\.io))[\w\d\.]+\/[\w\d\.]+\/[\w\d\.]+(:.+)?$/)}.join('\n') + '\n' )
```

PR #3748, opened 2023-09-25
HSEARCH-4969 IndexOutOfBoundsException in TextMultiValues

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4969

I also aligned the implementation of LongMultiValues, though I'm not sure that was strictly necessary. It's clearer anyway IMO.


yrodiere:
Alright, ok, I need to test this a bit more.

yrodiere:
Should be good now.

marko-bekhta (approved):
🎉

PR #3749, opened 2023-09-25
HSEARCH-4928 Rename hibernate-search-mapper-orm-coordination-outbox-polling to something shorter

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4928


integrationtest/java/modules/orm-outbox-polling-elasticsearch/src/main/java/org/hibernate/search/integrationtest/java/modules/orm/elasticsearch/coordination/outboxpolling/config/MyElasticsearchAnalysisConfigurer.java
yrodiere, line 1: You might want to change the package name in tests, to match the renaming of the module?

bom/public/pom.xml
```diff
@@ -84,7 +84,7 @@
             </dependency>
             <dependency>
                 <groupId>org.hibernate.search</groupId>
-                <artifactId>hibernate-search-mapper-orm-coordination-outbox-polling</artifactId>
+                <artifactId>hibernate-search-mapper-orm-outbox-polling</artifactId>
                 <version>${project.version}</version>
             </dependency>
```
yrodiere, line 89: Maybe relocation artifacts should be present in the BOM... WDYT?
Also, what about the JSR-352 ones?

yrodiere (commented):
Thanks. LGTM overall, just a few comments.

yrodiere (approved):

PR #3752, opened 2023-09-26
Skip ES tests for Lucene-specific builds and vice-versa

yrodiere:
To avoid unnecessary failures such as [this one](https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search-dependency-update/detail/main/151/pipeline/264), where we failed to start an ES container while testing Hibernate Search against Lucene 10...


marko-bekhta (approved):
🎉 thanks !

PR #3754, opened 2023-09-28
HSEARCH-4487 + HSEARCH-4980 Two fixes for Jakarta Batch Mass Indexing job: MySQL with jdbcFetchSize=Integer.MIN_VALUE + embedded ids

yrodiere:
* [HSEARCH-4980](https://hibernate.atlassian.net/browse/HSEARCH-4980): Jakarta Batch job fails for entities with embedded ids
* [HSEARCH-4487](https://hibernate.atlassian.net/browse/HSEARCH-4487): Jakarta Batch Mass Indexing job fails with @IndexedEmbedded entity and MySQL

For HSEARCH-4980, it's just that tests were disabled so the implementation was garbage and didn't even work. I would know, I committed that nonsense. We might want to backport the fix to 6.2, though I'm not sure it's easy.

For HSEARCH-4487, the problem was that we were using a scroll on entities. When using MySQL with jdbcFetchSize=Integer.MIN_VALUE (AKA "I mean scrolling for real, not loading the whole database in memory"), this is fundamentally incompatible with lazy loading (because scrolling basically takes over the whole connection and one can't use it for lazy loading anymore).

So, I solved this by not scrolling entities anymore: we're doing things like in the MassIndexer, with one session (connection) scrolling on IDs, and the other loading entities in batches.
This obviously changes the performance profile quite a lot, so we should brace for performance regressions.
This also implies some API changes as well: in particular, I changed the "custom HQL" feature to be closer to what we have in the MassIndexer, and removed the (now nonsensical) `sessionClearInterval` setting.
But AFAIK this is the only way to solve the problem.

I won't bother backporting HSEARCH-4487 to 6.2, as backporting without breaking changes would require too much work and duplication.

[HSEARCH-4980]: https://hibernate.atlassian.net/browse/HSEARCH-4980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HSEARCH-4487]: https://hibernate.atlassian.net/browse/HSEARCH-4487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Jenkinsfile
```diff
@@ -532,7 +532,6 @@ stage('Non-default environments') {
 					String mavenBuildAdditionalArgs = ''' \
 							-pl !documentation \
 							-pl !integrationtest/mapper/orm-spring \
-							-pl !integrationtest/mapper/orm-jakarta-batch \
```
marko-bekhta, line 535: 🤞🏻 let's hope we won't get the timeout on our super faaaast DBs 🙈

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/MassIndexerFilteringTypeStep.java
```diff
@@ -14,10 +14,17 @@ public interface MassIndexerFilteringTypeStep {
...
+	 * </pre>
+	 * To filter instances that have a manager whose level is strictly less than 2.
+	 * <p>
+	 * Parameters can be used, so assuming the parameter "max" is defined
+	 * in the {@link MassIndexerReindexParameterStep#param(String, Object) next step},
+	 * this is valid as well:
+	 * <pre>
+	 *     manager.level &lt; :max
 	 * </pre>
 	 * To filter instances that have a manager whose level is strictly less than 2.
```
marko-bekhta, line 29:
```suggestion
	 * To filter instances that have a manager whose level is strictly less than {@code :max}.
```
Maybe something like ^

yrodiere:
Ok, I had to push two updates:

1. One to use stateless sessions where relevant, mostly as an optimization
2. One to fix some JQAssistant constraint violations

marko-bekhta (approved):
😃 👍🏻

Jenkinsfile
yrodiere: Good point. I just checked and we seem to be using a 30s timeout for most tests, so hopefully that will be enough...

PR #3755, opened 2023-09-29
HSEARCH-4982 Upgrade to Lucene 9.8

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4982

### DIFF
--- a/ci/dependency-update/Jenkinsfile
+++ b/ci/dependency-update/Jenkinsfile
@@ -18,6 +18,12 @@ Map settings() {
 					// we need to recompile this module since it has incompatible return type and will result in a build error:
 					additionalMavenArgs: '-pl :hibernate-search-util-internal-integrationtest-mapper-orm'
 			]
+		case 'lucene9.8':
+			return [
+					updateProperties: ['version.org.apache.lucene'],
+					onlyRunTestDependingOn: ['hibernate-search-backend-lucene'],
+					requiresSourceModifications: true
+			]
 		case 'lucene9':
 			return [
 					updateProperties: ['version.org.apache.lucene'],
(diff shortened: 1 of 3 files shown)


ci/dependency-update/Jenkinsfile
```diff
@@ -18,6 +18,12 @@ Map settings() {
 					// we need to recompile this module since it has incompatible return type and will result in a build error:
 					additionalMavenArgs: '-pl :hibernate-search-util-internal-integrationtest-mapper-orm'
 			]
+		case 'lucene9.8':
+			return [
+					updateProperties: ['version.org.apache.lucene'],
+					onlyRunTestDependingOn: ['hibernate-search-backend-lucene'],
+					requiresSourceModifications: true
```
yrodiere, line 25: Does it though? :thinking:

Same for the other "lucene" jobs, by the way...

yrodiere (approved):
Thanks. LGTM, though there's something suspicious in the Jenkinsfile, I'll let you decide if it's correct or not...

ci/dependency-update/Jenkinsfile
marko-bekhta: `requiresSourceModifications` ? yeah.... maybe the name is not the best to describe what it actually does 😃 
this is to address the elasticsearch "upgrade" since there no version are modified in POM or any other file the build will fail with:
```
error "This job does not seem to update any dependency; perhaps it is misconfigured? The source code has not been updated, neither by merging a WIP branch nor by updating version properties."
```
and here we expect that the pom should be updated ...

yrodiere: Ah, ok. Then maybe `skipSourceModifiedCheck` would make more sense, and you'd set it to `true` for Elasticsearch-latest but omit it everywhere else.
But agree, that's unrelated to this PR.

PR #3762, opened 2023-10-02
[Hibernate Search 6.2] HSEARCH-4944 Recompile util module to make the ORM6.2 upgrade work correctly

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4944

I suppose with the next ORM upgrade, we can remove these additional build args as the changes will be in the main artifact already ...


yrodiere:
@marko-bekhta Doesn't work: https://ci.hibernate.org/view/Search/job/hibernate-search-dependency-update/job/6.2/43/console

PR #3766, opened 2023-10-02
HSEARCH-4984 Add test deserializing payload from older Avro versions

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4984


mapper/orm-outbox-polling/src/test/java/org/hibernate/search/mapper/orm/outboxpolling/avro/impl/EventPayloadSerializationUtilsTest.java
```diff
@@ -0,0 +1,90 @@
...
+		Files.createDirectories( testDataDirectory );
+		try ( OutputStream out = Files.newOutputStream(
+				testDataDirectory.resolve( "payload-avro-" + Protocol.class.getPackage().getImplementationVersion() ) ) ) {
+			out.write( serialized );
+		}
+	}
+
+	@Test
+	public void canDeserializePreviousPayloads() throws IOException {
+		try ( Stream<Path> payloads = Files.list( payloadTestResourceLocation() ) ) {
```
yrodiere, line 61: A fine candidate for a `@MethodSource` after the JUnit 5 upgrade  :D

yrodiere (approved):
I think you can drop the test payloads for Avro versions < 1.10 since we've never released a version of Search that would use those versions to serialize these payloads.

Otherwise this looks good, thanks :)

yrodiere (changes requested):
This broke the Windows build :D

marko-bekhta:
🪟🪟🪟 🥲

yrodiere (approved):

PR #3772, opened 2023-10-03
Add asm to the build dependency group

marko-bekhta:


marko-bekhta:
merging to update dependabot PRs

PR #3774, opened 2023-10-03
HSEARCH-4487 Follow-up: Jakarta Batch Mass Indexing job fixes

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4487

Follows up on #3754 and should fix bugs with some DBs (incorrect test with MariaDB, actual bugs with MSSQL/DB2).


yrodiere:
I just triggered a build to test against all databases: https://ci.hibernate.org/job/hibernate-search/job/PR-3774/2/

marko-bekhta (approved):
🤞🏻

yrodiere:
Ok, so this doesn't fix it, and from what I can see the bug is unrelated to HSEARCH-4487 anyway: it's just that we added tests for more DBs and thus we're just now noticing the problem, but the problem has been there for a long time.

I'll open a Jira issue and submit a new PR.

PR #3776, opened 2023-10-05
HSEARCH-4987 Upgrade -orm6 artifacts to Hibernate ORM 6.2.11.Final

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4987

We may need to create a new release in JIRA 😃🙈


yrodiere (approved):
Thanks, LGTM.

PR #3777, opened 2023-10-06
HSEARCH-3654 DO NOT MERGE 

marko-bekhta:

This is to get a test count from CI. The main build has more tests as all of the non-default envs are contributing to the total. 😃


yrodiere:
Damn, too slow :(

marko-bekhta:
😄

PR #3780, opened 2023-10-12
HSEARCH-4987 Upgrade -orm6 artifacts to Hibernate ORM 6.2.13.Final

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4987

Maybe we should somehow automate this as now there are going to be more ORM releases 🙈 😃


yrodiere:
> Maybe we should somehow automate this as now there are going to be more ORM releases 🙈 😃

Right... we could let dependabot handle this, though there's the problem of assigning a Jira issue to each PR...

yrodiere (approved):

PR #3783, opened 2023-10-12
HSEARCH-4989 Use a different property to skip JQAssistant analysis of some modules

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4989

The problem was coming from how the plugin decides if it is the last module in the execution run ... it expects to encounter all the modules, even those that skip plugin execution... but it still wants to see them as "executed" which never happens 😃 ...
using this other property `jqassistant.maven.module.skip` allows skipping the analysis of the module, which makes the plugin happy and it manages to correctly figure out when to run the analysis..

I have also looked into the rule for nested static class ... but the info is not there. from what it seems ASM does not add a static modifier to the descriptor and that field is never set to true in the generated DB, there's also no other info that can be used to cover that sooo I've just added another check into the filter in the JUnit upgrade PR...

### DIFF
--- a/build/config/pom.xml
+++ b/build/config/pom.xml
@@ -19,7 +19,7 @@
     <description>Hibernate Search common build configuration files</description>
 
     <properties>
-        <jqassistant.skip>true</jqassistant.skip>
+        <jqassistant.maven.module.skip>true</jqassistant.maven.module.skip>
 
         <!--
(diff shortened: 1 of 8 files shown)


marko-bekhta:
hmm Oracle container timed out  this time 😔, which looks unrelated, but it's a few times now that containers are timing out (Postgres, Elasticsearch, now Oracle...)

PR #3785, opened 2023-10-16
HSEARCH-4988 Container extractor resolution leads to StackOverflowError on properties of type Path

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4988

Best reviewed commit by commit, especially because there's one refactoring in the middle to allow for better testing.

EDIT: Considering the extent of the changes, I wouldn't backport unless we really need to.

### DIFF
--- a/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/extractor/impl/ContainerExtractorBinder.java
+++ b/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/extractor/impl/ContainerExtractorBinder.java
@@ -10,9 +10,11 @@ import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 
 import org.hibernate.search.engine.environment.bean.BeanHolder;
 import org.hibernate.search.engine.environment.bean.BeanResolver;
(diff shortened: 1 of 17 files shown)


yrodiere:
Rebased to use JUnit 5...

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/extractor/impl/ContainerExtractorBinder.java
```diff
@@ -282,15 +287,58 @@ void addCandidate(ExtractorContributor contributor) {
...
+					}
 					// Recurse as much as possible
 					tryAppend( state );
 					return true;
 				}
 			}
 			return false;
 		}
+
+		public boolean tryMatch(ExtractorResolutionState<?> state, List<String> toMatch) {
```
marko-bekhta, line 310: just checking if wether you were planning to pull this into `ExtractorContributor` since the `SingleExtractorContributor` class is private so won't make much difference...

marko-bekhta (approved):
👍🏻 🎉 🎸

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/extractor/impl/ContainerExtractorBinder.java
yrodiere: Well at the moment I don't have a use case for this method on anything other than the default contributor (`FirstMachingSomethingContributor`), so no, I didn't plan to.

Maybe when there's a reason to :shrug:

yrodiere:
Thanks, merging

PR #3793, opened 2023-10-18
Address build errors in Jenkinsfile for update job

marko-bekhta:
https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search-personal-marko/detail/build%2Fupdate-update-job/22/pipeline/11

that's a 6.2 run with the backports from Christian's repository.
I've reverted the configuration map, since CI was complaining that it cannot find the variable, then I've moved the ORM build to the same stage as pre-building Search so that all necessary jars are stashed. 
Dependency update worked looking at the logs:
```

-        <version.org.hibernate.orm>6.2.13.Final</version.org.hibernate.orm>

+        <version.org.hibernate.orm>6.2.14-SNAPSHOT</version.org.hibernate.orm>
```

I also had to add the dir cleanup, as otherwise, cloning of ORM was failing. 

Sorry for the formatting; it's because of an additional `stage('Build') {` that has a condition 🙈 .

Also cloning ORM with an ssh link fails, while using an HTTPS one works ok (I've added it to the comment on the job parameter, but just wanted to mention it here as well 😃)

### DIFF
--- a/ci/dependency-update/Jenkinsfile
+++ b/ci/dependency-update/Jenkinsfile
@@ -7,44 +7,46 @@
 
 @Library('hibernate-jenkins-pipeline-helpers@1.5') _
 
-def configurations = [
-		'jandex3'          :
-				[
-						onlyRunTestDependingOn: ['hibernate-search-mapper-pojo-base'],
-						// Need to rebuild this module in order to insert the Jandex 3 dependency in tests
-						additionalMavenArgs   : '-pl :hibernate-search-util-internal-test-common'
-				],
-		'orm5.6'           :
-				[
-						updateProperties      : ['version.org.hibernate'],
-						onlyRunTestDependingOn: ['hibernate-search-mapper-orm', 'hibernate-search-mapper-orm-jakarta']
-				],
-		'orm6.2'           :
-				[
-						updateProperties      : ['version.org.hibernate.orm'],
-						onlyRunTestDependingOn: ['hibernate-search-mapper-orm-orm6'],
-						// we need to recompile this module since it has incompatible return type and will result in a build error.
-						// Note: we *must* rebuild the `-orm6` artifact (not the regular one), since that's where the orm6 upgrade is applied:
-						additionalMavenArgs   : '-pl :hibernate-search-util-internal-integrationtest-mapper-orm-orm6',
-						skipSourceModifiedCheck: true
-				],
-		'orm6.3'           :
-				[
-						updateProperties      : ['version.org.hibernate.orm'],
-						onlyRunTestDependingOn: ['hibernate-search-mapper-orm-orm6'],
-						// we need to recompile this module since it has incompatible return type and will result in a build error.
-						// Note: we *must* rebuild the `-orm6` artifact (not the regular one), since that's where the orm6 upgrade is applied:
-						additionalMavenArgs   : '-pl :hibernate-search-util-internal-integrationtest-mapper-orm-orm6'
-				],
-		'orm6-in-main-code':
-				[
-						updateProperties: ['version.org.hibernate.orm']
-						// In this case we'll rebuild and test everything.
-				]
-]
-
+// NOTE: Remember to update the matrix axes below when adding/removing entries here.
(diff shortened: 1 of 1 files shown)


ci/dependency-update/Jenkinsfile
```diff
@@ -113,18 +117,45 @@ pipeline {
...
+						// include the time needed to provision a node.
+						timeout(time: 30, unit: 'MINUTES') {
+							withMavenWorkspace {
+								sh """ \
 							mvn clean install -U -Pdist -DskipTests \
 						"""
-						dir(env.WORKSPACE_TMP + '/.m2repository') {
-							stash name: 'original-build-result', includes:"org/hibernate/search/**"
+								dir(env.WORKSPACE_TMP + '/.m2repository') {
+									stash name: 'original-build-result', includes: "org/hibernate/**"
```
yrodiere, line 156: From what I understand the Hibernate Search build will not use the Hibernate ORM artifacts we just built, which is good. But it will download another version of ORM, and this version will end up in the stash, too...

Do you think it would be possible to just have both builds happen on different nodes, creating strictly separate stashes with strictly separate include patterns?

If I dared, I'd mention you could even have the builds run in parallel :D

ci/dependency-update/Jenkinsfile
```diff
@@ -143,68 +174,61 @@ pipeline {
...
-						}
-					}
-					stage('Build Hibernate ORM') {
+					stage('Build') {
 						when {
 							expression {
-								return params.ORM_REPOSITORY?.trim() || params.ORM_BRANCH?.trim()
+								return params.UPDATE_JOB?.trim() == 'all' || params.UPDATE_JOB?.trim() == env.DEPENDENCY_UPDATE_NAME
 							}
 						}
```
yrodiere, line 188: It seems this expression is evaluated after allocating a node to the stage, which is a waste...

Adding `beforeAgent true` should fix it; see https://www.jenkins.io/doc/book/pipeline/syntax/#evaluating-when-before-entering-agent-in-a-stage

ci/dependency-update/Jenkinsfile
```diff
@@ -113,18 +117,45 @@ pipeline {
...
-				timeout(time: 30, unit: 'MINUTES') {
-					withMavenWorkspace {
-						sh """ \
+			stages {
+				stage('Build Hibernate ORM') {
+					when {
+						expression {
+							return params.ORM_REPOSITORY?.trim() || params.ORM_BRANCH?.trim()
+						}
+					}
```
yrodiere, line 131: It seems this expression is evaluated after allocating a node to the stage, which will be a waste if you split this stage from the "Build Hibernate Search" stage as I recommended in another comment.

Adding `beforeAgent true` should fix it; see https://www.jenkins.io/doc/book/pipeline/syntax/#evaluating-when-before-entering-agent-in-a-stage

yrodiere (approved):
Thanks! LGTM. I added a few comments, feel free to dismiss or postpone to another PR what you don't consider important.

> I've reverted the configuration map, since CI was complaining that it cannot find the variable

You probably needed to annotate it with `@Field` or use `this.myvariable = ...` instead of `myvariable = ...`. See the main Jenkinsfile. Note I'm not sure this works for declarative Jenkinsfiles, but feel free to try if you think it's worth it.

> Also cloning ORM with an ssh link fails, while using an HTTPS one works ok (I've added it to the comment on the job parameter, but just wanted to mention it here as well 😃)

Right, I think cloning SSH is technically possible but requires additional configuration in our job, and I don't remember which. Anyway, not worth our time.

marko-bekhta:
> You probably needed to annotate it with @Field or use this.myvariable = ... instead of myvariable = .... See the main Jenkinsfile. Note I'm not sure this works for declarative Jenkinsfiles, but feel free to try if you think it's worth it.

The only reason I wanted to do that in the first place was so that we can reuse the map keys in axis and job parameters, but since that...  didn't worked out 😄 

I'll close the PR and try to build Search/ORM in parallel, and once that works I'll reopen it 😃 so that there's less emails 🙈

PR #3796, opened 2023-10-19
HSEARCH-4674 Use test containers

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4674

the testcontainer+dockerfile+dependabot combination seems  to work 🙈 😃 
Here're a few PRs opened for DB updates:
https://github.com/marko-bekhta/hibernate-search/pull/190
https://github.com/marko-bekhta/hibernate-search/pull/188

while I was testing it on my repo I've added:
```
    username: ${{secrets.MY_DOCKERHUB_USER}}
    password: ${{secrets.MY_DOCKERHUB_TOKEN}}
```
we probably would need to add some secrets (as I'm not sure it works without the registry)

There's still some cleanup needed, but wanted to see if this still would work on CI


marko-bekhta:
Added a test containers BOM and a couple of comments. aaand converting it from the draft 😃

util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/OrmSetupHelper.java
```diff
@@ -46,14 +46,20 @@ public class OrmSetupHelper
 	private static final Map<String, Object> DEFAULT_PROPERTIES;
 
 	static {
+		Map<String, Object> defaults = new LinkedHashMap<>();
+
+		String jdbcUrl = System.getProperty( "hibernate.connection.url" );
+		if ( jdbcUrl == null || jdbcUrl.trim().isEmpty() ) {
+			DatabaseContainer.configuration();
+		}
```
yrodiere, line 54: This feels hackish, no?

yrodiere (commented):
Thanks, here are a few comments.

On the `static` containers... it's not a hard no, I just feel uncomfortable with something that is so widely frowned upon. If this is standard practice when using test containers, then so be it. But I'd really have felt safer with a solution leveraging JUnit contexts where possible.

util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/DatabaseContainer.java
```diff
@@ -0,0 +1,486 @@
...
+		DATABASE = SupportedDatabase.from( name );
+
+		DATABASE_CONTAINER = DATABASE.container(
+				root.resolve( "build" ).resolve( "container" ).resolve( name + ".Dockerfile" ),
+				name
+		);
+	}
+
+	public static Configuration configuration() {
+		return configure( Configuration::addAsSystemProperties );
```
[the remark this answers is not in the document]
marko-bekhta: > I'm mainly worried about race conditions

+1 that was one of the issues, and it's also something that is bothering me 😄  ... One of the tests `EntityManagerFactoryRetrievalIT` that is not using helpers (maybe there's more I don't remember) is a "problem". I wasn't sure back then if we could switch to the  `createEntityManagerFactory` with properties map version (as I didn't want to change the tests themselves 😃) but I suppose it's reasonable to do that change 😄 

Similar in Spring tests.. I think that I've taken it from some testcontainers guide to pass the port in such way... BUT looking closer, there's a https://docs.spring.io/spring-framework/reference/testing/annotations/integration-spring/annotation-dynamicpropertysource.html that we probably can leverage instead... so will do that 😃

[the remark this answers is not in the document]
yrodiere: > One of the tests EntityManagerFactoryRetrievalIT that is not using helpers (maybe there's more I don't remember) is a "problem".

Then I'd suggest restricting this hack to those tests that actually need it, and perhaps in a safer way (e.g. with `org.hibernate.search.util.impl.test.SystemHelper#setSystemProperty`, like we do in `SystemConfigurationPropertySourceTest`)

util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/DatabaseContainer.java
```diff
@@ -0,0 +1,486 @@
...
+
+	private static final Duration REGULAR_TIMEOUT = Duration.ofMinutes( 5 );
+	private static final Duration EXTENDED_TIMEOUT = Duration.ofMinutes( 10 );
+
+	private DatabaseContainer() {
+	}
+
+	private static final Object LOCK = new Object();
+	private static final SupportedDatabase DATABASE;
+	private static final JdbcDatabaseContainer<?> DATABASE_CONTAINER;
```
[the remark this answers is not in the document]
yrodiere: Fair enough... if that's the recommended practice then we can go with this.

I'm pretty sure JUnit 5 allows storing context that spans multiple tests, but maybe it's not worth it here.

build/parents/integrationtest/pom.xml
```diff
@@ -72,348 +58,6 @@
                         </execution>
                     </executions>
                 </plugin>
-                <plugin>
-                    <groupId>io.fabric8</groupId>
-                    <artifactId>docker-maven-plugin</artifactId>
-                    <version>${version.docker.maven.plugin}</version>
-                    <configuration>
-                        <skip>${test.containers.run.skip}</skip>
```
yrodiere, line 80: I think you need to remove this property and all references to it (in GitHub Actions workflows, in CONTRIBUTING.md, ...)

yrodiere (commented):
Some notes about properties mentioned in CONTRIBUTING.md.

.github/dependabot.yml
```diff
@@ -4,6 +4,13 @@
 # https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
 
 version: 2
+registries:
+  dockerhub:
+    type: docker-registry
+    url: https://registry.hub.docker.com
+    username: ${{secrets.MY_DOCKERHUB_USER}}
+    password: ${{secrets.MY_DOCKERHUB_TOKEN}}
+    replaces-base: true
```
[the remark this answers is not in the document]
marko-bekhta: https://github.com/marko-bekhta/hibernate-search/runs/17991619122 :

```
Your .github/dependabot.yml contained invalid details
Dependabot encountered the following error when parsing your .github/dependabot.yml:

The property '#/registries/dockerhub/' did not contain a required property of 'username'
The property '#/registries/dockerhub/' did not contain a required property of 'password'
Please update the config file to conform with Dependabot's specification.
```
^ if we add a registry, it wants to see the username/password pair configured (BTW I've generated a read-only token on dockerhub, and that's what I've used for a password. 

Now do we actually need to provide a registry? I'm not sure. I don't see it anywhere here (https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#docker) written in bold that it's a *MUST* 😄 but then it didn't open an upgrade PR for mariadb for ~20 min and once I've added the registry back the PR was opened in under 5 min .... so IDK 🤷🏻‍♂️ I can leave one on the DB files using some older DB version and we can see how dependabot will behave on the main repo without this config ?

and if we are going forward with adding the registry ... maybe you have better suggestions for the variable name prefixes? 😄 maybe `HIBERNATE_SEARCH_DOCKERHUB_` ?

util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/SearchContainer.java
```diff
@@ -0,0 +1,147 @@
...
+				.withEnv( "cluster.routing.allocation.disk.threshold_enabled", "false" );
+	}
+
+	/*
+	 * Suppress "Resource leak: '<unassigned Closeable value>' is never closed". Testcontainers take care of closing
+	 * this resource in the end.
+	 */
+	@SuppressWarnings("resource")
+	private static GenericContainer<?> common(String image, String tag) {
+		return new GenericContainer<>( DockerImageName.parse( image ).withTag( tag ) )
```
yrodiere, line 141: This means we completely ignore the elasticsearch/opensearch Dockerfiles, doesn't it?

I mean, I suppose that's fair, as we will most likely update the version manually anyway, but in that case the Dockerfiles should probably be removed.

It would be nice to use the dockerfile at least for opensearch, at least by default when passing `-Dtest.elasticsearch.distribution=opensearch` or `-Dtest.elasticsearch.distribution=amazon-opensearch-service` and not passing a specific version/tag. But I guess this would make configuring the version in Hibernate Search (`hibernate.search.backend.version`) quite complex, so it's probably better not to do that.

build/parents/integrationtest/pom.xml
```diff
@@ -479,21 +120,7 @@
...
-        <profile>
-            <id>elasticsearch-test-skip</id>
-            <activation>
-                <!-- Activate if test.elasticsearch.connection.uris has been defined explicitly -->
-                <property>
-                    <name>test.elasticsearch.skip</name>
-                    <value>true</value>
-                </property>
-            </activation>
-            <properties>
```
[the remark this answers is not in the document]
marko-bekhta: This looks like a bad merge... `elasticsearch-do-not-run` profile above should go away... thanks!

.github/dependabot.yml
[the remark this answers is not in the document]
yrodiere: Ok, let's go with credentials then.

Please use `DOCKERHUB_USER` + `DOCKERHUB_TOKEN`. I'll add the credentials to the org.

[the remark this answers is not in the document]
yrodiere: Actually they already exist. Please use `DOCKERHUB_USERNAME` (not `USER`) + `DOCKERHUB_TOKEN`

pom.xml
```diff
@@ -388,43 +387,14 @@
...
-        <!-- CockroachDB -->
-        <!-- See https://hub.docker.com/r/cockroachdb/cockroach/tags -->
-        <test.database.run.cockroachdb.skip>true</test.database.run.cockroachdb.skip>
-        <test.database.run.cockroachdb.image.name>cockroachdb/cockroach</test.database.run.cockroachdb.image.name>
-        <test.database.run.cockroachdb.image.tag>v23.1.4</test.database.run.cockroachdb.image.tag>
+        <!-- Docker images to be pulled. To be redefined in specific profiles of Elasticsearch/OpenSearch \
+             or database specific ones. -->
+        <test.database.run.kind>h2</test.database.run.kind>
+
+        <test.searchengine.run.image.pull>true</test.searchengine.run.image.pull>
```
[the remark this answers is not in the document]
marko-bekhta: we don't need it anymore 🙈 
both the groovy script and the container java-side are looking at the `test.database.run.kind` if they should pull/start a container. if it's not an H2 then it means `test.database.run.image.pull==true`..

util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/DatabaseContainer.java
```diff
@@ -0,0 +1,486 @@
...
+					container.getUsername(),
+					container.getPassword(),
+					""
+			);
+		}
+
+		abstract String dialect();
+
+		abstract <SELF extends JdbcDatabaseContainer<SELF>> JdbcDatabaseContainer<SELF> container(Path dockerfile,
+				String name);
```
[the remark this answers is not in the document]
marko-bekhta: that's only to be able to use anonymous inner classes... I'll change it

pom.xml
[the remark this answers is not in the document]
yrodiere: Technically we could also be running (among others) a module that doesn't use Hibernate ORM with `-Pci-postgres`, in which case pulling is not necessary either. Though I'll admit it's far-fetched and probably doesn't matter in practice.

Similarly, we could be running Lucene tests only (e.g. with `-Dtest.elasticsearch.skip=true`), in which case we probably wouldn't want to pull the Elasticsearch/OpenSearch images... but I guess you preserved that behavior by setting `test.searchengine.run.image.pull=false` automatically?

marko-bekhta:
There're more changes to come ... I just wanted to test Ruyk on CI 🙈 😃.

util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/DatabaseContainer.java
```diff
@@ -0,0 +1,428 @@
...
+
+		public HibernateSearchJdbcDatabaseContainer(Path dockerfile, String name, String driverClassName, String jdbcUrlPattern,
+				int port, String username, String password, String testQueryString) {
+			this( dockerfile, name, driverClassName, jdbcUrlPattern, port, username, password, testQueryString, null );
+		}
+
+		public HibernateSearchJdbcDatabaseContainer(Path dockerfile, String name, String driverClassName, String jdbcUrlPattern,
+				int port, String username, String password, String testQueryString, WaitStrategy waitStrategy) {
+			// IMPORTANT: we do not want to delete the image on exit as then we cannot use container reuse.
+			// (these two options are somewhat mutually exclusive).
```
marko-bekhta, line 310: this is what was causing Oracle tests to fail. While at it I've added a custom wait condition; this way DB startup is cleaner for some of the containers.

build/container/oracle.Dockerfile
```diff
@@ -0,0 +1,3 @@
+# Oracle
+# See https://hub.docker.com/r/gvenzl/oracle-free/tags
+FROM docker.io/gvenzl/oracle-free:23.3-slim-faststart
```
marko-bekhta, line 3: Note that I've switched to `free` from `xe` as I've noticed that Quarkus and Reactive already switched 😃

integrationtest/v5migrationhelper/orm/src/test/java/org/hibernate/search/test/batchindexing/DatabaseMultitenancyTest.java
```diff
@@ -43,10 +43,6 @@
  * @author Sanne Grinovero
  * @since 5.2
  */
-@RequiresDialect(
```
marko-bekhta, line 46: Note that I've removed any `RequiresDialect` or `SkipForDialect` ... 
They do not work OK with testcontainers... the DialectContext in the ORM test utils wants to have connection properties in env and it even requires a Dialect variable too.... so it looks like we can just leverage the dialect from the configuration of the container and do not rely on this util.

yrodiere: Ok... does this mean we (well, more likely QE) can no longer reliably test Hibernate Search against non-container databases?

marko-bekhta: yeah ... 
there was this https://github.com/hibernate/hibernate-search/pull/3796#discussion_r1368828689 that was suppose to be something similar to ES config, if someone somehow provides a way to connect to the DB then we woulnd't start a container... but I figured you were saying that I should remove it 😄 

now if we want to be able to test against non-testcontainer DB... we can do that in the `DatabaseContainer#configuration()`. Just let me know which property we'd expect a user to pass to provide connection info (will `jakarta.persistence.jdbc.url` be ok?)

yrodiere: > but I figured you were saying that I should remove it 😄

I think my main problem was with getting the info from a system property...

> Just let me know which property we'd expect a user to pass to provide connection info (will jakarta.persistence.jdbc.url be ok?)

I'd expect people to do it more or less the same as today, so:

* For Elasticsearch/OpenSearch, I think it's documented in `CONTRIBUTING.md`
* For databases, I think currently you'd have to pass something like `-Pci-postgresql -Dtest.database.run.skip=true -Djdbc.url=...` and optionally (if using different settings than what Hibernate Search expects `-Djdbc.user=... -Djdbc.pass=...`.

yrodiere: > For databases, I think currently you'd have to pass something like `-Pci-postgresql -Dtest.database.run.skip=true -Djdbc.url=...` and optionally (if using different settings than what Hibernate Search expects `-Djdbc.user=... -Djdbc.pass=...`.

But feel free to align this on whatever you'll do for Elasticsearch/OpenSearch

marko-bekhta: ok 😃 
`test.database.run.skip=true` -- we wouldn't need this one anymore (same as it is for the Elasticsearch case)
`-Djdbc.url` -- will make this work 👍🏻 

In general, the idea is that the `*Container` checks whether the connection URL was provided, and only if not then it starts a container, otherwise -- it is using the provided URL.

pom.xml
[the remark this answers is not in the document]
yrodiere: > Similarly, we could be running Lucene tests only (e.g. with `-Dtest.elasticsearch.skip=true`), in which case we probably wouldn't want to pull the Elasticsearch/OpenSearch images... but I guess you preserved that behavior by setting `test.searchengine.run.image.pull=false` automatically?

Bump

pom.xml
```diff
@@ -378,53 +377,14 @@
         <!-- Container images for various integration tests -->
         <!-- The latest version of Elasticsearch tested against by default -->
         <version.org.elasticsearch.latest>8.10.4</version.org.elasticsearch.latest>
```
yrodiere, line 379: We'll need a reminder in the Dockerfile that this property needs to be updated as well...

yrodiere (commented):
LGTM overall,  here are my latest comments... 

Did you check how execution time compares to what we had previously with docker-maven-plugin?

build/parents/build/pom.xml
```diff
@@ -1182,10 +1188,24 @@
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-failsafe-plugin</artifactId>
                     <configuration>
+                        <environmentVariables>
+                            <!-- Just to be sure that we are pulling/using the Ryuk from docker.io -->
+                            <TESTCONTAINERS_RYUK_CONTAINER_IMAGE>docker.io/testcontainers/ryuk:0.5.1</TESTCONTAINERS_RYUK_CONTAINER_IMAGE>
```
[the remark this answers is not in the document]
marko-bekhta: good point! 😃

marko-bekhta:
the execution time is more or less the same compared to the maven docker plugin:

| Env | Test Containers | Maven Docker plugin |
|-------|-------|-------|
|Opensearch | 11m 38s | 12m 42s |
|Oracle | 15m 24s | 12m 40s |
|Postgres | 9m 53s | 9m 25s|
|Default build ITs | 25m 38s | 25m 16s|

I'll take a look at Oracle tests and why it took that much longer:
Oracle Testcontaienrs:
Hibernate Search ITs - ORM - Jakarta Batch ......... SUCCESS [04:26 min]
Oracle maven-docker-plugin:
Hibernate Search ITs - ORM - Jakarta Batch ......... SUCCESS [02:24 min]

But since previous builds show a time close to 3 min, maybe it's just the change from XE to FREE version of the Oracle container...

yrodiere:
> the execution time is more or less the same compared to the maven docker plugin:

Hm, that's underwhelming... I'd have expected that reusing the containers from one Maven project to the next would bring at least some improvement. Do we reuse containers from one Maven project to the next? I.e. [this](https://java.testcontainers.org/features/reuse/)?

I mean it's still worth it, be it just for easier running of tests in the IDE. But still.

> But since previous builds show a time close to 3 min, maybe it's just the change from XE to FREE version of the Oracle container...

Right... why did we do that again? I mean, I know we did it because ORM did, but why did ORM do it? @beikov any insight there?

marko-bekhta:
not 100% sure about ORM, but it is true for Reactive and Quarkus ( https://github.com/hibernate/hibernate-reactive/commit/0af4ebc9390d631c4e97032452344444e5455834 / https://github.com/quarkusio/quarkus/commit/546922cf13b4de2d84966550577c0f22ef27000c )

XE has versions `21c, 18c, 11g` available, while FREE is for `23c` ... I'm running the same batch tests against 21c to see if there's any difference 😃 

>Do we reuse containers from one Maven project to the next? I.e. [this](https://java.testcontainers.org/features/reuse/)?

Yes we do ... and I also did expect some improvements, but maybe the starting times were small anyway? 🤷🏻‍♂️ If I run a build locally, then all the containers are kept up even after the build is finished (which is what I'd expect from those reusable containers)

pom.xml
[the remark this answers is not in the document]
marko-bekhta: Yes, I think that's what you were asking about, right:
https://github.com/hibernate/hibernate-search/blob/0da07c6093a51bf479b27fd067efee642d94541c/build/parents/integrationtest/pom.xml#L83-L96

yrodiere:
> If I run a build locally, then all the containers are kept up even after the build is finished (which is what I'd expect from those reusable containers)
> and while at it .. I've double checked that the containers remained the same during the whole build 😃

Ok, nice, thanks for checking. Does Ryuk eventually kill these containers? If not, maybe this should be enabled on CI but opt-in for developers... I suspect people won't like useless containers staying up on their machine.

>  XE has versions 21c, 18c, 11g available, while FREE is for 23c ... I'm running the same batch tests against 21c to see if there's any difference 😃

Christian just told me "free" is the new name for container images, there won't be any more XE for the new versions. So, we have to use "free" moving forward.

He also mentioned this regarding performance:

> take a look at our setup script, we did quite a few optimizations with Loic

> btw before merging this one we probably want to run the entire build with all envs

+1 please do when you're done with other adjustments.

marko-bekhta:
>He also mentioned this regarding performance:

I've tried adding this stuff https://github.com/hibernate/hibernate-orm/blob/main/docker_db.sh#L558-L617 to the container, but that didn't help much (it even was slower in a few runs 😔)

marko-bekhta:
Env | Test Containers | Maven Docker plugin
-- | -- | --
Opensearch | 8m 7s | 12m 42s
Oracle | 11m 6s | 12m 40s
Postgres | 8m 36s | 9m 25s
Default build ITs | 25m 38s | 25m 16s

now that's looking a bit better 😃 I'll start a full build to see if other DB envs will work OK

Jenkinsfile
```diff
@@ -498,7 +498,7 @@ stage('Non-default environments') {
 				withMavenWorkspace {
 					// Re-run integration tests against the JARs produced by the default build,
 					// but using a different JDK to build and run the tests.
-					mavenNonDefaultBuild buildEnv, "-f integrationtest"
+					mavenNonDefaultBuild buildEnv, "-f integrationtest -Pci-build"
```
yrodiere, line 501: We'll need to enable this profile basically everywhere, no? If so you could just add it inside `mavenNonDefaultBuild`.

And you probably want to enable it in the default build as well?

yrodiere: Or, I guess, we could just set the testcontainers environment variable in Jenkinsfiles instead of using Maven.

marko-bekhta: we don't need it for compile-check sources (pre-build step) and for eclipse compiler case ... but then I guess it doesn't hurt to have it for them too 😃

build/container/elastic.Dockerfile
```diff
@@ -0,0 +1,8 @@
+# Elasticsearch
+# https://hub.docker.com/r/elastic/elasticsearch/tags
+#
+# IMPORTANT! When updating the version for Ryuk in this Dockerfile,
```
yrodiere, line 4:
```suggestion
# IMPORTANT! When updating the version of Elasticsearch in this Dockerfile,
```

build/container/amazon-opensearch-serverless.Dockerfile
```diff
@@ -0,0 +1,3 @@
+# OpenSearch
+# See https://hub.docker.com/r/opensearchproject/opensearch/tags
```
yrodiere, line 2:
```suggestion
# See https://hub.docker.com/r/opensearchproject/opensearch/tags
#
# IMPORTANT! When updating the version of OpenSearch in this Dockerfile,
# make sure to update `version.org.opensearch.latest` property in a POM file,
# and to update the version in opensearch.Dockerfile as well.
```

yrodiere: Note sure it's worth having two separate dockerfiles, BTW... things would be simpler if we just used the OpenSearch dockerfile for the amazon-opensearch-serverless distribution.

build/container/opensearch.Dockerfile
```diff
@@ -0,0 +1,3 @@
+# OpenSearch
+# See https://hub.docker.com/r/opensearchproject/opensearch/tags
```
yrodiere, line 2:
```suggestion
# See https://hub.docker.com/r/opensearchproject/opensearch/tags#
#
# IMPORTANT! When updating the version of OpenSearch in this Dockerfile,
# make sure to update `version.org.opensearch.latest` property in a POM file,
# and to update the version in amazon-opensearch-serverless.Dockerfile as well.
```

util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/SearchBackendContainer.java
```diff
@@ -0,0 +1,192 @@
...
+				// Disable disk-based shard allocation thresholds: on large, relatively full disks (>90% used),
+				// it will lead to index creation to get stuck waiting for other nodes to join the cluster,
+				// which will never happen since we only have one node.
+				// See https://www.elastic.co/guide/en/elasticsearch/reference/7.17/modules-cluster.html#disk-based-shard-allocation
+				.withEnv( "cluster.routing.allocation.disk.threshold_enabled", "false" );
+
+		ElasticsearchVersion version =
+				ElasticsearchVersion.of( ElasticsearchDistributionName.ELASTIC, dockerImageName.getVersionPart() );
+
+		if ( version.majorOptional().orElse( Integer.MIN_VALUE ) == 8 ) {
```
yrodiere, line 108:
```suggestion
		// Disable a few features that we don't use and that just slow up container startup.
		if ( version.majorOptional().orElse( Integer.MIN_VALUE ) == 8 ) {
```

util/internal/integrationtest/backend/elasticsearch/src/main/java/org/hibernate/search/util/impl/integrationtest/backend/elasticsearch/SearchBackendContainer.java
```diff
@@ -0,0 +1,192 @@
...
+						.withEnv( "slm.history_index_enabled", "false" );
+			}
+			if ( version.minor().orElse( Integer.MAX_VALUE ) > 8 ) {
+				container.withEnv( "stack.templates.enabled", "false" );
+			}
+			if ( version.minor().orElse( Integer.MAX_VALUE ) > 5 ) {
+				container.withEnv( "indices.lifecycle.history_index_enabled", "false" );
+			}
+			return container.withEnv( "xpack.ml.enabled", "false" )
+					.withEnv( "xpack.watcher.enabled", "false" );
```
yrodiere, line 135: We don't test against ES 7.9 or lower anymore:

```suggestion
			return container
					.withEnv( "indices.lifecycle.history_index_enabled", "false" )
					.withEnv( "stack.templates.enabled", "false" )
					.withEnv( "xpack.ml.enabled", "false" )
					.withEnv( "xpack.watcher.enabled", "false" );
```

yrodiere (approved):
Oh my, I just tried from the IDE: I just change the version in the Jenkinsfile, run the test and BOOM: a container gets started automatically with the right version. It's awesome.

I added a few comments below, could you please process them?

I also pushed a commit to your branch. The changes are very minor, but please take care to keep it when you rebase :)

Then once you're done rebasing on main (there's a conflict), and maybe after you've squashed a few commits if it makes sense, feel free to merge :)

build/container/amazon-opensearch-serverless.Dockerfile
marko-bekhta: yeah... I've only had this file to make the groovy script simple:
```groovy
def fromFile = parseImage( './build/container/${test.elasticsearch.distribution}.Dockerfile')
```
if we want to keep one file, then this would need to have some `if-else` logic; so maybe, let's keep the file for now, I suppose if the dependabot integration will work as we expect then we wouldn't need to update these files anyway. 🤞🏻😃

.github/dependabot.yml
```diff
@@ -99,3 +109,12 @@ updates:
...
       - dependency-name: "org.bsc.maven:maven-processor-plugin"
+  - package-ecosystem: "docker"
+    registries:
+      - dockerhub
+      - ibm
+    # For dependabot to find Docker files they all should be in the same directory.
+    # Dependabot is picking any files that has `dockerfile` in them.
+    directory: "/build/container/"
+    schedule:
+      interval: "monthly"
```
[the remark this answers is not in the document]
yrodiere: LGTM, please go ahead :+1:

yrodiere (approved):
LGTM, please go ahead and squash/fixup what's needed and merge!
Thanks for all this :)

.github/dependabot.yml
```diff
@@ -99,3 +109,50 @@ updates:
...
+      - ibm
+    # For dependabot to find Docker files they all should be in the same directory.
+    # Dependabot is picking any files that has `dockerfile` in them.
+    #
+    # This includes only Ryuk container, hence we want it to be checked/updated at the same time when our regular java dependencies are updated
+    directory: "/build/container/"
+    schedule:
+      interval: "weekly"
+      day: "tuesday"
+    assignees: ["yrodiere"]
```
yrodiere, line 124: I just need a group in order to more easily match the dependabot branches in CI job configuration:

```suggestion
    assignees: ["yrodiere"]
    groups:
      # This group combines all build containers dependencies.
      build-containers:
        patterns:
          # Include all:
          - "*"
```

PR #3805, opened 2023-10-26
HSEARCH-5002 Document dropped compatibility with older Elasticsearch/OpenSearch versions and Hibernate ORM 6.3 upgrade in migration guide

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-5002


marko-bekhta (approved):
🎉 😃

PR #3806, opened 2023-10-27
Remove update dependencies job throttling

marko-bekhta:
Here's an ORM job triggering the Search dependency update job and waiting for the results: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-orm-personal-marko/detail/ci%2Fadd-hibernate-search-tests-to-build-test/15/pipeline/

I've tried `git` `checkout` as well as `git clone -c 'core.sshCommand=ssh -v ...` for cloning a repo using an ssh URL, but with no luck... it looks like git wasn't able to find the known_hosts that would contain the keys we had on CI... Hence I've just adapted the job to "convert" the URL instead 🙈 🙈 🙈 
I'll backport to 6.2 once we merge the changes to main and then update the ORM PR so it can actually trigger the correct build 😃

### DIFF
--- a/ci/dependency-update/Jenkinsfile
+++ b/ci/dependency-update/Jenkinsfile
@@ -107,8 +107,7 @@ pipeline {
 	options {
 		buildDiscarder logRotator(daysToKeepStr: '10', numToKeepStr: '3')
 		disableConcurrentBuilds(abortPrevious: true)
-		// Run at most twice per week, to preserve CI workers for more urgent tasks.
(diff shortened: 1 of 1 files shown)


yrodiere (approved):

yrodiere:
LGTM, thanks

yrodiere:
>  I'll backport to 6.2 once we merge the changes to main and then update the ORM PR so it can actually trigger the correct build 😃

Please do :)

PR #3813, opened 2023-10-31
Ignore JIRA checks for commits targeting only build-related files

yrodiere:
See https://github.com/hibernate/hibernate-github-bot/pull/196

### DIFF
--- a/.github/hibernate-github-bot.yml
+++ b/.github/hibernate-github-bot.yml
@@ -5,3 +5,30 @@ jira:
       # See the `build-dependencies` group in the Dependabot's configuration file
     - user: dependabot[bot]
       titlePattern: "Bump the build-dependencies group with \\d+ updates?"
+  ignoreFiles:
+    # Git
+    - ".git*"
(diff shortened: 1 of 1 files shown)


.github/hibernate-github-bot.yml
```diff
@@ -4,4 +4,29 @@ jira:
...
+    - "copyright.txt"
+    - "README.md"
+    - "MAINTAINERS.md"
+    - "CONTRIBUTING.md"
+    # Misc. build files
+    - "build/config/"
+    - "build/enforcer/"
+    - "build/jqassistant/"
+    - "build/reports/"
+    - "build/script/"
```
marko-bekhta, line 32: There are also `build/configuration-properties-collector` and `build/container*`  that also fall under this category no?

marko-bekhta (approved):
🎉

.github/hibernate-github-bot.yml
yrodiere: Right, will do

PR #3814, opened 2023-11-03
HSEARCH-4948 Vector search using the Lucene backend

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4948

Hey Yoann,

I've put together a few things that would allow storing vector fields and thought that it would be a good time to check with you before moving forward with exploring the searching part...

I went with the property names for a vector field that are closer to the Lucene attributes rather than what Elasticsearch/OpenSearch are using (`m`/`ef_construction` do not sound like a nice option for method names 😆; hence it is  `maxConnections` and `beamWidth` instead).

With that said, note the `ScopedCodec` -- it turns out the only way to actually pass these two values is through a codec that is a part of the `IndexWriterConfig`.  I was hesitant about where to put `knnVectorFormat` but as it is somewhat codec-related I've put it there for now...

Let me know if you think we'd need something additional in the `VectorField` and what do you think about the naming of the public API side ... 😃


mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/VectorField.java
```diff
@@ -0,0 +1,91 @@
...
+	 */
+	int beamWidth() default 512;
+
+	/**
+	 * This seems like a great attribute name :D
+	 * The number of neighbors each node will be connected to in the HNSW graph. Defaults to 16 ES.
+	 * The number of bidirectional links that the plugin creates for each new element. Increasing and decreasing this value can have a large impact on memory consumption. Keep this value between 2 and 100.
+	 * @return
+	 */
+	int maxConnections() default 16;
```
yrodiere, line 81: Please use `-1` as the default value. See other comments in the same file.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/VectorField.java
```diff
@@ -0,0 +1,91 @@
...
+	 * @return How vector similarity is calculated.
+	 * @see VectorSimilarity
+	 */
+	VectorSimilarity vectorSimilarity() default VectorSimilarity.L2;
+
+	/**
+	 * Open Search has more parameters ...
+	 * https://opensearch.org/docs/latest/search-plugins/knn/knn-index/#hnsw-parameters-1
+	 */
+
```
yrodiere, line 62: Comment to be removed eventually.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/VectorField.java
```diff
@@ -0,0 +1,91 @@
...
+	 */
+	String name() default "";
+
+	/**
+	 * @return Whether this field should be projectable.
+	 * @see Searchable
+	 */
+	Projectable projectable() default Projectable.DEFAULT;
+
+	int dimension();
```
yrodiere, line 50: Missing javadoc.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/VectorField.java
```diff
@@ -0,0 +1,91 @@
...
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.processing.PropertyMapping;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.processing.PropertyMappingAnnotationProcessorRef;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.processing.impl.VectorFieldAnnotationProcessor;
+
+/**
+ * Maps an entity property to a vector field in the index.
+ * <p>
+ * This annotation will work for any {@code float/byte} array/list.
+ * <p>
+ * TODO: vector : docs
```
yrodiere, line 29: TODO indeed.

engine/src/main/java/org/hibernate/search/engine/backend/types/VectorSimilarity.java
```diff
@@ -0,0 +1,18 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.engine.backend.types;
+
+/**
+ * TODO: vector : docs
```
yrodiere, line 10: TODO indeed.

documentation/src/test/java17/org/hibernate/search/documentation/mapper/orm/projection/Author.java
```diff
@@ -29,10 +31,18 @@ public class Author {
 	@FullTextField(analyzer = "name", projectable = Projectable.YES)
 	private String lastName;
 
+	@VectorField(dimension = 4)
+	private byte[] embeddings;
+
+	@VectorField(dimension = 8)
+	private Float[] embeddingsFloat;
```
yrodiere, line 38: Not sure it makes sense to accept `Float[]`... I hope you didn't have to go out of your way to support it?
`List<Float>` would be more useful IMO.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/dsl/impl/LuceneFloatVectorFieldTypeOptionsStep.java
```diff
@@ -0,0 +1,71 @@
...
+	private static class FloatVectorCodec extends AbstractVectorFieldCodec<float[], float[]> {
+		public FloatVectorCodec(VectorSimilarity vectorSimilarity, Integer dimension, Storage storage,
+				float[] indexNullAsValue, KnnVectorsFormat knnVectorsFormat) {
+			super( vectorSimilarity, dimension, storage, indexNullAsValue, knnVectorsFormat );
+		}
+
+		@Override
+		public float[] decode(IndexableField field) {
+			KnnFloatVectorField byteVectorField = (KnnFloatVectorField) field;
+			return byteVectorField.vectorValue();
```
yrodiere, line 47:
```suggestion
			KnnFloatVectorField floatVectorField = (KnnFloatVectorField) field;
			return floatVectorField.vectorValue();
```

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/dsl/impl/AbstractLuceneVectorFieldTypeOptionsStep.java
```diff
@@ -0,0 +1,228 @@
...
+			case NO:
+				return false;
+			case YES:
+				return true;
+			default:
+				throw new AssertionFailure( "Unexpected value for Projectable: " + projectable );
+		}
+	}
+
+	protected abstract static class AbstractVectorFieldCodec<F, E> implements LuceneStandardFieldCodec<F, E> {
```
yrodiere, line 150: Can you please move this to its own file next to all other codecs?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/dsl/impl/LuceneByteVectorFieldTypeOptionsStep.java
```diff
@@ -0,0 +1,71 @@
...
+	private static class ByteVectorCodec extends AbstractVectorFieldCodec<byte[], byte[]> {
+		public ByteVectorCodec(VectorSimilarity vectorSimilarity, Integer dimension, Storage storage, byte[] indexNullAsValue,
+				KnnVectorsFormat knnVectorsFormat) {
+			super( vectorSimilarity, dimension, storage, indexNullAsValue, knnVectorsFormat );
+		}
+
+		@Override
+		public byte[] decode(IndexableField field) {
+			KnnByteVectorField byteVectorField = (KnnByteVectorField) field;
+			return byteVectorField.vectorValue();
```
yrodiere, line 47: Does this even work? I would be surprised, since the `toStoredField` method does not return a `KnnByteVectorField`.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/dsl/impl/LuceneByteVectorFieldTypeOptionsStep.java
```diff
@@ -0,0 +1,71 @@
...
+		return this;
+	}
+
+	@Override
+	protected AbstractVectorFieldCodec<byte[], ?> createCodec(VectorSimilarity vectorSimilarity, Integer dimension,
+			Storage storage, byte[] indexNullAsValue, KnnVectorsFormat knnVectorsFormat) {
+		return new ByteVectorCodec( vectorSimilarity, dimension, storage, indexNullAsValue, knnVectorsFormat );
+	}
+
+	private static class ByteVectorCodec extends AbstractVectorFieldCodec<byte[], byte[]> {
```
yrodiere, line 38: Can you please move this to a separate file next to all the other codecs?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/dsl/impl/LuceneFloatVectorFieldTypeOptionsStep.java
```diff
@@ -0,0 +1,71 @@
...
+		return this;
+	}
+
+	@Override
+	protected AbstractVectorFieldCodec<float[], ?> createCodec(VectorSimilarity vectorSimilarity, Integer dimension,
+			Storage storage, float[] indexNullAsValue, KnnVectorsFormat knnVectorsFormat) {
+		return new FloatVectorCodec( vectorSimilarity, dimension, storage, indexNullAsValue, knnVectorsFormat );
+	}
+
+	private static class FloatVectorCodec extends AbstractVectorFieldCodec<float[], float[]> {
```
yrodiere, line 38: Can you please move this to a separate file next to all the other codecs?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/dsl/impl/AbstractLuceneVectorFieldTypeOptionsStep.java
```diff
@@ -0,0 +1,228 @@
...
+			}
+
+		}
+
+		protected FieldType createType() {
+			FieldType type = new FieldType();
+			type.setVectorAttributes( dimension, vectorEncoding(), vectorSimilarityFunction() );
+			type.freeze();
+			return type;
+		}
```
yrodiere, line 218: It's probably not a good idea to re-create the type every time you create a field.

See how we do it for `LuceneStringFieldCodec#mainFieldType`.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/dsl/impl/AbstractLuceneVectorFieldTypeOptionsStep.java
```diff
@@ -0,0 +1,228 @@
...
+
+		protected abstract IndexableField toStoredField(String absoluteFieldPath, E encodedValue);
+
+		@Override
+		public boolean isCompatibleWith(LuceneFieldCodec<?> obj) {
+			if ( this == obj ) {
+				return true;
+			}
+			return getClass() == obj.getClass();
+		}
```
yrodiere, line 195: This seems suspiciously generous... the beam width, similarity, even dimensions do not matter and do not affect the `encode`/`decode` methods?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/dsl/impl/AbstractLuceneVectorFieldTypeOptionsStep.java
```diff
@@ -0,0 +1,228 @@
...
+			}
+
+			E encodedValue = encode( value );
+
+			documentBuilder.addField( createIndexField( absoluteFieldPath, value ) );
+			if ( Storage.ENABLED == storage ) {
+				documentBuilder.addField( toStoredField( absoluteFieldPath, encodedValue ) );
+			}
+			// For the "exists" predicate
+			documentBuilder.addFieldName( absoluteFieldPath );
```
yrodiere, line 184: This doesn't seem necessary as `org.apache.lucene.search.FieldExistsQuery` claims to support vector fields.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneFieldCodec.java
```diff
@@ -54,4 +55,8 @@ public interface LuceneFieldCodec<F> {
 	 * in doubt.
 	 */
 	boolean isCompatibleWith(LuceneFieldCodec<?> other);
+
+	default KnnVectorsFormat knnVectorFormat() {
+		return null;
+	}
```
yrodiere, line 61: Does this need to be this high in the hierarchy? It really doesn't seem relevant to anything but vector fields.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/logging/impl/Log.java
```diff
@@ -705,4 +705,13 @@ SearchException cannotHighlightInNestedContext(String currentNestingField,
...
 	SearchException cannotUseProjectionInNestedContext(String projection, String hint, @Param EventContext eventContext);
+
+	@Message(id = ID_OFFSET + 174,
+			value = "Vector '%1$s' cannot be equal to '%2$s'. It must be a positive integer value not greater than %3$s.")
+	SearchException vectorPropertyUnsupportedValue(String property, Integer value, int max);
+
+	@LogMessage(level = Level.WARN)
+	@Message(id = ID_OFFSET + 175,
+			value = "Vector field configuration property '%1$s' value '%2$s' is out of the recommended range [%3$d, %4$d]")
+	void vectorPropertyOutOfRecommendedRange(String property, Integer value, int min, int max);
```
yrodiere, line 716: Not sure such warning makes sense.

Either you think values out of the range will work fine and you don't warn, or you think these values will cause problems and then you just throw an exception.

If values out of the range may work, but have downsides, you should just make the accepted range configurable (through backend properties). Though this is definitely out of scope for this PR, so would rather be a follow-up Jira issue.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/ElasticsearchIndexFieldTypeFactoryImpl.java
```diff
@@ -234,6 +235,16 @@ public ScaledNumberIndexFieldTypeOptionsStep<?, BigInteger> asBigInteger() {
...
 
+	@Override
+	public VectorFieldTypeOptionsStep<?, byte[]> asByteVector() {
+		throw new UnsupportedOperationException( "Implement me" );
+	}
+
+	@Override
+	public VectorFieldTypeOptionsStep<?, float[]> asFloatVector() {
+		throw new UnsupportedOperationException( "Implement me" );
+	}
```
yrodiere, line 246: Come on, please think about the people who'll encounter the exception...

```suggestion
	@Override
	public VectorFieldTypeOptionsStep<?, byte[]> asByteVector() {
		throw new UnsupportedOperationException( "The Elasticsearch backend does not support vector field yet" );
	}

	@Override
	public VectorFieldTypeOptionsStep<?, float[]> asFloatVector() {
		throw new UnsupportedOperationException( "The Elasticsearch backend does not support vector field yet" );
	}
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/mapping/impl/BridgeResolver.java
```diff
@@ -325,6 +329,20 @@ private void addDefaults() {
...
+					.identifierBridge( DefaultBytePrimitiveArrayBridge.INSTANCE );
+			exactType( float[].class )
+					.valueBridge( DefaultFloatPrimitiveArrayBridge.INSTANCE )
+					.identifierBridge( DefaultFloatPrimitiveArrayBridge.INSTANCE );
+			exactType( Byte[].class )
+					.valueBridge( DefaultByteArrayBridge.INSTANCE )
+					.identifierBridge( DefaultByteArrayBridge.INSTANCE );
+			exactType( Float[].class )
+					.valueBridge( DefaultFloatArrayBridge.INSTANCE )
+					.identifierBridge( DefaultFloatArrayBridge.INSTANCE );
```
yrodiere, line 345: I'm not sure this makes a lot sense... But if you think it does, okay.

Note however that `List<Byte>` and `List<Float>` would make a lot more sense to support. Maybe create a follow-up Jira for that?

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/VectorField.java
```diff
@@ -0,0 +1,91 @@
...
+	/**
+	 * The number of candidates to track while assembling the list of nearest neighbors for each new node. Defaults to 100 in Elasticsearch.
+	 * The size of the dynamic list used during k-NN graph creation. Higher values lead to a more accurate graph but slower indexing speed. Defaults to 512 in Open Search.
+	 * @return
+	 */
+	int beamWidth() default 512;
+
+	/**
+	 * This seems like a great attribute name :D
+	 * The number of neighbors each node will be connected to in the HNSW graph. Defaults to 16 ES.
```
yrodiere, line 77:
```suggestion
	 * The number of neighbors each node will be connected to in the HNSW graph. Defaults to 16 in Elasticsearch.
```

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/VectorField.java
```diff
@@ -0,0 +1,91 @@
...
+
+	/**
+	 * The number of candidates to track while assembling the list of nearest neighbors for each new node. Defaults to 100 in Elasticsearch.
+	 * The size of the dynamic list used during k-NN graph creation. Higher values lead to a more accurate graph but slower indexing speed. Defaults to 512 in Open Search.
+	 * @return
+	 */
+	int beamWidth() default 512;
+
+	/**
+	 * This seems like a great attribute name :D
```
yrodiere, line 76: ?

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/VectorField.java
```diff
@@ -0,0 +1,91 @@
...
+
+	/**
+	 * @return A value used instead of null values when indexing.
+	 */
+	String indexNullAs() default AnnotationDefaultValues.DO_NOT_INDEX_NULL;
+
+	/**
+	 * The number of candidates to track while assembling the list of nearest neighbors for each new node. Defaults to 100 in Elasticsearch.
+	 * The size of the dynamic list used during k-NN graph creation. Higher values lead to a more accurate graph but slower indexing speed. Defaults to 512 in Open Search.
+	 * @return
```
yrodiere, line 71: You might as well put `@return` at the start of the javadoc, like in other attributes?

Same below.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/programmatic/impl/PropertyMappingVectorFieldStepImpl.java
```diff
@@ -0,0 +1,76 @@
...
+		implements PropertyMappingVectorFieldStep, PojoPropertyMetadataContributor {
+
+	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
+
+
+	PropertyMappingVectorFieldStepImpl(PropertyMappingStep parent, String relativeFieldName) {
+		super( parent, relativeFieldName, FieldModelContributorContext::vectorTypeOptionsStep );
+		if ( relativeFieldName != null && relativeFieldName.contains( FieldPaths.PATH_SEPARATOR_STRING ) ) {
+			throw log.invalidFieldNameDotNotAllowed( relativeFieldName );
+		}
```
yrodiere, line 32: I think this check is already present in the superclass constructor.

util/internal/integrationtest/common/src/main/java/org/hibernate/search/util/impl/integrationtest/common/stub/backend/types/dsl/impl/StubIndexFieldTypeFactory.java
```diff
@@ -157,6 +158,16 @@ public ScaledNumberIndexFieldTypeOptionsStep<?, BigInteger> asBigInteger() {
...
+	@Override
+	public VectorFieldTypeOptionsStep<?, byte[]> asByteVector() {
+		throw new UnsupportedOperationException( "Implement me" );
+	}
+
+	@Override
+	public VectorFieldTypeOptionsStep<?, float[]> asFloatVector() {
+		throw new UnsupportedOperationException( "Implement me" );
+	}
+
```
yrodiere, line 170: Implement me indeed.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/programmatic/PropertyMappingVectorFieldStep.java
```diff
@@ -0,0 +1,29 @@
...
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.mapper.pojo.mapping.definition.programmatic;
+
+import org.hibernate.search.engine.backend.types.Projectable;
+import org.hibernate.search.engine.backend.types.VectorSimilarity;
+
+/**
+ * TODO: vector : docs
```
yrodiere, line 13: TODO indeed.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/processing/impl/VectorFieldAnnotationProcessor.java
```diff
@@ -0,0 +1,38 @@
...
+public class VectorFieldAnnotationProcessor implements PropertyMappingAnnotationProcessor<VectorField> {
+
+	@Override
+	public final void process(PropertyMappingStep mappingContext, VectorField annotation,
+			PropertyMappingAnnotationProcessorContext context) {
+		String cleanedUpRelativeFieldName = context.toNullIfDefault( annotation.name(), "" );
+
+		PropertyMappingVectorFieldStep fieldContext = mappingContext.vectorField( cleanedUpRelativeFieldName )
+				.beamWidth( annotation.beamWidth() )
+				.maxConnections( annotation.maxConnections() )
```
yrodiere, line 26: You should probably handle defaults here.
See what we do with the decimal scale in `ScaledNumberFieldProcessor`.

yrodiere: And actually you did handle defaults for other fields below...

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/programmatic/PropertyMappingVectorFieldStep.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+public interface PropertyMappingVectorFieldStep extends PropertyMappingStep {
+
+	PropertyMappingVectorFieldStep projectable(Projectable projectable);
+
+	PropertyMappingVectorFieldStep vectorSimilarity(VectorSimilarity vectorSimilarity);
+
+	PropertyMappingVectorFieldStep dimension(Integer dimension);
```
yrodiere, line 21: If the dimension is mandatory (and I think it is?) you should probably pass it to `propertyMappingStep#vectorField` instead?

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/programmatic/PropertyMappingVectorFieldStep.java
```diff
@@ -0,0 +1,29 @@
...
+
+	PropertyMappingVectorFieldStep beamWidth(Integer beamWidth);
+
+	PropertyMappingVectorFieldStep maxConnections(Integer maxConnections);
```
yrodiere, line 25:
```suggestion
	PropertyMappingVectorFieldStep beamWidth(int beamWidth);

	PropertyMappingVectorFieldStep maxConnections(int maxConnections);
```

yrodiere (commented):
Thanks for this first pass. Here are a couple comments.

> I went with the property names for a vector field that are closer to the Lucene attributes rather than what Elasticsearch/OpenSearch are using (m/ef_construction do not sound like a nice option for method names 😆; hence it is maxConnections and beamWidth instead).

Agreed, `maxConnections` and `beamWidth` seem better... though it might be worth checking if `m`/`ef_construction` aren't standard names in the world of KNN search?

> With that said, note the ScopedCodec -- it turns out the only way to actually pass these two values is through a codec that is a part of the IndexWriterConfig. I was hesitant about where to put knnVectorFormat but as it is somewhat codec-related I've put it there for now...

I'm not sure where else you could put that? Makes sense to me to put it in the field type.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/document/model/impl/LuceneIndexModel.java
```diff
@@ -112,4 +120,19 @@ protected Analyzer getWrappedAnalyzer(String fieldName) {
 			return analyzer;
 		}
 	}
+
+	private class ScopedCodec extends Lucene95Codec {
```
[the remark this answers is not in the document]
marko-bekhta: I've tried to move things around (you'll see it in the next update 😉), but from what I see, at some point, we'll end up creating one of these "version-specific classes" 😔

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/vector/LuceneVectorFieldIT.java
```diff
@@ -0,0 +1,111 @@
...
+				.where( f -> f.matchAll() )
+				.toQuery();
+
+		List<Document> result = query.fetchAll().hits();
+		assertThat( result )
+				.hasSize( 2 )
+				.satisfies( containsDocument(
+						doc -> doc.hasField( "string", "keyword1" )
+								.hasField( "byteVector", Arrays.toString( BYTE_VECTOR_1 ) )
+								.hasField( "floatVector", Arrays.toString( FLOAT_VECTOR_1 ) )
```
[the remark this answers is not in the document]
yrodiere: Or -- hear me out -- don't allow projections on vector fields :D 

Any of the above is fine to me.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/VectorField.java
```diff
@@ -0,0 +1,91 @@
...
+	 */
+	int beamWidth() default 512;
```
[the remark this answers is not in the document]
marko-bekhta: these javadocs here were more of a notes on what's happening in other backends. but it does make sense to have it as `-1` and then let backends decide what they want to do about it 😃

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/ElasticsearchIndexFieldTypeFactoryImpl.java
marko-bekhta: This wasn't intended for the users 🙊  I was planning to move all of this vector-specific stuff under Lucene extensions, but for now, it would just make more noise, so these are just to make the compiler happy.

yrodiere: >  I was planning to move all of this vector-specific stuff under Lucene extensions

That would be nice, though I suspect you'll have a hard time doing that, as you need to call these methods from the POJO module, which cannot depend on the Lucene backend. And also, this effort would be a bit wasted because we'll most likely implement the same thing for Elasticsearch/OpenSearch in the near future.

IMO don't bother and just leave self-explanatory exceptions.

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/vector/LuceneVectorFieldIT.java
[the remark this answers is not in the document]
marko-bekhta: OK, I've tried to "serialize" floats as a byte array, and it doesn't look too terrible 😃, so maybe we can keep the projection. I think it might be helpful in some cases when the user wants to navigate the results going from a found item to a list of similar items.

documentation/src/test/java17/org/hibernate/search/documentation/mapper/orm/projection/Author.java
marko-bekhta: +1 I'm waiting for HSEARCH-5006 to become available to open it 😆 (btw there are a few more todos in the changes where we probably can open other followup tickets like making the max vector size configurable etc)

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/logging/impl/Log.java
marko-bekhta: yeah ... I've seen that there are these recommendations on the ranges for these values, so I thought it would be nice to show it to the user. And it is targeting the users who aren't sure about what to pick and are just "playing around" 😄. From what I understand, things might still work fine if the values are out of the range.... but maybe it would be ok to just move it to the docs, or Javadocs and remove the log message.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneFieldCodec.java
marko-bekhta: I've moved it to a "vector-specific" codec. My initial idea to put it there at the very top was to prevent from various `intsanceof` checks 😃. Ideally I'd put it inside of the field codec itself and pass the filter codec in, but I didn't see how to do that to make it work, hence this "getter" 😃

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/codec/impl/HibernateSearchLuceneCodec.java
```diff
@@ -0,0 +1,72 @@
...
+		this( new IndexModelBasedPerFieldKnnVectorsFormat( model ) );
+	}
+
+	public HibernateSearchLuceneCodec(KnnVectorsFormat knnVectorsFormat) {
+		super( HibernateSearchLuceneCodec.class.getSimpleName(), DEFAULT_CODEC );
+		this.knnVectorsFormat = knnVectorsFormat;
+	}
+
+	public HibernateSearchLuceneCodec() {
+		this( (KnnVectorsFormat) null );
```
marko-bekhta, line 37: Lucene is loading all these codecs through service loader and if we do not register one the whole thing fails to at the startup.. we also cannot use `Codec.getDefault()` as it can fall under race conditions depending on which codec gets loaded first...

> Please note: Don't call Codec.forName from the no-arg constructor of your own codec. When the SPI framework loads your own Codec as SPI component, SPI has not yet fully initialized! If you want to extend another Codec, instantiate it directly by calling its constructor.

that's why I've just used that constant for the codec and then added a test that will just fail if the `Codec.getDefault != Lucene95Codec` so that we'll notice if the thing will get updated...

Also I am not 100% convinced that this is a good approach, as this codec is still getting loaded and set in some lucene internals, but from some small experiments I've tried, it seemed to work ok...

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/predicate/impl/LuceneKnnPredicate.java
```diff
@@ -0,0 +1,119 @@
...
+import org.hibernate.search.backend.lucene.types.codec.impl.LuceneVectorFieldCodec;
+import org.hibernate.search.engine.search.predicate.SearchPredicate;
+import org.hibernate.search.engine.search.predicate.spi.KnnPredicateBuilder;
+import org.hibernate.search.util.common.logging.impl.LoggerFactory;
+
+import org.apache.lucene.search.KnnByteVectorQuery;
+import org.apache.lucene.search.KnnFloatVectorQuery;
+import org.apache.lucene.search.Query;
+
+public class LuceneKnnPredicate extends AbstractLuceneSingleFieldPredicate implements LuceneSearchPredicate {
```
marko-bekhta, line 25: I went back and forth on this one ... It seems like score and boost make no sense for this predicate, but then other things like nested checks might be useful? so I went with `AbstractLuceneSingleFieldPredicate` and added some unsuported operations in the builder 😄

documentation/src/test/java17/org/hibernate/search/documentation/mapper/orm/projection/Author.java
```diff
@@ -29,10 +31,18 @@ public class Author {
 	@FullTextField(analyzer = "name", projectable = Projectable.YES)
 	private String lastName;
 
+	@VectorField(dimension = 4)
```
marko-bekhta, line 34: please ignore this test 😃 I only have it to test an annotation with real db and backend at the same time, I'll have a proper test for the docs once I add a section for a new predicate 🙈

engine/src/main/java/org/hibernate/search/engine/backend/types/VectorSimilarity.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.engine.backend.types;
+
+/**
+ * Defines a function to calculate the vector similarity, i.e. distance between two vectors.
+ */
+public enum VectorSimilarity {
```
marko-bekhta, line 12: I'm wondering if we should be providing more information on these similarity functions... some of them require vectors to be normalized, and some wouldn't be able to handle a vector of `0` for example.... I'd suspect that people who will change it knows what they are doing and at the same time L2 norm looks like a safe default 😃

engine/src/main/java/org/hibernate/search/engine/backend/types/dsl/IndexFieldTypeFactory.java
```diff
@@ -40,6 +40,8 @@ public interface IndexFieldTypeFactory {
 	 */
 	<F> StandardIndexFieldTypeOptionsStep<?, F> as(Class<F> valueType);
 
+	<F> VectorFieldTypeOptionsStep<?, F> asVector(Class<F> valueType);
```
marko-bekhta, line 43: needs doc 👀

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/types/AbstractFieldTypeDescriptor.java
```diff
@@ -0,0 +1,88 @@
...
+
+	public final Class<F> getJavaType() {
+		return javaType;
+	}
+
+	public final String getUniqueName() {
+		return uniqueName;
+	}
+
+	public IndexFieldTypeOptionsStep<?, F> configure(IndexFieldTypeFactory fieldContext) {
```
marko-bekhta, line 48: I was trying to get vectors to fit this pattern for tests, but since vectors and other fields have a different base options step the most common one is this `IndexFieldTypeOptionsStep` and if I change the configure method to it then there's a looooot of tests that are getting affected and needs some casts 😔 So I'm not sure it is worth to fit the vectors into these regular fields tests...

yrodiere: I think you probably could get by with some generics? E.g. something like https://github.com/marko-bekhta/hibernate-search/pull/212

Note if you need to add a superinterface common to both `StandardIndexFieldTypeOptionsStep` and `VectorIndexFieldTypeOptionsStep`, that's fine by me. The only problem being the name...

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/AbstractLuceneVectorFieldCodec.java
```diff
@@ -0,0 +1,105 @@
...
+			return false;
+		}
+
+		AbstractLuceneVectorFieldCodec<?> other = (AbstractLuceneVectorFieldCodec<?>) obj;
+
+		return dimension == other.dimension
+				&& vectorSimilarity == other.vectorSimilarity
+				// not sure about this one :
+				// TODO : vector : need to test with different formats to see if that'll work ...
+				&& Objects.equals( knnVectorsFormat, other.knnVectorsFormat );
```
yrodiere, line 80: Maybe make it clear that the type of this field is actually `HibernateSearchKnnVectorsFormat`, so `equals` *is* implemented, unlike in the superclass `KnnVectorsFormat`?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneFieldCodec.java
yrodiere: Wouldn't it make more sense to retrieve the vector format directly from the field type, like we do for e.g. `org.hibernate.search.backend.lucene.types.impl.LuceneIndexValueFieldType#indexingAnalyzerOrNormalizer`?

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/programmatic/PropertyMappingStep.java
```diff
@@ -238,4 +238,8 @@ default PropertyMappingStep marker(MarkerBinder binder) {
 	 */
 	IndexingDependencyOptionsStep indexingDependency();
 
+	PropertyMappingVectorOptionsFieldStep vectorField(int dimension);
+
+	PropertyMappingVectorOptionsFieldStep vectorField(int dimension, String relativeFieldName);
+
```
yrodiere, line 244: Missing javadoc

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/predicate/impl/LuceneKnnPredicate.java
yrodiere: Makes sense.

As long as we don't expose unsupported options (boost, ...) through the API, I don't personally care :)

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/vector/LuceneVectorFieldIT.java
```diff
@@ -0,0 +1,464 @@
...
+				.satisfies( containsDocument(
+						doc -> doc.hasField( "string", "keyword2" )
+								.hasVectorField( "byteVector", BYTE_VECTOR_2 )
+								.hasVectorField( "floatVector", FLOAT_VECTOR_2 )
+								.andOnlyInternalFields()
+				) );
+	}
+
+	@Test
+	void simpleVectorSavedAndRetrievedViaProjection() {
```
yrodiere, line 79: Would be nice to rely on `FieldProjectionBaseIT` for that kind of things that are not specific to vector fields at all.

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/vector/LuceneVectorFieldIT.java
```diff
@@ -0,0 +1,464 @@
...
+		assertThat( result )
+				.hasSize( 2 )
+				.containsOnly(
+						new Object[] { "keyword1", BYTE_VECTOR_1, FLOAT_VECTOR_1 },
+						new Object[] { "keyword2", BYTE_VECTOR_2, FLOAT_VECTOR_2 }
+				);
+	}
+
+	@Test
+	void simpleVectorPredicateNoFilter() {
```
yrodiere, line 105: Eventually, could you try to move all tests related to the KNN predicate to a `KnnPredicateBaseIT`/`KnnPredicateSpecificsIT`? E.g. like `WildcardPredicateBaseIT`/`WildcardPredicateSpecificsIT`?

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/ExistsPredicateVectorIT.java
```diff
@@ -0,0 +1,91 @@
...
+import org.hibernate.search.util.impl.integrationtest.mapper.stub.BulkIndexer;
+import org.hibernate.search.util.impl.integrationtest.mapper.stub.SimpleMappedIndex;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.extension.RegisterExtension;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+class ExistsPredicateVectorIT {
```
yrodiere, line 31: I think you can remove this and just handle vector field types in ExistsPredicateBaseIT?

See https://github.com/marko-bekhta/hibernate-search/pull/212

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/KnnPredicateBaseIT.java
```diff
@@ -0,0 +1,108 @@
...
+import org.hibernate.search.integrationtest.backend.tck.testsupport.util.extension.SearchSetupHelper;
+import org.hibernate.search.util.impl.integrationtest.mapper.stub.BulkIndexer;
+import org.hibernate.search.util.impl.integrationtest.mapper.stub.SimpleMappedIndex;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+import org.junit.jupiter.params.provider.Arguments;
+
+class KnnPredicateBaseIT {
```
yrodiere, line 30: It would be nice to follow the same structure as e.g. `WildcardPredicateBaseIT`. That way you'd also tests things like multi-valued fields (if relevance), KNN predicate on unsupported types, etc.

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/vector/LuceneVectorFieldIT.java
```diff
@@ -0,0 +1,464 @@
...
+		setupHelper.start().withIndexes( index ).setup();
+		initDataSimpleNoFilter( index );
+
+		assertThatQuery(
+				index.createScope().query()
+						.select( f -> f.field( "byteVector" ) )
+						.where( f -> f.knn( 3 ).field( "byteVector" )
+								.matching( bytes( 4, (byte) 5 ) ) )
+		).hasTotalHitCount( 3 )
+				.hasHitsAnyOrder(
```
yrodiere, line 116: Why "any" order?

Wouldn't `bytes(4, 5)` be expected to be the first hit, since it's identical to the requested vector?

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/vector/LuceneVectorFieldIT.java
```diff
@@ -0,0 +1,464 @@
...
+												.from(
+														f.field( "nested.byteVector" ),
+														f.field( "nested.floatVector" )
+												)
+												.asList()
+												.multi()
+								).asList()
+						).where(
+								f -> f.nested( "nested" )
+										.add( f.knn( 1 ).field( "nested.byteVector" ).matching( bytes( 2, (byte) -120 ) ) )
```
yrodiere, line 234: I'm surprised this even works, considering it's not supported in Elasticsearch... does this also work with a multi-valued nested field? Because that's kind of the point of nested fields; single-valued nested fields are not useful.

If it does work, maybe it performs badly?

yrodiere (commented):
Thanks.

I didn't review everything, but at least I went through previous comments and the predicate API looks fine as well.

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/vector/LuceneVectorFieldIT.java
```diff
@@ -0,0 +1,464 @@
...
+						.field( "location" )
+						.matching( 5f, 4f )
+						.filter( f.range().field( "rating" ).between( 8, 10 ) )
+						.filter( f.terms().field( "parking" ).matchingAny( true ) )
+				).toQuery();
+
+		List<float[]> result = query.fetchAll().hits();
+
+		assertThat( result )
+				.hasSize( k ) // since that is how many neighbors we were asking for in the predicate
```
yrodiere, line 183: Ah, that's interesting. What if you do `f.or( f.knn(3)..., f.somethingThatMatchesMoreThan3Documents() )`? Will you get more than 3 hits, as expected?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/predicate/impl/LuceneKnnPredicate.java
```diff
@@ -0,0 +1,119 @@
...
+		}
+
+		@Override
+		public void filter(SearchPredicate filter) {
+			this.filter = LuceneSearchPredicate.from( scope, filter );
+		}
+
+		@Override
+		public void boost(float boost) {
+			throw new UnsupportedOperationException( "This predicate does not support boost" );
```
yrodiere, line 106: I wonder why you don't support boosts on KNN predicates... seems to me it's absolutely necessary that the KNN predicate returns some sort of score if we want to mix KNN with full-text search and use relevance sort... I think that was something that an OpenSearch hinted at: don't use KNN alone, but use it to improve the relevance of your full-text search?

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/vector/LuceneVectorFieldIT.java
marko-bekhta:
```java
f.or(
		f.knn( 3 )
				.field( "location" )
				.matching( 5f, 4f ),
		f.terms().field( "parking" ).matchingAny( true )
)
```
yields:
```
[[5.2f, 4.4f],
    [4.9f, 3.4f],
    [7.0f, 9.9f],
    [6.4f, 3.4f],
    [2.4f, 4.0f],
    [3.3f, 4.5f],
    [5.0f, 1.0f],
    [4.2f, 6.2f],
    [5.2f, 3.9f]]
```
soo it kinda works 😃

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/vector/LuceneVectorFieldIT.java
marko-bekhta: Yeah, honestly I didn't know what to expect there, and it was more of "let's try and see if it will not fail" 😄

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/vector/LuceneVectorFieldIT.java
marko-bekhta: Good point 😃

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/predicate/impl/LuceneKnnPredicate.java
marko-bekhta: I wasn't sure if the mix will work out and that only a knn predicate at the root will work ok. So I though I'll start without boosts and then see where it takes us 😃

yrodiere: Oh ok. Well starting small is perfectly fine. Just make sure you keep a list of the shortcuts you take... somewhere... so that we can add them to the backlog.

marko-bekhta:
oook, this should now build all of the modules and have some docs 😃 
I'll move it out of the draft and let me know if there's something missing for this PR 😃

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/codec/impl/HibernateSearchLuceneCodec.java
yrodiere: From what I understand, this loads the codec to read a segment that was written previously, and the name of the codec to load is included in the segment.

So, two things:

1. Yes, if it only uses the codec for the segment infos format, it probably doesn't need our codec wrapper.
2. We should make sure that the name of our own codec doesn't get written into the segment, because that would mean the segment would no longer include information about which "format" (Lucene version) it was written for. It would load our own codec, but delegating to the latest Lucene codec instead of the one it was written for... This just removes any chance of being able to read an index written with a previous version of Hibernate Search/Lucene.

I think #2 probably warrants a test, to be extra sure that the codec included in segment infos resolves to the Lucene one, not ours whose version is hard-coded...
We can access the codec from an `IndexReader` with `((StandardDirectoryReader) reader).getSegmentInfos().iterator().next().info`.
I don't know what would make more sense, a unit test in `backend/lucene/src/test/java/org/hibernate/search/backend/lucene/lowlevel/writer/impl` or an integration test in `integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/lowlevel/writer`. Whatever's simpler, I guess?

marko-bekhta: added something for this in the latest commit; let me know what do you think about it 😃

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/VectorBridge.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+package org.hibernate.search.mapper.pojo.bridge;
+
+import org.hibernate.search.util.common.annotation.Incubating;
+
+/**
+ * An extension of a regular {@link ValueBridge} for vector fields.
+ */
+@Incubating
+public interface VectorBridge<V, F> extends ValueBridge<V, F> {
```
yrodiere, line 15: This is a tagging interface... Not a fan :D 

But I'll see what it's used for...

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/binding/impl/ValueBindingContextImpl.java
```diff
@@ -188,7 +189,12 @@ private <F> IndexFieldTypeOptionsStep<?, F> inferFieldType(ValueBridge<?, F> bri
 				.orElseThrow( () -> new AssertionFailure( "Could not auto-detect the return type for value bridge '"
 						+ bridge + "'." ) );
 		if ( typeArgument instanceof Class ) {
-			return indexFieldTypeFactory.as( (Class<F>) typeArgument );
+			if ( bridge instanceof VectorBridge ) {
+				return indexFieldTypeFactory.asVector( (Class<F>) typeArgument );
+			}
```
yrodiere, line 194: If bridges want to use a specific field type, they should use a [value binder](https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#binding-valuebridge-valuebinder), so this shouldn't be necessary.

Why didn't you end up using the `FieldModelContributor` to change the default behavior depending on whether `@VectorField` or another annotation isused, as I suggested?

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/mapping/impl/BridgeResolver.java
```diff
@@ -325,6 +327,14 @@ private void addDefaults() {
...
 			exactType( GeoPoint.class )
 					.identifierBridge( DefaultGeoPointBridge.INSTANCE );
+
+			// arrays for vector fields:
+			exactType( byte[].class )
+					.valueBridge( DefaultByteArrayBridge.INSTANCE )
+					.identifierBridge( DefaultByteArrayBridge.INSTANCE );
+			exactType( float[].class )
+					.valueBridge( DefaultFloatArrayBridge.INSTANCE )
+					.identifierBridge( DefaultFloatArrayBridge.INSTANCE );
```
yrodiere, line 337: I don't really see the point of supporting identifiers of type `byte[]`/`float[]`, but... ok.

This makes me think, though: you'll probably need to add entries to `PropertyTypeDescriptor#all` to get these bridges tested.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/programmatic/PropertyMappingVectorOptionsFieldStep.java
```diff
@@ -0,0 +1,69 @@
...
+import org.hibernate.search.util.common.annotation.Incubating;
+
+/**
+ * The step in a property-to-index-field mapping where optional parameters can be set,
+ * when the index field is a vector field.
+ *
+ * @see VectorField
+ */
+@Incubating
+public interface PropertyMappingVectorOptionsFieldStep extends PropertyMappingStep {
```
yrodiere, line 22: I think you mean `PropertyMappingVectorFieldOptionsStep`?

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/programmatic/impl/PropertyMappingVectorOptionsFieldStepImpl.java
```diff
@@ -0,0 +1,66 @@
...
+
+import org.hibernate.search.engine.backend.types.Projectable;
+import org.hibernate.search.engine.backend.types.Searchable;
+import org.hibernate.search.engine.backend.types.VectorSimilarity;
+import org.hibernate.search.mapper.pojo.extractor.mapping.programmatic.ContainerExtractorPath;
+import org.hibernate.search.mapper.pojo.mapping.building.spi.PojoPropertyMetadataContributor;
+import org.hibernate.search.mapper.pojo.mapping.definition.programmatic.PropertyMappingStep;
+import org.hibernate.search.mapper.pojo.mapping.definition.programmatic.PropertyMappingVectorOptionsFieldStep;
+
+class PropertyMappingVectorOptionsFieldStepImpl
```
yrodiere, line 17: I think you mean `PropertyMappingVectorFieldOptionsStepImpl`?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/codec/impl/HibernateSearchLuceneCodec.java
yrodiere: Thanks. The approach looks good to me, I added a few comments.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/codec/impl/HibernateSearchKnnVectorsFormat.java
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	@Override
+	public KnnVectorsReader fieldsReader(SegmentReadState state) throws IOException {
+		return delegate.fieldsReader( state );
+	}
+
+	@Override
+	public int getMaxDimensions(String fieldName) {
+		// TODO : vector : we can make this configurable, apparently there are models that produce larger vectors than this default allows.
```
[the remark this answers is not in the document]
marko-bekhta: https://hibernate.atlassian.net/browse/HSEARCH-5020 😃

backend/lucene/src/main/resources/META-INF/services/org.apache.lucene.codecs.Codec
```diff
@@ -0,0 +1 @@
+org.hibernate.search.backend.lucene.lowlevel.codec.impl.HibernateSearchLuceneCodec
```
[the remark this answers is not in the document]
marko-bekhta: yeah, this should be addressed with a later commit now...

build/parents/build/pom.xml
```diff
@@ -30,6 +30,7 @@
 
         <!-- >>> Lucene -->
         <!-- When upgrading Lucene, make sure to align the dependency to com.carrotsearch.hppc (see below) -->
+        <!-- When upgrading Lucene, check HibernateSearchKnnVectorsFormat to see if its internals should be updated too. -->
```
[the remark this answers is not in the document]
marko-bekhta: yes 😃 
> this.delegate = new Lucene95HnswVectorsFormat( maxConn, beamWidth );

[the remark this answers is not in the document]
yrodiere: Ah, right... Maybe this should be tested as well? Like the version of the codec?

documentation/src/main/asciidoc/public/reference/_search-dsl-predicate.adoc
```diff
@@ -1464,6 +1464,53 @@ either on a per-field basis with a call to `.boost(...)` just after `.field(...)
...
+====
+
+[[search-dsl-predicate-knn-argument-type]]
+=== Expected type of arguments
+
+The `knn` predicate expects arguments to the `matching(...)` method
+to have the same type as the entity property corresponding to the target field.
+
+For example, if an entity property is of a byte array type (`byte[]`),
+`.matching(...)` will expect its argument to be a byte array (`byte[]`) only.
```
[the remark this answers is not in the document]
marko-bekhta: > If the property is of type List<Byte> and we set up a custom bridge, it will expect a List<Byte>, right? Not byte[]? And there's a test for that?

mmm not really the predicate will only take an array (`KnnPredicateVectorStep`)

documentation/src/test/java/org/hibernate/search/documentation/search/predicate/LucenePredicateDslIT.java
```diff
@@ -44,10 +49,56 @@ class LucenePredicateDslIT {
...
 	void setup() {
-		entityManagerFactory = setupHelper.start().setup( Book.class, Author.class, EmbeddableGeoPoint.class );
+		entityManagerFactory = setupHelper.start()
+				.withProperty(
+						HibernateOrmMapperSettings.MAPPING_CONFIGURER,
+						(HibernateOrmSearchMappingConfigurer) context -> context.programmaticMapping()
+								.type( Book.class )
+								.property( "coverImage" )
+								.vectorField( 128 )
+				)
```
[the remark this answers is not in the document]
marko-bekhta: Adding an annotation to the entity will lead to Elasticsearch suite failing all the tests at init stage, since it doesn't support the annotation so far... so the idea is to keep the same entities and map the vector only in Lucene-specific IT.

documentation/src/test/java/org/hibernate/search/documentation/search/predicate/LucenePredicateDslIT.java
```diff
@@ -44,10 +49,56 @@ class LucenePredicateDslIT {
...
 	@BeforeEach
 	void setup() {
-		entityManagerFactory = setupHelper.start().setup( Book.class, Author.class, EmbeddableGeoPoint.class );
+		entityManagerFactory = setupHelper.start()
+				.withProperty(
+						HibernateOrmMapperSettings.MAPPING_CONFIGURER,
+						(HibernateOrmSearchMappingConfigurer) context -> context.programmaticMapping()
+								.type( Book.class )
+								.property( "coverImage" )
+								.vectorField( 128 )
```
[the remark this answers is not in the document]
marko-bekhta: just a random thing 😃, I haven't researched any image processing models (for now), and  this isn't visible to the readers of the docs so should be ok to put any value.

documentation/src/main/asciidoc/public/reference/_search-dsl-predicate.adoc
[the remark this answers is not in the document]
yrodiere: Ok, then the description is incorrect. The expected type has nothing to do with the entity property type. You could have `byte[]` in the entity and use a `ValueBridge<byte[], float[]>`, and then the KNN predicate expects a `float[]`.

documentation/src/test/java/org/hibernate/search/documentation/search/predicate/LucenePredicateDslIT.java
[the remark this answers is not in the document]
yrodiere: Ok, makes sense... Could you move this all to PredicateDslIT and make the mapping + test conditional, so it's clearer?

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/ExistsPredicateSpecificsIT.java
```diff
@@ -38,12 +39,12 @@
 import org.junit.jupiter.params.provider.MethodSource;
 
 class ExistsPredicateSpecificsIT<F> {
-	private static final List<FieldTypeDescriptor<?>> supportedFieldTypes = FieldTypeDescriptor.getAll();
-	private static final List<FieldTypeDescriptor<?>> supportedFieldTypesWithDocValues = new ArrayList<>();
+	private static final List<StandardFieldTypeDescriptor<?>> supportedFieldTypes = FieldTypeDescriptor.getAllStandard();
```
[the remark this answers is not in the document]
marko-bekhta: the tests here needed a flattened field that results in a "multivalue" vector field and that just fails on test init. I'll make some changes...

engine/src/main/java/org/hibernate/search/engine/cfg/spi/ParseUtils.java
```diff
@@ -366,4 +367,32 @@ public static <T> T parseDiscreteValues(T[] allowedValues, Function<T, String> s
 						.collect( Collectors.toList() )
 		);
 	}
+
+	public static byte[] parseBytePrimitiveArray(String value) {
+		try {
+			String[] values = value.replaceAll( "(^\\s*\\[\\s*)|(\\s*\\]\\s*$)", "" ).split( "[,;\\s]+" );
```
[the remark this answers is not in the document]
marko-bekhta: IDK... I was thinking more like a format of `Arrays.toString(..)` so that's where this came from 😃

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/KnnPredicateSpecificsIT.java
```diff
@@ -0,0 +1,731 @@
...
+			StubMappingScope scope = index.createScope( indexDifferentMaxConnection );
+			SearchPredicateFactory f = scope.predicate();
+
+			String fieldPath = index.binding().field.get( fieldType ).relativeFieldName;
+
+			assertThat( scope.query().select()
+					.where( predicate( f, fieldPath, fieldType ).toPredicate() )
+					.fetchAllHits() )
+					.hasSize( 1 );
+		}
```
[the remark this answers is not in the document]
marko-bekhta: Right, at least that's what it seems like. I had these two as such that must match initially, but once I tried it in the test only dimension resulted in an error from Lucene, and as for the similarity -- it just didn't make sense to me to mix them. as for the max connections and beam width, these two are used for storage, I didn't see how it may affect the search algorithm, but at least it's not failing with exceptions 😃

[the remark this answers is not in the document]
yrodiere: Lucene has been notoriously permissive in the past, so just because it doesn't throw an exception, it doesn't mean that it supports what you're doing. They're improving these days, but still...

Considering we don't have many tests that check that the computed distance actually makes sense, I'm not sure we're in a position to promise this will work... 

Honestly I'd just lock all of this down, sending exceptions when the beamwidth/maxConnections are inconsistent, like you do for the dimension. If this bothers someone one day, well... they can open a Jira issue :shrug:

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/projection/FieldProjectionBaseIT.java
```diff
@@ -76,7 +77,7 @@ class InObjectProjectionIT<F> extends InObjectProjectionConfigured<F> {
 
 	abstract static class InObjectProjectionConfigured<F>
 			extends AbstractProjectionInObjectProjectionIT<F, F, FieldProjectionTestValues<F>> {
-		private static final List<FieldTypeDescriptor<?>> supportedFieldTypes = FieldTypeDescriptor.getAll();
+		private static final List<StandardFieldTypeDescriptor<?>> supportedFieldTypes = FieldTypeDescriptor.getAllStandard();
```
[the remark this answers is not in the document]
marko-bekhta: yeah, it's all the multivalued field variations in the tests...  I'll add something to the docs

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/projection/FieldProjectionSingleValuedBaseIT.java
```diff
@@ -44,12 +44,12 @@
 
 class FieldProjectionSingleValuedBaseIT<F> {
 
-	private static final List<FieldTypeDescriptor<?>> supportedFieldTypes = FieldTypeDescriptor.getAll();
+	private static final List<StandardFieldTypeDescriptor<?>> supportedFieldTypes = FieldTypeDescriptor.getAllStandard();
```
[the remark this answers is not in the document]
marko-bekhta: Even though it says that it is a single-value test some multivalued fields sneaked in 😄. will make some changes...

backend/lucene/src/main/resources/META-INF/services/org.apache.lucene.codecs.KnnVectorsFormat
```diff
@@ -0,0 +1 @@
+org.hibernate.search.backend.lucene.lowlevel.codec.impl.HibernateSearchKnnVectorsFormat
```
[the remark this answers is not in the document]
marko-bekhta: +1 this is the same thing as with the codec ... I've made some similar changes to it as well. this is also loaded on init if it was written into the index...

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/binding/impl/ValueBindingContextImpl.java
marko-bekhta: > Why didn't you end up using the FieldModelContributor to change the default behavior depending on whether @VectorField or another annotation isused, as I suggested?

I'm not sure I got your idea correctly 😃, but I've made some changes in this commit here https://github.com/hibernate/hibernate-search/pull/3814/commits/4336a365d28285797e8c4b8be9f9874c614c87bb let me know if that's what you had in mind 😃

yrodiere: Close enough, here's what I had in mind: https://github.com/marko-bekhta/hibernate-search/pull/215
Either will work though :shrug:

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/VectorField.java
```diff
@@ -0,0 +1,99 @@
...
+ * @see org.hibernate.search.engine.search.predicate.dsl.KnnPredicateFieldStep
+ */
+@Documented
+@Target({ ElementType.METHOD, ElementType.FIELD })
+@Retention(RetentionPolicy.RUNTIME)
+@Repeatable(VectorField.List.class)
+@PropertyMapping(processor = @PropertyMappingAnnotationProcessorRef(type = VectorFieldAnnotationProcessor.class,
+		retrieval = BeanRetrieval.CONSTRUCTOR))
+@Incubating
+public @interface VectorField {
```
yrodiere, line 42: I was looking into problems one might encounter because of the mandatory number of dimensions, and opened https://hibernate.atlassian.net/browse/HSEARCH-5021, but I just noticed... this annotation doesn't support passing a bridge/binder at all?

Feel free to repurpose https://hibernate.atlassian.net/browse/HSEARCH-5021 to address this later if necessary.

marko-bekhta:
I've added a few more commits for the bridges/binders and some checks. you'll find them after the commit for step initiatiors 😃. Also eclipse compiler was complaining so I had some more changes applied to the TCK tests

engine/src/main/java/org/hibernate/search/engine/cfg/spi/ParseUtils.java
yrodiere, line 373: Sonar is warning about DOS attacks here. We'll need to address this in a follow-up issue.

yrodiere (approved):
Thanks! I couldn't read it all (100 changed files since last time...) but what I've seen is looking great.

I think we'll have to adjust bridge/binder support a bit, but it's already a great start (I didn't think about solving it that way!), so we'll see about that in https://hibernate.atlassian.net/browse/HSEARCH-5021

Merging!

yrodiere:
> Started a build for all envs here https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-3814/26/pipeline to make sure there will be no surprises once merged (I've tried to build with JDK 11 and had to update some of the new tests 😔)

The cockroachdb failure is definitely a flaky test. As for the eclipse compiler thing... I don't know, it probably got cancelled when the other one failed. Let's merge and see what happens on main.

Thanks for all this work!

marko-bekhta:
> As for the eclipse compiler thing

I've just tried it locally and it passed, so hopefully no surprises there 🤞🏻

engine/src/main/java/org/hibernate/search/engine/cfg/spi/ParseUtils.java
yrodiere: https://hibernate.atlassian.net/browse/HSEARCH-5023

PR #3815, opened 2023-11-03
HSEARCH-4995 Use suppressed exceptions instead of logging for nested failures in RootFailureCollector

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4995

We still log such exceptions just in case, but at TRACE level (which should usually be disabled).

Please do not merge in 7.0, this is 7.1 material.


engine/src/main/java/org/hibernate/search/engine/reporting/spi/RootFailureCollector.java
```diff
@@ -230,12 +229,11 @@ private void doAdd(Throwable failure, String failureMessage) {
 			appendContextTo( contextJoiner );
 			log.newCollectedFailure( root.process, contextJoiner.toString(), failure );
```
marko-bekhta, line 230: I wonder, now that this log has become a TRACE, is it worth rendering the context if the log level is not TRACE?

marko-bekhta (approved):
🎉 😃 

> this is 7.1 material.

😨 😈 😏

engine/src/main/java/org/hibernate/search/engine/reporting/spi/RootFailureCollector.java
yrodiere: Right... I'll see about that.

yrodiere:
I addressed your comment, thanks!

yrodiere:
So, we're heading towards a 7.0.0.CR2 after all, so... let's merge this in 7.0 after all?

marko-bekhta:
yes ! 😄

PR #3820, opened 2023-11-08
HSEARCH-5005 Upgrade to Elasticsearch client 8.11.0 and test against Elasticsearch 8.11.0

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5005

Tests passed locally, let's see what CI will tell us.
Should we also backport to 6.2?


yrodiere (approved):
Thanks, LGTM.

>  Should we also backport to 6.2?

Not worth it IMO.

PR #3826, opened 2023-11-10
HSEARCH-4999 Use geo_shape instead of geo_polygon in Elasticsearch spatial predicates

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4999

Luckily, Elasticsearch 7.10 already support this query 🎉. I've tried using the relation operator (`WITHIN`/`CONTAINS`) 
(https://www.elastic.co/guide/en/elasticsearch/reference/8.11/query-dsl-geo-shape-query.html#geo-shape-spatial-relations) but since older Elasticsearch versions (e.g. 7.10 ...) can only work with the default `INTERSECTS` value I've removed it 😃 
tested it locally with Elasticsearch 7.10.1/7.17.14/8.11.0 and OpenSearch 1.3.13/2.11.0


yrodiere (approved):
Works for me if this gets rid of the warnings; thanks!

PR #3828, opened 2023-11-13
HSEARCH-5010 Disable skipping in comparators

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5010

I can only reproduce the test failure from time to time... the problem is caused by this change:
https://github.com/marko-bekhta/lucene/commit/d9109907bcafd9e3ac6a23bff9e5b1b2605bb2f8#diff-79c6a57519ecd1ef504629e62e13d17859a4ffedc58f4602e583ce758a15adc8R288-R295

Since our RawNumericDocValues throws exceptions, the code in the mentioned block just fails. When I've tried to "implement" the two methods throwing the exception -- instead of a failed test because of the exception, the order was incorrect. But again, the issue cannot be reproduced consistently ... 😔

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/comparator/impl/DoubleValuesSourceComparator.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/comparator/impl/DoubleValuesSourceComparator.java
@@ -34,6 +34,7 @@ public class DoubleValuesSourceComparator extends DoubleComparator {
 	private class DoubleValuesSourceLeafComparator extends DoubleLeafComparator {
 		DoubleValuesSourceLeafComparator(LeafReaderContext context) throws IOException {
 			super( context );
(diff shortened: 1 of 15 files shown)


yrodiere (approved):
LGTM, please port to the lucene9 branch as well :)

> But again, the issue cannot be reproduced consistently ... 😔

This doesn't worry me too much, as skipping probably is only possible when documents are distributed a specific way across fragments, and that distribution is notably unpredictable (it depends on the order of indexing operations coming from multiple threads, I think).

> Since our RawNumericDocValues throws exceptions, the code in the mentioned block just fails

I think we'll want to solve this eventually, because skipping most likely enables better performance.

I created https://hibernate.atlassian.net/browse/HSEARCH-5014, could you please add the stracktrace there in a comment?

marko-bekhta:
Thanks for taking a look 😃. Rebased and pushed both (lucene9/lucene10) branches. Closing this draft.

PR #3838, opened 2023-11-21
Update Groovy script to list required test containers

marko-bekhta:
to work with non-docker-io registries (e.g. db2/mssql)

Right now, these won't be prefetched. It should still work out ok in the end as they can be downloaded without docker auth. But let's fetch them the same as we are doing for other  containters.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -967,7 +967,7 @@
                                         }
                                     }
                                     // Exclude images from non-dockerhub repositories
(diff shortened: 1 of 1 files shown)


yrodiere (approved):
Right, I guess for images that explicitly target a different registry than docker.io... pulling them anyway cannot hurt, as we know we'll use them later.
And this code is simpler, so it's better!

PR #3840, opened 2023-11-28
HSEARCH-5022 Mention WildFly in the documentation

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-5022

I didn't add links about specifics (like outbox-polling); we'll need to add that to the WildFly documentation.

Also, I'll send a different PR to 6.0 to mention something in the migration guide.

### DIFF
--- a/documentation/src/main/asciidoc/public/reference/_compatibility.adoc
+++ b/documentation/src/main/asciidoc/public/reference/_compatibility.adoc
@@ -105,7 +105,7 @@ link:{wildflyUrl}[WildFly] includes modules for <<mapper-orm,Hibernate Search wi
 using either the <<backend-lucene,Lucene backend>> or the <<backend-elasticsearch,Elasticsearch backend>>.
 
 To start using Hibernate Search within WildFly,
-see the https://docs.wildfly.org/30/Developer_Guide.html#using-hibernate-search[Hibernate Search section in the WildFly Developer Guide]:
+see the link:{wildflyDeveloperGuideUrl}#using-hibernate-search[Hibernate Search section in the WildFly Developer Guide]:
 it covers all the specifics of WildFly.
 
 [[compatibility-framework-spring-boot]]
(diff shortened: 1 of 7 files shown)


documentation/src/main/asciidoc/public/reference/_compatibility.adoc
```diff
@@ -89,14 +89,25 @@ is also relevant for the OpenSearch distribution.
...
 _and_ it covers the specifics of Quarkus.
 
+[[compatibility-framework-wildfly]]
+=== WildFly
+
+link:{wildflyUrl}[WildFly] includes modules for <<mapper-orm,Hibernate Search with Hibernate ORM>>
+using either the <<backend-lucene,Lucene backend>> or the <<backend-elasticsearch,Elasticsearch backend>>.
+
+To start using Hibernate Search within WildFly,
+see the https://docs.wildfly.org/30/Developer_Guide.html#using-hibernate-search[Hibernate Search section in the WildFly Developer Guide]:
```
marko-bekhta, line 108: hmm I didn't see any "latest" or something like that link to the docs so that'd mean that we should keep track of it? 
I wonder if we maybe can add https://repo1.maven.org/maven2/org/wildfly/bom/wildfly/30.0.0.Final/wildfly-30.0.0.Final.pom to dependency management and parse the version out of it and use as a property?

marko-bekhta (approved):
😃

documentation/src/main/asciidoc/public/reference/_compatibility.adoc
yrodiere: > hmm I didn't see any "latest" or something like that link to the docs so that'd mean that we should keep track of it?

Neither did I. They just don't have a "latest" URL.

>  I wonder if we maybe can add https://repo1.maven.org/maven2/org/wildfly/bom/wildfly/30.0.0.Final/wildfly-30.0.0.Final.pom to dependency management and parse the version out of it and use as a property?

I guess that would be considered a build dependency, then? Like Spring. Ok, I'll add that.

yrodiere: Done! And I checked on my fork that Dependabot actually updates the dependency.

PR #3844, opened 2023-11-28
Address Spring Boot upgrade changes

marko-bekhta:
From what it looks the nested jars got updated. The new format is:
```
nested:/home/example/my.jar/!BOOT-INF/lib/my-nested.jar
nested:/home/example/my.jar/!BOOT-INF/classes/
```
and I see there's a filesystem for these nested jars. so the classes should get loaded just fine on their own as long as all those components are registered (and they should if it's a Spring app 😃)


util/common/src/test/java/org/hibernate/search/util/common/jar/impl/CodeSourceTest.java
```diff
@@ -22,27 +22,49 @@
...
+	static void beforeAll() {
+		preconfiguredPackages = System.getProperty( PROTOCOL_HANDLER_PACKAGES, "" );
+	}
+
+	@BeforeEach
+	void setUp() {
+		// Spring has some of their own handlers configured through this property.
+		// Handlers.register(); is required to get them registered.
+		// We want to have them available for tests that rely on Spring Boot.
+		// We don't want them to interfere with non-spring tests, so we reset the property before running a test.
```
yrodiere, line 63: Maybe use `SystemHelper.setSystemProperty`?

See `org.hibernate.search.integrationtest.performance.backend.elasticsearch.SmokeIT`

util/common/src/main/java/org/hibernate/search/util/common/jar/impl/CodeSource.java
```diff
@@ -38,6 +38,7 @@ class CodeSource implements Closeable {
 	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
 
 	private static final String JAR_URI_PATH_SEPARATOR = "!/";
+	private static final String NESTED_SPEC_PREFIX = "nested:";
```
yrodiere, line 41: Maybe add a comment to make it clear this is about Spring stuff?

yrodiere (approved):
Thanks. LGTM, though I suggested a few improvements.

yrodiere (approved):

yrodiere (approved):

PR #3848, opened 2023-11-29
HSEARCH-5025 Add container extraction support to @VectorField

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5025

I've dropped the `AbstractBaseFieldAnnotationProcessor` that was there only to *not* support extractors by vector fields 😃 and moved things back to the original place in `AbstractFieldAnnotationProcessor`.

if  we just enable the extraction then the default ones will build a path up to extracting from the arrays included and we'd end up having a byte/float value, which won't be considered as a vector. 

So I wonder if it's reasonable enough to ask the user to explicitly provide a path, and keep things disabled by default. WDYT?


yrodiere (changes requested):
Ideally we'd update the algorithm that resolves the default container extractors to take into account our expectations. I.e. we want the algorithm to stop extracting at the first extracted type that either:

* is accepted by the explicit bridge/binder (if one is set)
* or resolved to a default bridge/binder that results in field type (explicit or inferred) compatible with the annotation (`@VectorField`/`@FullTextField`).

However. That would be a big change, probably backwards-incompatible, and probably complex to implement, not to mention to test.

An alternative would be to "tell" the algorithm, when using `@VectorField`, that the default extract path should ignore all multi-valued extractors. That might be simpler to implement, and given that we don't support multi-valued vector fields anyway, shouldn't be too much of a problem.
I'm thinking that a simple boolean (`allowMultiValued`) passed to `org.hibernate.search.mapper.pojo.mapping.building.impl.PojoIndexModelBinder#bindExtractorPath` could do the trick.
We can test whether an extractor is multi-valued in `org.hibernate.search.mapper.pojo.extractor.impl.ContainerExtractorBinder.SingleExtractorContributor#tryAppend` using new metadata extracted from `ContainerExtractorDefinition` and passed to the constructor of `SingleExtractorContributor`; I don't think `ContainerExtractor#multiValued` would help here.
And this could be useful elsewhere, e.g. in `IdentityMappingCollectorPropertyNode`.
Not sure how intuitive that would be, though.

Anyway, if the above is too complex, then yes, I think your solution is reasonable.

Please consider the alternative I mentioned above, and in any case (whether you stick to the current solution or try my alternative) please update the reference documentation to take your changes into account? I think you'll need a note in the documentation of `@VectorField` as well as the `extractors` attribute and maybe [here](https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#mapping-containerextractor) too.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/extractor/mapping/programmatic/ContainerExtractorPath.java
```diff
@@ -36,11 +36,18 @@
...
-			true, Collections.emptyList()
+			DefaultExtractorsApplicability.ANY, Collections.emptyList()
+	);
+	private static final ContainerExtractorPath DEFAULT_SINGLE_VALUED = new ContainerExtractorPath(
+			DefaultExtractorsApplicability.SINGLE_VALUED, Collections.emptyList()
 	);
 	private static final ContainerExtractorPath NONE = new ContainerExtractorPath(
-			false, Collections.emptyList()
+			DefaultExtractorsApplicability.NO, Collections.emptyList()
 	);
```
marko-bekhta, line 51: Does it make sense to have it like this? It's not what you were suggesting, but it looked less intrusive in terms of adding the boolean parameter everywhere ... 🙈

yrodiere: I was hoping my suggestion would only force you to add the boolean (or enum, whatever) from `org.hibernate.search.mapper.pojo.mapping.building.spi.PojoIndexMappingCollectorPropertyNode#value` down to `org.hibernate.search.mapper.pojo.extractor.impl.ContainerExtractorBinder#bindPath`, and that it wouldn't have too many side-effects... I guess I was wrong?

Changing `ContainerExtractorPath` is a possibility, but it has ramifications as well, this time on APIs... We'd need to start thinking about what it means for users to use `defaultSingleValued()` extractor paths elsewhere...

And in fact I suspect that having multiple intepretations of what is the "default" will lead to problems with `org.hibernate.search.mapper.pojo.extractor.impl.ContainerExtractorBinder#isDefaultExtractorPath`...

Maybe it would be better to stick to what you initially did for now, and introduce support for "default" extractors once we have the time to design a future-proof solution that takes into account all the possible corner cases?

marko-bekhta: yeah, it seems like that would be a safer option ... see the second commit I've just added... as soon as I update `ContainerExtractorBinder#bindPath` and I haven't even changed the `ContainerExtractorBinder#tryBindPath` 😔

yrodiere: I'll give you my perfectly objective, non-biased and very nuanced opinion: `indexNullAs` is shit ;D

I mean, what does it even mean to index a missing value as `0` or `"somestring"` or `[0,0,0]`? Why would you want missing values to be considered close to `[0,0,0]`? Can one even come up with a vector (`[0,0,0]` or other) that properly represents a "missing value" in the vector space of your AI model? I doubt it.

`Optional<float[]>` would just be a more fancy way to say "I may not have data for this". E.g. "this book doesn't have a cover", or "I haven't turned the cover of this book into a vector yet".
So, yeah, maybe not a mainstream use case. But a relevant one nonetheless. 

Anyway. Let's go with your first solution, it has the advantage of being clear and straightforward, allows people to map `Option<float[]>` if they want to, and opens the door to supporting default extractors in the future, with some more changes.

Thanks for investigating!

yrodiere (approved):

PR #3856, opened 2023-12-05
HSEARCH-5004 Default to SqlTypes.UUID instead of CHAR for the OutboxEvent/Agent ID in databases supporting it

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5004

### DIFF
--- a/documentation/src/main/asciidoc/migration/index.adoc
+++ b/documentation/src/main/asciidoc/migration/index.adoc
@@ -65,11 +65,197 @@ older indexes can be read from and written to without reindexing.
 [[outboxpolling]]
 === Outbox polling database tables
 
-The event and agent database tables used for
-link:https://docs.jboss.org/hibernate/search/{hibernateSearchVersionShort}/reference/en-US/html_single/#coordination-database-polling[outbox-polling]
-in Hibernate Search {hibernateSearchVersion}
-are backward-compatible with Hibernate Search {hibernateSearchPreviousStableVersionShort}:
-no database schema update is necessary for these tables.
+If you use the incubating link:{hibernateSearchDocUrl}#coordination-outbox-polling[`outbox-polling` coordination strategy],
+you may be impacted by the changes to entity mapping for entities that represents the outbox event and agent,
+requiring database schema changes.
+
+Changes are only applicable if the `hibernate.search.coordination.entity.mapping.*.uuid_type` was set to `default` (implicitly or explicitly).
+In case when schema changes are not possible to apply at the moment, the configuration settings can be provided
+to specify the type to represent UUIDs in the database:
+
+[source, properties]
+----
+hibernate.search.coordination.entity.mapping.outboxevent.uuid_type=CHAR
+hibernate.search.coordination.entity.mapping.agent.uuid_type=CHAR
+----
+
+The `id` column changes its type from the current one to the one used by the corresponding Hibernate ORM Database dialect
+by default for representing the uuid type, except the following databases where no changes are expected:
+
+* MSSQL: no migration required. Type of the `id` remains `binary(16)` for both tables.
+
+For other databases see suggested migration scripts:
+
+.Postgresql:
+[,sql]
+----
+-- change outbox event `id` column type to uuid:
+alter table hsearch_outbox_event
+    alter column id TYPE uuid USING cast(id as uuid);
+
+-- change agent `id` column type to uuid:
+alter table hsearch_agent
+    alter column id TYPE uuid USING cast(id as uuid);
+----
+
+.CockroachDB:
+[,sql]
+----
+-- change outbox event `id` column type to uuid:
+-- altering type directly is not supported: https://go.crdb.dev/issue-v/47636/v22.1
+alter table hsearch_outbox_event
+    add tmp uuid;
+update hsearch_outbox_event
+set tmp = cast(id as uuid)
+where 1 = 1;
+alter table hsearch_outbox_event
+    alter column tmp set not null;
+alter table hsearch_outbox_event
+    alter primary key using columns (tmp);
+alter table hsearch_outbox_event
+    drop column id;
+alter table hsearch_outbox_event
+    rename column tmp to id;
+
+-- change agent `id` column type to char:
+alter table hsearch_agent
+    add tmp uuid;
+update hsearch_agent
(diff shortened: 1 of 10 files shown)


mapper/orm-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/outboxpolling/cfg/impl/UuidDataTypeUtils.java
```diff
@@ -61,12 +61,12 @@ else if ( UUID_BINARY.equalsIgnoreCase( value ) ) {
 		return TypeCodeConverter.convert( value );
 	}
 
-	public static int defaultUuidType(Dialect dialect) {
+	public static Integer defaultUuidType(Dialect dialect) {
 		// See HSEARCH-4749 why MSSQL is treated differently
 		if ( dialect instanceof SQLServerDialect ) {
 			return SqlTypes.BINARY;
 		}
```
yrodiere, line 68: Do we still need this `if` block? Binary would be the default on that dialect anyway, no?

documentation/src/main/asciidoc/migration/index.adoc
```diff
@@ -65,11 +65,197 @@ older indexes can be read from and written to without reindexing.
...
-link:https://docs.jboss.org/hibernate/search/{hibernateSearchVersionShort}/reference/en-US/html_single/#coordination-database-polling[outbox-polling]
-in Hibernate Search {hibernateSearchVersion}
-are backward-compatible with Hibernate Search {hibernateSearchPreviousStableVersionShort}:
-no database schema update is necessary for these tables.
+If you use the incubating link:{hibernateSearchDocUrl}#coordination-outbox-polling[`outbox-polling` coordination strategy],
+you may be impacted by the changes to entity mapping for entities that represents the outbox event and agent,
+requiring database schema changes.
+
+Changes are only applicable if the `hibernate.search.coordination.entity.mapping.*.uuid_type` was set to `default` (implicitly or explicitly).
+In case when schema changes are not possible to apply at the moment, the configuration settings can be provided
```
yrodiere, line 73:
```suggestion
If schema changes are not possible to apply at the moment, the following settings can be provided
```

documentation/src/main/asciidoc/migration/index.adoc
```diff
@@ -65,11 +65,197 @@ older indexes can be read from and written to without reindexing.
...
-The event and agent database tables used for
-link:https://docs.jboss.org/hibernate/search/{hibernateSearchVersionShort}/reference/en-US/html_single/#coordination-database-polling[outbox-polling]
-in Hibernate Search {hibernateSearchVersion}
-are backward-compatible with Hibernate Search {hibernateSearchPreviousStableVersionShort}:
-no database schema update is necessary for these tables.
+If you use the incubating link:{hibernateSearchDocUrl}#coordination-outbox-polling[`outbox-polling` coordination strategy],
+you may be impacted by the changes to entity mapping for entities that represents the outbox event and agent,
+requiring database schema changes.
+
+Changes are only applicable if the `hibernate.search.coordination.entity.mapping.*.uuid_type` was set to `default` (implicitly or explicitly).
```
yrodiere, line 72:
```suggestion
Changes are only necessary if configuration properties `hibernate.search.coordination.entity.mapping.*.uuid_type` are not set or are set to `default`.
```

documentation/src/main/asciidoc/migration/index.adoc
```diff
@@ -65,11 +65,197 @@ older indexes can be read from and written to without reindexing.
...
+to specify the type to represent UUIDs in the database:
+
+[source, properties]
+----
+hibernate.search.coordination.entity.mapping.outboxevent.uuid_type=CHAR
+hibernate.search.coordination.entity.mapping.agent.uuid_type=CHAR
+----
+
+The `id` column changes its type from the current one to the one used by the corresponding Hibernate ORM Database dialect
+by default for representing the uuid type, except the following databases where no changes are expected:
```
yrodiere, line 83: I must admit I have a hard time understanding this sentence... Maybe rephrase this way, and add a link to the ORM docs?

```suggestion
The default type for UUIDs in outbox-polling tables now match
https://docs.jboss.org/hibernate/orm/6.4/userguide/html_single/Hibernate_User_Guide.html#basic-uuid[the default type for UUIDs in Hibernate ORM].

The default behavior should not change for the following databases:
```

documentation/src/main/asciidoc/migration/index.adoc
yrodiere, line 1: I think you will probably need to update the reference documentation as well?

You'll probably need to add a link to https://docs.jboss.org/hibernate/orm/6.4/userguide/html_single/Hibernate_User_Guide.html#basic-uuid there (don't hardcode the ORM version though, I think we have an Asciidoc property for links to the ORM documentation)

yrodiere (changes requested):
Thanks.
So Hibernate ORM default to binary for UUIDs in most dialects. That's... underwhelming.
Ah, well, that's not in our hands :shrug:

mapper/orm-outbox-polling/src/main/java/org/hibernate/search/mapper/orm/outboxpolling/cfg/impl/UuidDataTypeUtils.java
marko-bekhta: let me double-check if that's the default for MSSQL as well.. also I left it in as this was something that we wanted to force (even if something changes in the dialect we'd still pick the binary type ourselves)

documentation/src/main/asciidoc/migration/index.adoc
marko-bekhta: https://docs.jboss.org/hibernate/search/7.1/reference/en-US/html_single/#_custom_schematable_nameetc

we had it vaguely described, so it still is true with the changes 😄 but let me add the link you've mentioned too

yrodiere (approved):

marko-bekhta:
the failed cockroach test seems to be the one failing in other branches .... 
as for the mssql one -- it was an incorrect expected type, so I've updated it. new build for all DBs: https://ci.hibernate.org/job/hibernate-search/job/PR-3856/5/

yrodiere (approved):

PR #3863, opened 2023-12-11
HSEARCH-5021 Improve support for custom value bridges for vector fields

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5021

This is just a draft for now so we can talk over the code 😃

### DIFF
--- a/integrationtest/mapper/pojo-standalone-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/standalone/realbackend/mapping/VectorFieldIT.java
+++ b/integrationtest/mapper/pojo-standalone-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/standalone/realbackend/mapping/VectorFieldIT.java
@@ -1,0 +1,106 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.integrationtest.mapper.pojo.standalone.realbackend.mapping;
+
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import java.lang.invoke.MethodHandles;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Objects;
+
+import org.hibernate.search.integrationtest.mapper.pojo.standalone.realbackend.testsupport.BackendConfigurations;
+import org.hibernate.search.mapper.pojo.bridge.ValueBridge;
+import org.hibernate.search.mapper.pojo.bridge.binding.ValueBindingContext;
+import org.hibernate.search.mapper.pojo.bridge.mapping.annotation.ValueBinderRef;
+import org.hibernate.search.mapper.pojo.bridge.mapping.programmatic.ValueBinder;
+import org.hibernate.search.mapper.pojo.bridge.runtime.ValueBridgeFromIndexedValueContext;
+import org.hibernate.search.mapper.pojo.bridge.runtime.ValueBridgeToIndexedValueContext;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.DocumentId;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.Indexed;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.VectorField;
+import org.hibernate.search.util.common.SearchException;
+import org.hibernate.search.util.impl.integrationtest.common.reporting.FailureReportUtils;
+import org.hibernate.search.util.impl.integrationtest.mapper.pojo.standalone.StandalonePojoMappingSetupHelper;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+class VectorFieldIT {
+
+	private static final String INDEX_NAME = "IndexName";
+
+	@RegisterExtension
+	public StandalonePojoMappingSetupHelper setupHelper = StandalonePojoMappingSetupHelper.withSingleBackend(
+			MethodHandles.lookup(), BackendConfigurations.simple() );
+
+	/*
+	* This test relies on a backend implementation to make sure that the vector dimension was somehow set for the field.
+	* hence it requires a real backend.
+	 */
+	@Test
+	void customBridge_vectorDimensionUnknown() {
+		@Indexed(index = INDEX_NAME)
+		class IndexedEntity {
+			@DocumentId
+			Integer id;
+			@VectorField(valueBinder = @ValueBinderRef(type = ValidImplicitTypeBridge.ValidImplicitTypeBinder.class))
+			Collection<Float> floats;
+		}
+
+		assertThatThrownBy( () -> setupHelper.start().expectCustomBeans().setup( IndexedEntity.class ) )
+				.isInstanceOf( SearchException.class )
+				.satisfies( FailureReportUtils.hasFailureReport()
+						.typeContext( IndexedEntity.class.getName() )
+						.pathContext( ".floats" )
+						.indexContext( INDEX_NAME )
+						.failure(
+								"Invalid index field type: missing vector dimension."
+										+ " Define the vector dimension explicitly."
+										// hint:
+										+ " Either specify dimension as an annotation property (@VectorField(dimension = somePositiveInteger)), or define a value binder (@VectorField(valueBinder = @ValueBinderRef(..))) that explicitly declares a vector field specifying the dimension."
+						) );
+	}
+
+	@SuppressWarnings("rawtypes")
+	public static class ValidImplicitTypeBridge implements ValueBridge<Collection, float[]> {
+
+		public static class ValidImplicitTypeBinder implements ValueBinder {
+
+			@Override
+			public void bind(ValueBindingContext<?> context) {
+				context.bridge( Collection.class, new ValidImplicitTypeBridge() );
+			}
+		}
+
+		@Override
+		public float[] toIndexedValue(Collection value, ValueBridgeToIndexedValueContext context) {
+			if ( value == null ) {
+				return null;
+			}
+			float[] result = new float[value.size()];
+			int index = 0;
+			for ( Object o : value ) {
+				result[index++] = Float.parseFloat( Objects.toString( o, null ) );
+			}
+			return result;
+		}
+
+		@Override
+		public Collection fromIndexedValue(float[] value, ValueBridgeFromIndexedValueContext context) {
+			if ( value == null ) {
+				return null;
+			}
+			List<Float> floats = new ArrayList<>( value.length );
+			for ( float v : value ) {
+				floats.add( v );
+			}
+			return floats;
+		}
+	}
+}
--- a/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/programmatic/impl/PropertyMappingVectorFieldOptionsStepImpl.java
+++ b/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/programmatic/impl/PropertyMappingVectorFieldOptionsStepImpl.java
@@ -36,14 +36,7 @@ class PropertyMappingVectorFieldOptionsStepImpl
 						if ( dimension == null ) {
 							throw log.vectorDimensionNotSpecified();
 						}
-						// we consider arrays (byte[] and float[] in particular to be vector types, anything else will be
-						// treated as a non-vector field and delegated to a regular as(..) call.
-						if ( clazz.isArray() ) {
-							return factory.asVector( dimension, clazz );
-						}
-						else {
-							return factory.as( clazz );
-						}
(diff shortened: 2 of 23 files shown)


mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/processing/impl/AbstractFieldAnnotationProcessor.java
```diff
@@ -46,8 +46,15 @@ public final void process(PropertyMappingStep mappingContext, A annotation,
...
 				context
 		);
+		Map<String, Object> params = context.toMap( annotation );
+		for ( Map.Entry<String, Object> entry : context.toMap( valueBinder.params() ).entrySet() ) {
+			Object overriddenValue = params.put( entry.getKey(), entry.getValue() );
+			if ( overriddenValue != null ) {
+				log.annotationAttributeOverriddenWithParameter( annotation.annotationType(), entry.getKey(), overriddenValue,
+						entry.getValue() );
+			}
+		}
```
marko-bekhta, line 56: > They need to know about the number of dimensions, because that’s a mandatory parameter of IndexFieldTypeFactory#asVector…  but they don't have access to the number passed to @VectorField#dimension.

the idea here that we just give access to all the attributes from the annotation as parameters to the binder, and then binder implementers can do what they want with the info...

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/programmatic/impl/PropertyMappingVectorFieldOptionsStepImpl.java
```diff
@@ -36,14 +36,7 @@ public <F> IndexFieldTypeOptionsStep<?, F> initiate(IndexFieldTypeFactory factor
...
 						}
-						// we consider arrays (byte[] and float[] in particular to be vector types, anything else will be
-						// treated as a non-vector field and delegated to a regular as(..) call.
-						if ( clazz.isArray() ) {
-							return factory.asVector( dimension, clazz );
-						}
-						else {
-							return factory.as( clazz );
-						}
+						return factory.asVector( dimension, clazz );
```
marko-bekhta, line 39: if we get rid of this check then we are never getting to this check to fail:
https://github.com/hibernate/hibernate-search/blob/ce6e483c2322eee9a1cc8f2dbda73f3823035e94/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/binding/impl/FieldModelContributorContextImpl.java#L82-L91

since we have only one vector-type interface, so no matter what we pass in to `factory#asVector(...)`  we are either getting a vector step (so not going to fail here) or we are passing something that will fail inside the factory itself with `log#cannotGuessVectorFieldType(..)`...

So it seems if we'd remove this check from here, then we'd need to update the `log#cannotGuessVectorFieldType` and all the info from the other message about using bridges and stuff... 

WDYT?

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/processing/impl/AbstractFieldAnnotationProcessor.java
yrodiere: I'm not sure this is a good idea, as it essentially breaks the clear separation we used to have between annotations and the rest of the code, which allowed it to work in contexts where Java annotations don't make sense (e.g. Infinispan schema definitions).

As for suggestions... IDK, I was thinking of simply making `dimensions()` optional in the annotation and in `IndexFieldTypeFactory`, moving it back to an optional parameter. We could handle it a bit like we handle `decimalScale`.
Then binders would be able to set the number dimensions, and mapping (through annotations or programmatic API) would be able to set it explicitly when the binder doesn't (or even to override the binder's choice, though I doubt that is useful...).
Admittedly that would not be a great API but at least it would be consistent with the existing APIs and behaviors. 

Ideally the number of dimensions should really be part of the type: we don't map to a `float[]`, but to a `@Dimensions(10) float[]`. I can imagine an API supporting something like that (e.g. `implements ValueBridge<ImageContent, @Dimensions(1024) float[]>`, `private @Dimensions(1024) float[] embeddings;`, ...). We could even handle it right now with "synthetic" subtypes of `float[]` represented by a `RawTypeModel` that has some additional metadata attached. But I suspect this would have side-effects and require extensive care to make sure edge cases work correctly and/or lead to clear error messages, including in places where we actually don't care about vectors.... So, probably an idea best kept on the shelve :x

`decimalScale()` is a bit the same thing as `dimensions()`, so we might as well handle dimensions the same way... I guess?

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/programmatic/impl/PropertyMappingVectorFieldOptionsStepImpl.java
yrodiere: I think the check could still fail with a custom binder that sets the field type to a non-vector type, no?
E.g. `@VectorField(valueBinder = @ValueBinderRef(type = MyIntegerBinder.class))`.

> So it seems if we'd remove this check from here, then we'd need to update the log#cannotGuessVectorFieldType and all the info from the other message about using bridges and stuff...

Can you point me to the test where you think the error message wouldn't be clear enough without this additional info?

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/processing/impl/AbstractFieldAnnotationProcessor.java
marko-bekhta: Right ... if we don't add new API, then we already have dimensions as optional on the annotation as well as on `PropertyMappingStep#vectorField`, so that part is covered? I was thinking more about the imaginary case 😄 
 where someone would add dimension to the annotation and then also provide a custom binder, and would want to somehow access that dimension value from the annotation. And for those who do not rely on annotations, it would be as is (just using the programmatic API, I assume?)

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/programmatic/impl/PropertyMappingVectorFieldOptionsStepImpl.java
yrodiere: I see.

If you make the number of dimensions "optional" (i.e. set later through the DSL), we could in theory expose a `asAny(Class)` method in `IndexFieldTypeFactory` that would return a standard or vector interface, whatever makes sense, and then the check you commented on would be more useful. We'd probably have to rename`IndexFieldTypeFactory#as()` to `asStandard` to be consistent though...

But I'm honestly not sure this is worth your time or the annoyance that renaming `as()` will cause... Just forwarding the exception from the backend, which should say something similar to what the stub backend says, would be enough IMO.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/processing/impl/AbstractFieldAnnotationProcessor.java
yrodiere: > where someone would add dimension to the annotation and then also provide a custom binder, and would want to somehow access that dimension value from the annotation

Unsupported, they should use a custom annotation and annotation processor that passes that info to the binder.

> And for those who do not rely on annotations, it would be as is (just using the programmatic API, I assume?)

Not sure what you mean by "as is".

marko-bekhta: > Not sure what you mean by "as is".

I meant that this change was only for the annotation processor, so only relevant for those who are using the annotations, and others who don't use them would not notice any difference 🙈 
But then if you are saying that in this case a user should just create a custom annotation, then I'll just drop this commit 😃

yrodiere: > I meant that this change was only for the annotation processor, so only relevant for those who are using the annotations, and others who don't use them would not notice any difference 🙈

Yes, people who use the programmatic API actually don't have this problem as they can pass information directly through the binder's constructor.

>  But then if you are saying that in this case a user should just create a custom annotation, then I'll just drop this commit 😃

:+1:

yrodiere:
Hey @marko-bekhta is this ready for review?

marko-bekhta:
hmm, good question 😃, looking at the changes, I don't see something to add...

yrodiere (commented):
From https://hibernate.atlassian.net/browse/HSEARCH-5021:

> At the moment (as of https://hibernate.atlassian.net/browse/HSEARCH-4948 ), if someone was to write a custom binder for vector fields with a custom field type:
>
> 1. They need to know about the number of dimensions, because that’s a mandatory parameter of IndexFieldTypeFactory#asVector…  but they don't have access to the number passed to `@VectorField#dimension`.
> 2. If they want to force the number of dimensions (because their bridge always produces vectors with a certain number of dimensions), users of the binder will still need to pass something to `@VectorField#dimension` in the mapping. Even though that value will never be used…

Does this PR addresses any of that? I'm honestly not sure :x 

Perhaps it would be best to add/find an integration test for each of those:

1. A custom bridge that turns for example a `List<Float>` into a `float[]`, and declares the field type explicitly as projectable by default, but doesn't know about the number of dimensions. Users must provide the number of dimensions in `@VectorField`.
2. A custom bridge that turns for example a `Triplet<Float, Float, Float>` into a `float[]`, and declares the field type explicitly with dimensions=3. Users must not be forced to provide a number of dimensions in `@VectorField`.

I think `org.hibernate.search.integrationtest.mapper.pojo.mapping.definition.VectorFieldIT#customBridge_explicitFieldType` already addresses the second item.

I can't find something about the first, though... Which I guess makes sense, since `org.hibernate.search.engine.backend.types.dsl.IndexFieldTypeFactory#asVector` forces you to pass a number of dimensions ATM?

marko-bekhta:
> I can't find something about the first, though... Which I guess makes sense, since org.hibernate.search.engine.backend.types.dsl.IndexFieldTypeFactory#asVector forces you to pass a number of dimensions ATM?

Right. And the initial idea was to add annotation attributes (`@VectorField` ones in particular) to the parameters so that the user would have access to dimension value from the annotation and could pass it to `asVector` call ... but then you were saying that that would be a users-writes-their-own-annotation-and-processor case so I assumed that that covers it and nothing to be done about it 😄... 

Now if we want to allow calls like `asVector()` without an initial dimension value ... I suppose we can do that, but that would mean reverting the changes we've done in the other PR... let me put something together and we can check it tomorrow.

yrodiere (approved):
> Now if we want to allow calls like `asVector()` without an initial dimension value ... I suppose we can do that

Thanks, looks like this addresses the issue completely! A few comments below.

> but that would mean reverting the changes we've done in the other PR... 

Yeah sorry about that. I guess I failed to consider this use case when I suggested that change in the initial PR...

> the initial idea was to add annotation attributes (@VectorField ones in particular) to the parameters so that the user would have access to dimension value from the annotation and could pass it to asVector call ...

Right, and that created a few [problems](https://github.com/hibernate/hibernate-search/pull/3863#discussion_r1422611393).

> but then you were saying that that would be a users-writes-their-own-annotation-and-processor case so I assumed that that covers it and nothing to be done about it 😄...

Yeah I was just talking about the specific case where the binder needs to get the number of dimensions as a parameter. Not about cases where the binder doesn't need to know about the number of dimensions (because it's implied by input).

marko-bekhta:
yeah, no worries 😃 and sorry that I didn't catch it that we should make these changes from your initial review 🙈 😃. I've applied your suggestions and cleaned up `PropertyMappingVectorFieldOptionsStepImpl` a bit more.

integrationtest/mapper/pojo-standalone-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/standalone/realbackend/mapping/VectorFieldIT.java
```diff
@@ -0,0 +1,106 @@
...
+	@RegisterExtension
+	public StandalonePojoMappingSetupHelper setupHelper = StandalonePojoMappingSetupHelper.withSingleBackend(
+			MethodHandles.lookup(), BackendConfigurations.simple() );
+
+	/*
+	* This test relies on a backend implementation to make sure that the vector dimension was somehow set for the field.
+	* hence it requires a real backend.
+	 */
+	@Test
+	void customBridge_vectorDimensionUnknown() {
```
marko-bekhta, line 47: Note that I've moved this test to a "real-backend" since the exception is based on the backend implementation. We probably could add something to a stub backend to make it work, but that to me seems like making assertions  `assertTrue(true)` in tests 😄

yrodiere (approved):
LGTM, thanks!

PR #3865, opened 2023-12-12
HSEARCH-3909 Allow looking up the capabilities of each field in the metamodel

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-3909

Best reviewed commit by commit, as the first commit is essentially a refactoring of tests to allow for easier testing of this new feature. (you'll probably want to skip reviewing that)

See the changes to the documentation and to `SearchMappingIndexedEntitiesIT` for a (short) description of the feature.

Note I chose to use strings for traits in anticipation for https://hibernate.atlassian.net/browse/HSEARCH-3633: if we want to be able to use traits in annotations, we can't use a custom type, and enums aren't great either as they aren't extensible by the Lucene/Elasticsearch backends.


yrodiere:
What the hell is this: https://sonarcloud.io/organizations/hibernate/rules?open=java%3AS5841&rule_key=java%3AS5841
I don't know if the trait sets are empty, and I don't care, I just want to check that they don't contain a particular trait :(
I'll see if I can alter the Sonar profiles.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/highlight/AbstractHighlighterIT.java
```diff
@@ -127,6 +129,52 @@ static void setup() {
 
 	abstract HighlighterOptionsStep<?> highlighter(SearchHighlighterFactory factory);
 
+	@Test
+	void highlightable_enabled_trait() {
+		assertThat( Arrays.asList( "string", "objectFlattened.string" ) )
+				.allSatisfy( fieldPath -> assertThat( index.toApi().descriptor().field( fieldPath ) )
+						.hasValueSatisfying( fieldDescriptor -> assertThat( fieldDescriptor.type().traits() )
+								.as( "traits of field '" + fieldPath + "'" )
+								.contains( "projection:highlight" ) ) );
```
marko-bekhta, line 138:
```suggestion
								.contains( IndexFieldTraits.Projections.HIGHLIGHT ) ) );
```

marko-bekhta (approved):
nice 👍🏻 
I've noticed that in tests we aren't always using trait constants but just strings. But I can't say that there's much benefit in going through all the tests to find them 😃.
I was wondering about the identifier field: is it needed for Infinispan?

yrodiere:
> I've noticed that in tests we aren't always using trait constants but just strings.

Yes that's on purpose, as this makes sure that tests will fail if we change constants (that's an API change, shouldn't happen).

We (generally) do the same thing in tests of configuration properties.

> But I can't say that there's much benefit in going through all the tests to find them 😃.

Agreed :)

>  I was wondering about the identifier field: is it needed for Infinispan?

The identifier is not a field... Not sure what you're referring to?

AFAIK we only allow the id predicate and id projection on identifers, and we always allow them, so the "traits" for the identifier would never change... doesn't seem useful?

marko-bekhta:
> The identifier is not a field... Not sure what you're referring to?
> 
> AFAIK we only allow the id predicate and id projection on identifers, and we always allow them, so the "traits" for the identifier would never change... doesn't seem useful?

Yes, I'm just not sure how Infinispan is going to use these traits and if they are treating an id as a field or not using the id at all in their queries, so that was why I've asked if we need to somehow include the id into all this 😃

yrodiere:
> > The identifier is not a field... Not sure what you're referring to?
> > AFAIK we only allow the id predicate and id projection on identifers, and we always allow them, so the "traits" for the identifier would never change... doesn't seem useful?
> 
> Yes, I'm just not sure how Infinispan is going to use these traits and if they are treating an id as a field or not using the id at all in their queries, so that was why I've asked if we need to somehow include the id into all this 😃

Okay, well as I said the capabilities of the identifier don't change, so they can hardcode it. And if they index the id as a field, they'll have metadata for that field. So, all good until we get feedback hinting otherwise :shrug: 

I'm not sure keys are indexed in Infinispan at the moment, though. That was a planned feature at some point, no idea if it was implemented.

yrodiere:
Hey @fax4ever , here's the PR to allow you to look up field capabilities ("traits"), in case you want to have a look. This will allow you to determine if the `knn` predicate can be used, among other things.

PR #3869, opened 2023-12-13
HSEARCH-5034 Allow passing BeanReference<? extends T> when registering beans to BeanConfigurationContext

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-5034


marko-bekhta (approved):
makes sense 😃 maybe let's build it with the Eclipse compiler as well once it's fixed (knowing how Eclipse one sometimes likes generics 🙈)

yrodiere:
Eclipse likes my changes! Merging.
Thanks for the review.

PR #3871, opened 2023-12-13
HSEARCH-4950 Vector search using Elasticsearch/OpenSearch

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4950

Opening as a draft as we need to address an additional parameter for the elasticsearch knn predicate before finishing up the integration.

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/dialect/model/impl/OpenSearch2ModelDialect.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/dialect/model/impl/OpenSearch2ModelDialect.java
@@ -1,0 +1,23 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.elasticsearch.dialect.model.impl;
+
+import org.hibernate.search.backend.elasticsearch.types.dsl.provider.impl.ElasticsearchIndexFieldTypeFactoryProvider;
+import org.hibernate.search.backend.elasticsearch.types.dsl.provider.impl.OpenSearch1IndexFieldTypeFactoryProvider;
+
+import com.google.gson.Gson;
+
+/**
+ * The model dialect for OpenSearch 2.x.
+ */
+public class OpenSearch2ModelDialect implements ElasticsearchModelDialect {
+
+	@Override
+	public ElasticsearchIndexFieldTypeFactoryProvider createIndexTypeFieldFactoryProvider(Gson userFacingGson) {
+		return new OpenSearch1IndexFieldTypeFactoryProvider( userFacingGson );
+	}
+}
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/dialect/model/impl/OpenSearch2ModelDialect.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/dialect/model/impl/OpenSearch2ModelDialect.java
@@ -7,9 +7,9 @@
 package org.hibernate.search.backend.elasticsearch.dialect.model.impl;
 
 import org.hibernate.search.backend.elasticsearch.types.dsl.provider.impl.ElasticsearchIndexFieldTypeFactoryProvider;
-import org.hibernate.search.backend.elasticsearch.types.dsl.provider.impl.OpenSearch1IndexFieldTypeFactoryProvider;
+import org.hibernate.search.backend.elasticsearch.types.dsl.provider.impl.OpenSearch2IndexFieldTypeFactoryProvider;
 import org.hibernate.search.backend.elasticsearch.validation.impl.ElasticsearchPropertyMappingValidatorProvider;
-import org.hibernate.search.backend.elasticsearch.validation.impl.OpenSearch1PropertyMappingValidatorProvider;
+import org.hibernate.search.backend.elasticsearch.validation.impl.OpenSearch2PropertyMappingValidatorProvider;
 
 import com.google.gson.Gson;
 
@@ -20,11 +20,11 @@ public class OpenSearch2ModelDialect implements ElasticsearchModelDialect {
 
 	@Override
 	public ElasticsearchIndexFieldTypeFactoryProvider createIndexTypeFieldFactoryProvider(Gson userFacingGson) {
-		return new OpenSearch1IndexFieldTypeFactoryProvider( userFacingGson );
+		return new OpenSearch2IndexFieldTypeFactoryProvider( userFacingGson );
 	}
 
 	@Override
 	public ElasticsearchPropertyMappingValidatorProvider createElasticsearchPropertyMappingValidatorProvider() {
-		return new OpenSearch1PropertyMappingValidatorProvider();
+		return new OpenSearch2PropertyMappingValidatorProvider();
 	}
 }
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/dialect/model/impl/OpenSearch2ModelDialect.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/dialect/model/impl/OpenSearch2ModelDialect.java
@@ -8,6 +8,8 @@ package org.hibernate.search.backend.elasticsearch.dialect.model.impl;
 
 import org.hibernate.search.backend.elasticsearch.types.dsl.provider.impl.ElasticsearchIndexFieldTypeFactoryProvider;
 import org.hibernate.search.backend.elasticsearch.types.dsl.provider.impl.OpenSearch1IndexFieldTypeFactoryProvider;
+import org.hibernate.search.backend.elasticsearch.validation.impl.ElasticsearchPropertyMappingValidatorProvider;
+import org.hibernate.search.backend.elasticsearch.validation.impl.OpenSearch1PropertyMappingValidatorProvider;
 
 import com.google.gson.Gson;
 
@@ -19,5 +21,10 @@ public class OpenSearch2ModelDialect implements ElasticsearchModelDialect {
 	@Override
 	public ElasticsearchIndexFieldTypeFactoryProvider createIndexTypeFieldFactoryProvider(Gson userFacingGson) {
 		return new OpenSearch1IndexFieldTypeFactoryProvider( userFacingGson );
+	}
+
+	@Override
+	public ElasticsearchPropertyMappingValidatorProvider createElasticsearchPropertyMappingValidatorProvider() {
+		return new OpenSearch1PropertyMappingValidatorProvider();
 	}
 }
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/lowlevel/index/mapping/impl/PropertyMapping.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/lowlevel/index/mapping/impl/PropertyMapping.java
@@ -18,8 +18,11 @@ import com.google.gson.annotations.SerializedName;
  * See https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping.html#mapping-type
  */
 /*
- * CAUTION: JSON serialization is controlled by a specific adapter, which must be
+ * CAUTION:
+ * 1. JSON serialization is controlled by a specific adapter, which must be
  * updated whenever fields of this class are added, renamed or removed.
+ *
+ * 2. Whenever adding more properties consider adding property validation to PropertyMappingValidator.
  */
 @JsonAdapter(PropertyMappingJsonAdapterFactory.class)
 public class PropertyMapping extends AbstractTypeMapping {
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/lowlevel/index/mapping/impl/PropertyMapping.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/lowlevel/index/mapping/impl/PropertyMapping.java
@@ -86,6 +86,52 @@ public class PropertyMapping extends AbstractTypeMapping {
 	@SerializedName("term_vector")
 	private String termVector;
 
+	/*
+	 * Dense vector datatype
+	 * https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html
+	 */
+	@SerializedName("element_type")
+	private String elementType;
+
+	/*
+	 * Dense vector datatype
+	 * https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html
+	 */
+	private Integer dims;
+
+	/*
+	 * Dense vector datatype
+	 * https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html
+	 */
+	private String similarity;
+
+	/*
+	 * Dense vector datatype
+	 * https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html
+	 */
+	@SerializedName("index_options")
+	private JsonElement indexOptions;
+
+	/*
+	 * k-NN vector datatype
+	 * https://opensearch.org/docs/latest/field-types/supported-field-types/knn-vector/
+	 */
+	private Integer dimension;
+
+	/*
+	 * k-NN vector datatype
+	 * https://opensearch.org/docs/latest/field-types/supported-field-types/knn-vector/
+	 */
+	private JsonElement method;
+
+	/*
+	 * k-NN vector datatype
+	 * https://opensearch.org/docs/latest/field-types/supported-field-types/knn-vector/
+	 */
+	@SerializedName("data_type")
+	private String dataType;
+
+
 	public String getType() {
 		return type;
 	}
@@ -172,5 +218,61 @@ public class PropertyMapping extends AbstractTypeMapping {
 
 	public void setTermVector(String termVector) {
 		this.termVector = termVector;
+	}
+
+	public String getElementType() {
+		return elementType;
+	}
+
+	public void setElementType(String elementType) {
+		this.elementType = elementType;
+	}
+
+	public Integer getDims() {
+		return dims;
+	}
+
+	public void setDims(Integer dims) {
+		this.dims = dims;
+	}
+
+	public String getSimilarity() {
+		return similarity;
+	}
+
+	public void setSimilarity(String similarity) {
+		this.similarity = similarity;
+	}
+
+	public JsonElement getIndexOptions() {
+		return indexOptions;
+	}
+
+	public void setIndexOptions(JsonElement indexOptions) {
+		this.indexOptions = indexOptions;
+	}
+
+	public Integer getDimension() {
+		return dimension;
+	}
+
+	public void setDimension(Integer dimension) {
+		this.dimension = dimension;
+	}
+
+	public JsonElement getMethod() {
+		return method;
+	}
+
+	public void setMethod(JsonElement method) {
+		this.method = method;
+	}
+
+	public String getDataType() {
+		return dataType;
+	}
+
+	public void setDataType(String dataType) {
+		this.dataType = dataType;
 	}
 }
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/lowlevel/index/mapping/impl/PropertyMapping.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/lowlevel/index/mapping/impl/PropertyMapping.java
@@ -113,7 +113,7 @@ public class PropertyMapping extends AbstractTypeMapping {
 	 * https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html
 	 */
 	@SerializedName("index_options")
-	private JsonElement indexOptions;
+	private ElasticsearchDenseVectorIndexOptions indexOptions;
 
 	/*
 	 * k-NN vector datatype
@@ -125,7 +125,7 @@ public class PropertyMapping extends AbstractTypeMapping {
 	 * k-NN vector datatype
 	 * https://opensearch.org/docs/latest/field-types/supported-field-types/knn-vector/
 	 */
-	private JsonElement method;
+	private OpenSearchVectorTypeMethod method;
 
 	/*
 	 * k-NN vector datatype
@@ -247,11 +247,11 @@ public class PropertyMapping extends AbstractTypeMapping {
 		this.similarity = similarity;
 	}
 
-	public JsonElement getIndexOptions() {
+	public ElasticsearchDenseVectorIndexOptions getIndexOptions() {
 		return indexOptions;
 	}
 
-	public void setIndexOptions(JsonElement indexOptions) {
+	public void setIndexOptions(ElasticsearchDenseVectorIndexOptions indexOptions) {
 		this.indexOptions = indexOptions;
 	}
 
@@ -263,11 +263,11 @@ public class PropertyMapping extends AbstractTypeMapping {
 		this.dimension = dimension;
 	}
 
-	public JsonElement getMethod() {
+	public OpenSearchVectorTypeMethod getMethod() {
 		return method;
 	}
 
-	public void setMethod(JsonElement method) {
+	public void setMethod(OpenSearchVectorTypeMethod method) {
 		this.method = method;
 	}
 
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/dsl/impl/ElasticsearchKnnPredicateFieldStep.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/dsl/impl/ElasticsearchKnnPredicateFieldStep.java
@@ -1,0 +1,41 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.elasticsearch.search.predicate.dsl.impl;
+
+import org.hibernate.search.backend.elasticsearch.search.predicate.dsl.ElasticsearchKnnPredicateOptionsStep;
+import org.hibernate.search.backend.elasticsearch.search.predicate.spi.ElasticsearchKnnPredicateBuilder;
+import org.hibernate.search.engine.search.predicate.dsl.SearchPredicateFactory;
+import org.hibernate.search.engine.search.predicate.dsl.spi.AbstractKnnPredicateFieldStep;
+import org.hibernate.search.engine.search.predicate.dsl.spi.SearchPredicateDslContext;
+import org.hibernate.search.engine.search.predicate.spi.PredicateTypeKeys;
+
+public final class ElasticsearchKnnPredicateFieldStep
+		extends AbstractKnnPredicateFieldStep<ElasticsearchKnnPredicateOptionsStep<?>, ElasticsearchKnnPredicateBuilder>
+		implements
+		ElasticsearchKnnPredicateOptionsStep<ElasticsearchKnnPredicateOptionsStep<?>> {
+
+	public ElasticsearchKnnPredicateFieldStep(SearchPredicateFactory factory, SearchPredicateDslContext<?> dslContext,
+			int k) {
+		super( factory, dslContext, k );
+	}
+
+	@Override
+	protected ElasticsearchKnnPredicateBuilder createBuilder(String fieldPath) {
+		return (ElasticsearchKnnPredicateBuilder) dslContext.scope().fieldQueryElement( fieldPath, PredicateTypeKeys.KNN );
+	}
+
+	@Override
+	public ElasticsearchKnnPredicateOptionsStep<?> numberOfCandidates(int numberOfCandidates) {
+		this.builder.numberOfCandidates( numberOfCandidates );
+		return this;
+	}
+
+	@Override
+	protected ElasticsearchKnnPredicateOptionsStep<?> thisAsT() {
+		return this;
+	}
+}
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchNestablePredicate.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchNestablePredicate.java
@@ -35,7 +35,6 @@ public abstract class AbstractElasticsearchNestablePredicate extends AbstractEla
 
 	@Override
 	public JsonObject toJsonQuery(PredicateRequestContext context) {
-		checkNestableWithin( context.getNestedPath() );
 
 		List<String> nestedPathHierarchy = getNestedPathHierarchy();
 		String expectedNestedPath = nestedPathHierarchy.isEmpty()
@@ -44,6 +43,7 @@ public abstract class AbstractElasticsearchNestablePredicate extends AbstractEla
 
 		if ( Objects.equals( context.getNestedPath(), expectedNestedPath ) ) {
 			// Implicit nesting is not necessary
+			checkNestableWithin( context.getNestedPath() );
 			return super.toJsonQuery( context );
 		}
 
@@ -52,6 +52,7 @@ public abstract class AbstractElasticsearchNestablePredicate extends AbstractEla
 		// so that the predicate is actually executed in this context.
 		PredicateRequestContext contextAfterImplicitNesting =
 				context.withNestedPath( expectedNestedPath );
+		checkNestableWithin( expectedNestedPath );
 
 		JsonObject result = super.toJsonQuery( contextAfterImplicitNesting );
 
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchNestablePredicate.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchNestablePredicate.java
@@ -25,7 +25,7 @@ public abstract class AbstractElasticsearchNestablePredicate extends AbstractEla
 
 
 	@Override
-	public void doCheckNestableWithin(PredicateNestingContext context) {
+	public void checkNestableWithin(PredicateNestingContext context) {
 		List<String> nestedPathHierarchy = getNestedPathHierarchy();
 		String expectedParentNestedPath = context.getNestedPath();
 
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchNestablePredicate.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchNestablePredicate.java
@@ -23,9 +23,11 @@ public abstract class AbstractElasticsearchNestablePredicate extends AbstractEla
 		super( builder );
 	}
 
+
 	@Override
-	public void checkNestableWithin(String expectedParentNestedPath) {
+	public void doCheckNestableWithin(PredicateNestingContext context) {
 		List<String> nestedPathHierarchy = getNestedPathHierarchy();
+		String expectedParentNestedPath = context.getNestedPath();
 
 		if ( expectedParentNestedPath != null && !nestedPathHierarchy.contains( expectedParentNestedPath ) ) {
 			throw log.invalidNestedObjectPathForPredicate( this, expectedParentNestedPath,
@@ -34,7 +36,7 @@ public abstract class AbstractElasticsearchNestablePredicate extends AbstractEla
 	}
 
 	@Override
-	public JsonObject buildJsonQuery(PredicateRequestContext context) {
+	public JsonObject toJsonQuery(PredicateRequestContext context) {
 		List<String> nestedPathHierarchy = getNestedPathHierarchy();
 		String expectedNestedPath = nestedPathHierarchy.isEmpty()
 				? null
@@ -42,8 +44,7 @@ public abstract class AbstractElasticsearchNestablePredicate extends AbstractEla
 
 		if ( Objects.equals( context.getNestedPath(), expectedNestedPath ) ) {
 			// Implicit nesting is not necessary
-			checkNestableWithin( context.getNestedPath() );
-			return super.buildJsonQuery( context );
+			return super.toJsonQuery( context );
 		}
 
 		// The context we expect this predicate to be built in.
@@ -51,9 +52,8 @@ public abstract class AbstractElasticsearchNestablePredicate extends AbstractEla
 		// so that the predicate is actually executed in this context.
 		PredicateRequestContext contextAfterImplicitNesting =
 				context.withNestedPath( expectedNestedPath );
-		checkNestableWithin( expectedNestedPath );
 
-		JsonObject result = super.buildJsonQuery( contextAfterImplicitNesting );
+		JsonObject result = super.toJsonQuery( contextAfterImplicitNesting );
 
 		// traversing the nestedPathHierarchy in the inverted order
 		int hierarchyLastIndex = nestedPathHierarchy.size() - 1;
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchNestablePredicate.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchNestablePredicate.java
@@ -34,8 +34,7 @@ public abstract class AbstractElasticsearchNestablePredicate extends AbstractEla
 	}
 
 	@Override
-	public JsonObject toJsonQuery(PredicateRequestContext context) {
-
+	public JsonObject buildJsonQuery(PredicateRequestContext context) {
 		List<String> nestedPathHierarchy = getNestedPathHierarchy();
 		String expectedNestedPath = nestedPathHierarchy.isEmpty()
 				? null
@@ -44,7 +43,7 @@ public abstract class AbstractElasticsearchNestablePredicate extends AbstractEla
 		if ( Objects.equals( context.getNestedPath(), expectedNestedPath ) ) {
 			// Implicit nesting is not necessary
 			checkNestableWithin( context.getNestedPath() );
-			return super.toJsonQuery( context );
+			return super.buildJsonQuery( context );
 		}
 
 		// The context we expect this predicate to be built in.
@@ -54,7 +53,7 @@ public abstract class AbstractElasticsearchNestablePredicate extends AbstractEla
 				context.withNestedPath( expectedNestedPath );
 		checkNestableWithin( expectedNestedPath );
 
-		JsonObject result = super.toJsonQuery( contextAfterImplicitNesting );
+		JsonObject result = super.buildJsonQuery( contextAfterImplicitNesting );
 
 		// traversing the nestedPathHierarchy in the inverted order
 		int hierarchyLastIndex = nestedPathHierarchy.size() - 1;
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchPredicate.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchPredicate.java
@@ -14,7 +14,7 @@ import org.hibernate.search.engine.search.predicate.spi.SearchPredicateBuilder;
 
 import com.google.gson.JsonObject;
 
-public abstract class AbstractElasticsearchPredicate implements ElasticsearchSearchPredicate {
+public abstract class AbstractElasticsearchPredicate extends ElasticsearchSearchPredicate {
 
 	private static final JsonAccessor<Float> BOOST_ACCESSOR = JsonAccessor.root().property( "boost" ).asFloat();
 
@@ -37,7 +37,7 @@ public abstract class AbstractElasticsearchPredicate implements ElasticsearchSea
 	}
 
 	@Override
-	public JsonObject toJsonQuery(PredicateRequestContext context) {
+	public JsonObject buildJsonQuery(PredicateRequestContext context) {
 		JsonObject outerObject = new JsonObject();
 		JsonObject innerObject = new JsonObject();
 
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchPredicate.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchPredicate.java
@@ -12,7 +12,6 @@ import org.hibernate.search.backend.elasticsearch.gson.impl.JsonAccessor;
 import org.hibernate.search.backend.elasticsearch.search.common.impl.ElasticsearchSearchIndexScope;
 import org.hibernate.search.engine.search.predicate.spi.SearchPredicateBuilder;
 
-import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 
 public abstract class AbstractElasticsearchPredicate implements ElasticsearchSearchPredicate {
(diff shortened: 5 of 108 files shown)


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/dialect/model/impl/OpenSearch2ModelDialect.java
```diff
@@ -0,0 +1,23 @@
...
+
+import org.hibernate.search.backend.elasticsearch.types.dsl.provider.impl.ElasticsearchIndexFieldTypeFactoryProvider;
+import org.hibernate.search.backend.elasticsearch.types.dsl.provider.impl.OpenSearch1IndexFieldTypeFactoryProvider;
+
+import com.google.gson.Gson;
+
+/**
+ * The model dialect for OpenSearch 2.x.
+ */
+public class OpenSearch2ModelDialect implements ElasticsearchModelDialect {
```
marko-bekhta, line 17: I'm not sure that we need `OpenSearch1` and `OpenSearch2` as they are identical, I've just followed the Elasticsearch example.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchKnnPredicate.java
```diff
@@ -0,0 +1,261 @@
...
+			//            "type": "illegal_argument_exception",
+			//            "reason": "Required [num_candidates]"
+			//          }
+			//        ],
+			//        "type": "illegal_argument_exception",
+			//        "reason": "Required [num_candidates]"
+			//      },
+			//      "status": 400
+			//    }
+			NUM_CANDIDATES_ACCESSOR.set( object, 100 );
```
marko-bekhta, line 177: this is a required filed  for a knn search in Elasticsearch and currently we do not read it. so I was wondering how should we do it... just add it as an optional property to the predicate or create some predicate that is only available through an extension ... (NOTE that OpenSearch does not need this attribute)

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/AccumulatingSourceExtractor.java
```diff
@@ -87,7 +87,7 @@ else if ( fieldValue.isJsonNull() ) {
 			// Present, but null
 			return accumulator.accumulate( accumulated, extract( projectionHitMapper, hit, fieldValue, context ) );
 		}
-		else if ( fieldValue.isJsonArray() ) {
+		else if ( !canDecodeArrays() && fieldValue.isJsonArray() ) {
```
marko-bekhta, line 90: Note this change, it is isolated in its own commit. problem here was that accumulator would start feeding vector elements to the codec that expects to get an entire array...

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/types/ByteVectorFieldTypeDescriptor.java
```diff
@@ -73,6 +74,13 @@ protected List<byte[]> createNonMatchingValues() {
 		);
 	}
 
+	@Override
+	public List<byte[]> unitLengthVectors() {
```
marko-bekhta, line 78: had to update how we were generating the sample data in both vector descriptors. Elastricsearch/OpenSearch both have more strict rules it seems and some of the values were not accepted with some of the similarity functions...

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/lowlevel/index/mapping/impl/PropertyMapping.java
```diff
@@ -86,6 +86,52 @@ public class PropertyMapping extends AbstractTypeMapping {
...
+	private JsonElement method;
+
+	/*
+	 * k-NN vector datatype
+	 * https://opensearch.org/docs/latest/field-types/supported-field-types/knn-vector/
+	 */
+	@SerializedName("data_type")
+	private String dataType;
+
+
```
yrodiere, line 134: If you add stuff here, I think you need to also add validation rules to `PropertyMappingValidator`. And maybe some tests :x

yrodiere: Might be worth adding another warning about this at the top of this file :/

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchPredicate.java
```diff
@@ -53,6 +77,10 @@ public JsonObject toJsonQuery(PredicateRequestContext context) {
 	protected abstract JsonObject doToJsonQuery(PredicateRequestContext context,
 			JsonObject outerObject, JsonObject innerObject);
 
+	protected JsonElement doToJsonKnn(PredicateRequestContext context) {
+		return null;
+	}
```
yrodiere, line 82: This is probably worth some comments... ?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchSearchPredicate.java
```diff
@@ -24,8 +25,14 @@ public interface ElasticsearchSearchPredicate extends SearchPredicate {
 
 	void checkNestableWithin(String expectedParentNestedPath);
 
+	default ElasticsearchSearchPredicate checkAcceptableAsBoolPredicateClause(String clauseType) {
+		return this;
+	}
+
 	JsonObject toJsonQuery(PredicateRequestContext context);
 
+	JsonElement toJsonKnn(PredicateRequestContext rootPredicateContext);
```
yrodiere, line 34: Ah, so it wasn't possible to just add an `addKnn` method to `PredicateRequestContext` as we discussed?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/projection/impl/AccumulatingSourceExtractor.java
yrodiere: This will probably lead to problems with arrays of arrays, but since vector fields cannot currently be multi-valued... +1 let's do this.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchSearchPredicate.java
```diff
@@ -24,8 +25,14 @@ public interface ElasticsearchSearchPredicate extends SearchPredicate {
 
 	void checkNestableWithin(String expectedParentNestedPath);
 
+	default ElasticsearchSearchPredicate checkAcceptableAsBoolPredicateClause(String clauseType) {
+		return this;
+	}
```
yrodiere, line 30: This is dangerous, as composite predicates (i.e. `bool`, `nested`) need to propagate the check to their children.

I'd recommend not having a default implementation here, or moving the default to a more specific class that can be reasonably assumed to be usable in any bool clause, e.g. `AbstractElasticsearchSingleFieldPredicate`.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchKnnPredicate.java
yrodiere: > and currently we do not read it

What do you mean? Read from where?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchKnnPredicate.java
```diff
@@ -0,0 +1,261 @@
...
+
+	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
+
+	protected final ElasticsearchSearchPredicate filter;
+	protected final int k;
+	protected final JsonArray vector;
+
+	private ElasticsearchKnnPredicate(AbstractKnnBuilder<?> builder) {
+		super( builder );
+		this.filter = builder.filter;
```
yrodiere, line 39: Dumb question maybe but... what happens if the filter itself is a KNN predicate?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchSearchPredicate.java
```diff
@@ -24,8 +25,14 @@ public interface ElasticsearchSearchPredicate extends SearchPredicate {
 
 	void checkNestableWithin(String expectedParentNestedPath);
 
+	default ElasticsearchSearchPredicate checkAcceptableAsBoolPredicateClause(String clauseType) {
```
yrodiere, line 28: Maybe use an enum rather than a String for the clause type?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/dialect/model/impl/OpenSearch2ModelDialect.java
yrodiere: I highly doubt that OpenSearch 1.x supports vector search, so you'll probably need to make them different, be it only to throw a clear error when people try to use vector search on older versions of OpenSearch/Elasticsearch.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/indexnull/IndexNullAsValueIT.java
```diff
@@ -74,6 +77,10 @@ void indexNullAsValue_spatial() {
 
 	@Test
 	void indexNullAsValue_vector() {
+		assumeTrue(
+				TckConfiguration.get().getBackendFeatures().supportsVectorSearchIndexNullsAs(),
+				"These tests only make sense for a backend where Vector Search is supported and implemented."
+		);
```
yrodiere, line 83: Maybe test that we get a clear error message when it's *not* supported?

Or just remove the `indexNullAs` feature for vectors :)
Or handle it on the client side.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/KnnPredicateSpecificsIT.java
```diff
@@ -476,7 +472,7 @@ private static class IndexBinding {
 
 			public IndexBinding(IndexSchemaElement root, Collection<? extends VectorFieldTypeDescriptor<?>> fieldTypes,
 					VectorSimilarity vectorSimilarity) {
-				field = SimpleFieldModelsByType.mapAll( fieldTypes, root, "", c -> c.vectorSimilarity( vectorSimilarity ) );
+				field = SimpleFieldModelsByType.mapAll( fieldTypes, root, "", c -> c.projectable( Projectable.YES ).vectorSimilarity( vectorSimilarity ) );
```
yrodiere, line 475: It would probably be best to use a different field to test projections. Otherwise you can't be sure knn works without `Projectable.YES`... We've had problems like that in the past, with `Sortable.YES` adding doc-values to the field, which turned out to be necessary for some predicates :x

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchKnnPredicate.java
marko-bekhta: I meant we do not expose a way to pass it from the search predicate (or through any other way) 😃. So should I just add it to `KnnPredicateOptionsStep` like this:
```java
public interface KnnPredicateOptionsStep extends PredicateScoreStep<KnnPredicateOptionsStep>, PredicateFinalStep {

	KnnPredicateOptionsStep filter(SearchPredicate searchPredicate);

	default KnnPredicateOptionsStep filter(PredicateFinalStep searchPredicate) {
		return filter( searchPredicate.toPredicate() );
	}

	KnnPredicateOptionsStep filter(Function<? super SearchPredicateFactory, ? extends PredicateFinalStep> clauseContributor);

	/**
	 * throws exceptions on OpenSearch and Lucene
	 */
	KnnPredicateOptionsStep numberOfCandidates(int numberOfCandidates);

}
```

yrodiere: Doesn't look like a great solution if it doesn't make any sense on OpenSearch/Lucene... I need to understand what the point of this parameter is. I'll have a look.

marko-bekhta: Here's a link if you haven't looked at it already:
https://www.elastic.co/guide/en/elasticsearch/reference/current/knn-search.html#tune-approximate-knn-for-speed-accuracy

I was thinking about setting it to something like `k/num-of-shards+1` 😃 in the background ..

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/dialect/model/impl/OpenSearch2ModelDialect.java
marko-bekhta: I hoped that it would work since there's a doc for it https://opensearch.org/docs/1.3/search-plugins/knn/knn-index in 1.3 version... But I've run a quick test and it results in:
```json
{
  "error": {
    "root_cause": [
      {
        "type": "mapper_parsing_exception",
        "reason": "Invalid engine: lucene"
      }
    ],
    "type": "mapper_parsing_exception",
    "reason": "Failed to parse mapping [_doc]: Invalid engine: lucene",
    "caused_by": {
      "type": "mapper_parsing_exception",
      "reason": "Invalid engine: lucene"
    }
  },
  "status": 400
}
```
even though there's an example in that doc 🤷🏻‍♂️😔:
```
{
    "type": "knn_vector",
    "dimension": 100,
    "method": {
        "name":"hnsw",
        "engine":"lucene",
        "space_type": "l2",
        "parameters":{
            "m":2048,
            "ef_construction": 245
        }
    }
}
```
Then, on the Elasticsearch side, from what I see, this API has been available as a preview since 8.0

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/types/ByteVectorFieldTypeDescriptor.java
marko-bekhta: created https://hibernate.atlassian.net/browse/HSEARCH-5048

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchKnnPredicate.java
marko-bekhta: I've tried 
```json
.where( f -> f.knn( 3 )
	.field( "location" )
	.matching( 5f, 4f )
	.filter(
			f.knn( 6 )
					.field( "location" )
					.matching( 6f, 3f )
	)
```
And it seems to work correctly, it "finds 6 close to 6-3 and then finds 3 close to 5-4 among those 6 ones...

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchPredicate.java
marko-bekhta: not anymore 😄 (see commit with message `HSEARCH-4950 Hide knn clauses in a predicate request context`)

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchSearchPredicate.java
marko-bekhta: I wasn't sure how it all will play out so I went with the most straightforward thing. I've moved things around and this method is gone now.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/indexnull/IndexNullAsValueIT.java
marko-bekhta: added null-indexing into the codec.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchKnnPredicate.java
```diff
@@ -0,0 +1,250 @@
...
+			JsonObject knn = new JsonObject();
+			FIELD_ACCESSOR.set( knn, absoluteFieldPath );
+			K_ACCESSOR.set( knn, k );
+			if ( filter != null ) {
+				FILTER_ACCESSOR.set( knn, filter.toJsonQuery( context ) );
+			}
+			NUM_CANDIDATES_ACCESSOR.set( knn, numberOfCandidates != null ? numberOfCandidates : k );
+			QUERY_VECTOR_ACCESSOR.set( knn, vector );
+
+			context.contributeKnnClause( knn );
```
marko-bekhta, line 170: One of the missing things is a check for where knn predicate is added i.e. is it a top-level predicate/should clause of a top-level bool predicate/a clause/filter of some predicate that itself is a clause to some other predicate ...
I don't think we have anything to determine at which level `toJsonQuery` is called,so I was thinking should we add a parent to `PredicateRequestContext` and create  a new context setting current one as a parent any time we call toJsonQuery, so that the context is a tree node and we can decide if the location of a knn predicate is acceptable?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/dialect/model/impl/OpenSearch2ModelDialect.java
yrodiere: IMO you shouldn't bother too much with older releases. If it takes longer than just implementing clear "Feature not available on this version" error messages, your time is better invested elsewhere :) It is an incubating feature, after all.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchKnnPredicate.java
yrodiere: Do I understand correctly it returns 6 documents?

I'd expect this query to only return 3 documents, since the top-level one has k=3.

And if it does return 3 documents with your current implementation... does that mean Elasticsearch executes a conjunction of all knn predicates? If so, we need to reconsider how we require knn predicates to be structured in the Search DSL.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchKnnPredicate.java
yrodiere: >  I don't think we have anything to determine at which level toJsonQuery is called,so I was thinking should we add a parent to PredicateRequestContext and create a new context setting current one as a parent any time we call toJsonQuery, so that the context is a tree node and we can decide if the location of a knn predicate is acceptable?

+1 makes sense, but be careful: we already create new contexts right now, to handle nested contexts. That one could have been modeled as a tree too, but it hasn't. You'll have to make sure things work, and ideally make it a bit more consistent.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchKnnPredicate.java
marko-bekhta: no no, I should've posted a complete example: https://github.com/hibernate/hibernate-search/pull/3871/files#diff-03164f33f7d3eedeb9d14e03be09490770e107dfe30661215c26dd3c3ebd857cR648-R667
what I meant was that the filter removes some elements and then the top knn finds 3 "matches"  (among those that inner knn kept). so the result is 3 elements. 
For Elasticsearch ... If I'd execute:
```json
{
  "knn": [
    {
      "field": "location",
      "k": 6,
      "num_candidates": 6,
      "query_vector": [
        6.0,
        3.0
      ]
    },
    {
      "field": "location",
      "k": 3,
      "num_candidates": 3,
      "query_vector": [
        5.0,
        4.0
      ]
    }
  ],
  "_source": false
}
```
I'd get 6 docs: `["ID:2", "ID:3", "ID:1", "ID:6", "ID:12", "ID:4"]`, so it's a disjunction for Elasticsearch. And it seems "fine" the way we are allowing them to be added as should clauses of a top level bool. 
As for this particular scenario I'd say  we just need to not allow a knn as a filter if it's an Elasticsearch distribution (which should be doable with that other comment on predicate context in mind 😃 🙈 )

yrodiere: > And it seems "fine" the way we are allowing them to be added as should clauses of a top level bool.

To be clear the current behavior on Elasticsearch is definitely not fine, because the results are incorrect. There should only be 3.

It would be fine if that case was simply forbidden in Elasticsearch, because then users wouldn't experience this (incorrect) result.

>  As for this particular scenario I'd say we just need to not allow a knn as a filter if it's an Elasticsearch distribution (which should be doable with that other comment on predicate context in mind 😃 🙈 )

Agreed. A "filter" clause is not a "should" clause, after all. This would only be consistent with other constraints in boolean predicates.

marko-bekhta: Sorry 🙈 so the example with Elasticsearch is just adding two knn predicates (if we want to do that in Search it's through two should clauses). if I'd try to actually put knn as a filter for Elasticsearch it's an exception. 
I was just saying that adding multiple knn clauses for Elasticsearch is a disjunction

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/AbstractElasticsearchNestablePredicate.java
```diff
@@ -34,26 +34,26 @@ public void checkNestableWithin(String expectedParentNestedPath) {
...
+			checkNestableWithin( context.getNestedPath() );
+			return super.buildJsonQuery( context );
 		}
 
 		// The context we expect this predicate to be built in.
 		// We'll make sure to wrap it in nested predicates as appropriate in the next few lines,
 		// so that the predicate is actually executed in this context.
 		PredicateRequestContext contextAfterImplicitNesting =
 				context.withNestedPath( expectedNestedPath );
+		checkNestableWithin( expectedNestedPath );
```
marko-bekhta, line 54: btw, note this change: previously we'd just assume that the nested path would be ok for the predicate without double-checking.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchSearchPredicate.java
```diff
@@ -16,17 +16,21 @@
...
 
-	void checkNestableWithin(String expectedParentNestedPath);
+	public abstract void checkNestableWithin(String expectedParentNestedPath);
 
-	JsonObject toJsonQuery(PredicateRequestContext context);
+	public final JsonObject toJsonQuery(PredicateRequestContext context) {
+		return buildJsonQuery( context.predicateContext( this ) );
+	}
+
+	public abstract JsonObject buildJsonQuery(PredicateRequestContext context);
```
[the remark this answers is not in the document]
marko-bekhta: hmm 😃. My main goal here with this change was to make sure that the context is always created for us and we do not need to think when to create a new one...

> Pushed a bit further, if you called checkNestableWithin whenever a predicate is added to e.g. a boolean predicate clause, this could allow you to throw errors earlier

Wouldn't there be a problem if someone does something like:
```java
var p1 = f.bool().should(f.knn()..); // all looks fine and should would accept the knn
var p2 = f.bool().must(f.match()...).filter(p1); // this should now fail as it would result in a knn predicate being in the filter. But that would mean that we'd need to check p1 is "nestable" and then recursively if any predicates that are nested within p1 are nestable within this new context...
```

I've added a couple of tests for knn in sort/aggregation filters and they are failing (not allow to add knn) right now since Elasticsearch knn predicate just throws exceptions on any nonnull path in `checkNestableWithin` and these filters want to be applied to a field of a nested object.

But then .. I'm probably missing something here 😃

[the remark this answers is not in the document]
yrodiere: > Wouldn't there be a problem if someone does something like:
> 
> ```java
> var p1 = f.bool().should(f.knn()..); // all looks fine and should would accept the knn
> var p2 = f.bool().must(f.match()...).filter(p1); // this should now fail as it would result in a knn predicate being in the filter. But that would mean that we'd need to check p1 is "nestable" and then recursively if any predicates that are nested within p1 are nestable within this new context...
> ```

Why would this be a problem? `ElasticsearchBooleanPredicate` does indeed [check "nestability" recursively](https://github.com/hibernate/hibernate-search/blob/b19562cea319cf90524448f413e4d019b519fd59/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchBooleanPredicate.java#L66-L72). It always has.

As I mentioned there would be an impact on performance, but I think it's acceptable provided the leaf checks are quick enough.


> I've added a couple of tests for knn in sort/aggregation filters and they are failing (not allow to add knn) right now since Elasticsearch knn predicate just throws exceptions on any nonnull path in checkNestableWithin and these filters want to be applied to a field of a nested object.

These sort/aggregations filters are specifically about nested fields, and nested fields are not supported for knn. So we expect those tests to fail. So all is good, I'm not sure I see the problem?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/dsl/impl/ElasticsearchKnnPredicateFieldStep.java
```diff
@@ -9,16 +9,16 @@
 import org.hibernate.search.backend.elasticsearch.search.predicate.dsl.ElasticsearchKnnPredicateOptionsStep;
 import org.hibernate.search.backend.elasticsearch.search.predicate.spi.ElasticsearchKnnPredicateBuilder;
 import org.hibernate.search.engine.search.predicate.dsl.SearchPredicateFactory;
-import org.hibernate.search.engine.search.predicate.dsl.impl.AbstractKnnPredicateFieldStep;
+import org.hibernate.search.engine.search.predicate.dsl.spi.AbstractKnnPredicateFieldStep;
 import org.hibernate.search.engine.search.predicate.dsl.spi.SearchPredicateDslContext;
 import org.hibernate.search.engine.search.predicate.spi.PredicateTypeKeys;
 
-public final class ElasticsearchKnnPredicateFieldStepImpl
+public final class ElasticsearchKnnPredicateFieldStep
```
marko-bekhta, line 16: this seems to make sense, but I didn't want it to sneak in so I've made these changes in a temporary commit so you can see them isolated 😄

documentation/src/test/java/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java
```diff
@@ -1141,6 +1149,49 @@ void knn() {
...
+							.should( f.knn( 10 ).field( "coverImageEmbeddings" ).matching( coverImageEmbeddingsVector ) )
+							.should( f.knn( 5 ).field( "alternativeCoverImageEmbeddings" )
+									.matching( alternativeCoverImageEmbeddingsVector ) )
+					)
+					.fetchHits( 20 );
+			// end::knn-should[]
+			assertThat( hits )
+					.extracting( Book::getId )
+					.containsExactlyInAnyOrder( BOOK1_ID, BOOK2_ID, BOOK3_ID, BOOK4_ID );
+		} );
```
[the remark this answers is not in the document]
marko-bekhta: I've kept this one, as it is part of the section about limitations of Elasticsearch distribution, but I've also added an example on how to combine knn and regular text predicates.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/dsl/ElasticsearchKnnPredicateFieldStep.java
```diff
@@ -0,0 +1,17 @@
...
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.elasticsearch.search.predicate.dsl;
+
+import org.hibernate.search.engine.search.predicate.dsl.KnnPredicateFieldStep;
+
+/**
+ * The initial step in a "knn" predicate definition, where the target field can be set.
+ */
+public interface ElasticsearchKnnPredicateFieldStep<T extends ElasticsearchKnnPredicateVectorStep<?>>
```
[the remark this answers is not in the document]
marko-bekhta: Yes, and removing these interfaces worked out for the better. (I think) 😄

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchSearchPredicate.java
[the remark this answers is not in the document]
marko-bekhta: > These sort/aggregations filters are specifically about nested fields, and nested fields are not supported for knn. So we expect those tests to fail. So all is good, I'm not sure I see the problem?

No, there's no problem with that. I was just saying that these cases happened to work as you'd expected them to work purely by chance 😃 

I've removed the `checkNestableWithin` from the building json query and made some more changes around it in the last commit, let me know if that works for you 😃

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/validation/impl/PropertyMappingValidator.java
```diff
@@ -115,4 +120,131 @@ private void validateIndexOptions(ValidationErrorCollector errorCollector, Prope
...
+	static class Elasticsearch7PropertyMappingValidator extends PropertyMappingValidator {
+
+		@Override
+		protected void validateVectorMapping(ValidationErrorCollector errorCollector, PropertyMapping expectedMapping,
+				PropertyMapping actualMapping) {
+
+		}
+	}
+
+	static class Elasticsearch8PropertyMappingValidator extends PropertyMappingValidator {
```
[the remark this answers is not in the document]
marko-bekhta: ok, this validation and coverage is still on my todo list....  and regarding this validation -- yes, it shouldn't cause any problems to have it all together, but then it was easier to see what fields are for what backend, I suppose I could've just added comments and grouped them like that... but I liked it this way 🙈 😃

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/dsl/impl/ElasticsearchKnnPredicateFieldStep.java
yrodiere: If jqassistant is happy, I'm happy: feel free to squash this.

backend/elasticsearch/src/test/java/org/hibernate/search/backend/elasticsearch/gson/spi/GsonClassesTest.java
```diff
@@ -80,6 +80,9 @@ void testNoMissingGsonContractImplementations() {
 			for ( ClassInfo implementor : backendElasticsearchIndex.getAllKnownSubclasses( gsonContract ) ) {
 				classes.add( implementor.name() );
 			}
+			for ( ClassInfo implementor : backendElasticsearchIndex.getAllKnownImplementors( gsonContract ) ) {
```
marko-bekhta, line 83: Since `TypeAdapterFactory` is an interface, none of those factories were actually added to the `classes` set 🙈

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/validation/impl/PropertyMappingValidator.java
```diff
@@ -115,4 +119,150 @@ private void validateIndexOptions(ValidationErrorCollector errorCollector, Prope
...
+			);
+
+			OpenSearchVectorTypeMethod expectedMethod = expectedMapping.getMethod();
+			if ( expectedMethod != null ) {
+				methodValidator.validate( errorCollector, expectedMethod, actualMapping.getMethod() );
+			}
+		}
+	}
+
+	private static class ElasticsearchDenseVectorIndexOptionsValidator
```
marko-bekhta, line 198: @yrodiere WDYT?

current:
```
Stub context: 
        failures: 
          - Validation of the existing index in the Elasticsearch cluster failed. See below for details.
        field 'vectorB': 
            attribute 'index_options.m': 
                failures: 
                  - Invalid value. Expected '20', actual is '30'
            attribute 'index_options.ef_construction': 
                failures: 
                  - Invalid value. Expected '2', actual is '3'
        field 'vectorF': 
            attribute 'index_options.m': 
                failures: 
                  - Invalid value. Expected '50', actual is '60'
            attribute 'index_options.ef_construction': 
                failures: 
                  - Invalid value. Expected '5', actual is '6'
```
vs pushing another element to `errorCollector`:
```
    Stub context: 
        failures: 
          - Validation of the existing index in the Elasticsearch cluster failed. See below for details.
        field 'vectorB': 
            attribute 'index_options': 
                attribute 'm': 
                    failures: 
                      - Invalid value. Expected '20', actual is '30'
                attribute 'ef_construction': 
                    failures: 
                      - Invalid value. Expected '2', actual is '3'
        field 'vectorF': 
            attribute 'index_options': 
                attribute 'm': 
                    failures: 
                      - Invalid value. Expected '50', actual is '60'
                attribute 'ef_construction': 
                    failures: 
                      - Invalid value. Expected '5', actual is '6'
```

documentation/src/main/asciidoc/public/reference/_search-dsl-predicate.adoc
```diff
@@ -1504,6 +1504,56 @@ include::{sourcedir}/org/hibernate/search/documentation/search/predicate/Predica
...
+
+[[search-dsl-predicate-knn-limitations]]
+=== Backend specifics and limitations
+
+With the Elasticsearch backend a `knn` predicate can only be added as a top-level predicate,
+i.e. a predicate directly passed to a where clause of a search query,
+or as part of a top-level disjunction,
+i.e. as `should` clauses of a top-level <<search-dsl-predicate-boolean,`boolean` predicate>>
+(the same can be achieved by using an <<search-dsl-predicate-or,`or` predicate>>).
+Any other usages of a `knn` predicate, with this backend, would lead to an exception being thrown.
```
yrodiere, line 1529: "Elasticsearch backend" can technically be either Elasticsearch or OpenSearch. Try this:

```suggestion
With the Elasticsearch backend and when using <<backend-elasticsearch-compatibility-elasticsearch,Elasticsearch>>,
 a `knn` predicate can only be added as a top-level predicate,
i.e. a predicate directly passed to a where clause of a search query,
or as part of a top-level disjunction,
i.e. as `should` clauses of a top-level <<search-dsl-predicate-boolean,`boolean` predicate>>
(the same can be achieved by using an <<search-dsl-predicate-or,`or` predicate>>).
Any other usages of a `knn` predicate, with this backend, would lead to an exception being thrown.

<<backend-elasticsearch-compatibility-opensearch,OpenSearch>> does not have this limitation.
```

I'd also suggest to update the sections I linked to, to list this limitation and add a link back to this section.

documentation/src/main/asciidoc/public/reference/_search-dsl-predicate.adoc
```diff
@@ -1504,6 +1504,56 @@ include::{sourcedir}/org/hibernate/search/documentation/search/predicate/Predica
...
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java[tags=knn-should]
+----
+====
+
+The Elasticsearch backend also allows configuring a backend-specific `knn` predicate option: the number of candidates.
+This option specifies a number of approximate nearest neighbor candidates to be found on each shard,
+then the results from each shard are merged and the top `k` are selected.
+When not specified explicitly, Hibernate Search will default the number of candidates value to `k`.
+See the Elasticsearch link:{elasticsearchDocUrl}/knn-search.html#tune-approximate-knn-for-speed-accuracy[documentation] for more details.
```
yrodiere, line 1543:
```suggestion
The Elasticsearch backend, when using <<backend-elasticsearch-compatibility-elasticsearch,Elasticsearch>>,
also allows configuring a backend-specific `knn` predicate option: the number of candidates.
This option specifies a number of approximate nearest neighbor candidates to be found on each shard,
then the results from each shard are merged and the top `k` are selected.
When not specified explicitly, Hibernate Search will default the number of candidates value to `k`.
See the Elasticsearch link:{elasticsearchDocUrl}/knn-search.html#tune-approximate-knn-for-speed-accuracy[documentation] for more details.

<<backend-elasticsearch-compatibility-opensearch,OpenSearch>> does not expose this option.

```

documentation/src/test/java/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java
```diff
@@ -1106,7 +1113,8 @@ void not() {
 
 	@Test
 	void knn() {
-		// TODO HSEARCH-4950 Remove this assumption when we support KNN search using Elasticsearch/OpenSearch
+		// NOTE: To keep this documentation example simple there is no testing with Elasticsearch/OpenSearch
+		// as not all versions have integration implemented e.g. Elasticsearch 7 or OpenSearch 1.3 will throw exceptions
```
yrodiere, line 1117: I think you can call `org.hibernate.search.util.impl.integrationtest.backend.elasticsearch.dialect.ElasticsearchTestDialect#isActualVersion(...)` from here, if it helps?

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/util/ElasticsearchTckBackendFeatures.java
```diff
@@ -248,6 +248,28 @@ public boolean supportsExplicitRefresh() {
 
 	@Override
 	public boolean supportsVectorSearch() {
-		return false;
+		return isActualVersion(
+				esVersion -> !esVersion.isLessThan( "8.0" ),
+				osVersion -> !osVersion.isLessThan( "2.0" ),
+				osVersion -> true
```
yrodiere, line 254:
```suggestion
				es -> !es.isLessThan( "8.0" ),
				os -> !os.isLessThan( "2.0" ),
				aoss -> true
```

Also, you're sure Amazon OpenSearch Serverless supports vector search?

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/util/ElasticsearchTckBackendFeatures.java
```diff
@@ -248,6 +248,28 @@ public boolean supportsExplicitRefresh() {
...
+		);
+	}
+
+	@Override
+	public boolean supportsVectorSearchInsideOtherPredicates() {
+		return isActualVersion(
+				es -> false,
+				os -> true,
+				aoss -> false
```
yrodiere, line 263: I find it odd the behavior of OpenSearch and Amazon OpenSearch Serverless is different here?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchSearchPredicate.java
```diff
@@ -22,13 +22,13 @@ public abstract class ElasticsearchSearchPredicate implements SearchPredicate {
 
 	public abstract Set<String> indexNames();
 
-	public abstract void checkNestableWithin(String expectedParentNestedPath);
-
-	public final JsonObject toJsonQuery(PredicateRequestContext context) {
-		return buildJsonQuery( context.predicateContext( this ) );
+	public final void checkNestableWithin(PredicateNestingContext context) {
+		doCheckNestableWithin( context.wrap( this ) );
```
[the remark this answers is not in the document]
marko-bekhta: Consider this case:
```java
// all good we do not call any `checkNestableWithin` as this knn is at the very top level:
SearchPredicate knn = scope.predicate().knn( 15 ).field( "location" ).matching( 50.0f, 50.0f ).toPredicate();
// here we add it in a should clause so we know it can accept the knn 
// and we create a `PredicateNestingContext.acceptsKnn()` context:
SearchPredicate boolKnnInShould = scope.predicate().bool().should( knn )
		.should( scope.predicate().match().field( "parking" ).matching( Boolean.TRUE ) ).toPredicate();
// now here ...  
// the first should will create a `PredicateNestingContext.acceptsKnn()` context. 
// i.e it will receive a boolean predicate on which we are calling `imaBoolPredicateWithKnnShouldClause.checkNestableWithin( PredicateNestingContext.acceptsKnn() );`
// so this first call is ok, but then any inner predicates of this `imaBoolPredicateWithKnnShouldClause` 
// should not allow knn preidcates, right ? soo when this is executed:
// 	@Override
// 	public void doCheckNestableWithin(PredicateNestingContext context) {
// 		checkAcceptableWithin( context, mustClauses );
//             // context here now should not allow accepting knn predicates, so it should somehow be updated... 
// 		checkAcceptableWithin( context, shouldClauses );
// 		checkAcceptableWithin( context, filterClauses );
// 		checkAcceptableWithin( context, mustNotClauses );
// 	}
scope.predicate().bool().should( boolKnnInShould );
```

we probably could try replacing `doCheckNestableWithin` in a bool predicate with something like
```java
@Override
public void doCheckNestableWithin(PredicateNestingContext context) {
	PredicateNestingContext updatedContext = PredicateNestingContext.doesNotAcceptKnn( context.getNestedPath() );
	checkAcceptableWithin( updatedContext, mustClauses );
	checkAcceptableWithin( updatedContext, shouldClauses );
	checkAcceptableWithin( updatedContext, filterClauses );
	checkAcceptableWithin( updatedContext, mustNotClauses );
}
```
but then that wrap is doing it for us already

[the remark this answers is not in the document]
marko-bekhta: you can check this test to see how it behaves 
https://github.com/hibernate/hibernate-search/blob/cbda56b8ab9a36637d1ae8210e2d11228a3bbf5d/integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/search/ElasticsearchKnnPredicateSpecificsIT.java#L182-L209

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/util/ElasticsearchTckBackendFeatures.java
marko-bekhta: I cannot see it here https://docs.aws.amazon.com/opensearch-service/latest/developerguide/serverless-genref.html
but then there's https://aws.amazon.com/opensearch-service/serverless-vector-engine/

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchSearchPredicate.java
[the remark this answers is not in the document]
yrodiere: I'd actually have expected your specific example (a `knn` within a `should` within a `should`) to work without throwing an exception... After all, `X OR (Y OR Knn)` is the same as `Knn OR (X OR Y)`.

Though I see the documentation does mention something about "top-level"... any reason to that? Score when `must` clauses are added to the mix, maybe?

I mean if it's possible to support knn predicates in nested should clauses *and* it simplifies implementation, I'd definitely be in favor.
And personally even if we don't want to support it, I'd favor a solution that doesn't involve double dispatch if it's not much more complicated.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/search/ElasticsearchKnnPredicateSpecificsIT.java
```diff
@@ -183,29 +184,64 @@ void knnPredicateInWrongPlace_addingPrebuiltKnn() {
...
-		knnPredicateInWrongPlace( () -> scope.predicate().bool().must( boolKnnInShould ) );
-		knnPredicateInWrongPlace( () -> scope.predicate().bool().mustNot( boolKnnInShould ) );
-		knnPredicateInWrongPlace( () -> scope.predicate().bool().filter( boolKnnInShould ) );
-
-		// adding as a knn filter:
-		knnPredicateInWrongPlace(
-				() -> scope.predicate().knn( 10 ).field( "location" ).matching( 50.0f, 50.0f ).filter( knn ) );
-		knnPredicateInWrongPlace(
-				() -> scope.predicate().knn( 10 ).field( "location" ).matching( 50.0f, 50.0f ).filter( boolKnnInShould ) );
+		SearchPredicate inlineKnn = scope.predicate().knn( 15 ).field( "location" ).matching( 50.0f, 50.0f ).toPredicate();
```
marko-bekhta, line 187: Things weren't going too terrible, but then I thought that we could pass a knn predicate as  a named predicate ... 🙈 
I've pushed what I have done... but maybe you have an idea for something more simple?

marko-bekhta: No worries 😃. I've made it so all these tests pass, but I think you may not like the way it is done 😄. 
BTW speaking of a named predicate, I was wondering if we should just remove the predicate and have the `ElasticsearchNamedPredicate$Builder` just return the `providedPredicate`. But let's discuss that later 😃

documentation/src/main/asciidoc/public/reference/_search-dsl-predicate.adoc
```diff
@@ -856,6 +856,13 @@ but can be <<search-dsl-predicate-common-constantScore,made constant with `.cons
 * The score of an `and` predicate can be <<search-dsl-predicate-common-boost,boosted>>
 with a call to `.boost(...)`.
 
+[[search-dsl-predicate-and-limitations]]
+=== Limitations
+
+Keep in mind that adding a <<search-dsl-predicate-knn,knn>> as an `and` clause will lead to an exception when using
+<<backend-elasticsearch-compatibility-elasticsearch,an Elastic distribution of Elasticsearch>>.
+See <<search-dsl-predicate-knn-limitations,`knn` limitations>> for more details.
```
yrodiere, line 864: I think there was a misunderstanding, I meant you should add the limitation to `backend-elasticsearch-compatibility-elasticsearch`.

You certainly can keep this, though, it doesn't hurt.

yrodiere (approved):
I don't really see where you handled these new checks, but if all tests pass... alright, let's merge :) I've been nitpicking long enough.
Thanks for working on this @marko-bekhta! Please merge as soon  as CI passes.

documentation/src/main/asciidoc/public/reference/_search-dsl-predicate.adoc
marko-bekhta: added this paragraph:
https://github.com/hibernate/hibernate-search/blob/764d121153449680246cf7b91f677a3db30982a8/documentation/src/main/asciidoc/public/reference/_backend-elasticsearch.adoc?plain=1#L50-L53

marko-bekhta:
Thanks @yrodiere !!! 😃 I'll merge this in and then if we find anything needs adjusting I'll create a follow-up PR

PR #3876, opened 2023-12-15
HSEARCH-5052 Filter out documents with vectors below a "similarity limit"

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5052

Hey Yoann 😃 
You've probably seen the discussion we had yesterday on Zulip here: https://infinispan.zulipchat.com/#narrow/stream/118645-infinispan/topic/vector.20search/near/407994976 
I thought I'd give it a try and see what can we do to filter things out on our side; I've looked into creating custom queries and that's what I've put together. Do you think this could be something useful to add?

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/query/impl/VectorSimilarityFilterQuery.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/query/impl/VectorSimilarityFilterQuery.java
@@ -1,0 +1,222 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.backend.lucene.lowlevel.query.impl;
+
+import java.io.IOException;
+import java.util.Objects;
+
+import org.hibernate.search.util.common.AssertionFailure;
+
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.index.VectorSimilarityFunction;
+import org.apache.lucene.search.DocIdSetIterator;
+import org.apache.lucene.search.Explanation;
+import org.apache.lucene.search.FilterWeight;
+import org.apache.lucene.search.IndexSearcher;
+import org.apache.lucene.search.KnnByteVectorQuery;
+import org.apache.lucene.search.KnnFloatVectorQuery;
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.QueryVisitor;
+import org.apache.lucene.search.ScoreMode;
+import org.apache.lucene.search.Scorer;
+import org.apache.lucene.search.TwoPhaseIterator;
+import org.apache.lucene.search.Weight;
+import org.apache.lucene.util.VectorUtil;
+
+public class VectorSimilarityFilterQuery extends Query {
+
+	private final Query query;
+	private final float similarityAsScore;
+
+	public static VectorSimilarityFilterQuery create(KnnByteVectorQuery query, float similarityLimit, int dimension,
+			VectorSimilarityFunction vectorSimilarityFunction) {
+		// We assume that `similarityLimit` is a distance so we need to convert it to the score using a formula from a
+		// similarity function:
+		return new VectorSimilarityFilterQuery(
+				query, byteSimilarityDistanceToScore( similarityLimit, dimension, vectorSimilarityFunction ) );
+	}
+
+	public static VectorSimilarityFilterQuery create(KnnFloatVectorQuery query, float similarityLimit,
+			VectorSimilarityFunction vectorSimilarityFunction) {
+		// We assume that `similarityLimit` is a distance so we need to convert it to the score using a formula from a
+		// similarity function:
+		return new VectorSimilarityFilterQuery(
+				query, floatSimilarityDistanceToScore( similarityLimit, vectorSimilarityFunction ) );
+	}
+
+	private VectorSimilarityFilterQuery(Query query, float similarityAsScore) {
+		this.query = query;
+		this.similarityAsScore = similarityAsScore;
+	}
+
+	@Override
+	public Query rewrite(IndexSearcher indexSearcher) throws IOException {
+		Query rewritten = query.rewrite( indexSearcher );
+		if ( rewritten == query ) {
+			return this;
+		}
+		// Knn queries are rewritten and we need to use a rewritten one to get the weights and scores:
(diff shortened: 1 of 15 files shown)


integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/KnnPredicateSpecificsIT.java
```diff
@@ -698,6 +699,53 @@ void nestedVector() {
...
 
+		@ParameterizedTest
+		@MethodSource
+		void similarityFilter(float similarity, float score, int matches) {
+			assertThat( index.createScope().query()
+					.select( SearchProjectionFactory::score )
+					.where( f -> f.knn( 10 )
+							.field( "location" )
+							.matching( 5f, 4f )
+							.similarity( similarity )
```
yrodiere, line 710: `minSimilarity`, maybe? Or `maxDistance`, IDK how it's supposed to be used.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/predicate/impl/LuceneKnnPredicate.java
```diff
@@ -31,23 +33,29 @@ public class LuceneKnnPredicate extends AbstractLuceneSingleFieldPredicate imple
...
+		this.similarityFunction = builder.similarityFunction;
 	}
 
 	@Override
 	protected Query doToQuery(PredicateRequestContext context) {
 		if ( vector instanceof byte[] ) {
-			return new KnnByteVectorQuery(
+			KnnByteVectorQuery query = new KnnByteVectorQuery(
 					absoluteFieldPath, (byte[]) vector, k, filter == null ? null : filter.toQuery( context ) );
+			return similarity == null ? query : VectorSimilarityFilterQuery.create( query, similarity, similarityFunction );
```
yrodiere, line 53: Isn't there a way to plug into the `Knn*Query`, so that you filter out the documents *while collecting them*? That could improve performance significantly in some edge cases (large `k`, high `minSimilarity`,  ...).

yrodiere (commented):
> Do you think this could be something useful to add?

For casual users, I very much doubt it's useful.

For advanced users... I can't say, I'm not familiar enough with the topic. I'm not sure the "similarity" value is meaningful enough for users to even be able to provide a meaningful limit, but then... I have little experience with this.

If you can find good use cases, and it's not too hard to support on Elasticsearch/OpenSearch, well... why not. But we'd have to be extra clear this is for very specific use cases.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/predicate/impl/LuceneKnnPredicate.java
marko-bekhta: I didn't see any, and from what I've seen the collecting part lives somewhere at the reader level when searching for the vectors, but I'll double-check 🙈

yrodiere:
> If you can find good use cases, and it's not too hard to support on Elasticsearch/OpenSearch, well... why not. But we'd have to be extra clear this is for very specific use cases.

Ok, this gives a rather interesting use case: https://www.elastic.co/guide/en/elasticsearch/reference/current/knn-search.html#knn-similarity-search

Basically it's not about relevance, it's about performance. A similarity filter stops the search early if the engine can't find any vector that is reasonably similar to the provided one, avoiding unnecessary computation. In other words, we know that vectors that have a similarity lower than a given number won't affect the score that much, so we don't even look for them.

So yes, I think you're right, this is a valuable feature. Please create a Jira?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/query/impl/VectorSimilarityFilterQuery.java
```diff
@@ -0,0 +1,218 @@
...
+
+	@Override
+	public int hashCode() {
+		return Objects.hash( query, similarityAsScore );
+	}
+
+	private static float floatSimilarityDistanceToScore(float distance, VectorSimilarityFunction similarityFunction) {
+		switch ( similarityFunction ) {
+			case EUCLIDEAN:
+				return 1.0f / ( 1.0f + distance );
```
marko-bekhta, line 106: NOTE: from what it seems in Elasticsearch a different formula is used to convert the similarity into score:
```
l2_norm: sqrt((1 / _score) - 1) 
```
from which score is:
```
l2_norm:  _score = (1 / d^2+1) 
```
but then in a Lucene function: `1 / (1 + squareDistance(v1, v2))`

documentation/src/main/asciidoc/public/reference/_search-dsl-predicate.adoc
```diff
@@ -1534,6 +1534,35 @@ include::{sourcedir}/org/hibernate/search/documentation/search/predicate/Predica
...
+----
+<1> Create a knn predicate as usual.
+<2> Specify the required minimum similarity value, to filter out irrelevant results.
+====
+
+NOTE: This configuration is only available for a <<backend-lucene, Lucene backend>>
+and for the <<backend-elasticsearch-compatibility-elasticsearch,Elastic>> distribution of
+an <<backend-elasticsearch,Elasticsearch backend>>.
+<<backend-elasticsearch-compatibility-opensearch,OpenSearch>> does not have this option available,
+and trying to use it will lead to an exception being thrown.
```
yrodiere, line 1564: Maybe mention this limitation in the section about OpenSearch :D

yrodiere (approved):
Thanks! LGTM, just one comment.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/query/impl/VectorSimilarityFilterQuery.java
```diff
@@ -0,0 +1,222 @@
...
+				return Explanation.noMatch( "Similarity limit is greater than the vector similarity.", explanation );
+			}
+			return explanation;
+		}
+
+		@Override
+		public Scorer scorer(LeafReaderContext context) throws IOException {
+			Scorer scorer = super.scorer( context );
+			if ( scorer == null ) {
+				return null;
```
marko-bekhta, line 154: I was running some tests locally and noticed that sometimes the super-scorer can be null, leading to a `NullPointerExeception` in `MinScoreScorer`. So I've added this if statement. See the knn scorer part that would return a `null`

https://github.com/apache/lucene/blob/2a0b7f2056745a2cafbda8cafe47eb7bcb94d28a/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L349-L351

yrodiere (approved):

PR #3892, opened 2024-01-03
HSEARCH-5031 Mass indexing ID loading queries with discriminator multi-tenancy may cross tenant boundaries in some cases

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5031

Opening this as a draft for now, as this is just a test + a fix for a multitenancy helper

### DIFF
--- a/util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/multitenancy/impl/H2LazyMultiTenantConnectionProvider.java
+++ b/util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/multitenancy/impl/H2LazyMultiTenantConnectionProvider.java
@@ -6,13 +6,22 @@
  */
 package org.hibernate.search.util.impl.integrationtest.mapper.orm.multitenancy.impl;
 
+import static org.hibernate.testing.env.ConnectionProviderBuilder.URL_FORMAT;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
+import java.util.Properties;
 
 import org.hibernate.HibernateException;
+import org.hibernate.cfg.Environment;
+import org.hibernate.cfg.JdbcSettings;
 import org.hibernate.engine.jdbc.connections.spi.AbstractMultiTenantConnectionProvider;
 import org.hibernate.engine.jdbc.connections.spi.ConnectionProvider;
 import org.hibernate.search.util.common.impl.Closer;
+import org.hibernate.search.util.impl.integrationtest.mapper.orm.DatabaseContainer;
 import org.hibernate.service.spi.Stoppable;
 
 import org.hibernate.testing.env.ConnectionProviderBuilder;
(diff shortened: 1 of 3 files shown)


mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/HibernateOrmMassIndexingContext.java
```diff
@@ -173,10 +173,12 @@ public PojoMassIdentifierLoader createIdentifierLoader(PojoMassIndexingIdentifie
 					.collect( Collectors.toList() );
 
 			HibernateOrmQueryLoader<E, I> typeQueryLoader = createQueryLoader( typeContexts );
+			// NOTE: we are creating a regular and not a stateless session,
+			//   since a stateless session does not apply multitenant filter (for now)
```
yrodiere, line 177: Maybe add a link to the Jira ticket in ORM?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/HibernateOrmMassIndexingContext.java
```diff
@@ -173,10 +173,12 @@ public PojoMassIdentifierLoader createIdentifierLoader(PojoMassIndexingIdentifie
 					.collect( Collectors.toList() );
 
 			HibernateOrmQueryLoader<E, I> typeQueryLoader = createQueryLoader( typeContexts );
+			// NOTE: we are creating a regular and not a stateless session,
```
yrodiere, line 176: Hmm. I have no idea what kind of impact this could have... Given we only execute a scroll on identifiers, I'd hope not much, but who knows...
I think we should be very careful about backporting this. Ideally we would not, and would backport the fix on the ORM side instead.

yrodiere (approved):
Thanks! LGTM, but as I mentioned in one comment, it's probably a bit dangerous to backport...

Do you have a link to the relevant issue in ORM? Seems pretty serious, so I guess we could hope for a fix and backports relatively soon?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/HibernateOrmMassIndexingContext.java
marko-bekhta: I've only opened a draft PR so far https://github.com/hibernate/hibernate-orm/pull/7675  🙈  let me create tickets ...

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/impl/HibernateOrmMassIndexingContext.java
marko-bekhta: yeah, I am 50-50 on the change... if we get a fix into ORM, then we won't need the change in Search...

util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/multitenancy/impl/H2LazyMultiTenantConnectionProvider.java
```diff
@@ -61,14 +70,34 @@ public ConnectionProvider selectConnectionProvider(String tenantIdentifier) {
...
+
+			for ( String tenantId : tenantIds ) {
+				properties.put( JdbcSettings.URL, String.format( Locale.ROOT, URL_FORMAT, tenantId ) );
+				ConnectionProvider connectionProvider =
+						(ConnectionProvider) buildConnectionProvider.invoke( null, properties, false );
+				delegates.put( tenantId, connectionProvider );
+			}
+		}
+		catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
+			throw new IllegalStateException( "Unable to prepare connection providers", e );
```
marko-bekhta, line 98: Note, there's also a PR https://github.com/hibernate/hibernate-orm/pull/7672 so we may not need to do something like that

marko-bekhta:
I've tested it with 6.4.2.Final locally and all seems fine (with stateless session) let's see if CI also agrees.

yrodiere (approved):

PR #3897, opened 2024-01-09
Stop testing against JDK 20

yrodiere:
It EOL'd on Sept 19th, 2023.
See https://endoflife.date/oracle-jdk

Also, we already test against the next version (21).


marko-bekhta (commented):
there's also a `testWithJdk20` mvn profile ... we probably should drop it now 😃

yrodiere:
Right, good catch. Done.

marko-bekhta (approved):
🎉

PR #3905, opened 2024-01-16
Assign dependabot PRs to Marko instead of Yoann

yrodiere:
Because let's be real, Marko is the one doing all that work lately.


marko-bekhta (approved):
:wasnt-me: 😃

PR #3906, opened 2024-01-16
HSEARCH-5057 Bump software.amazon.awssdk:auth from 2.22.1 to 2.23.3

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5057

https://github.com/hibernate/hibernate-search/pull/3904

### DIFF
--- a/build/parents/build/pom.xml
+++ b/build/parents/build/pom.xml
@@ -81,7 +81,7 @@
         <version.org.opensearch.latest>2.11.0</version.org.opensearch.latest>
 
         <version.com.google.code.gson>2.9.1</version.com.google.code.gson>
-        <version.software.amazon.awssdk>2.22.1</version.software.amazon.awssdk>
+        <version.software.amazon.awssdk>2.23.3</version.software.amazon.awssdk>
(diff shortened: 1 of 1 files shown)


yrodiere (approved):

marko-bekhta:
BTW I see that AWS has 2.11 available https://docs.aws.amazon.com/opensearch-service/latest/developerguide/what-is.html#choosing-version
should we update 
https://github.com/hibernate/hibernate-search/blob/82a86e90d5bb583e9f893084613ea9faebf4dcc9/Jenkinsfile#L278-L280

at some point?

yrodiere:
> BTW I see that AWS has 2.11 available https://docs.aws.amazon.com/opensearch-service/latest/developerguide/what-is.html#choosing-version should we update
> 
> https://github.com/hibernate/hibernate-search/blob/82a86e90d5bb583e9f893084613ea9faebf4dcc9/Jenkinsfile#L278-L280
> 
> at some point?

Ah, shoot, I missed that. I'll update the infra accordingly.

We'll need to update this in all maintained branches, removing this test altogether on branches that just don't support OpenSearch 2.11. I don't want to run many different versions of this service, as it's expensive. Can you please open those PRs?

marko-bekhta:
ok will do 👍🏻 😃

PR #3907, opened 2024-01-16
Backport of HSEARCH-5057 to 7.0 - Bump software.amazon.awssdk:auth from 2.22.1 to 2.23.3

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5057


yrodiere:
Hmm, no, let's not backport a minor version upgrade of a dependency. Especially since there's no particular reason for the upgrade :)

Closing, let's focus on #3906.

PR #3908, opened 2024-01-16
Test against OpenSearch 2.11 on AWS [7.1]

marko-bekhta:


yrodiere:
Thanks. I updated the infra, let's see.

PR #3920, opened 2024-01-22
HSEARCH-5062 Target types as precise as possible for entity loading

marko-bekhta:
Hey Yoann,

I was looking into the https://discourse.hibernate.org/t/polymorphic-mass-indexing-with-discriminator/8869  to see what we can do there.
One of the problems is that we are creating the query factory from the root entity but in that case it makes ORM  join all the tables, so the idea here is to create a query factory from a most specific common supertype... 

https://hibernate.atlassian.net/browse/HSEARCH-5062

### DIFF
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingInheritanceIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingInheritanceIT.java
@@ -76,13 +76,32 @@ class MassIndexingInheritanceIT {
 				// .withProperty( "hibernate.session_factory.statement_inspector", statementInspector )
 				.withProperty( "hibernate.show_sql", true )
 				.withAnnotatedTypes( Car.class, Truck.class, DooredVehicle.class, BedVehicle.class, BaseVehicle.class,
-						BaseEntity.class );
+						BaseEntity.class, Van.class,
+						A.class, AA.class, BA.class, ABA.class, BBA.class, AABA.class, AAABA.class,
+						Root.class, ARoot.class, BRoot.class,
+						RootTablePerClass.class, ARootTablePerClass.class, BRootTablePerClass.class,
+						RootSingleTable.class, ARootSingleTable.class, BRootSingleTable.class
+				);
 
 		// We add the schema expectation as a part of a configuration, and as a last configuration.
 		// this way we will only set the expectation only when the entire config was a success:
 		setupContext.withConfiguration(
 				ignored -> backendMock.expectAnySchema( Car.INDEX ).expectAnySchema( Truck.INDEX )
 						.expectAnySchema( BedVehicle.INDEX )
+						.expectAnySchema( A.INDEX )
+						.expectAnySchema( AA.INDEX )
+						.expectAnySchema( BA.INDEX )
+						.expectAnySchema( BBA.INDEX )
+						.expectAnySchema( AAABA.INDEX )
+						.expectAnySchema( Root.INDEX )
+						.expectAnySchema( ARoot.INDEX )
+
+						.expectAnySchema( RootTablePerClass.INDEX )
+						.expectAnySchema( ARootTablePerClass.INDEX )
+						.expectAnySchema( BRootTablePerClass.INDEX )
+
+						.expectAnySchema( RootSingleTable.INDEX )
+						.expectAnySchema( ARootSingleTable.INDEX )
 		);
 		sessionFactory = setupContext.setup();
 
@@ -91,6 +110,26 @@ class MassIndexingInheritanceIT {
 			session.persist( Car.create( 2L ) );
 			session.persist( Car.create( 3L ) );
 			session.persist( Truck.create( 10L ) );
+			session.persist( BedVehicle.create( 100L ) );
+			session.persist( BedVehicle.create( 101L ) );
+			session.persist( Van.create( 10_000L ) );
+			session.persist( A.create( 100_000L ) );
+			session.persist( AA.create( 100_001L ) );
+			session.persist( BA.create( 100_002L ) );
+			session.persist( ABA.create( 100_003L ) );
+			session.persist( BBA.create( 100_004L ) );
+			session.persist( AAABA.create( 100_005L ) );
+			session.persist( Root.create( 1_000_000L ) );
+			session.persist( ARoot.create( 1_000_001L ) );
+			session.persist( BRoot.create( 1_000_002L ) );
+
+			session.persist( RootSingleTable.create( 2_000_000L ) );
+			session.persist( ARootSingleTable.create( 2_000_001L ) );
+			session.persist( BRootSingleTable.create( 2_000_002L ) );
+
+			session.persist( RootTablePerClass.create( 3_000_000L ) );
+			session.persist( ARootTablePerClass.create( 3_000_001L ) );
+			session.persist( BRootTablePerClass.create( 3_000_002L ) );
 		} );
 	}
 
@@ -158,6 +197,65 @@ class MassIndexingInheritanceIT {
 				.anyMatch( "select [a-z0-9_.,]+ from car [a-z0-9_.]+ "
 						+ "join basevehicle [a-z0-9_.]+ on [a-z0-9_.=]+ "
 						+ "join dooredvehicle [a-z0-9_.]+ on [a-z0-9_.=]+ "
+						+ "where .+" );
+
+		backendMock.verifyExpectationsMet();
+	}
+
+	@Test
+	void singleEntity_notTopOne() {
+		with( sessionFactory ).runNoTransaction( session -> {
+			SearchSession searchSession = Search.session( session );
+			MassIndexer indexer = searchSession.massIndexer( BedVehicle.class );
+
+			backendMock.expectWorks(
+					Truck.INDEX, DocumentCommitStrategy.NONE, DocumentRefreshStrategy.NONE
+			)
+					.add( "10", b -> b
+							.field( "id", 10L )
+					);
+			backendMock.expectWorks(
+					BedVehicle.INDEX, DocumentCommitStrategy.NONE, DocumentRefreshStrategy.NONE
+			)
+					.add( "100", b -> b
+							.field( "id", 100L )
+					)
+					.add( "101", b -> b
+							.field( "id", 101L )
+					);
+
+			backendMock.expectIndexScaleWorks( BedVehicle.INDEX )
+					.purge()
+					.mergeSegments()
+					.flush()
+					.refresh();
+			backendMock.expectIndexScaleWorks( Truck.INDEX )
+					.purge()
+					.mergeSegments()
+					.flush()
+					.refresh();
+
+			try {
+				indexer.startAndWait();
+			}
+			catch (InterruptedException e) {
+				fail( "Unexpected InterruptedException: " + e.getMessage() );
+			}
+
+		} );
+
+		statementInspector.hasSelects( 3 )
+				// select count(bv1_0.id) from bedvehicle bv1_0
+				.anyMatch( "select count\\([a-z0-9_.]+\\) from bedvehicle [a-z0-9_.]+" )
+				// select bv1_0.id from bedvehicle bv1_0
+				.anyMatch( "select [a-z0-9_.]+ from bedvehicle [a-z0-9_.]+" )
+				// select bv1_0.id,bv1_1.type,bv1_2.bodytype,bv1_3.doortype,bv1_0.bedtype,bv1_4.truckroof
+				//   from bedvehicle bv1_0 join baseentity bv1_1 on bv1_0.id=bv1_1.id join basevehicle bv1_2 on bv1_0.id=bv1_2.id join dooredvehicle bv1_3 on bv1_0.id=bv1_3.id left join truck bv1_4 on bv1_0.id=bv1_4.id where bv1_0.id=?
+				.anyMatch( "select [a-z0-9_.,]+ from bedvehicle [a-z0-9_.]+ "
+						+ "join baseentity [a-z0-9_.]+ on [a-z0-9_.=]+ "
+						+ "join basevehicle [a-z0-9_.]+ on [a-z0-9_.=]+ "
+						+ "join dooredvehicle [a-z0-9_.]+ on [a-z0-9_.=]+ "
+						+ "left join truck [a-z0-9_.]+ on [a-z0-9_.=]+ "
 						+ "where .+" );
 
 		backendMock.verifyExpectationsMet();
@@ -252,6 +350,15 @@ class MassIndexingInheritanceIT {
 					.add( "10", b -> b
 							.field( "id", 10L )
 					);
+			backendMock.expectWorks(
+					BedVehicle.INDEX, DocumentCommitStrategy.NONE, DocumentRefreshStrategy.NONE
+			)
+					.add( "100", b -> b
+							.field( "id", 100L )
+					)
+					.add( "101", b -> b
+							.field( "id", 101L )
+					);
 
 			// purgeAtStart and mergeSegmentsAfterPurge are enabled by default,
 			// so we expect 1 purge, 1 mergeSegments and 1 flush calls in this order:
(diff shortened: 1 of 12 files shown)


mapper/orm/src/main/java/org/hibernate/search/mapper/orm/loading/impl/AbstractHibernateOrmLoadingStrategy.java
```diff
@@ -54,17 +51,37 @@ public HibernateOrmQueryLoader<E, I> createQueryLoader(List<LoadingTypeContext<?
 				includedTypesFilter.add( typeContext.typeIdentifier().javaClass() );
 			}
 		}
+		EntityMappingType commonSuperType = toMostSpecificCommonEntitySuperType( typeContexts );
```
yrodiere, line 54: We probably need to test for null here, and throw an exception? We have that in the subclass...

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/loading/impl/HibernateOrmEntityIdEntityLoadingStrategy.java
```diff
@@ -13,33 +13,39 @@
...
-public class HibernateOrmEntityIdEntityLoadingStrategy<E, I>
-		extends AbstractHibernateOrmLoadingStrategy<E, I> {
+public class HibernateOrmEntityIdEntityLoadingStrategy<E, I> extends AbstractHibernateOrmLoadingStrategy<E, I> {
 
 	public static HibernateOrmEntityLoadingStrategy<?, ?> create(SessionFactoryImplementor sessionFactory,
 			EntityMappingType entityMappingType) {
+		// 		EntityMappingType rootEntityMappingType = entityMappingType.getRootEntityDescriptor();
+		//		TypeQueryFactory<?, ?> queryFactory = TypeQueryFactory.create( sessionFactory, entityMappingType,
+		//				entityMappingType.getIdentifierMapping().getAttributeName() );
+		//		return new HibernateOrmEntityIdEntityLoadingStrategy<>( sessionFactory, entityMappingType, queryFactory );
```
yrodiere, line 30: It goes without saying, but this will need to go :)

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/loading/impl/HibernateOrmEntityIdEntityLoadingStrategy.java
```diff
@@ -147,25 +153,14 @@ private PojoSelectionEntityLoader<?> doCreate(EntityMappingType entityMappingTyp
 		// We must pass rootEntityMappingType here, to avoid getting a WrongClassException when loading from the cache,
 		// even if we know we actually want instances from the most specific entity type,
 		// because that exception cannot be recovered from.
```
yrodiere, line 155: I hope we have a test for this :thinking:

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/loading/impl/HibernateOrmNonEntityIdPropertyEntityLoadingStrategy.java
```diff
@@ -144,4 +144,11 @@ private AssertionFailure multipleTypesException(Set<? extends LoadingTypeContext
 								.collect( Collectors.toList() )
 		);
 	}
+
+	@Override
+	@SuppressWarnings("unchecked")
+	protected TypeQueryFactory<E, I> createFactory(EntityMappingType commonSuperType) {
+		return (TypeQueryFactory<E, I>) TypeQueryFactory.create( sessionFactory, commonSuperType,
+				documentIdSourcePropertyName );
+	}
```
yrodiere, line 153: The fact that we have to implement this method differently in the entity-id/non-entity-id subclasses makes me think that some conditions in the constructor of `HqlTypeQueryFactory` are probably unnecessary... we could have two separate constructors / factory methods.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/loading/impl/HibernateOrmEntityIdEntityLoadingStrategy.java
```diff
@@ -147,25 +153,14 @@ private PojoSelectionEntityLoader<?> doCreate(EntityMappingType entityMappingTyp
...
-		return new HibernateOrmSelectionEntityByIdLoader<>( rootEntityMappingType, queryFactory, sessionContext,
+		return new HibernateOrmSelectionEntityByIdLoader<>( rootEntityMappingType,
+				TypeQueryFactory.create(
+						session.getSessionFactory(), entityMappingType,
+						entityMappingType.getIdentifierMapping().getAttributeName() ),
```
yrodiere, line 159:
```suggestion
				createFactory( entityMappingType ),
```

Also, I'm confused as to why you don't need the same change in `HibernateOrmNonEntityIdPropertyEntityLoadingStrategy.java`?

Also, does this mean you're also fixing a problem for entity loading during search? That would be awesome.

yrodiere (commented):
Makes sense, let's see if it works.

At the risk of stating the obvious:

1. We'll need non-regression tests. Unit tests will probably be challenging, so I guess they would be integrationt tests that spy on executed SQL? We probably already do that somewhere.
2. I think the Jira I created mentions other concerns to be addressed... do you want to move that to a separate issue?

marko-bekhta:
> At the risk of stating the obvious:

😃 

> 1. We'll need non-regression tests. Unit tests will probably be challenging, so I guess they would be integrationt tests that spy on executed SQL? We probably already do that somewhere.

+1, I think we can try something similar to this:

https://github.com/hibernate/hibernate-search/blob/f9e4f397f72f132c856f6d7d92df96c47f4b8cf9/integrationtest/mapper/orm-outbox-polling/src/test/java/org/hibernate/search/integrationtest/mapper/orm/outboxpolling/automaticindexing/OutboxPollingAutomaticIndexingProcessingOrderIT.java#L68-L72


> 2. I think the Jira I created mentions other concerns to be addressed... do you want to move that to a separate issue?

no no it's ok, let's do it all here, I was just looking around to see what can be done and wasn't 100% sure if creating the query factories like this, instead of having them initialized in the constructors was an OK approach 😃

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/loading/impl/HibernateOrmEntityIdEntityLoadingStrategy.java
marko-bekhta: > Also, I'm confused as to why you don't need the same change in HibernateOrmNonEntityIdPropertyEntityLoadingStrategy.java?

I've probably just missed it since I was mostly looking at the entity loading query 🙈 

> Also, does this mean you're also fixing a problem for entity loading during search? That would be awesome.

I think it should; since I had some tests failing related to entity projection loading while making changes here, but I'll doublecheck 😃

yrodiere:
Hey @marko-bekhta , FYI my work on https://hibernate.atlassian.net/browse/HSEARCH-5068 is most likely going to conflict with this. So, whoever gets his stuff merged last will have to deal with conflict resolution.

May the best man win :D

marko-bekhta:
I can make a guess ... and since this one needs the next ORM release... 😄

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/massindexing/MassIndexer.java
```diff
@@ -225,4 +225,16 @@ public interface MassIndexer {
...
+	 * Sets whether the massindexer should try to merge types from the same entity hierarchy into groups when performing
+	 * loading operations or have a single group for each type.
+	 * <p>
+	 * Defaults to {@code true}.
+	 *
+	 * @param merge Whether to merge types into groups.
+	 * @return {@code this} for method chaining
+	 */
+	@Incubating
+	MassIndexer mergeTypeGroups(boolean merge);
```
yrodiere, line 238: So... technically, this is already controlled by whether those types have compatible loading strategies (according to `equals`), *and* by `org.hibernate.search.mapper.pojo.loading.spi.PojoMassLoadingStrategy#groupingAllowed`.

I'd argue that if there's a reason not to merge two types together... they should probably just have distinct loading strategies, or at least strategies that return the correct value from `groupingAllowed`?

Also, the same problem exists for selection loading (search), where merging is also based on whether loading strategies are equal.

TBH I'm not entirely sure this is even necessary, especially after your fix. And if it is... I'd rather get a bug report to explain to me why it is, so that we can think of an automated solution.

For example we might need something that _automatically_ sets the loading strategy not just by entity hierarchy, but by something finer grained... the ORM "table mapping" for example?
E.g. entities with `InheritanceType.TABLE_PER_CLASS` automatically get distinct strategies, those with `InheritanceType.SINGLE_TABLE` automatically get a shared strategy, and those with `InheritanceType.JOINED` will get a shared strategy, but one which would only enable grouping when targeting all concrete subtypes of a common entity supertype.
That, of course, would definitely be a separate Jira though.

yrodiere (commented):
My two cents below.

Also, sorry if I missed it, but did you address those items from the Jira?

> * At the moment, there is an optimization that tries to load multiple types belonging to the same hierarchy using a single query. We might want to disable that optimization when we’re not loading all types of that hierarchy, as this seems to hurt performance on large hierarchies. See   https://discourse.hibernate.org/t/polymorphic-mass-indexing-with-discriminator/8869/2?u=yrodiere
> * When using JOINED inheritance, we might want to look for a way to load the IDs from tables of targeted types only. There’s no need to perform any joins for ID loading…

EDIT: I think those two items would be addressed if we added some way of cancelling grouping for JOINED entities when not all concrete entities are present in the group. Somehow :D

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/mapping/impl/HibernateOrmEntityTypeMetadataContributor.java
```diff
@@ -58,6 +59,7 @@ public void contributeAdditionalMetadata(PojoAdditionalMetadataCollectorTypeNode
 		// the Hibernate ORM name internally when using Hibernate ORM features (Session.load, ...).
 		node.entityName( persistentClass.getJpaEntityName() );
 		node.secondaryEntityName( persistentClass.getEntityName() );
+		node.canPotentiallyBeJoined( !( persistentClass instanceof JoinedSubclass ) );
```
yrodiere, line 62: It's a bit weird to expose this metadata in pojo-base where the concept of "join" isn't even defined.

Any reason to add this to the type metadata instead of including it in the (mapper-specific) loading strategy?

For ORM you could pass it to the loading strategy in `org.hibernate.search.mapper.orm.mapping.impl.HibernateOrmEntityLoadingBinder#createLoadingStrategy`, and simply have `org.hibernate.search.mapper.orm.loading.impl.AbstractHibernateOrmLoadingStrategy#groupingAllowed` return `false` for non-joinable types.

I think that's what I suggested earlier... but maybe there was a problem I didn't anticipate?

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingIndexedTypeGroup.java
```diff
@@ -48,16 +48,18 @@ public static List<PojoMassIndexingIndexedTypeGroup<?>> disjoint(PojoMassIndexin
...
-			ListIterator<PojoMassIndexingIndexedTypeGroup<?>> iterator = typeGroups.listIterator();
-			while ( iterator.hasNext() ) {
-				PojoMassIndexingIndexedTypeGroup<?> mergeResult = iterator.next().mergeOrNull( typeGroup );
-				if ( mergeResult != null ) {
-					// We found an existing group that can be merged with this one.
-					// Remove that group, we'll add the merge result to the list later.
-					typeGroup = mergeResult;
-					iterator.remove();
-					// Continue iterating through existing groups, as we may be able to merge with multiple groups.
+			if ( !typeContext.canPotentiallyBeJoined() ) {
```
yrodiere, line 51: Are you sure about this implementation? Believe it or not, when I wrote "cancelling grouping for JOINED entities _when not all concrete entities are present in the group_", the "not all concrete entities" part was there on purpose :grin: 

Your current solution could actually result in worse results when e.g. we have classes `A`, `B extends A`, and `C extends A`, all of them concrete (non-abstract), and we try to reindex A (plus subclasses): we'd end up issuing one query for B (fine), one for C (fine), and one for A (not so fine, would be something like `A join B join C where B.id is null and C.id is null`). All that when a single query targeting all these types would have worked just fine...

Critically, what I mentioned just now is probably much more frequent than the problem you're effectively solving here, which is when we try to reindex A (which is indexed), but C is not indexed.

I'd actually expect this:

* TABLE_PER_CLASS => no grouping
* SINGLE_TABLE => grouping
* JOINED => grouping, except when not targeting all concrete subtypes of the common supertype

Considering the only way of ending up in a situation where one type is being mass indexed but not its concrete subtype, is when that subtype is not indexed (`@Indexed(enabled=false)`), I think an acceptable approximation would be:

* TABLE_PER_CLASS => no grouping
* SINGLE_TABLE => grouping
* JOINED => grouping, except when one of the types involved has non-indexed concrete subtypes

I think the information "an indexed type has non-indexed concrete subtypes" could easily be inferred from current metadata in the constructor of `org.hibernate.search.mapper.pojo.mapping.impl.PojoTypeManagerContainer`: you can list subtypes using `org.hibernate.search.mapper.pojo.automaticindexing.building.impl.PojoImplicitReindexingResolverBuildingHelper#getConcreteEntitySubTypesForEntitySuperType` and check whether they are indexed using `org.hibernate.search.mapper.pojo.mapping.impl.PojoTypeManagerContainer.Builder#indexed` (change the map key type to `PojoRawTypeIdentifier` if that helps).
Checking whether that type is concrete is... left as an exercise to the reader ;D Not sure the info is available right there, but it's in `org.hibernate.search.engine.mapper.model.spi.MappableTypeModel#isAbstract` and in a couple other places.

From there you can pass the info to `org.hibernate.search.mapper.pojo.mapping.impl.AbstractPojoTypeManager.Builder#build` and expose it in `PojoLoadingTypeContext#hasNonIndexedConcreteSubtypes()`.
You'll be able to leverage it to prevent grouping in `org.hibernate.search.mapper.orm.loading.impl.AbstractHibernateOrmLoadingStrategy#groupingAllowed`.

yrodiere (commented):
Thanks, here are my comments... Let's discuss on Zulip if something is wrong/unclear, this is becoming rather complex.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingInheritanceIT.java
```diff
@@ -292,6 +376,95 @@ void multipleSameBranch() {
...
+				//            where b1_1.type in (?)
+				.anyMatch( "select count\\([a-z0-9_.]+\\) from bba [a-z0-9_.]+" )
+				.anyMatch( "select count\\([a-z0-9_.]+\\) from aaaba [a-z0-9_.]+" )
+				.anyMatch( "select count\\([a-z0-9_.]+\\) from ba [a-z0-9_.]+ "
+						+ "join a [a-z0-9_.]+ on [a-z0-9_.=]+ "
+						+ "left join aaaba [a-z0-9_.]+ on [a-z0-9_.=]+ "
+						+ "left join aba [a-z0-9_.]+ on [a-z0-9_.=]+ "
+						+ "left join bba [a-z0-9_.]+ on [a-z0-9_.=]+ "
+						+ "where .+"
+				)
```
marko-bekhta, line 435: from the looks of it ... ORM probably could do a bit better in this case, as all those left joins aren't really needed.

yrodiere: Right... Probably something to request as an enhancement?

It's not *very* critical in our case as the query is only executed once, but it would be nice to get it solved.

yrodiere (commented):
LGTM, with a slight doubt on the "joined" check (see below). You might want to add a test for that.

Can you confirm this seems to solve all problems mentioned in the original Discourse topic?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/loading/impl/AbstractHibernateOrmLoadingStrategy.java
```diff
@@ -37,14 +41,58 @@
...
+		},
+		ONLY_FOR_NO_NON_INDEXED_CONCRETE_SUBTYPES {
+			@Override
+			public boolean allowed(boolean hasNonIndexedConcreteSubtypes) {
+				return !hasNonIndexedConcreteSubtypes;
+			}
+		};
+
+		public static GroupingAllowed determine(PersistentClass persistentClass) {
+			if ( isFromInheritanceType( persistentClass, JoinedSubclass.class ) ) {
```
marko-bekhta, line 65: This should also address the root classes 🙈

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingInheritanceIT.java
```diff
@@ -292,6 +399,289 @@ void multipleSameBranch() {
...
+			catch (InterruptedException e) {
+				fail( "Unexpected InterruptedException: " + e.getMessage() );
+			}
+
+		} );
+
+		statementInspector.hasSelects( 9 )
+				// select count(atpc1_0.id) from aroottableperclass atpc1_0
+				// select count(btpc1_0.id) from broottableperclass btpc1_0
+				// select count(rtpc1_0.id) from (select id, 0 as clazz_ from roottableperclass union all select id, 1 as clazz_ from aroottableperclass union all select id, 2 as clazz_ from broottableperclass) rtpc1_0 where rtpc1_0.clazz_ in (?)
```
marko-bekhta, line 613: this looks strage, with a table per class inheritance 🤔 but ... IDK I'm probably missing something 🙈

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/common/impl/HibernateOrmUtils.java
```diff
@@ -181,4 +181,8 @@ public static List<Property> sortedNonSyntheticProperties(Iterator<Property> pro
 	public static boolean isDiscriminatorMultiTenancyEnabled(Metadata metadata) {
 		return metadata.getFilterDefinition( TenantIdBinder.FILTER_NAME ) != null;
 	}
+
+	public static boolean isJoinedInheritancePersistentClass(PersistentClass persistentClass) {
+		return persistentClass instanceof JoinedSubclass;
```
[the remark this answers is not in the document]
marko-bekhta: yeah, I just wanted to check with you if I got your idea correctly before adding more to his 😃  (in particular other inheritance types)

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingInheritanceIT.java
yrodiere: What looks strange to me is that we (probably?) use `where type in` in the HQL query to begin with... ? Aren't we supposed to detect that we target all subtypes and thus that the `where` is unnecessary?

The generated query doesn't look too odd if you consider that Hibernate ORM can't optimize `where type in (something)` when translating from HQL to SQL if `something` is a parameter. I'm not sure such optimization is implemented anyway.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/loading/impl/AbstractHibernateOrmLoadingStrategy.java
yrodiere: "Should" as in "you added a test and you're sure"?

yrodiere (approved):
Works for me if you're confident in your tests :)
Thanks! I'll let you merge.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingInheritanceIT.java
marko-bekhta: hmm yeah ... 
> TABLE_PER_CLASS => no grouping

sooo if we do not group types in table-per-class - then we aren't targeting all types, aaaand if we aren't targeting all types then ... Since 
```
select e.id from RootTablePerClass
```
should return ids of `RootTablePerClass` and its subclasses ... we can't do that since we aren't grouping subclasses ...so we don't want subclasses and we need a type filter...
So while a simple 
```sql
select id from roottableperclass
```
would've been enough I'm not sure we can do it 🙈 (if that makes sense)

yrodiere: Ah, right, so this is the query for the root class, because it's not abstract.

I'd say having a non-abstract root class is rather unusual, or at least is generally just an oversight.

Anyway, for table-per-class, we have a choice between:

* Grouping, and getting unions without the `where _clazz in (...)` all over the place.
* Not grouping, and getting unions with the `where _clazz in (...)` but only for the root class and only if it's concrete.

Since in this case the `where _class in` relies on a "synthetic" column (no idea what's the proper name) that isn't backed by actual data on disk, I think the database is likely to optimize it, so we can probably ignore it.

So... I'd say not grouping is the better of two evils?

yrodiere: Note we might be able to get rid of `where _clazz in (...)` if Hibernate ORM gave us a way to tell it "target this type only but not its subtypes" in HQL, without relying on a `where` clause.
I don't think that's currently possible though.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/massindexing/MassIndexingInheritanceIT.java
marko-bekhta: https://hibernate.atlassian.net/browse/HHH-17806

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/loading/impl/AbstractHibernateOrmLoadingStrategy.java
marko-bekhta: Yup, there are tests that try indexing from the root as well as from entities deeper in the hierarchy. We definitely can have more tests, but since adding more of these hierarchies to add new variations of possible configurations make the test grow too-big-to-read (and potentially to update the regexps) I'd say let's keep it at what we have and add in the future if some new case comes up. 🙈 😃

marko-bekhta:
Merging 🤞🏻 thanks for taking a look!

yrodiere:
@marko-bekhta Looks like SQL Server tests are failing: https://ge.hibernate.org/s/vt7vnfkdiswm4/tests/overview?outcome=FAILED
(and maybe others, I didn't check)

marko-bekhta:
If only there was a lib that converts some generic SQL into a DB specific one .... 😆  
thanks, will take a look 👍🏻

PR #3924, opened 2024-01-24
HSEARCH-5065 Two-phase boot for the standalone POJO mapper

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-5065

### DIFF
--- a/mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/cfg/spi/StandalonePojoMapperSpiSettings.java
+++ b/mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/cfg/spi/StandalonePojoMapperSpiSettings.java
@@ -18,12 +18,17 @@ public final class StandalonePojoMapperSpiSettings {
 
 	public static final String BEAN_PROVIDER = PREFIX + Radicals.BEAN_PROVIDER;
 
+	public static final String INTEGRATION_PARTIAL_BUILD_STATE =
+			PREFIX + Radicals.INTEGRATION_PARTIAL_BUILD_STATE;
+
 	public static class Radicals {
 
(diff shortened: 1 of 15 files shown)


mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/cfg/spi/StandalonePojoMapperSpiSettings.java
```diff
@@ -18,12 +18,17 @@ private StandalonePojoMapperSpiSettings() {
...
+			PREFIX + Radicals.INTEGRATION_PARTIAL_BUILD_STATE;
+
 	public static class Radicals {
 
 		private Radicals() {
 		}
 
 		public static final String BEAN_PROVIDER = "bean_provider";
+
+		public static final String INTEGRATION_PARTIAL_BUILD_STATE = "integration_partial_build_state";
```
marko-bekhta, line 31: This wouldn't be a problem if one day we decide to mix ORM and standalone mappers? 🙈

marko-bekhta (approved):
🎉 😃

mapper/pojo-standalone/src/main/java/org/hibernate/search/mapper/pojo/standalone/cfg/spi/StandalonePojoMapperSpiSettings.java
yrodiere: It might, yes, but... we're still a long way from that :)
Also, I exposed this as SPI to mirror what we do the the ORM mapper, but if you look at the implementation, technically this property is internal: it's only ever set or consumed by us, not by users or service providers. So I guess we could change its name fairly easily.

PR #3926, opened 2024-01-24
HSEARCH-5066 Upgrade -orm6 artifacts to Hibernate ORM 6.2.22.Final

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5066

And I thought that it was going to be just a few releases 😃 we might want to automate this kind of update to our maintenance branches as well 😄


yrodiere:
> we might want to automate this kind of update to our maintenance branches as well 😄

Maybe, though then again if we start automating release on those branches... maybe not.

Anyway, feel free to try to automate this on 6.2/7.0, we can always disable it later.

yrodiere (approved):

PR #3927, opened 2024-01-24
HSEARCH-5067 Allow targeting entities by name in the Standalone POJO Mapper

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-5067

### DIFF
--- a/documentation/src/main/asciidoc/public/reference/_entrypoints.adoc
+++ b/documentation/src/main/asciidoc/public/reference/_entrypoints.adoc
@@ -198,9 +198,12 @@ The scope's generic type parameter can be any common supertype of those entity t
 <4> A scope will always target all subtypes of the given classes, and the given classes do not need to be indexed entity types themselves.
 This creates a `SearchScope` targeting all (indexed entity) subtypes of the `Person` interface;
 in our case this will target both the `Associate` entity type and the `Manager` entity type.
-<5> For advanced use cases and with the <<mapper-orm,Hibernate ORM integration>> only,
-it is possible to target (indexed entity) types by their name,
-e.g. the JPA entity name.
+<5> For advanced use cases,
+it is possible to target entity types by their name.
+For <<mapper-orm,Hibernate ORM>> this would be the JPA entity name,
+and for the <<mapper-pojo-standalone,Standalone POJO Mapper>>
+this would be the name assigned to the entity type upon declaration in the <<mapper-pojo-standalone-startup,configurer>>.
+In both cases, the entity name is the simple name of the Java class by default.
 <6> Passing `Object.class` will create a scope targeting every single indexed entity types.
 ====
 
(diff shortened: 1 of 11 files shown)


documentation/src/test/java/org/hibernate/search/documentation/mapper/pojo/standalone/entrypoints/StandalonePojoEntryPointsIT.java
```diff
@@ -115,12 +115,12 @@ void searchSession_withOptions() {
 	@Test
 	void searchScope_fromSearchMapping() {
 		SearchMapping searchMapping = theSearchMapping;
-		// tag::searchScope-fromSearchMapping[]
```
marko-bekhta, line 118: Hmm, since these tests aren't used in the docs, should we keep them, or move them to the standalone tests, if it's not yet covered there? 🙈

documentation/src/main/asciidoc/public/reference/_entrypoints.adoc
```diff
@@ -198,9 +198,12 @@ The scope's generic type parameter can be any common supertype of those entity t
 <4> A scope will always target all subtypes of the given classes, and the given classes do not need to be indexed entity types themselves.
 This creates a `SearchScope` targeting all (indexed entity) subtypes of the `Person` interface;
 in our case this will target both the `Associate` entity type and the `Manager` entity type.
-<5> For advanced use cases and with the <<mapper-orm,Hibernate ORM integration>> only,
-it is possible to target (indexed entity) types by their name,
-e.g. the JPA entity name.
+<5> For advanced use cases,
+it is possible to target entity types by their name.
+For Hibernate ORM this would be the JPA entity name,
```
marko-bekhta, line 203:
```suggestion
For <<mapper-orm,Hibernate ORM>> this would be the JPA entity name,
```
maybe let's keep the link?

marko-bekhta (approved):
looks nice 😃 👍🏻

documentation/src/test/java/org/hibernate/search/documentation/mapper/pojo/standalone/entrypoints/StandalonePojoEntryPointsIT.java
yrodiere: I don't know... While we only show the ORM snippets in the documentation, we present them as working for both mappers... So it seems to make sense that we also test the Standalone POJO Mapper...

Also, it seems like to keep it all consistent we'd have to create similar tests for ORM, but then there's duplication... 

All that to say: I won't object to merging a PR with such changes, but I'm not convinced it would be better than the current state of things, so... not worth it?

yrodiere:
I addressed your comments, thanks!

marko-bekhta:
merged! 🎉 😃

PR #3929, opened 2024-01-25
HSEARCH-5070 Change macro to correctly display formulas in HTML docs / HSEARCH-5020 Add an assumption to a test on vector search capability of a backend

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5070
https://hibernate.atlassian.net/browse/HSEARCH-5020

I've noticed that https://docs.jboss.org/hibernate/search/7.1/reference/en-US/html_single/#mapping-directfieldmapping-vectorSimilarity shows unrendered formulas ... 🙈 
this should address that.... though as for the pdf ... it won't https://docs.asciidoctor.org/pdf-converter/latest/stem/ and as far as I can see there's no `asciidoctorJ-mathematical` to add it as a dependency 🙈  so  in pdf it shows up as:
image

### DIFF
--- a/integrationtest/mapper/pojo-standalone-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/standalone/realbackend/mapping/VectorFieldIT.java
+++ b/integrationtest/mapper/pojo-standalone-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/standalone/realbackend/mapping/VectorFieldIT.java
@@ -8,6 +8,7 @@ package org.hibernate.search.integrationtest.mapper.pojo.standalone.realbackend.
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 import java.lang.invoke.MethodHandles;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


yrodiere:
Is there another syntax that's available for PDF? You can use conditionals to include different code depending on the renderer...

marko-bekhta:
From what I got - no... PDF relies on something external to render the formula into an image... So maybe let's just make it more pseudo-java looking for PDF, or... Or just drop the formulas and point to some external doc

yrodiere:
> Or just drop the formulas and point to some external doc

Yeah, that's fine by me. Maybe only for PDF, but whatever is simpler for you.

marko-bekhta:
image
this should look less like a script that didn't get converted... I think 🙈

yrodiere (approved):

yrodiere:
@marko-bekhta looks like this works and you just need a Jira to merge it?

integrationtest/mapper/pojo-standalone-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/standalone/realbackend/mapping/VectorFieldIT.java
```diff
@@ -57,6 +59,14 @@ class VectorFieldIT {
 	public StandalonePojoMappingSetupHelper setupHelper = StandalonePojoMappingSetupHelper.withSingleBackend(
 			MethodHandles.lookup(), BackendConfigurations.simple() );
 
+	@BeforeAll
+	static void beforeAll() {
+		assumeTrue(
```
marko-bekhta, line 64: I've added this assumption to address the main build failure  🙈

yrodiere: Maybe use a related Jira issue, though?
Backporting becomes quite hard once you start attaching unrelated stuff to Jira issues :)

yrodiere:
Merged, thanks

PR #3931, opened 2024-01-26
HSEARCH-5033 Better tests and documentation of limitations around nested fields for highlighting

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5033

I didn't find much to rename anymore, but I've moved tests out of an `AbstractHighlighterIT` that do not really need to be tested against different highlighter types.

### DIFF
--- a/documentation/src/main/asciidoc/public/reference/_search-dsl-projection.adoc
+++ b/documentation/src/main/asciidoc/public/reference/_search-dsl-projection.adoc
@@ -1378,7 +1378,6 @@ The <<mapping-directfieldmapping-highlightable,highlightable>> default may alrea
 in some cases. For more details see <<mapping-directfieldmapping-highlightable-default, how the `DEFAULT` highlightable value behaves>>.
 ====
 
-
 [[search-dsl-projection-highlight-syntax]]
 === Syntax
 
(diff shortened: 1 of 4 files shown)


documentation/src/main/asciidoc/public/reference/_search-dsl-projection.adoc
```diff
@@ -1515,6 +1514,18 @@ include::{sourcedirJava17}/org/hibernate/search/documentation/search/projection/
...
 ====
 
+[[search-dsl-projection-highlight-limitations]]
+=== Highlight limitations
+
+When it comes to highlighting of fields located in <<mapping-indexedembedded-structure-nested,nested>>
+or <<mapping-indexedembedded-structure-flattened,flattened>> objects, Hibernate Search will not allow it in most scenarios right now.
+Trying to do so will result in an exception being thrown.
+The only exception to this is highlighting a field of a <<mapping-indexedembedded-structure-flattened,flattened>> object
+that is a property of a root entity, i.e. the one having a path `[flattenedObjectPropertyName].[propertyName]`.
```
yrodiere, line 1524: Am I really bad at reading, or is this saying "it's not supported on nested fields nor on flattened fields, with the exceptions of flattened fields"... ???

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/highlight/HighlightProjectionTypeIndependentIT.java
```diff
@@ -67,6 +94,21 @@ void objectField_nested() {
 				.hasMessageContaining( "Cannot use 'projection:highlight' on field '" + fieldPath + "'" );
 	}
 
+	/*
+	 * Trying to create a highlight projection on any filed located inside a `ObjectStructure.NESTED` object
```
yrodiere, line 98:
```suggestion
	 * Trying to create a highlight projection on any field located inside a `ObjectStructure.NESTED` object
```

documentation/src/main/asciidoc/public/reference/_search-dsl-projection.adoc
marko-bekhta: How is that for a "better documentation" 😆 ohh 🙈 sorry, it's probably best to show it with the test and then we'll see how to update it...
sooo there is this case:

https://github.com/hibernate/hibernate-search/blob/169c8ad83ea8704346a9166a7e143fac2fb7937c/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/highlight/HighlightProjectionTypeIndependentIT.java#L188-L194

here it doesn't matter what's the structure of `object` in `object + ".string"`, since because we are trying to add a highlight inside a `f.object( object )`.

There's also this case:

https://github.com/hibernate/hibernate-search/blob/169c8ad83ea8704346a9166a7e143fac2fb7937c/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/highlight/AbstractHighlighterIT.java#L627-L632

where it works fine ... 🙈

yrodiere: The first case uses a multi-valued object projection, which implies reconstituting the structure of the object field.

While this kind of reconstitution is restricted to nested fields on Lucene, it can work with a flattened field on Elasticsearch since Elasticsearch can use the `_source` JSON to recover the structure of an object field.

However, for highlighting, we just can't apply this "hack": we don't rely on `_source`, or at least not directly.

In the second example, you are not trying to reconstitute the structure of the object field with an object projection... so it works.

yrodiere (approved):
LGTM, thanks.

PR #3933, opened 2024-01-26
HSEARCH-5020 Increase the max allowed dimension for a Lucene backend

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5020

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/codec/impl/HibernateSearchKnnVectorsFormat.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/codec/impl/HibernateSearchKnnVectorsFormat.java
@@ -20,7 +20,10 @@ import org.apache.lucene.index.SegmentReadState;
 import org.apache.lucene.index.SegmentWriteState;
 
 public class HibernateSearchKnnVectorsFormat extends KnnVectorsFormat {
-	public static final int DEFAULT_MAX_DIMENSIONS = KnnVectorsFormat.DEFAULT_MAX_DIMENSIONS;
+	// OpenSearch has a limit of 16000
+	// Elasticsearch has a limit of 4096
+	// We'll keep it at 4096 for now as well:
+	public static final int DEFAULT_MAX_DIMENSIONS = 4096;
 	private static final KnnVectorsFormat DEFAULT_KNN_VECTORS_FORMAT = new HibernateSearchKnnVectorsFormat();
 
 	public static KnnVectorsFormat defaultFormat() {
@@ -59,7 +62,6 @@ public class HibernateSearchKnnVectorsFormat extends KnnVectorsFormat {
 
 	@Override
 	public int getMaxDimensions(String fieldName) {
-		// TODO: HSEARCH-5020: we can make this configurable, apparently there are models that produce larger vectors than this default allows.
 		return DEFAULT_MAX_DIMENSIONS;
 	}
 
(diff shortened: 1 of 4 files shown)


integrationtest/mapper/pojo-standalone-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/standalone/realbackend/mapping/VectorFieldIT.java
```diff
@@ -24,21 +29,140 @@
...
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 
 class VectorFieldIT {
 
 	private static final String INDEX_NAME = "IndexName";
+	private static final int BATCHES = 20;
+	private static final int BATCH_SIZE = 1_000;
```
yrodiere, line 54: Not sure why you need so many documents? Wouldn't just one document be enough?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/codec/impl/HibernateSearchKnnVectorsFormat.java
```diff
@@ -20,7 +20,10 @@
 import org.apache.lucene.index.SegmentWriteState;
 
 public class HibernateSearchKnnVectorsFormat extends KnnVectorsFormat {
-	public static final int DEFAULT_MAX_DIMENSIONS = KnnVectorsFormat.DEFAULT_MAX_DIMENSIONS;
+	// OpenSearch has a limit of 16000
+	// Elasticsearch has a limit of 4096
+	// We'll keep it at 4096 for now as well:
+	public static final int DEFAULT_MAX_DIMENSIONS = 4096;
```
yrodiere, line 26: Isn't there any documentation (asciidoc/javadoc) to update as a result of this bump? We didn't use to document the max?

yrodiere (approved):
LGTM, thanks. Just a few questions.

integrationtest/mapper/pojo-standalone-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/standalone/realbackend/mapping/VectorFieldIT.java
marko-bekhta: Well.... 😄 
I wanted to see how it all behaves when there are more documents since we are mainly testing with indexes containing 1-20 docs. Hence that much documents in this test.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/codec/impl/HibernateSearchKnnVectorsFormat.java
marko-bekhta: Thanks for pointing that out! updated and added links to ES/OS docs so we don't need to track  any changes on their side.

it's a good thing that I've left those `TODO: HSEARCH-5020` todos in docs and code so whenever we come to work on this ticket, we don't miss any places to update... and then just ignore them 🙈 😭 🤦🏻‍♂️

integrationtest/mapper/pojo-standalone-realbackend/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/standalone/realbackend/mapping/VectorFieldIT.java
yrodiere: Fair enough. Though that probably should be a separate test, with intentions clearly stated either in the test name or in a comment :)

yrodiere (approved):

yrodiere (approved):

PR #3934, opened 2024-01-29
HSEARCH-5064 Integrate Elasticsearch's knn query introduced in 8.12

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5064

Another dialect doesn't look too terrible 😃 ... it seems ... 😃


documentation/src/main/asciidoc/public/reference/_search-dsl-predicate.adoc
```diff
@@ -1578,7 +1578,7 @@ While when an <<backend-elasticsearch-compatibility-opensearch,OpenSearch>> dist
 when the index is configured to have more than one shard.
 See this section of the OpenSearch link:{openSearchDocUrl}/search-plugins/knn/approximate-knn/#get-started-with-approximate-k-nn[documentation] for more details.
 
-With the Elasticsearch backend and when using <<backend-elasticsearch-compatibility-elasticsearch,Elasticsearch>>,
+With the Elasticsearch backend and when using <<backend-elasticsearch-compatibility-elasticsearch,Elasticsearch>> version < 8.12,
```
yrodiere, line 1581: See  that's the kind of things I'm talking about. If we keep support for KNN predicates in Elasticsearch 8.11 and below, documentation is more complex, even for people using Elasticsearch 8.12, because of these limitations. And questions from people trying to use it on 8.11... :face_exhaling: I don't want to think about that.

So yeah, I really think we should remove support for knn predicates in ES 8.11 and below. You can do it in a single commit if you want: that way, if we ever need to revert restore such support, it would be relatively straightforward.

yrodiere (approved):
LGTM, thanks, though I still think we should remove support for 8.11 (along with all the limitation warnings spread through the docs).

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/dialect/model/impl/Elasticsearch8ModelDialect.java
```diff
@@ -6,25 +6,25 @@
...
+import org.hibernate.search.backend.elasticsearch.validation.impl.Elasticsearch7PropertyMappingValidatorProvider;
 import org.hibernate.search.backend.elasticsearch.validation.impl.ElasticsearchPropertyMappingValidatorProvider;
 
 import com.google.gson.Gson;
 
 /**
- * The model dialect for Elasticsearch 8.x.
+ * The model dialect for Elasticsearch [8.0-8.11].
  */
 public class Elasticsearch8ModelDialect implements ElasticsearchModelDialect {
```
marko-bekhta, line 19: we had a 8 dialect before all those changes for knn, so I've kept it as it was, even though we could just reuse the 7 one for Elasticsearch 8.0-8.11.

marko-bekhta:
sorry noticed that I've missed to rename one test method 🙈

PR #3937, opened 2024-01-31
HSEARCH-4563 queryString predicate for advanced, Lucene-syntax query strings / HSEARCH-5073 Add minimum_should_match to simple query string predicate

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4563
https://hibernate.atlassian.net/browse/HSEARCH-5073


engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/QueryStringPredicateOptionsStep.java
```diff
@@ -0,0 +1,235 @@
...
+
+import org.hibernate.search.engine.search.common.BooleanOperator;
+import org.hibernate.search.engine.search.common.Rewrite;
+
+/**
+ * The final step in an "simple query string" predicate definition, where optional parameters can be set.
+ *
+ * @param <S> The "self" type (the actual exposed type of this step).
+ */
+public interface QueryStringPredicateOptionsStep<S extends QueryStringPredicateOptionsStep<?>>
```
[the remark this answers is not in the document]
marko-bekhta: I went ahead with those options and here are some tests for the predicate options:
https://github.com/hibernate/hibernate-search/blob/02479424969ce2ed7ab98fe0c0f35ebfa5108f9a/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/QueryStringPredicateSpecificsIT.java#L31-L268

and here https://github.com/hibernate/hibernate-search/blob/02479424969ce2ed7ab98fe0c0f35ebfa5108f9a/integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/search/LuceneQueryStringPredicateSpecificsIT.java#L67-L111 as this options needs a custom analyzer so it didn't fit nicely into a regular TCK test 🙈 

let me know if these options make sense and if I should try adding more 😈 😄

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchCommonQueryStringPredicate.java
```diff
@@ -0,0 +1,175 @@
...
+			}
+		}
+
+		@Override
+		public final void queryString(String queryString) {
+			this.queryString = queryString;
+		}
+
+		@Override
+		public final FieldState field(String fieldPath) {
```
yrodiere, line 136: `FieldState` is specific to simple query strings... there's probably something wrong in the type hierarchy?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/ElasticsearchStringIndexFieldTypeOptionsStep.java
```diff
@@ -213,7 +213,7 @@ public IndexFieldType<String> toIndexFieldType() {
 			builder.queryElementFactory( PredicateTypeKeys.REGEXP, new ElasticsearchTextRegexpPredicate.Factory() );
 			builder.queryElementFactory( PredicateTypeKeys.TERMS, new ElasticsearchTermsPredicate.Factory<>( codec ) );
 			builder.queryElementFactory( ElasticsearchPredicateTypeKeys.SIMPLE_QUERY_STRING,
-					new ElasticsearchSimpleQueryStringPredicateBuilderFieldState.Factory() );
+					new ElasticsearchCommonQueryStringPredicateBuilderFieldState.Factory() );
```
yrodiere, line 216: Seems weird that the key is still named `SIMPLE_QUERY_STRING`?

Also, this won't play well with the metamodel: the field type will have a trait named `predicate:simple-query-string` and no trait named `predicate:query-string`.

You probably need two separate keys here.

And you need to look at the metamodel test for this predicate, cause there's something missing or wrong if this didn't get detected :D

documentation/src/test/java/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java
```diff
@@ -874,6 +874,27 @@ void simpleQueryString() {
...
+					.where( f -> f.queryString()
+							.field( "title" ).boost( 20.0f )
+							.field( "description" ).boost( 2.0f )
+							.matching( "robots" )
+							.minimumShouldMatch()
+							.ifMoreThan( 0 )
+							.thenRequireNumber( 2 )
+							.end()
+					).toQuery().queryString() );
+		} );
```
yrodiere, line 895: Adding a comment so that you don't forget to adapt/remove this.

Also, you'll probably want to use a more complicated example for the documentation. Since the main advantage of `queryString` over `simpleQueryString` is that it's more powerful.

engine/src/main/java/org/hibernate/search/engine/logging/impl/Log.java
```diff
@@ -570,4 +571,14 @@ SearchException cannotIncludeAndExcludePathsWithinSameFilter(Set<String> include
 			value = "Multiple configuration providers are available for scope '%1$s'. "
 					+ "They will be taken under consideration in the following order: '%2$s'.")
 	void multipleConfigurationProvidersAvailable(String scope, List<ConfigurationProvider> configurationProviders);
+
+	@Message(id = ID_OFFSET + 121,
+			value = "Trying to use a rewrite method '%1$s' that requires parameter N without specifying the parameter. "
+					+ "Use an overloaded method that allows passing the parameter N.")
```
yrodiere, line 577: When the user made a mistake, could you please follow this wording? I did a pass around the 6.0 release to make sure we used consistent wording, though I admit things may have drifted a bit since then.

Also I'm afraid "overloaded" may not be newbie-friendly...

```suggestion
			value = "Cannot use rewrite method '%1$s': this method requires parameter 'n', which was not specified."
					+ "Use another version of the rewrite(...) method that accepts parameter 'n'.")
```

Though it would be even better if we used a more self-explanatory name than `n`...

engine/src/main/java/org/hibernate/search/engine/logging/impl/Log.java
```diff
@@ -570,4 +571,14 @@ SearchException cannotIncludeAndExcludePathsWithinSameFilter(Set<String> include
...
+	SearchException parameterizedRewriteWithoutParameter(Rewrite rewrite);
+
+	@Message(id = ID_OFFSET + 122,
+			value = "Trying to use a rewrite method '%1$s' that does not require any parameters with a specified parameter N. "
+					+ "Use an overloaded method that does not require passing the parameter N.")
```
yrodiere, line 582:
```suggestion
			value = "Cannot use rewrite method '%1$s': this method does not accept parameter 'n', but it was specified."
					+ "Use another version of the rewrite(...) method that does not accept parameter 'n'.")
```

engine/src/main/java/org/hibernate/search/engine/search/common/Rewrite.java
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * The rewrite parameter determines:
+ *
+ * <ul>
+ *     <li>How Lucene calculates the relevance scores for each matching document</li>
+ *     <li>Whether Lucene changes the original query to a bool query or bit set</li>
+ *     <li>If changed to a bool query, which term query clauses are included</li>
+ * </ul>
+ */
+public enum Rewrite {
```
yrodiere, line 18: `RewriteMethod`, maybe? You call it that in error messages.

engine/src/main/java/org/hibernate/search/engine/search/common/Rewrite.java
```diff
@@ -0,0 +1,51 @@
...
+	 */
+	TOP_TERMS_BLENDED_FREQS_N,
+	/**
+	 * Assigns each matching document a relevance score equal to the boost parameter.
+	 */
+	TOP_TERMS_BOOST_N,
+	/**
+	 * Calculates a relevance score for each matching document.
+	 */
+	TOP_TERMS_N;
```
yrodiere, line 49: My, that's complicated.

Two suggestions:

* You might want to check those names are consistent across Lucene/Elasticsearch/OpenSearch, and if not, pick the most self-explanatory set of names.
* Maybe add some info to the Javadoc regarding use cases for each rewrite method? Or at least to the reference documentation.

It would probably help

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/CommonQueryStringPredicateFieldMoreStep.java
```diff
@@ -0,0 +1,57 @@
...
+ * @param <S> The "self" type (the actual exposed type of this step).
+ * @param <N> The type of the next step.
+ */
+public interface CommonQueryStringPredicateFieldMoreStep<
+		S extends CommonQueryStringPredicateFieldMoreStep<?, N>,
+		N extends CommonQueryStringPredicateOptionsStep<?>>
+		extends CommonQueryStringPredicateMatchingStep<N>, MultiFieldPredicateFieldBoostStep<S> {
+
+	/**
+	 * Target the given field in the simple query string predicate,
```
yrodiere, line 24: You'll need to update the javadoc if you really want to share these interfaces.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/CommonQueryStringPredicateFieldStep.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * The initial step in an "abstract" "query string" predicate definition, where the target field can be set.
+ *
+ * @param <N> The type of the next step.
+ */
+public interface CommonQueryStringPredicateFieldStep<N extends CommonQueryStringPredicateFieldMoreStep<?, ?>> {
+
+	/**
+	 * Target the given field in the simple query string predicate.
```
yrodiere, line 17: Same here, javadoc needs some love if you really want to share these interfaces.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/CommonQueryStringPredicateOptionsStep.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.engine.search.predicate.dsl;
+
+import org.hibernate.search.engine.search.common.BooleanOperator;
+
+/**
+ * The final step in an "abstract" "query string" predicate definition, where optional parameters can be set.
```
yrodiere, line 12: I'd go with simply `a query string predicate definition`. "abstract" is a bit confusing here.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/QueryStringPredicateFieldMoreStep.java
```diff
@@ -0,0 +1,23 @@
...
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.engine.search.predicate.dsl;
+
+/**
+ * The step in a "query string" predicate definition where the query string to match can be set
+ * (see the superinterface {@link QueryStringPredicateMatchingStep}),
```
yrodiere, line 11: Haha, see, sharing superinterfaces also has downsides :)

Anyway, I guess it's for the best, as it allows callers to write generic code.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/QueryStringPredicateOptionsStep.java
```diff
@@ -0,0 +1,202 @@
...
+	//	 * @return {@code this}, for method chaining.
+	//	 */
+	//	S lenient(boolean lenient);
+
+	//	/**
+	//	 *
+	//	 * @param maxDeterminizedStates
+	//	 * @return {@code this}, for method chaining.
+	//	 */
+	//	S maxDeterminizedStates(int maxDeterminizedStates);
```
yrodiere, line 106: I'd wait for an actual feature request before working on these.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/QueryStringPredicateOptionsStep.java
```diff
@@ -0,0 +1,202 @@
...
+
+	/**
+	 * When true, resulting queries are aware of position increments.
+	 * This setting is useful when the removal of stop words leaves an unwanted “gap” between terms.
+	 * Default is true.
+	 *
+	 * @param enablePositionIncrements
+	 * @return {@code this}, for method chaining.
+	 */
+	S enablePositionIncrements(boolean enablePositionIncrements);
```
yrodiere, line 52: Doesn't that require enabling term vectors or something?

Not sure it's worth it, but if you already implemented and tested it, why not... You'll need to document it though.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/QueryStringPredicateOptionsStep.java
```diff
@@ -0,0 +1,202 @@
...
+ */
+public interface QueryStringPredicateOptionsStep<S extends QueryStringPredicateOptionsStep<?>>
+		extends CommonQueryStringPredicateOptionsStep<S> {
+
+	/**
+	 * Specifies whether * and ? are allowed as first characters of a search term. Default is true.
+	 * @param allowLeadingWildcard
+	 * @return {@code this}, for method chaining.
+	 */
+	S allowLeadingWildcard(boolean allowLeadingWildcard);
```
yrodiere, line 26: Ok.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/QueryStringPredicateOptionsStep.java
```diff
@@ -0,0 +1,202 @@
...
+
+	//	/**
+	//	 * Specifies whether backend should attempt to analyze wildcard terms.
+	//	 * Default is false.
+	//	 *
+	//	 * @param analyzeWildcard
+	//	 * @return {@code this}, for method chaining.
+	//	 */
+	//	S analyzeWildcard(boolean analyzeWildcard);
```
yrodiere, line 35: Seems useful, but:

1. Please be careful to use a default that matches that of the `wildcard` predicate.
2. Please give up if it's not already implemented for Lucene.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/QueryStringPredicateOptionsStep.java
```diff
@@ -0,0 +1,202 @@
...
+
+	//	/**
+	//	 * Specifies whether to create a match phrase query automatically for multi-term synonyms. For example, if you specify ba, batting average as synonyms and search for ba, OpenSearch searches for ba OR "batting average" (if this option is true) or ba OR (batting AND average) (if this option is false). Default is true.
+	//	 * @param autoGenerateSynonymsPhraseQuery
+	//	 * @return {@code this}, for method chaining.
+	//	 */
+	//	S autoGenerateSynonymsPhraseQuery(boolean autoGenerateSynonymsPhraseQuery);
```
yrodiere, line 42: I'd wait for an actual feature request before working on this.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/QueryStringPredicateOptionsStep.java
```diff
@@ -0,0 +1,202 @@
...
+	/**
+	 * The maximum number of words that are allowed between the matched words.
+	 * If phrase_slop is 2, a maximum of two words is allowed between matched words in a phrase.
+	 * Transposed words have a slop of 2.
+	 * Default is 0 (an exact phrase match where matched words must be next to each other).
+	 *
+	 * @param phraseSlop
+	 * @return {@code this}, for method chaining.
+	 */
+	S phraseSlop(Integer phraseSlop);
```
yrodiere, line 149: There's a phrase predicate in Hibernate Search, please make sure the naming/javadoc is consistent with that predicate (I didn't check).

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/QueryStringPredicateOptionsStep.java
```diff
@@ -0,0 +1,202 @@
...
+
+	//	/**
+	//	 * This option supports searching for exact matches (surrounded with quotation marks) using a different analysis method than non-exact matches use.
+	//	 * For example, if quote_field_suffix is .exact and you search for \"lightly\" in the title field,
+	//	 * OpenSearch searches for the word lightly in the title.exact field.
+	//	 * This second field might use a different type (for example, keyword rather than text) or a different analyzer.
+	//	 * @param quoteFieldSuffix
+	//	 * @return {@code this}, for method chaining.
+	//	 */
+	//	S quoteFieldSuffix(String quoteFieldSuffix);
```
yrodiere, line 169: I'd wait for an actual feature request before working on these.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/QueryStringPredicateOptionsStep.java
```diff
@@ -0,0 +1,202 @@
...
+
+	/**
+	 * Determines how OpenSearch rewrites and scores multi-term queries.
```
yrodiere, line 172: You may want to read through the Javadoc to check for things like this :)

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/QueryStringPredicateOptionsStep.java
```diff
@@ -0,0 +1,202 @@
...
+
+	//	/**
+	//	 * Specifies the number of hours to offset the desired time zone from UTC.
+	//	 * You need to indicate the time zone offset number if the query string contains a date range.
+	//	 * For example, set time_zone": "-08:00" for a query with a date range such as "query": "wind rises release_date[2012-01-01 TO 2014-01-01]").
+	//	 * The default time zone format used to specify number of offset hours is UTC.
+	//	 * @param timeZone
+	//	 * @return {@code this}, for method chaining.
+	//	 */
+	//	S timeZone(String timeZone);
```
yrodiere, line 200: I'd wait for an actual feature request before working on this.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/QueryStringPredicateOptionsStep.java
```diff
@@ -0,0 +1,202 @@
...
+
+	MinimumShouldMatchConditionStep<? extends S> minimumShouldMatch();
+
+	/**
+	 *
+	 * @param constraintContributor
+	 * @return {@code this}, for method chaining.
+	 */
+	S minimumShouldMatch(
+			Consumer<? super MinimumShouldMatchConditionStep<?>> constraintContributor);
```
yrodiere, line 138: OK, but please make sure the javadoc is consistent with what we already have elsewhere.

Also... Apparently it's supported for the simple query string query too: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html
So if you add it here, maybe open a Jira for simple query string? Even if you solve it in this PR: we need to document the improvement in the changelog.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/search/query/ElasticsearchQueryStringPredicateSpecificsIT.java
```diff
@@ -0,0 +1,129 @@
...
+
+		SearchQuery<DocumentReference> query = scope.query()
+				.where( f -> f.queryString()
+						.field( absoluteFieldPath )
+						.matching( "\"out of the room\"" ) )
+				.toQuery();
+
+		assertThatQuery( query )
+				.hasDocRefHitsAnyOrder( index.typeName(), DOCUMENT_2 );
+	}
```
yrodiere, line 106: Why is this specific to Elasticsearch?

If it's just about the analyzer definition, you can make it backend-independent; see:

https://github.com/hibernate/hibernate-search/blob/6354c347a466be9f8a04d4bf985736b3bcdf68cc/integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/analysis/AnalysisCustomIT.java#L259

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/AbstractBaseQueryStringPredicateBaseIT.java
```diff
@@ -0,0 +1,605 @@
...
+		}
+
+		@Override
+		protected void tryPredicate(SearchPredicateFactory f, String fieldPath) {
+			predicate( f ).field( fieldPath );
+		}
+
+		@Override
+		protected String predicateTrait() {
+			return "predicate:simple-query-string";
```
yrodiere, line 415: I think not :)

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/QueryStringPredicateSpecificsIT.java
```diff
@@ -0,0 +1,275 @@
...
+				.matching( TERM_1 );
+
+		SearchPredicate predicate = ( n == null ? optionsStep.rewrite( rewrite ) : optionsStep.rewrite( rewrite, n ) )
+				.toPredicate();
+
+		SearchQuery<DocumentReference> query = scope.query()
+				.where( predicate )
+				.toQuery();
+		assertThatQuery( query )
+				.hasDocRefHitsAnyOrder( index.typeName(), DOCUMENT_1, DOCUMENT_2 );
```
yrodiere, line 48: If the query string and search results are always the same, does this really test the rewrite options, or just that none blows up? :thinking:

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/QueryStringPredicateSpecificsIT.java
```diff
@@ -0,0 +1,275 @@
...
+				Arguments.of( phraseReorderedWords, 0, null ),
+				// Transposed terms have a slop of 2
+				Arguments.of( phraseReorderedWords, 2, DOCUMENT_1 ),
+				// not enough so should not match anything
+				Arguments.of( phraseReorderedWords, 1, null ),
+				Arguments.of( phraseWithoutTwoWordsInTheMiddle, 1, null ),
+				Arguments.of( phraseWithoutTwoWordsInTheMiddle, 2, DOCUMENT_1 ),
+				Arguments.of( phraseWithoutTwoWordsInTheMiddle, 5, DOCUMENT_1 )
+		);
+	}
```
yrodiere, line 268: This would probably be easier to read if each "parameter method" was just after the corresponding test method... no?

yrodiere (commented):
Alright, here's a first pass :) I think you're all set regarding the options, no need to add more.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/search/query/ElasticsearchQueryStringPredicateSpecificsIT.java
marko-bekhta: oh nice, yeah, only beacuse of the analyzer, to make some meaningful test for this option there should be a stop filter 🙈, thanks for the pointer 😃 👍🏻

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/QueryStringPredicateSpecificsIT.java
marko-bekhta: > or just that none blows up

just that, I haven't figured out yet how to make a test where a rewrite is making a "visible difference" 😃

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/QueryStringPredicateSpecificsIT.java
marko-bekhta: good point 😃 will update it

engine/src/main/java/org/hibernate/search/engine/logging/impl/Log.java
marko-bekhta: for all these "parameterized" rewrites it will eventually reach
https://github.com/apache/lucene/blob/9caeb9395dad6a535882888335915e4ea78e8b43/lucene/core/src/java/org/apache/lucene/search/MultiTermQuery.java#L207

where the parameter is called `size`, but I  thought since we have the `N` in the enum name... it would make sense to keep it as `N`

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/QueryStringPredicateOptionsStep.java
marko-bekhta: ok, yeah looks like it needs some work for Lucene... so:

>I'd wait for an actual feature request before working on this. 

😃

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/QueryStringPredicateOptionsStep.java
marko-bekhta: created https://hibernate.atlassian.net/browse/HSEARCH-5073 for tracking

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/util/ElasticsearchTckBackendFeatures.java
```diff
@@ -279,4 +280,16 @@ public boolean supportsSimilarity(VectorSimilarity vectorSimilarity) {
...
+
+	@Override
+	public boolean supportQueryStringRewriteMethod(RewriteMethod rewriteMethod) {
+		return isActualVersion(
+				es -> true,
+				// OpenSearch only has 6 options for the rewrites available:
+				//   constant_score, scoring_boolean, constant_score_boolean,
+				//   top_terms_N, top_terms_boost_N, and top_terms_blended_freqs_N
+				os -> !RewriteMethod.CONSTANT_SCORE_BLENDED.equals( rewriteMethod ),
+				aoss -> !RewriteMethod.CONSTANT_SCORE_BLENDED.equals( rewriteMethod )
```
marko-bekhta, line 292: Turns out that this rewrite method is not available on the OpenSearch side..
Since this (rewrite methods) is probably an advanced option anyway, would it be enough to simply leave a note in the Javadocs for users to see what options are allowed by their backends, or should we do something similar to what was done for vector search, where each distribution would have own list of allowed methods and throw an early exception if something unsupported is supplied? 
Right now if constant score blended is supplied to OpenSearch it results in:
```
Response: 400 'Bad Request' from 'http://localhost:32769' with body 
{
  "error": {
    "root_cause": [
      {
        "type": "query_shard_exception",
        "reason": "failed to create query: Failed to parse rewrite_method [constant_score_blended]",
        "index": "indexname-000001",
        "index_uuid": "sm4DikTfSyyMyxLercR4fQ"
      }
    ],
    "type": "search_phase_execution_exception",
    "reason": "all shards failed",
    "phase": "query",
    "grouped": true,
    "failed_shards": [
      {
        "shard": 0,
        "index": "indexname-000001",
        "node": "sjCTMoEDR_CxWrvQkOX4Ng",
        "reason": {
          "type": "query_shard_exception",
          "reason": "failed to create query: Failed to parse rewrite_method [constant_score_blended]",
          "index": "indexname-000001",
          "index_uuid": "sm4DikTfSyyMyxLercR4fQ",
          "caused_by": {
            "type": "illegal_argument_exception",
            "reason": "Failed to parse rewrite_method [constant_score_blended]"
          }
        }
      }
    ]
  },
  "status": 400
}
```
at query execution time ....

Or just drop the option?

documentation/src/main/asciidoc/public/reference/_search-dsl-predicate.adoc
```diff
@@ -1572,6 +1590,176 @@ but can be <<search-dsl-predicate-common-constantScore,made constant with `.cons
...
+but this query is broken down as two clauses for `robot` and `?etecitve` where in the second
+clause, the `?` wildcard becomes a leading character.
+
+[[search-dsl-predicate-query-string-enable-position-increments]]
+=== Enabling position increments
+
+Position increments are enabled by default.
+When disabled, resulting queries are not aware of position increments
+resulted from removing stop words by a `stopwords` filter.
+This, for example, can lead to a change in the behaviour of phrase queries:
```
yrodiere, line 1691:
```suggestion
Position increments are enabled by default, allowing phrase queries to take into account stop words removed by a `stopwords` filter.

Position increments can be disabled as shown below,
leading to a change in the behaviour of phrase queries:
```

documentation/src/main/asciidoc/public/reference/_search-dsl-predicate.adoc
```diff
@@ -1572,6 +1590,176 @@ but can be <<search-dsl-predicate-common-constantScore,made constant with `.cons
...
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java[tags=queryString-enable-position-increments]
+----
+====
+
+[[search-dsl-predicate-query-string-rewrite-method]]
+=== Rewrite method
+
+The rewrite method determines how the backend's query parser rewrites and scores multi-term queries.
+To change the default `CONSTANT_SCORE` rewrite method, one of the allowed `RewriteMethod` enum values can be passed
```
yrodiere, line 1707:
```suggestion
The rewrite method determines how the backend's query parser rewrites and scores multi-term queries.
To change the default `CONSTANT_SCORE` rewrite method, one of the allowed `RewriteMethod` enum values can be passed
```
```suggestion
The rewrite method determines how the backend's query parser rewrites and scores multi-term queries.

To change the default `CONSTANT_SCORE` rewrite method, one of the allowed `RewriteMethod` enum values can be passed
```

documentation/src/main/asciidoc/public/reference/_search-dsl-predicate.adoc
```diff
@@ -1572,6 +1590,176 @@ but can be <<search-dsl-predicate-common-constantScore,made constant with `.cons
...
+
+IMPORTANT: If targeted fields have different analyzers, an exception will be thrown.
+You can avoid this by <<search-dsl-predicate-common-overriding-analysis,picking an analyzer explicitly>>,
+but make sure you know what you're doing.
+
+[[search-dsl-predicate-query-string-other]]
+=== Other options
+
+* The score of a `queryString` predicate is variable by default,
+but can be <<search-dsl-predicate-common-constantScore,made constant with `.constantScore()`>>.
```
yrodiere, line 1755: You said above that the default rewrite method was "CONSTANT_SCORE"... might want to explain that this doesn't mean, in fact, that the score is constant...

documentation/src/main/asciidoc/public/reference/_search-dsl-predicate.adoc
```diff
@@ -1572,6 +1590,176 @@ but can be <<search-dsl-predicate-common-constantScore,made constant with `.cons
...
+----
+====
+
+[[search-dsl-predicate-query-string-rewrite-method]]
+=== Rewrite method
+
+The rewrite method determines how the backend's query parser rewrites and scores multi-term queries.
+To change the default `CONSTANT_SCORE` rewrite method, one of the allowed `RewriteMethod` enum values can be passed
+to `.rewriteMethod(RewriteMethod)`/`rewriteMethod(RewriteMethod, int)`.
+We will not go into details about different rewrite methods in this guide.
```
yrodiere, line 1709:
```suggestion
to `.rewriteMethod(RewriteMethod)`/`rewriteMethod(RewriteMethod, int)`.

We will not go into details about different rewrite methods in this guide.
```

documentation/src/test/java/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java
```diff
@@ -872,6 +873,128 @@ void simpleQueryString() {
...
+	}
+
+	@Test
+	void queryString() {
+		withinSearchSession( searchSession -> {
+			// tag::queryString-query[]
+			List<Book> hits = searchSession.search( Book.class )
+					.where( f -> f.queryString().field( "description" )
+							.matching(
+									"robots +(crime || investigation || disappearance)^10 +\"investigation help\"~2 -/(dis)?a[p]+ea?ance/" ) ) // <1>
```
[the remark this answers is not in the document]
marko-bekhta: hmm right, well I guess in this particular case it wouldn't matter anyway, since it'll get converted to should clauses so I'll just remove that 😃 .... but I was looking at https://lucene.apache.org/core/9_9_2/queryparser/org/apache/lucene/queryparser/classic/package-summary.html#OR (and https://lucene.apache.org/core/9_9_2/queryparser/org/apache/lucene/queryparser/classic/package-summary.html#Escaping_Special_Characters)
and also here https://www.elastic.co/guide/en/elasticsearch/reference/8.12/query-dsl-query-string-query.html#_boolean_operators 
and I thought just to throw lots of different things into that query (even if some may not make much sense) just to show that the query can be a bit harder to read 😃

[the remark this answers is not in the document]
yrodiere: Hu, that's interesting. So `||` *is* supported. Please ignore me, then.

PR #3949, opened 2024-02-07
Do not use prepared statement to update instance column values in tests

marko-bekhta:
Sooo I was trying to reproduce that problem I mentioned in https://hibernate.zulipchat.com/#narrow/stream/132092-hibernate-search-dev/topic/HSEARCH-5004 🙈 

It turned out the problem was in our test... what was happening with MSSQL:
1. an event is saved with an instant value of `2024-02-07T16:40:00.000Z`
2. our select in prepare statement would return `2024-02-07T17:40:00.000Z` (+1h because of the timezone)
3. new value is saved back
4. the test would need to wait for an hour to find any events available 🙈 

this should let ORM figure out all the JDBC strangeness.... (detach/merge since we are changing the ids)


yrodiere (approved):

marko-bekhta:
https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search/detail/PR-3949/2/pipeline 🙈

PR #3952, opened 2024-02-08
HSEARCH-4961 API to list all analyzer names

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4961

WDYT?  😃 I've ended up adding backend-specific methods to backends since, in the case of the ES backend, we'd need to pass an index name + the name of the analyzer... so it won't match the Lucene case.. and forcing to pass index name for Lucene... I wasn't sure it would be ok 😃

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/analysis/model/impl/ElasticsearchAnalysisDefinitionRegistry.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/analysis/model/impl/ElasticsearchAnalysisDefinitionRegistry.java
@@ -7,16 +7,25 @@
 package org.hibernate.search.backend.elasticsearch.analysis.model.impl;
 
 import java.lang.invoke.MethodHandles;
+import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
 import java.util.TreeMap;
+import java.util.stream.Collectors;
 
+import org.hibernate.search.backend.elasticsearch.analysis.impl.ElasticsearchAnalysisDescriptor;
 import org.hibernate.search.backend.elasticsearch.logging.impl.Log;
 import org.hibernate.search.backend.elasticsearch.lowlevel.index.analysis.impl.AnalyzerDefinition;
 import org.hibernate.search.backend.elasticsearch.lowlevel.index.analysis.impl.CharFilterDefinition;
 import org.hibernate.search.backend.elasticsearch.lowlevel.index.analysis.impl.NormalizerDefinition;
 import org.hibernate.search.backend.elasticsearch.lowlevel.index.analysis.impl.TokenFilterDefinition;
 import org.hibernate.search.backend.elasticsearch.lowlevel.index.analysis.impl.TokenizerDefinition;
+import org.hibernate.search.engine.backend.analysis.AnalyzerDescriptor;
+import org.hibernate.search.engine.backend.analysis.NormalizerDescriptor;
+import org.hibernate.search.engine.backend.analysis.spi.AnalysisDescriptorRegistry;
 import org.hibernate.search.util.common.logging.impl.LoggerFactory;
 
 /**
@@ -26,7 +35,7 @@ import org.hibernate.search.util.common.logging.impl.LoggerFactory;
  * (see {@link #getAnalyzerDefinitions} for instance).
  *
  */
-public final class ElasticsearchAnalysisDefinitionRegistry {
+public final class ElasticsearchAnalysisDefinitionRegistry implements AnalysisDescriptorRegistry {
 
 	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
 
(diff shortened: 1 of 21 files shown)


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/ElasticsearchBackend.java
```diff
@@ -23,4 +30,22 @@ public interface ElasticsearchBackend extends Backend {
...
+	 * @return An {@link Optional#empty() empty optional} if there is no analyzer configured with the given index and name.
+	 */
+	@Incubating
+	Optional<ElasticsearchAnalyzerDescriptor> analyzerDescriptor(String indexName, String analyzerName);
+
+	/**
+	 * @return A map of built-in and configured analyzer descriptors by indexes configured for this backend.
+	 */
+	@Incubating
+	Map<String, Collection<ElasticsearchAnalyzerDescriptor>> analyzerDescriptors();
```
yrodiere, line 49: This feels dangerous, as there may be conflicts: analyzers with the same name in multiple indexes, but a different definition.

I mean right now it's okay since you only expose the analyzer name, but the possibility of such conflicts will prevent this API from evolving, and makes the whole concept of an "analyzer descriptor" pointless.

IMO you should not expose anything on ElasticsearchBackend. What you added to `IndexDescriptor` is enough, and if one day we need more, we'll make `IndexDescriptor` evolve to expose backend-specific stuff (through extensions or other).

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/LuceneBackend.java
```diff
@@ -28,4 +31,19 @@ public interface LuceneBackend extends Backend {
...
+	 */
+	@Incubating
+	Optional<LuceneAnalyzerDescriptor> analyzerDescriptor(String name);
+
+	/**
+	 * @return A collection of built-in and configured analyzer descriptors available to this backend.
+	 */
+	@Incubating
+	Collection<LuceneAnalyzerDescriptor> analyzerDescriptors();
+
```
yrodiere, line 48: IMO this is not really necessary as `IndexDescriptor` + the current `analyzer(...)` methods already expose all one could need.

But if you feel strongly about it... :shrug:

engine/src/main/java/org/hibernate/search/engine/backend/metamodel/IndexDescriptor.java
```diff
@@ -52,4 +54,19 @@ public interface IndexDescriptor {
 	 */
 	Collection<IndexFieldDescriptor> staticFields();
 
+	/**
+	 * Looks up the built-in and configured analyzers available to the index represented by this descriptor.
```
yrodiere, line 58: Makes me wonder... we'd need the same thing for normalizers, no?

engine/src/main/java/org/hibernate/search/engine/backend/metamodel/IndexDescriptor.java
```diff
@@ -52,4 +54,19 @@ public interface IndexDescriptor {
...
+	 *
+	 * @param name The name of the analyzer.
+	 * @return An {@link Optional#empty() empty optional} if there is no analyzer configured with the given name.
+	 */
+	@Incubating
+	Optional<? extends AnalyzerDescriptor> analyzerDescriptor(String name);
```
yrodiere, line 64: Why not just `analyzer(...)`? We don't use the `Decriptor` suffix anywhere else in the metamodel methods.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/ElasticsearchBackend.java
marko-bekhta: > analyzers with the same name in multiple indexes

right... that's why I thought a map <index-name, list-of-analyzers> would work here, but anyways 😃 I'll drop the backend API changes

engine/src/main/java/org/hibernate/search/engine/backend/metamodel/IndexDescriptor.java
marko-bekhta: yeah, that's because of this one

https://github.com/hibernate/hibernate-search/blob/4a8b3af583676d3f71f4475d0a8fa66b87079124/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/LuceneBackend.java#L23

and since I wanted to keep the same names for backend/index descriptor ... 🙈 but with backend stuff gone it should be ok 😃

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/analysis/model/impl/ElasticsearchAnalysisDefinitionRegistry.java
```diff
@@ -110,4 +119,40 @@ public Map<String, CharFilterDefinition> getCharFilterDefinitions() {
 		return Collections.unmodifiableMap( charFilterDefinitions );
 	}
 
+	@Override
+	public Optional<ElasticsearchAnalyzerDescriptor> analyzerDescriptor(String name) {
+		if ( analyzerDefinitions.containsKey( name ) ) {
+			return Optional.of( new ElasticsearchAnalysisDescriptor( name ) );
+		}
+		else {
+			return Optional.ofNullable( ElasticsearchAnalysisDescriptor.BUILT_IN.get( name ) );
```
yrodiere, line 128: I'm not sure it's a good idea to expose built-in analyzers in metadata when we have no idea what they look like on the server...

For metadata, it would probably be safer to only expose what is actually defined, not what *might* exist on the server?

Note it's not the same with Lucene, because with that backend we actually have a full view of which analyzers are defined.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/analysis/LuceneAnalyzerDescriptor.java
```diff
@@ -0,0 +1,24 @@
...
+import org.hibernate.search.engine.backend.analysis.AnalyzerDescriptor;
+import org.hibernate.search.util.common.annotation.Incubating;
+
+import org.apache.lucene.analysis.Analyzer;
+
+/**
+ * The Lucene backend-specific extension of the analyzer descriptor.
+ */
+@Incubating
+public interface LuceneAnalyzerDescriptor extends AnalyzerDescriptor {
```
yrodiere, line 18: I couldn't find integration tests for this... How do you intend to expose it to users? Ask them to cast? :/

Unless you intend to provide some `.extension` method on the index descriptor, I'd say those interfaces should remain implementation for now (or disappear if they are not useful).

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/analysis/LuceneAnalyzerDescriptor.java
```diff
@@ -0,0 +1,24 @@
...
+
+	/**
+	 * @return The Lucene analyzer described by this descriptor.
+	 */
+	Analyzer analyzer();
```
yrodiere, line 23: This means people will end up writing `indexDescriptor.analyzer("foo").analyzer()`, no?

Maybe `instance()` would be a better name?

Though if you're going to remove this interface, I guess it doesn't matter.

yrodiere (approved):
Thanks. LGTM, though I'm really not sure about the "built-in" stuff on Elasticsearch.

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/testsupport/util/ElasticsearchTckTestRunner.java
```diff
@@ -26,6 +26,6 @@
 // Defines a "root" package, subpackages are included. Use Include/Exclude ClassNamePatterns annotations to limit the executed tests:
 @SelectPackages("org.hibernate.search.integrationtest.backend.tck")
 // Default class pattern does not include IT tests, hence we want to customize it a bit:
-@IncludeClassNamePatterns({ ".*Test", ".*IT" })
+@IncludeClassNamePatterns({ ".*Analysis.*IT" })
```
yrodiere, line 29: Probably shouldn't be included.

PR #3953, opened 2024-02-08
HSEARCH-5079 Test for indexing a @JdbcTypeCode(SqlTypes.JSON)

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5079

Hey, I thought I'd try it out. (probably would be better to simply include Gson as a dependency rather than doing the regexp though 😄) and it seems to work fine


yrodiere (approved):

yrodiere:
Well that's good news :)

> Hey, I thought I'd try it out. (probably would be better to simply include Gson as a dependency rather than doing the regexp though 😄) and it seems to work fine

I agree it's a bit rough, but I like how your solution spares us the dependency: the more dependencies, the less we're sure Hibernate Search works without any of them...

If you find a simpler way to do that (some test utils we already have in the classpath? Apache stuff, or JsonAssert stuff?), that's great, but otherwise I'd merge as-is.

Thanks!

marko-bekhta:
There's probably something with the `FormatMapper`... I've tried it yesterday with gson on Postgtres and all was good 🙈 
I'll take a look once I'm done with the review for your PR 🙈 😃

yrodiere:
Can I assume this is ready for review @marko-bekhta ? Since it was... you know... reviewed? :)

marko-bekhta:
and the build for DBs seems to pass https://ci.hibernate.org/job/hibernate-search/job/PR-3953/7/pipeline-graph/

yrodiere:
Merged, thanks!

PR #3964, opened 2024-02-13
HSEARCH-5039 Tenant filter should be a pre-filter for knn predicates

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5039

I thought I'd add a couple of tests and ... well. ... . .  😃 
The outer tenant filter is applied "too late". It finds similar vectors among all tenants, and only then it applies the tenant filter, returning fewer results than the user expected.

With the Elasticsearch backend, the change was much cleaner, while with the Lucene one... needed to delay converting some things to Lucene queries 😞 

I also noticed an unused method on the collector interface, so I've dropped it since it is in the impl package.

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
@@ -880,4 +880,9 @@ public interface Log extends BasicLogger {
 					+ " Try using a different similarity type and refer to the OpenSearch documentation for more details.")
 	SearchException vectorSimilarityNotSupportedByOpenSearchBackend(VectorSimilarity vectorSimilarity);
 
+	@LogMessage(level = Level.WARN)
+	@Message(id = ID_OFFSET + 189,
+			value = "Using a knn predicate in the nested context when tenant or routing filters are required "
+					+ "will lead to unpredictable results and may return fewer documents then expected.")
+	void knnUsedInNestedContextRequiresFilters();
 }
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchKnnPredicate.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchKnnPredicate.java
@@ -13,6 +13,7 @@ import org.hibernate.search.backend.elasticsearch.gson.impl.JsonAccessor;
 import org.hibernate.search.backend.elasticsearch.gson.impl.JsonArrayAccessor;
 import org.hibernate.search.backend.elasticsearch.gson.impl.JsonObjectAccessor;
 import org.hibernate.search.backend.elasticsearch.logging.impl.Log;
+import org.hibernate.search.backend.elasticsearch.lowlevel.query.impl.Queries;
 import org.hibernate.search.backend.elasticsearch.search.common.impl.AbstractElasticsearchCodecAwareSearchQueryElementFactory;
 import org.hibernate.search.backend.elasticsearch.search.common.impl.ElasticsearchSearchIndexScope;
 import org.hibernate.search.backend.elasticsearch.search.common.impl.ElasticsearchSearchIndexValueFieldContext;
@@ -43,6 +44,19 @@ public abstract class ElasticsearchKnnPredicate extends AbstractElasticsearchSin
 		this.similarity = builder.similarity;
 		builder.filter = null;
 		builder.vector = null;
+	}
+
+	protected JsonObject prepareFilter(PredicateRequestContext context) {
+		String tenantIdentifier = context.getTenantId();
+		if ( tenantIdentifier != null ) {
+			JsonObject tenantFilter = context.getSearchIndexScope().filterOrNull( tenantIdentifier );
+			if ( tenantFilter != null ) {
+				JsonArray filters = new JsonArray();
+				filters.add( tenantFilter );
+				return filter == null ? tenantFilter : Queries.boolFilter( filter.toJsonQuery( context ), filters );
+			}
+		}
+		return filter == null ? null : filter.toJsonQuery( context );
 	}
 
 	public static class Elasticsearch812Factory<F>
@@ -163,8 +177,9 @@ public abstract class ElasticsearchKnnPredicate extends AbstractElasticsearchSin
 			NUM_CANDIDATES_ACCESSOR.set( innerObject, k );
 			VECTOR_ACCESSOR.set( innerObject, vector );
 
+			JsonObject filter = prepareFilter( context );
 			if ( filter != null ) {
-				FILTER_ACCESSOR.set( innerObject, filter.toJsonQuery( context ) );
+				FILTER_ACCESSOR.set( innerObject, filter );
 			}
 			if ( similarity != null ) {
 				SIMILARITY_ACCESSOR.set( innerObject, similarity );
@@ -210,8 +225,9 @@ public abstract class ElasticsearchKnnPredicate extends AbstractElasticsearchSin
 			KNN_ACCESSOR.set( outerObject, field );
 
 			field.add( absoluteFieldPath, innerObject );
+			JsonObject filter = prepareFilter( context );
 			if ( filter != null ) {
-				FILTER_ACCESSOR.set( innerObject, filter.toJsonQuery( context ) );
+				FILTER_ACCESSOR.set( innerObject, filter );
 			}
 			K_ACCESSOR.set( innerObject, k );
 			VECTOR_ACCESSOR.set( innerObject, vector );
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchKnnPredicate.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchKnnPredicate.java
@@ -49,7 +49,13 @@ public abstract class ElasticsearchKnnPredicate extends AbstractElasticsearchSin
 	protected JsonObject prepareFilter(PredicateRequestContext context) {
 		JsonObject mainFilter = filter == null ? null : filter.toJsonQuery( context );
 		JsonArray filters = context.tenantAndRoutingFilters();
-		return Queries.boolCombineMust( mainFilter, filters );
+		if ( context.getNestedPath() == null ) {
+			return Queries.boolCombineMust( mainFilter, filters );
+		}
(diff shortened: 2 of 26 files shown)


marko-bekhta:
And I guess the routing filter is in the same situation, right?

yrodiere:
> And I guess the routing filter is in the same situation, right?

Now that you mention it, it probably is, yes. Would it be a big change to handle that in this PR as well?

yrodiere (approved):
Thanks. I added a few suggestions below, I'll let you be the judge of which make sense. Feel free to merge when you're satisfied.

yrodiere:
Oh and there's the routing filter as well, of course:

> > And I guess the routing filter is in the same situation, right?
> 
> Now that you mention it, it probably is, yes. Would it be a big change to handle that in this PR as well?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchKnnPredicate.java
```diff
@@ -45,6 +46,12 @@ private ElasticsearchKnnPredicate(AbstractKnnBuilder<?> builder) {
 		builder.vector = null;
 	}
 
+	protected JsonObject prepareFilter(PredicateRequestContext context) {
+		JsonObject mainFilter = filter == null ? null : filter.toJsonQuery( context );
+		JsonArray filters = context.tenantAndRoutingFilters();
+		return Queries.boolCombineMust( mainFilter, filters );
```
marko-bekhta, line 52: This works OK for the regular case when knn is not nested. 
As soon as we add a knn predicate to a nested one ... 😞 the results are all over the place between the backends..
for Lucene the results seem fine, with Elasticsearch -- it sometimes works, sometimes it fails with 
```json
{
        "type": "illegal_state_exception",
        "reason": "Parent query must not match any docs besides parent filter. Combine them as must (+) and must-not (-) clauses to find a problem doc. docID\u003d8"
      }
```
and OpenSearch "ignores" the tenant and routes i.e. it doesn't fail, but it doesn't apply the filtering correctly. 

Soooo... I'm thinking if it would make sense to include the tenant id and route key in the nested objects so we can use them for filtering.  would it make sense to go this route, or maybe you have some other idea to test?

marko-bekhta: I thought about logging/throwing an exception if we need to apply tenant/route filters and the knn is nested... but ... didn't want to add new limitations for this predicate as we've just removed some of them 😞

yrodiere: > I'm thinking if it would make sense to include the tenant id and route key in the nested objects so we can use them for filtering

You mean automatically adding the tenant ID/routing filter to `nested` predicates?

I wouldn't have expected that to be necessary, but... if it is, why not. I don't think it will hurt.

marko-bekhta: not really 🙈 😃. Initially, I thought that this should work for a nested query:

```json
{
  "query": {
    "bool": {
      "must": {
        "nested": {
          "path": "nestedObject",
          "query": {
            "knn": {
              "field": "nestedObject.floats",
              "num_candidates": 1,
              "query_vector": [
                1000.0,
                1000.0
              ]
            }
          }
        }
      },
      "filter": [
        {
          "term": {
            "_tenant_id": "tenant_1"
          }
        }
      ]
    }
  }
}
```
Having a filter at the root level and no additional filter in the knn that lives in the nested query... but that does not work, meaning having two docs:
```
1. tenant: tenant_1   nestedObject.floats: [10.0, 10.0]
2. tenant: tenant_2   nestedObject.floats: [1000.0, 1000.0]
```
and running that search query for `tenant_1` results in ... no matches 😞 
what I was considering: adding a "system" field to `nestedObject` -- `nestedObject._tenant_id` so that we'd be able to say:
```json
{
  "query": {
    "bool": {
      "must": {
        "nested": {
          "path": "nestedObject",
          "query": {
            "knn": {
              "field": "nestedObject.floats",
              "num_candidates": 1,
              "query_vector": [
                1000.0,
                1000.0
              ],
              "filter": [
                {
                  "term": {
                    "nestedObject._tenant_id": "tenant_1"
                  }
                }
              ]
            }
          }
        }
      },
      "filter": [
        {
          "term": {
            "_tenant_id": "tenant_1"
          }
        }
      ]
    }
  }
}
```

yrodiere: That's... reasonable. And also an annoying breaking change :D 

I think in the short term your only option is to document that with Elasticsearch/OpenSearch, knn inside nested fields with multitenancy is not supported.

You might try to report this as a bug to Elasticsearch/OpenSearch; too. It seems strange that in the context of a nested query, the knn query doesn't apply only to children matched by the parent query. But I think this is unlikely to get wide attention, unfortunately...

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/sort/impl/AbstractLuceneDocumentValueSort.java
```diff
@@ -32,24 +34,27 @@ public abstract class AbstractLuceneDocumentValueSort extends AbstractLuceneReve
...
 
-	protected final SortField sortField;
+	protected final Function<LuceneSearchSortContext, LuceneFieldComparatorSource> sortFieldCreator;
+	private final String absoluteFieldPath;
 
 	protected AbstractLuceneDocumentValueSort(AbstractBuilder builder) {
 		super( builder );
-		LuceneFieldComparatorSource fieldComparatorSource = builder.toFieldComparatorSource();
-		sortField = new SortField( builder.absoluteFieldPath, fieldComparatorSource, order == SortOrder.DESC );
+		sortFieldCreator = builder::toFieldComparatorSource;
```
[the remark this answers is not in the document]
yrodiere: You just shifted the problem around xD. In things like this:

https://github.com/hibernate/hibernate-search/pull/3964/files#diff-28633312525f6cf27d6995d8708c75770f7a91529e01c66fb19e3b55a2b242beR105-R108

```
		protected Function<SortRequestContext, LuceneFieldComparatorSource> toFieldComparatorSource() {
			return context -> new LuceneGeoPointDistanceComparatorSource( nestedDocumentPath, center,
					getEffectiveMissingValue(),
					getMultiValueMode(), getNestedFilter( context )
```

... the lambda still holds a reference to the builder, so that it'll be able to call `getEffectiveMissingValue()`.

Really, you'll need to either move `toFieldComparatorSource` to the sort itself (passing additional arguments to the constructor), or to call these `getEffectiveMissingValue()` methods outside of the lambda to avoid capturing a reference to the builder. Though the latter seems more prone to future mistakes.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/logging/impl/Log.java
```diff
@@ -734,4 +734,8 @@ SearchException vectorKnnMatchVectorTypeDiffersFromField(String absoluteFieldPat
 
 	@Message(id = ID_OFFSET + 180, value = "An error occurred while parsing the query string '%1$s': %2$s")
 	SearchException queryStringParseException(String query, String message, @Cause Exception cause);
+
+	@Message(id = ID_OFFSET + 181,
+			value = "An unexpected attempt to create additional predicate filters (tenant/routing etc.).")
+	SearchException predicateRequestContextUnawareOfFilters();
```
yrodiere, line 740: If this can result from a user errors, we need an actionable message: what did they do wrong, how to make it right?
If this cannot result from a user error, `AssertionFailure` would probably be a better fit (and wouldn't need to be declared in the logger).

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -879,4 +879,10 @@ SearchException vectorKnnMatchVectorTypeDiffersFromField(String absoluteFieldPat
 			value = "The OpenSearch distribution does not allow using %1$s as a space type for a Lucene engine."
 					+ " Try using a different similarity type and refer to the OpenSearch documentation for more details.")
 	SearchException vectorSimilarityNotSupportedByOpenSearchBackend(VectorSimilarity vectorSimilarity);
+
+	@LogMessage(level = Level.WARN)
+	@Message(id = ID_OFFSET + 189,
+			value = "Using a knn predicate in the nested context when tenant or routing filters are required "
+					+ "will lead to unpredictable results and may return less documents then expected.")
```
yrodiere, line 886:
```suggestion
					+ "will lead to unpredictable results and may return fewer documents then expected.")
```

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchKnnPredicate.java
```diff
@@ -45,6 +46,18 @@ private ElasticsearchKnnPredicate(AbstractKnnBuilder<?> builder) {
 		builder.vector = null;
 	}
 
+	protected JsonObject prepareFilter(PredicateRequestContext context) {
+		JsonObject mainFilter = filter == null ? null : filter.toJsonQuery( context );
+		JsonArray filters = context.tenantAndRoutingFilters();
+		if ( context.getNestedPath() == null ) {
+			return Queries.boolCombineMust( mainFilter, filters );
+		}
+		else if ( filters != null ) {
```
yrodiere, line 55: From what I can see, `filters` will never be null... ?

```java
	public JsonArray tenantAndRoutingFilters() {
		JsonArray filters = new JsonArray();
		JsonObject filter = searchIndexScope.filterOrNull( sessionContext.tenantIdentifier() );
		if ( filter != null ) {
			filters.add( filter );
		}
		if ( !routingKeys.isEmpty() ) {
			filters.add( Queries.anyTerm( "_routing", routingKeys ) );
		}
		return filters;
	}
```

yrodiere (approved):
Thanks, I think there are a few oversights, but feel free to merge when you're confident you've addressed them.

yrodiere:
@marko-bekhta Merged, thanks; can you backport this please?

marko-bekhta:
yes 🏃🏻‍♂️ 🏃🏻‍♂️ 🏃🏻‍♂️ 🏃🏻‍♂️ 😄
thansk!

PR #3966, opened 2024-02-14
HSEARCH-5081 Do not trace-log in HashTableUniformityTest

marko-bekhta:
that makes it run in ~10-12sec compared to ~3-3.5min with tracing enabled on my workstation 🙈

### DIFF
--- a/util/common/src/test/resources/log4j2.properties
+++ b/util/common/src/test/resources/log4j2.properties
@@ -14,6 +14,15 @@ rootLogger.appenderRef.stdout.ref=STDOUT
 logger.jboss.name=org.jboss
 logger.jboss.level=info
 
+### Search logger for test utils: org.hibernate.search.util.impl.test.logging.Log
+# NOTE: we want to have a separate configuration for this particular logger,
+# since otherwise it will pick up the configs from `logger.search.name`
+# resulting in enabled trace logging in tests like HashTableUniformityTest.
+# And that will slow things down too much!
+logger.searchutil.name=org.hibernate.search.util.impl.test.logging
+# See the appender threshold above to actually log trace events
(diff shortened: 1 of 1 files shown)


yrodiere (approved):
Great, thanks for taking care of that!

Do you know if raising the log level for the rest of Hibernate Search results in a significantly faster build overall? Or maybe in some tests at least?

That would probably be worth some investigation, but later: we can merge this right now.

marko-bekhta:
I've tried to run:
```bash
mvn clean install -am -pl :hibernate-search-integrationtest-backend-lucene,:hibernate-search-integrationtest-mapper-orm,:hibernate-search-documentation
```

with debug and info logs, and in both cases, it was around 11mins... I guess it makes the difference only if there are a lot of iterations, and we do not have that many tests where we add lots of documents or something similar...

I did add the commit to up the levels to match the appender one, as the lower level ones would be ignored anyway.

util/common/src/test/resources/log4j2.properties
```diff
@@ -14,12 +14,21 @@ rootLogger.appenderRef.stdout.ref=STDOUT
...
+# And that will slow things down too much!
+logger.searchutil.name=org.hibernate.search.util.impl.test.logging
+# See the appender threshold above to actually log trace events
+logger.searchutil.level=info
+
 ### Search logger
 logger.search.name=org.hibernate.search
 # See the appender threshold above to actually log trace events
-logger.search.level=trace
+logger.search.level=debug
```
yrodiere, line 29: The point of this was to allow us to enable tracing everywhere, just by changing the appender level.

If you want to remove that trick, fine, but then maybe remove the threshold filter/level from the appender.

PR #3967, opened 2024-02-16
HSEARCH-4963 API to run analysis on a given String

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4963

I've added a new incubating interface `AnalysisPerformer` to run these analyzers/normalizers and then added it to the IndexManager, since I didn't feel that the index descriptor would be a nice place for these methods. I also went with the 
```java
List<? extends AnalysisToken> analyze(String analyzerName, String terms)
```
instead of a simple list of strings, so that we give a bit more info than the tokens themselves. 

As for the docs, it didn't feel right to add this to the metamodel section, and as there's no non-backend-specific section on analysis I've "duplicated" the section in Lucene/Elasticsearch backends...


documentation/src/main/asciidoc/public/reference/_backend-lucene.adoc
```diff
@@ -783,6 +783,40 @@ see the javadoc of `Similarity` and Lucene's source code.
 You can also find useful resources on the web, for example in Elasticsearch's documentation.
 ====
 
+[[backend-lucene-analysis-run-analysis]]
+=== Applying configured analyzers/normalizers to a string
```
yrodiere, line 787: I get where you're coming from, but it feels like a shame that we end up duplicating content.

What do you think of doing this instead:

* Introduce a top-level section called "Explicit backend/index operations", perhaps just after "Searching" and before "Lucene backend"
* Introduce a subsection there called "Applying analyzers/normalizers to a string"

Then we could move more stuff there. For example, these look like they could belong in this new section:

* https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#mapper-orm-indexing-manual-largescale
* https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#backend-lucene-access-analyzers
* https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#backend-lucene-access-size
* https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#backend-lucene-access-index-reader
* https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#backend-elasticsearch-access-client

I've always thought it was weird that we put backend-specific predicates in the predicates section, but backend-specific operations in the backend sections... That would solve it.

We can totally do the moving of all these sections in a separate PR, though.

documentation/src/main/asciidoc/public/reference/_backend-lucene.adoc
```diff
@@ -783,6 +783,40 @@ see the javadoc of `Similarity` and Lucene's source code.
...
+
+include::../components/_incubating-warning.adoc[]
+
+Hibernate Search provides an API that runs a configured analyzer/normalizer against a given string.
```
yrodiere, line 791: It should work with built-in analyzers, too. Also I'm not sure one "runs" an analyzer?

```suggestion
Hibernate Search provides an API that applies an analyzer/normalizer to a given string.
```

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/impl/AnalyzeWork.java
```diff
@@ -0,0 +1,149 @@
...
+		JsonObject body = response.body();
+
+		// we don't use Gson to keep ElasticsearchAnalysisToken immutable.
+		List<ElasticsearchAnalysisToken> tokens = new ArrayList<>();
+		for ( JsonElement element : TOKENS_ACCESSOR.getOrCreate( body, JsonArray::new ) ) {
+			JsonObject token = element.getAsJsonObject();
+			tokens.add( new ElasticsearchAnalysisToken(
+					TOKEN_ACCESSOR.get( token ).orElse( "" ),
+					START_OFFSET_ACCESSOR.get( token ).orElse( Integer.MIN_VALUE ),
+					END_OFFSET_ACCESSOR.get( token ).orElse( Integer.MIN_VALUE ),
```
yrodiere, line 52: If you have to rely on this... Maybe startOffset/endOffset should be `OptionalInt`?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/impl/AnalyzeWork.java
```diff
@@ -0,0 +1,149 @@
...
+	protected List<? extends AnalysisToken> generateResult(ElasticsearchWorkExecutionContext context,
+			ElasticsearchResponse response) {
+		JsonObject body = response.body();
+
+		// we don't use Gson to keep ElasticsearchAnalysisToken immutable.
+		List<ElasticsearchAnalysisToken> tokens = new ArrayList<>();
+		for ( JsonElement element : TOKENS_ACCESSOR.getOrCreate( body, JsonArray::new ) ) {
+			JsonObject token = element.getAsJsonObject();
+			tokens.add( new ElasticsearchAnalysisToken(
+					TOKEN_ACCESSOR.get( token ).orElse( "" ),
```
yrodiere, line 50: Is this even possible? If not I'd use `orElseThrow` here, and if yes... I don't know, I'd like to understand why.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/impl/AnalyzeWork.java
```diff
@@ -0,0 +1,149 @@
...
+					START_OFFSET_ACCESSOR.get( token ).orElse( Integer.MIN_VALUE ),
+					END_OFFSET_ACCESSOR.get( token ).orElse( Integer.MIN_VALUE ),
+					TYPE_ACCESSOR.get( token ).orElse( "" )
```
yrodiere, line 53: Is this even possible? If not I'd use `orElseThrow` here, and if yes... I don't know, I'd like to understand why.

engine/src/main/java/org/hibernate/search/engine/backend/index/IndexManager.java
```diff
@@ -28,6 +29,12 @@ public interface IndexManager {
 	 */
 	IndexDescriptor descriptor();
 
+	/**
+	 * @return The object to perform analysis operations like analyzing or normalizing a given string using the configured
+	 * analyzers/normalizer for this index.
+	 */
+	AnalysisPerformer analysisPerformer();
```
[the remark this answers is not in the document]
yrodiere: Also, if you keep this method, I think you need an `@Incubating` annotation.

yrodiere (commented):
Thanks, I added a few comments below.

documentation/src/main/asciidoc/public/reference/_backend-lucene.adoc
marko-bekhta: ok I've made an attempt on moving these around 😃 🙈

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/impl/AnalyzeWork.java
marko-bekhta: no, I don't think that it is. I was just hesitant to throw an exception here. And was thinking that if for some unknown reason, it won't be there we'd still try to give a user the parts that are there 🙈

engine/src/main/java/org/hibernate/search/engine/backend/index/IndexManager.java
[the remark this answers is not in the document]
marko-bekhta: I've removed the interface from the public API but kept the Lucene/ES impls in their own classes to not make the index impl classes to complex 😃

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/index/impl/ElasticsearchIndexManagerImpl.java
```diff
@@ -264,6 +263,13 @@ clazz, ElasticsearchIndexManager.class, getBackendAndIndexEventContext()
 		);
 	}
 
+	private ElasticsearchAnalysisPerformer analysisPerformer() {
+		if ( analysisPerformer == null ) {
+			analysisPerformer = backendContext.createAnalysisPerformer( model );
+		}
+		return analysisPerformer;
+	}
```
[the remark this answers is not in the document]
marko-bekhta: hmm I didn't think that IndexManager should be safe... 

I guess we could:
- just create an instance every time: `backendContext.createAnalysisPerformer( model ).analyze( .... )` 
- make `ElasticsearchAnalysisPerformer` a util class and add analyze/normalize methods to `backendContext` and delegate to them...
- init the `analysisPerformer` in the start method ... 
- something else?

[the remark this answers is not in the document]
yrodiere: > init the analysisPerformer in the start method ...

That's what we do for everything else, so I'd favor that option?

Util classes or just shoving everything directly into the index manager are both fine to me too.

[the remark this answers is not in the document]
marko-bekhta: ok 😃 
The last commit is optional -- I'm not super happy about context having those analyze/normalize methods, and if we don't put them there, then we'd need to add some getters for `workOrchestrator` and `workFactory` 

sooo if you do not like it either -- I'll drop it 😄

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/index/impl/ElasticsearchIndexManagerImpl.java
```diff
@@ -222,6 +230,30 @@ public ElasticsearchIndexDescriptor descriptor() {
...
+	@Override
+	public List<? extends AnalysisToken> analyze(String analyzerName, String terms) {
+		return Futures.unwrappedExceptionJoin(
+				analyzeAsync( analyzerName, terms, OperationSubmitter.rejecting() ).toCompletableFuture() );
+	}
+
+	@Override
+	public AnalysisToken normalize(String normalizerName, String terms) {
+		return Futures.unwrappedExceptionJoin(
+				normalizeAsync( normalizerName, terms, OperationSubmitter.rejecting() ).toCompletableFuture() );
```
yrodiere, line 239: If you're going to block anyway in the end, you might as well use `OperationSubmitter.blocking()`.

documentation/src/main/asciidoc/public/reference/_explicit_backend_index_operations.adoc
```diff
@@ -0,0 +1,13 @@
+[[explicit-backend-index-operations]]
+= Explicit backend/index operations
+:leveloffset: +1
+
+include::_explicit_operations_apply-analazyer-normalyzer.adoc[]
```
yrodiere, line 5: Typo in the file name.

That being said, splitting this into multiple files seems a little bit extreme... no?

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/index/impl/ElasticsearchIndexManagerImpl.java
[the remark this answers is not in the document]
yrodiere: > The last commit is optional -- I'm not super happy about context having those analyze/normalize methods, and if we don't put them there, then we'd need to add some getters for `workOrchestrator` and `workFactory`
> 
> sooo if you do not like it either -- I'll drop it 😄

+1 let's drop that last commit.

yrodiere (approved):
LGTM, just minor comments. Feel free to merge!

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/index/impl/ElasticsearchIndexManagerImpl.java
marko-bekhta: true 👍🏻 I was just following the example for `computeSizeInBytesAsync` 😃

yrodiere: `computeSizeInBytes` you mean? That one probably should be fixed as well... but it's not that important

yrodiere (approved):

PR #3971, opened 2024-02-20
HSEARCH-5086 Upgrade to Lucene 9.10.0

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5086
https://hibernate.atlassian.net/browse/HSEARCH-5084


yrodiere:
@marko-bekhta CI doesn't agree :|

yrodiere:
> Yeah 😞 and what's sad about it is that it complains about
> 
> ```
> [WARNING] /mnt/workdir/jenkins/workspace/hibernate-search_PR-3971/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/extraction/impl/ExtractionRequirements.java:[89,40] [deprecation] create(Sort,int,int) in TopFieldCollector has been deprecated
> [WARNING] /mnt/workdir/jenkins/workspace/hibernate-search_PR-3971/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/extraction/impl/LuceneCollectors.java:[96,17] [deprecation] search(Query,Collector) in IndexSearcher has been deprecated
> [WARNING] /mnt/workdir/jenkins/workspace/hibernate-search_PR-3971/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/extraction/impl/LuceneCollectors.java:[171,16] [deprecation] search(Query,Collector) in IndexSearcher has been deprecated
> ```
> 
> Which was supposed to happen in Lucene 10... but I guess 10.. 9.10 🤷🏻 😃

Well, that's fair from them since it's just a deprecation.

I think we can suppress the warnings for now. But we'll really need to work on the switch soon.

BTW, we may want to enable deprecation warnings in dependency update jobs... WDYT?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/extraction/impl/ExtractionRequirements.java
```diff
@@ -184,4 +195,54 @@ else if ( !requiredStoredFields.isEmpty() ) {
...
+		}
+
+		@Override
+		public TopFieldDocs reduce(Collection<TopFieldCollector> collectors) {
+			final TopFieldDocs[] topDocs = new TopFieldDocs[collectors.size()];
+			int i = 0;
+			for ( TopFieldCollector collector : collectors ) {
+				topDocs[i++] = collector.topDocs();
+			}
+			return TopDocs.merge( sort, offset, limit == null ? numHits : Math.min( numHits, offset + limit ), topDocs );
```
yrodiere, line 245: As far as I can see, `numHits` here is equal to `maxDocs` in `LuceneSearcherImpl`, and that is already defined as:

```
		int maxDocs = getMaxDocs( indexSearcher.getIndexReader(), offset, limit );
```

Which is:

```
	private int getMaxDocs(IndexReader reader, int offset, Integer limit) {
		if ( limit == null ) {
			return reader.maxDoc();
		}
		else if ( limit.equals( 0 ) ) {
			return 0;
		}
		else {
			return Math.min( offset + limit, reader.maxDoc() );
		}
	}
```

So... I don't think you need to carry around the limit all the way down. Just using `numHits` here should lead to the same results... I think?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/extraction/impl/HibernateSearchMultiCollectorManager.java
```diff
@@ -0,0 +1,233 @@
...
+			final int size = collectors.size();
+
+			Map<CollectorKey<?, ?>, Object> results = new HashMap<>();
+
+			int i = 0;
+			for ( Map.Entry<CollectorKey<?, ?>, CollectorManager<Collector, ?>> entry : collectorManagers.entrySet() ) {
+				List<Collector> toReduce = new ArrayList<>( size );
+				for ( Collector possiblyWrappedCollector : collectors ) {
+					Collector collector = unwrapTimeLimitingCollectorIfNecessary( possiblyWrappedCollector );
+					if ( collector instanceof MultiCollector ) {
```
yrodiere, line 113: I think this will always be true, no? Since you override `reduce` in `HibernateSearchMultiCollectorManagerSingle`.

I wouldn't bother with the check; just cast it.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/extraction/impl/HibernateSearchMultiCollectorManager.java
```diff
@@ -0,0 +1,233 @@
...
+						toReduce.add(
+								unwrapTimeLimitingCollectorIfNecessary( ( (MultiCollector) collector ).getCollectors()[i] ) );
```
yrodiere, line 115: From what I can see, unwrapping is unnecessary here? You only wrap the top-level `MultiCollector`.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/extraction/impl/HibernateSearchMultiCollectorManager.java
```diff
@@ -0,0 +1,233 @@
...
+public abstract class HibernateSearchMultiCollectorManager
+		implements CollectorManager<Collector, HibernateSearchMultiCollectorManager.MultiCollectedResults> {
+
+	protected HibernateSearchMultiCollectorManager(TimeoutManager timeoutManager) {
+		this.timeoutManager = timeoutManager;
+	}
+
+	public static HibernateSearchMultiCollectorManager create(
+			TimeoutManager timeoutManager, Map<CollectorKey<?, ?>, CollectorManager<Collector, ?>> collectorManagers) {
+		if ( collectorManagers.size() == 1 ) {
```
yrodiere, line 42: Might as well keep the comment, to understand why we're making code more complicated :x 

```suggestion
		if ( collectorManagers.size() == 1 ) {
			// avoid using a multi collector if we have just one collector,
			// as MultiCollector explicitly ignores the total hit count optimization
```

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/extraction/impl/LuceneCollectors.java
```diff
@@ -104,31 +106,34 @@ public void collectMatchingDocs(int offset, Integer limit) throws IOException {
...
 				resultTotal = SimpleSearchResultTotal.lowerBound( 0 );
 			}
 			return;
 		}
+		else {
+			extractTopDocs( offset );
+		}
 
-		extractTopDocs( topDocsCollector, offset, limit );
+		// extractTopDocs( topDocs, offset, limit );
```
yrodiere, line 136: ?

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/extraction/impl/LuceneCollectors.java
```diff
@@ -192,27 +197,20 @@ public TopDocs getTopDocs() {
...
+					: new TopDocs( topDocs.totalHits, new ScoreDoc[0] );
+
 
-		if ( limit == null ) {
-			topDocs = topDocsCollector.topDocs( offset );
-		}
-		else {
-			topDocs = topDocsCollector.topDocs( offset, limit );
 		}
 	}
```
yrodiere, line 215: A few things:

1. You removed "the code below" mentioned in the comment, so the comment does not make sense anymore.
2. Will this even work, since you're no longer calling `.topDocs(0,1)` and the comment suggests that without that, we don't get a correct total hit count? I mean we should have integration tests for this and they passed, but it's probably worth double-checking (finding the relevant integration test).
3. Is this code even necesssary nowadays, and especially after your changes? Won't the topdocs already contain empty topdocs and a correct total hit count?

yrodiere (commented):
Thanks! A few comments below.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/extraction/impl/ExtractionRequirements.java
```diff
@@ -63,19 +71,20 @@ public LuceneCollectors createCollectors(IndexSearcher indexSearcher, Query orig
...
 
-		CollectorSet.Builder collectorsForAllMatchingDocsBuilder =
-				new CollectorSet.Builder( executionContext, timeoutManager );
+		HibernateSearchMultiCollectorManager.Builder collectorsForAllMatchingDocsBuilder =
+				new HibernateSearchMultiCollectorManager.Builder( executionContext, timeoutManager );
 
 		if ( maxDocs > 0 ) {
 			if ( sort == null || isDescendingScoreSort( sort ) ) {
-				topDocsCollector = TopScoreDocCollector.create( maxDocs, totalHitCountThreshold );
+				topDocsCollectorManager = new HibernateSearchTopScoreDocCollectorManager( offset, maxDocs, null,
```
marko-bekhta, line 86: note this `null` .... 

> after – the previous doc after which matching docs will be collected.

I guess we could try to leverage that, but it would only work in the case of scrolling since,  as with a simple offset, we don't have the doc...

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/extraction/impl/LuceneCollectors.java
yrodiere: > we have this one:

Wrong link? It doesn't look like the offset is higher than the document count or limit in this test.
Keep in mind that a `null` limit is essentially no limit, so higher than anything.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/extraction/impl/ExtractionRequirements.java
yrodiere: We can't leverage that right now, it's a separate feature that would need its own API, and even then it won't change anything for users that don't explicitly opt-in that feature.

See https://hibernate.atlassian.net/browse/HSEARCH-2601

yrodiere: Also we can't use that for scrolling in all conditions since it requires unique sort keys.

Let alone the performance implications of going from an offset to a "searchAfter". I'm pretty sure the performance profile would be quite different.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/extraction/impl/LuceneCollectors.java
yrodiere: Actually scratch that, the limit doesn't even matter. What we're after is a test where the offset is higher than the number of documents in the index.

yrodiere: Why, you expected me to _scroll_? That's putting too much faith in my abilities.

Sorry :x  

Looks like the assertion you just quoted is missing something about the total hit count, but the assertion in `org.hibernate.search.integrationtest.backend.tck.search.query.SearchQueryFetchIT#fetch_offset_limit` and `org.hibernate.search.integrationtest.backend.tck.search.query.SearchQueryFetchIT#fetch_offset_limit_defaultSort` actually check the total hit count, so we're fine.

> I've dropped it and the Lucene tests still pass.

LGTM, thanks!

yrodiere (approved):

PR #3985, opened 2024-02-26
HSEARCH-5092 Make `HibernateOrmSearchQueryAdapter` compatible with a future ORM 6.5 release

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5092

Note the different method name -- it's so it does not clash with the one that is currently still visible to this adapter.


marko-bekhta:
And I'll keep the https://github.com/hibernate/hibernate-search/tree/wip/main/dependency-update/orm6.5plus branch that'll be adding the override annotation once we actually do the upgrade to ORM 6.5

yrodiere:
> And I'll keep the https://github.com/hibernate/hibernate-search/tree/wip/main/dependency-update/orm6.5plus branch that'll be adding the override annotation once we actually do the upgrade to ORM 6.5

If we don't forget about it :D Anyway, +1 that's worth a try.

yrodiere (approved):

yrodiere:
I think the failure is unrelated, but we'll need to look into this: https://hibernate.atlassian.net/browse/HSEARCH-5093

I'll restart the build

PR #3990, opened 2024-02-26
First iteration of the build cache support

gsmet:
I'm shooting a bit in the dark without the build scan support but this could be a first step, at least for local caching.

As for remote caching, we will need to set up an access key on CI but it can be done as a second step.

Note that I'm not entirely sure if you want to actually enable caching on CI, and especially local caching as it seems the workspaces can be reused.

Remote caching will in any case require additional work as there are probably some system properties that we will need to ignore for the cache to be portable across environments. But we can't really detect that without having build scans being published.

I also created https://github.com/hibernate/hibernate-release-scripts/pull/3 to make sure this is disabled when releasing.

Let's discuss this when you have time.


gsmet:
Closing in favor of a PR coming from an upstream branch.

yrodiere:
Closing for real :)

PR #3994, opened 2024-02-28
HSEARCH-5094 First iteration of the build cache support

gsmet:
@yrodiere FYI, let's work on this upstream branch now, it will be easier to debug things.


yrodiere:
So, build scans appear at the bottom of this page: https://ci.hibernate.org/job/hibernate-search/job/build-cache/17/

Which I guess is better than nothing, but not exactly straightforward. We'll need to think of something.

.mvn/extensions.xml
```diff
@@ -0,0 +1,17 @@
...
+        <groupId>com.gradle</groupId>
+        <artifactId>common-custom-user-data-maven-extension</artifactId>
+        <version>1.12.5</version>
+    </extension>
+    <extension>
+        <groupId>org.hibernate.search.develocity</groupId>
+        <artifactId>hibernate-search-develocity-extension</artifactId>
+        <version>1.0.11.Final</version>
+    </extension>
+</extensions>
```
marko-bekhta, line 17: I guess these won't be managed by Dependabot, right?

marko-bekhta (approved):
🎉 😃

.mvn/extensions.xml
yrodiere: I don't think they will. Though I doubt we'll need to update them frequently anyway.

@jprinet suggested that we add the top two as dependencies of `hibernate-search-develocity-extension`, which would somewhat mitigate the problem, at the cost of having to re-release `hibernate-search-develocity-extension` every time we need to upgrade one of these extensions.

jprinet: FYI Dependabot also monitors `.mvn/extensions.xml`.
There is a [compatiblity matrix](https://docs.gradle.com/enterprise/compatibility/#develocity_compatibility_2) between Develocity backend and `gradle-enterprise-maven-extension` though. 

Given that your instance is managed, the time between the extension publication and the backend upgrade is usually minimal. In worst case scenario, you may have some Build Scans not being published if the extension was upgraded to a not yet supported version by the backend.

Hope this clarifies.

yrodiere:
> Easier access to build scans: https://hibernate.atlassian.net/browse/HSEARCH-5102

I enabled [enriched build scans](https://plugins.jenkins.io/gradle/#plugin-content-enriched-build-scans) on Jenkins, so hopefully that should make things easier in the meantime.

PR #3995, opened 2024-02-29
WIP: HSEARCH-5019 Automatically normalize indexed vectors where necessary

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5019

super-early-pre-alpha 😃😁 only to discuss some things.

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/AbstractLuceneVectorFieldCodec.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/AbstractLuceneVectorFieldCodec.java
@@ -19,7 +19,8 @@ import org.apache.lucene.index.VectorSimilarityFunction;
 import org.apache.lucene.util.BytesRef;
 
 public abstract class AbstractLuceneVectorFieldCodec<F> implements LuceneVectorFieldCodec<F> {
-
+	// TODO: should we make this one configurable ?
+	protected static final double EPS = 0.0000000001;
 	protected final FieldType fieldType;
 	protected final VectorSimilarityFunction vectorSimilarity;
 	private final int dimension;
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneByteVectorCodec.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneByteVectorCodec.java
@@ -6,7 +6,16 @@
  */
 package org.hibernate.search.backend.lucene.types.codec.impl;
 
+import static org.apache.lucene.util.VectorUtil.cosine;
+import static org.apache.lucene.util.VectorUtil.dotProduct;
+
+import java.lang.invoke.MethodHandles;
+import java.util.function.Function;
+
+import org.hibernate.search.backend.lucene.logging.impl.Log;
 import org.hibernate.search.backend.lucene.lowlevel.codec.impl.HibernateSearchKnnVectorsFormat;
+import org.hibernate.search.util.common.AssertionFailure;
+import org.hibernate.search.util.common.logging.impl.LoggerFactory;
 
 import org.apache.lucene.document.KnnByteVectorField;
 import org.apache.lucene.index.IndexableField;
@@ -14,9 +23,47 @@ import org.apache.lucene.index.VectorEncoding;
 import org.apache.lucene.index.VectorSimilarityFunction;
 
 public class LuceneByteVectorCodec extends AbstractLuceneVectorFieldCodec<byte[]> {
+
+	private static final Log log = LoggerFactory.make( Log.class, MethodHandles.lookup() );
+
+	private final Function<byte[], byte[]> normalizer;
+
 	public LuceneByteVectorCodec(VectorSimilarityFunction vectorSimilarity, int dimension, Storage storage, Indexing indexing,
-			byte[] indexNullAsValue, HibernateSearchKnnVectorsFormat knnVectorsFormat) {
+			byte[] indexNullAsValue, Float magnitude, HibernateSearchKnnVectorsFormat knnVectorsFormat) {
 		super( vectorSimilarity, dimension, storage, indexing, indexNullAsValue, knnVectorsFormat );
+		if ( magnitude != null ) {
+			normalizer = vector -> normalize( vector, magnitude );
+		}
+		else {
+			normalizer = this::normalizeNoop;
+		}
+	}
+
+	private byte[] normalizeNoop(byte[] vector) {
+		// do nothing
+		return vector;
+	}
+
+	private byte[] normalize(byte[] vector, float magnitude) {
+		float norm = norm( vector, vectorSimilarity );
+		if ( Math.abs( norm - magnitude ) > EPS ) {
+			throw log.byteVectorLengthExpectationFailed( vector, magnitude, norm );
+		}
+		return vector;
+	}
+
+	private static float norm(byte[] vector, VectorSimilarityFunction similarityFunction) {
+		switch ( similarityFunction ) {
+			case EUCLIDEAN:
+				return (float) Math.sqrt( dotProduct( vector, vector ) );
+			case MAXIMUM_INNER_PRODUCT:
+			case DOT_PRODUCT:
+				return dotProduct( vector, vector );
+			case COSINE:
+				return cosine( vector, vector );
+			default:
+				throw new AssertionFailure( "Unknown similarity function: " + similarityFunction );
+		}
 	}
 
 	@Override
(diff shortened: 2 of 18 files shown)


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneByteVectorCodec.java
```diff
@@ -6,17 +6,64 @@
...
+			case EUCLIDEAN:
+				return (float) Math.sqrt( dotProduct( vector, vector ) );
+			case MAXIMUM_INNER_PRODUCT:
+			case DOT_PRODUCT:
+				return dotProduct( vector, vector );
+			case COSINE:
+				return cosine( vector, vector );
+			default:
+				throw new AssertionFailure( "Unknown similarity function: " + similarityFunction );
+		}
```
marko-bekhta, line 66: to double-check and if all ok then create our own utils, since these are currently from `org.apache.lucene.util.VectorUtil` ...

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/mapping/LuceneVectorFieldIT.java
```diff
@@ -54,4 +68,166 @@ void test(int dimension, int efConstruction, int m, String property, int value,
...
+		// l2norm3 == 119.4026800369238
+		assertThat( l2norm3 ).isCloseTo( magnitude, Percentage.withPercentage( 1 ) );
+		assertThat( vector3 ).contains(
+				111,
+				22,
+				22,
+				22,
+				22
+		);
+	}
```
marko-bekhta, line 232: This is why I'm not "happy" to normalize byte[] vectors ...  (that is if my understanding of it is correct)
because of the precision/data loss, we won't get the norm value we expect ...

yrodiere: I understand, but I think we need to answer this before we can take a decision:

1. Is there anything about normalizing byte vectors in OpenSearch?
2. For similarities that require normalized vectors, what magnitude do they require for byte vectors?
3. Does requring a magnitude different from whatever they require make sense in practice, and if not do we have the same problem for that magnitude? E.g. we can certainly ensure that all byte vectors have a magnitude of 255, since a `byte` magnitude cannot be more than 255, so rounding all coordinates **up** (instead of down like in this example) would do the trick.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/VectorField.java
```diff
@@ -93,6 +93,14 @@
 	 */
 	int m() default AnnotationDefaultValues.DEFAULT_M;
 
+	/**
+	 * @return The vector magnitude.
+	 * <p>
+	 * With float vectors, if set to {@code > 0} automatic normalization will be applied, i.e. {@code ||vector|| == magnitude }.
+	 * With byte vectors, if set to {@code > 0} a check will be performed that the passed vector is of required magnitude.
+	 */
+	float magnitude() default AnnotationDefaultValues.DEFAULT_MAGNITUDE;
```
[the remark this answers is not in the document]
yrodiere: I think it's fine to require normalization to be explicit, but you might want to make the checks implicit?
Maybe `@VectorField(magnitude = ..., normalization = ...)` with normalization defined as `enum Normalization { DEFAULT, CHECK, NORMALIZE }`, and `DEFAULT` effectively identical to `CHECK` for now, at least for the Lucene backend?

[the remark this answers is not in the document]
yrodiere: > `DEFAULT` effectively identical to `CHECK` for now, at least for the Lucene backend?

Scratch that, that's dumb. Better idea would be to have `DEFAULT` be equal to `CHECK` for similarities that require normalization, and `IGNORE` otherwise.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/AbstractLuceneVectorFieldCodec.java
```diff
@@ -19,7 +19,8 @@
 import org.apache.lucene.util.BytesRef;
 
 public abstract class AbstractLuceneVectorFieldCodec<F> implements LuceneVectorFieldCodec<F> {
-
+	// TODO: should we make this one configurable ?
+	protected static final double EPS = 0.0000000001;
```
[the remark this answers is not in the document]
yrodiere: I'm fine with a sensible hardcoded default, at least for a start.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/codec/impl/LuceneByteVectorCodec.java
yrodiere: Did you check that we needed client-side normalization/checks for Elasticsearch/OpenSearch? If recent versions of Elasticsearch/OpenSearch implement normalization/checks themselves, I wouldn't bother.

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/mapping/LuceneVectorFieldIT.java
marko-bekhta: > dot_product
Computes the dot product of two unit vectors. This option provides an optimized way to perform cosine similarity. 
>
> When element_type is float, all vectors must be unit length.
>
> When element_type is byte, all vectors must have the same length.

^ from what I see only `dot_product` requires vectors to be normalized. (and cosine cannot accept a vector ith zero magnitude)

> Is there anything about normalizing byte vectors in OpenSearch?

I haven't seen anything, but since OpenSearch only accepts `L2` and `Consine` (https://github.com/opensearch-project/k-NN/blob/e1a7cb7aa1a523cf75beba3446985a87682f4d2c/src/main/java/org/opensearch/knn/index/util/Lucene.java#L45) with a Lucene engine, it doesn't really need it ... 

As for Elasticsearch: it has a check for floats:

```
The [dot_product] similarity can only be used with unit-length vectors. Preview of invalid vector: [1.0, 2.0, 3.0]
```
but I can put any byte vectors in with no exceptions...

which matches the checks here : 

https://github.com/elastic/elasticsearch/blob/eeecdbf87b08ec03682dabe7def7272055119252/server/src/main/java/org/elasticsearch/index/mapper/vectors/DenseVectorFieldMapper.java#L322-L338 

> For similarities that require normalized vectors, what magnitude do they require for byte vectors?

so from that `dot_product` description `all vectors must have the same length`... my understanding is that it can be any length e.g. 23 😃 , but just all of them should be of the same length...


When I'm thinking about it, I'd expect that the model used to produce the vector should provide some guidance here, i.e. does it produce normalized vectors or not and based on that user decides what to do.. IDK if it makes sense to transform the vectors produced by the model...

yrodiere: > When I'm thinking about it, I'd expect that the model used to produce the vector should provide some guidance here, i.e. does it produce normalized vectors or not and based on that user decides what to do.. IDK if it makes sense to transform the vectors produced by the model...

Makes sense... in that case checks would be enough (i.e. HSEARCH-5048 but not HSEARCH-5019)... but! We need an expected magnitude, or we cannot run checks.

From what you say here, I understand that with some similarities at least, we can infer the expected magnitude:

> As for Elasticsearch: it has a check for floats:
> 
> ```
> The [dot_product] similarity can only be used with unit-length vectors. Preview of invalid vector: [1.0, 2.0, 3.0]
> ```

So to sum up:

* Normalization probably doesn't really make sense, so we'd skip that (close HSEARCH-5019)
* Checks may make sense so we'd keep them (HSEARCH-5048)
* We'd enable checks by default depending on the configured similarity -- I think only for `dot_product` with `float` vectors?
* We may want to provide a way to disable checks (see other conversation about an enum)
* We may want to keep the custom magnitude and a way to enable checks explicitly, but we're not sure that really makes sense, so maybe we shouldn't.

marko-bekhta: Thanks! I've started to look into the checking part .... 

> We may want to provide a way to disable checks (see other conversation about an enum)

I guess we won't be able to disable it for elasticsearch backends ... as it will fail on the server side, so maybe start with just adding a check to the Lucene impl and see if users request a change, oooor add a Lucene-specific backend config property to disable the checks?

yrodiere: I was more thinking of people who have normalized vectors anyway, so don't want to waste time on checks that are not useful.

But ok, +1 to not allow disabling it for a start.

Implementing it for the Elasticsearch backend would still have value, since the error message will probably be clearer, but if it's harder (e.g. you need to rely on Lucene classes), ok, let's not bother.

marko-bekhta: no no, it should be fine to implement for Elastcisearch too, I was just thinking that for Elasticsearch we'd be doing the check twice (the second one happening on the server side), and if a user disables it on our side, Elasticsearch will still fail ...
maybe we just need to come up with a name that tells the user that it happens before we send the data to backend? but I don't have a good example of such name 🙈

yrodiere: > ok, let's not bother.

:)

marko-bekhta: https://github.com/hibernate/hibernate-search/pull/4022

PR #3997, opened 2024-03-01
HSEARCH-5006 Non-string tenant identifiers

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5006

Hey 😃 this is another draft PR. To make this work, we have to update the user-facing methods like `massindexer(teantIds)` and others so that they can accept `Object` tenant IDs. Then, on the consumption side, we have to either open the session (in ORM), where we need an Object value, or we want to write into the index, where we need the String value. 
So, I was thinking that the `TenancyConfiguration` could be a good place to keep the converter logic in. 
Also, after a few back-and-forths with these tenant ids ... I'd like it if we had the value object for the tenant id rather than a string or object... but the change is spreading through a lot of classes, so I've limited it to only places that are needed so that it reaches the ORM ... But since the changes are breaking some of the API/SPI anyway, should we replace it everywhere? 
also should we try limiting the changes to ORM mapper or apply them to the standalone one as well?

-------------------------
Idea v2 😃 
In this iteration the tenant identifier is converted to string as soon as we receive it from the user and is passed around as string everywhere. This way we have a string ready for indexing, and in those few cases where a session should be opened the string is converted back to an object representation....


mapper/orm/src/main/java/org/hibernate/search/mapper/orm/tenancy/spi/TenancyConfiguration.java
```diff
@@ -34,41 +41,122 @@ public class TenancyConfiguration {
...
+		}
+
+		@Override
+		public TenantIdentifier<?> convert(Object tenantIds) {
+			return TenantIdentifier.NO_TENANT;
+		}
+	}
+
+	private abstract static class Multi extends TenancyConfiguration {
+		private final BeanHolder<? extends TenantIdentifierConverter> converterBeanHolder;
```
yrodiere, line 85: A `BeanHolder` must be closed eventually.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/tenancy/spi/TenancyConfiguration.java
```diff
@@ -34,41 +41,122 @@ public class TenancyConfiguration {
...
 		switch ( tenancyMode ) {
 			case SINGLE_TENANCY:
-				return new TenancyConfiguration( Optional.of( Collections.emptySet() ),
-						tenantIdsConfigurationPropertyKey );
+				return new Single();
 			case MULTI_TENANCY:
-				return new TenancyConfiguration(
-						MULTI_TENANCY_TENANT_IDS.getAndMap( configurationPropertySource, LinkedHashSet::new ),
-						tenantIdsConfigurationPropertyKey );
+				// get from property:
```
yrodiere, line 51: I guess this is a TODO?

I agree it would make sense to make this configurable, though I wonder if we couldn't also provide a decent default.

`toString()` is not the way to go, that's for sure, but assuming we can get some metadata from Hibernate ORM, we might be able to use something like `org.hibernate.type.descriptor.java.JavaType#toString(T)`/`org.hibernate.type.descriptor.java.JavaType#fromString(String)`...


I wonder if we actually need to make this configurable...

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/tenancy/spi/TenancyConfiguration.java
```diff
@@ -34,41 +41,122 @@ public class TenancyConfiguration {
...
+		}
+
+		@Override
+		public SearchException invalidTenantId(TenantIdentifier<?> tenantId) {
+			throw new AssertionFailure( "Single tenancy is used." );
+		}
+
+		@Override
+		public TenantIdentifier<?> convert(Object tenantIds) {
+			return TenantIdentifier.NO_TENANT;
```
yrodiere, line 80: This is a good idea, but I wonder how many `tenantIdentifier() != null` we have out there... we'll need to catch them all and replace them with `tenantIdentifier().isEmpty()` or something like that, otherwise we'll introduce many bugs for people using single-tenancy, which is most people.

yrodiere (commented):
Thanks! Here are a few comments.

> To make this work, we have to update the user-facing methods like `massindexer(teantIds)` and others so that they can accept `Object` tenant IDs.

Agreed, but see comments, this needs to be done differently to avoid a breaking change.

> Then, on the consumption side, we have to either open the session (in ORM), where we need an Object value, or we want to write into the index, where we need the String value.

Makes sense.

> So, I was thinking that the `TenancyConfiguration` could be a good place to keep the converter logic in.

Sure, makes sense.

> Also, after a few back-and-forths with these tenant ids ... I'd like it if we had the value object for the tenant id rather than a string or object... but the change is spreading through a lot of classes, so I've limited it to only places that are needed so that it reaches the ORM ...

Are you sure you need to keep both the string and object value everywhere?

I'd have expected that you would only need to convert from string to object when reading the configuration (because we'd ultimately create Hibernate ORM sessions from there), and from object to string when creating a HibernateSearchSession (because we'll need the string for Hibernate Search internals. I'm surprised this change has such far-reaching effects... ?

Note that it can be fine to convert from string then back to string if it's very localized and not in the hot path.

> But since the changes are breaking some of the API/SPI anyway, should we replace it everywhere?

Breaking SPIs or the incubating Standalone mapper API is fine if you can't do it any other way, but breaking non-incubating API is a no-go.

> also should we try limiting the changes to ORM mapper or apply them to the standalone one as well?

I'd say you'll probably have an easier time applying them to both.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/tenancy/spi/impl/HibernateOrmTenantIdentifierConverter.java
```diff
@@ -0,0 +1,50 @@
...
+		//	      at org.hibernate.search.util.common.impl.CancellableExecutionCompletableFuture$CompletingRunnable.run(CancellableExecutionCompletableFuture.java:70)
+		//	      at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
+		//	      at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
+		//	      at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
+		//	      at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
+		//	      at java.base/java.lang.Thread.run(Thread.java:840)
+		//
+		//     Sooooooooo it's probably best to have a simple implementation that works with strings,
+		//     and throw an exception for any other type letting the user know that that need to provide a custom impl and ...
+		return tenantId == null ? null : tenantIdentifierJavaType.fromString( tenantId );
```
marko-bekhta, line 48: I guess I'll drop this one, and users will need to provide the converter 😃

mapper/orm-jakarta-batch/core/src/main/java/org/hibernate/search/jakarta/batch/core/massindexing/util/impl/PersistenceUtil.java
```diff
@@ -70,6 +70,7 @@ public static StatelessSession openStatelessSession(EntityManagerFactory entityM
 		@SuppressWarnings("rawtypes")
 		StatelessSessionBuilder builder = sessionFactory.withStatelessOptions();
 		if ( StringHelper.isNotEmpty( tenantId ) ) {
+			// TODO: needs a converter here too!
 			builder.tenantIdentifier( (Object) tenantId );
```
marko-bekhta, line 74: I'm not sure if we could pass the converter to these batch utils...  or say something like "the converter should be a named bean "blablabla" and try injecting it as:
```java
@Inject
@Named("blablabla")
private TenantIdentifierConverter tenantIdentifierConverter;
```

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/tenancy/spi/TenancyConfiguration.java
marko-bekhta: we probably have to 😔 .... https://github.com/hibernate/hibernate-search/pull/3997#discussion_r1511047204

documentation/src/main/asciidoc/migration/index.adoc
```diff
@@ -84,13 +84,43 @@ The https://hibernate.org/community/compatibility-policy/#code-categorization[AP
 in Hibernate Search {hibernateSearchVersion}
 is backward-compatible with Hibernate Search {hibernateSearchPreviousStableVersionShort}.
 
+With Hibernate ORM allowing Object tenant identifiers some of the Hibernate Search methods got deprecated and alternatives were introduced:
```
yrodiere, line 87:
```suggestion
With Hibernate ORM allowing Object tenant identifiers some of the Hibernate Search methods got deprecated and alternatives were introduced:
```
```suggestion
With Hibernate ORM allowing non-`String` tenant identifiers, some of the Hibernate Search methods got deprecated and alternatives were introduced:
```

documentation/src/main/asciidoc/public/reference/_mapper-pojo-standalone.adoc
```diff
@@ -122,6 +122,21 @@ include::{sourcedir}/org/hibernate/search/documentation/mapper/pojo/standalone/m
 <4> Build the new session.
 ====
 
+[NOTE]
+====
+When using non-string tenant identifiers, a custom `TenantIdentifierConverter` should be provided:
```
yrodiere, line 127:
```suggestion
When using non-string tenant identifiers, a custom `TenantIdentifierConverter` must be provided:
```

yrodiere (approved):
Makes sense to me, thanks. I added a few comments on documentations, but feel free to merge.

documentation/src/main/asciidoc/public/reference/_mapper-pojo-standalone.adoc
marko-bekhta: thanks!

PR #3998, opened 2024-03-04
Allow dependency jobs to specify JDK version

marko-bekhta:
With Lucene bumping min JDK to 21.... 🙈


yrodiere:
Wonderful :| I'll start discussions downstream...
Thanks!

PR #4003, opened 2024-03-05
Follow up on HSEARCH-5062 Update regexps to address mssql queries

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5062


yrodiere (approved):

marko-bekhta:
ok, this passed on all DBs, soooo let's try merging it 😃

PR #4025, opened 2024-03-08
HSEARCH-5016 Allow binding a @HighlightProjection to a single-valued String (instead of List<String>) when using numberOfFragments(1)

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5016

### DIFF
--- a/documentation/src/main/asciidoc/public/reference/_search-dsl-projection.adoc
+++ b/documentation/src/main/asciidoc/public/reference/_search-dsl-projection.adoc
@@ -1381,8 +1381,9 @@ in some cases. For more details see <<mapping-directfieldmapping-highlightable-d
 [[search-dsl-projection-highlight-syntax]]
 === Syntax
 
-The `highlight` projection always returns a list of string values per highlighted field, no matter if the field is
-a single or a multivalued one:
+The `highlight` projection, by default, returns a list of string values per highlighted field, no matter if the field is
+a single or a multivalued one, since there can be multiple highlighted terms in a field value and depending on the
+highlighter configuration that can result in multiple text fragments with highlighted terms in them:
 
 .Returning highlights for matched documents
 ====
@@ -1405,6 +1406,20 @@ As an example, the result may look like:
 <1> First hit.
 <2> Second hit.
 <3> Third hit with multiple highlighted snippets.
+====
+
+In some scenarios, when we know that there is going to be only one highlighted fragment returned, it may be helpful to
+force the highlight projection to produce a single `String` rather than a `List<String`.
+This is only possible when the <<search-dsl-highlighting-highlighter-options-fragment-size, number of fragments>> is explicitly set to `1`.
+
+.Forcing a single-valued highlight projection
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/projection/ProjectionDslIT.java[tags=highlight-single-value]
+----
+<1> Force the single-valued highlight projection by calling `.single()`.
+<2> Force the number of fragments to `1` so that the highlighter returns a single highlighted fragment at most.
 ====
 
 [[search-dsl-projection-highlight-multivalued]]
@@ -1466,7 +1481,7 @@ See <<search-dsl-highlighting>> for more information on highlighter configuratio
 To achieve a `highlight` projection inside a <<search-dsl-projection-mapped,projection to an annotated custom type>>,
 use the `@HighlightProjection` annotation on the constructor parameter:
 
-.Returning highlights for matched documents within a projection constructor
+.Returning highlights for matched documents within a projection constructor. Multiple highlighted fragments
 ====
 [source, JAVA, indent=0, subs="+callouts"]
 ----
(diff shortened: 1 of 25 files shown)


yrodiere (commented):
Thanks, this looks good but I didn't see a test for `@HighlightProjection` on a `String` parameter (instead of `List<String>`). Did I miss it?
You might need to adjust the corresponding documentation as well.

documentation/src/main/asciidoc/public/reference/_search-dsl-projection.adoc
```diff
@@ -1502,6 +1503,29 @@ include::{sourcedirJava17}/org/hibernate/search/documentation/search/projection/
...
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedirJava17}/org/hibernate/search/documentation/search/projection/ProjectionDslJava17IT.java[tags=projection-constructor-highlight-single]
+----
+<1> Pass the custom projection type to `.select(...)`.
+<2> Configure the highlighter to return a single highlighted fragment.
+<3> Each hit will be an instance of the custom projection type,
+populated with the requested highlight and field.
+====
+
```
marko-bekhta, line 1528: I didn't want to add a single highlight to the previous example as that would mean that we'd need to make a query more complex (matching on multiple fields)
But then I can't say that I really like this approach with two similar examples 🙈 ...

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/highlight/AbstractHighlighterIT.java
```diff
@@ -426,6 +426,23 @@ void numberOfFragmentsSingle() {
 				);
 	}
 
+	@Test
+	void numberOfFragmentsSingleNamedHighlighter() {
+		StubMappingScope scope = index.createScope();
+
+		SearchQuery<String> highlights = scope.query().select(
+				f -> f.highlight( "anotherString" ).single().highlighter( "single" )
```
marko-bekhta, line 434: While updating the processor I've noticed that the previous commit didn't address the case when there's a single fragment + a named highlighter, so I moved highlighter step methods around 🙈

engine/src/main/java/org/hibernate/search/engine/search/projection/dsl/CommonHighlightProjectionOptionsStep.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.engine.search.projection.dsl;
+
+import java.util.function.Function;
+
+/**
+ * The step in a highlight definition, where optional parameters can be set.
+ */
+public interface CommonHighlightProjectionOptionsStep<N extends ProjectionFinalStep<?>> {
```
yrodiere, line 14: Not sure introducing this is a good idea... the name is so-so, and it's not a big deal if people always need to call `highlighter(...)` _before_ `.single()`.

yrodiere (approved):

engine/src/main/java/org/hibernate/search/engine/search/projection/dsl/CommonHighlightProjectionOptionsStep.java
marko-bekhta: ok 😃 , maybe like this then? 
the thing is that we'd need to have a `.single()` both after calling highlighter and without calling it  🙈

yrodiere: I don't get why you don't just add the `single()` method in `HighlightProjectionOptionsStep` and make it return `ProjectionFinalStep`?

A bit like `FieldProjectionValueStep`.

Anyway... what you did will work fine as well.

marko-bekhta: it's just then `.highlighter(..)` and `single()`  will be in the same interface, and single -- must end up returning `ProjectionFinalStep<String>` while `.highlighter(.. )` can still end with `ProjectionFinalStep<List<String>>` 

and if `.highlighter(..)`  returns the `HighlightProjectionOptionsStep` step, it could work; but users would be able to call   `.highlighter(..)`  multiple times, which doesn't make much sense 🙈

yrodiere: > users would be able to call .highlighter(..) multiple times, which doesn't make much sense 🙈

I agree it's not perfect, but it's fine by me. Pretty much every single other setter in `OptionsStep` interfaces can be called multiple times, whether it makes sense or not: it's just simpler that way.

marko-bekhta: ok 👍🏻 😃

yrodiere (approved):

PR #4034, opened 2024-03-12
Downgrade plexus-compiler to 2.14.2

marko-bekhta:
Until we figure out what to do about the logging AP which may need a new release of it?


yrodiere (approved):
Sure, but you'll need to adjust dependabot config to avoid getting another upgrade

marko-bekhta:
Mm ok 😃 I was hoping just to click "close" on the PR with plexus-compiler and then hope that Dependabot would ignore that version 🤞🏻 🙈 😄 
but this should be safer 😃

yrodiere (approved):

yrodiere:
LGTM but it seems Eclipse doesn't like our code anymore...

PR #4035, opened 2024-03-12
HSEARCH-4558 WIP: simpleQueryString for numeric/date fields / HSEARCH-5109 Introduce `ValueConvert.PARSE` to provide built-in converters for string values

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4558
https://hibernate.atlassian.net/browse/HSEARCH-5109

Opening only as a draft to get some early feedback 😉 

I've tried running queries against an Elasticsearch instance to see what syntax is allowed for non-string fields with the simple query and that should match the results of `LuceneSimpleQuerySearchPredicateIT#simple()`  I have it there for now so it's just easier to debug. 

From what I've seen the impl for a regular query string, the approach would be similar, just that more queries will have to be overridden (e.g. the one for the ranges)


backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/predicate/impl/LuceneSimpleQueryStringPredicate.java
```diff
@@ -47,7 +56,54 @@ public SearchPredicate build() {
...
+				protected Query createFieldQuery(Analyzer analyzer, BooleanClause.Occur operator, String field,
+						String queryText, boolean quoted, int phraseSlop) {
+					var state = fieldStatesRemapped.get( field );
+					LuceneFieldCodec<?> codec = state.field().type().codec();
+					if ( codec instanceof AbstractLuceneNumericFieldCodec<?, ?> ) {
+						Number number = ( (AbstractLuceneNumericFieldCodec<?, ?>) codec ).fromString( queryText );
+						return ( (AbstractLuceneNumericFieldCodec<?, Number>) codec ).getDomain()
+								.createExactQuery( field, number );
+					}
+					return super.createFieldQuery( analyzer, operator, field, queryText, quoted, phraseSlop );
```
marko-bekhta, line 75: I remember that you were not too excited about adding more and more things to the codec 😃 🙈 
but we need to somehow parse the string into something we can pass to the "createQuery" one way or the other... but if you have an idea ... 😃

yrodiere: I wasn't, but in this case... I don't think there's a way around it, at least for parsing.

Though I'd suggest avoiding casts and using proper encapsulation, e.g. something like this:

```suggestion
					var state = fieldStatesRemapped.get( field );
					LuceneFieldCodec<?> codec = state.field().type().codec();
					return codec.createExactQuery( analyzer, operator, field, queryText, quoted, phraseSlop );
```

Or even, for better integration and error messages:

```suggestion
					var state = fieldStatesRemapped.get( field );
					LuceneFieldCodec<?> codec = state.field().type().codec();
					var builder state.field().queryElement(PredicateTypeKeys.MATCH, searchContext);
					Object parsed = codec.parse(queryText);
					builder.value(parsed, ValueConvert.NO);
					// ... some more setters ...
					return builder.build();
```

Or, even bolder, bring string parsing as a first-class feature available straight from the DSL; you'll still need some "parse" method in the codec, but it will be called indirectly by predicate builders:

```suggestion
					var state = fieldStatesRemapped.get( field );
					LuceneFieldCodec<?> codec = state.field().type().codec();
					var builder state.field().queryElement(PredicateTypeKeys.MATCH, searchContext);
					builder.value(queryText, ValueConvert.PARSE); // ValueConvert.PARSE is new, and means "parse value from string"
					// ... some more setters ...
					return builder.build();
```

marko-bekhta: thanks! 😃

yrodiere: Np.

For the parsing format, I'd suggest aligning with "Parsing method for 'indexNullAs'" in https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#mapping-directfieldmapping-supported-types.

Maybe we'll want to update that to say "Parsing method for 'indexNullAs' and query strings" instead :thinking: And link to there from the query string predicate's documentation.

yrodiere: > For the parsing format, I'd suggest aligning with "Parsing method for 'indexNullAs'" in https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#mapping-directfieldmapping-supported-types.

That would, of course, require that bridges be able to push their parsing method to the backend, e.g. for enums.

Maybe some `queryStringConverter(ToDocumentValueConverter<String, ? extends F>)` method you could add alongside `org.hibernate.search.engine.backend.types.dsl.IndexFieldTypeConverterStep#dslConverter`?

But with defaults for at least some of the field types, so that fields declared through something else than `ValueBridge` still have decent parsing support.

yrodiere: See also:

https://github.com/hibernate/hibernate-search/blob/7b6b15c04f0ce6ec98e8d4983605b97f36d34fd5/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/ValueBridge.java#L50-L60

https://github.com/hibernate/hibernate-search/blob/b476e2958180757b7de968a043bf7be6d1be005d/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/binding/impl/ValueBindingContextImpl.java#L164-L168

https://github.com/hibernate/hibernate-search/blob/2fd171a5962eae87bd9821d3155217b2d54182e4/documentation/src/test/java/org/hibernate/search/documentation/mapper/orm/binding/fieldtypes/dsl/dslconverter/ISBNBinder.java#L30-L33

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/AbstractElasticsearchIndexFieldTypeOptionsStep.java
```diff
@@ -40,6 +41,13 @@ public <V> S projectionConverter(Class<V> valueType, FromDocumentValueConverter<
 		return thisAsS();
 	}
 
+	@Override
+	public S queryStringConverter(ToDocumentValueConverter<String, ? extends F> fromQueryStringConverter) {
+		Contracts.assertNotNull( fromQueryStringConverter, "fromQueryStringConverter" );
+		builder.queryStringConverter( fromQueryStringConverter );
+		return thisAsS();
+	}
+
```
marko-bekhta, line 50: This converter doesn't make much sense in the case of an Elasticsearch backend... but I wasn't sure what to do about it if we were going this route of using a converter...

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/predicate/impl/LuceneQueryStringPredicate.java
```diff
@@ -145,6 +148,74 @@ private static QueryParser.Operator toOperator(BooleanOperator operator) {
...
+			var state = fieldStates.get( field );
+
+			if ( !state.field().type().valueClass().isAssignableFrom( String.class ) ) {
+				var builder = state.field().queryElement( PredicateTypeKeys.MATCH, scope );
+				builder.value( queryText, ValueConvert.PARSE );
+
+				return LuceneSearchPredicate.from( scope, builder.build() ).toQuery( contextForField( state ) );
+			}
+
+			return super.newFieldQuery( analyzer, field, queryText, quoted );
```
marko-bekhta, line 176: I've kept delegating to original parser logic for strings since, from what it seems, there's more to it than just a simple ` return new MatchSomethingQuery(...)`.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/AbstractElasticsearchIndexFieldTypeOptionsStep.java
yrodiere: Good point... And making this Lucene-specific isn't an option either, since it would make integration with value bridges much harder...

However, I see you added `ValueConvert.PARSE` though, which means this converter is not only about query strings: it can be used for many predicates, including in particular the `match` predicate; and in that case, it can be used even with Elasticsearch./

So:

1. Shouldn't this method be named something like `parser(ToDocumentValueConverter<String, ? extends F>)` instead?
2. Shouldn't there be tests for `ValueConvert.PARSE` in the TCK for other predicates? See in particular `AbstractPredicateTypeCheckingAndConversionIT`.

engine/src/main/java/org/hibernate/search/engine/search/common/spi/SearchIndexValueFieldTypeContext.java
```diff
@@ -49,6 +59,11 @@ default DslConverter<?, F> dslConverter(ValueConvert convert) {
 		switch ( convert ) {
 			case NO:
 				return rawProjectionConverter();
+			case PARSE:
+				throw log.parseConverterNotAllowed( ValueConvert.class.getSimpleName(),
+						ValueConvert.PARSE,
+						List.of( ValueConvert.YES, ValueConvert.NO )
+				);
```
marko-bekhta, line 66: There are a few places like this where we need a `FromDocumentValueConverter` instead of a `ToDocumentValueConverter`, which means that the parser doesn't make much sense... 
It means that we cannot pass `ValueConvert.PARSE` to projections, id-match-query, terms-aggregations.

and it works for terms/range/match queries, and sorts

yrodiere: Ok but in this specific case it does make sense, no? It would mean parsing the string into an identifier value, then shoving that value through the DSL converter.

yrodiere: Hmm sorry I looked at the wrong place.

You're right about this one, but I was talking about that one: https://github.com/hibernate/hibernate-search/pull/4035/files#diff-83eac4231be79581524a307f41fc0d0ac7ff0316ec7662c5d26c20e0222db7c9R37-R41

engine/src/main/java/org/hibernate/search/engine/search/common/ValueConvert.java
```diff
@@ -57,6 +57,14 @@ public enum ValueConvert {
...
 	 * Please refer to the reference documentation for more information.
 	 */
-	NO
+	NO,
+
+	/**
+	 * Used during parsing a query string to convert string token to a value that can be passed to a Lucene query.
+	 * <p>
+	 * It is only applicable to the Lucene backend, as with the Elasticsearch backend the query parsing is not happening
+	 * on the Hibernate Search side.
```
yrodiere, line 66: Needs updating.

I'd suggest using the javadoc of `YES` as a template.

engine/src/main/java/org/hibernate/search/engine/search/common/spi/SearchIndexValueFieldTypeContext.java
marko-bekhta: right, with the ID, I wasn't sure, meaning this parser takes a string and transforms it to whatever the thing should be, which in case of an ID is a string? no? so it basically would be string->string, and if we are already passing a string, then why does Search need to "convert" it?

yrodiere: Same type doesn't mean same data. The backend-side ID "String" might have a different format than whatever we expect users to provide.

E.g. for an `Instant` identifier one would want to use standard ISO-8601 for parsing, but on the backend side may chose some custom format where special characters are replaced with `-`.

That being said, while in theory this suggestion makes sense, I'm not sure it's easy to implement in practice, as I'm not sure the output of parsing has the same type as the input of DSL converters. Might be worth checking though :shrug:

yrodiere: > E.g. for an `Instant` identifier one would want to use standard ISO-8601 for parsing, but on the backend side may chose some custom format where special characters are replaced with `-`.

Another example: parsing may perform some normalization, such as trimming the string before parsing it. In which case `ValueConvert.NO` is not the same as `ValueConvert.PARSE`.

marko-bekhta: ok 😃 thanks! if it's a possible valid use case, then I'll try to make it work 😃👍🏻

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/IdentifierBridge.java
```diff
@@ -50,6 +51,18 @@ default boolean isCompatibleWith(IdentifierBridge<?> other) {
...
+	 * Parse an input String to the raw index field value.
+	 *
+	 * @param value The value to parse.
+	 * @return The raw index field value.
+	 * @throws RuntimeException If the value cannot be parsed to the raw index field value.
+	 */
+	@Incubating
+	default String parseToString(String value) {
+		return value;
+	}
```
marko-bekhta, line 64: I couldn't use the  `parse(string)` method name here as it clashes with the parse method from the bridge interface, which has a different return type. Otherwise this last commit is about allowing `PARSE` on match-id predicates.

marko-bekhta: >  You mean there's a clash for classes implementing both IdentifierBridge and ValueBridge?

yes, that 😃 

```java
public interface ValueBridge<V, F> {
    // @param <F> The type of raw index field values, on the index side of the bridge.
    default F parse(String value)
}
```
so parse returns the document value... (which is string for the id)

> Regardless... It's weird that this method returns a String rather than a I, no? I'd have expected Hibernate Search to apply toDocumentIdentifier automatically on the resulting "parsed" value (of type I), if necessary. Then your parseIdentifierLiteral method (or whatever you name it) could possibly be used for something else in the future.

mmm so it'll be something like:

```java
	@Override
	public String toDocumentValue(String value, ToDocumentValueConvertContext context) {
		return bridge.toDocumentIdentifier( bridge.parseIdentifierLiteral( value ), context );
	}
```

yrodiere: > mmm so it'll be something like:
> 
> ```java
> 	@Override
> 	public String toDocumentValue(String value, ToDocumentValueConvertContext context) {
> 		return bridge.toDocumentIdentifier( bridge.parseIdentifierLiteral( value ), context );
> 	}
> ```

Either that or you'd handle it on the backend side, with something like `dslConverter.toDocumentValue(parserConverter.toDocumentValue(string, context), context)`.

I don't know what feels more consistent with what you do for fields :thinking:

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/QueryStringPredicateSpecificsIT.java
```diff
@@ -226,6 +233,204 @@ void enablePositionIncrements_default() {
...
+						.fetchAllHits()
+		).hasDocRefHitsAnyOrder( indexForSyntaxParsingCheck.typeName(), "2" );
+
+		// ranges:
+		assertThatHits(
+				scope.query()
+						.where( f -> f.queryString().field( field )
+								.matching( String.format( Locale.ROOT, "[%s TO %s]", value1, value2 ) ) )
+						.fetchAllHits()
+		).hasDocRefHitsAnyOrder( indexForSyntaxParsingCheck.typeName(), "1", "2" );
```
yrodiere, line 292: Awesome!

BTW, I know that the fact non-string fields were not supported is not really explicit in the documentation, but now they _are_ supported you might want to explicitly mention it in the documentation of `simpleQueryString`/`queryString`? E.g. a subsection section titled "Expected format of field values" would totally make sense.

Good luck for the Elasticsearch part though ;D Maybe you should link to the table in `mapping-directfieldmapping-supported-types` for Lucene, but to the table in https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#backend-elasticsearch-field-types-available for Elasticsearch/OpenSearch.

It would probably also make sense to mention somewhere that some syntax constructs in the `queryString` predicate are only available on text fields -- e.g. regexp, phrase, etc.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/QueryStringPredicateSpecificsIT.java
```diff
@@ -226,6 +233,204 @@ void enablePositionIncrements_default() {
...
+								Instant.parse( "2000-01-01T01:01:01Z" ) ).replace( ":", "\\:" ),
+						backendFeatures.formatForQueryStringPredicate( InstantFieldTypeDescriptor.INSTANCE,
+								Instant.parse( "2000-02-02T02:02:02Z" ) ).replace( ":", "\\:" ),
+						backendFeatures.formatForQueryStringPredicate( InstantFieldTypeDescriptor.INSTANCE,
+								Instant.parse( "2222-02-02T02:02:02Z" ) ).replace( ":", "\\:" ),
+						"not-an-instant" ),
+				Arguments.of( "localDate", "2000-01-01", "2000-02-02", "2222-02-02", "not-an-localDate" )
+		);
+	}
+
```
yrodiere, line 433: Could you try to integrate this with `FieldTypeDescriptor`, so that we're sure we're testing all available types, including those that will get added in the future?

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/types/YearMonthFieldTypeDescriptor.java
```diff
@@ -69,7 +69,9 @@ protected List<YearMonth> createSingle() {
 	@Override
 	protected List<YearMonth> createUniquelyMatchableValues() {
 		List<YearMonth> values = new ArrayList<>();
-		for ( Year year : YearFieldTypeDescriptor.INSTANCE.getIndexableValues().getSingle() ) {
+		// NOTE: we are not going with the other years like (createUniquelyMatchableValues) because
+		//  years like 18353 cannot be parsed back from a string for some reason ...
```
yrodiere, line 73: Years beyond `9999` are supposed to be prefixed with a `+`, e.g. `+10000`; that's the `ISO_YEAR` format used in `ParseUtils`.

I guess `Year.toString()` doesn't output the year in ISO format. If that's the case I'd suggest exposing a `format` method in `FieldTypeDescriptor` and using that instead of `toString()` in relevant tests.

We can also certainly discuss changing the format used for parsing years... But I feel that's a broader change that deserves its own Jira.

yrodiere (approved):
Thanks. A few comments below; feel free to merge as soon as you're fine with the result.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/QueryStringPredicateSpecificsIT.java
```diff
@@ -281,94 +281,130 @@ void queryStringSyntax(String field, String value1, String value2, String noMatc
...
-
-		assertThatHits(
-				scope.query()
-						.where( f -> f.queryString().field( field ).matching( noMatch ) )
-						.fetchAllHits()
-		).isEmpty();
+		//   We are not running these range tests if the value contains square brackets [] because there is a problem in a Lucene query parser
+		//   and even if the query is escaped as `[2018\-01\-01T12\:58\:30.000000000\+02\:00\[Africa\/Cairo\] TO 2018\-02\-01T08\:15\:30.000000000\-02\:00\[\-02\:00\]]`
+		//   Fails for both Lucene and Elasticsearch backends...
+		if ( !( value1.contains( "]" ) || value2.contains( "]" ) ) ) {
```
marko-bekhta, line 290: note this "condition" 😃 
while the range queries do work with ZDT fields, as soon as there's that optional text part in brackets parser fails to parse the query. it fails for Elasticsearch as well. 
I've tried looking into the parser internals, but 😨 🙈

yrodiere: Makes sense. About the escape, do you mean that `\` is indeed intended for escapes but doesn't work, or is it just that `\` isn't supported? The escape character might be different...

yrodiere: BTW I hope our parser for `ZonedDateTime` is lenient then... e.g. does it accept something with just an offset, but no timezone?
For that matter, does it accept a date without a time (but with an offset)? Those are the formats that end users are more likely to use...

marko-bekhta: https://lucene.apache.org/core/2_9_4/queryparsersyntax.html#Escaping%20Special%20Characters

don't look at the version 🙈 
but yeah it is suppose to escape special characters, and it does in all other cases, otherwise `-10` is treated as `not 10`. but when it comes to range query string it seems theres' a problem in the lucene's parser, and it doesn't correctly escape the brackets

yrodiere: Okay then, probably something to report to Lucene. But not a blocker for us.

marko-bekhta: > BTW I hope our parser for ZonedDateTime is lenient then... e.g. does it accept something with just an offset, but no timezone?

We use  this: 
```java
ISO_ZONED_DATE_TIME = new DateTimeFormatterBuilder()
                .append(ISO_OFFSET_DATE_TIME)
                .optionalStart()
                .appendLiteral('[')
                .parseCaseSensitive()
                .appendZoneRegionId()
                .appendLiteral(']')
                .toFormatter(ResolverStyle.STRICT, IsoChronology.INSTANCE);
```
and I've tried the strings with just offset and it's fine, I even tried to do `toOffsetDateTime().toZoneDateTime()` in these tests to get rid of the zone, but then there were problems with Elasticsearch and I didn't want to deal with those 🙈

yrodiere: > but then there were problems with Elasticsearch and I didn't want to deal with those 🙈

Ok. You might want to keep them in mind though, because that possibly means we should add more "input" formats for our ZonedDateTime field types. Or just solve https://hibernate.atlassian.net/browse/HSEARCH-2354.

PR #4039, opened 2024-03-14
Ignore Jackson updates in dependabot config

marko-bekhta:
- since we align the version with the one used in Wildfly


yrodiere:
> since we align the version with the one used in Wildfly

FWIW you might want to give that decision some more thought, since Quarkus likely won't follow the same upgrade policy as WildFly, and we need to be compatible with both.

marko-bekhta:
😔 hmmm, maybe add an upgrade job that bumps Jackson to the version that Quarkus uses? while the main branch is using the version from Wildfly?

PR #4044, opened 2024-03-15
HSEARCH-5094 Follow-up: Enable testcontainers reuse on CI through workflow/pipeline config rather than the POM

yrodiere:

Because if we do this through the POM, the environment variable is taken
into account for build caching, which results in cache entries
contributed by CI being ignored when building locally.


marko-bekhta (approved):
😃

PR #4047, opened 2024-03-15
HSEARCH-5094 follow-up: HSEARCH-5094 Make sure the content of internal/incubating reports is reproducible  

yrodiere:

So that Maven goal executions relying on it (e.g. checkstyle) can be cached.


marko-bekhta (approved):
Makes sense 😃

PR #4051, opened 2024-03-19
Revert maven-compiler-plugin to 3.12.1

yrodiere:
Because:

> a new field forceLegacyJavacApi will make the compiler goal not cacheable until bumping up the Gradle Enterprise extension to 1.21 (should be available before end of month).

I hope dependabot will open another PR to upgrade again; if so let's leave that PR open until we do the Gradle Enterprise upgrade.


marko-bekhta (approved):
A lot of reverts on all these dependency updates lately 🙈 🙈 🙈

PR #4054, opened 2024-03-20
HSEARCH-5110 Error message when the version of Elasticsearch/Opensearch is too old for vector search states wrong requirements

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-5110


marko-bekhta (approved):
thanks! 🙈

PR #4057, opened 2024-03-20
HSEARCH-4806 Basic support for parameters at the query level

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4806

So far, it should cover query parameters for predicates, but the approach would be more or less the same for projections.
Docs and a lot of tests are still on their way 😃 
It should be ok to look at this by-commit.


engine/src/main/java/org/hibernate/search/engine/search/query/spi/QueryParameters.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import org.hibernate.search.engine.logging.impl.Log;
+import org.hibernate.search.util.common.annotation.Incubating;
+import org.hibernate.search.util.common.impl.Contracts;
+import org.hibernate.search.util.common.logging.impl.LoggerFactory;
+
+@Incubating
+public class QueryParameters {
```
marko-bekhta, line 19: I went with a "dedicated" wrapper around the map for these query parameters so that we can have various parameter checks in one place. See also a couple of TODOs in this class. I think that it doesn't make sense to have a parameter with a blank name or a null value. At least I'm not sure that there's a useful use case when the value is `null` ...

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/predicate/impl/LuceneKnnPredicate.java
```diff
@@ -26,17 +26,17 @@
 import org.apache.lucene.search.KnnFloatVectorQuery;
 import org.apache.lucene.search.Query;
 
-public class LuceneKnnPredicate extends AbstractLuceneSingleFieldPredicate implements LuceneSearchPredicate {
+public abstract class LuceneKnnPredicate<T> extends AbstractLuceneSingleFieldPredicate implements LuceneSearchPredicate {
```
marko-bekhta, line 33: Just noticed that it would be easier to work with query parameters here once there are "dedicated' predicates for bytes and floats

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/search/predicate/impl/LuceneKnnPredicate.java
```diff
@@ -81,10 +85,24 @@ public void k(int k) {
...
 		public void vector(Object vector) {
+			this.vectorProvider = simple(
+					objectToVector( vector, vectorCodec, absoluteFieldPath, vectorElementsType, indexedVectorsDimension ) );
+		}
+
+		@Override
+		public void vectorParam(String parameterName) {
+			this.vectorProvider =
+					parameter( parameterName, Object.class, vector -> objectToVector( vector, vectorCodec, absoluteFieldPath,
+							vectorElementsType, indexedVectorsDimension ) );
```
marko-bekhta, line 97: it seems safe to have such lambdas keeping scope/codec etc, since those are around for a long time, but let me know if that's not the case 😃

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchMatchIdPredicate.java
```diff
@@ -94,7 +98,15 @@ static class Builder extends AbstractElasticsearchPredicate.AbstractBuilder impl
...
-			values.add( converter.unknownTypeToDocumentValue( value, context ) );
+			valueProviders.add( simple( converter.unknownTypeToDocumentValue( value, context ) ) );
+		}
+
+		@Override
+		public void param(String parameterName, ValueConvert valueConvert) {
+			DslConverter<?, String> converter = scope.identifier().dslConverter( valueConvert );
+			ToDocumentValueConvertContext context = scope.toDocumentValueConvertContext();
+			valueProviders.add(
+					parameter( parameterName, String.class, value -> converter.unknownTypeToDocumentValue( value, context ) ) );
```
marko-bekhta, line 109: similar to Lucene's case, seems safe to use the converter in the lambda 🙈

engine/src/main/java/org/hibernate/search/engine/search/query/spi/QueryParametersValueProvider.java
```diff
@@ -0,0 +1,108 @@
...
+		public ParameterCollectionOrSingleValueProvider(String parameterName, Function<Collection<?>, ? extends F> mapping) {
+			this.parameterName = parameterName;
+			this.mapping = mapping;
+		}
+
+		@Override
+		public F provide(QueryParametersContext context) {
+			Object parameter = context.parameter( parameterName );
+			return mapping.apply( parameter instanceof Collection ? (Collection<?>) parameter : List.of( parameter ) );
+		}
```
marko-bekhta, line 106: not 100% on this one, it is to address the case of various terms predicates,  those predicates expect 1-n terms
and with our current DSL user can use a vararg and pass a single term or an array/list of them. But with a parameter, there's no way to tell what an expectation is, so the idea is that we always want to get a collection, and if the user gives a single element, then we wrap it into a list.

marko-bekhta (commented):
added a few inline comments 🙈

marko-bekhta:
I think this one is ok to take a look at, there is no ref-docs update for now; I'll update them once there are no changes to what predicates/sorts/aggregations/projections are parameterized

marko-bekhta:
closing in favor of https://github.com/hibernate/hibernate-search/pull/4061

PR #4059, opened 2024-03-21
Test against Hibernate ORM 6.5+ snapshots in dependency update jobs

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5112

### DIFF
--- a/v5migrationhelper/orm/src/main/java/org/hibernate/search/query/hibernate/impl/FullTextQueryImpl.java
+++ b/v5migrationhelper/orm/src/main/java/org/hibernate/search/query/hibernate/impl/FullTextQueryImpl.java
@@ -197,6 +197,10 @@ public class FullTextQueryImpl extends AbstractQuery implements FullTextQuery {
 		return hSearchQuery.getResultSize();
 	}
 
+	public long getResultCount() {
+		return doGetResultSize();
+	}
+
 	@Override
 	public FullTextQuery applyGraph(RootGraph graph, GraphSemantic semantic) {
 		applyGraph( (RootGraphImplementor) graph, semantic );
--- a/ci/dependency-update/Jenkinsfile
+++ b/ci/dependency-update/Jenkinsfile
@@ -16,6 +16,11 @@ Map settings() {
 					updateProperties: ['version.org.hibernate.orm'],
 					onlyRunTestDependingOn: ['hibernate-search-mapper-orm'],
 			]
+		case 'orm6.5':
+			return [
+					updateProperties: ['version.org.hibernate.orm'],
+					onlyRunTestDependingOn: ['hibernate-search-mapper-orm'],
+			]
 		case 'lucene9.9':
 			return [
 					updateProperties: ['version.org.apache.lucene'],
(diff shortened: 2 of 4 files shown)


v5migrationhelper/orm/src/main/java/org/hibernate/search/query/hibernate/impl/FullTextQueryImpl.java
```diff
@@ -197,6 +197,10 @@ public int doGetResultSize() {
 		return hSearchQuery.getResultSize();
 	}
 
+	public long getResultCount() {
+		return doGetResultSize();
+	}
```
marko-bekhta, line 202: we weren't testing the migration helper in the upgrade jobs, so this one slipped out, but it's a new method in ORM 6.5+ sooo would need that 🙈

ci/dependency-update/rules-orm6.5plus.xml
```diff
@@ -0,0 +1,12 @@
+<ruleset comparisonMethod="maven"
+         xmlns="http://mojo.codehaus.org/versions-maven-plugin/rule/2.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://mojo.codehaus.org/versions-maven-plugin/rule/2.0.0 http://mojo.codehaus.org/versions-maven-plugin/xsd/rule-2.0.0.xsd">
+  <ignoreVersions>
+    <!-- Ignore Alpha/Beta/CRs versions in the repo, because the versions-maven-plugin considers those
+         as higher priority than snapshots, which is wrong. -->
+    <ignoreVersion type="regex">.*\.(Alpha|Beta|CR).*</ignoreVersion>
+    <!-- Restrict allowed versions to a particular major/minor, because forbidding major/minor upgrades
+         in the version-maven-plugin configuration doesn't always work for some reason. -->
+    <ignoreVersion type="regex">(?!6\.[5-9]\.).*</ignoreVersion>
```
yrodiere, line 10: I don't think we need 6.5+, just 6.5.0+. People wanting to use 6.6 (not even planned yet) will most likely be required to use Search 7.2.

v5migrationhelper/orm/src/main/java/org/hibernate/search/query/hibernate/impl/FullTextQueryImpl.java
yrodiere: Ok, then we'll need a Jira (I think there is one, this is just a backport?) and a 7.1.1 :/

marko-bekhta: https://hibernate.atlassian.net/browse/HSEARCH-5092

We did this for the `HibernateOrmSearchQueryAdapter` but since it's already released in 7.1.0 ... probably a new one is needed now 😔

yrodiere: No that's fine, under `Actions` you have ` Backport` action to assign an issue to another Fix version.

yrodiere: image

yrodiere: Ah... 7.1.0. Right. We need another Jira.

marko-bekhta: 🙈

PR #4061, opened 2024-03-22
HSEARCH-4806 Alternative basic support for parameters at the query level

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4806

This essentially is doing the same as the named predicate, but pushes the `predicateDefinition.create( context )`  to the predicate's "to-query" stage


engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/SearchPredicateFactory.java
```diff
@@ -301,6 +301,9 @@ SimpleBooleanPredicateOptionsStep<?> or(PredicateFinalStep firstSearchPredicate,
 	 */
 	KnnPredicateFieldStep knn(int k);
 
+	@Incubating
+	ParameterizedPredicateOptionsStep parameterized();
```
marko-bekhta, line 305: that'll look as:
```java
var query = index.query().where(
    f -> f.bool()
        .must(
            f.parameterized().predicate(
                ctx -> ctx.predicate()
                    .match()
                    .field( "some-field" )
                    .matching( ctx.param( "my-query-param", String.class ) )
                    .toPredicate() )
        )
        .must(
            f.parameterized().predicate(
                ctx -> ctx.predicate()
                    .simpleQueryString()
                    .field( "some-field" )
                    .matching( ctx.param( "my-simple-query-match-param", String.class ) )
                    .defaultOperator(
                        ctx.param( "my-simple-query-bool-param", BooleanOperator.class ) )
                    .toPredicate() )
        )
)
    .param( "my-query-param", "match me" )
    .param( "my-simple-query-match-param", "I'm a simple query string" )
    .param( "my-simple-query-bool-param", BooleanOperator.OR )
    .toQuery();
```

WDYT @yrodiere

yrodiere: That's about what I had in mind... what about:

1. passing the lambda directly to `parameterized`
2. expecting the lambda to return a `PredicateFinalStep` instead of a `Predicate`, to remove the odd `toPredicate()` and be consistent with e.g. `org.hibernate.search.engine.search.predicate.dsl.GenericBooleanPredicateClausesStep#should(java.util.function.Function<? super org.hibernate.search.engine.search.predicate.dsl.SearchPredicateFactory,? extends org.hibernate.search.engine.search.predicate.dsl.PredicateFinalStep>)`?

marko-bekhta: ok
ok

makes sense will do 😃 thanks

documentation/src/test/java/org/hibernate/search/documentation/search/aggregation/AggregationDslIT.java
```diff
@@ -410,6 +410,35 @@ void filter() {
...
 	}
 
+	@Test
+	void withParameters() {
+		withinSearchSession( searchSession -> {
+			// tag::with-parameters[]
+			AggregationKey<Map<Range<Double>, Long>> countsByPriceKey = AggregationKey.of( "countsByPrice" );
+			SearchResult<Book> result = searchSession.search( Book.class )
+					.where( f -> f.matchAll() )
+					.aggregation( countsByPriceKey, f -> f.withParameters( ctx -> ctx.aggregation().range() // <1>
```
yrodiere, line 420: Not sure exposing the factory here is really necessary... you could just do this:

```suggestion
					.aggregation( countsByPriceKey, f -> f.withParameters( params -> f.range() // <1>
```

And instead of a context, you'd directly pass a `NamedValues` class similar to what's suggested in https://hibernate.atlassian.net/browse/HSEARCH-5075, so that the param retrieval reads `params.get("bound0", Double.class)`.

At least I _think_ this would feel more natural... WDYT?

documentation/src/main/asciidoc/public/reference/_search-dsl-aggregation.adoc
```diff
@@ -322,6 +322,27 @@ See <<search-dsl-argument-type>> for more information.
...
 
+[[search-dsl-aggregation-withparameters]]
+== `withParameters`: create aggregations using query parameters
+
+include::../components/_incubating-warning.adoc[]
+
+The `withParameters` aggregation allows building aggregations using <<search-dsl-query-parameters,query parameters>>.
+
+This type of aggregation gives user a context exposing an aggregation factory and access to query parameters and delays creating
+an actual aggregation to the query building time allowing all query parameters to be set.
```
yrodiere, line 333: I would avoid the word "user" since it's unclear whether we talk about developers or end users.

```suggestion
This type of aggregation requires a function that accepts query parameters and returns an aggregation. That function will get called at query building time.
```

Same for other documentation.

documentation/src/main/asciidoc/public/reference/_search-dsl-query.adoc
```diff
@@ -622,6 +622,21 @@ The former is used for debugging purposes and in particular to find out why a qu
...
+Some query elements may leverage the use of query parameters. Call `.param(..)` at the query level to set these:
+
+.Setting query parameters
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/query/QueryDslIT.java[tags=parameters]
+----
+====
+
```
yrodiere, line 639:
```
See also:

* <links to relevant predicates/aggregations/...>
* ...
````

File: documentation/src/main/asciidoc/public/reference/_search-dsl-sort.adoc [hunk 5]
```diff
@@ -181,6 +181,27 @@ can be <<search-dsl-sort-common-multi-value-mode,controlled explicitly with `.mo
 * For fields in nested objects, all nested objects are considered by default,
 but that can be <<search-dsl-sort-common-filter,controlled explicitly with `.filter(...)`>>.
 
+[[search-dsl-sort-withparameters]]
+== `withParameters`: create sorting using query parameters
```
yrodiere, line 185:
```suggestion
== `withParameters`: create sorts using query parameters
```

documentation/src/test/java/org/hibernate/search/documentation/search/query/QueryDslIT.java
```diff
@@ -528,6 +529,26 @@ void graph() {
...
+		with( entityManagerFactory ).runInTransaction( entityManager -> {
+			SearchSession searchSession = Search.session( entityManager );
+			// tag::parameters[]
+			List<Manager> managers = searchSession.search( Manager.class )
+					.where(
+							//...
+							// end::parameters[]
+							f -> f.matchAll()
+			// tag::parameters[]
+					).param( "param1", "name" )
```
yrodiere, line 543:
```suggestion
					)
					.param( "param1", "name" )
```

engine/src/main/java/org/hibernate/search/engine/logging/impl/Log.java
```diff
@@ -581,4 +581,15 @@ SearchException cannotIncludeAndExcludePathsWithinSameFilter(Set<String> include
...
 	SearchException nonParameterizedRewriteMethodWithParameter(RewriteMethod rewriteMethod);
+
+	@Message(id = ID_OFFSET + 123,
+			value = "Query parameter '%1$s' value is not set."
+					+ " Use `.param(..)` methods on the query to set any parameters that the query requires.")
+	SearchException cannotFindQueryParameter(String parameter);
+
+	@Message(id = ID_OFFSET + 124,
+			value = "Expecting the query parameter '%1$s' value to be of the %2$s type,"
+					+ " but instead got a value of the %3$s type.")
```
yrodiere, line 592:
```suggestion
			value = "Expecting value of query parameter '%1$s' to be of type %2$s,"
					+ " but instead got a value of type %3$s.")
```

engine/src/main/java/org/hibernate/search/engine/logging/impl/Log.java
```diff
@@ -581,4 +581,15 @@ SearchException cannotIncludeAndExcludePathsWithinSameFilter(Set<String> include
 			value = "Cannot use rewrite method '%1$s': this method does not accept parameter 'n', but it was specified."
 					+ " Use another version of the rewrite(...) method that does not accept parameter 'n'.")
 	SearchException nonParameterizedRewriteMethodWithParameter(RewriteMethod rewriteMethod);
+
+	@Message(id = ID_OFFSET + 123,
+			value = "Query parameter '%1$s' value is not set."
```
yrodiere, line 586:
```suggestion
			value = "Query parameter '%1$s' is not set."
```

yrodiere (approved):
LGTM overall, thanks. I added a few suggestions regarding APIs, but we can handle them as a second step, especially the part about https://hibernate.atlassian.net/browse/HSEARCH-5075

documentation/src/test/java/org/hibernate/search/documentation/search/aggregation/AggregationDslIT.java
marko-bekhta: hmm, IDK 😃 🙈, there's nothing stopping the user from doing it right now (meaning ignoring the factory from the context and going with the one from the "outer lambda" 😃) I wasn't sure though if it is ok to do such "nesting" of lambdas. But if this won't be a problem (using `f` from outer lambda), I don't mind making the adjustments 😃 

this got me thinking .... what if we add to `SearchQueryWhereStep` something like:
```java
N where(BiFunction<? super PDF, ? extends NamedValues, ? extends PredicateFinalStep> predicateContributor);
```
and for all other entry points -- projection/sort/aggregation...

then the user could've done something like:
```java
.where( (f, param) -> f.match().field( "title" ).matching( param.get( "myParam" ) ) )
```
the problem is that this kinda clashes with the 
```
N where(BiConsumer<? super PDF, ? super SimpleBooleanPredicateClausesCollector<?>> predicateContributor);
```
(when defining lambdas) 😔

yrodiere: > there's nothing stopping the user from doing it right now (meaning ignoring the factory from the context and going with the one from the "outer lambda" 😃)

Sure. I just felt that this would work better with HSEARCH-5075: one would probably rather write `params.get(...)` than `ctx.params().get(...)`.

> I wasn't sure though if it is ok to do such "nesting" of lambdas. But if this won't be a problem (using f from outer lambda), I don't mind making the adjustments 😃

It technically means lambdas would be capturing, but currently any lambda used in the DSL that relies on user input (most likely the vast majority of them) is capturing. So... no big deal, I think.

> this got me thinking .... what if we add to SearchQueryWhereStep something like:

Tempting, but as you just saw it's also not possible ;D Also this basically means the entire predicate tree would be created later, with the downsides we know when it comes to stacktraces.

I think we just can't cover every use case at the root, and because of the kind of clashes you just noticed, we need to be conservative. So I'd wait a bit for the API to mature before thinking about this.

marko-bekhta:
```
2024-03-27T16:28:42.1170340Z [ERROR] --[ Constraint Violation ]-----------------------------------------
2024-03-27T16:28:42.1171995Z [ERROR] Constraint: hsearch:PublicTypesMayNotDeclareMethodsWithGetOrSetPrefix
2024-03-27T16:28:42.1173179Z [ERROR] Severity: MAJOR
2024-03-27T16:28:42.1222689Z [ERROR] Number of rows: 1
2024-03-27T16:28:42.1236027Z [ERROR] type=org.hibernate.search.engine.search.common.NamedValues, method=java.util.Optional getOptional(java.lang.String,java.lang.Class)
```
😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 

😆 
so maybe:

```java
public interface NamedValues {
	<T> T get(String name, Class<T> paramType);
	<T> Optional<T> optional(String name, Class<T> paramType);
}
```
m?

yrodiere: Rules are made to be broken :)

Try `@SuppressJQAssistant`, and if that doesn't work adjust that specific constraint to take the annotation into account (there are examples in other constraints).

PR #4062, opened 2024-03-22
HSEARCH-4971 Switch project license to Apache License 2.0

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-4971

See https://in.relation.to/2023/11/18/license/

~~Creating as draft while we're reviewing some final details.~~ => All good!

Best reviewed commit by commit, due to the huge, annoying commit that changes file headers (which, obviously, doesn't need reviewing -- there are automated checks for that).

I've gone through past contributions and, unless I am mistaken, we've got permission for this change from all authors of non-trivial contributions that are still present in the repo (some contributions have been deleted in subsequent refactorings, mainly Hibernate Search 6.0).

I've also had to remove a few tests inherited from Search 5, which nowadays are mainly relevant to the v5migrationhelper modules, so... they would probably have been removed in 8.0 along with the v5 migration helpers anyway. Not to detract from those contributions, they were definitely useful (and will continue to be) in v5; it's just that a lot has changed since then, so in newer versions, we can live without them.


build/config/src/main/resources/checkstyle.xml
```diff
@@ -13,12 +13,12 @@
     <property name="cacheFile" value="${checkstyle.cache.file}" />
 
     <module name="Header">
-        <property name="header" value="/*\n * Hibernate Search, full-text search for your domain model\n *\n * License: GNU Lesser General Public License (LGPL), version 2.1 or later\n * See the lgpl.txt file in the root directory or &lt;http://www.gnu.org/licenses/lgpl-2.1.html&gt;.\n */"/>
+        <property name="headerFile" value="classpath:/org/hibernate/search/headers/java.header"/>
         <property name="fileExtensions" value="java"/>
```
marko-bekhta, line 16: Would that include Jenkins files as well?

also on a related note, do we need to "license" docker files too?

marko-bekhta (approved):
🎉

build/config/src/main/resources/checkstyle.xml
yrodiere: Nope, I forgot about checking those -- I'll add something.
I hope your Dockerfile parsing utilities take comments into account -- let's see ;D

yrodiere: Actually neither Jenkinsfiles nor Dockerfiles are part of a Maven project with a checkstyle execution... They're technically part of the root project (I guess), and I'm not sure what the effect of adding a checkstyle to the root project would be, so... I'll leave this out for now.

Sanne (commented):
It's approved. Congratulations!

PR #4064, opened 2024-03-25
Safely publish build scans even for pull requests 

yrodiere:
And do it synchronously (before the build ends) so that Hibernate GitHub Bot is able to list build scans immediately.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,9 @@ on:
       - '!wip/**/dependency-update/**'
     tags:
       - '**'
-  pull_request:
+  # WARNING: Using pull_request_target to access secrets, but we check out the merge commit.
+  # See checkout action for details.
+  pull_request_target:
     types: [opened, synchronize, reopened, ready_for_review]
     branches:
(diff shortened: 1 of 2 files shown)


yrodiere:
Seems to be working fine for PRs in Jenkins:

image

image

So... let's merge and see about the main branch in Jenkins + GitHub Actions in general (you can't test changes to GitHub Actions  workflows without merging the changes)

marko-bekhta:
🤞🏻 😃

yrodiere:
Ok so I had to push a few follow-up fix commits, but it seems to be working now.


On a PR, we get build scans:

https://github.com/hibernate/hibernate-search/pull/4062/checks?check_run_id=23054112481

image

And the build cache behaves as expected (reads from remote cache, does not write):

image

On push events, everything seems to be working fine too: https://github.com/hibernate/hibernate-search/runs/23053729751

yrodiere:
Well okay, there's *one* slightly annoying thing, it's that PRs don't get tagged with `PR-someNumber`, and instead get tagged with `HEAD`. But personally I can live with that.

PR #4078, opened 2024-03-28
Require approval for Jenkins builds on PRs from external contributors

yrodiere:
It looks like this in practice:

image

Anybody in the Hibernate organization is able to approve something like this.

Approval is not required for non-PR builds, or builds of PRs submitted by members of the organization, or builds of PRs explicitly triggered (through the UI) by members of the organization.


marko-bekhta (approved):
nice! 🎉

PR #4088, opened 2024-04-02
HSEARCH-5119 Use `IndexSearcher#setTimeout` instead of a TimeLimitingCollector

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5119


integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/query/SearchQueryTimeoutIT.java
```diff
@@ -40,7 +40,8 @@ class SearchQueryTimeoutIT {
 	private static final int NON_MATCHING_INTEGER = 739;
 
 	// Increasing this will increase query execution time.
-	private static final int DOCUMENT_COUNT = 100;
+	// Should be more than TimeLimitingBulkScorer#INTERVAL
```
marko-bekhta, line 43: https://github.com/apache/lucene/blob/baecaf556f8fe5db69d130f0a9094e83c2f5f226/lucene/core/src/java/org/apache/lucene/search/TimeLimitingBulkScorer.java#L35

PR #4095, opened 2024-04-04
Test full build of 7.1 on CI with 6.5 just in case

marko-bekhta:


yrodiere:
@marko-bekhta Note that being able to build against 6.5 is not the same as being able to run against 6.5.

The dependency-update jobs check that you can run against 6.5, which is what matters.

You may not be able to build against 6.5 due to e.g. deprecation warnings (which cause build failures) or problems related to generics (which are erased anyway and don't always matter at runtime).

yrodiere:
That being said, the compilation error we're seing in this build is indeed suspicious. But I think you sent a patch to add a default implementation for that method in ORM... didn't you?

PR #4109, opened 2024-04-09
HSEARCH-5128 Update the supported versions of Java in the documentation

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-5128


marko-bekhta (approved):
🎉

PR #4116, opened 2024-04-12
HSEARCH-3200 Allow setting the minimum number of should clauses required to match for full-text predicates in the query DSL

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-3200


documentation/src/main/asciidoc/public/reference/_search-dsl-predicate.adoc
```diff
@@ -268,6 +268,42 @@ include::{sourcedir}/org/hibernate/search/documentation/search/predicate/Predica
...
+Changing the operator can be done by calling `.operator(...)`.
+
+.`match` predicate with `AND` operator
+====
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/predicate/PredicateDslIT.java[tags=match-operator]
+----
+====
+
```
yrodiere, line 306: I'm not sure how this is related to this PR... but you definitely should have a look at https://hibernate.atlassian.net/browse/HSEARCH-917 . See the comments in particular.

You probably will want to have a look at https://hibernate.atlassian.net/browse/HSEARCH-3327 too, because I think Elasticsearch exposes a single settings controlling both multi-term handling and multi-field handling.

marko-bekhta: > I'm not sure how this is related to this PR...

I've convinced myself before taking a look at the query builder that it uses a `MUST` clause and thought .. how would I test min should match if the clauses are must clauses? As in this comment 😄:

> From my experience with Hibernate Search, a lot of people are mistaken about how matching() works so it might be a good idea to give them an hint in the API.

But when I actually looked at the query builder, I already had the operator ready, so I thought I'd keep it and ask you what you think about it 😃

yrodiere: I think it's a risky move if you don't have the whole problem in mind, but if you're confident, well... :)

PR #4117, opened 2024-04-15
HSEARCH-5075 Refactor parameter handling during binding

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5075

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/search/predicate/spi/NamedValuesBasedPredicateDefinitionContext.java
+++ b/engine/src/main/java/org/hibernate/search/engine/search/predicate/spi/NamedValuesBasedPredicateDefinitionContext.java
@@ -1,0 +1,41 @@
+/*
+ * Hibernate Search, full-text search for your domain model
+ *
+ * License: GNU Lesser General Public License (LGPL), version 2.1 or later
+ * See the lgpl.txt file in the root directory or <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ */
+package org.hibernate.search.engine.search.predicate.spi;
+
+import java.util.Map;
+import java.util.function.Function;
+
+import org.hibernate.search.engine.search.common.NamedValues;
+import org.hibernate.search.engine.search.common.spi.MapNamedValues;
+import org.hibernate.search.engine.search.predicate.definition.PredicateDefinitionContext;
+import org.hibernate.search.engine.search.predicate.dsl.SearchPredicateFactory;
+import org.hibernate.search.util.common.SearchException;
+import org.hibernate.search.util.common.annotation.Incubating;
+
+@Incubating
+public final class NamedValuesBasedPredicateDefinitionContext implements PredicateDefinitionContext {
+
+	private final SearchPredicateFactory factory;
+	private final NamedValues parameters;
+
+	public NamedValuesBasedPredicateDefinitionContext(SearchPredicateFactory factory, Map<String, Object> params,
+			Function<String, SearchException> namedValueMissing) {
+		this.factory = factory;
+		this.parameters = MapNamedValues.fromMap( params, namedValueMissing );
+	}
+
+	@Override
+	public SearchPredicateFactory predicate() {
+		return factory;
+	}
+
+	@Override
+	public NamedValues params() {
+		return parameters;
+	}
+
+}
--- a/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/binding/impl/AbstractBindingContext.java
+++ b/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/binding/impl/AbstractBindingContext.java
@@ -8,12 +8,12 @@ package org.hibernate.search.mapper.pojo.bridge.binding.impl;
 
 import java.lang.invoke.MethodHandles;
 import java.util.Map;
-import java.util.Optional;
 
 import org.hibernate.search.engine.environment.bean.BeanResolver;
+import org.hibernate.search.engine.search.common.NamedValues;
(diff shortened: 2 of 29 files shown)


mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/binding/impl/AbstractBindingContext.java
```diff
@@ -35,22 +36,18 @@ public BeanResolver beanResolver() {
...
-		if ( value == null ) {
-			throw log.paramNotDefined( name );
-		}
-
-		return paramType.cast( value );
+		return params.get( name, paramType );
 	}
 
 	@Override
 	public <T> Optional<T> paramOptional(String name, Class<T> paramType) {
```
yrodiere, line 43: Shouldn't older methods get deprecated?

marko-bekhta: well that backfired nicely 😅 
```
Error:  Constraint: hsearch:TypesExtendingTypeFromAnotherModuleMustHaveModuleSpecificKeywordInName
Error:  Severity: MAJOR
Error:  Number of rows: 22
Error:  
Error:  Main (non-test) types extending/implementing a Hibernate Search type from another module must have a module-specific keyword in their name,
Error:  either at the very start or just after "Abstract".
Error:  This allows to more easily understand which module a given type comes from.
Error:  Exceptions are allowed when:
Error:  - the misnamed type is an anonymous type or a generated type;
Error:  - or the misnamed type is deprecated;
Error:  - or the implemented type is deprecated;
Error:  - or the implementing type is an inner/nested type and is non-public (implementation) --
Error:  in which case only the surrounding type must comply, if relevant;
Error:  - or the implemented type is in a util module,
Error:  in which case the implemented interface may just be a detail.
Error:  
Error:  misnamedTypeArtifact.moduleSpecificKeyword=Pojo, misnamedType=org.hibernate.search.mapper.pojo.bridge.binding.impl.PropertyBindingContextImpl, externalParentType=org.hibernate.search.engine.search.common.NamedValues
```
Once I made `BindingContext extends NamedValues` 😄

yrodiere: Yeah, don't do that :)

build/jqassistant/rules/rules.xml
```diff
@@ -600,6 +601,7 @@
                 AND NOT misnamedType:Anonymous
                 AND NOT misnamedType:Deprecated
                 AND NOT externalParentType:Deprecated
+                AND NOT externalParentType:SuppressJQAssistant
```
marko-bekhta, line 604: > Yeah, don't do that :)

ohh so this will go away then 🙈

yrodiere: Well this can stay, but the `@SuppressJQAssistant` annotation you probably added should go away.

Not sure why `BindingContext` would extend `NamedValues`... the whole point what to get people to write `ctx.params().get("something")`, wasn't it? `ctx.get("something")` could get confusing.

marko-bekhta: For now, we only have the named values in the `withParameters(...)` and there:
https://github.com/hibernate/hibernate-search/pull/4061#discussion_r1540942440

soooo that's why I thought the idea was to do the same thing here (don't add additional params but just get methods). 🙈 

as for the suppressing annotation -- it was there already as we want the get/getoptional methods and JQ won't let us do that on "public API classes" 🙈.
I've updated the code to have the params method.

engine/src/main/java/org/hibernate/search/engine/search/predicate/spi/NamedValuesBasedPredicateDefinitionContext.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.hibernate.search.engine.search.common.NamedValues;
+import org.hibernate.search.engine.search.common.spi.MapNamedValues;
+import org.hibernate.search.engine.search.predicate.definition.PredicateDefinitionContext;
+import org.hibernate.search.engine.search.predicate.dsl.SearchPredicateFactory;
+import org.hibernate.search.util.common.SearchException;
+import org.hibernate.search.util.common.annotation.Incubating;
+
+@Incubating
+public class NamedValuesBasedPredicateDefinitionContext implements PredicateDefinitionContext {
```
yrodiere, line 20: If you make this final you can shield yourself from (some) abuse :)

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/bridge/binding/impl/AbstractBindingContext.java
```diff
@@ -34,23 +34,14 @@ public BeanResolver beanResolver() {
...
-	@Override
-	public <T> Optional<T> paramOptional(String name, Class<T> paramType) {
-		Contracts.assertNotNull( name, "name" );
-		Contracts.assertNotNull( paramType, "paramType" );
+	private static class BindingContextParameters extends MapNamedValues {
 
-		return Optional.ofNullable( params.get( name ) ).map( paramType::cast );
+		public BindingContextParameters(Map<String, Object> values) {
+			super( values, log::paramNotDefined );
+		}
```
yrodiere, line 45: This could have been a factory method!

(my version of "this could have been an email")

yrodiere (approved):
Thanks, LGTM. A few comments below, but honestly it's more for awareness, you could merge this as-is.

PR #4125, opened 2024-04-22
HSEARCH-5132 Do not directly rely on the Security Manager in EmbeddedThreadProvider

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5132


yrodiere:
@marko-bekhta Did you test this against the latest build of JDK 23? https://jdk.java.net/23/

If the tests pass on that version, I'd merge this PR. Even if they didn't remove the security manager yet; when they do, we'll know quite quickly whether this patch is enough or not.

marko-bekhta:
Started a JDK 23 build here https://ci.hibernate.org/job/hibernate-search/job/PR-4125/12/

I've looked at it on the previous build, and SM was still there so I just kept the PR open 🙈, let's see how the build goes 😃

PR #4130, opened 2024-04-23
HSEARCH-5135 Improve test execution time for some of the orm-based tests

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5135

### DIFF
--- a/util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/DatabaseContainer.java
+++ b/util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/DatabaseContainer.java
@@ -16,8 +16,12 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.function.BiConsumer;
 
-import org.hibernate.cfg.JdbcSettings;
+import javax.script.ScriptException;
 
+import org.hibernate.cfg.JdbcSettings;
+import org.hibernate.search.util.common.AssertionFailure;
+
+import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.github.dockerjava.api.model.HostConfig;
 import com.github.dockerjava.api.model.Ulimit;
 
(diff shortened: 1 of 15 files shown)


util/internal/integrationtest/mapper/orm/src/main/java/org/hibernate/search/util/impl/integrationtest/mapper/orm/DatabaseContainer.java
```diff
@@ -284,7 +289,23 @@ HibernateSearchJdbcDatabaseContainer container(Path dockerfile, String name) {
...
+					@Override
+					protected void containerIsStarted(InspectContainerResponse containerInfo) {
+						super.containerIsStarted( containerInfo );
+						try {
+							// see https://www.cockroachlabs.com/docs/stable/configure-replication-zones#replication-zone-variables
+							// Smaller values can save disk space and improve performance if values are frequently overwritten or for queue-like workloads
+							ScriptUtils.executeDatabaseScript( getDatabaseDelegate(), null,
+									"ALTER DATABASE defaultdb CONFIGURE ZONE USING gc.ttlseconds = 60;\n"
+											+ "ALTER RANGE default CONFIGURE ZONE USING gc.ttlseconds = 60;"
+							);
```
marko-bekhta, line 302: Maybe this won't help with our tests much, but that's something to consider for those using outbox-polling with CockroachDB I guess 😃 (just in case you are interested @yrodiere  😃)

yrodiere: Makes sense...

I wonder though; those other changes you're making seem like an awful lot of work. Isn't there rather an option like:

```
ALTER DATABASE defaultdb CONFIGURE ZONE USING
schema.dont-be-dumb-i-dont-want-a-distrubuted-database-just-some-fast-stuff.so-make-ddl-fast = true;
```

? :grin:

marko-bekhta: it's hidden somewhere and they don't want to share!!! 🤣

yrodiere:
Nice, it seems like you scraped ~20 minutes off the ~1 hour 40 minutes build:

* Before: https://ge.hibernate.org/s/itrmwxt4trqge/tests/overview
* After https://ge.hibernate.org/s/sbjlrwypkosgm/tests/overview

PR #4142, opened 2024-05-01
HSEARCH-5107 Add a test checking segment files on various index operations

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5107
https://hibernate.atlassian.net/browse/HSEARCH-5108

### DIFF
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/reader/impl/IndexReaderProvider.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/reader/impl/IndexReaderProvider.java
@@ -15,15 +15,10 @@ public interface IndexReaderProvider {
 	/**
 	 * Closes and drops any cached resources (index readers in particular).
 	 * <p>
-	 * Should be used when stopping the index or to clean up upon error.
+	 * Should be used when stopping the index, to clean up upon error,
+	 * or simply to force the creation of a new reader (refresh) on the next call to {@link #getOrCreate()}.
 	 */
 	void clear() throws IOException;
-
-	/**
-	 * Make sure the index reader returned by the next call to {@link #getOrCreate()}
-	 * will return an up-to-date view of the index.
-	 */
-	void refresh();
 
 	/**
 	 * @return A ready-to-use index reader, with its reference count already increased.
--- a/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/reader/impl/IndexReaderProvider.java
+++ b/backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/reader/impl/IndexReaderProvider.java
@@ -28,4 +28,7 @@ public interface IndexReaderProvider {
 	 */
 	DirectoryReader getOrCreate() throws IOException;
 
+	default DirectoryReader getCurrentForTests() throws IOException {
+		return getOrCreate();
+	}
 }
--- a/integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/lowlevel/reader/LuceneIndexSegmentFilesIT.java
+++ b/integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/lowlevel/reader/LuceneIndexSegmentFilesIT.java
@@ -14,9 +14,12 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
 import org.hibernate.search.backend.lucene.LuceneExtension;
+import org.hibernate.search.backend.lucene.index.impl.LuceneIndexManagerImpl;
+import org.hibernate.search.backend.lucene.index.impl.Shard;
 import org.hibernate.search.engine.backend.common.DocumentReference;
 import org.hibernate.search.engine.backend.document.IndexFieldReference;
 import org.hibernate.search.engine.backend.document.model.dsl.IndexSchemaElement;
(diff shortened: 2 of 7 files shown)


yrodiere:
Back to this... we discussed it in a video call, but I don't remember exactly what your problem was. It was testing, right? I assume the tests described in the Jira issue are impossible to implement.

IMO:

1. For https://hibernate.atlassian.net/browse/HSEARCH-5107 , set up a test where you run a search, check that Hibernate Search's index reader is still open, then call `merge()`, then check that Hibernate Search's index reader is no longer open. It's imperfect as it won't check that the reader closing happens before the merging, but still, it's something.
2. For https://hibernate.atlassian.net/browse/HSEARCH-5108 , set up a test where you run a search, check that Hibernate Search's index reader is still open, then call `refresh()`, then check that Hibernate Search's index reader is no longer open.

OR just forget about integration tests and add appropriate unit tests -- I think we have a few already for the low-level Lucene stuff:

1. For https://hibernate.atlassian.net/browse/HSEARCH-5107 , just check that the `merge()` operation triggers an attempt to close readers before it does its thing.
1. For https://hibernate.atlassian.net/browse/HSEARCH-5108 , just check that the `refresh()` operation triggers an attempt to close readers before it does its thing.

yrodiere:
> I was trying to get that, but even with what we have right now (without closing anything), everything seems to work just fine. see this test for example:

I'd certainly expect things to work fine right now on Linux; this is a Windows-specific problem. On Linux (at last in every case I've encountered) you can delete a file that's still being read -- on Windows (at least in some cases) this will fail.

If it works on Windows, well... I don't know, I'd have to debug to find out how the merge manages to delete an open file, and I really don't want to :)

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/lowlevel/reader/LuceneIndexSegmentFilesIT.java
```diff
@@ -0,0 +1,236 @@
...
+			LuceneIndexManagerImpl luceneIndexManager = index.unwrapForTests( LuceneIndexManagerImpl.class );
+			assertThat( luceneIndexManager.getShardsForTests() ).hasSize( 1 );
+			Shard shard = luceneIndexManager.getShardsForTests().get( 0 );
+
+			IndexReader indexReaderBefore = shard.indexAccessorForTests().getCurrentReaderForTests();
+			assertThat( indexReaderBefore ).isNotNull();
+			IndexWorkspace workspace = index.createWorkspace();
+			operation.accept( workspace );
+
+			assertThat( shard.indexAccessorForTests().getCurrentReaderForTests() ).isNull();
```
yrodiere, line 136: Maybe also check that `indexReaderBefore` is closed? You could check that `.getRefCount() == 0`.

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/lowlevel/reader/LuceneIndexSegmentFilesIT.java
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	void noReader(Consumer<IndexWorkspace> operation) throws Exception {
+		basicTestSteps( () -> {
+			LuceneIndexManagerImpl luceneIndexManager = index.unwrapForTests( LuceneIndexManagerImpl.class );
+			assertThat( luceneIndexManager.getShardsForTests() ).hasSize( 1 );
+			Shard shard = luceneIndexManager.getShardsForTests().get( 0 );
+
+			IndexReader indexReaderBefore = shard.indexAccessorForTests().getCurrentReaderForTests();
+			assertThat( indexReaderBefore ).isNotNull();
```
yrodiere, line 132: Maybe also check that `indexReaderBefore` is open? You could check that `.getRefCount() > 0`.

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/reader/impl/IndexReaderProvider.java
```diff
@@ -12,18 +12,11 @@
...
-	 * Should be used when stopping the index or to clean up upon error.
-	 */
-	void clear() throws IOException;
-
 	/**
 	 * Make sure the index reader returned by the next call to {@link #getOrCreate()}
 	 * will return an up-to-date view of the index.
 	 */
-	void refresh();
+	void refresh() throws IOException;
```
yrodiere, line 19: Considering what `IndexReaderProvider#refresh` does and how it's used, it should probably be called `clear`. Yeah I know, this was my suggestion, but it was wrong :x 

Also, for the javadoc:

```
	/**
	 * Closes and drops any cached resources (index readers in particular).
	 * <p>
	 * Should be used when stopping the index, to clean up upon error, or simply to force the creation of a new reader (refresh) on the next call to {@link #getOrCreate()}.
	 */
```

yrodiere (commented):
>     * this test here https://github.com/hibernate/hibernate-search/blob/a380bbac912bb581fba68511cda154b43bda9c55/integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/lowlevel/reader/LuceneIndexSegmentFilesIT.java#L126-L139
>        is close to your suggestion 1

Agreed; added a few comments below.

>     * As for suggestion 2, I think you are talking about the IndexAccessorTest; which mostly covers the scenario here
>       https://github.com/hibernate/hibernate-search/blob/a380bbac912bb581fba68511cda154b43bda9c55/backend/lucene/src/test/java/org/hibernate/search/backend/lucene/lowlevel/index/impl/IndexAccessorTest.java#L157C1-L165C3

Not quite, as it doesn't check that `IndexReaderProvider` will actually close readers immediately on calls to `refresh`. _But_, you check that in the integration test above, so you're right, this is enough.

integrationtest/backend/lucene/src/test/java/org/hibernate/search/integrationtest/backend/lucene/lowlevel/reader/LuceneIndexSegmentFilesIT.java
```diff
@@ -0,0 +1,238 @@
...
+			assertThat( indexReaderBefore.getRefCount() ).isPositive();
+			IndexWorkspace workspace = index.createWorkspace();
+			operation.accept( workspace );
+
+			assertThat( shard.indexAccessorForTests().getCurrentReaderForTests() ).isNull();
+			assertThat( index.query().select().where( f -> f.matchAll() ).fetchTotalHitCount() ).isPositive();
+
+			assertThat( shard.indexAccessorForTests().getCurrentReaderForTests() ).isNotNull().isNotSameAs( indexReaderBefore );
+			assertThat( indexReaderBefore.getRefCount() ).isZero();
```
yrodiere, line 141: Shouldn't this be checked before the query? Just after you perform the refresh/merge, really?

yrodiere (approved):
Thanks. Feel free to merge, but see my comment

PR #4144, opened 2024-05-02
HSEARCH-5133 Support basic metrics aggregations

fax4ever:
https://hibernate.atlassian.net/browse/HSEARCH-5133

~~1. we need to implement count, count distinct and avg~~ done!
~~2. test filtering on nested~~ done!
~~3. test explicit ValueConvert.YES/NO~~ done!
~~4. support and test aggregations for temporal types~~ done!
~~5. we need to implement the Lucene backend~~ done!
~~6. write good tests covering all the fields types~~ done from my side (check with the Hibernate team)
~~7. write some doc~~ done!

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/aggregation/impl/ElasticsearchMetricFieldAggregation.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/aggregation/impl/ElasticsearchMetricFieldAggregation.java
@@ -17,6 +17,7 @@ import org.hibernate.search.engine.backend.types.converter.runtime.FromDocumentV
 import org.hibernate.search.engine.backend.types.converter.spi.ProjectionConverter;
 import org.hibernate.search.engine.search.aggregation.spi.FieldMetricAggregationBuilder;
 import org.hibernate.search.engine.search.common.ValueModel;
+import org.hibernate.search.util.common.AssertionFailure;
 
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
@@ -126,9 +127,7 @@ public class ElasticsearchMetricFieldAggregation<F, K> extends AbstractElasticse
 		@Override
 		public <T> Builder<F, T> type(Class<T> expectedType, ValueModel valueModel) {
 			ProjectionConverter<F, ? extends T> projectionConverter = null;
-			if ( !Double.class.isAssignableFrom( expectedType )
-					||
-					field.type().projectionConverter( valueModel ).valueType().isAssignableFrom( expectedType ) ) {
+			if ( useProjectionConverter( expectedType, valueModel ) ) {
 				projectionConverter = field.type().projectionConverter( valueModel )
 						.withConvertedType( expectedType, field );
 			}
@@ -136,6 +135,22 @@ public class ElasticsearchMetricFieldAggregation<F, K> extends AbstractElasticse
 					projectionConverter,
 					operation
 			);
+		}
+
+		private <T> boolean useProjectionConverter(Class<T> expectedType, ValueModel valueModel) {
+			if ( !Double.class.isAssignableFrom( expectedType ) ) {
+				if ( ValueModel.RAW.equals( valueModel ) ) {
+					throw new AssertionFailure(
+							"Raw projection converter is not supported with metric aggregations at the moment" );
+				}
+				return true;
+			}
+
+			// expectedType == Double.class
+			if ( ValueModel.RAW.equals( valueModel ) ) {
+				return false;
+			}
+			return field.type().projectionConverter( valueModel ).valueType().isAssignableFrom( Double.class );
 		}
 	}
 
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/aggregation/impl/ElasticsearchMetricFieldAggregation.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/aggregation/impl/ElasticsearchMetricFieldAggregation.java
@@ -1,0 +1,186 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.search.backend.elasticsearch.search.aggregation.impl;
+
+import java.util.List;
+import java.util.Optional;
+
+import org.hibernate.search.backend.elasticsearch.gson.impl.JsonAccessor;
+import org.hibernate.search.backend.elasticsearch.search.common.impl.AbstractElasticsearchCodecAwareSearchQueryElementFactory;
+import org.hibernate.search.backend.elasticsearch.search.common.impl.ElasticsearchSearchIndexScope;
+import org.hibernate.search.backend.elasticsearch.search.common.impl.ElasticsearchSearchIndexValueFieldContext;
+import org.hibernate.search.backend.elasticsearch.search.predicate.impl.ElasticsearchSearchPredicate;
+import org.hibernate.search.backend.elasticsearch.types.codec.impl.ElasticsearchFieldCodec;
+import org.hibernate.search.engine.backend.types.converter.runtime.FromDocumentValueConvertContext;
+import org.hibernate.search.engine.backend.types.converter.spi.ProjectionConverter;
+import org.hibernate.search.engine.search.aggregation.spi.FieldMetricAggregationBuilder;
+import org.hibernate.search.engine.search.common.ValueModel;
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+
+/**
+ * @param <F> The type of field values.
+ * @param <K> The type of returned value. It can be {@code F}, {@link Double}
+ * or a different type if value converters are used.
+ */
+public class ElasticsearchMetricFieldAggregation<F, K> extends AbstractElasticsearchNestableAggregation<K> {
+
+	private static final JsonAccessor<JsonObject> SUM_PROPERTY_ACCESSOR =
+			JsonAccessor.root().property( "sum" ).asObject();
+
+	private static final JsonAccessor<JsonObject> MIN_PROPERTY_ACCESSOR =
+			JsonAccessor.root().property( "min" ).asObject();
+
+	private static final JsonAccessor<JsonObject> MAX_PROPERTY_ACCESSOR =
+			JsonAccessor.root().property( "max" ).asObject();
+
+	private static final JsonAccessor<JsonObject> AVG_PROPERTY_ACCESSOR =
+			JsonAccessor.root().property( "avg" ).asObject();
+
+	private static final JsonAccessor<String> FIELD_PROPERTY_ACCESSOR =
+			JsonAccessor.root().property( "field" ).asString();
+
+	private static final JsonAccessor<Double> VALUE_ACCESSOR =
+			JsonAccessor.root().property( "value" ).asDouble();
+
+	public static <F> ElasticsearchMetricFieldAggregation.Factory<F> sum(ElasticsearchFieldCodec<F> codec) {
+		return new ElasticsearchMetricFieldAggregation.Factory<>( codec, SUM_PROPERTY_ACCESSOR );
+	}
+
+	public static <F> ElasticsearchMetricFieldAggregation.Factory<F> min(ElasticsearchFieldCodec<F> codec) {
+		return new ElasticsearchMetricFieldAggregation.Factory<>( codec, MIN_PROPERTY_ACCESSOR );
+	}
+
+	public static <F> ElasticsearchMetricFieldAggregation.Factory<F> max(ElasticsearchFieldCodec<F> codec) {
+		return new ElasticsearchMetricFieldAggregation.Factory<>( codec, MAX_PROPERTY_ACCESSOR );
+	}
+
+	public static <F> ElasticsearchMetricFieldAggregation.Factory<F> avg(ElasticsearchFieldCodec<F> codec) {
+		return new ElasticsearchMetricFieldAggregation.Factory<>( codec, AVG_PROPERTY_ACCESSOR );
+	}
+
+	private final String absoluteFieldPath;
+	private final ProjectionConverter<F, ? extends K> fromFieldValueConverter;
+	private final ElasticsearchFieldCodec<F> codec;
+	private final JsonAccessor<JsonObject> operation;
+
+	private ElasticsearchMetricFieldAggregation(Builder<F, K> builder) {
+		super( builder );
+		this.absoluteFieldPath = builder.field.absolutePath();
+		this.fromFieldValueConverter = builder.fromFieldValueConverter;
+		this.codec = builder.codec;
+		this.operation = builder.operation;
+	}
+
+	@Override
+	protected final JsonObject doRequest(AggregationRequestContext context) {
+		JsonObject outerObject = new JsonObject();
+		JsonObject innerObject = new JsonObject();
+
+		operation.set( outerObject, innerObject );
+		FIELD_PROPERTY_ACCESSOR.set( innerObject, absoluteFieldPath );
+		return outerObject;
+	}
+
+	@Override
+	protected Extractor<K> extractor(AggregationRequestContext context) {
+		return new MetricFieldExtractor( nestedPathHierarchy, filter );
+	}
+
+	private static class Factory<F>
+			extends
+			AbstractElasticsearchCodecAwareSearchQueryElementFactory<FieldMetricAggregationBuilder.TypeSelector, F> {
+
+		private final JsonAccessor<JsonObject> operation;
+
+		private Factory(ElasticsearchFieldCodec<F> codec, JsonAccessor<JsonObject> operation) {
+			super( codec );
+			this.operation = operation;
+		}
+
+		@Override
+		public FieldMetricAggregationBuilder.TypeSelector create(ElasticsearchSearchIndexScope<?> scope,
+				ElasticsearchSearchIndexValueFieldContext<F> field) {
+			return new ElasticsearchMetricFieldAggregation.TypeSelector<>( codec, scope, field, operation );
+		}
+	}
+
+	private static class TypeSelector<F> implements FieldMetricAggregationBuilder.TypeSelector {
+		private final ElasticsearchFieldCodec<F> codec;
+		private final ElasticsearchSearchIndexScope<?> scope;
+		private final ElasticsearchSearchIndexValueFieldContext<F> field;
+		private final JsonAccessor<JsonObject> operation;
+
+		private TypeSelector(ElasticsearchFieldCodec<F> codec,
+				ElasticsearchSearchIndexScope<?> scope, ElasticsearchSearchIndexValueFieldContext<F> field,
+				JsonAccessor<JsonObject> operation) {
+			this.codec = codec;
+			this.scope = scope;
+			this.field = field;
+			this.operation = operation;
+		}
+
+		@Override
+		public <T> Builder<F, T> type(Class<T> expectedType, ValueModel valueModel) {
+			ProjectionConverter<F, ? extends T> projectionConverter = null;
+			if ( !Double.class.isAssignableFrom( expectedType )
+					||
+					field.type().projectionConverter( valueModel ).valueType().isAssignableFrom( expectedType ) ) {
+				projectionConverter = field.type().projectionConverter( valueModel )
+						.withConvertedType( expectedType, field );
+			}
+			return new Builder<>( codec, scope, field,
+					projectionConverter,
+					operation
+			);
+		}
+	}
+
+	private class MetricFieldExtractor extends AbstractExtractor<K> {
+		protected MetricFieldExtractor(List<String> nestedPathHierarchy, ElasticsearchSearchPredicate filter) {
+			super( nestedPathHierarchy, filter );
+		}
+
+		@Override
+		@SuppressWarnings("unchecked")
+		protected K doExtract(JsonObject aggregationResult, AggregationExtractContext context) {
+			FromDocumentValueConvertContext convertContext = context.fromDocumentValueConvertContext();
+			Optional<Double> value = VALUE_ACCESSOR.get( aggregationResult );
+			JsonElement valueAsString = aggregationResult.get( "value_as_string" );
+
+			if ( fromFieldValueConverter == null ) {
+				Double decode = value.orElse( null );
+				return (K) decode;
+			}
+			return fromFieldValueConverter.fromDocumentValue(
+					codec.decodeAggregationValue( value, valueAsString ),
+					convertContext
+			);
+		}
+	}
+
+	private static class Builder<F, K> extends AbstractBuilder<K>
+			implements FieldMetricAggregationBuilder<K> {
+
+		private final ElasticsearchFieldCodec<F> codec;
+		private final ProjectionConverter<F, ? extends K> fromFieldValueConverter;
+		private final JsonAccessor<JsonObject> operation;
+
+		private Builder(ElasticsearchFieldCodec<F> codec, ElasticsearchSearchIndexScope<?> scope,
+				ElasticsearchSearchIndexValueFieldContext<F> field,
+				ProjectionConverter<F, ? extends K> fromFieldValueConverter, JsonAccessor<JsonObject> operation) {
+			super( scope, field );
+			this.codec = codec;
+			this.fromFieldValueConverter = fromFieldValueConverter;
+			this.operation = operation;
+		}
+
+		@Override
+		public ElasticsearchMetricFieldAggregation<F, K> build() {
+			return new ElasticsearchMetricFieldAggregation<>( this );
+		}
+	}
+}
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/aggregation/impl/ElasticsearchMetricLongAggregation.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/aggregation/impl/ElasticsearchMetricLongAggregation.java
@@ -1,0 +1,108 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.search.backend.elasticsearch.search.aggregation.impl;
+
+import java.util.List;
+
+import org.hibernate.search.backend.elasticsearch.gson.impl.JsonAccessor;
+import org.hibernate.search.backend.elasticsearch.gson.impl.JsonElementTypes;
+import org.hibernate.search.backend.elasticsearch.search.common.impl.AbstractElasticsearchCodecAwareSearchQueryElementFactory;
+import org.hibernate.search.backend.elasticsearch.search.common.impl.ElasticsearchSearchIndexScope;
+import org.hibernate.search.backend.elasticsearch.search.common.impl.ElasticsearchSearchIndexValueFieldContext;
+import org.hibernate.search.backend.elasticsearch.search.predicate.impl.ElasticsearchSearchPredicate;
+import org.hibernate.search.backend.elasticsearch.types.codec.impl.ElasticsearchFieldCodec;
+import org.hibernate.search.engine.search.aggregation.spi.SearchFilterableAggregationBuilder;
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+
+public class ElasticsearchMetricLongAggregation extends AbstractElasticsearchNestableAggregation<Long> {
+
+	private static final JsonAccessor<JsonObject> COUNT_PROPERTY_ACCESSOR =
+			JsonAccessor.root().property( "value_count" ).asObject();
+
+	private static final JsonAccessor<JsonObject> COUNT_DISTINCT_PROPERTY_ACCESSOR =
+			JsonAccessor.root().property( "cardinality" ).asObject();
+
+	private static final JsonAccessor<String> FIELD_PROPERTY_ACCESSOR =
+			JsonAccessor.root().property( "field" ).asString();
+
+	public static <F> ElasticsearchMetricLongAggregation.Factory<F> count(ElasticsearchFieldCodec<F> codec) {
+		return new ElasticsearchMetricLongAggregation.Factory<>( codec, COUNT_PROPERTY_ACCESSOR );
+	}
+
+	public static <F> ElasticsearchMetricLongAggregation.Factory<F> countDistinct(ElasticsearchFieldCodec<F> codec) {
+		return new ElasticsearchMetricLongAggregation.Factory<>( codec, COUNT_DISTINCT_PROPERTY_ACCESSOR );
+	}
+
+	private final String absoluteFieldPath;
+	private final JsonAccessor<JsonObject> operation;
+
+	private ElasticsearchMetricLongAggregation(Builder builder) {
+		super( builder );
+		this.absoluteFieldPath = builder.field.absolutePath();
+		this.operation = builder.operation;
+	}
+
+	@Override
+	protected final JsonObject doRequest(AggregationRequestContext context) {
+		JsonObject outerObject = new JsonObject();
+		JsonObject innerObject = new JsonObject();
+
+		operation.set( outerObject, innerObject );
+		FIELD_PROPERTY_ACCESSOR.set( innerObject, absoluteFieldPath );
+		return outerObject;
+	}
+
+	@Override
+	protected Extractor<Long> extractor(AggregationRequestContext context) {
+		return new MetricLongExtractor( nestedPathHierarchy, filter );
+	}
+
+	private static class Factory<F>
+			extends
+			AbstractElasticsearchCodecAwareSearchQueryElementFactory<SearchFilterableAggregationBuilder<Long>, F> {
+
+		private final JsonAccessor<JsonObject> operation;
+
+		private Factory(ElasticsearchFieldCodec<F> codec, JsonAccessor<JsonObject> operation) {
+			super( codec );
+			this.operation = operation;
+		}
+
+		@Override
+		public SearchFilterableAggregationBuilder<Long> create(ElasticsearchSearchIndexScope<?> scope,
+				ElasticsearchSearchIndexValueFieldContext<F> field) {
+			return new Builder( scope, field, operation );
+		}
+	}
+
+	private static class MetricLongExtractor extends AbstractExtractor<Long> {
+		protected MetricLongExtractor(List<String> nestedPathHierarchy, ElasticsearchSearchPredicate filter) {
+			super( nestedPathHierarchy, filter );
+		}
+
+		@Override
+		protected Long doExtract(JsonObject aggregationResult, AggregationExtractContext context) {
+			JsonElement value = aggregationResult.get( "value" );
+			return JsonElementTypes.LONG.fromElement( value );
+		}
+	}
+
+	private static class Builder extends AbstractBuilder<Long> implements SearchFilterableAggregationBuilder<Long> {
+		private final JsonAccessor<JsonObject> operation;
+
+		private Builder(ElasticsearchSearchIndexScope<?> scope, ElasticsearchSearchIndexValueFieldContext<?> field,
+				JsonAccessor<JsonObject> operation) {
+			super( scope, field );
+			this.operation = operation;
+		}
+
+		@Override
+		public ElasticsearchMetricLongAggregation build() {
+			return new ElasticsearchMetricLongAggregation( this );
+		}
+	}
+}
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/codec/impl/ElasticsearchBigDecimalFieldCodec.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/codec/impl/ElasticsearchBigDecimalFieldCodec.java
@@ -12,6 +12,7 @@ import java.math.RoundingMode;
 import org.hibernate.search.backend.elasticsearch.gson.impl.JsonElementTypes;
 import org.hibernate.search.backend.elasticsearch.logging.impl.Log;
 import org.hibernate.search.engine.cfg.spi.NumberScaleConstants;
+import org.hibernate.search.engine.cfg.spi.NumberUtils;
 import org.hibernate.search.util.common.logging.impl.LoggerFactory;
 
 import com.google.gson.Gson;
@@ -63,6 +64,11 @@ public class ElasticsearchBigDecimalFieldCodec extends AbstractElasticsearchFiel
 		}
 
 		return JsonElementTypes.BIG_DECIMAL.fromElement( element );
+	}
+
+	@Override
+	public BigDecimal decode(Double value) {
+		return NumberUtils.toBigDecimal( value ).setScale( decimalScale, RoundingMode.FLOOR );
 	}
 
 	@Override
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/codec/impl/ElasticsearchFieldCodec.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/codec/impl/ElasticsearchFieldCodec.java
@@ -4,7 +4,10 @@
  */
 package org.hibernate.search.backend.elasticsearch.types.codec.impl;
 
+import java.util.Optional;
+
 import org.hibernate.search.backend.elasticsearch.lowlevel.syntax.search.impl.ElasticsearchSearchSyntax;
+import org.hibernate.search.util.common.AssertionFailure;
 
 import com.google.gson.JsonElement;
 
@@ -37,6 +40,10 @@ public interface ElasticsearchFieldCodec<F> {
 
 	F decode(JsonElement element);
 
+	default F decode(Double element) {
+		throw new AssertionFailure( this + " codec is not expected to handle decoding metric aggregation values." );
+	}
+
 	/**
 	 * Decodes the key returned by a term aggregation.
 	 * @param key The "key" property  returned by the aggregation.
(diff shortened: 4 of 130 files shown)


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/codec/impl/AbstractElasticsearchJavaTimeFieldCodec.java
```diff
@@ -60,6 +60,11 @@ public T decodeAggregationKey(JsonElement key, JsonElement keyAsString) {
 		return decode( keyAsString );
 	}
 
+	@Override
+	public T convertFromDoubleNullSafe(Double value) {
+		return null;
```
marko-bekhta, line 65: Have you tried date fields with the Elasticsearch backend? Is it not possible to do any metric aggregations with them? if not, this probably should fail with an assertion error and corresponding aggregations just not added to the `ElasticsearchIndexValueFieldType.Builder`...

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/codec/impl/AbstractElasticsearchVectorFieldCodec.java
```diff
@@ -53,6 +53,11 @@ public F decode(JsonElement element) {
 		return fromJsonArray( JsonElementTypes.ARRAY.fromElement( element ) );
 	}
 
+	@Override
+	public F convertFromDoubleNullSafe(Double value) {
+		return null;
```
marko-bekhta, line 58: It's unlikely that we'd want any aggregations for vectors, so maybe just an assertion error.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/codec/impl/ElasticsearchBigDecimalFieldCodec.java
```diff
@@ -65,6 +65,11 @@ public BigDecimal decode(JsonElement element) {
 		return JsonElementTypes.BIG_DECIMAL.fromElement( element );
 	}
 
+	@Override
+	public BigDecimal convertFromDoubleNullSafe(Double value) {
+		return BigDecimal.valueOf( value );
```
marko-bekhta, line 70: It's probably best to move these converter methods to the utils similar to `ParseUtils`.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/AbstractElasticsearchNumericFieldTypeOptionsStep.java
```diff
@@ -65,6 +68,19 @@ protected final void complete() {
...
+					new ElasticsearchMetricFieldAggregation.Factory<>( codec, "min" ) );
+			builder.queryElementFactory( AggregationTypeKeys.MAX,
+					new ElasticsearchMetricFieldAggregation.Factory<>( codec, "max" ) );
+			builder.queryElementFactory( AggregationTypeKeys.COUNT,
+					new ElasticsearchMetricLongAggregation.Factory<>( codec, "value_count" ) );
+			builder.queryElementFactory( AggregationTypeKeys.COUNT_DISTINCT,
+					new ElasticsearchMetricLongAggregation.Factory<>( codec, "cardinality" ) );
+			builder.queryElementFactory( AggregationTypeKeys.AVG,
+					new ElasticsearchMetricDoubleAggregation.Factory<>( codec, "avg" )
+			);
```
marko-bekhta, line 83: Maybe let's try testing these aggregations with other types as well and see which are supported (I'd hope that at least temporal types would work with some of these). When I say test, I mean just to create an index and run some simple aggregation HTTP request so that we don't need to add all the code if the backend does not even support it 🙈 😃.

engine/src/main/java/org/hibernate/search/engine/search/aggregation/dsl/impl/AvgAggregationFieldStepImpl.java
```diff
@@ -0,0 +1,30 @@
...
+	private final SearchAggregationDslContext<?, ? extends PDF> dslContext;
+
+	public AvgAggregationFieldStepImpl(SearchAggregationDslContext<?, ? extends PDF> dslContext) {
+		this.dslContext = dslContext;
+	}
+
+	@Override
+	public AvgAggregationOptionsStep<?, PDF> field(String fieldPath) {
+		MetricDoubleAggregationBuilder builder = dslContext.scope()
+				.fieldQueryElement( fieldPath, AggregationTypeKeys.AVG ).type();
```
marko-bekhta, line 27: a double average may not be the best for the temporal types. any chance we can follow the same pattern as for the sum?

integrationtest/backend/elasticsearch/src/test/java/org/hibernate/search/integrationtest/backend/elasticsearch/tmp/ElasticsearchMetricAggregationsIT.java
```diff
@@ -0,0 +1,165 @@
...
+		initData();
+	}
+
+	@Test
+	public void test_filteringResults() {
+		StubMappingScope scope = mainIndex.createScope();
+		SearchQuery<DocumentReference> query = scope.query()
+				.where( f -> f.match().field( "style" ).matching( "bla" ) )
+				.aggregation( sumIntegers, f -> f.sum().field( "integer", Integer.class ) )
+				.aggregation( sumConverted, f -> f.sum().field( "converted", String.class ) )
```
marko-bekhta, line 58: Since you have these new metric aggregations extend `AggregationFilterStep` maybe let's  have a test with nested objects and a filter to make sure that there are no surprises there 🙈 😃.

Also, maybe let's try to have some explicit `ValueConvert.YES/NO` ?

marko-bekhta (commented):
Nice progress 😃,

I've added some inline comments here and there. As for the `ElasticsearchFieldCodec#convertFromDouble(Double value)`:
- it is probably fine for now, while we are iterating on the implementation.
- let's figure out what it means for non-numeric types 
- if we don't go with doubles for the Lucene backend and it will stay only as an implementation detail for the Elasticsearch backend, we probably wouldn't need to expose it for any user customization so it can stay in the codec. otherwise maybe it'll make more sense to have something in `SearchIndexValueFieldTypeContext`

engine/src/main/java/org/hibernate/search/engine/search/aggregation/spi/MetricDoubleAggregationBuilder.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+package org.hibernate.search.engine.search.aggregation.spi;
+
+import org.hibernate.search.engine.search.predicate.SearchPredicate;
+
+public interface MetricDoubleAggregationBuilder extends SearchAggregationBuilder<Double> {
+
+	interface TypeSelector {
+		MetricDoubleAggregationBuilder type();
+	}
```
yrodiere, line 13: Note this "TypeSelector" step, as it is, is probably pointless. The only point of type selectors in other builders is to allow passing an expected type, e.g. like this:

https://github.com/hibernate/hibernate-search/blob/0b9e3f0d87250daa800f10223bcc81cdb0da773a/engine/src/main/java/org/hibernate/search/engine/search/aggregation/spi/TermsAggregationBuilder.java#L14-L16

I'm not saying you should remove this `TypeSelector` interface, but maybe there's something missing, e.g. a parameter to select the expected return type?
That's more or less what you're doing in `MetricFieldAggregationBuilder`...

fax4ever: You're right. In the last commit I remove the `type()` methods to support non-fields aggregations. The code is simplified now.

engine/src/main/java/org/hibernate/search/engine/search/aggregation/spi/MetricDoubleAggregationBuilder.java
```diff
@@ -0,0 +1,17 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.search.engine.search.aggregation.spi;
+
+import org.hibernate.search.engine.search.predicate.SearchPredicate;
+
+public interface MetricDoubleAggregationBuilder extends SearchAggregationBuilder<Double> {
```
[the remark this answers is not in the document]
fax4ever: Removed those interfaces. Thanks

fax4ever:
I just added two last commits to address the last @yrodiere's suggestions. Once we're fine with them, I'll reorganize the commits.

engine/src/main/java/org/hibernate/search/engine/search/aggregation/spi/AggregationTypeKeys.java
```diff
@@ -18,5 +18,17 @@ private AggregationTypeKeys() {
...
+	public static final SearchQueryElementTypeKey<MetricFieldAggregationBuilder.TypeSelector> MIN =
+			of( IndexFieldTraits.Aggregations.MIN );
+	public static final SearchQueryElementTypeKey<MetricFieldAggregationBuilder.TypeSelector> MAX =
+			of( IndexFieldTraits.Aggregations.MAX );
+	public static final SearchQueryElementTypeKey<SearchAggregationFilteredBuilder<Long>> COUNT =
+			of( IndexFieldTraits.Aggregations.COUNT );
+	public static final SearchQueryElementTypeKey<SearchAggregationFilteredBuilder<Long>> COUNT_DISTINCT =
+			of( IndexFieldTraits.Aggregations.COUNT_DISTINCT );
+	public static final SearchQueryElementTypeKey<SearchAggregationFilteredBuilder<Double>> AVG =
+			of( IndexFieldTraits.Aggregations.AVG );
```
yrodiere, line 32: I think the names you've selected are inconsistent with existing naming conventions.

* `MetricFieldAggregationBuilder` => `FieldMetricAggregationBuilder`
* `SearchAggregationFilteredBuilder` => `SearchFilteredAggregationBuilder` ... And I'd even say: `SearchFilterableAggregationBuilder`, since it's not necessarily filtered.

engine/src/main/java/org/hibernate/search/engine/search/aggregation/spi/MetricFieldAggregationBuilder.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @param <A> the type of the result of the aggregation
+ */
+public interface MetricFieldAggregationBuilder<A> extends SearchAggregationBuilder<A> {
+
+	interface TypeSelector {
+		<A> MetricFieldAggregationBuilder<A> type(Class<A> expectedType, ValueConvert convert);
+	}
+
+	void filter(SearchPredicate filter);
```
yrodiere, line 22: Shouldn't this interface extend `SearchAggregationFilteredBuilder`, then?

engine/src/main/java/org/hibernate/search/engine/search/aggregation/spi/AggregationTypeKeys.java
```diff
@@ -18,5 +18,17 @@ private AggregationTypeKeys() {
...
+			of( IndexFieldTraits.Aggregations.SUM );
+	public static final SearchQueryElementTypeKey<MetricFieldAggregationBuilder.TypeSelector> MIN =
+			of( IndexFieldTraits.Aggregations.MIN );
+	public static final SearchQueryElementTypeKey<MetricFieldAggregationBuilder.TypeSelector> MAX =
+			of( IndexFieldTraits.Aggregations.MAX );
+	public static final SearchQueryElementTypeKey<SearchAggregationFilteredBuilder<Long>> COUNT =
+			of( IndexFieldTraits.Aggregations.COUNT );
+	public static final SearchQueryElementTypeKey<SearchAggregationFilteredBuilder<Long>> COUNT_DISTINCT =
+			of( IndexFieldTraits.Aggregations.COUNT_DISTINCT );
+	public static final SearchQueryElementTypeKey<SearchAggregationFilteredBuilder<Double>> AVG =
```
yrodiere, line 31: This seems wrong: as a user, I'd expect an average on a `LocalTime` field to result in a `LocalTime` value. A `Double` would not mean anything.

IMO this case is very similar to `sum`, where you decided to go for `MetricFieldAggregationBuilder` (which I agree with), and you should do the same here.

yrodiere (commented):
Thanks. Since you asked for my opinion, here's some -- I'll let @marko-bekhta take over though.

engine/src/main/java/org/hibernate/search/engine/search/aggregation/spi/AggregationTypeKeys.java
fax4ever: I think you're right. Going to changing it to use `MetricFieldAggregationBuilder.TypeSelector` as sum.

fax4ever: I think you're right. Going to changing it to use `MetricFieldAggregationBuilder.TypeSelector` as sum.

fax4ever: I think you're right. Going to changing it to use `MetricFieldAggregationBuilder.TypeSelector` as sum.

fax4ever: I think you're right. Going to changing it to use `MetricFieldAggregationBuilder.TypeSelector` as sum.

marko-bekhta: 🙈🙈🙈🙈😃

fax4ever: @yrodiere changing the `avg` implementation from `ElasticsearchMetricDoubleAggregation` to `ElasticsearchMetricFieldAggregation`, it means that if for instance the field has type `Integer` the average will become an integer, losing the floating point values. I'm not sure it is actually a good idea. Anyway I'll keep this change in a different commit, so that we could easily rollback it in case.

fax4ever:
Thanks @yrodiere I applied all the great suggestions here.
My only doubt is a about the last change that I added in a different commit: https://github.com/hibernate/hibernate-search/pull/4144/commits/a807e2c39d8b1bb3dad7e708138f0a37631be94d.

In my opinion returning a double would be better, but I have an idea...
we could expose two avg operations, one returning always a double and one returning the field type to support the exotic cases such as avg of a data. I don't know what you think.
What you think about it @marko-bekhta?

engine/src/main/java/org/hibernate/search/engine/search/aggregation/spi/AggregationTypeKeys.java
yrodiere: Makes sense. Then maybe let's not provide a default? Change the DSL to force users to provide a type?

yrodiere:
> we could expose two avg operations, one returning always a double and one returning the field type to support the exotic cases such as avg of a date. I don't know what you think.

IMO we should just force users to pass a type. They ask for a date, we check if that's possible, and if so we return a date. They ask for an integer, same thing. They as for a double, ditto.

fax4ever:
> > we could expose two avg operations, one returning always a double and one returning the field type to support the exotic cases such as avg of a date. I don't know what you think.
> 
> IMO we should just force users to pass a type. They ask for a date, we check if that's possible, and if so we return a date. They ask for an integer, same thing. They as for a double, ditto.

IMO if you give the user the flexibility of requiring any type, it would be very complex (in term of API) and also error prone. I think that the solution of exposing a double for numeric avg and a field avg for all the types, it would safer and easier to use.

yrodiere:
> > > we could expose two avg operations, one returning always a double and one returning the field type to support the exotic cases such as avg of a date. I don't know what you think.
> > 
> > 
> > IMO we should just force users to pass a type. They ask for a date, we check if that's possible, and if so we return a date. They ask for an integer, same thing. They as for a double, ditto.
> 
> IMO if you give the user the flexibility of requiring any type, it would be very complex (in term of API) and also error prone. I think that the solution of exposing a double for numeric avg and a field avg for all the types, it would safer and easier to use.

I don't get this, they already have the flexibility to require any type: https://github.com/hibernate/hibernate-search/pull/4144/files#diff-6b1cb9769f69a0a38bbf7c3d35a53612d9d88e45f8180ba2cbe9eea36c1baea0R27-R41

And in fact there's already no default: if they get an `Integer` average, it can only be because they asked for it.

Also, please note I'm not saying we should comply with every request. IMO there are only two types that make sense: the field type (`ValueConvert.YES`) and the underlying, unconverted type (`ValueConvert.NO`, would be `double` I guess).

fax4ever:
> > > > we could expose two avg operations, one returning always a double and one returning the field type to support the exotic cases such as avg of a date. I don't know what you think.
> > > 
> > > 
> > > IMO we should just force users to pass a type. They ask for a date, we check if that's possible, and if so we return a date. They ask for an integer, same thing. They as for a double, ditto.
> > 
> > 
> > IMO if you give the user the flexibility of requiring any type, it would be very complex (in term of API) and also error prone. I think that the solution of exposing a double for numeric avg and a field avg for all the types, it would safer and easier to use.
> 
> I don't get this, they already have the flexibility to require any type: https://github.com/hibernate/hibernate-search/pull/4144/files#diff-6b1cb9769f69a0a38bbf7c3d35a53612d9d88e45f8180ba2cbe9eea36c1baea0R27-R41
> 
> And in fact there's already no default: if they get an `Integer` average, it can only be because they asked for it.
> 
> Also, please note I'm not saying we should comply with every request. IMO there are only two types that make sense: the field type (`ValueConvert.YES`) and the underlying, unconverted type (`ValueConvert.NO`, would be `double` I guess).

If I try to ask for a double a get this error in my test:

```
org.hibernate.search.util.common.SearchException: HSEARCH000584: Invalid type for returned values: 'java.lang.Double'. Expected 'java.lang.Integer' or a supertype.
Context: field 'integer'

	at org.hibernate.search.engine.backend.types.converter.spi.ProjectionConverter.withConvertedType(ProjectionConverter.java:75)
	at org.hibernate.search.backend.elasticsearch.search.aggregation.impl.ElasticsearchMetricFieldAggregation$TypeSelector.type(ElasticsearchMetricFieldAggregation.java:89)
	at org.hibernate.search.backend.elasticsearch.search.aggregation.impl.ElasticsearchMetricFieldAggregation$TypeSelector.type(ElasticsearchMetricFieldAggregation.java:71)
	at org.hibernate.search.engine.search.aggregation.dsl.impl.AvgAggregationFieldStepImpl.field(AvgAggregationFieldStepImpl.java:29)
	at org.hibernate.search.engine.search.aggregation.dsl.AvgAggregationFieldStep.field(AvgAggregationFieldStep.java:28)
	at org.hibernate.search.integrationtest.backend.elasticsearch.tmp.ElasticsearchMetricAggregationsIT.lambda$test_filteringResults$11(ElasticsearchMetricAggregationsIT.java:65)
	at org.hibernate.search.engine.search.query.dsl.spi.AbstractSearchQueryOptionsStep.aggregation(AbstractSearchQueryOptionsStep.java:179)
	at org.hibernate.search.integrationtest.backend.elasticsearch.tmp.ElasticsearchMetricAggregationsIT.test_filteringResults(ElasticsearchMetricAggregationsIT.java:65)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
```

fax4ever:
> unconverted type (ValueConvert.NO, would be double I guess).

OK so I can revert the last commit?

yrodiere:
> If I try to ask for a double a get this error in my test:

With `ValueConvert.NO`?

> > unconverted type (ValueConvert.NO, would be double I guess).
> 
> OK so I can revert the last commit?

From what I've seen that commit is fine and the DSL changes at least should stay, but I might be mistaken.

Can you point me to the exact failing test?

fax4ever:
> > If I try to ask for a double a get this error in my test:
> 
> With `ValueConvert.NO`?
> 
> > > unconverted type (ValueConvert.NO, would be double I guess).
> > 
> > 
> > OK so I can revert the last commit?
> 
> From what I've seen that commit is fine and the DSL changes at least should stay, but I might be mistaken.
> 
> Can you point me to the exact failing test?

Sure! Here is the commit I which I tried to ask for a double avg => https://github.com/hibernate/hibernate-search/pull/4144/commits/6abad63b20195e1039f6fd0304d07698d2ddf808. Thanks

marko-bekhta (commented):
Hey 👋🏻 😃 

You've made good progress on this one 👍🏻 Thanks! I've added some comments inline.

engine/src/main/java/org/hibernate/search/engine/search/aggregation/dsl/SearchAggregationFactory.java
```diff
@@ -61,6 +61,47 @@ public interface SearchAggregationFactory {
...
+	 * @return The next step.
+	 */
+	CountDistinctAggregationFieldStep<?> countDistinct();
+
+	/**
+	 * Perform the avg metrics aggregation.
+	 *
+	 * @return the next step.
+	 */
+	AvgAggregationFieldStep<?> avg();
```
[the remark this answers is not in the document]
fax4ever: I think it is a great idea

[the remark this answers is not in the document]
marko-bekhta: @fax4ever the `ValueModel` is now merged, so maybe you can use it instead of the `ValueConvert` in the new API? 🙈 🤞🏻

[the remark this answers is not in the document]
fax4ever: done!

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/aggregation/impl/ElasticsearchMetricLongAggregation.java
```diff
@@ -0,0 +1,90 @@
...
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+
+public class ElasticsearchMetricLongAggregation extends AbstractElasticsearchNestableAggregation<Long> {
+
+	private final String absoluteFieldPath;
+	private final String operation;
+
+	public ElasticsearchMetricLongAggregation(Builder builder) {
```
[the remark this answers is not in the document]
fax4ever: fixed thanks

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/dsl/impl/AbstractElasticsearchTemporalIndexFieldTypeOptionsStep.java
```diff
@@ -77,6 +79,18 @@ protected final void complete() {
...
+			builder.queryElementFactory( AggregationTypeKeys.MIN,
+					new ElasticsearchMetricFieldAggregation.Factory<>( codec, "min" ) );
+			builder.queryElementFactory( AggregationTypeKeys.MAX,
+					new ElasticsearchMetricFieldAggregation.Factory<>( codec, "max" ) );
+			builder.queryElementFactory( AggregationTypeKeys.COUNT,
+					new ElasticsearchMetricLongAggregation.Factory<>( codec, "value_count" ) );
+			builder.queryElementFactory( AggregationTypeKeys.COUNT_DISTINCT,
+					new ElasticsearchMetricLongAggregation.Factory<>( codec, "cardinality" ) );
+			builder.queryElementFactory( AggregationTypeKeys.AVG,
+					new ElasticsearchMetricFieldAggregation.Factory<>( codec, "avg" ) );
```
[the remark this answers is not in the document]
fax4ever: it looks a great idea!

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/codec/impl/ElasticsearchFieldCodec.java
```diff
@@ -49,6 +53,18 @@ default F decodeAggregationKey(JsonElement key, JsonElement keyAsString) {
 		return decode( key );
 	}
 
+	default F decodeAggregationValue(JsonElement value, JsonElement valueAsString) {
```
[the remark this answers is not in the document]
fax4ever: I think that we don't need to override this method.

> which filed types have valueAsString returned?

All the temporal fields.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/aggregation/impl/ElasticsearchMetricFieldAggregation.java
```diff
@@ -0,0 +1,150 @@
...
+		this.operation = builder.operation;
+	}
+
+	@Override
+	protected final JsonObject doRequest(AggregationRequestContext context) {
+		JsonObject outerObject = new JsonObject();
+		JsonObject innerObject = new JsonObject();
+
+		outerObject.add( operation, innerObject );
+		innerObject.addProperty( "field", absoluteFieldPath );
```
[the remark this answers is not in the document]
fax4ever: I have an idea of merging this idea with the factory methods... I'm pushing it

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/aggregation/impl/LuceneMetricNumericFieldAggregation.java
```diff
@@ -0,0 +1,194 @@
...
+			context.requireCollector( collectorFactory );
+		}
+		else if ( "avg".equals( operation ) ) {
+			SumCollectorFactory sumCollectorFactory = new SumCollectorFactory( source );
+			CountCollectorFactory countCollectorFactory = new CountCollectorFactory( source );
+			collectorKey = sumCollectorFactory.getCollectorKey();
+			countCollectorKey = countCollectorFactory.getCollectorKey();
+			context.requireCollector( sumCollectorFactory );
+			context.requireCollector( countCollectorFactory );
+		}
```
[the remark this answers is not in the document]
fax4ever: It's a good idea... doing it!

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/lowlevel/aggregation/collector/impl/KahanSummation.java
```diff
@@ -0,0 +1,79 @@
...
+	 * Increments the Kahan sum by adding two sums, and updating the correction term for reducing numeric errors.
+	 */
+	public KahanSummation add(double value, double delta) {
+		// If the value is Inf or NaN, just add it to the running tally to "convert" to
+		// Inf/NaN. This keeps the behavior bwc from before kahan summing
+		if ( Double.isFinite( value ) == false ) {
+			this.value = value + this.value;
+		}
+
+		if ( Double.isFinite( this.value ) ) {
```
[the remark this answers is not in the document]
fax4ever: I took this code from OpenSearch. Anyway, it is a good idea to improve it!

engine/src/main/java/org/hibernate/search/engine/search/aggregation/dsl/SearchAggregationFactory.java
```diff
@@ -61,6 +61,47 @@ public interface SearchAggregationFactory {
 	 */
 	TermsAggregationFieldStep<?> terms();
 
+	/**
+	 * Perform the sum metrics aggregation.
```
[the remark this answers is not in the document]
fax4ever: It seems that Elasticsearch doc uses the term `metrics aggregation`, while OpenSearch one uses `metric aggregation`.  About this contribution, so far I'm using `metric`  in the code and `metrics` only in commit comments... I would stick with `metric` if you agree with that.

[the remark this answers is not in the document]
marko-bekhta: ok 😃 thanks!

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/aggregation/MetricTemporalFieldsAggregationsIT.java
```diff
@@ -0,0 +1,177 @@
...
+import org.hibernate.search.util.impl.integrationtest.mapper.stub.BulkIndexer;
+import org.hibernate.search.util.impl.integrationtest.mapper.stub.SimpleMappedIndex;
+import org.hibernate.search.util.impl.integrationtest.mapper.stub.StubLoadingOptionsStep;
+import org.hibernate.search.util.impl.integrationtest.mapper.stub.StubMappingScope;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+public class MetricTemporalFieldsAggregationsIT {
```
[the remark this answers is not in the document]
fax4ever: sure! working on it!

fax4ever:
Hi @marko-bekhta, the commit [HSEARCH-5133 Test all types with metric aggregations](https://github.com/hibernate/hibernate-search/pull/4144/commits/9abb07f596266759166bf40ae67660369c775228) is in progress and I still need to address the other unresolved conversation.
I will continue to work on it when I come back from PTO in two weeks. Thanks for the support

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/aggregation/MetricTemporalFieldsAggregationsIT.java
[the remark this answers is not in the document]
fax4ever: I added it!

fax4ever:
With this last push I think I have only one comment to address, the one about introducing `abstract List<? extends CollectorFactory> requiredCollectors()`...

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/aggregation/impl/LuceneMetricNumericFieldAggregation.java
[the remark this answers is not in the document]
fax4ever: I tried to apply it

fax4ever:
With the last push I should have addressed all the comments.
If so, is there anything to improve here?
Thanks @marko-bekhta and @yrodiere

marko-bekhta (commented):
Thanks Fabio! 
I think we are getting close 🤞🏻😃. I've left some comments inline, but the main thing to solve is to address the `ValueModel.RAW` case; right now, it can fail with an exception 🙈

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/types/AnalyzedStringFieldTypeDescriptor.java
```diff
@@ -62,6 +63,11 @@ protected List<List<String>> createMultiResultingInSingleAfterMedian() {
 		};
 	}
 
+	@Override
+	public MetricAggregationsValues<String> metricAggregationsValues() {
+		return null;
```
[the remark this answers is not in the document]
fax4ever: I think it is a great idea!

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/operations/MetricAggregationsTestCase.java
```diff
@@ -0,0 +1,183 @@
...
+		result.apply( query );
+		return new Result<>( typeDescriptor.getJavaType(), metricAggregationsValues, result );
+	}
+
+	@Override
+	public String toString() {
+		return "Case{" + typeDescriptor + '}';
+	}
+
+	public static class Result<F> {
```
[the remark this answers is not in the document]
fax4ever: The idea is not to support sum but all the other functions for those types. They are `Year` and `YearMonth`. But I agree with you. I think we can live without supporting aggregations for these two types. :+1:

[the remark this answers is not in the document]
marko-bekhta: mmm, what I was trying to say is that maybe we should remove the sum aggregation from here:
https://github.com/hibernate/hibernate-search/pull/4144/files#diff-66afc25d1cae294503e1f467a45527a70e365c9688412d9e2bb14fce3f431fd5R95-R96 :

```java
builder.queryElementFactory( AggregationTypeKeys.SUM,
					new AbstractLuceneMetricNumericFieldAggregation.Factory<>( codec, "sum" ) );
```
(and same for the ES backend), but just for these two types you've mentioned.

My thinking here is that if the backend gives some strange result, then we shouldn't even give the user an option to request it ....

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/aggregation/impl/AbstractLuceneMetricNumericFieldAggregation.java
```diff
@@ -187,8 +163,22 @@ public Builder(AbstractLuceneNumericFieldCodec<F, E> codec, LuceneSearchIndexSco
...
+			}
+			else if ( "max".equals( operation ) ) {
+				return new LuceneMaxNumericFieldAggregation<>( this );
+			}
+			else if ( "avg".equals( operation ) ) {
+				return new LuceneAvgNumericFieldAggregation<>( this );
+			}
+			else {
+				throw new AssertionFailure( "Aggregation operation not supported: " + operation );
+			}
```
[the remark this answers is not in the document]
fax4ever: I think that we can do a step further and push also the builder in each `*NumericFieldAddregation`, but I guess that if we really want to remove the if-else chain, we need to push also the `TypeSelector` and the `Factory` in each `*NumericFieldAddregation`. I'm trying to do a different commit. I'm not sure the overall code is nicer. Usually I agree it is better to avoid if and use subclasses, but maybe here the use case is quite complex, since we have for each case sub classes like: a factory, a type selector and the builder. Anyway, let's try to do that.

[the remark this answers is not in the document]
fax4ever: Another option would be to introduce a method to create the factory, passing the codec and the method reference to each `*NumericFieldAddregation` constructor... I don't know what you think about it

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/testsupport/operations/MetricAggregationsTestCase.java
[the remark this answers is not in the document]
fax4ever: Ok sorry. I didn't understand. I'll push the fix on the next force push on this contribution (soon).

backend/lucene/src/main/java/org/hibernate/search/backend/lucene/types/aggregation/impl/AbstractLuceneMetricNumericFieldAggregation.java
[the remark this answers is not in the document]
fax4ever: I tried!

fax4ever:
With last push... again just one comment left!

fax4ever:
Hi @marko-bekhta. Comment addressed (I hope :P)

To fully support the RAW model type, I opened [HSEARCH-5230 Support RAW model type with metric aggregations](https://hibernate.atlassian.net/browse/HSEARCH-5230).
Let me know if we can improve something else.
Thanks!

documentation/src/main/asciidoc/public/reference/_search-dsl-aggregation.adoc
```diff
@@ -324,6 +324,117 @@ See <<search-dsl-argument-type>> for more information.
...
+include::{sourcedir}/org/hibernate/search/documentation/search/aggregation/AggregationDslIT.java[tags=max]
+----
+<1> Define the target field path to which you want to apply the aggregation function and the expected returned type.
+====
+
+=== Count metric aggregation
+
+The `count` aggregation counts the field values.
+
+.Count anytime the price field has a value among all the science fiction books
```
marko-bekhta, line 396:
```suggestion
.Count the number of the science fiction books with prices
```

documentation/src/main/asciidoc/public/reference/_search-dsl-aggregation.adoc
```diff
@@ -324,6 +324,117 @@ See <<search-dsl-argument-type>> for more information.
...
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/aggregation/AggregationDslIT.java[tags=max]
+----
+<1> Define the target field path to which you want to apply the aggregation function and the expected returned type.
+====
+
+=== Count metric aggregation
+
+The `count` aggregation counts the field values.
```
marko-bekhta, line 394: It's
```suggestion
The `count` aggregation counts the number of documents in which the field is not empty. 
```

or not ? 🙈

documentation/src/main/asciidoc/public/reference/_search-dsl-aggregation.adoc
```diff
@@ -324,6 +324,117 @@ See <<search-dsl-argument-type>> for more information.
...
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/aggregation/AggregationDslIT.java[tags=count]
+----
+<1> Define the target field path to which you want to apply the aggregation function. For this function a `Long.class` value is always returned.
+====
+
+=== Count distinct metric aggregation
+
+The `count distinct` aggregation counts the different field values.
```
marko-bekhta, line 407:
```suggestion
The `count distinct` aggregation counts the number of unique field values.
```
does it also count an empty value towards the total?

documentation/src/main/asciidoc/public/reference/_search-dsl-aggregation.adoc
```diff
@@ -324,6 +324,117 @@ See <<search-dsl-argument-type>> for more information.
...
+include::{sourcedir}/org/hibernate/search/documentation/search/aggregation/AggregationDslIT.java[tags=count-distinct]
+----
+<1> Define the target field path to which you want to apply the aggregation function. For this function a `Long.class` value is always returned.
+====
+
+=== Avg metric aggregation
+
+Applies the SQL `avg` aggregation function on a given numeric or temporal field.
+
+.Compute the average of the prices all of the science fiction books
```
marko-bekhta, line 422:
```suggestion
.Compute the average price of all science fiction books
```

documentation/src/main/asciidoc/public/reference/_search-dsl-aggregation.adoc
```diff
@@ -324,6 +324,117 @@ See <<search-dsl-argument-type>> for more information.
...
+[source, JAVA, indent=0, subs="+callouts"]
+----
+include::{sourcedir}/org/hibernate/search/documentation/search/aggregation/AggregationDslIT.java[tags=count-distinct]
+----
+<1> Define the target field path to which you want to apply the aggregation function. For this function a `Long.class` value is always returned.
+====
+
+=== Avg metric aggregation
+
+Applies the SQL `avg` aggregation function on a given numeric or temporal field.
```
marko-bekhta, line 420: I think we've missed this one as we were replacing the sql-based definition in the previous iteration 🙈. Maybe
```suggestion
The `avg` aggregation calculates the average value of a given numeric or temporal field among the selected documents.
```

engine/src/main/java/org/hibernate/search/engine/search/aggregation/dsl/AvgAggregationFieldStep.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.hibernate.search.engine.search.common.ValueModel;
+import org.hibernate.search.engine.search.predicate.dsl.SearchPredicateFactory;
+
+/**
+ * The initial step in an "avg" aggregation definition, where the target field can be set.
+ *
+ * @param <PDF> The type of factory used to create predicates in {@link AggregationFilterStep#filter(Function)}.
+ */
+public interface AvgAggregationFieldStep<PDF extends SearchPredicateFactory> {
```
marko-bekhta, line 17: could you please also go through these new DSL interfaces and mark them as incubating? You've already did it for the new methods in the factories, and we probably won't change these interfaces, but just to be on the safe side ... 🙈 😃 thanks!

marko-bekhta (commented):
Thanks Fabio!
This looks nice 🎉 . Let's make sure we marked everything as incubating so that we will have some freedom to adjust if needed, and then we'll continue in the follow-up PRs as needed.

engine/src/main/java/org/hibernate/search/engine/search/aggregation/dsl/AvgAggregationFieldStep.java
fax4ever: Sure!

marko-bekhta: thanks!

fax4ever:
I integrate the corrections for the doc and incubating rebasing the contribution on top of the current main branch. Thanks

marko-bekhta (approved):
Thanks @fax4ever ! 🎉 

I'll merge this one and we can do the followups to address the `RAW` `ValueModel` and tests for other value models.

fax4ever:
Thank you Marko and Yoann

PR #4156, opened 2024-05-09
HSEARCH-3319 Type-safe field references / HSEARCH-5300 Annotation processor for the Search static metamodel

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-3319
https://hibernate.atlassian.net/browse/HSEARCH-5300

----------------------

----------------------


engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/SearchPredicateFactory.java
```diff
@@ -348,7 +368,22 @@ PredicateFinalStep withParameters(
...
+	 * This is used to call reusable methods that apply the same predicate
+	 * on different object fields that have same structure (same sub-fields).
+	 *
+	 * @param objectFieldReference The reference representing the path from the current root to an object field that will become the new root.
+	 * @return A new predicate factory using the given object field as root.
+	 */
+	@Incubating
+	default SearchPredicateFactory<SR> withRoot(ObjectFieldReference<? super SR> objectFieldReference) {
+		return withRoot( objectFieldReference.absolutePath() );
+	}
```
marko-bekhta, line 386: Even though I've added these `withRoot(..)` methods to factories, I am not sure that they make much sense with the approach references are taking...

Assuming we somehow address the paths, the factory would still point to the same `<SR>` root type, as the properties in the embedded objects of the "generated" metamodel will use them:

```java
public static class EntityA_ {
	public static ValueFieldReference1<EntityA_, String, String, String> stringA;

	public static SomeNestedObject_ someNestedObject;

	public static RootReferenceScope<EntityA_, EntityA> scope;

	static {
		stringA = ValueFieldReference1.of( "stringA", EntityA_.class, String.class, String.class, String.class );
		someNestedObject = new SomeNestedObject_();

		scope = RootReferenceScopeImpl.of( EntityA_.class, EntityA.class );
	}

	private static class SomeNestedObject_ implements ObjectFieldReference<EntityA_> {

		public final ValueFieldReference1<EntityA_, String, String, String> nestedStringA;

		public SomeNestedObject_() {
			this.nestedStringA = ValueFieldReference1.of( "someNestedObject.stringA", EntityA_.class, String.class, String.class, String.class );
		}

		@Override
		public String absolutePath() {
			return "someNestedObject";
		}

		@Override
		public Class<EntityA_> scopeRootType() {
			return EntityA_.class;
		}
	}
}
```

If we go with the different `<SR>` for `nestedStringA` then we wouldn't be able to use it in a regular case e.g.:
```java
...
.where( f -> f.match().field( EntityA_.someNestedObject.nestedStringA ).matching( "a" ) )
```
Also, if we change the `SR` on `withRoot` call, the current idea for these embedded objects to create a class-per-field would not fly 😔. 

If we ignore the `<SR>` type change, then any property can be passed to the factory methods, which makes the static metamodel pointless 😔 (in this case)

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/scope/SearchScope.java
```diff
@@ -38,7 +38,8 @@
  *
  * @param <E> A supertype of all types in this scope.
  */
-public interface SearchScope<E> {
+@SuppressWarnings("deprecation")
+public interface SearchScope<E> extends org.hibernate.search.engine.mapper.scope.SearchScope<E, org.hibernate.search.mapper.orm.common.EntityReference> {
```
marko-bekhta: Obviously, JQAssistant does not like these changes 😔 😃. However, I couldn't come up with an alternative to allow scope to be created from the "generated" static metamodel classes. Unless we generate a model specific to the mapper 😖 😕 

@yrodiere if you can take a quick look at the approach in this PR whenever you have some free time 😃

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/SearchPredicateFactory.java
yrodiere: Ideally, I'd expect after using `withRoot()`, people using the factory would use references such as `SomeNestedObject_.nestedStringA`, with `nestedStringA` being a *static* field. Just like they would on root types.

So you'd have two classes per object field:

* `SomeNestedObject$Absolute` (the name doesn't matter much, but must not conflict with other classes), which would be the type of `EntityA_.someNestedObject`, and which would expose non-static "sub" fields of type `ValueFieldReference<EntityA_, ...>`.
* `SomeNestedObject_` which would be the type users of `withRoot()` would rely on, and which would expose non-static "sub" fields of type `ValueFieldReference<SomeNestedObject_, ...>`.

I don't think you need an inheritance relationship between these two classes, but you may need to redefine `ObjectFieldReference` to have two generic type parameters: the first being its "holding" type, and the second being the `SR` of factories obtained by calling `withRoot`.

That being said... Since `SomeNestedObject_` is completely specific to the `nestedString` field of `EntityA`, I'm wondering how useful `withRoot()` would be. The main point of `withRoot` is to obtain a factory that can be passed to some util method, which is then able to build a predicate. That method is supposed to be used with different object fields that are assumed to have a similar schema.
If that method works for a specific object field only (with a specific metamodel type, e.g. `SomeNestedObject_`), then... what's the point?

I suppose it all comes down to the "mapped superclass"/inheritance support I mentioned the other time. If we could make sure that object field metamodel uses inheritance, i.e. the metamodel type `EntityB_.SomeNestedObject_` extends `EntityA_.SomeNestedObject_` as soon as `EntityB` extends `EntityA`, then you'd at least be able to create a root from `EntityB_.someNestedObject` and then use `EntityA_.SomeNestedObject_.nestedStringA` on the resulting factory.

Union types at the object field level could be a thing too, but that seems... hard. We'd essentially need `SomeNestedObject_` to extend all its union types. That means multiple inheritance, which means `SomeNestedObject_` must be an interface. Possible, since interfaces can have static fields, but needs investigation.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/scope/SearchScope.java
yrodiere: > Unless we generate a model specific to the mapper 😖 😕

That seems like a very valid solution to me?

Not for every Hibernate Search type in the metamodel, of course -- i.e. I wouldn't make `ValueFieldReference` specific to each mapper. But it seems perfectly fine to me to give `EntityA_.scope` a mapper-specific type.

marko-bekhta: > But it seems perfectly fine to me to give EntityA_.scope a mapper-specific type.

mm ok, we'd need to tell the plugin generating these to know which mappers are expected and then have 
```java
EntityA_.ormScope
EntityA_.standaloneScope
```
we probably can detect that from the dependencies (which mappers are present) and have an explicit property if needed I suppose 🤔

yrodiere: > mm ok, we'd need to tell the plugin generating these to know which mappers are expected and then have

The plugin would already need to know, since it needs to start Hibernate Search in offline mode to create backends and their index descriptors.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/SearchPredicateFactory.java
marko-bekhta: Hmm ok I've come up with the next example. Assuming we have a model as
```java
@Indexed
	public static class EntityA {

		@IndexedEmbedded(includePaths = { "a", "b" })
		public EmbeddedThing embA;

		@IndexedEmbedded(includePaths = { "b", "c" })
		public EmbeddedThing embB;
	}

	public static class EmbeddedThing {
		@FullTextField
		String a;
		@FullTextField
		String b;
		@FullTextField
		String c;
	}
```
and want to have an util to filter by `smth.b` using this `withRoot` approach...

we can generate the metamodel as:
```java
	public static class EntityA_ {

		public static a_ embA;
		public static b_ embB;

		// These two are for the absolute paths: 
		public static class a_ extends ObjectFieldReference<EntityA_, EntityA_a_>{
			public static ValueFieldReference<EntityA_, String, String, String> a; // path = embA.a
			public static ValueFieldReference<EntityA_, String, String, String> b; // path = embA.b
		}

		public static class b_ extends ObjectFieldReference<EntityA_, EntityA_b_>{
			public static ValueFieldReference<EntityA_, String, String, String> b; // path = embB.b
			public static ValueFieldReference<EntityA_, String, String, String> c; // path = embB.c
		}
	}

	// These two are for "withRoot"/  relative paths 
	public static class EntityA_a_ {
		public static ValueFieldReference<EntityA_a_, String, String, String> a; // path = a
		public static ValueFieldReference<EntityA_a_, String, String, String> b; // path = b
	}

	public static class EntityA_b_ {
		public static ValueFieldReference<EntityA_b_, String, String, String> b; // path = b
		public static ValueFieldReference<EntityA_b_, String, String, String> c; // path = c
	}
```
That won't help much for the with `withRoot` ... so we can try:
```java
	// ideally I guess we'd want to generate:
	public static class EntityA_a_b_common_ {
		public static ValueFieldReference<EntityA_a_b_common_, String, String, String> b; // path = b
	}

	public static class EntityA_a_ extends EntityA_a_b_common_ {
		public static ValueFieldReference<EntityA_a_, String, String, String> b; // path = b
	}
	
	public static class EntityA_b_ extends EntityA_a_b_common_ {
		public static ValueFieldReference<EntityA_b_, String, String, String> c; // path = c
	}
	
	// and then the util method could use  EntityA_a_b_common_ with access to `b` 
	//   and we pass withRoot(EntityA_.embA) withRoot(EntityA_.embB) to the util method ... 
```

that could probably work.. but if I understand it correctly... currently that withRoot is not tied to a particular embedded type, meaning we can have:
```java
public static class EntityA {
	@IndexedEmbedded
	EmbeddedThing1 e1;
	@IndexedEmbedded
	EmbeddedThing1 e2;
}

public static class EmbeddedThing1 {
	@FullTextField
	String a;
	// some other fields
}

public static class EmbeddedThing2 {
	@FullTextField
	String a;
	// some other fields maybe different from EmbeddedThing1
}

public static class EntityB {
	@IndexedEmbedded
	EmbeddedThing3 e3;
}

public static class EmbeddedThing3 {
	@FullTextField
	String a;
	// some other fields maybe different from EmbeddedThing1/EmbeddedThing2
}

//util method
private SearchPredicate matchFirstAndLastName(SearchPredicateFactory f, String avalue) {
	return f.match().field( "a" ).matching( avalue ).toPredicate();
}
```
it is ok now to pass any of 
```java
withRoot("e1");
withRoot("e2");
withRoot("e3");
```
but I'm not sure how we'd tie them all to something in the metamodel to be used in that util... 😕

yrodiere: > but I'm not sure how we'd tie them all to something in the metamodel to be used in that util... 😕

Essentially, to go beyond simple entity inheritance, we'll need to introduce union types at the object field level.

That's what I was talking about here:

> Union types at the object field level could be a thing too, but that seems... hard. We'd essentially need `SomeNestedObject_` to extend all its union types. That means multiple inheritance, which means `SomeNestedObject_` must be an interface. Possible, since interfaces can have static fields, but needs investigation.

marko-bekhta: ok next iteration 🙈 I've tried to look at the interfaces and whether we can somehow use them to our advantage. This time, the model is:
```java
@MappedSuperclass
public static class MappedSuperclassThing {
	@Id
	Long id;

	@FullTextField
	String a;
}

@Indexed
@Entity
public static class ContainingA extends MappedSuperclassThing {
	@IndexedEmbedded
	EmbeddedThing1 e1;
	@IndexedEmbedded
	EmbeddedThing1 e2;
}

public static class EmbeddedThing1 {
	@FullTextField
	String a;
	@FullTextField
	String b;
	// some other fields
}

public static class EmbeddedThing2 {
	@FullTextField
	String a;
	// some other fields maybe different from EmbeddedThing1
}

@Entity
@Indexed
public static class ContainingB extends MappedSuperclassThing{
	@IndexedEmbedded
	EmbeddedThing3 e3;
}

public static class EmbeddedThing3 {
	@GenericField
	Integer a;
	// some other fields maybe different from EmbeddedThing1/EmbeddedThing2
}
```
There are string/int fields `a` and some embeddables. Ultimately, we'd want to be able to have a util method as:
```java
public static SearchPredicate utilMethodForPredicate(SearchPredicateFactory<? extends Property_String_a_> factory) {
	return factory.match().field( Property_String_a_.a ).matching( "a" ).toPredicate();
}
```
where we want to accept the factories for "scope roots" that have a string property `a` in them. Which can be based on `ContainingA`/`ContainingB`/`EmbeddedThing1`/`EmbeddedThing2`. For the util to work `a` should be a part of the `Property_String_a_`:
```java
interface Property_String_a_ {
	ValueFieldReference1<Property_String_a_, String, String, String> a = ....;
}
```
then we can have the metamodel as:
```java
public static class MappedSuperclassThing_ implements Property_String_a_ {

	public static HibernateOrmRootReferenceScope<MappedSuperclassThing_, MappedSuperclassThing> scope;

	static {
		//a = ValueFieldReference1.of( "a", MappedSuperclassThing_.class, String.class, String.class, String.class );

		scope = RootReferenceScopeImpl.of( MappedSuperclassThing_.class, MappedSuperclassThing.class );
	}
}

public static class ContainingA_ extends MappedSuperclassThing_ {
	public static e1_.Absolute e1;
	public static e2_.Absolute e2;

	public static HibernateOrmRootReferenceScope<ContainingA_, ContainingA> scope;

	static {
		e1 = new e1_.Absolute();
		e2 = new e2_.Absolute();

		scope = RootReferenceScopeImpl.of( ContainingA_.class, ContainingA.class );
	}

	public static class e1_ implements ObjectFieldReference<ContainingA_>, Property_String_a_, Property_String_b_ {

		@Override
		public String absolutePath() {
			return "e1";
		}

		@Override
		public Class<ContainingA_> scopeRootType() {
			return ContainingA_.class;
		}

		static class Absolute extends e1_ {
			public static ValueFieldReference1<ContainingA_, String, String, String> a;
			public static ValueFieldReference1<ContainingA_, String, String, String> b;

			static {
				a = ValueFieldReference1.of( "e1.a", ContainingA_.class, String.class, String.class, String.class );
				b = ValueFieldReference1.of( "e1.b", ContainingA_.class, String.class, String.class, String.class );
			}
		}
	}

	public static class e2_ implements ObjectFieldReference<ContainingA_>, Property_String_a_, Property_String_b_ {

		@Override
		public String absolutePath() {
			return "e2";
		}

		@Override
		public Class<ContainingA_> scopeRootType() {
			return ContainingA_.class;
		}

		static class Absolute extends e2_ {
			public static ValueFieldReference1<ContainingA_, String, String, String> a;
			public static ValueFieldReference1<ContainingA_, String, String, String> b;

			static {
				a = ValueFieldReference1.of( "e2.a", ContainingA_.class, String.class, String.class, String.class );
				b = ValueFieldReference1.of( "e2.b", ContainingA_.class, String.class, String.class, String.class );
			}
		}
	}
}
```
Since we use `? super SR` we can do
```java
SearchScope<ContainingA_, ContainingA> scope = ContainingA_.scope.create( searchSession );

searchSession.search( scope )
		.select( f -> f.field( ContainingA_.a ) )
		.where( f -> utilMethodForPredicate( f ) )
		.fetchHits( 20 )
```
as `ContainingA_` extends `Property_String_a_` it works both for projection and util.
Hence, I'm thinking - we can create an interface-per-filed, where we reuse the interfaces with the same name and capabilities(traits)...

marko-bekhta (commented):
Hey @yrodiere 😃  the approach we discussed last Friday seems to work well so far. (see the new inline comments or just take a look at the `FieldReferenceIT` )

One thing I'm not sure how to deal with is named predicates. They are defined in binders, where there's no notion of search scope. Since binders would define the index structure, does it make sense to use the generated metamodel in them? as its generation depends on the binder ... 🤔 🙈

documentation/src/test/java/org/hibernate/search/documentation/search/predicate/FieldReferenceIT.java
```diff
@@ -0,0 +1,589 @@
...
+
+	// IMPL_NOTE: note cannot use the EntityClassName_ since ORM picks it up and tries to its thing...
+	// so we'd need to come up with a different naming strategy...
+	public static class ContainingA__
+			implements HibernateOrmRootReferenceScope<ContainingA__, ContainingA>,
+			ContainingA_e1_e2_Intersection<ContainingA__> {
+
+		// IMPL_NOTE: Maybe let's use the INDEX name?
+		// also I'm thinking we can make it configurable and let the user decide how to call this variable...
+		public static final ContainingA__ INDEX = new ContainingA__();
```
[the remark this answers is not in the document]
yrodiere: But then for union scopes, you'd have... `INDEXES`?

Wouldn't `SCOPE` be a better alternative? Then you'd have `ContainingA__` implement `SearchScopeReferenceWhatever`.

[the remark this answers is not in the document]
yrodiere: Or `SCOPE_REF`, or `scopeRef()`, or...

documentation/src/test/java/org/hibernate/search/documentation/search/predicate/FieldReferenceIT.java
```diff
@@ -0,0 +1,589 @@
...
+				MatchPredicateFieldReference<SR, T>,
+				ExistsPredicateFieldReference<SR>,
+				QueryStringPredicateFieldReference<SR, T> {
+
+			@Override
+			default ValueConvert valueConvert() {
+				return ValueConvert.YES;
+			}
+		}
+	}
```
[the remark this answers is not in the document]
yrodiere: Are you sure there's a need to have special handling for the `TraitsInteresection` interface?

I mean, we already need to generate interfaces/classes for combination of traits, so we already assume something will be generating a `FieldProjectionAndMatchPredicateAndExistsPredicateAndQueryStringPredicateFieldReference`; can't we use that here, too?

Unrelated note: naming these "trait combination" interfaces will be a challenge in itself.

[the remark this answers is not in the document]
marko-bekhta: > Are you sure there's a need to have special handling for the TraitsInteresection interface?

thinking more about it, yeah we can probably reuse the interfaces... I was thinking about a scenario:

```java
@KeywordField(projectable = Projectable.YES, sortable = Sortable.NO)
private String title;

@KeywordField(projectable = Projectable.NO, sortable = Sortable.YES)
private String title;
```
where the intersection would be just predicates and we wouldn't have it since it is not needed for an actual field, but we could just keep creating these combinations and add them to a lookup for reuse, so +1 😃 

> Unrelated note: naming these "trait combination" interfaces will be a challenge in itself.

yeah... At some point I was thinking to have a sorted list of trait interfaces and then "map" them to some numbered keys, e.g. T1, T2... Tn, and the name of a combination will be based on these keys, e.g. `T1T5T24FieldReferece`

marko-bekhta:
So, in the new commit, I've tried adding a Maven plugin that would generate things. The approach I've taken here is to have model+ search config classes in their own jar that is passed as a dependency to the Maven plugin. The plugin then starts the Search. I've used some of our test helpers to start things since, well, it's just to test how it'll work.

It generates something like:

```java
package org.hibernate.search.metamodel.generator.model;

class MyEntity__ {

	public String text;	
}
```
```java
package org.hibernate.search.metamodel.generator.model;

class MyProgrammaticEntity__ {

	public String number;
	public String text;	
}
```
And since it happens on the generate sources phase, these are getting compiled automatically by Maven itself. 

As for the plugin configuration:
```xml
<configuration>
    <annotatedTypes>
        <type>org.hibernate.search.metamodel.generator.model.MyEntity</type>
        <type>org.hibernate.search.metamodel.generator.model.MyProgrammaticEntity</type>
    </annotatedTypes>
    <properties>
        <hibernate.search.mapping.configurer>org.hibernate.search.metamodel.generator.model.MyConfigurer</hibernate.search.mapping.configurer>
    </properties>
</configuration>
```
There's a way to pass any properties that should be "forwarded" to the Search through `properties`. In this case, I've tried passing a configurer to apply some programmatic mapping 🙈.

now.... next thing I'd want to try is to run the plugin in the same module that the model is. From what I've seen so far, it is easy to get the source root, and the dependencies (things that were missing in the AP approach).

metamodel/processor/src/main/java/org/hibernate/search/metamodel/processor/impl/IndexedEntityMetamodelAnnotationProcessor.java
```diff
@@ -0,0 +1,150 @@
...
+													"Unexpected location of the " + indexedAnnotation.getQualifiedName()
+															+ ". Expected to be placed on an indexed type.",
+													el
+											);
+								}
+							}
+
+						} )
+				).build()
+				.boot() ) {
```
marko-bekhta, line 101: The main idea behind this approach is that we are starting a standalone mapper with all possible "annotation discovery" disabled since an annotation processor won't play nice with here 😃.
Then there is a mapping configurer that iterates through the `@Indexed ` entities and uses programmatic mapping to build the model.

metamodel/processor/src/main/java/org/hibernate/search/metamodel/processor/impl/IndexedEntityMetamodelAnnotationProcessor.java
```diff
@@ -0,0 +1,150 @@
...
+			for ( SearchIndexedEntity<?> entity : searchMapping.allIndexedEntities() ) {
+				context.messager().printMessage( Diagnostic.Kind.NOTE, entity.name() );
+				entity.indexManager().descriptor().staticFields()
+						.forEach( f -> context.messager().printMessage( Diagnostic.Kind.NOTE, f.toString() ) );
+			}
```
marko-bekhta, line 106: for now I just print the fields out so it's easy to see what is produced, but I've already started looking into how to generate classes. From what I see, Hibernate ORM/Tools are writing the classes with custom code, I also found this https://github.com/palantir/javapoet and was considering whether it would make sense to pull a dependency, or just do yet another custom class writer 😄

metamodel/processor/src/main/java/org/hibernate/search/metamodel/processor/impl/ProcessorElementUtils.java
```diff
@@ -0,0 +1,94 @@
...
+						if ( types.isSameType( returnType, annotationType ) ) {
+							return ( (List<? extends AnnotationMirror>) entry.getValue().getValue() ).stream();
+						}
+					}
+				}
+
+			}
+		}
+		return Stream.of( annotationMirror );
+	}
```
marko-bekhta, line 65: had to "reinvent" the wheel 😄 as these repeatable annotations are wrapped in the list one, even if the user didn't do that explicitly 🫨 (just adapted the code used in the mapper to expend these repeatable annotations)

metamodel/processor/src/main/java/org/hibernate/search/metamodel/processor/mapping/ProcessorPojoModelsBootstrapIntrospector.java
```diff
@@ -0,0 +1,76 @@
...
+				k -> {
+					Optional<Class<?>> loadableType = BuiltInBridgeResolverTypes.loadableType( typeElement.asType() );
+					if ( loadableType.isPresent() ) {
+						return typeModel( loadableType.get() );
+					}
+					else {
+						return new ProcessorPojoRawTypeModel<>( typeElement, context.processorContext(), this );
+					}
+				} );
+	}
```
marko-bekhta, line 66: you'll see a few options to get the types by this introspector...
to leverage all built-in bridges etc, we have to pass them the Java types they use in a regular run, rather than typelements. There's a list of "loadable" types (I'm planning to add a test to make sure that it is kept in sync with what is actually present). When we are going through the model and encounter a property of this loadable type, we will end up loading a class and creating a regular type model. 
for other cases, when we are working with the type elements, there's `ProcessorPojoRawTypeModel` and `ProcessorPojoPropertyModel`.  These types implement most of the API, except the annotations, as those aren't available and we are working with the annotation mirrors...

metamodel/processor/src/main/java/org/hibernate/search/metamodel/processor/annotation/processing/impl/ProcessorKeywordFieldProcessor.java
```diff
@@ -0,0 +1,61 @@
...
+		if ( !normalizer.isEmpty() ) {
+			optionsStep.normalizer( normalizer );
+		}
+		return Optional.of( optionsStep );
+	}
+
+	protected Norms getNorms(AnnotationMirror annotation) {
+		return Norms.valueOf( getAnnotationValueAsString( annotation, "norms", "DEFAULT" ) );
+	}
+}
```
marko-bekhta, line 61: In general all these processors are the same as the ones in the mapper, but they are based on the `javax.lang.model.*` types. The other difference is that they have two process methods, one for the regular use of the search annotations and the other one for "injectable fields" in the binders.

metamodel/processor/src/main/java/org/hibernate/search/metamodel/processor/annotation/processing/impl/ProcessorPropertyBindingProcessor.java
```diff
@@ -0,0 +1,90 @@
...
+			ProcessorAnnotationProcessorContext context) {
+		AnnotationMirror propertyBinder = getAnnotationProperty( annotation, "binder" );
+		if ( propertyBinder != null ) {
+			AnnotationValue type = getAnnotationValue( propertyBinder, "type" );
+			if ( type != null ) {
+				TypeMirror binder = (TypeMirror) type.getValue();
+
+				mapping.binder( new ProcessorPropertyBinder( context, binder ) );
+			}
+		}
```
marko-bekhta, line 37: at first I thought we could just continue using the programmatic mapping to build the model from the "injected" fields, but that didn't work. So instead, I've added this binder ^

metamodel/processor/src/main/java/org/hibernate/search/metamodel/processor/annotation/processing/impl/ProcessorPropertyBindingProcessor.java
```diff
@@ -0,0 +1,90 @@
...
+										bindingContext,
+										annotationMirror,
+										context,
+										member
+								) );
+					}
+				}
+			}
+
+			bindingContext.bridge( DoNothingPropertyBridge.INSTANCE );
```
marko-bekhta, line 77: the binder takes the binder type from the annotation attribute and then goes through the fields from that type, and that's where the "second" processing method for the annotations is used since, at this point, we are working with the index elements and not the programmatic api, hence the "original"  processing method won't apply here...

metamodel/processor/src/main/java/org/hibernate/search/metamodel/processor/annotation/processing/impl/ProcessorVectorFieldProcessor.java
```diff
@@ -0,0 +1,143 @@
...
+		}
+		else {
+			context.messager().printMessage( Diagnostic.Kind.ERROR, "Only float[]/byte[] fields are allowed.", element );
+			return Optional.empty();
+		}
+
+		VectorFieldTypeOptionsStep<?, ?> optionsStep = bindingContext.typeFactory().asVector( vectorType )
+				// it doesn't really matter since we won't be actually using this value anywhere,
+				// but since it's "optional" let's just keep it simple for now here:
+				.dimension( 10 );
```
marko-bekhta, line 86: this is just an interesting thing to point out, as we don't really care about some things when building this model, we can skip reading them and just hardcode some values instead.

marko-bekhta (commented):
Hey @yrodiere,
I've added the AP experiments to this branch as in the end we'd want to use the interfaces introduced here to generate that model.... as if the PR wasn't big enough already 😄, but the experiments are all in a single commit, and I've added comments to the places I think are of interest. There's a lot of boring code similar to the one we have in the annotation processors, as these just work with a different model. 

the next steps, as I see it, are:
- keep adding the missing bits (e.g. type level binders etc.). As I encounter some new code example from the docs, or just when I remember some missing part, I add it to the test and then see what has to be added to the processor to get it working. 
- write a simple class from the model. 
- take care of "various backends". I think that we'd just add all backends (lucene,elasticsearch) and have a config property for the annotation processor to let the user specify the backend to use (or even both).
- deal with the ORM mapper. I don't think that it would be feasible to make ORM somehow generate the model from `javax.lang.model.element` types. Hence, my thinking here is to add a specific processor that reads the ORM's `@Id` and treats them as `@DocumentId`; ignore the access type for now 🙈

yrodiere:
Hey. I'm finally having a look, but first some answers 

> take care of "various backends". I think that we'd just add all backends (lucene,elasticsearch) and have a config property for the annotation processor to let the user specify the backend to use (or even both).

Backends should be in the user classpath already, no? If you rely on that, you can make the annotation processor work without any configuration for the most common case (only one backend type in the classpath).

> the next steps, as I see it, are: [...]

Makes sense.

Except...

> deal with the ORM mapper. I don't think that it would be feasible to make ORM somehow generate the model from javax.lang.model.element types

As stated, I agree. But:

1. We don't need the whole ORM metamodel. Most of the things we do around the Hibernate ORM metamodel in Search are related to dependency resolution, which is pointless when generating Search's static metamodel. We would only need to infer access type, somehow, and that could reasonably be done by getting rid of [our current way of doing it](https://github.com/hibernate/hibernate-search/blob/4ec1af24e116fe79e4b206baa4d717399a87f06e/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/model/impl/HibernateOrmClassRawTypeModel.java#L96-L136) and directly applying the JPA spec ourselves instead.
2. We don't need ORM to do it, we can do it ourselves. It's annoying, but possible.
3. Perhaps Hibernate Models could help someday.

Critically, this task can even be extracted to a separate issue, and we can ask someone who knows this AccessType stuff to help with a first solution that just implements this stuff directly in Hibernate Search :)

> Hence, my thinking here is to add a specific processor that reads the ORM's @Id and treats them as @DocumentId; ignore the access type for now 🙈

That's a very reasonable temporary solution.

metamodel/processor/src/main/java/org/hibernate/search/metamodel/processor/impl/IndexedEntityMetamodelAnnotationProcessor.java
yrodiere: If it's not too much work, yes I'd start with custom code, possibly copy-pasted. It doesn't look very complex nor likely to require much maintenance, considering our needs?

metamodel/processor/src/main/java/org/hibernate/search/metamodel/processor/impl/ProcessorElementUtils.java
yrodiere: Gotta do what you gotta do :)

metamodel/processor/src/main/java/org/hibernate/search/metamodel/processor/annotation/processing/impl/ProcessorKeywordFieldProcessor.java
yrodiere: > In general all these processors are the same as the ones in the mapper, but they are based on the javax.lang.model.* types. 

So, something that would (eventually) use an abstraction instead?

I've been wondering about having an equivalent of `PojoRawTypeModel` for annotations, in particular so we can implement `org.hibernate.search.mapper.pojo.model.spi.PojoPropertyModel#annotations` even when annotations aren't actually loaded (Hibernate Models using Jandex, Processor, ...).

> The other difference is that they have two process methods, one for the regular use of the search annotations and the other one for "injectable fields" in the binders.

Seems reasonable as a temporary thing, but ultimately we might want to reuse the corresponding binder... That would allow binder/bridge composition!
It might require that we use a different type than `IndexFieldReference<T>` though. Something more like `IndexFieldHandle<T>` which could potentially become as crazy as `@GeoPointBinding IndexFieldHandle<MyPointType>`.

metamodel/processor/src/main/java/org/hibernate/search/metamodel/processor/annotation/processing/impl/ProcessorPropertyBindingProcessor.java
yrodiere: Makes sense.

metamodel/processor/src/main/java/org/hibernate/search/metamodel/processor/annotation/processing/impl/ProcessorPropertyBindingProcessor.java
yrodiere: Right. The programmatic mapping API was never intended for composition like this, so it makes sense.

Someday, to introduce composable binders, I think it would make sense to expose something like this through API:

https://github.com/hibernate/hibernate-search/blob/0b9e3f0d87250daa800f10223bcc81cdb0da773a/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/building/impl/PojoIndexModelBinder.java#L185

Then you would not declare your field directly, but instead call this method and use e.g. the default `String` binder + the corresponding field model contributor (which, unfortunately, currently [lives in the programmatic model](https://github.com/hibernate/hibernate-search/blob/0b9e3f0d87250daa800f10223bcc81cdb0da773a/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/programmatic/impl/PropertyMappingFullTextFieldOptionsStepImpl.java#L34)).

Anyway, that's a problem for another day. This is a great first implementation.

metamodel/processor/src/main/java/org/hibernate/search/metamodel/processor/annotation/processing/impl/ProcessorVectorFieldProcessor.java
yrodiere: Again fine as a temporary solution. Ultimately I'd hope we'd unify the processor implementations and just always read the value -- and trigger errors when invalid, such as `0` here.

yrodiere (commented):
No complaints, everything seems very reasonable :)

Maybe now would be the time to start merging things? It looks like we're going to go ahead with this solution, so it's not like we're going to revert it entirely. You'll just keep adding/updating code from now on, I believe?

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/SearchPredicateFactory.java
```diff
@@ -354,7 +373,7 @@ PredicateFinalStep withParameters(
 	 * @return A new predicate factory using the given object field as root.
 	 */
 	@Incubating
-	SearchPredicateFactory withRoot(String objectFieldPath);
+	SearchPredicateFactory<SR> withRoot(String objectFieldPath);
```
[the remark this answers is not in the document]
yrodiere: Uh, that was had been pending on your PR for quite some time. Sorry about that.

documentation/src/main/asciidoc/public/reference/_static-metamodel-overview.adoc
```diff
@@ -0,0 +1,68 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Red Hat Inc. and Hibernate Authors
+[[static-metamodel-overview]]
+= Overview of the static metamodel
```
marko-bekhta, line 4: @yrodiere this commit (https://github.com/hibernate/hibernate-search/pull/4156/commits/24465fae98c562209953d8fc5d2d4995ccdc575e) adds the docs 😃. only pinging you here so it'll be easier to find this change 😃 🙈

documentation/src/main/asciidoc/public/reference/_static-metamodel-processor.adoc
```diff
@@ -0,0 +1,116 @@
...
+</plugin>
+----
+
+<1> Provide the fully qualified class name of the annotation processor that generates the metamodel.
+<2> Add the `org.hibernate.search:hibernate-search-metamodel-processor` dependency as an annotation processor path for the Maven compiler plugin to be able to actually find the processor.
+<3> Add the backend dependency, in this example, the <<backend-lucene,Lucene backend>>, as an annotation processor path.
+It is important to include the same backend that the application is using to make sure that the generated metamodel classes reflect all the backend specifics.
+For example, backends might have different defaults, resulting in a different set of field traits depending on the backend.
+
+NOTE: The version of both annotation processor and backend dependencies are skipped in the definition of the annotation paths, as in case the Hibernate Search BOM is imported via dependency management, the compiler plugin will use the managed versions.
```
yrodiere, line 50:
```suggestion
NOTE: The version of both annotation processor and backend dependencies can be ommitted in the definition of the annotation paths, because they are defined in the Hibernate Search BOM, which we recommend you import via dependency management.
```

documentation/src/main/asciidoc/public/reference/_static-metamodel-processor.adoc
```diff
@@ -0,0 +1,116 @@
...
+----
+
+<1> Pass the annotation processor parameters using the `-A` key.
+<2> Pass any other compiler arguments required by the build.
+<3> Further compiler plugin configuration.
+====
+
+The following annotation processor configuration properties are available:
+
+[[static-metamodel-processor-configuration-generated_annotation-add]]`org.hibernate.search.metamodel.processor.generated_annotation.add`::
```
yrodiere, line 86: Same here, you might want to drop the `.metamodel` part?

documentation/src/main/asciidoc/public/reference/_static-metamodel-processor.adoc
```diff
@@ -0,0 +1,116 @@
...
+Description:::
+Defines whether the `@Generated` annotation includes the `date` attribute.
+Having the date attribute will result in non-reproducible builds, as the timestamp will be different for each compilation.
+Hence, it is disabled by default.
+Default value:::
+`false`
+
+[[static-metamodel-processor-configuration-add_generated_annotation]]`org.hibernate.search.metamodel.processor.backend.version`::
+Description:::
+Explicitly define the backend version. By default, the processor will use the latest compatible version of the backend.
```
yrodiere, line 102: Do you mean the Elasticsearch version?

If so, you should probably be more specific, and maybe link to the relevant section of the docs.

documentation/src/main/asciidoc/public/reference/_static-metamodel-processor.adoc
```diff
@@ -0,0 +1,116 @@
...
+This option can be used if the static metamodel is required for an older backend version.
+Default value:::
+`<latest>`
+
+[[static-metamodel-processor-limitations]]
+== Current annotation processor limitations
+
+While the annotation processor is in its early stages of development, it has a few limitations:
```
yrodiere, line 110:
```suggestion
While the annotation processor is very much functional and we encourage you to try it, it is still in the early stages of development, and thus has a few limitations:
```

documentation/src/main/asciidoc/public/reference/_static-metamodel-processor.adoc
```diff
@@ -0,0 +1,116 @@
...
+
+* Any use of <<binding-valuebridge-valuebinder,custom binders>> will be ignored and should produce a compiler warning.
+This means that if the search entities rely on custom binders, fields that those binders produce will be missing from the generated metamodel.
+* <<mapper-orm-custom-annotations,Custom mapping annotations>> are not supported.
```
yrodiere, line 114:
```suggestion
* <<mapper-orm-custom-annotations,Custom mapping annotations>> are ignored without warning.
```

documentation/src/main/asciidoc/public/reference/_static-metamodel-processor.adoc
```diff
@@ -0,0 +1,116 @@
...
+* <<mapping-programmatic,Programmatic mapping>> is also unsupported by the annotation processor.
+As the annotation processor cannot easily read programmatic mapping definitions, this limitation is here to stay.
```
yrodiere, line 116:
```suggestion
As the annotation processor cannot possibly execute programmatic mapping defined in the code that is being compiled, this limitation is here to stay.
```

documentation/src/test/java/org/hibernate/search/documentation/search/metamodel/MetamodelIT.java
```diff
@@ -0,0 +1,186 @@
...
+			assertThat( hits )
+					.extracting( Book::getId )
+					.containsExactlyInAnyOrder( BOOK1_ID, BOOK2_ID, BOOK3_ID, BOOK4_ID );
+		} );
+	}
+
+	@Test
+	void compileCheck() {
+		withinSearchSession( searchSession -> {
+			SearchScope<Book__, Book> scope = Book__.INDEX.scope( searchSession );
```
yrodiere, line 76: Not a problem here because you don't include the code in the docs, but if you do include similar code in the docs, I would encourage using `var scope`... Otherwise this feels very verbose.

documentation/src/main/asciidoc/public/reference/_static-metamodel-overview.adoc
```diff
@@ -0,0 +1,68 @@
...
+
+include::../components/_incubating-warning.adoc[]
+
+The static metamodel class describes the index structure. Each indexed entity (index) is represented by a single class
+that may contain inner classes describing the embeddables (i.e. nested/flattened objects). These classes contain <<static-metamodel-processor-field-reference-types,field references>>
```
yrodiere, line 9:
```suggestion
that may contain inner classes describing object fields (e.g. from <<TODO LINK,indexed-embedded properties>>). These classes contain <<static-metamodel-processor-field-reference-types,field references>>
```

documentation/src/test/java/org/hibernate/search/documentation/search/metamodel/Author__.java
```diff
@@ -0,0 +1,125 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.search.documentation.search.metamodel;
+
+/**
+ * Note: this class was created by an annotation processor and copied to the sources,
+ * so that we do not run the AP on an entire documentation module.
```
yrodiere, line 9: Can't you run the annotation processor on a subset of the sources?

I'm pretty sure you can add a separate execution of the maven-compiler-plugin that _only_ performs annotation processing, in which case you could possibly filter the sources to include only this package.

documentation/src/test/java/org/hibernate/search/documentation/search/metamodel/MetamodelIT.java
```diff
@@ -0,0 +1,186 @@
...
+			// tag::compileCheck-pass[]
+			List<String> titles = searchSession.search( scope )
+					.select( f -> f.field( Book__.INDEX.title ) ) // <1>
+					.where( f -> f.matchAll() )
+					.fetchHits( 20 );
```
yrodiere, line 81: Awesome :heart:

engine/src/main/java/org/hibernate/search/engine/search/aggregation/dsl/RangeAggregationFieldStep.java
```diff
@@ -56,6 +58,17 @@ default <F> RangeAggregationRangeStep<?, PDF, F> field(String fieldPath, Class<F
 	 * See {@link ValueModel}.
 	 * @return The next step.
 	 */
-	<F> RangeAggregationRangeStep<?, PDF, F> field(String fieldPath, Class<F> type, ValueModel valueModel);
+	<F> RangeAggregationRangeStep<SR, ?, PDF, F> field(String fieldPath, Class<F> type, ValueModel valueModel);
+
+	/**
+	 * Target the given field in the range aggregation.
+	 *
+	 * @param fieldReference The field reference representing a <a href="SearchAggregationFactory.html#field-paths">path</a> to the index field to aggregate.
```
yrodiere, line 66: I'm afraid linking to `#field-paths` won't be very helpful here.
Wouldn't it be better to add a section there about "static metamodel field references", with some generic info about what the static metamodel is and a recommendation to check it out in the reference documentation?

documentation/src/main/asciidoc/migration/index.adoc
```diff
@@ -117,6 +119,34 @@ to be in a more consistent format compared with other logging categories.
...
+SearchScope<Book, Book> scope = searchSession.scope( Book.class ); // <2>
+SearchScope<ReadingMaterial, ReadingMaterial> scope = searchSession.scope( List.of( Book.class, Magazine.class ) ); // <3>
+SearchScope<Book__, Book> scope = searchSession.scope( Book.class ); // <4>
+SearchScope<SomeRandomClass, Book> scope = searchSession.scope( Book.class ); // <5>
+----
+<1> Use the `?` wildcard.
+<2> Use the same type as your search entity.
+<3> Use the common supertype for a scope of multiple search entities.
+<4> Use the class generated for the static metamodel of this search entity.
+<5> Use any class.
```
yrodiere, line 148: Or... Better option, use `var`? That's what people should do when using DSL classes anyway.

yrodiere (approved):
I couldn't review it all unfortunately, but here are a few comments on the documentation.
Great stuff :)

documentation/src/main/asciidoc/public/reference/_static-metamodel-processor.adoc
marko-bekhta: It's passed to both backends, while it probably makes close to 0 difference for the Lucene one ... it still will pass it over, who knows if we will have some dependency on the index version 😃.

yrodiere:
Great to see this in! Congrats :)

PR #4160, opened 2024-05-10
HSEARCH-4542 .within() and .withinAny() for the range predicate

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4542

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/RangePredicateMatchingStep.java
+++ b/engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/RangePredicateMatchingStep.java
@@ -6,6 +6,9 @@
  */
 package org.hibernate.search.engine.search.predicate.dsl;
 
+import java.util.Arrays;
+import java.util.Collection;
+
 import org.hibernate.search.engine.search.common.ValueConvert;
 import org.hibernate.search.util.common.data.Range;
 import org.hibernate.search.util.common.data.RangeBoundInclusion;
@@ -53,7 +56,7 @@ public interface RangePredicateMatchingStep<N extends RangePredicateOptionsStep<
 	 * @return The next step.
 	 */
 	default N between(Object lowerBound, Object upperBound, ValueConvert convert) {
-		return range( Range.between( lowerBound, upperBound ), convert );
+		return within( Range.between( lowerBound, upperBound ), convert );
 	}
 
 	/**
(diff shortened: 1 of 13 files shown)


engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/RangePredicateMatchingStep.java
```diff
@@ -225,7 +227,50 @@ default N range(Range<?> range) {
...
+	 * See {@link ValueConvert} for more information.
+	 * @param convert Controls how the range bounds should be converted
+	 * before Hibernate Search attempts to interpret them as a field value.
+	 * See {@link ValueConvert} for more information.
+	 * @return The next step.
+	 */
+	N within(Range<?> range, ValueConvert convert);
+
+	/**
+	 * Require at least one of the targeted fields to be in the given range.
```
yrodiere, line 265:
```suggestion
	 * Require at least one of the targeted fields to be in any of the given ranges.
```

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/RangePredicateMatchingStep.java
```diff
@@ -225,7 +227,50 @@ default N range(Range<?> range) {
...
+	 *
+	 * @param ranges The ranges to match.
+	 * The signature of this method defines this parameter as a range with bounds of any type,
+	 * but a specific type is expected depending on the targeted field.
+	 * See {@link ValueConvert#YES} for more information.
+	 * @return The next step.
+	 */
+	N withinAny(Range<?>... ranges);
 
+	RangePredicateMatchingWithinAnyMoreStep<?> withinAny();
```
yrodiere, line 275: Unless you already found similar syntax elsewhere, I would recommend sticking to `withinAny(Range<?>...)` / `withinAny(Collection<? extends Range<?>>)` / `withinAny(Collection<? extends Range<?>>, ValueConvert)`.

See `org.hibernate.search.engine.search.predicate.dsl.TermsPredicateMatchingStep`.

If people want to use a different `ValueConvert` for each range, let them find a solution for themselves. There are limits to how flexible we should be, IMO.

yrodiere (commented):
Thanks; a few comments below.

FWIW, I think aligning on the syntax of the `terms` predicate is better than aligning on the syntax of the `range` aggregation.

I even wonder if we should change the syntax of the `range` aggregation... This `.range(....).range(...)` is quite awkward, too. Not in this PR for sure, though :)

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/RangePredicateMatchingStep.java
marko-bekhta: Thanks for taking a look at this one! 😃 👍🏻 ok will do😃

yrodiere (approved):

PR #4165, opened 2024-05-16
HSEARCH-5152 Use env variable name instead of the passphrase in GPG configuration

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5152


yrodiere:
:x true... Thanks for spotting that.

Would you mind backporting to relevant branches once it's merged (no need for PRs, just push it)? I'll see if I should remove now unnecessary code in our Develocity extension.

PR #4170, opened 2024-05-21
Update instructions to run JMH performance tests

yrodiere:
Loosely based on @franz1981's PR: https://github.com/hibernate/hibernate-orm-benchmark/pull/5/files
I tried to keep things short, at the expense of details :)


marko-bekhta (approved):
Thanks!

PR #4171, opened 2024-05-22
HSEARCH-5155 Build with JDK 21

yrodiere:
https://hibernate.atlassian.net/browse/HSEARCH-5155

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -303,7 +303,7 @@
         <version.com.puppycrawl.tools.checkstyle>10.17.0</version.com.puppycrawl.tools.checkstyle>
         <version.versions.plugin>2.16.2</version.versions.plugin>
         <version.maven-wrapper-plugin>3.3.2</version.maven-wrapper-plugin>
-        <version.impsort-maven-plugin>1.10.0</version.impsort-maven-plugin>
+        <version.impsort-maven-plugin>1.11.0</version.impsort-maven-plugin>
         <version.formatter-maven-plugin>2.24.1</version.formatter-maven-plugin>
         <version.org.eclipse.m2e.lifecycle-mapping>1.0.0</version.org.eclipse.m2e.lifecycle-mapping>
         <version.sisu-maven-plugin>0.9.0.M3</version.sisu-maven-plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -202,7 +202,7 @@
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
         <!-- JDK version required for the build; we target 11 but require at least 17 for the build -->
-        <jdk.min.version>17</jdk.min.version>
+        <jdk.min.version>21</jdk.min.version>
         <!-- The lowest supported version of Java for applications using Hibernate Search -->
         <!-- Set statically, independently from the current JDK: we want our code to comply with this version -->
         <java-version.main.release>11</java-version.main.release>
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,10 +63,10 @@ jobs:
       matrix:
         os:
           - {
-            name: "Linux JDK 17",
+            name: "Linux JDK 21",
             runs-on: 'ubuntu-latest',
             java: {
-              version: 17
+              version: 21
             },
             maven: {
               # Since we only start an Elasticsearch container on Linux we add the profile to enable container reuse here:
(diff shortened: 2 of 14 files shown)


marko-bekhta:
Looks nice 😃 thanks! 
Version in GH actions should probably be updated too 🙈

yrodiere:
> Version in GH actions should probably be updated too 🙈

Right, sorry. Done.

yrodiere:
So... turns out this won't work because impsort maven plugin needs a release... see https://github.com/revelc/impsort-maven-plugin/pull/83.
Moving to draft.

pom.xml
```diff
@@ -1012,6 +1012,8 @@
                         <staticAfter>false</staticAfter>
                         <skip>${format.skip}</skip>
                         <removeUnused>true</removeUnused>
+                        <!-- Fixes impsort for JDK 19+, see https://github.com/revelc/impsort-maven-plugin/pull/83 -->
+                        <ignoreParseErrorsBelowImports>true</ignoreParseErrorsBelowImports>
```
marko-bekhta, line 1016: @yrodiere I wanted to start a build of this PR as there's a new version of impsort published and I've merged it in, but now there's a merge conflict 🙈 🙈 🙈 sorrrrry 🙈.
I've also noticed this comment: 
https://github.com/revelc/impsort-maven-plugin/blob/384083158a635353b4700a7a3ad2243aefe432be/src/main/java/net/revelc/code/impsort/maven/plugin/AbstractImpSortMojo.java#L246-L252

we probably have to remove this property 😕 as we keep the `removeUnused` ... but it should be ok since the build now passes on newer JDKs

yrodiere: Rebased and removed `ignoreParseErrorsBelowImports`, thanks.

yrodiere:
Draft again... https://github.com/revelc/formatter-maven-plugin/issues/881

yrodiere:
I added an upgrade to formatter-maven-plugin 2.24.0, which supposedly solves the Java 21 compatibility problem... but then it has [other problems](https://github.com/quarkusio/search.quarkus.io/pull/271#issuecomment-2144431538), so we'll probably have to wait for 2.24.1 or 2.25.0 :(

yrodiere:
> I added an upgrade to formatter-maven-plugin 2.24.0, which supposedly solves the Java 21 compatibility problem... but then it has [other problems](https://github.com/quarkusio/search.quarkus.io/pull/271#issuecomment-2144431538), so we'll probably have to wait for 2.24.1 or 2.25.0 :(

Alright, 2.24.1 is out, I updated the PR... and...

```
[ERROR] Failed to execute goal net.revelc.code:impsort-maven-plugin:1.10.0:sort (import-sorting) on project hibernate-search-parent-integrationtest: Execution import-sorting of goal net.revelc.code:impsort-maven-plugin:1.10.0:sort failed: No enum constant com.github.javaparser.ParserConfiguration.LanguageLevel.JAVA_21 -> [Help 1]
```

:facepalm: 

https://github.com/revelc/impsort-maven-plugin/issues/101#issuecomment-2140991247

:cry:

yrodiere:
Added an upgrade to a version of impsort that supports JDK 21, and rebased.

:crossed_fingers:

marko-bekhta:
😖 😕 there seems to be some precision loss with JDK 21 switch... `RangeAggregationSpecificsIT` fails in this PR (but only on OpenSearch) as the upper bound on JDK 17 `1.58451456E9` and on JDK 21 `1.5845146E9`

yrodiere:
This is very odd... If this is related to the jdk, I'd expect this to affect Elasticsearch as well... ? Or maybe there's some opensearch-specific testing or Jason handling at play?

Did you pinpoint where the precision loss occurs exactly? Compilation of a constant, or Jason parsing/formatting, or... ?

marko-bekhta:
yeah ... 
looking into this right now 🙈 that info I posted yesterday was by looking at the generated query string. I'll post here once I find something.

marko-bekhta:
https://bugs.openjdk.org/browse/JDK-8202555
https://bugs.openjdk.org/browse/JDK-8291475

^ looks like that's the reason 😕 I've tested with JDK 17 (ok) 18 (ok) 19 (not ok) and 21 (not ok) 😄
This "loss" happens inside of the `JsonPrimitive` when the query is sent.
OpenSearch has a JDK 21 bundled while Elasticsearch has JDK 22. I've tried starting OpenSearch with JDK 22 just to see if that would make a difference, but no. 
I also tried running this test on OpenSearch 2.15 - 2.1 and got the same result on all.

I guess let's just update the test value and move on? ( `1584514514.000000184f` -> `1584514414.000000184f` )

yrodiere:
> I guess let's just update the test value and move on? ( `1584514514.000000184f` -> `1584514414.000000184f` )

+1 if it's a jdk bug, and it's already been reported, not much we can do.

marko-bekhta (approved):

marko-bekhta:
The build passed! thanks 😃!

PR #4172, opened 2024-05-23
HSEARCH-4930 Use dropAndCreateSchemaOnStart by default on Amazon OpenSearch Service for MassIndexer and JSR-352 integration

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4930

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/execution/impl/ElasticsearchIndexWorkspace.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/execution/impl/ElasticsearchIndexWorkspace.java
@@ -14,9 +14,11 @@ import org.hibernate.search.backend.elasticsearch.multitenancy.impl.MultiTenancy
 import org.hibernate.search.backend.elasticsearch.orchestration.impl.ElasticsearchParallelWorkOrchestrator;
 import org.hibernate.search.backend.elasticsearch.util.spi.URLEncodedString;
 import org.hibernate.search.backend.elasticsearch.work.factory.impl.ElasticsearchWorkFactory;
+import org.hibernate.search.engine.backend.schema.management.spi.IndexSchemaManager;
 import org.hibernate.search.engine.backend.work.execution.OperationSubmitter;
 import org.hibernate.search.engine.backend.work.execution.spi.IndexWorkspace;
 import org.hibernate.search.engine.backend.work.execution.spi.UnsupportedOperationBehavior;
+import org.hibernate.search.util.common.function.TriFunction;
 
 import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
@@ -28,6 +30,8 @@ public class ElasticsearchIndexWorkspace implements IndexWorkspace {
 	private final ElasticsearchParallelWorkOrchestrator orchestrator;
 	private final URLEncodedString indexName;
 	private final Set<String> tenantIds;
+	private final TriFunction<OperationSubmitter, UnsupportedOperationBehavior, Boolean, CompletableFuture<?>> purgeOrDrop;
+	private final IndexSchemaManager indexSchemaManager;
 
 	public ElasticsearchIndexWorkspace(ElasticsearchWorkFactory workFactory,
 			MultiTenancyStrategy multiTenancyStrategy, ElasticsearchParallelWorkOrchestrator orchestrator,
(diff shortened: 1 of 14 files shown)


engine/src/main/java/org/hibernate/search/engine/backend/work/execution/spi/IndexWorkspace.java
```diff
@@ -61,4 +61,22 @@ CompletableFuture<?> flush(OperationSubmitter operationSubmitter,
...
+	/**
+	 * Depending on which operation is available for the underlying backend either purge or schema drop/create will be performed to clear out the index.
+	 *
+	 * @param operationSubmitter The behavior to adopt when submitting the operation to a full queue/executor.
+	 * @param unsupportedOperationBehavior The behavior to adopt if the operation is not supported in this index.
+	 * @param mergeSegmentsAfterPurge Defines whether to {@link #mergeSegments(OperationSubmitter, UnsupportedOperationBehavior) merge the segments}
+	 * if the purge was executed.
+	 * @return A completion stage for the executed operation, or a completed stage if the operation is not supported.
+	 */
+	default CompletableFuture<?> purgeOrDrop(OperationSubmitter operationSubmitter,
```
yrodiere, line 73: Putting the logic that far down looks a bit dodgy to me: if a mass indexer or pojo workspace targets multiple indexes from different backends, we could end up doing a purge in some indexes and a drop-and-create in others. Do we really want that?

On the other hand, automatically applying the default to all indexes as soon as there's just *one* Amazon OpenSearch Service index involved feels dodgy as well, as behavior for a given scope could change just by adding another index to the scope.

Maybe a better approach would be to have a global (mapper-scoped) setting instead, have it default to "purge" in general, but throw an exception on startup if there's an Amazon OpenSearch Service set up, asking people to set it explicitly to `drop-and-create`, and plan to change the default in the future to `drop-and-create`? It seems to me that `drop-and-create` is a much better default in general anyway, even on Lucene.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/execution/impl/ElasticsearchIndexWorkspace.java
```diff
@@ -38,6 +42,29 @@ public ElasticsearchIndexWorkspace(ElasticsearchWorkFactory workFactory,
 		this.orchestrator = orchestrator;
 		this.indexName = indexManagerContext.getElasticsearchIndexWriteName();
 		this.tenantIds = tenantIds;
+		this.indexSchemaManager = indexManagerContext.schemaManager();
+		this.purgeOrDrop = workFactory.isPurgeSupported() ? this::purgeAndMerge : this::dropAndCreate;
```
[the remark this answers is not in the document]
marko-bekhta: yeah, I probably should've just used the `isDeleteByQuerySupported`. As to moving the logic into..., it drops and creates the schema, so it depends on the schema manager, which itself depends on the workFactory. Or maybe I'm missing something here🙈

engine/src/main/java/org/hibernate/search/engine/backend/work/execution/spi/IndexWorkspace.java
marko-bekhta: yeah, that's what I was thinking... since we can only know what can be applied to indexes within the scope, by accessing indexes themselves, I thought that it would make sense to let them deal with it and pick the operation they can perform. 

> Maybe a better approach would be to have a global (mapper-scoped) setting instead

If you think that `drop-and-create` is a better option moving forward, then maybe let's just switch to it in 7.2 and add a deprecated config property to allow users to switch to the `purge` option, just in case they want an old behavior... since they'd need to either accept the new logic eventually or make an explicit choice when creating the indexer...

With that said, using `drop-and-create` can be a problem when multitenancy is involved. There is a check in  Jakarta Batch that would prevent calling `drop-and-create` when a tenant id is set, since it'll drop all, but index just one tenant...

yrodiere: > If you think that `drop-and-create` is a better option moving forward, then maybe let's just switch to it in 7.2 and add a deprecated config property to allow users to switch to the `purge` option, just in case they want an old behavior... But since they'd need to either accept the new logic eventually or make an explicit choice when creating the indexer... do we even need that property?

Changing default behavior is a backwards-incompatible change, it shouldn't happen in minors. The setting would provide a migration path, and we could switch the default in 8.0 (while retaining the setting until some later time).

> With that said, using `drop-and-create` can be a problem when multitenancy is involved.

Fair point. I suppose we could have the default be `purge` only when multitenancy is enabled, though? Note this also means you can't easily use multitenancy on Amazon OpenSearch Serverless, at least not how it's implemented right now...

Anyway... in light of this, I agree it's less clear what's the best solution. I'll approve the PR and let you decide what's best: merge your solution and forget about the "global config" idea , or work on the "global config" idea right now, or merge your solution and plan the "global config" idea for later.

Longer term, we'd probably want to make `drop-and-create` work for multitenancy... I can imagine an "index-per-tenant" multitenancy strategy be a solution, as then drop-and-create *would* be doable per tenant. If we made that "index-per-tenant" multitenancy strategy the default, then `drop-and-create` would itself be a justifiable default.
But that strategy doesn't exist, and would likely involve quite a few changes, and a few challenges: with Elasticsearch we can probably get by with index templates and aliases, but how would be know what the current indexes are on startup with Lucene?

>  There is a check in Jakarta Batch that would prevent calling `drop-and-create` when a tenant id is set, since it'll drop all, but index just one tenant...

Can you please reassure me and tell me this check is in the mass indexer as well? :see_no_evil:

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/work/execution/impl/ElasticsearchIndexWorkspace.java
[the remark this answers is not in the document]
yrodiere: Ah, right. So that's why :)

yrodiere (approved):

engine/src/main/java/org/hibernate/search/engine/backend/work/execution/spi/IndexWorkspace.java
marko-bekhta: Thanks, ok 😃 will do the global setting then and as for ... 

> Can you please reassure me and tell me this check is in the mass indexer as well? 🙈

I will do so a bit later once I'll add it 🙈 🤐

marko-bekhta:
closing in favor of https://github.com/hibernate/hibernate-search/pull/4177

PR #4181, opened 2024-06-04
HSEARCH-5162 Make ORM mass indexer work when JPA compliance is enabled

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5162

### DIFF
--- a/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/loading/impl/ConditionalExpressionQueryFactory.java
+++ b/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/loading/impl/ConditionalExpressionQueryFactory.java
@@ -11,7 +11,7 @@ import java.util.List;
 import java.util.Set;
 
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
-import org.hibernate.metamodel.mapping.EntityMappingType;
+import org.hibernate.metamodel.model.domain.EntityDomainType;
 import org.hibernate.query.Query;
 import org.hibernate.search.mapper.orm.loading.spi.ConditionalExpression;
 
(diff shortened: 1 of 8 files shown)


mapper/orm/src/main/java/org/hibernate/search/mapper/orm/loading/impl/ConditionalExpressionQueryFactory.java
```diff
@@ -35,22 +35,22 @@ public final boolean uniquePropertyIsTheEntityId() {
 	}
 
 	@Override
-	public Query<Long> createQueryForCount(SharedSessionContractImplementor session, EntityMappingType entityMappingType,
+	public Query<Long> createQueryForCount(SharedSessionContractImplementor session, EntityDomainType<?> entityDomainType,
```
yrodiere, line 38: I'm surprised tests are passing... I'd expect you need to change more than this, since there are other places where we insert entity names in HQL queries.

There's `HqlTypeQueryFactory` in particular, but there may be other places.

marko-bekhta: yeah, I wasn't sure if that's enough, but `HqlTypeQueryFactory` is only used for dynamic entities so the names there are the same, no? 
btw, locally I've updated the ORM helper to always enable the JPA compliance and there were only a couple of tests that failed

yrodiere (approved):
LGTM, though I'd recommend testing dynamic-map entities with JPA compliance enabled.

yrodiere (approved):

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/loading/impl/ConditionalExpressionQueryFactory.java
marko-bekhta: I double-checked these dynamic map entities, and from what I see, the JPA and ORM names are the same for them and are the ones defined in the XML. I've updated a few existing tests for dynamic entities to be parameterized and test the compliance, so if something changes, we'll notice. 

Thanks for taking a look at this one!

PR #4183, opened 2024-06-05
HSEARCH-5156 Address Spring Boot 3.3 upgrade nested JARs test failures

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5156

Our integration test with the real app in `hibernate-search-integrationtest-spring-repackaged-application` works fine with the `3.3.0`. There were a few changes in Boot around nested jars and all related. I haven't pinpointed the exact change that caused the issue with the `CodeSourceTest` test, but.... I think it wasn't entirely correct here. `jar_jarScheme_springBoot_classesInSubDirectory` tests a sub-directory, so there is no need to work with a nested jar since it's just a dir within a root jar and not another jar ...

### DIFF
--- a/build/parents/build/pom.xml
+++ b/build/parents/build/pom.xml
@@ -177,7 +177,7 @@
 
         <!-- >>> Spring integration tests -->
         <!-- When updating Spring Boot check if it supports JDK23 and enable testing with it if so (see testWithJdk23 profile). -->
-        <version.org.springframework.boot>3.2.0</version.org.springframework.boot>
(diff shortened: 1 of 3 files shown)


yrodiere:
Does it still work with Spring 3.2? If so LGTM.

marko-bekhta:
> Does it still work with Spring Boot 3.2? If so LGTM.

yes 👍🏻 😃 
Thanks for taking a look!

PR #4193, opened 2024-06-11
DONOTMERGE: removing 1-1 associations and setting values to null

marko-bekhta:
Hey @yrodiere,

I was looking more into https://hibernate.atlassian.net/browse/HHH-18212 and how it affects Search ...

if the user sets both sides to nulls (so that there's no TransientObjectException) we will end up with just a delete event sent to our listener... this delete event will have the entity with the null for the other side of the relation, so we won't be able to identify which entities have to be re-indexed ... if I add this bit to the ondelete handling:
```java
if ( dirtyAssociationPaths != null ) {
	plan.updateAssociationInverseSide( dirtyAssociationPaths, null, event.getDeletedState() );
}
```
Then things would work even if the user set both sides to `null` (which we probably should do regardless of what we end up doing for HHH-18212). But we'll get the same NPE on 6.6 again ... just coming from a different starting point (`event.getDeletedState()`)

cc @mbladel

### DIFF
--- a/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/association/bytype/AbstractAutomaticIndexingAssociationBaseIT.java
+++ b/integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/association/bytype/AbstractAutomaticIndexingAssociationBaseIT.java
@@ -1153,6 +1153,8 @@ public abstract class AbstractAutomaticIndexingAssociationBaseIT<
 			// and the deletion even will simply be ignored.
 			containedAssociation.get( containedEntity );
 
+			containedAssociation.set( containedEntity, null );
+			containingAssociation.set( entity1, null );
 			session.remove( containedEntity );
 
 			backendMock.expectWorks( _indexed().indexName() )
@@ -3764,8 +3766,9 @@ public abstract class AbstractAutomaticIndexingAssociationBaseIT<
 			// But DO force loading on contained side:
 			// if the association is lazy and unloaded, it won't be loadable after the deletion
 			// and the deletion even will simply be ignored.
-			containedAssociation.get( containedEntity );
 
+			containedAssociation.set( containedEntity, null );
+			containingAssociation.set( containing, null );
 			session.remove( containedEntity );
 
 			backendMock.expectWorks( _indexed().indexName() )
--- a/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/event/impl/HibernateSearchEventListener.java
+++ b/mapper/orm/src/main/java/org/hibernate/search/mapper/orm/event/impl/HibernateSearchEventListener.java
@@ -106,6 +106,12 @@ public final class HibernateSearchEventListener
 
 		Object providedId = typeContext.toIndexingPlanProvidedId( event.getId() );
 		plan.delete( providedId, null, entity );
+
+		BitSet dirtyAssociationPaths = typeContext.dirtyContainingAssociationFilter().all();
+
+		if ( dirtyAssociationPaths != null ) {
+			plan.updateAssociationInverseSide( dirtyAssociationPaths, null, event.getDeletedState() );
+		}
 	}
 
 	@Override


integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/association/bytype/AbstractAutomaticIndexingAssociationBaseIT.java
```diff
@@ -1153,6 +1153,8 @@ void directImplicitAssociationUpdateThroughDeleteWithAlreadyLoadedAssociation_in
 			// and the deletion even will simply be ignored.
 			containedAssociation.get( containedEntity );
 
+			containedAssociation.set( containedEntity, null );
+			containingAssociation.set( entity1, null );
```
yrodiere, line 1157: If you look at the comment above, this specific test only makes sense if you don't update associations.

Don't we have a test that does what you're doing here already?

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/association/bytype/AbstractAutomaticIndexingAssociationBaseIT.java
```diff
@@ -3764,8 +3766,9 @@ void indirectImplicitAssociationUpdateThroughDeleteWithAlreadyLoadedAssociation_
 			// But DO force loading on contained side:
 			// if the association is lazy and unloaded, it won't be loadable after the deletion
 			// and the deletion even will simply be ignored.
-			containedAssociation.get( containedEntity );
 
+			containedAssociation.set( containedEntity, null );
+			containingAssociation.set( containing, null );
```
yrodiere, line 3771: Same here, the test apparently only makes sense if you don't update associations.

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/event/impl/HibernateSearchEventListener.java
```diff
@@ -106,6 +106,12 @@ public void onPostDelete(PostDeleteEvent event) {
 
 		Object providedId = typeContext.toIndexingPlanProvidedId( event.getId() );
 		plan.delete( providedId, null, entity );
+
+		BitSet dirtyAssociationPaths = typeContext.dirtyContainingAssociationFilter().all();
+
+		if ( dirtyAssociationPaths != null ) {
+			plan.updateAssociationInverseSide( dirtyAssociationPaths, null, event.getDeletedState() );
+		}
```
yrodiere, line 114: I'm not sure of the implications of this... I'll try to have a closer look tomorrow.

yrodiere (commented):
From the changes to the test, I suspect this essentially removes a feature... but let me have another look tomorrow.

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/association/bytype/AbstractAutomaticIndexingAssociationBaseIT.java
marko-bekhta: Yeah, I've seen the comments... But if I understood correctly, that's what we (or users) would have to do with the ORM 6.6 upgrade and that `TransientObjectException`. Or have I missed something?

(I've just used these tests as they were failing on the upgrade branch)

yrodiere: So the answer to this is "no"?

> Don't we have a test that does what you're doing here already?

marko-bekhta: > So the answer to this is "no"?

I'd say yes, it's "no" 😃. I haven't looked through all the tests we have, but since without the change to the listener, all tests (when not modified) in the orm-mapper test module were passing, I'd assume we don't test this particular combination... and TBH... it does feel a bit unnatural to set nulls when I'm just trying to remove an entity, especially the one that owns the association.

yrodiere: OK.

> TBH... it does feel a bit unnatural to set nulls when I'm just trying to remove an entity, especially the one that owns the association.

Yeah I suspect this additional check isn't here to stay, it'll just make everyone's life miserable. I predict we'll have to roll back the change in ORM, or at least enable it only when JPA compliance is enabled. But we'll see.

So I'd suggest:

* Keeping the pre-existing test as-is.
* Adding a test which sets the unowned association to `null`, copy-pasted from the pre-existing one but with a proper method name (it's not really an "implicit" change anymore) and updated comments.
   BTW from what I understand you only need to set the association to `null` for the side that points to a deleted entity. No need to set both sides to `null`.
* In a separate commit, add an `@Disabled` annotation on the pre-existing test, which we'll keep as long as Hibernate ORM doesn't support it. Please link to the appropriate Jira :)

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/event/impl/HibernateSearchEventListener.java
yrodiere: Okay so this was copy-pasted from `onPostInsert`/`onPostUpdate`...
It should be at least as safe as it is there, then.

I wonder what happens when the target of the assocation in `deletedState` has also been deleted, but to be fair if that's a problem, it's probably a problem in the `oldState` of `onPostUpdate` too. So, out of scope here.

I'd simply suggest you add a comment just like we did on  `onPostInsert`/`onPostUpdate` :)

integrationtest/mapper/orm/src/test/java/org/hibernate/search/integrationtest/mapper/orm/automaticindexing/association/bytype/AbstractAutomaticIndexingAssociationBaseIT.java
mbladel: Hey guys, chiming in here with just a couple notes:

> Yeah I suspect this additional check isn't here to stay, it'll just make everyone's life miserable

This is what we discussed last week, and IIRC @yrodiere you mentioned that forgetting to set associations to `null` when their targets have been removed from the database _will_ cause trouble in Search too with regards to indices (?). It will impact some behaviors of ORM for sure (e.g. caching). The check is probably here to stay, we just need to find a reasonable way of reintroducing it (may be in 7.0 instead of 6.6) so that impacted users have the chance to adjust.

> from what I understand you only need to set the association to null for the side that points to a deleted entity. No need to set both sides to null

This is correct. There shouldn't be any references to the deleted entity instance, but on the other side (sice it's already been deleted) no need to do anything more.

yrodiere: > This is what we discussed last week, and IIRC @yrodiere you mentioned that forgetting to set associations to `null` when their targets have been removed from the database _will_ cause trouble in Search too with regards to indices (?).

I said something like that, and indeed, as far as entities in Search are concerned, the check does help.

For entities that are not indexed/cached, well... It really imposes constraints for seemingly no reason, which is the kind of think that usually causes an outcry.

And that's basically what we discussed in that meeting as well: let's try to introduce the check, and reconsider if there's too much backlash. Which is something I agree with.

I just think the backlash is very likely to happen :)

> > from what I understand you only need to set the association to null for the side that points to a deleted entity. No need to set both sides to null
> 
> This is correct. There shouldn't be any references to the deleted entity instance, but on the other side (sice it's already been deleted) no need to do anything more.

:+1:

marko-bekhta:
opened https://github.com/hibernate/hibernate-search/pull/4195 to add the tests

PR #4206, opened 2024-06-24
Update list of JDK bundles in the forbidden APIs plugin

marko-bekhta:
The new bundle files didn't result in any new reports, except for the two URL constructors that we've already identified and planned to update when possible.

I've moved the constructors to small methods so that we can suppress the warnings just on them and leave everything else still scanned.

Otherwise I guess it's ok to drop the JDKs that are lower than the min one and add the others that are now available.


yrodiere:
I think there is a conflict with my other PR about using JDK 21 for the build?

PR #4214, opened 2024-06-28
HSEARCH-5187 Deprecate `ValueConvert` in favor of more a more descriptive enum with more consistent handling

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5187

----------------------

----------------------


mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/FieldProjection.java
```diff
@@ -41,9 +41,9 @@
 
 	/**
 	 * @return A value controlling how the data fetched from the backend should be converted.
-	 * @see ValueConvert
-	 * @see SearchProjectionFactory#field(String, Class, ValueConvert)
+	 * @see ProjectionValueConvert
+	 * @see SearchProjectionFactory#field(String, Class, ProjectionValueConvert)
 	 */
-	ValueConvert convert() default ValueConvert.YES;
+	ProjectionValueConvert convert() default ProjectionValueConvert.MAPPING;
```
marko-bekhta, line 47: I'm not sure that there's a nice way to "deprecated" things for this annotation 😕

marko-bekhta (commented):
Hey @yrodiere 👋🏻 😃 

I've ended up with two SPI enums:
- `InputValueConvert`: what are we going to do with the user input
- `OutputValueConvert`: what are we going to do with the backend output values

and as for the user-facing enums, I thought we can go with an enum per DSL-"thing":
- `AggregationInputValueConvert` 
- `AggregationOutputValueConvert`
- `PredicateValueConvert` (maybe `PredicateInputValueConvert` to be consistent?)
- `ProjectionValueConvert`  (maybe `ProjectionInputValueConvert` to be consistent?)
- `SortValueConvert`  (maybe `SortInputValueConvert` to be consistent?)

I've ended up having two enums for aggregations since the user can use one to pass range values to us and the other to tell us what output is expected ... 

Then since we've added `ValueConvert.PARSE` in https://hibernate.atlassian.net/browse/HSEARCH-5109 as part of 7.2.0.Alpha1 I wonder if it would be ok to just drop that value from the enum in favor of these new enums ..🙈 

Does such granularity make sense to you?

yrodiere (commented):
Thanks for having a look.

I have a few problems with this projection/predicate/sort/aggregation + input/output + api/spi split:

1. It's a bit verbose and inconvenient (many different enums) -- and seemingly not necessary, since a few of the enums are similar (Predicate/Sort/AggInput, AggOutput/Projection).
2. I'm not sure how future-proof it will be. We might need other conversions... For example you're currently using the `AggregationOutputConvert` for aggregation buckets (keys) -- will you want the same enum for metrics (values: sum, ...)?

Naming is also a bit awkward: `ValueConvert` was bad to begin with (I would know, I picked it :) )

Maybe let's get back to the drawing board on Zulip... I may be missing something :shrug:

engine/src/main/java/org/hibernate/search/engine/search/aggregation/dsl/RangeAggregationFieldStep.java
```diff
@@ -35,9 +35,27 @@ default <F> RangeAggregationRangeStep<?, PDF, F> field(String fieldPath, Class<F
...
+	 * Target the given field in the range aggregation.
+	 *
+	 * @param fieldPath The <a href="SearchAggregationFactory.html#field-paths">path</a> to the index field to aggregate.
+	 * @param type The type of field values.
+	 * @param <F> The type of field values.
+	 * @param convert Controls how the ranges passed to the next steps and fetched from the backend should be converted.
+	 * See {@link ValueModel}.
+	 * @return The next step.
+	 */
+	<F> RangeAggregationRangeStep<?, PDF, F> field(String fieldPath, Class<F> type, ValueModel convert);
```
yrodiere, line 59:
```suggestion
	 * @param valueModel The model of range values, used to determine how range values passed to the next steps and fetched from the backend should be converted.
	 * See {@link ValueModel}.
	 * @return The next step.
	 */
	<F> RangeAggregationRangeStep<?, PDF, F> field(String fieldPath, Class<F> type, ValueModel valueModel);
```

This is probably something to fix in other places too.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/aggregation/SingleFieldAggregationTypeCheckingAndConversionIT.java
```diff
@@ -492,33 +492,11 @@ void withConverter_conversionDisabled(SupportedSingleFieldAggregationExpectation
...
-		}
-		else {
-			assertThatThrownBy( () -> scenario.setupWithConverterSetting( mainIndex.createScope().aggregation(), fieldPath,
-					ValueConvert.PARSE ) )
-					.isInstanceOf( SearchException.class )
-					.hasMessageContainingAll(
-							"Cannot use ValueConvert.PARSE as a converter. Use one of the allowed values [YES, NO] instead" );
-		}
-	}
-
```
yrodiere, line 521: I get that this shouldn't fail anymore, but there's still behavior to be tested, isn't there?

engine/src/main/java/org/hibernate/search/engine/cfg/spi/ParseUtils.java
```diff
@@ -400,4 +400,89 @@ private static String[] arrayValues(String value) {
...
+		return ISO_YEAR.format( value );
+	}
+
+	public static String format(YearMonth value) {
+		return ISO_YEAR_MONTH.format( value );
+	}
+
+	public static String format(ZonedDateTime value) {
+		return DateTimeFormatter.ISO_ZONED_DATE_TIME.format( value );
+	}
```
yrodiere, line 487: Shouldn't this be in `FormatUtils`? Maybe not, IDK.

In any case, this probably warrants some tests, e.g. in `org.hibernate.search.engine.cfg.spi.ParseUtilsTest`.

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/ProjectionConstructorFieldProjectionIT.java
```diff
@@ -134,8 +134,10 @@ class IndexedEntity {
 		class MyProjection {
 			public final String myEnum;
 
+			@SuppressWarnings("deprecation")
```
yrodiere, line 137: ... why? You're not using `ValueConvert` anymore, are you?

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/ProjectionConstructorFieldProjectionIT.java
```diff
@@ -134,8 +134,10 @@ class IndexedEntity {
...
 			@ProjectionConstructor
-			public MyProjection(@FieldProjection(convert = ValueConvert.NO) String myEnum) {
+			public MyProjection(
+					@FieldProjection(valueModel = ValueModel.INDEX) String myEnum) {
```
yrodiere, line 140: That's nice but shouldn't `ValueConvert` be tested as well?

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/aggregation/SingleFieldAggregationTypeCheckingAndConversionIT.java
```diff
@@ -313,7 +313,7 @@ void invalidFieldType_conversionDisabled(SupportedSingleFieldAggregationExpectat
 		AggregationScenario<?> scenario = expectations.withFieldType( TypeAssertionHelper.wrongType( wrongType ) );
 
 		assertThatThrownBy( () -> scenario.setupWithConverterSetting(
-				mainIndex.createScope().aggregation(), fieldPath, ValueConvert.NO
+				mainIndex.createScope().aggregation(), fieldPath, ValueModel.INDEX
```
yrodiere, line 316: That's nice but shouldn't `ValueConvert` be tested as well?

documentation/src/main/asciidoc/public/reference/_search-dsl-aggregation.adoc
```diff
@@ -318,23 +318,6 @@ include::{sourcedir}/org/hibernate/search/documentation/search/aggregation/Aggre
...
-alternatively see how it can be <<binding-valuebridge-parse,customized with bridges>>.
-
-.Using string values to create ranges
-====
-[source, JAVA, indent=0, subs="+callouts"]
-----
-include::{sourcedir}/org/hibernate/search/documentation/search/aggregation/AggregationDslIT.java[tags=range-parse]
-----
-====
-
```
yrodiere, line 337: I suspect there's a tiny bit more documentation to update :grin:

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/ProjectionConstructorFieldProjectionIT.java
marko-bekhta: that's related to your comment above: https://github.com/hibernate/hibernate-search/pull/4214#discussion_r1660706416

I've removed the value convert so that's why the test looks like that.... 

Now, going back to that comment... I've started with `@FieldProjection` having two attributes, one existing for value convert and the other for a new enum. The problem with that is handling two values... I had some code in the processor that would check that both value convert and new enum values are not conflicting. But that means that the user has to set both values at the same time... because if the user changes one value but keep the other default ... what should we pick... so IDK but it seems that just removing the attribute makes it cleaner... an alternative I could think of would be to create a new `FieldProjection` annotation and deprecate existing one, but then how to call the new annotation ... 😖😖😖😖😖😖😖😖😖 😕

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/aggregation/SingleFieldAggregationTypeCheckingAndConversionIT.java
marko-bekhta: Thanks for taking a look! 🎉 
Yeah, we'd need to add more tests for `RAW` and `STRING` now that they should work everywhere 🙈
For now, I was just trying to make it work and see if what we'll get would be something we'll keep, as I wasn't convinced with a single enum 😄

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/ProjectionConstructorFieldProjectionIT.java
yrodiere: Maybe do what you did in your first attempt, but ignore `convert` if it's set to the default `YES`? Not great, but at least it sort of works and detects most misuses -- with the remaining ones having a warning.

engine/src/main/java/org/hibernate/search/engine/cfg/spi/NormalizeUtils.java
```diff
@@ -0,0 +1,43 @@
...
+					.toOffsetTime();
+		}
+
+		if ( encoded > ALMOST_WHOLE_DAY ) {
+			return instant.atOffset( ZoneOffset.ofHours( -instant.atOffset( ZoneOffset.UTC ).getHour() - 1 ) )
+					.toOffsetTime();
+		}
+
+		return instant.atOffset( ZoneOffset.UTC ).toOffsetTime();
+	}
```
marko-bekhta, line 35: hey @yrodiere  👋🏻 🙈 

I wanted to ask you what you'd think about such "timzeone-offset-magic" 🪄🪄🪄... 

All this is to make using `RAW` on input work fine... we are decoding and then encoding the values through a codec. If a user passes us a long-as-offset-time we'll end up losing the fact that e.g. the time is in the previous/next day... e.g. 

Let's assume we have a `17-00+18`, this would encode to -1hour (in nanos). Decoding that will result in `23:01Z`. 

This means if we had a range of `( 17-00+18 ; 19-00+18 )`, which seems legit, if such range goes through encode decode, currently it'll result in `( 23:01Z  ; 01:01Z)` and here if we encode the lower and upper values again, we'll have `lower > upper` leading to an exception (`82860000000000 ; 3660000000000`). With the change ... we'll get `(00:01+01:00 ; 01:01Z)` or, if encoded (`-3540000000000 ; 3660000000000`), preserving the "order". 

The alternative is to bypass the dsl converters entirely, but that means we'll have something like:
```java
if (valueModel == RAW) {
    if (codec.getEncodedType() != compatible with requested type) {
        throw log.someExcetpion();
    }

    return value;
} else {
    var converter = getDslConverter();

    codec.encode(converter.covert(value)
}
```
and I've tried that & but didn't like it 🙈 😕... + that won't work fine with ES backend as there we didn't want to expose JSON types to the user ...

yrodiere: > The alternative is to bypass the dsl converters entirely

That's actually what I'd expect. I mean, raw is raw. No conversion, pass what the backend needs directly.

> and I've tried that & but didn't like it 🙈 😕... 

Can't argue about feelings :shrug: 

> + that won't work fine with ES backend as there we didn't want to expose JSON types to the user ...

What about expecting whatever we feed into JSON types? E.g. expect a String if we're using `new JsonPrimitive(<String>)`.

Yes it's brutally simple and potentially brittle, but then again, it _is_ called "raw".

yrodiere: > > * that won't work fine with ES backend as there we didn't want to expose JSON types to the user ...
> 
> What about expecting whatever we feed into JSON types? E.g. expect a String if we're using `new JsonPrimitive(<String>)`.
> 
> Yes it's brutally simple and potentially brittle, but then again, it _is_ called "raw".

Or you could expect a JSON string and parse it, if that makes life easier for you (e.g. for native field types).

marko-bekhta: > Or you could expect a JSON string and parse it, if that makes life easier for you (e.g. for native field types).

That's what I've essentially done for now (at least I think so 😄). I'm using a parser converter to parse a string that the user passed through the DSL to us. The parser should be able to parse strings in the format that ES expects, so that seems to work okay.

marko-bekhta: > Yes it's brutally simple and potentially brittle, but then again, it is called "raw".
> Can't argue about feelings 🤷

ok, you've convinced me 😃. I'll give it another try. There was a nice rainstorm over the weekend and maybe it changed my feelings towards it 😄😄😄

marko-bekhta: I've been looking at this ... and it seems that to make it work and to keep it clean, we'd need to move the codec encode/decode into dsl/projection converter ... does that sound like a good idea to you?

marko-bekhta: > to keep it clean, we'd need to move the codec encode/decode into dsl/projection converter ...

ok, so this idea didn't work out 😖😖😖 elasticsearch codecs have multiple encode functions for various parts so having encode/decode packed inside the dsl/projection converter. I'll try a straightforward approach to just check things in place. I'll ping you once I have something workable.

engine/src/main/java/org/hibernate/search/engine/cfg/spi/ParseUtils.java
marko-bekhta: added `org.hibernate.search.engine.cfg.spi.FormatUtils`

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/ProjectionConstructorFieldProjectionIT.java
marko-bekhta: Ok, so I've added back the convert attribute, but it still looks a bit useless to me there...
If the convert value is not a default one i.e. `convert != YES` it means that there's a conflict between `convert` and a default `modelValue` (which is `MAPPING`). If the model value is set to non-default `INDEX` to match `convert=NO` then it means that the user went in and made a change to the annotation ... so to me, it seems if you are going to apply the change anyway, then why not drop the deprecated attribute and use the new one... and if the user does that then it is the same as if we just drop the `convert` attribute entirely... 
Logging a warning instead of throwing an exception, in this case, seems like something that would lead to problems at runtime.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/impl/ElasticsearchIndexValueFieldType.java
```diff
@@ -29,13 +34,19 @@ public class ElasticsearchIndexValueFieldType<F>
...
 
 	public ElasticsearchIndexValueFieldType(Builder<F> builder) {
 		super( builder );
 		this.elasticsearchTypeAsJson = builder.elasticsearchTypeAsJson();
 		this.codec = builder.codec;
 		this.mapping = builder.mapping;
 		this.indexSettingsContributor = builder.indexSettingsContributor;
+
+		this.rawProjectionConverter = ProjectionConverter.passThrough( String.class );
+		this.rawDslConverter = new DslConverter<>( String.class, RawNoParsingDslConverter.INSTANCE );
```
yrodiere, line 49: This seems... wrong? Why no parsing? It means all raw values people will pass to the DSL will be rendered as strings. Never numbers, never objects... Just strings.

IMO if you want to represent raw values as strings, you should use the "userFacingGson":

* where a `DslConverter` is appropriate, you'd use `userFacingGson.fromJson(userProvidedString, JsonElement.class)` to convert a user-provided JSON-formatted String (which could reprensent a number, an object, ...) to a `JsonElement` you can integrate into your request to Elasticsearch.
* where a `ProjectionConverter` is appropriate, you'd use `userFacingGson.toJson(jsonElementFromResponse)` to convert the `JsonElement` from the response to JSON-formatted String (which could reprensent a number, an object, ...) the user can then manipulate.

Now do you want to represent raw values as strings... that's another question. I guess in the case of avg aggregations on date/numeric fields, you don't, since you want a `double` there. Maybe we need something more subtle, like:

* if we expect JsonPrimitive on the JSON side, the "raw" model will be String/Number/Boolean/Character, depending on the case
* if we expect JsonArray on the JSON side, `List<T>` with T being determined using the same rules
* if we expect JsonObject on the JSON side... I don't know. `JsonObject` maybe? And document this might change in the future.

yrodiere (commented):
Pretty sure you won't like what I have to say, but... feel free to argue or ignore me :)

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/converter/impl/ElasticsearchDslProjectionHelper.java
```diff
@@ -0,0 +1,85 @@
...
+import org.hibernate.search.backend.elasticsearch.search.common.impl.ElasticsearchSearchIndexValueFieldContext;
+import org.hibernate.search.backend.elasticsearch.types.codec.impl.ElasticsearchFieldCodec;
+import org.hibernate.search.engine.backend.types.converter.spi.DslConverter;
+import org.hibernate.search.engine.reporting.spi.EventContexts;
+import org.hibernate.search.engine.search.common.ValueModel;
+import org.hibernate.search.util.common.logging.impl.LoggerFactory;
+
+import com.google.gson.JsonElement;
+
+public final class ElasticsearchDslProjectionHelper {
```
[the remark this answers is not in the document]
marko-bekhta: mm none that I can think of at the moment 😄 will move things around.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/codec/impl/ElasticsearchGeoPointFieldCodec.java
```diff
@@ -52,6 +53,12 @@ public GeoPoint decode(JsonElement element) {
 		return GeoPoint.of( latitude, longitude );
 	}
 
+	@Override
+	public String raw(JsonElement element) {
+		GeoPoint point = decode( element );
+		return point == null ? null : FormatUtils.format( point );
```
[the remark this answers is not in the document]
marko-bekhta: > Wouldn't Gson.toJson(element) make more sense?

Maybe ... I guess it all comes down to the question how exactly we want to represent the raw values for the Elasticsearch backend.... I'd be happy to just accept JsonElement and let the users figure everything out 😄 
But then I think last time we discussed it, you said we shouldn't expose the JSON lib to the user ...

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/codec/impl/AbstractElasticsearchVectorFieldCodec.java
```diff
@@ -51,6 +51,11 @@ public F decode(JsonElement element) {
 		return fromJsonArray( JsonElementTypes.ARRAY.fromElement( element ) );
 	}
 
+	@Override
+	public String raw(JsonElement element) {
+		return element.getAsJsonArray().asList().toString();
```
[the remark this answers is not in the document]
marko-bekhta: mm ok, so I guess we would keep a reference to `Gson` in the codec?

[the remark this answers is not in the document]
yrodiere: See https://github.com/hibernate/hibernate-search/pull/4214/files/b1a41f4edc7cb1ca78f20065ef11d1a637f618f9#diff-5340406498b1c3572ff15fbfe5c800765ec9d38d9db1ddc5e59815b6cc4b80a7

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/impl/ElasticsearchIndexValueFieldType.java
yrodiere: BTW if you want to just return `JsonElement`, document this as unstable and change the behavior when we integrate the "raw avg as double", that's fine by me.

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/codec/impl/AbstractElasticsearchFieldCodec.java
```diff
@@ -0,0 +1,30 @@
...
+		return gson.fromJson( value, JsonElement.class );
+	}
+
+	@Override
+	public String fromJsonElementToString(JsonElement value) {
+		if ( value == null || value.isJsonNull() ) {
+			return null;
+		}
+		return gson.toJson( value );
+	}
```
marko-bekhta, line 29: oooooooook 😄 
So something like this I guess ... and then a lot of noise for the codec change (so may want to ignore that) though some codecs now are not singletons anymore

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/impl/ElasticsearchIndexValueFieldType.java
```diff
@@ -128,17 +131,46 @@ private JsonPrimitive elasticsearchTypeAsJson() {
...
+			return codec.fromJsonElementToString( value );
+		}
+
+		@Override
+		public boolean isCompatibleWith(FromDocumentValueConverter<?, ?> other) {
+			if ( !( other instanceof RawDslConverter ) ) {
+				return false;
+			}
+			return codec.isCompatibleWith( ( (RawDslConverter<?>) other ).codec );
+		}
```
marko-bekhta, line 174: and Converters for the Elasticsearch backend become this then

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/impl/ElasticsearchIndexValueFieldType.java
marko-bekhta: I mean ... you had a reason for not wanting to expose the Gson 😄 so unless we reaaaaaaaly need to let's work around it 😃

yrodiere (approved):
Thanks! I didn't read everything (obviously) but at least the documentation and key parts of implementation look good to me.

Let's merge once CI agrees :)

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/ProjectionConstructorFieldProjectionIT.java
yrodiere: Oh I forgot about this one... let me have a look

backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/types/impl/ElasticsearchIndexValueFieldType.java
marko-bekhta: I've pushed a change so that we work with strings but through a Gson... now if we want to do this 

> if we expect JsonPrimitive on the JSON side, the "raw" model will be String/Number/Boolean/Character, depending on the case

hmm IDK ... so we'd update the codec to be `interface ElasticsearchFieldCodec<F, *V*> ` and those couple methods I've added would become:
```
	JsonElement fromJsonStringToElement(V value);

	V fromJsonElementToString(JsonElement value);
```
(obviously with a different name 😄). If we consider the `.avg( "myDateTimeField", Double.class, ModelValue.RAW)` it may work if we just have something like:

```java
double d = DoubleCodec.fromJsonElementToString(value); // just reusing the same name as above...
if (valuemodel == RAW) {
    return d;
} else {
  // I think Fabio had something in his PR to convert double to the field type....
  return actualFieldCodec.smth(d);
}
```

maybe it'll work 😃

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/mapping/definition/ProjectionConstructorFieldProjectionIT.java
yrodiere: I see. What about only detecting conflicts when both have a non-default value?

It would work if developers set one of the two, of course:

```
@FieldProjection => MAPPING
@FieldProjection(convert = YES) => MAPPING
@FieldProjection(model = MAPPING) => MAPPING
@FieldProjection(convert = NO) => INDEX
@FieldProjection(model = INDEX) => INDEX
@FieldProjection(model = STRING) => STRING
@FieldProjection(model = RAW) => RAW
```

It would correctly fail in most cases when developers explicitly set two to non-default values:

```
@FieldProjection(convert = NO, model = INDEX) => ERROR
@FieldProjection(convert = NO, model = RAW) => ERROR
@FieldProjection(convert = NO, model = STRING) => ERROR
```

It incorrectly ignore the conflict when developers explicitly set one default and one non-default value:

```
@FieldProjection(convert = YES, model = INDEX) => INDEX ... :/
@FieldProjection(convert = YES, model = RAW) => RAW ... :/
@FieldProjection(convert = YES, model = STRING) => STRING ... :/
@FieldProjection(convert = NO, model = MAPPING) => INDEX ... :/
```

But then those last few examples are far-fetched... no?

yrodiere: > I just thought ... maybe a bad idea, but what if we add `ValueConvert.DEFAULT` and `ValueModel.DEFAULT` and use these in the annotation, mark both as deprecated, once we remove ValueConvert, we'll remove the `ValueModel.DEFAULT` constant too... that way we will know that if the value is not `DEFAULT` the user set it explicitly ...

+1 even better.

Adding an enum value is technically a breaking change since people might be doing `switch(valueConvert) { case YES:  /* ... */ return; case NO: /* ... */ return; }` somewhere, and will now end up with a compilation error.
But then, we were planning to add `ValueConvert.PARSE` before, and that would have broke their app just the same :shrug:

yrodiere:
Merged, thanks a lot!

marko-bekhta:
thank you ! 😃 🎉

PR #4230, opened 2024-07-15
HSEARCH-5182 Enable knn similarity filter for the OpenSearch backend / HSEARCH-5183 Add a score based alternative to similarity filter in the knn predicate

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5182
https://hibernate.atlassian.net/browse/HSEARCH-5183

----------------------

----------------------

### DIFF
--- a/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchKnnPredicate.java
+++ b/backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchKnnPredicate.java
@@ -217,9 +217,14 @@ public abstract class ElasticsearchKnnPredicate extends AbstractElasticsearchSin
 		private static final JsonAccessor<Integer> K_ACCESSOR = JsonAccessor.root().property( "k" ).asInteger();
 
 		private static final JsonObjectAccessor FILTER_ACCESSOR = JsonAccessor.root().property( "filter" ).asObject();
+		private static final JsonAccessor<Float> MAX_DISTANCE = JsonAccessor.root().property( "max_distance" ).asFloat();
+		private static final JsonAccessor<Float> MIN_SCORE = JsonAccessor.root().property( "min_score" ).asFloat();
+
+		private final Float score;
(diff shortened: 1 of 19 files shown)


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/impl/ElasticsearchKnnPredicate.java
```diff
@@ -226,21 +237,37 @@ protected JsonObject doToJsonQuery(PredicateRequestContext context, JsonObject o
...
+			if ( similarity != null ) {
+				MAX_DISTANCE.set( innerObject, similarity );
+			}
+			if ( score != null ) {
+				MIN_SCORE.set( innerObject, score );
+			}
+			if ( similarity == null && score == null ) {
+				// [knn] requires exactly one of k, distance or score to be set
+				K_ACCESSOR.set( innerObject, k );
+			}
```
marko-bekhta, line 249: This may be a change of "interest" 😃 
as per that comment, OpenSearch only allows setting one of the options. Having `k` sent along with score/distance leads to an exception returned by the cluster. 
Hence my thinking here - if we have score/distance filter we ignore the `k` value and mention this behavior in the ref docs.

integrationtest/backend/tck/src/main/java/org/hibernate/search/integrationtest/backend/tck/search/predicate/KnnPredicateSpecificsIT.java
```diff
@@ -1036,13 +1033,66 @@ public static List<? extends Arguments> similarityFilterFloat() {
 			return args;
 		}
 
+		@ParameterizedTest
+		@MethodSource
+		void similarityFilterFloatOpenSearch(SimpleMappedIndex<IndexBinding> index, float similarity, float score,
```
marko-bekhta, line 1038: I couldn't get OpenSearch to behave the same as Lucene and Elasticsearch for this similarity/distance filter, so I've just added a different test for it instead...

PR #4233, opened 2024-07-16
Follow up on HSEARCH-5193 - Move the reproducibility check to the nightly build

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5193

https://hibernate.zulipchat.com/#narrow/stream/132092-hibernate-search-dev/topic/Build.20reproducibility.20check

We can start with something like this, I guess?

And the test build execution here https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search-personal-marko/detail/build%2FHSEARCH-5193-move-repro-check-to-nightly/2/pipeline/

----------------------

----------------------

### DIFF
--- a/ci/nightly/Jenkinsfile
+++ b/ci/nightly/Jenkinsfile
@@ -1,0 +1,71 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright Red Hat Inc. and Hibernate Authors
+ */
+
+@Library('hibernate-jenkins-pipeline-helpers@1.17') _
+
+def withMavenWorkspace(Closure body) {
+	withMaven(jdk: 'OpenJDK 21 Latest', maven: 'Apache Maven 3.9',
+			mavenLocalRepo: env.WORKSPACE_TMP + '/.m2repository',
+			options: [
+					// Artifacts are not needed and take up disk space
+					artifactsPublisher(disabled: true),
+					// stdout/stderr for successful tests is not needed and takes up disk space
+					// we archive test results and stdout/stderr as part of the build scan anyway,
+					// see https://ge.hibernate.org/scans?search.rootProjectNames=Hibernate%20Search
+					junitPublisher(disabled: true)
+			]) {
+		withCredentials([string(credentialsId: 'ge.hibernate.org-access-key',
+				variable: 'DEVELOCITY_ACCESS_KEY')]) {
+			withGradle {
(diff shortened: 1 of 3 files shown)


marko-bekhta:
@yrodiere let me know if you want to look at this one. 
I've kept the develocity config and the test container env variable, even though both aren't really needed for this particular check (we don't want build cache to test the reproducibility, and we aren't running tests, so there are no containers). I did that so that when we add more things to the nightly build, we do not forget to have the build cache and other things.

I've also updated the corresponding HV pr https://github.com/hibernate/hibernate-validator/pull/1376, that had a similar check, to do the same thing (have a nightly build).

ci/nightly/Jenkinsfile
```diff
@@ -0,0 +1,71 @@
...
+			}
+		}
+	}
+}
+
+pipeline {
+	agent none
+	triggers {
+		cron '@midnight'
+	}
```
yrodiere, line 33: We'll need to make sure the job created through the Jenkins web UI excludes pull requests and dependabot branches, otherwise this cron will be very expensive.

ci/nightly/Jenkinsfile
```diff
@@ -0,0 +1,71 @@
...
+			agent {
+				label 'Worker&&Containers'
+			}
+			steps {
+				// The timeout cannot be in stage options, because that would
+				// include the time needed to provision a node.
+				timeout(time: 30, unit: 'MINUTES') {
+					withMavenWorkspace {
+						echo "Generate the artifacts."
+						sh "mvn clean install -Pskip-checks -Preproducible"
```
yrodiere, line 53: > `-Preproducible`

I won't lie, this gave me pause. Any reason why the build is not reproducible by default, without needing a profile to be enabled explicitly?

yrodiere (approved):
Thanks, LGTM. I added a few comments but feel free to merge.

ci/nightly/Jenkinsfile
marko-bekhta: Yeah... I guess we should target `main` only at this time and 7.2+ branches.. any other branches will fail the check

ci/nightly/Jenkinsfile
yrodiere: `-Preproducibility-check`?

ci/nightly/Jenkinsfile
yrodiere: Older branches are not an issue: they don't have the Jenkinsfile so won't get a run anyway.

PR #4281, opened 2024-08-13
HSEARCH-5223 Regenerate avro Dtos / add a sources-not-modified check to the build

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5223

I've tried the change in this job https://ci.hibernate.org/blue/organizations/jenkins/hibernate-search-personal-marko/detail/build%2FHSEARCH-5223-add-source-update-check/3/pipeline/

I'm not sure what's going on with the Avro plugin... when I reverted the changes committed by CI and tried running a build, the Dtos were not updated. Only when I just removed them entirely did they get regenerated, and the imports changed...

so the source modification check won't catch this problem.

I thought about dropping them from git and letting them be generated into target each time, but looking at the comments here https://hibernate.atlassian.net/browse/HSEARCH-4638 .. 😖 

I thought about printing the diff if the check fails but then thought that it may be potentially dangerous, so didn't do that 😄 

----------------------

----------------------

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -447,6 +447,9 @@ stage('Default build') {
 							install \
 					"} \
 			"""
+			// Quick check after the initial build, if we spot the change here, no need to even proceed..
+			checkNoSourceFilesModifiedOrFail()
+
 			dir(helper.configuration.maven.localRepositoryPath) {
 				stash name:'default-build-result', includes:"org/hibernate/search/**"
 				stash name:'default-build-cache', includes:".develocity/**"
@@ -962,6 +965,8 @@ void withMavenWorkspace(Map args, Closure body) {
 		sh 'ci/docker-cleanup.sh'
 		tryFinally(body, { // Finally
 			sh 'ci/docker-cleanup.sh'
+		}, {
+			checkNoSourceFilesModifiedOrFail()
 		})
 	})
 }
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -1155,3 +1160,14 @@ def tryFinally(Closure main, Closure ... finallies) {
 		throw mainFailure
 	}
 }
+
+void checkNoSourceFilesModifiedOrFail() {
+	echo 'Checking sources for changes'
+	// We are ignoring the failsafe-summary.xml as this file may get modified by integration tests running on
+	//  non-default envs and we don't want to change that at the moment:
```
yrodiere, line 1167: Shouldn't this just be in `.gitignore`?

In fact I'd put `/.empty/*` in gitignore; we should never commit anything in a directory named `.empty` whose only purpose is being empty...

marko-bekhta: I thought about it yes, but there are a couple files in that "empty" directory 🙈 so I thought I'd better keep it as is 😃

yrodiere: Ah, ok, I misunderstood. Yes I see `.empty` is not supposed to be empty, but `.empty/directory` is, and `.empty/failsafe-summary.xml` is actually a checked-in file representing an empty test run.

Now I'm wondering why test runs would end up modifying this file... There must be something wrong :/

marko-bekhta: I noticed that the summary file was only modified by the OpenSearch build (so probably non-default builds that run tests ... ). But if you are saying that it should've stayed unchanged, I'll look into why it gets updated.

PR #4284, opened 2024-08-14
HSEARCH-5226 Ignore collecting scale on formula properties

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5226

----------------------

----------------------


marko-bekhta:
If the build passes, I'll backport this one to other branches since the resulting error is a bit obscure, it mentions formulas and so on, but the problem is that Search is trying to find an inverse side of a relation and stumbles upon a field that happens to be a formula and ....

PR #4291, opened 2024-08-21
HSEARCH-5208 Add a fail-fast configuration option to the mass indexer

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5208

----------------------

----------------------

### DIFF
--- a/documentation/src/main/asciidoc/public/reference/_indexing-massindexer.adoc
+++ b/documentation/src/main/asciidoc/public/reference/_indexing-massindexer.adoc
@@ -363,6 +363,13 @@ Defaults to a threshold defined by the failure handler in use; see `MassIndexing
 `FailureHandler#failureFloodingThreshold`.
 For the default log-based failure handler, the default threshold is 100.
 
+|`failFast(boolean)`
+| `false`
+|*This feature is _incubating_: it is still under active development.*
+An option to stop the indexing right after an error is encountered during the process,
+without waiting for the process to attempt indexing the remaining entities.
+With fail-fast enabled, the mass indexer will attempt to cancel any mass-indexing internal processes after the first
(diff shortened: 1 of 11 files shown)


mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/impl/PojoMassIndexingBatchCoordinator.java
```diff
@@ -83,6 +84,10 @@ public PojoMassIndexingBatchCoordinator(PojoMassIndexingMappingContext mappingCo
 				mappingContext.threadPoolProvider(),
 				mappingContext.failureHandler()
 		);
+
+		if ( failFast ) {
+			getNotifier().addOnFailureHandlerCallback( this::cleanUpOnFailure );
```
marko-bekhta, line 89: Hey  @yrodiere 

The idea for this fail-fast implementation is to cancel the coordinator once the failure handler is called. Cancelling the coordinator also cancels the sub-tasks that the coordinator started. So that seems like a simple solution....
Do you think this approach makes sense, or maybe you'd have some other suggestion? Thanks! 😃

marko-bekhta: mmm a fair point 😃 
I guess... I wanted it to be more explicit, i.e. call the cancel because we were in the "fail-fast-if-statement" 🙈.

With the boolean and an exception, this seems to work ok (throwing a `MassIndexingOperationHandledFailureException` so that we do not need to add anything to the exception handling)

yrodiere (commented):
LGTM, only reference documentation is missing :)

PR #4314, opened 2024-09-11
HSEARCH-2945 Configure default mass indexing logging monitor

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-2945

I've tried a few approaches, and it seems that having a "dedicated" method to configure the default monitor looks better than others...
Having more options on the mass indexer itself may work for a config like "don't do counts at all" but not for the "count the entities before indexing starts. 

Let me know if you have any other ideas we could try. Opening as a draft for now as some more tests are needed and the impl of the monitor should be cleaned up a bit.

----------------------

----------------------


yrodiere:
It's a shame that this will only work with the default monitor, no? :thinking: 

> Having more options on the mass indexer itself may work for a config like "don't do counts at all"

Ok.

> but not for the "count the entities before indexing starts.

Why?

---

Regarding the solution... In order to make it work with the default monitor, but also custom ones, we could imagine allowing the monitor itself to influence the behavior? I.e.:

```
class MyMonitor1 implements MassIndexingMonitor {

    void initialize(MassIndexingMonitorContext context) {
        context.countOnStart(this::onCount); // Callback called immediately
    }

    void onCount(EntityTypeReference type, long count) {
         // ...
    }
}
class MyMonitor2 implements MassIndexingMonitor {

    void initialize(MassIndexingMonitorContext context) {
        context.countBeforeTypeStart(this::onCount); // Callback called later
    }

    void onCount(EntityTypeReference type, long count) {
         // ...
    }
}
class MyMonitor2 implements MassIndexingMonitor {

    void initialize(MassIndexingMonitorContext context) {
        context.countOnStartAndBeforeTypeStart(this::onCount, this::onCountAdjustment); // First callback called now, secont callback called later
    }

    void onCount(EntityTypeReference type, long count) {
         // ...
    }
    void onCountAdjustment(EntityTypeReference type, long initialCount, long updatedCount) {
         // ...
    }
}
```

I'm not suggesting this *exact* API -- names are bad, in particular -- but you get the idea.

Then making it work with the default monitor would be a matter of exposing that monitor -- e.g.:

```java
massIndexer.monitor(DefaultMassIndexingMonitor.builder().countOnStart(true));
// OR 
massIndexer.monitor(DefaultMassIndexingMonitor.builder().countOnBeforeType(true));
// OR ...
```

marko-bekhta:
> Ok.
> 
> > but not for the "count the entities before indexing starts.
> 
> Why?

well ... the call to get the total count is currently made by mass indexer itself, and the value is passed to the monitor (no matter what the impl of that monitor is, right). So if we config this at the mass indexer level it makes sense to allow disable the count call for both the built-in one and custom ones.
Now this "new thing" to get the counts before anything even starts and have an approximation that may get adjusted later is something that would be specific to our own monitor implementation. I didn't think that we'd want to call it from the mass indxer and pass the values around to custom monitor implementations... That was the reason why I was thinking that exposing this other property (to do the count-all-before-things-start) won't work .... 

what you wrote made me think ... we have this in the type group monitor now:

https://github.com/hibernate/hibernate-search/blob/684ba49771e82db07b4504da349b5bfa8a341436/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/MassIndexingTypeGroupMonitor.java#L43-L48

what if instead of the total count or nothing, we pass the context that exposes some access to the loading strategy ( that can do the counting) through a context class and have something like:

```java
public interface MassIndexingTypeGroupMonitor {
	void documentsAdded(long increment);

	// gets called when we create a workspace for the type group
	// so essentially can be used by us to do the counts before the indexing starts
	void typeGroupInitialized(SomeContext context);
	
	// gets called when the identifier loading runnable starts:
	void indexingStarted(SomeContext context);

	// gets called when all indexing runnables for the group are completed.
	void indexingCompleted(SomeContext context);
}
```

then in the main mass indexing monitor we ~~drop~~ deprecate the `addToTotalCount` method with `MassIndexingTypeGroupMonitor#indexingStarted` being it's replacement ?

yrodiere:
Your suggestion makes sense to me.

This is starting to get complex though, so I'd suggest:

1. Being careful about naming, to make things extra clear. `documentsAdded` => `documentsIndexed` would make sense I think, and `typeGroupInitialized` => `initialize()` (since it's about initializing the monitor itself, more than anything else) or `indexingPlanned()`, if you want to keep it event-like? IDK.
2. Not forgetting about providing ways to configure the default monitor -- see my suggestion above.
3. Making it obvious that counting may, in some cases, not return anything -- because the loading strategy may not support it. Using `Optional` in APIs would make sense I think.
4. Deprecating *all* methods in the main monitor that duplicate what you're doing -- `documentsBuilt`, `documentsIndexed`, etc.
5. Making sure that existing implementations (implementing deprecated methods) still work -- we'll need tests for that.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/MassIndexingMonitor.java
```diff
@@ -21,9 +21,21 @@
...
+	 * @param context Describes the type group for which the monitor is requested.
+	 * @return A type group mass indexing monitor. By default, a no-op monitor is returned.
+	 * @see MassIndexingTypeGroupMonitor
+	 */
 	@Incubating
-	default MassIndexingTypeGroupMonitor typeGroupMonitor(MassIndexingTypeGroupMonitorContext context) {
-		return NoOpMassIndexingTypeGroupMonitor.INSTANCE;
+	default MassIndexingTypeGroupMonitor typeGroupMonitor(MassIndexingTypeGroupMonitorCreateContext context) {
+		return new LegacyDelegatingMassIndexingTypeGroupMonitor( this, context );
 	}
```
marko-bekhta, line 39: I did not add the initialize method to `MassIndexingTypeGroupMonitor` since that'll be called in the exact same place where we call this one to create a type group, so users can do whatever they wanted in init in the constructor of the type group monitor.

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/massindexing/MassIndexingTypeGroupMonitor.java
```diff
@@ -29,22 +27,22 @@ public interface MassIndexingTypeGroupMonitor {
...
 	 * otherwise an empty optional is supplied.
 	 */
-	void indexingStarted(OptionalLong totalCount);
+	void indexingStarted(MassIndexingTypeGroupMonitorContext context);
 
 	/**
 	 * Notify the monitor that indexing of the type group is completed.
 	 */
-	void indexingCompleted();
+	void indexingCompleted(MassIndexingTypeGroupMonitorContext context);
```
marko-bekhta, line 46: I used the same context for both of these, and for now there's only a total count

integrationtest/mapper/pojo-base/src/test/java/org/hibernate/search/integrationtest/mapper/pojo/massindexing/MassIndexingDefaultMonitorIT.java
```diff
@@ -0,0 +1,237 @@
...
+			// so we expect 1 purge, 1 mergeSegments and 1 flush calls in this order:
+			backendMock.expectIndexScaleWorks( Book.NAME, searchSession.tenantIdentifierValue() )
+					.purge()
+					.mergeSegments()
+					.flush()
+					.refresh();
+
+			try {
+				indexer.monitor( DefaultMassIndexingMonitor.builder().countOnBeforeType( doCounts ).build() )
+						.startAndWait();
```
marko-bekhta, line 112: So that works 😄 but ! I'll have to go through the docs + add something to the migration guide and add more tests

marko-bekhta:
@yrodiere let me know if you'd want to have a look at this one 😉

yrodiere:
Seems you still need to add something to the migration guide?

yrodiere (approved):

PR #4315, opened 2024-09-11
HSEARCH-4720 Throw exceptions instead of logging warnings for failures to configure association inverse path resolution

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4720

In one of the places marked to throw an exception, the code can enter the condition for valid reasons. I've tried to describe that with an example and kept the code simply returning early as before instead of throwing an exception. 

----------------------

----------------------


marko-bekhta:
> [25.0% Coverage on New Code](https://sonarcloud.io/component_measures?id=org.hibernate.search%3Ahibernate-search-parent&pullRequest=4315&metric=new_coverage&view=list) (required ≥ 80%)

I guess that's good 😄 since we don't expect those exceptions to actually be thrown 😃

PR #4329, opened 2024-09-26
HSEARCH-4577 @ProjectionConstructor: aggregating multi-valued field/object projections in Set, SortedSet, etc. instead of List

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4577

~~for now, just to get a CI run~~ 🙈 
still need more tests, docs and migration guide udpates

----------------------

----------------------

### DIFF
--- a/engine/src/main/java/org/hibernate/search/engine/search/projection/dsl/FieldProjectionValueStep.java
+++ b/engine/src/main/java/org/hibernate/search/engine/search/projection/dsl/FieldProjectionValueStep.java
@@ -25,7 +25,7 @@ public interface FieldProjectionValueStep<N extends FieldProjectionOptionsStep<?
 	/**
 	 * Defines the projection as multi-valued, i.e. returning {@code List<T>} instead of {@code T}.
 	 * <p>
-	 * Calling {@link #multi()} is mandatory for multi-valued fields,
+	 * Calling {@link #multi()}/{@link #multi(MultiProjectionTypeReference)} is mandatory for multi-valued fields,
 	 * otherwise the projection will throw an exception upon creating the query.
 	 *
 	 * @return A new step to define optional parameters for the multi-valued projections.
@@ -37,7 +37,7 @@ public interface FieldProjectionValueStep<N extends FieldProjectionOptionsStep<?
 	/**
 	 * Defines the projection as multi-valued, i.e. returning a collection, e.g. {@code List<T>}, instead of {@code T}.
 	 * <p>
-	 * Calling {@link #multi()} is mandatory for multi-valued fields,
+	 * Calling {@link #multi()}/{@link #multi(MultiProjectionTypeReference)} is mandatory for multi-valued fields,
 	 * otherwise the projection will throw an exception upon creating the query.
 	 * <p>
 	 * Requires a collection type reference, either a built-in (see {@link MultiProjectionTypeReference}) or a custom one.
--- a/engine/src/main/java/org/hibernate/search/engine/search/projection/dsl/FieldProjectionValueStep.java
+++ b/engine/src/main/java/org/hibernate/search/engine/search/projection/dsl/FieldProjectionValueStep.java
@@ -30,6 +30,22 @@ public interface FieldProjectionValueStep<N extends FieldProjectionOptionsStep<?
 	 *
 	 * @return A new step to define optional parameters for the multi-valued projections.
 	 */
-	FieldProjectionOptionsStep<?, List<T>> multi();
+	default FieldProjectionOptionsStep<?, List<T>> multi() {
+		return multi( MultiProjectionTypeReference.list() );
+	}
+
+	/**
+	 * Defines the projection as multi-valued, i.e. returning a collection, e.g. {@code List<T>}, instead of {@code T}.
+	 * <p>
+	 * Calling {@link #multi()} is mandatory for multi-valued fields,
+	 * otherwise the projection will throw an exception upon creating the query.
+	 * <p>
+	 * Requires a collection type reference, either a built-in (see {@link MultiProjectionTypeReference}) or a custom one.
+	 *
+	 * @param collectionTypeReference Collection type reference that specifies the expected collection into which the values have to be collected into.
+	 * @return A new step to define optional parameters for the multi-valued projections.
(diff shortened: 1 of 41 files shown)


engine/src/main/java/org/hibernate/search/engine/search/projection/dsl/FieldProjectionValueStep.java
```diff
@@ -25,11 +25,27 @@ public interface FieldProjectionValueStep<N extends FieldProjectionOptionsStep<?
...
+	 * Calling {@link #multi()}/{@link #multi(MultiProjectionTypeReference)} is mandatory for multi-valued fields,
+	 * otherwise the projection will throw an exception upon creating the query.
+	 * <p>
+	 * Requires a collection type reference, either a built-in (see {@link MultiProjectionTypeReference}) or a custom one.
+	 *
+	 * @param collectionTypeReference Collection type reference that specifies the expected collection into which the values have to be collected into.
+	 * @return A new step to define optional parameters for the multi-valued projections.
+	 * @see MultiProjectionTypeReference
+	 */
+	<C> FieldProjectionOptionsStep<?, C> multi(MultiProjectionTypeReference<C, T> collectionTypeReference);
```
marko-bekhta, line 49: I wonder whether renaming this from multi to container and then adding built-in type references for optional is a good idea ...

engine/src/main/java/org/hibernate/search/engine/search/projection/dsl/MultiProjectionTypeReference.java
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Defines the projection collection type to use for a multi-valued fields.
+ * <p>
+ * Either one of the built-in type references, provided by this interface, can be used,
+ * or a custom one can be implemented for other collection types.
+ * @param <C> The type of the collection.
+ * @param <V> The type of the elements in the collection.
+ */
+@Incubating
+public interface MultiProjectionTypeReference<C, V> {
```
marko-bekhta, line 27: I couldn't come up with a better name for this one ....  so any suggestions are more than welcome 😃

yrodiere: Couldn't you just promote `ProjectionAccumulator` to API, and use `ProjectionAccumulator.Provider`? It's admittedly a complex interface, but it's what we use under the hood, and if you provide implementations for the main container types, users will only need to implement it very rarely.

Then you could rename `ProjectionAccumulator#single()` to `nullable()` and introduce `optional()`.

Well, you'll probably need to create a new interface, and juggle with inheritance and deprecations, but you see what I mean.

engine/src/main/java/org/hibernate/search/engine/search/projection/dsl/FieldProjectionValueStep.java
yrodiere: +1 for `container` as the default is already "multi", so it feels a bit weird to introduce a "multi".

However, don't we expose "multi" methods in other DSLs, especially in the search DSL? We'd need to be consistent...

engine/src/main/java/org/hibernate/search/engine/search/projection/dsl/MultiProjectionTypeReference.java
marko-bekhta: > It's admittedly a complex interface

yeah ... that's why I was thinking that since we are using this index-based approach for transformation operations in the accumulator that most implementations would still be backed by a list as a temporary storage option and the only difference would be the `R finish(A accumulated)` that performs the final transformation. Hence, I thought extracting this method (as `convert()`) would work for the users with some custom collections and we'd just delegate to in the list-based accumulator.

but let me see how promoting the accumulator to the API would work out 😃

engine/src/main/java/org/hibernate/search/engine/search/projection/dsl/FieldProjectionValueStep.java
marko-bekhta: yeah we also have that `HighlightProjectionOptionsStep#single()` ... 
It would be nice if we can just deprecate all that and do something like

```java
.select(
	f -> f.composite()
		.from(
			f.highlight( "anotherString" ),  // default list as it is now
			f.highlight( "anotherString" ).accumulator(ProjectionAccumulator.single() ), // instead of calling  f.highlight( "anotherString" ).single(),
			f.highlight( "anotherString" ).accumulator(ProjectionAccumulator.optional() ), // still single but wraps the string into an optional
			f.highlight( "anotherString" ).accumulator(ProjectionAccumulator.sortedSet() ), // multi but non-default collection type
			f.highlight( "anotherString" ).accumulator(ProjectionAccumulator.sortedSet(customComparator) ), // multi but non-default collection type with custom comparator
			f.highlight( "anotherString" ).accumulator(ProjectionAccumulator.array() ), // multi but non-default collection type
			f.highlight( "anotherString" ).accumulator(CustomUserDefinedThing.smth() ), // whatever the user came up with
		).asList()
)
```

engine/src/main/java/org/hibernate/search/engine/search/projection/dsl/MultiProjectionTypeReference.java
yrodiere: Fair enough, either solution has its merits. I'll let you have a look and decide :)

> Hence, I thought extracting this method (as `convert()`) would work for the users with some custom collections and we'd just delegate to in the list-based accumulator.

Note you could also provide a `static <V, C> Provider<V, C> simple(Function<List<V>, C> converter)` "factory method" to allow simple list-based implementations.

yrodiere: > Note you could also provide a `static <V, C> Provider<V, C> simple(Function<List<V>, C> converter)` "factory method" to allow simple list-based implementations.

https://github.com/hibernate/hibernate-search/pull/4334 :grinning:

marko-bekhta:
superseded by https://github.com/hibernate/hibernate-search/pull/4344

PR #4344, opened 2024-10-03
HSEARCH-4577 @ProjectionConstructor: aggregating multi-valued field/object projections in Set, SortedSet, etc. instead of List 

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-4577

----------------------

----------------------


backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/logging/impl/Log.java
```diff
@@ -602,7 +602,7 @@ SearchException invalidFieldValueType(@FormatWith(ClassFormatter.class) Class<?>
 	@Message(id = ID_OFFSET + 113,
 			value = "Invalid cardinality for projection on field '%1$s': the projection is single-valued,"
 					+ " but this field is multi-valued."
-					+ " Make sure to call '.multi()' when you create the projection.")
+					+ " Make sure to call '.accumulator(...)' when you create the projection.")
```
yrodiere, line 605: This makes me wonder about the name... While `ProjectionAccumulator` was fine(-ish) for an SPI that most likely nobody except us will use, it's arguably a bit weird and I don't think people will look for that word when looking for this feature.

I suppose `ProjectionCollector` is a better name, but then people might wonder why we didn't just use `java.util.stream.Collector`. From the javadoc it's mainly because of unnecessary features in `Collector` and because of the `transform` stuff.

yrodiere (commented):
Just noticed this was still pending -- did you encounter any problem? Did I forget a review? :x

marko-bekhta:
> Just noticed this was still pending -- did you encounter any problem? Did I forget a review? :x

Problems ? no no problems, only opportunities 😄. 
From my notes I only have: take another look at the docs.

> I suppose ProjectionCollector is a better name, but then people might wonder why we didn't just use java.util.stream.Collector. From the javadoc it's mainly because of unnecessary features in Collector and because of the transform stuff.

mm ok 😃 I can rename it 👍🏻 (it was good I left the PR hanging here 😄) So if we go with `ProjectionCollector ` then corresponding `.accumulator(...)` methods become `.collector(...)` ? ooor ...

yrodiere:
> mm ok 😃 I can rename it 👍🏻 (it was good I left the PR hanging here 😄) So if we go with `ProjectionCollector ` then corresponding `.accumulator(...)` methods become `.collector(...)` ? ooor ...

I guess that's what makes most sense, yes. `collect(collector)` would be closer to `java.util.stream`, but confusing since our method doesn't actually perform the collecting, which happens later.

BTW I wonder if we could just use `java.util.stream.Collector` :D Probably something to investigate in a different ticket, though -- if we even investigate it.

marko-bekhta:
> I guess that's what makes most sense, yes. collect(collector) would be closer to java.util.stream, but confusing since our method doesn't actually perform the collecting, which happens later.

ok 😃, yeah that's what I thought too ... that we are not collecting in that particular place ... 

> BTW I wonder if we could just use java.util.stream.Collector :D Probably something to investigate in a different ticket, though -- if we even investigate it.

hah I was just looking at the collector interface 😄, but yeah I agree that it's for the other PR. But if we want to explore that option, should I mark `ProjectionCollector` as incubating for now? (if we end up replacing it with collector)

yrodiere:
> should I mark `ProjectionCollector` as incubating for now? (if we end up replacing it with collector)

That's probably a good idea regardless of our plans, yes.

engine/src/main/java/org/hibernate/search/engine/search/projection/spi/ArrayProjectionCollector.java
```diff
@@ -0,0 +1,59 @@
...
+	private ArrayProjectionCollector(Class<? super V> elementType) {
+		this.elementType = elementType;
+	}
+
+	private final Class<? super V> elementType;
+
+	@SuppressWarnings("unchecked")
+	@Override
+	public V[] doFinish(List<V> accumulated) {
+		V[] array = (V[]) Array.newInstance( elementType, accumulated.size() );
```
marko-bekhta, line 33: Soooo following the discussion at https://hibernate.zulipchat.com/#narrow/channel/132094-hibernate-orm-dev/topic/StandardStack.20and.20reflection
I guess this won't fly well with Quarkus too 😖 😃....

I can change this to use `IntFunction<T[]>` and simply use `accumulated.toList( ... )` but then..... it won't work fine here:

https://github.com/hibernate/hibernate-search/blob/12cd6cb0fefbcd068285a73daef5d2635fb067ff/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/search/definition/binding/impl/ProjectionBindingContextImpl.java#L417-L419

so we could do something like for sorted sets and throw an exception. Or give two ways to create array collector with class and with function. Use class in the `ProjectionBindingContextImpl` and it'll fail for users so they'd need to register their arrays themselves .... @yrodiere WDYT ? 🙈

yrodiere: I think it's fine-ish, as this only happens when a user explicitly requests an array, so it's reasonable to expect them to register that array type for reflection.

For `ProjectionBindingContextImpl`, I suppose we can add something to Quarkus that would detect projection constructors and register for reflection any array type used in their parameters. But only if we consider this important enough -- I personally don't. I wouldn't even have provided the feature TBH, because using arrays in domain models is a big no-no in my books. But now that it's there, I guess we might as well keep it.

PR #4347, opened 2024-10-07
HSEARCH-5045 Raise various default max-connection config properties for the Elasticsearch backend

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5045

----------------------

----------------------


marko-bekhta:
I wonder whether it would make sense to make the total max connections some high number (e.g. `Integer.MAX_VALUE`) by default and let the per-route value keep things limited?

yrodiere:
> I wonder whether it would make sense to make the total max connections some high number (e.g. `Integer.MAX_VALUE`) by default and let the per-route value keep things limited?

Seems to me this could lead to problems with large Elasticsearch cluster? I.e. with a 20-node cluster, one app instance could end up opening 400 connections...

marko-bekhta:
good point! 😃 thanks!

PR #4364, opened 2024-10-25
Rework/simplify GH Actions jobs

yrodiere:
* Use `pull_request` instead of `pull_request_target`
* Move Develocity build scan publishing for untrusted code to a separate workflow
* Split caches between trusted and untrusted code
* Update secrets to use "Develocity" name instead of "Gradle Enterprise"

Opening as draft as I need to check that this actually works, and will be checking on the ORM side: https://github.com/hibernate/hibernate-orm/pull/9163

----------------------

----------------------

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@
 # See https://docs.github.com/en/actions
 # for more information about GitHub actions.
 
-name: GitHub Actions Build
+name: GH Actions CI
 
 on:
   push:
@@ -24,9 +24,7 @@ on:
       - '!wip/**/dependency-update/**'
     tags:
       - '**'
-  # WARNING: Using pull_request_target to access secrets, but we check out the merge commit.
-  # See checkout action for details.
-  pull_request_target:
+  pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
     branches:
       # Pattern order matters: the last matching inclusion/exclusion wins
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -131,55 +118,39 @@ jobs:
...
-      - name: Publish Develocity build scan for previous build (pull request)
-        # Don't fail a build if publishing fails
-        continue-on-error: true
-        if: "${{ !cancelled() && github.event_name == 'pull_request_target' && github.repository == 'hibernate/hibernate-search' }}"
-        run: |
-          ./mvnw $MAVEN_ARGS develocity:build-scan-publish-previous
-        env:
-          # WARNING: exposes secrets, so must only be passed to a step that doesn't run unapproved code.
-          DEVELOCITY_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY_PR }}
+          DEVELOCITY_ACCESS_KEY: "${{ secrets.DEVELOCITY_ACCESS_KEY || '' }}"
```
marko-bekhta, line 141: Should this workflow have any reference to secrets at all ?

yrodiere: This workflow executes on `push` and on `pull_request` events.

On pull request, we don't trust the code being executed, but that's fine, because GitHub will not give access to secrets. So this particular line has no effect, at best it sets the key to an empty string.

On push, we trust the code being executed (because we reviewed it before pushing, right? :smirk: ), and GitHub Actions gives access to secrets. Which, critically, allows populating the Develocity remote cache, so that the next PR builds can benefit from it.

marko-bekhta (approved):

marko-bekhta:
Thanks!

PR #4419, opened 2024-12-06
HSEARCH-5285 Add Pojo Standalone mapper Javadocs to the bundle

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5285

While really the config was missing `mapper/pojo-standalone/src/main/java` and had a redundant `mapper/lucene/src/main/java`  this PR also adds an enforcer check, and while adding it it pointed out that the generated sources aren't included. So I've added those too, just in case.

----------------------

----------------------

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -165,12 +165,13 @@
                                 ${basedir}/../util/common/src/main/java;
                                 ${basedir}/../mapper/pojo-base/src/main/java;
                                 ${basedir}/../mapper/orm/src/main/java;
+                                ${basedir}/../mapper/pojo-standalone/src/main/java;
                                 ${basedir}/../backend/elasticsearch/src/main/java;
                                 ${basedir}/../backend/elasticsearch-aws/src/main/java;
                                 ${basedir}/../backend/lucene/src/main/java;
-                                ${basedir}/../mapper/lucene/src/main/java;
                                 ${basedir}/../mapper/orm-outbox-polling/src/main/java;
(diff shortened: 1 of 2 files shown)


distribution/pom.xml
```diff
@@ -160,17 +160,32 @@
...
+                                ${basedir}/../mapper/pojo-base/target/generated-sources/annotations;
+                                ${basedir}/../mapper/orm/target/generated-sources/annotations;
+                                ${basedir}/../mapper/pojo-standalone/target/generated-sources/annotations;
+                                ${basedir}/../backend/elasticsearch/target/generated-sources/annotations;
+                                ${basedir}/../backend/elasticsearch-aws/target/generated-sources/annotations;
+                                ${basedir}/../backend/lucene/target/generated-sources/annotations;
+                                ${basedir}/../mapper/orm-outbox-polling/src/main/avro/generated;
+                                ${basedir}/../mapper/orm-outbox-polling/target/generated-sources/annotations;
+                                ${basedir}/../mapper/orm-jakarta-batch/core/target/generated-sources/annotations;
+                                ${basedir}/../mapper/orm-jakarta-batch/jberet/target/generated-sources/annotations;
```
marko-bekhta, line 188: Hey @yrodiere 😃 
wanted to check with you if this makes sense to you, or would you say that we shouldn't include the generated sources here?

yrodiere: I guess it cannot hurt, since there is not a lot of generated sources and even less API generated sources. Sure, let's do it :)

PR #4433, opened 2024-12-16
Use specific Jenkins nodes for releases

yrodiere:
This should be safer. 

It requires moving the snapshot publishing to a dedicated job, but ORM already does that, so we might as well align.

I took this opportunity to clean up sections about CI in `CONTRIBUTING.md`/`MAINTAINERS.md`.

We'll need to backport this to all branches where we want to perform releases or publish snapshots...

----------------------

----------------------

### DIFF
--- a/ci/snapshot-publish/Jenkinsfile
+++ b/ci/snapshot-publish/Jenkinsfile
@@ -14,7 +14,7 @@ if (currentBuild.getBuildCauses().toString().contains('BranchIndexingCause')) {
 
 pipeline {
 	agent {
-		label 'Worker&&Containers'
+		label 'Release'
 	}
 	tools {
 		maven 'Apache Maven 3.9'
--- a/ci/snapshot-publish/Jenkinsfile
+++ b/ci/snapshot-publish/Jenkinsfile
@@ -1,0 +1,53 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright Red Hat Inc. and Hibernate Authors
+ */
+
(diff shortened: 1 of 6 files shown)


ci/snapshot-publish/Jenkinsfile
```diff
@@ -0,0 +1,53 @@
...
+	stages {
+		stage('Publish') {
+			steps {
+				script {
+					withMaven(mavenSettingsConfig: 'ci-hibernate.deploy.settings.maven',
+							mavenLocalRepo: env.WORKSPACE_TMP + '/.m2repository') {
+						sh """mvn \
+								-Pci-build \
+								-DskipTests \
+								clean install \
```
marko-bekhta, line 41:
```suggestion
								clean deploy \
```

we probably need a `deploy` goal to publish the snapshot...

marko-bekhta (approved):
Thanks! This looks like a nice improvement 🎉

ci/snapshot-publish/Jenkinsfile
yrodiere: That will indeed work better :]

yrodiere:
> We'll need to backport this to all branches where we want to perform releases or publish snapshots...

Discussed elsewhere, but here's the conclusion. We (or rather @marko-bekhta :grimacing:) will backport:

* the removal of snapshot publishing in the main Jenkinsfile to all branches
* the whole patch (except maybe changes to `.md` files) to still-maintained branches

PR #4451, opened 2024-12-20
HSEARCH-5235 WIP: draft of injecting field references

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5235

----------------------

----------------------


mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/impl/AnnotationMappingConfigurationContextImpl.java
```diff
@@ -154,6 +157,8 @@ public void configure(MappingBuildContext buildContext, PojoMappingConfiguration
 					new PojoAnnotationTypeMetadataDiscoverer( contributorFactory, alreadyContributedTypes );
 			collector.collectDiscoverer( discoverer );
 		}
+
+		injectableBinderCollector.processDiscoveredBinders();
```
marko-bekhta, line 161: I've tried a few things and here's where I ended up so far 😃 

My thinking is that since we want these new binders as a way to have annotated properties so we can read that info in an AP, it seems that making them "available" through the programmatic mapping probably doesn't make that much sense... 
Hence, I thought that we could "detect" used binders when an annotation processor (current Search ones to get info from annotations) use a corresponding binder method ( which matches the case when a user adds a binder. So it'll include those coming from programmatic API too ... but IDK if we should just prohibit that use from the programmatic mapping or simply "fail nicely"). Then, once we "discovered" all annotations, we process the info we collected on binders by creating "something" that later will be able to inject required fields into a binder instance.

Since binder metadata (fields to inject) will not change depending on where the binder is used, we can have an instance per type rather than per occurrence. 

Then, the injection of fields happens here right before the bind method is called. We do have that delegating binder, which creates a bit of a problem since we cannot simply get our hands on the actual binder instance. I've worked around that for now by updating the delegating binder.

I've started by trying to reuse our annotation processors and basically building an "index model" for a binder but discarded that idea as it didn't work out that nicely 😄
Thinking more about it, for an annotation processor ... we won't be able to rely on user custom annotations and their processors anyway... so in this patch, I've just put together some simple processing to get the required data to create a field reference just for this injection part... thinking even more about it since we will be limiting the number of processors to a specific subset maaaybe I could make it work (reuse current processors at least partially).

Speaking of custom annotations... that's something that also won't play nice with the static metamodel, I suppose, unless we also do some similar "injection of field references" there, too .... 


Does that sound to you like the direction to keep exploring @yrodiere ? or should I revisit the approach 😃 🙈

marko-bekhta: Oh yeah sure 😃 I only wanted to put those few notes so I don't forget about them 🙈 let's definitely leave this discussion for the new year 😃 

> I think.we talked about some dynamism involving an expression language, which might change a few things such as analyzers based on where the binder is used ?

Right.. we did. I think in that case the dynamic part will come from the parameters, which we will have at the binding time, so my plan so far, was to pass those to the "injector" (at binding time) and it will do the work. But yes - let's get back to that next year 😃

marko-bekhta: And thank you for the answers 😃

mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/mapping/definition/annotation/impl/AnnotationPojoInjectableBinderCollector.java
```diff
@@ -0,0 +1,251 @@
...
+	}
+
+	private static class KeywordFieldInjectableBinderAnnotationProcessor
+			implements InjectableBinderAnnotationProcessor<KeywordField> {
+
+		static final KeywordFieldInjectableBinderAnnotationProcessor INSTANCE =
+				new KeywordFieldInjectableBinderAnnotationProcessor();
+
+		@Override
+		public InjectableBinderFieldInjector process(PojoInjectablePropertyModel<?> property, KeywordField annotation) {
```
yrodiere, line 200: Note the fact that you're manipulating a property model here means we won't be able to use this in javac annotation processors -- because I doubt we can implement `PojoPropertyModel` on top of annotation processor APIs. I'd love to be proven wrong though.
The javac annotation processor code that infers the index model from annotations will likely be completely separate...

PR #4564, opened 2025-04-10
HSEARCH-5351 Switch to JReleaser for nexus publishing

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5351

and the script for snapshots publishing through jreleaser: https://github.com/hibernate/hibernate-release-scripts/pull/39

----------------------

----------------------

### DIFF
--- a/ci/release/Jenkinsfile
+++ b/ci/release/Jenkinsfile
@@ -60,6 +60,7 @@ pipeline {
 
 					def releaseVersion = Version.parseReleaseVersion(params.RELEASE_VERSION)
 					def developmentVersion = Version.parseDevelopmentVersion(params.DEVELOPMENT_VERSION)
+					env.JRELEASER_DRY_RUN = params.RELEASE_DRY_RUN
 					echo "Performing full release for version ${releaseVersion.toString()}"
 
 					withMaven(mavenSettingsConfig: params.RELEASE_DRY_RUN ? null : 'ci-hibernate.deploy.settings.maven',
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -7,6 +7,7 @@
 # Our build scripts - necessary for our release jobs
 hibernate-noorm-release-scripts
 hibernate-release-scripts
+jreleaser
```
yrodiere, line 10: This is probably no longer needed, since the scripts now "install" jreleaser in `hibernate-release-scripts`?

jreleaser.yml
```diff
@@ -0,0 +1,62 @@
...
+# Deploy JARs and POMs to Maven Central
+deploy:
+  maven:
+    nexus2:
+      maven-central:
+        active: RELEASE
+        url: https://oss.sonatype.org/service/local
+        snapshotUrl: https://oss.sonatype.org/content/repositories/snapshots/
+        closeRepository: true
+        releaseRepository: false
```
yrodiere, line 27: So we want to keep logging into the Nexus repository web UI to click the "release" button?

Fine by me, but I wonder: is there an equivalent for the Maven Central publishing API? If not, we'll probably want to make sure we don't rely on this manual step...

jreleaser.yml
```diff
@@ -0,0 +1,62 @@
...
+      maven-central:
+        # TODO: HSEARCH-5354 Change to RELEASE once switching to Maven-Central:
+        active: NEVER
+        url: https://central.sonatype.com/api/v1/publisher
+        snapshotSupported: false
+        applyMavenCentralRules: true
+        stagingRepositories:
+          - target/staging-deploy/maven
+        # Deployment identifier used for publication.
+        # deploymentId: dd9991b0-18a7-41e7-b1fe-37b8ea936f85
```
yrodiere, line 50: This is not a secret?

ci/release/Jenkinsfile
```diff
@@ -62,13 +62,19 @@ pipeline {
 					def developmentVersion = Version.parseDevelopmentVersion(params.DEVELOPMENT_VERSION)
 					echo "Performing full release for version ${releaseVersion.toString()}"
 
-					withMaven(mavenSettingsConfig: params.RELEASE_DRY_RUN ? null : 'ci-hibernate.deploy.settings.maven',
-							mavenLocalRepo: env.WORKSPACE_TMP + '/.m2repository') {
+					withMaven(mavenSettingsConfig: null, mavenLocalRepo: env.WORKSPACE_TMP + '/.m2repository') {
```
yrodiere, line 65: The goal of the custom Maven settings here was also to override the default, which uses our Nexus mirror. We probably don't want to use the Nexus mirror during releases, and use "actual" dependencies downloaded from Maven central instead. Just in case our cache was poisoned somehow (mistakes happen).

See also how we use the custom Maven settings in snapshot publishing.

Note that we indeed don't need the credentials in Maven settings anymore. We should probably remove them from these maven settings on Jenkins once we switched to JReleaser everywhere.

yrodiere (approved):
I added a few comments, but LGTM overall. Thanks!

jreleaser.yml
marko-bekhta: the idea here was not to release the repository while testing the migration, as for the future ... we can flip the flag for maintenance branches, as for the main one... idk... if we aren't adding new artifacts then it's probably fine to release the repo automatically, but if we are .. it'll probably be safer to go check the staged bundle before releasing, no?

as for the new APIs, I think they have some alternative to this: 

> The MavenCentral Maven deployer allows executing deployments in stages that is, you may set the stage property to UPLOAD, PUBLISH, and/or FULL

https://jreleaser.org/guide/latest/reference/deploy/maven/maven-central.html#_staged_deployments

but ... you know, I haven't tested this yet 😄 
and while I was writing the comment, I thought I can (probably) try it on a local nexus, unless it's a Pro  feature ...

jreleaser.yml
marko-bekhta: tbh ... I'm not sure if this one will even be required, I only copied it from the docs, and it is marked as optional. I think that it might only be needed if we do a release in multiple steps... but will have to test it to see 😃 

> The MavenCentral deployer requires the deploymentId property to be set when the stage is set to PUBLISH.

https://jreleaser.org/guide/latest/reference/deploy/maven/maven-central.html#_staged_deployments

PR #4570, opened 2025-04-15
HSEARCH-5358 Fail faster if an enclosing instance parameter is encountered in a projection constructor on JDK 25+

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5358

----------------------

----------------------

### DIFF
--- a/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/search/definition/binding/impl/ProjectionConstructorParameterBinder.java
+++ b/mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/search/definition/binding/impl/ProjectionConstructorParameterBinder.java
@@ -56,7 +56,16 @@ class ProjectionConstructorParameterBinder<P> implements EventContextProvider {
 			// and it's often useful to be able to declare a method-local type for projections
 			// (those types have a "surrounding instance" parameter in their constructor
 			// even if they don't use it).
-			return ConstantProjectionDefinition.nullValue();
+
+			// NOTE: with JDK 25+ this is no longer the case,
+			// and will fail further at runtime when such projection is created.
+			// Hence, let's fails faster instead;
+			if ( parameter.enclosingInstanceCanBeNull() ) {
+				return ConstantProjectionDefinition.nullValue();
+			}
+			else {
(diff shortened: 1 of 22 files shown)


mapper/pojo-base/src/main/java/org/hibernate/search/mapper/pojo/search/definition/binding/impl/ProjectionConstructorParameterBinder.java
```diff
@@ -56,7 +56,16 @@ BeanHolder<? extends ProjectionDefinition<?>> bind() {
...
+
+			// NOTE: with JDK 25+ this is no longer the case,
+			// and will fail further at runtime when such projection is created.
+			// Hence, let's fails faster instead;
+			if ( parameter.enclosingInstanceCanBeNull() ) {
+				return ConstantProjectionDefinition.nullValue();
+			}
+			else {
+				throw ProjectionLog.INSTANCE.nullEnclosingParameterInProjectionConstructorNotAllowed( eventContext() );
+			}
```
marko-bekhta, line 68: @yrodiere this one might be of interest to you once you are back 😃 
(https://jdk.java.net/25/release-notes#JDK-8164714)

yrodiere: Makes sense. So... all our tests using method-local types for projections are now failing on JDK 25+?

marko-bekhta: well... they did ... so I moved them out to be static inner classes 🙈 . I kept 1-2 with `assertTrue` to skip on JDK 25 so we test those method-local scenarios while they still work.

yrodiere: Wonderful. I was about to ask for a link to the PR, but on second thought I'd rather not see that :grin: 

Sorry for the technical debt, thanks for cleaning it up :)

PR #4618, opened 2025-05-16
HSEARCH-**** Attempt to make the migration simpler

marko-bekhta:
[Please describe here what your change is about]

----------------------

----------------------


engine/src/main/java/org/hibernate/search/engine/search/projection/definition/ProjectionDefinition.java
```diff
@@ -31,13 +31,13 @@ public interface ProjectionDefinition<P> {
...
+	 * {@link ProjectionDefinition#create(TypedSearchProjectionFactory, ProjectionDefinitionContext)} returns.
 	 * @param context The context in which the definition is applied.
 	 * @return The created {@link SearchPredicate}.
 	 * @throws RuntimeException If the creation of the projection fails.
-	 * @see SearchPredicateFactory
+	 * @see TypedSearchPredicateFactory
 	 * @see ProjectionDefinitionContext
 	 */
-	SearchProjection<? extends P> create(SearchProjectionFactory<?, ?, ?> factory, ProjectionDefinitionContext context);
+	SearchProjection<? extends P> create(TypedSearchProjectionFactory<?, ?, ?> factory, ProjectionDefinitionContext context);
```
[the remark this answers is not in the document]
yrodiere: It seems to me that in order for type-safe factories to be useful in these definitions, we'd need the implementor of the definition to be able to check/require a specific root type, no? E.g.:

```java
public class BookAuthorNamesProjectionDef implements ProjectionDefinition<Book__, P> {
		@Override
		public SearchProjection<List<String>> create(SearchProjectionFactory<Book__, ?, ?> factory,
				ProjectionDefinitionContext context) {
			return factory.field( Book__.authors.name )
					.collector( ProjectionCollector.list() ) // <4>
					.toProjection();
		}
}
```

IMO, either this ^ can be made to work, or typed scoped can be ignored here.

Either way... this is incubating API. I'd say it's fine to do whatever is easier here, and create a follow-up issue?

engine/src/main/java/org/hibernate/search/engine/search/predicate/definition/TypedPredicateDefinition.java
```diff
@@ -0,0 +1,33 @@
...
+public interface TypedPredicateDefinition<SR> {
+
+	/**
+	 * Creates a predicate.
+	 * @param context The context, exposing in particular a {@link TypedSearchPredicateFactory}.
+	 * @return The created {@link SearchPredicate}.
+	 */
+	SearchPredicate create(TypedPredicateDefinitionContext<SR> context);
+
+}
```
marko-bekhta, line 33: Hey @yrodiere 

soooo here I've tried to explore what can be done about the predicate/projection definitions... 
I didn't extend `TypedPredicateDefinition` with `PredicateDefinition` or vice-versa as that makes things more cumbersome .... since they expect the context (which has a factory either typed or not) we'd have to add the context type to the type arguments and that makes defining the typed one as:

```java
new TypedPredicateDefinition<SomeScope, TypedPredicateDefinitionContext<SomeScope>> {
   SearchPredicate create(TypedPredicateDefinitionContext<SomeScope> context) {
   ...
   }
}

with this approach there's some ambiguity when creating these definitions as lambdas as it's unclear which context user wants in `root.namedPredicate(name definition)`. Having a `root.namedTypedPredicate()` ... didn't look nice to me 😄.

and then in the actual impl... since this `SR` is only in the type definition and we never pass it anywhere... we can just do an "ugly-unchecked-cast".
I wonder if we should keep the scope type at runtime and make additional checks, or if it's not worth it, as the main point here was to have a compile-level check? WDYT?

Comment (yrodiere) [reply engine/src/main/java/org/hibernate/search/engine/search/predicate/definition/TypedPredicateDefinition.java:33]: I think additional runtime checks would make sense, if possible.

I also think having the definition context type as a generic type parameter is a bit odd.

And... I think it's fine to leave this for later, too, as it's incubating API that you can break and break again :)

File: backend/elasticsearch/src/main/java/org/hibernate/search/backend/elasticsearch/search/predicate/dsl/impl/ElasticsearchSearchPredicateFactoryImpl.java [hunk 3]
```diff
@@ -21,16 +23,36 @@ public class ElasticsearchSearchPredicateFactoryImpl<SR>
...
+		// e.g. we have one tied to a Book__ and we want to use some "raw" string paths in a named predicate.
+		scopeRootType.equals( NonStaticMetamodelScope.class )
+				// scope type is in the same hierarchy:
+				|| this.scopeRootType.isAssignableFrom( scopeRootType )
+				|| scopeRootType.isAssignableFrom( this.scopeRootType )
+		) {
+			return new ElasticsearchSearchPredicateFactoryImpl<>( scopeRootType, dslContext );
+		}
+		throw QueryLog.INSTANCE.incompatibleScopeRootType( scopeRootType, this.scopeRootType );
+	}
```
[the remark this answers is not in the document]
yrodiere: I'm not entirely sure how safe downcasting would be, but... whatever, if it's incubating, let's try.

engine/src/main/java/org/hibernate/search/engine/search/predicate/dsl/ExtendedSearchPredicateFactory.java
```diff
@@ -15,9 +18,14 @@
  * @param <S> The self type, i.e. the exposed type of this factory.
  */
 public interface ExtendedSearchPredicateFactory<SR, S extends ExtendedSearchPredicateFactory<SR, ?>>
-		extends SearchPredicateFactory<SR> {
+		extends TypedSearchPredicateFactory<SR> {
```
yrodiere, line 21: So IIUC, it's no longer possible to pass an `ElasticsearchSearchPredicateFactory` to a method expecting a `SearchPredicateFactory`? E.g. a method like the one in [this example of the documentation](https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#search-dsl-paths-relative)? 

```
private SearchPredicate matchFirstAndLastName(SearchPredicateFactory f,
        String firstName, String lastName) {
    return f.and(
            f.match().field( "firstName" ) 
                    .matching( firstName ),
            f.match().field( "lastName" )
                    .matching( lastName )
    )
            .toPredicate();
}
```

It's fine if this is a voluntary limitation, i.e. if it's a compromise you found between complexity and compatibility. But I thought you should be aware :)

yrodiere (approved):

marko-bekhta:
superseded by https://github.com/hibernate/hibernate-search/pull/4623

PR #4712, opened 2025-07-18
HSEARCH-5356 Introduce the extended (platform) BOM

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5356

----------------------

----------------------


marko-bekhta:
hey @yrodiere 🙂 👋🏻 if you have the time ... I'd like your opinion on what's included in these platform poms 🫣 🙂

pom.xml
```diff
@@ -158,6 +158,8 @@
 
     <modules>
         <module>bom/public</module>
+        <module>bom/platform</module>
+        <module>bom/platform-next</module>
```
yrodiere, line 162: Maven 4 wording would be `bom/skinny` vs. `bom/fat`, or something like that. Though I'm sure someone will get offended at the use of the word "fat" :roll_eyes: 

Did you look into conventions before going with the `platform` name? And yes, I know that's what Hibernate ORM uses.

pom.xml
```diff
@@ -315,6 +317,34 @@
         <version.spotless-maven-plugin>2.45.0</version.spotless-maven-plugin>
         <version.maven-antrun-plugin>3.1.0</version.maven-antrun-plugin>
 
+        <!-- >>> Version of dependencies that we need to keep in the root to share between build and "bom" pom files -->
```
yrodiere, line 320: Question: if we import the platform BOM into the skinny BOM, would the "flattening" be able to only keep the dependencies explicitly referenced in the skinny BOM, applying versions coming from the platform BOM?

build/enforcer/src/main/java/org/hibernate/search/build/enforcer/DependencyManagementIncludesAllGroupIdArtifactsRule.java
```diff
@@ -0,0 +1,161 @@
...
+import org.apache.maven.execution.MavenSession;
+import org.apache.maven.model.Dependency;
+
+@Named("dependencyManagementIncludesAllGroupIdArtifactsRule") // rule name - must start with lowercase character
+public class DependencyManagementIncludesAllGroupIdArtifactsRule extends AbstractEnforcerRule {
+	/**
+	 * See <a href="https://central.sonatype.org/search/rest-api-guide/">Maven Central REST API</a>
+	 */
+	private static final String BASE_URL_FORMAT =
+			"https://search.maven.org/solrsearch/select?q=%s&core=gav&start=%d&rows=%d&wt=json";
```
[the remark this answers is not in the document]
yrodiere: > so maybe a warning if the API fails and then it would mean to be more careful with the updates and look if the rule works or not

+1 

> *My understading was that we want to "manage" more than we currently use so that if someone pulls an extra dependency for their needs the version would be compatible with what our artifacts use

Makes sense, though I'd really prefer relying on a Lucene BOM for that. But I guess there isn't any :/

bom/platform-next/pom.xml
```diff
@@ -0,0 +1,352 @@
...
+                <artifactId>byte-buddy-android-test</artifactId>
+                <version>${version.net.bytebuddy}</version>
+            </dependency>
+            <dependency>
+                <groupId>net.bytebuddy</groupId>
+                <artifactId>byte-buddy-gradle-plugin</artifactId>
+                <version>${version.net.bytebuddy}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.agroal</groupId>
```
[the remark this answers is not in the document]
yrodiere: Agreed. If a dependency version doesn't matter when using just Hibernate Search and its dependencies, without adding optional Hibernate ORM artifacts, well... there's no reason to put it in the Hibernate Search platform.

So I think your rule should be more something like:

* Manage every Hibernate Search artifact
* Manage every dependency of those, transitively
* Manage every artifact in the same groupId (so, all ORM artifacts, all Lucene artifacts)
* Do _not_ manage transitive dependencies of that last item (so, no Agroal, no optional Lucene dependency)

bom/platform-next/pom.xml
```diff
@@ -0,0 +1,1206 @@
...
+        <version.bom.jakarta.xml.bind>4.0.2</version.bom.jakarta.xml.bind>
+        <version.bom.org.jberet>3.1.0.Final</version.bom.org.jberet>
+    </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <!--
+                Hibernate Search artifacts listed explicitly as we cannot exclude some from the bom import,
+                and we want to skip the lucene-next ones for this pom to not mix different Lucene versions
+            -->
```
yrodiere, line 47: Makes me think you probably need a `hibernate-search-next-bom` too?

bom/platform-next/pom.xml
```diff
@@ -0,0 +1,1206 @@
...
+                <artifactId>third-party-slf4j-api</artifactId>
+                <version>${version.bom.software.amazon.awssdk}</version>
+            </dependency>
+            <dependency>
+                <groupId>software.amazon.awssdk</groupId>
+                <artifactId>third-party</artifactId>
+                <version>${version.bom.software.amazon.awssdk}</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
```
yrodiere, line 870: Would it make sense to avoid repetition by either:

1. Importing the normal BOM and overriding just lucene dependencies? But then you'd be managing `hibernate-search-backend-lucene` (not next) even though it can't work with the managed version of Lucene.
2. Importing a separate POM dedicated to just dependencies, not Hibernate Search artifacts, in _both_ platform poms, and overriding the Lucene version here?

yrodiere (commented):
LGTM, thanks. Some ideas below, but feel free to hate/ignore :)

bom/platform-next/pom.xml
marko-bekhta: This one seems to work unexpectedly ok 😃 
Since it does not manage Lucene directly each backend transitevely brings the version it wants as long as you do not use two lucene backends at the same time 😃

bom/platform-next/pom.xml
marko-bekhta: > Importing a separate POM dedicated to just dependencies, not Hibernate Search artifacts, in both platform poms, and overriding the Lucene version here?

this one would be nice yeah ... but I guess if we'd make some bom/platform-common we'd need to either publish it as well ... or don't publish but inline all the dependencies... 

let me pull the common part and see if it would look better that way, and if so -- I'll just inline the dependencies, after all ... it is to manage dependencies and not to be read by a human on a daily basis 😁 🙈

PR #4893, opened 2025-11-12
DRAFT: Switch from the Search session holder to an ORM session extension

marko-bekhta:
Based on https://github.com/hibernate/hibernate-orm/pull/11093

----------------------

----------------------


mapper/orm/src/main/java/org/hibernate/search/mapper/orm/session/impl/HibernateOrmSearchSessionHolder.java
```diff
@@ -1,111 +0,0 @@
...
-		if ( holder != null ) {
-			return holder;
-		}
-		boolean closedSessionAndTransaction = session.isClosed() && session.isTransactionInProgress();
-		if ( closedSessionAndTransaction ) {
-			// This can happen when using JTA.
-			// Only do this as a fallback: if somehow the session holder was added to the session properties
-			// before the session was closed, we definitely want to use it and avoid the static map.
-			holder = holderPerClosedSessionTransaction.get( session.accessTransaction() );
-		}
```
marko-bekhta: Hi @yrodiere 👋🏻 🙂 

I've tried to integrate the ORM extension, and it made me wonder if this ^ could even happen with holder being attached to a session as soon as it (session) is created.
The tests passed without this static map.. but ... I'm not sure whether we actualy have any tests to hit this scenario, or we do?

yrodiere: Hey. It's been a long time... but I don't see a reason this would still be needed if the holder is added when creating the session.

Also, the PR that added this code seems to be specifically about closed sessions: https://github.com/hibernate/hibernate-search/pull/2450
So if the tests from that PR are still around, and still pass... +1 to remove all this code?

mapper/orm/src/main/java/org/hibernate/search/mapper/orm/session/impl/HibernateOrmSearchSession.java
```diff
@@ -213,6 +220,11 @@ public Session toOrmSession() {
 		return sessionImplementor;
 	}
 
+	@Override
+	public SessionImplementor session() {
+		return sessionImplementor;
+	}
+
```
marko-bekhta: we have these publicly exposed ... so I wonder what should we do about them ... (once we start the stateless session work...)

marko-bekhta:
* superseded by https://github.com/hibernate/hibernate-search/pull/5137

PR #5117, opened 2026-05-04
HSEARCH-5614 Enable automated releases

marko-bekhta:
https://hibernate.atlassian.net/browse/HSEARCH-5614

----------------------

----------------------

### DIFF
--- a/ci/release/Jenkinsfile
+++ b/ci/release/Jenkinsfile
@@ -7,10 +7,49 @@
 
 import org.hibernate.jenkins.pipeline.helpers.version.Version
 
+env.PROJECT = "search"
+// Set to `true` *only* on branches where you want a scheduled release.
+// Also uncomment the `FULLY_AUTOMATIC` property and related steps.
+def RELEASE_ON_SCHEDULE = false
+
+// Avoid running the pipeline on branch indexing
+if (currentBuild.getBuildCauses().toString().contains('BranchIndexingCause')) {
+	print "INFO: Build skipped due to trigger being Branch Indexing"
+	currentBuild.result = 'NOT_BUILT'
+	return
+}
+
+def manualRelease = currentBuild.getBuildCauses().toString().contains( 'UserIdCause' )
+def cronRelease = currentBuild.getBuildCauses().toString().contains( 'TimerTriggerCause' )
+
+// Only do automatic release on branches where we opted in
+if ( !manualRelease && !cronRelease ) {
+	print "INFO: Build skipped because automated releases on push are disabled on this branch."
+	currentBuild.result = 'NOT_BUILT'
+	return
+}
+
+if ( !manualRelease && cronRelease && !RELEASE_ON_SCHEDULE ) {
+	print "INFO: Build skipped because automated releases are disabled on this branch. See constant RELEASE_ON_SCHEDULE in ci/release/Jenkinsfile"
+	currentBuild.result = 'NOT_BUILT'
+	return
+}
+
+def checkoutReleaseScripts() {
+	dir('.release/scripts') {
+		checkout scmGit(branches: [[name: '*/main']], extensions: [],
+				userRemoteConfigs: [[credentialsId: 'ed25519.Hibernate-CI.github.com',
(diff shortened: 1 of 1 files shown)


ci/release/Jenkinsfile
```diff
@@ -90,6 +210,46 @@ pipeline {
...
+                script {
+                    checkoutReleaseScripts()
+
+                    configFileProvider([
+                            configFile(fileId: 'release.config.ssh', targetLocation: "${env.HOME}/.ssh/config"),
+                            configFile(fileId: 'release.config.ssh.knownhosts', targetLocation: "${env.HOME}/.ssh/known_hosts")
+                    ]) {
+                        withCredentials([
+                                gitUsernamePassword(credentialsId: 'username-and-token.Hibernate-CI.github.com', gitToolName: 'Default')
+                        ]) {
```
marko-bekhta: hey @yrodiere 👋🏻 🙂 

I've copied this part from ORM... but it makes me wonder... we are using the ssh key for git operations, was there a reason we also need to wrap that call in a withcredentials+GH token ? or maybe we can just remove this part and have the ssh agent only ? 🤔

yrodiere: Hey. I don't know. Try it? :)

yrodiere (commented):
Forgot to press "submit review"... sorry. Especially since my answer isn't very useful.

ci/release/Jenkinsfile
marko-bekhta:
```
+ ../scripts/website-release.sh null search 8.3.1.Final
[detached HEAD 0768c471] search
 1 file changed, 1 insertion(+)
 create mode 100644 _data/projects/null/releases/search/search.yml
Attempt 1 of 5: Pushing changes...
fatal: could not read Username for 'https://github.com/': No such device or address
Push failed. Waiting for 5 seconds before trying again...
Rebasing and will try again...
From https://github.com/hibernate/hibernate.org
 * branch              production -> FETCH_HEAD
HEAD is up to date.
Attempt 2 of 5: Pushing changes...
fatal: could not read Username for 'https://github.com/': No such device or address
Push failed. Waiting for 5 seconds before trying again...
Rebasing and will try again...
From https://github.com/hibernate/hibernate.org
 * branch              production -> FETCH_HEAD
 ```
 
 sooo I think we got the answer 😁 
 
 but it's probably because we are missing 
 ```
 git config --local user.name "Hibernate CI"
git config --local user.email "ci@hibernate.org"
```
+the url in the checkout is an http one and not the ssh ... I'll try making some changes ... and we can see how it goes with the next release 🙂

yrodiere: > but it's probably because we are missing
> 
> ```
> git config --local user.name "Hibernate CI"
> git config --local user.email "ci@hibernate.org"
> ```

That part should be configured as part of the CI job. Or at least I think it's possible and I'd very much prefer us to -- because changing the config like this makes a mess when testing scripts locally.

yrodiere: > the url in the checkout is an http one and not the ssh

That's also configurable in the CI job; there is an advanced property along the lines of "Check out using SSH"

yrodiere: I just checked an both are set on the Hibernate Search release job

marko-bekhta: thanks! right ... so we just need to fix the url here then 🤞🏻 :
https://github.com/hibernate/hibernate-search/blob/ab00f13d1856f5a47648e10a802ec63ed5fbc8fa/ci/release/Jenkinsfile#L237-L241
with something like
```groovy
checkout scmGit(
    branches: [[name: '*/production']],
    extensions: [],
    userRemoteConfigs: [[
        credentialsId: 'ed25519.Hibernate-CI.github.com', 
        url: 'git@github.com:hibernate/hibernate.org.git'
    ]]
)
```

marko-bekhta: (and do that for other projects too ... 🙂)

yrodiere: Ah... right. You might have to set the git username/email there, too; I'm not sure if the CI job config is global or local.

